PR #1, opened 2018-05-09
Sync patients from users to server

govindkrjoshi:
Issue: https://github.com/resolvetosavelives/redapp/issues/18

This PR should not be merged as is, is only here for visibility

- First cut of syncing patients from users
- Sending back data to users isn't done yet
- The merging logic itself is fairly generically applicable to all entities. We'll refine this with time.

-- 
Srihari and Govind


app/controllers/application_controller.rb
```diff
@@ -1,3 +1,2 @@
 class ApplicationController < ActionController::Base
-  protect_from_forgery with: :exception
```
timcheadle: We will need this. I expect this app to serve an admin interface so I don't want to strip out CSRF protection.

app/helpers/merge_record.rb
```diff
@@ -0,0 +1,22 @@
...
+      model.create(where_clauses.merge(record))
+    else
+      if record[:updated_at] > existing_record.updated_at
+        existing_record.update(record)
+      end
+      existing_record
+    end
+  end
+
+  def self.bulk_merge_on_id(model, records)
```
timcheadle: I think calling it `records` is confusing here when really it's an array of param hashes.

app/helpers/merge_record.rb
```diff
@@ -0,0 +1,22 @@
+module MergeRecord
+
+  # make this an ActiveSupport::Concern ?
+
+  def self.merge(model, where_clauses, record)
+    existing_record = model.find_by(where_clauses)
+    if existing_record.nil?
+      model.create(where_clauses.merge(record))
+    else
+      if record[:updated_at] > existing_record.updated_at
```
timcheadle: If `record[:updated_at]` is nil, this crashes.

Maybe use this:
```ruby
if record.has_key?(:updated_at) && record[:updated_at] > existing_record.updated_at
```

app/helpers/merge_record.rb
```diff
@@ -0,0 +1,22 @@
+module MergeRecord
+
+  # make this an ActiveSupport::Concern ?
+
+  def self.merge(model, where_clauses, record)
```
timcheadle: What happens when one of the param hashes contains invalid data and `update` fails?

config/application.rb
```diff
@@ -26,6 +26,11 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded.
 
     # Don't generate system test files.
-    config.generators.system_tests = nil
+    config.generators do |g|
+      g.system_tests false
+      g.assets false
+      g.helper false
+      g.view false
```
timcheadle: Do we want to disable view generators? I can see us using those for the admin UI.

config/routes.rb
```diff
@@ -1,3 +1,11 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+
+  namespace :api do
+    namespace :v1 do
+      resource :patients, only: [] do
```
timcheadle: This feels weird. Why not `namespace :patients` here if you aren't using any resources?

spec/controllers/api/v1/patients_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.describe Api::V1::PatientsController, type: :controller do
+  describe "POST sync" do
+
+    let(:patients) {{patients: FactoryBot.attributes_for_list(:patient, 10)}}
```
timcheadle: Add whitespace after this; delete it before this

app/controllers/api/v1/patients_controller.rb
```diff
@@ -0,0 +1,19 @@
+class Api::V1::PatientsController < ApplicationController
+  def sync_from_user
+    MergeRecord.bulk_merge_on_id(Patient, patient_params)
+    render json: nil, status: :ok
+  end
+
+  def sync_to_user
```
timcheadle: Let's avoid using placeholder methods that do nothing.

spec/spec_helper.rb
```diff
@@ -1,96 +1,17 @@
...
-  # Many RSpec users commonly either run the entire suite or an individual
-  # file, and it's useful to allow more verbose output when running an
-  # individual spec file.
-  if config.files_to_run.one?
-    # Use the documentation formatter for detailed output,
-    # unless a formatter has already been configured
-    # (e.g. via a command-line flag).
-    config.default_formatter = "doc"
+  config.after :each do
+    DatabaseCleaner.clean_with(:truncation, reset_ids: true)
```
timcheadle: Why do this here instead of just using transactions?

spec/controllers/api/v1/patients_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    it "creates new patients" do
+      post :sync_from_user, params: patients
+      expect(Patient.count).to eq 10
+      expect(response).to have_http_status(200)
+    end
+
+    it "updates the existing patients" do
+      existing_patients = FactoryBot.create_list(:patient, 10)
+      updated_patients = existing_patients.take(5).each do |p|
+        FactoryBot.attributes_for(
```
timcheadle: Shouldn't this actually update the record `p`? Also, please don't use one-letter variables; make it `patient` instead for readability.

spec/factories/patients.rb
```diff
@@ -0,0 +1,14 @@
+FactoryBot.define do
+  factory :patient do
+    common_names = { female: %w[anjali divya ishita priya priyanka riya shreya tanvi tanya vani],
```
timcheadle: Let's try and abstract these somewhere for re-use. I can see us wanting names in the `User` model as well.

db/migrate/20180509061518_create_patients.rb
```diff
@@ -0,0 +1,11 @@
+class CreatePatients < ActiveRecord::Migration[5.1]
+  def change
+    create_table :patients, id: false do |t|
+      t.uuid :id, primary_key: true
```
timcheadle: What's the reason for using UUID instead of integers?

timcheadle:
Looks promising overall. Needs unit tests for the merge helper.

config/application.rb
govindkrjoshi: We don't think we'll need the view generators for the near future, given we'll mostly be working on the APIs. We can add views manually, or add this generator back when we need it.

config/routes.rb
govindkrjoshi: Fixed in aa7a18d8b98d3d400cb2b081e3e3a5ab2a3ea315.

app/controllers/api/v1/patients_controller.rb
govindkrjoshi: Just wanted to leave that specific one there so we don't forget the reasoning for the method names :) . Removed it now.
-- Srihari and Govind

spec/spec_helper.rb
govindkrjoshi: We mostly just defaulted to what we've used previously. But since this is just one line, we can use transactions. Will change this.

spec/controllers/api/v1/patients_controller_spec.rb
govindkrjoshi: Hrm, ya. This might be our clojure/haskell practices rubbing off on us, will be wary. This is fixed though.

db/migrate/20180509061518_create_patients.rb
govindkrjoshi: All the records are created on devices, and sync'd across devices. Using UUIDs makes it easy to track the same patient across devices. Using serial ids would inevitably cause collisions. For performance reasons, UUIDs are stored and treated like large numbers, and should be fairly performant.

We aren't actually sure how this would work using only integers. Did you have something in mind?

-- Srihari and Govind

config/application.rb
timcheadle: I'm ok with that.

db/migrate/20180509061518_create_patients.rb
timcheadle: Good point; I wasn't considering collisions.

spec/spec_helper.rb
timcheadle: I don't think we need DatabaseCleaner for this project yet. I've used that in the past when dealing with Capybara and Javascript headless browser testing, but we won't yet be doing that here.

app/models/phone_number.rb
```diff
@@ -0,0 +1,5 @@
+class PhoneNumber < ApplicationRecord
+  PHONE_TYPE = %w[mobile landline].freeze
+  validates_presence_of :number, :created_at, :updated_at
+  has_and_belongs_to_many :patients
```
timcheadle: This should be a `has_many :through` relationship, as we may want to flag phone numbers as inactive for a certain patient.

Also, HABTM relationships get messy quickly. I strongly prefer having an explicit join table.

app/models/patient.rb
```diff
@@ -0,0 +1,18 @@
+class Patient < ApplicationRecord
+  GENDERS = %w[male female transgender].freeze
+  STATUSES = %w[active dead migrated unresponsive inactive]
+
+  belongs_to :address, optional: true
+  has_and_belongs_to_many :phone_numbers
```
timcheadle: Let's make this a `has_many :through`.

app/models/patient.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  validates_presence_of :created_at, :updated_at
```
timcheadle: This will cause us to manually set these fields before we can save any new records, not ideal.

db/migrate/20180510090438_create_addresses.rb
```diff
@@ -0,0 +1,22 @@
...
+  def change
+    create_table :addresses, id: false do |t|
+      t.uuid :id, primary_key: true
+      t.string :street_address
+      t.string :colony
+      t.string :village
+      t.string :district
+      t.string :state
+      t.string :country
+      t.string :pin
```
timcheadle: This all is very India-specific. We should model these fields to be more generic (`postal_code` for example). Let's discuss.

spec/spec_helper.rb
ssrihari: Fixed in f7f62fca8d98972ae80d8524da6c4eab4402bf55

app/controllers/application_controller.rb
ssrihari: Fixed in 0f3bccfde3ad3757193cc0710d54a1d1ac0c8320

app/helpers/merge_record.rb
ssrihari: This method has since been removed. Also, `MergeRecord` now takes real records, not hashes.

app/models/patient.rb
ssrihari: These fields need to be set on the front-end devices, and are required params in the API. But we could perhaps look into naming these fields differently in the API and model.

app/controllers/api/v1/patients_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      errors
+    end
+
+    response = { errors: errors.nil? ? nil : errors }
+    render json: response, status: :ok
+  end
+
+  def sync_to_user
+    latest_record_timestamp = params[:latest_record_timestamp].to_time
+    patients_to_sync        = Patient.distinct
```
timcheadle: Let's avoid unnecessary alignment of equal signs; it makes too much whitespace.

app/controllers/api/v1/patients_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+                                .left_outer_joins(:address)
+                                .left_outer_joins(:phone_numbers)
+                                .where('patients.updated_on_server_at >= ?
+                                        or addresses.updated_on_server_at >= ?
```
timcheadle: Can we go with 4 space tabs?

app/services/merge_patient_service.rb
```diff
@@ -0,0 +1,33 @@
+class MergePatientService
```
timcheadle: I'm wondering if this shouldn't just live in the `Patient` model...

app/services/merge_patient_service.rb
```diff
@@ -0,0 +1,33 @@
...
+  private
+
+  attr_reader :patient_record
+
+  def merge_address(address_params)
+    address = Address.new(address_params)
+    MergeRecord.merge_by_id(address)
+  end
+
+  def merge_phone_number(phone_number_params)
```
timcheadle: I'd put this method below the plural form as it's called later

spec/api/spec.rb
```diff
@@ -0,0 +1,97 @@
+module Spec
+
+  ###############
+  # Models
+
+  def patient_spec
+    { type:       :object,
+      properties: { id:               { type: :string, format: :uuid },
```
timcheadle: Let's put hash keys on a new line after the `{` so we don't have so much left whitespace

spec/controllers/api/v1/patients_controller_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+    it 'creates new patients' do
+      patients = (1..10).map { build_patient }
+      post(:sync_from_user, params: { patients: patients })
+      expect(Patient.count).to eq 10
+      expect(Address.count).to eq 10
+      expect(PhoneNumber.count).to eq(patients.sum { |patient| patient['phone_numbers'].count })
+      expect(response).to have_http_status(200)
+    end
+
+    it 'creates new patients without address' do
```
timcheadle: Should we test for creating patients without a phone number?

spec/controllers/api/v1/patients_controller_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+      post(:sync_from_user, params: { patients: [build_patient.except('address')] })
+      expect(Patient.count).to eq 1
+      expect(Address.count).to eq 0
+      expect(response).to have_http_status(200)
+    end
+
+    it 'returns errors for invalid records' do
```
timcheadle: What should the HTTP status be?

spec/controllers/api/v1/patients_controller_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+        patients_hash['updated_on_server_at']            = 15.minutes.ago
+        patients_hash['address']['updated_on_server_at'] = 15.minutes.ago
+        patients_hash['phone_numbers'].each do |phone_number|
+          phone_number['updated_on_server_at'] = 15.minutes.ago
+        end
+        MergePatientService.new(patients_hash).merge
+      end
+    end
+
+    it 'Gets all the patients updated since last sync' do
```
timcheadle: s/Gets/returns/

app/helpers/merge_record.rb
```diff
@@ -0,0 +1,19 @@
+module MergeRecord
+  def self.merge(where_clauses, record)
+    return record if record.invalid?
```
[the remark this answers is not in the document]
ssrihari: Fixed in 5c5d8fc.

spec/controllers/api/v1/patients_controller_spec.rb
ssrihari: Yeah, added in 5c5d8fc.

app/models/patient.rb
```diff
@@ -0,0 +1,33 @@
...
+  STATUSES = %w[active dead migrated unresponsive inactive].freeze
+
+  belongs_to :address, optional: true
+  has_many :patient_phone_numbers
+  has_many :phone_numbers, through: :patient_phone_numbers
+
+  validates_presence_of :created_at, :updated_at
+  validates_inclusion_of :gender, in: GENDERS
+  validates_inclusion_of :status, in: STATUSES
+  validate :presence_of_age
```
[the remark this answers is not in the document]
ssrihari: Good catch! Thanks. We've added validation for full_name in 5c5d8fc.

app/models/patient.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def presence_of_age
+    unless date_of_birth.present? || age_when_created.present?
+      errors.add(:age, 'Either date_of_birth or age_when_created should be present')
+    end
+  end
+
+  def has_errors?
```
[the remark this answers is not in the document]
ssrihari: Thanks! Fixed in 5c5d8fc.

app/models/patient.rb
```diff
@@ -0,0 +1,33 @@
+class Patient < ApplicationRecord
+  GENDERS  = %w[male female transgender].freeze
+  STATUSES = %w[active dead migrated unresponsive inactive].freeze
+
+  belongs_to :address, optional: true
+  has_many :patient_phone_numbers
+  has_many :phone_numbers, through: :patient_phone_numbers
+
+  validates_presence_of :created_at, :updated_at
```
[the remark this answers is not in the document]
ssrihari: I think the dissonance here comes from the fact that created_at and updated_at mean different things in our app. Perhaps we should rename these to `device_created_at` and `device_updated_at` at the controller so it does not overload rails' conventional definitions of the fields?

Ideally, all of these validations belong at the API level. Perhaps we could move all of these into a Form Object? And I'd prefer to do these as a part of the next PR if that's okay, this PR has been around for a long time.

app/services/merge_patient_service.rb
ssrihari: Unsure about this right now. Since there's a lot of room for reuse of this kind of code across models, perhaps we can change this design after we implement sync for BP, and Users as well.

ssrihari:
This PR is now ready for merge. @timcheadle

ssrihari:
@timcheadle merging this. We've addressed almost all the comments. There's a couple of things pending: renaming `created_at`, `updated_at` and use of the form object pattern. We'll do that in the PRs to come, since they are more widely applicable.

PR #9, opened 2018-05-18
Sync patients to user

ssrihari:
Issue: https://github.com/resolvetosavelives/redapp/issues/16

- [x] `first_time: true` should return records from beginning of time
- [x] use `latest_record_timestamp` from request to compute records to send
- [x] update patient's `latest_record_timestamp` when it's address or phone_number is updated
- [x] send `latest_record_timestamp` with the response
- [x] accept `number_of_records` as "batch size" from the request, send only those many records
- [x] add API documentation
- [x] setup config (dotenv?)
- [x] test push + pull over http
- [x] document caveats of design choices: sequencing using timestamps

Things to do/refactor [not in this PR]:

- use form-object pattern to validate and process patient sync requests
- rename `updated_at` and `created_at` for all entities within backend
- we need to store `age`, and `age_updated_at`, to compute the current age.
- `patient_phone_numbers` should be one to many, not many to many
- merge the colony and village fields in address


ssrihari:
@timcheadle how would you prefer to do configs, normally? Would you prefer we use `dotenv`? We have a `default_batch_size` we need to define somewhere. Technically, this is a constant that can be changed for an environment.

timcheadle:
@ssrihari `dotenv` for sure. Usually I like to set up a `.env.development` that's checked into the repo and use `.env.production` for production. We'll make sure that `.env` and `.env.production` are in `.gitignore`

ssrihari:
@timcheadle this PR is ready for review and merge.

app/helpers/mergeable_record.rb
```diff
@@ -0,0 +1,43 @@
+class MergeableRecord
```
timcheadle: This feels like it should be implemented as a "mergeable" model concern instead of a class you pass a model into.

app/controllers/api/v1/patients_controller.rb
```diff
@@ -27,8 +42,24 @@ def patients_params
...
         phone_numbers: [permitted_phone_number_params],
-        address: permitted_address_params
+        address:       permitted_address_params
       )
     end
   end
-end
+
+  def latest_record_timestamp
+    if params[:first_time].present? && params[:first_time] == 'true'
```
timcheadle: I'd rather call this `sync_all`, as it's more descriptive of what's actually happening.

app/controllers/api/v1/patients_controller.rb
```diff
@@ -27,8 +42,24 @@ def patients_params
...
+      Time.new(0)
+    else
+      params.require(:latest_record_timestamp).to_time
```
timcheadle: I'd prefer to rename this param to `:since` for simplicity.

app/controllers/api/v1/patients_controller.rb
```diff
@@ -27,8 +42,24 @@ def patients_params
...
+    end
+  end
+
+  def number_of_records
+    if params[:number_of_records].present?
```
timcheadle: I'd prefer to rename this param to `:limit` for simplicity

app/controllers/api/v1/patients_controller.rb
```diff
@@ -10,11 +10,26 @@ def sync_from_user
     render json: response, status: :ok
   end
 
+  def sync_to_user
+    patients_to_sync = Patient.updated_on_server_since(latest_record_timestamp, number_of_records)
+
+    next_latest_record_timestamp =
```
timcheadle: This variable name is confusing. How about `most_recent_record_timestamp`?

app/controllers/api/v1/patients_controller.rb
```diff
@@ -10,11 +10,26 @@ def sync_from_user
...
+      if patients_to_sync.empty?
+        latest_record_timestamp
+      else
+        patients_to_sync.last.updated_on_server_at
+      end
+
+    render json:   { patients:                patients_to_sync.map(&:nested_hash),
```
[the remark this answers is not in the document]
timcheadle: Also doesn't Rails default to `status: :ok` if you're rendering something? Do we need to add this explicitly?

app/helpers/mergeable_record.rb
timcheadle: Also this seems unrelated to this PR. Let's try and avoid PRs that accomplish too many things.

app/controllers/api/v1/patients_controller.rb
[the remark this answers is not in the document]
ssrihari: Sure, we'll change this one in this PR, and create another PR with the alignment settings changed.

app/helpers/mergeable_record.rb
ssrihari: > This feels like it should be implemented as a "mergeable" model concern

Yup, that makes sense. Will do.

> Also this seems unrelated to this PR.

This came up when we were changing the merging functionality for the GET sync call, where we set `updated_on_server_at` for patients depending on the merged `status` of it's children. We refactored the `MergeRecord` module into this.

app/controllers/api/v1/patients_controller.rb
ssrihari: `since` could be understood as a request/response timestamp as well. I would like to strongly separate the notions of request/response timestamps from record timestamps, which is what this is.

Over lunch today, we thought `checkpoint` could be a nicer name for this, because it also makes it independent of it's current implementation as a timestamp. If we choose to use a sequence of numbers for this later, the field name wouldn't change.

app/controllers/api/v1/patients_controller.rb
ssrihari: Hrm. `limit` _is_ simpler.

app/controllers/api/v1/patients_controller.rb
ssrihari: Yup, nice catch. This seemed so correct to @govindkrjoshi and me at the time.

ssrihari:
Hey @timcheadle, we're building the form-object pattern changes in another branch, but that work would conflict with the other PRs that depend on this PR. I guess this is still a bit early in the project to expect PRs to be independent.

We'll still address any comments on master or in another PR.

-- Govind and Srihari

PR #10, opened 2018-05-22
Change patient - phone_number association from many-to-many to has-many

govindkrjoshi:
- Add migration to delete patient_phone_number table, and add patient_id
  to phone_number
- Remove PatientPhoneNumber model
- Change factories to create new associations
- Add patient_id to phone_number before merging
- Fix model association assertions


ssrihari:
Should we call this model `PatientPhoneNumbers`? We might have phone numbers for facility and user as well.

db/migrate/20180522110435_change_patient_phone_number_association.rb
```diff
@@ -0,0 +1,8 @@
+class ChangePatientPhoneNumberAssociation < ActiveRecord::Migration[5.1]
+  def change
+    drop_table :patient_phone_numbers do
```
timcheadle: Why the empty block here?

PR #15, opened 2018-05-25
Validate request using JSON schema 

ssrihari:


app/payloads/api/v1/patient_payload.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def schema_with_definitions
+    schema.merge(definitions: Api::V1::Spec.all_definitions)
+  end
+
+  def errors_hash
+    errors.to_hash.merge(id: id)
+  end
+
+  def validate_schema
```
timcheadle: I'd move this up above `presence_of_age` to match the order of the validations

app/payloads/api/v1/patient_payload.rb
```diff
@@ -0,0 +1,48 @@
+class Api::V1::PatientPayload
```
timcheadle: Does this need a constructor? (looking at how it's called up above)

app/payloads/api/v1/spec.rb
```diff
@@ -61,85 +61,117 @@ def phone_number_spec
...
+      error_spec: error_spec }
+  end
+
+  def self.swagger_info
+    {
+      description: I18n.t('api_description'),
+      version:     'v1',
+      title:       'RedApp Server',
+      'x-logo'     => {
+        url:             'https://static1.squarespace.com/static/59945d559f7456b755d759f2/t/59aebc5ecf81e0ac9b4b6f59/1526304079797/?format=1500w',
```
timcheadle: Is it possible to use an asset in the repo for this?

app/payloads/api/v1/patient_payload.rb
govindkrjoshi: The constructor is provided by ActiveModel::Model

app/payloads/api/v1/patient_payload.rb
govindkrjoshi: This would ideally move out to common payload class, or module. We'll fix this in another PR.

app/payloads/api/v1/spec.rb
govindkrjoshi: Moved the strings to locale/en.yml for now. This would be convenient for internationalizing our API docs.

govindkrjoshi: Oops, I misunderstood this.
Added the logo as an image asset.

PR #17, opened 2018-05-27
Set updated_on_server_at only if patient is valid.

ssrihari:


ssrihari:
This is addressed by https://github.com/resolvetosavelives/redapp-server/pull/15.

PR #18, opened 2018-05-28
Rename created at and updated at

ssrihari:


app/controllers/api/v1/patients_controller.rb
```diff
@@ -1,11 +1,10 @@
...
-
   def merge_patient(single_patient_params)
-    patient_payload = Api::V1::PatientPayload.new(single_patient_params)
+    patient_payload = Api::V1::PatientPayload.new(single_patient_params.to_hash.with_indifferent_access)
     logger.debug "Patient had errors: #{patient_payload.errors_hash}" if patient_payload.invalid?
     return patient_payload.errors_hash if patient_payload.invalid?
 
-    MergePatientService.new(single_patient_params).merge
+    MergePatientService.new(patient_payload.model_attributes).merge
     nil
```
timcheadle: Returning nil is always a code smell. I'd flip the logic here, something like this:

```ruby
  unless patient_payload.invalid?
    MergePatientService.new(patient_payload.model_attributes).merge
  end

  patient_payload.errors_hash
end

File: app/controllers/api/v1/patients_controller.rb [hunk 2]
```diff
@@ -23,13 +22,13 @@ def sync_to_user
...
       else
-        patients_to_sync.last.updated_on_server_at
+        patients_to_sync.last.updated_at
       end
 
     render(
       json:   {
         patients:        patients_to_sync.map(&:nested_hash),
-        processed_since: most_recent_record_timestamp
+        processed_since: most_recent_record_timestamp.strftime('%FT%T.%3NZ')
```
timcheadle: We may want to make that format string a constant up above; makes the code more legible and maintainable

app/models/patient.rb
```diff
@@ -18,10 +18,22 @@ def errors_hash
...
   def nested_hash(options = {})
-    as_json(options.merge(
-      except:  %i[address_id  updated_on_server_at],
-      include: { address:       { except: :updated_on_server_at },
-                 phone_numbers: { except: :updated_on_server_at } }))
+    with_payload_keys(attributes)
+      .except('address_id')
+      .merge(
+        'address'       => with_payload_keys(address.attributes),
+        'phone_numbers' => phone_numbers.map { |phno| with_payload_keys(phno.attributes).except('patient_id') }
```
timcheadle: What is `phno`?

timcheadle: Oh, phone number! Let's make it `phone_number` (like below) for consistency

spec/controllers/api/v1/patients_controller_spec.rb
```diff
@@ -72,7 +72,7 @@
 
         patients_payload.each do |updated_patient|
           db_patient = Patient.find(updated_patient['id'])
-          expect(db_patient.attributes.with_int_timestamps.except('address_id', 'updated_on_server_at'))
+          expect(with_payload_keys(db_patient.attributes).with_int_timestamps.except('address_id'))
```
timcheadle: Where is this expecting to find `with_payload_keys` defined? Isn't it an instance method for the patient model?

spec/factories/patients.rb
```diff
@@ -15,21 +15,29 @@
...
-    updated_at { Time.now }
-    updated_on_server_at { Time.now }
+    device_created_at { Time.now }
+    device_updated_at { Time.now }
     association :address, strategy: :build
     phone_numbers { build_list(:patient_phone_number, rand(1..3), patient_id: id) }
   end
 end
 
+def with_payload_keys(attributes)
```
timcheadle: Nevermind, found it! This code duplication smells wrong. Shouldn't this test the model code itself?

PR #19, opened 2018-05-29
Sync Blood Pressures

ssrihari:
Issue: https://github.com/resolvetosavelives/redapp/issues/17


ssrihari:
All the duplication in the code until now is present intentionally. Will refactor.

app/models/blood_pressure.rb
```diff
@@ -0,0 +1,17 @@
+class BloodPressure < ApplicationRecord
+  include Mergeable
+  belongs_to :patient, optional: true
+
+  def with_payload_keys(attributes)
```
timcheadle: Seems like we could turn this into a model concern if we're copying this code into multiple models.

spec/controllers/api/v1/blood_pressures_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+  describe 'POST sync: send data from device to server;' do
+
+    describe 'creates new blood pressures' do
+
+      it 'returns 400 when there are no blood pressures in the request' do
+        post(:sync_from_user, params: { blood_pressures: [] })
+        expect(response.status).to eq 400
+      end
+
+      it 'creates new blood pressures without associated patient' do
```
timcheadle: Hmm... is this something we want? Creating BPs without a patient?

spec/requests/blood_pressure_sync_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    get sync_route, params: {}, headers: headers
+
+    response_body = JSON(response.body)
+    expect(response.status).to eq 200
+    expect(response_body['blood_pressures']).to eq([])
+    expect(response_body['processed_since']).to eq(Time.new(0).strftime('%FT%T.%3NZ'))
+  end
+
+  it 'push a new valid blood_pressure and pull first time' do
```
timcheadle: Typo: should be "pushes"

spec/controllers/api/v1/blood_pressures_controller_spec.rb
govindkrjoshi: This is intentional. 
BPs will have weak references to patients so can independently sync patients and blood pressures.

PR #24, opened 2018-06-03
Add Facilities

timcheadle:
Add support for Facilities.

- [x] Admin UI for CRUD actions
- [x] Read-only sync API
- [x] Associate with Patients
- [x] Associate with Blood Pressures


app/models/facility.rb
```diff
@@ -0,0 +1,11 @@
+class Facility < ApplicationRecord
+  include Mergeable
+
+  has_many :facility_patients
+  has_many :patients, through: :facility_patients
+
+  validates :name, presence: true
+  validates :district, presence: true
+  validates :state, presence: true
+  validates :country, presence: true
```
ssrihari: Since district, state and country are required fields, we should probably mark them required, and non-empty in the API spec as well.

app/models/patient.rb
```diff
@@ -8,6 +8,12 @@ class Patient < ApplicationRecord
   has_many :phone_numbers, class_name: 'PatientPhoneNumber'
   has_many :blood_pressures
 
+  has_many :facility_patients
+  has_many :facilities, through: :facility_patients
+
+  validates :device_created_at, presence: true
+  validates :device_updated_at, presence: true
+
```
ssrihari: Where is the facility assigned to a patient? Should that be a part of the patient sync API?

ssrihari:
The join table for blood pressures would be the same, yes?

db/migrate/20180603184000_enable_pgcrypto_extension.rb
```diff
@@ -0,0 +1,5 @@
+class EnablePgcryptoExtension < ActiveRecord::Migration[5.1]
+  def change
+    enable_extension 'pgcrypto'
```
ssrihari: Should we perhaps use https://www.postgresql.org/docs/9.6/static/uuid-ossp.html? It seems more relevant than `pgcrypto`, which has more crypto stuff.

db/schema.rb
```diff
@@ -38,6 +39,28 @@
     t.datetime "device_updated_at", null: false
   end
 
+  create_table "facilities", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
ssrihari: This is a nicer approach than the `before_create` callback we used for protocols. We'll change that.

ssrihari:
We should also have a relationship between facility and protocol. A facility `has a` protocol?

db/migrate/20180603184000_enable_pgcrypto_extension.rb
timcheadle: We discussed this; ActiveRecord has moved to `pgcrypto` in PG >= 9.4
http://edgeguides.rubyonrails.org/active_record_postgresql.html#uuid

PR #25, opened 2018-06-04
Protocols

ssrihari:
Issue: https://github.com/resolvetosavelives/redapp/issues/9


app/controllers/protocols_controller.rb
```diff
@@ -0,0 +1,61 @@
+class ProtocolsController < ApplicationController
```
timcheadle: These should live in an `Admin::` namespace

app/controllers/protocol_drugs_controller.rb
```diff
@@ -0,0 +1,66 @@
+class ProtocolDrugsController < ApplicationController
```
timcheadle: These should live in an `Admin::` namespace

app/models/protocol_drug.rb
```diff
@@ -0,0 +1,11 @@
+class ProtocolDrug < ApplicationRecord
+  belongs_to :protocol
+  before_create :assign_id
+
+  validates :name, presence: true
+  validates :dosage, presence: true
+
+  def assign_id
+    self.id = SecureRandom.uuid
```
timcheadle: We should discuss this. I think we should switch all of these records to create UUIDs as ID by default, or move UUID to a separate field and use bigint IDs. As it is, this is messy; I ran into the same issue with Facility.

app/controllers/protocols_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  def edit
+  end
+
+  def create
+    @protocol = Protocol.new(protocol_params)
+
+    respond_to do |format|
+      if @protocol.save
+        format.html { redirect_to @protocol, notice: 'Protocol was successfully created.' }
+        format.json { render :show, status: :created, location: @protocol }
```
timcheadle: I don't think we need JSON replies in the admin UI

app/models/protocol_drug.rb
ssrihari: Creating UUIDs by default in the database sounds like the way to go. We'll do this for all entities that are created via the admin interface.

app/controllers/protocols_controller.rb
ssrihari: Yeah, we'll remove these.

app/models/protocol_drug.rb
ssrihari: We'll do this after the facilities PR is merged in, so that we don't conflict with your changes of adding pgcrypto or another extension.

PR #30, opened 2018-06-07
Add, and sync prescription drugs

ssrihari:


spec/controllers/api/v1/prescription_drugs_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  let(:invalid_record) { build_invalid_payload.call }
+  let(:update_payload) { lambda { |prescription_drug| updated_prescription_drug_payload prescription_drug } }
+  let(:number_of_schema_errors_in_invalid_payload) { 2 }
+
+  describe 'POST sync: send data from device to server;' do
+    it_behaves_like 'a working sync controller creating records'
+    it_behaves_like 'a working sync controller updating records'
+
+    describe 'creates new prescription drugs' do
+      it 'creates new prescription drugs with associated patient' do
```
govindkrjoshi: We should also test creating new prescriptiong drugs with associated facilites

db/schema.rb
```diff
@@ -78,6 +78,18 @@
     t.datetime "device_updated_at", null: false
   end
 
+  create_table "prescription_drugs", id: :uuid, default: nil, force: :cascade do |t|
```
govindkrjoshi: This can be `id: :uuid` instead?

ssrihari: This is true right now for all models/records that are _only_ synced from user. We can change this uniformly for all such models later.

PR #31, opened 2018-06-08
Use capistrano to deploy app

govindkrjoshi:


.env
```diff
@@ -0,0 +1 @@
+DEFAULT_NUMBER_OF_RECORDS=10
```
timcheadle: We need to rename this file to `.env.development`. Let's also add empty values for `SENTRY_DSN` and `NEW_RELIC_LICENSE_KEY`

Capfile
```diff
@@ -0,0 +1,13 @@
+require "capistrano/setup"
+require "capistrano/deploy"
+
+require "capistrano/scm/git"
+install_plugin Capistrano::SCM::Git
+
+require "capistrano/rbenv"
+require "capistrano/bundler"
```
timcheadle: Let's just require `capistrano/rails` instead of breaking it out into `bundler`, `rails/assets`, etc.

config/deploy.rb
```diff
@@ -0,0 +1,13 @@
+# config valid for current version and patch releases of Capistrano
+lock "~> 3.11.0"
+
+set :application, "redapp-server"
+set :repo_url, "https://github.com/resolvetosavelives/redapp-server.git"
+set :deploy_to, -> { "/home/deploy/apps/#{fetch(:application)}" }
+set :rbenv_ruby, '2.3.4'
```
timcheadle: Let's use 2.5.1 instead. We should add a `.ruby-version` file with this as well.

config/deploy/production.rb
```diff
@@ -0,0 +1 @@
+server "13.127.138.54", user: "deploy", roles: %w(web app db)
```
timcheadle: Let's set up subdomains (even if temporary) instead of check in an IP.

config/deploy/staging.rb
```diff
@@ -0,0 +1 @@
+server "13.127.222.182", user: "deploy", roles: %w(web app db)
```
timcheadle: Let's set up subdomains (even if temporary) instead of check in an IP.

PR #54, opened 2018-06-26
Add feature toggle to selectively enable sync apis

govindkrjoshi:


app/controllers/api/v1/sync_controller.rb
```diff
@@ -24,6 +25,17 @@ def __sync_to_user__(response_key)
 
   private
 
+  def short_circuit_disabled_api
+    return if sync_api_toggled_on?
+    logger.info "Short circuiting #{request.env['PATH_INFO']} since it's a disabled feature"
+    head :ok
```
govindkrjoshi: Unsure about the HTTP status. Chose `:ok` so that the client go into an error flow. A better status would be `:not_found` or `:not_implemented`

PR #63, opened 2018-07-03
Slightly better layout, and fixed 'back' links.

ssrihari:


app/views/admin/facilities/show.html.erb
```diff
@@ -1,85 +1,92 @@
...
+              <li>
+                <%= if (user.logged_in_at.present?)
+                      'Logged in at:' + user.logged_in_at.iso8601
+                    else
+                      'Has not logged in yet'
+                    end
+                %>
+              </li>
+              <li>
+                Login link: <%= link_to "https://simple.org/register?otp=#{user.otp}" %>
```
timcheadle: This needs to duplicate the URL as a second argument to `link_to`, otherwise it links to the same page itself

PR #66, opened 2018-07-16
Add validations to ensure patient date of birth is in the past

govindkrjoshi:


app/payloads/api/v1/patient_payload_validator.rb
```diff
@@ -16,14 +16,21 @@ class Api::V1::PatientPayloadValidator < Api::V1::PayloadValidator
...
   validate :presence_of_age
+  validate :past_date_of_birth
 
   def presence_of_age
     unless date_of_birth.present? || (age.present? && age_updated_at.present?)
       errors.add(:age, 'Either date_of_birth or age and age_updated_at should be present')
     end
   end
 
+  def past_date_of_birth
```
timcheadle: This is fine, but it seems like we should delegate these kinds of validations to the model somehow so we don't repeat code here.

spec/payloads/patient_payload_validator_spec.rb
```diff
@@ -32,6 +32,12 @@ def new_patient_payload(attrs = {})
                                  'date_of_birth' => nil).valid?).to be false
     end
 
+    it 'validates patient date_of_birth is less than today' do
```
timcheadle: Let's make this test name match the one in the model spec

```ruby
it 'validates that date of birth is not in the future' do
```

PR #68, opened 2018-07-17
Add supervisor role for CVHO and STS

timcheadle:
*Story:* https://www.pivotaltracker.com/story/show/158967518

Add user roles and authorization (using [Pundit](https://github.com/varvet/pundit)) to allow CVHOs and STSs to take certain actions in the backend UI


app/models/admin.rb
```diff
@@ -1,4 +1,11 @@
 class Admin < ApplicationRecord
   devise :database_authenticatable, :lockable, :recoverable, :rememberable,
          :timeoutable, :trackable, :validatable
+
+  enum role: [
+    :owner,
```
govindkrjoshi: The name owner is ambiguous.

app/views/admin/facilities/index.html.erb
```diff
@@ -28,13 +28,17 @@
         <td><%= facility.country %></td>
         <td><%= facility.pin %></td>
         <td><%= facility.facility_type %></td>
-        <td><%= link_to 'Edit', edit_admin_facility_path(facility) %></td>
-        <td><%= link_to 'Destroy', [:admin, facility], method: :delete, data: { confirm: 'Are you sure?' } %></td>
+        <% if policy(facility).edit? %>
+          <td><%= link_to 'Edit', edit_admin_facility_path(facility) %></td>
+          <td><%= link_to 'Destroy', [:admin, facility], method: :delete, data: { confirm: 'Are you sure?' } %></td>
```
govindkrjoshi: We should check if the user has destroy policy as a separate condition

govindkrjoshi:
Should check for authorization for update and delete in the controllers as well?
The PR only shows edit or delete button if the user has permissions, but the controller should also check for permissions before allowing the actions.

app/models/admin.rb
timcheadle: We both agree, but are unsure what to call it. We'll keep it as `:owner` for now and change in the future.

app/views/admin/facilities/index.html.erb
timcheadle: 👍

PR #69, opened 2018-07-17
Allow setting branch name to deploy from env

govindkrjoshi:


config/deploy.rb
```diff
@@ -7,7 +7,11 @@
 set :rbenv_ruby, '2.5.1'
 set :rails_env, 'production'
 
-ask :branch, `git rev-parse --abbrev-ref HEAD`.chomp
+if ENV['DEPLOY_BRANCH']
```
timcheadle: I prefer this simpler method:

```ruby
set :branch, ENV["BRANCH"] || "master"
```

I don't like asking every time; I almost always want to deploy master to staging.

govindkrjoshi: Makes sense, I'll update this.

PR #72, opened 2018-07-20
Use Bootstrap for Admin UI

timcheadle:
**Stories:**
- https://www.pivotaltracker.com/story/show/159119611
- https://www.pivotaltracker.com/story/show/159119539

Update the admin UI to use Bootstrap and have a nicer, responsive look and feel.


timcheadle:
This is nearly ready to go, but I broke the admin controller specs for `User` and `ProtocolDrug` when i made them nested resources. @govindkrjoshi do you mind picking that fix up?

govindkrjoshi:
@timcheadle We are using Model.find in many places in the admin UI. 

Should we add rescue statements for RecordNotFound and StandardError in Application controller?

PR #73, opened 2018-07-26
Nurse self registration

govindkrjoshi:
Issue: https://www.pivotaltracker.com/story/show/159268022


app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    logger.debug "User registration params had errors: #{validator.errors_hash}" if validator.invalid?
+    if validator.invalid?
+      render json: { errors: validator.errors }, status: :bad_request
+    end
+  end
+
+  def registration_params
+    params.require(:user)
+      .permit(
+        :id,
```
govindkrjoshi: IDs are created on the client.
This is so that if the user continues to use the app while registration happens in the backend, the user id remains consistent

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+        :full_name,
+        :phone_number,
+        :password_digest,
```
govindkrjoshi: Registration API will be called a synchronously on the client. The API accepts password digest instead of password so that the app doesn't have to persist the plain text password

app/models/user.rb
```diff
@@ -12,8 +18,9 @@ class User < ApplicationRecord
   before_create :set_access_token
 
   validates :full_name, presence: true
-  validates :phone_number, presence: true
+  validates :phone_number, presence: true, uniqueness: true
```
govindkrjoshi: Phone numbers for users need to be unique so that the user can be looked up with just phone number

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      it 'responds with 400' do
+        post :register, params: request_params
+
+        expect(response.status).to eq(400)
+      end
+    end
+
+    describe 'registration payload is valid' do
+      let(:facility) { FactoryBot.create(:facility) }
+      let(:user) do
```
timcheadle: I'd call this `user_params` since it isn't actually a user.

spec/factories/users.rb
```diff
@@ -7,6 +7,7 @@
     association :facility, strategy: :build
     device_updated_at { Time.now }
     device_created_at { Time.now }
+    sync_approval_status { User::SYNC_STATUSES.values.sample }
```
timcheadle: Why not set this to a specific status? Looks like it would create some funky debugging situations if it's random.

app/models/user.rb
```diff
@@ -1,6 +1,12 @@
 class User < ApplicationRecord
   include Mergeable
 
+  SYNC_STATUSES = {
```
timcheadle: Can we simplify this and make it an `enum` instead?

spec/controllers/api/v1/users_controller_spec.rb
govindkrjoshi: Updated.

spec/factories/users.rb
govindkrjoshi: The factory is used to create random patients, so it made sense to also choose a random status. Agreed that it might make debugging a little difficult. We can be explicit when we need random statuses. Will change this in patient factory as well.

app/models/user.rb
govindkrjoshi: Enums convert the value into an int column and the value is order specific. We can use a hash to define the Enum instead, in which case the order does not matter. 
Eg: `enum sync_approval_status: { waiting: 0, approved: 2, disapproved: 1}` 

But converting the value to int makes it rails specific. Someone looking at the DB will not directly know what the status is. 

We can define the enum as 
```enum sync_approval_status {
 waiting: 'waiting_for_approval', 
disapproved: 'disapproved_for_syncing',
approved: 'approved_for_syncing',
```
to store strings in the database.
I've tested this and it works, but the behavior is undocumented.

app/models/user.rb
```diff
@@ -1,6 +1,12 @@
 class User < ApplicationRecord
   include Mergeable
 
+  enum sync_approval_status: {
+    waiting: 'waiting',
```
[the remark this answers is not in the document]
ssrihari: Also from a user's perspective, let's call the other statuses `allowed` and `denied`. The buttons can be 'allow access', and `deny access`. Further, the actions to allow/deny access post registration can also remain the same.

PR #77, opened 2018-08-01
Allow managing and inviting owners and supervisor users

timcheadle:
**Story:** https://www.pivotaltracker.com/story/show/158967543

Use `devise_invitable` to invite admin users.


app/controllers/application_controller.rb
```diff
@@ -1,5 +1,24 @@
...
+  # Send a user to the admins index after sending invitations
+  def after_invite_path_for(inviter, invitee)
+    admins_path
+  end
+
+  # Customize which fields Devise allows for Admins
+  # See https://github.com/plataformatec/devise/tree/v3.5.2#strong-parameters
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:invite, keys: [:email, :role])
+  end
```
ssrihari: These seem orthogonal to the API controller, and seem better encapsulated in the `AdminController`'s concerns. Why is it a better idea to have them here?

-- Govind and Srihari

timcheadle: It's because `DeviseController` inherits from `ApplicationController`. It was either this or implement our own controllers for all of devise, which seemed excessive.

PR #83, opened 2018-08-07
Supervisor approves nurse

ssrihari:
Story: https://www.pivotaltracker.com/story/show/158963884

- Setup approval emails post registration, using action mailer
  - Provide information within email for admin to make decision
  - Provide link to login and allow or deny access
- Email is sent asynchronously
- Use sendgrid api key to login
- Setup owner and supervisor emails in ENV
- Derive Users controller from AdminController, and specify authorisations
- Login admin before each test
- Test that emails get sent out, and that access is granted or denied.


app/policies/user_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+  def destroy?
+    create?
+  end
+
+  def show?
+    index?
+  end
+
+  def disable_access?
+    show?
```
govindkrjoshi: This should be the same as `update?`

app/policies/user_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def enable_access?
+    show?
```
govindkrjoshi: Similarly here.

app/views/approval_notifier_mailer/approval_email.html.erb
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta content='text/html; charset=UTF-8' http-equiv='Content-Type'/>
+</head>
+<body>
+<h2>User <%= @user.full_name %> is requesting access.</h2>
```
govindkrjoshi: This can also be a locale string.

app/views/approval_notifier_mailer/approval_email.html.erb
```diff
@@ -0,0 +1,31 @@
...
+<p>Please allow or deny.</p>
```
govindkrjoshi: Same here

app/views/approval_notifier_mailer/approval_email.html.erb
```diff
@@ -0,0 +1,31 @@
...
+</p>
+
+<p>
+  <strong>Facility:</strong>
+  <%= @user.facility.name %>
+</p>
+
+<p>
+  <strong>First login in at:</strong>
+  <%= @user.logged_in_at.try(:iso8601) || 'Not logged in yet' %>
```
govindkrjoshi: The user would have never logged in on the server when the email is sent. Is this information necessary?

PR #87, opened 2018-08-10
Multiple facility for user

govindkrjoshi:
Allow users to select multiple facilities during registration

Issue: https://www.pivotaltracker.com/story/show/158963795

### DIFF
--- a/app/controllers/api/v1/sync_controller.rb
+++ b/app/controllers/api/v1/sync_controller.rb
@@ -4,7 +4,7 @@ class Api::V1::SyncController < APIController
   def __sync_from_user__(params)
     errors = params.flat_map do |single_entity_params|
       res = merge_if_valid(single_entity_params)
-      AuditLog.merge_log(current_user, res[:record]) if res[:record].present?
+      AuditLog.merge_log(current_user, res[:record]) if current_user.present? && res[:record].present?
       res[:errors_hash] || []
     end
 
@@ -15,7 +15,7 @@ class Api::V1::SyncController < APIController
 
   def __sync_to_user__(response_key)
     records_to_sync = find_records_to_sync(processed_since, limit)
-    records_to_sync.each { |record| AuditLog.fetch_log(current_user, record) }
+    records_to_sync.each { |record| AuditLog.fetch_log(current_user, record) } if current_user.present?
     render(
       json:   {
         response_key      => records_to_sync.map { |record| transform_to_response(record) },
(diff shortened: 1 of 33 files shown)


app/controllers/api/v1/sync_controller.rb
```diff
@@ -38,25 +37,6 @@ def sync_api_toggled_on?
     FeatureToggle.enabled_for_regex?('MATCHING_SYNC_APIS', controller_name)
   end
 
-  def current_user
```
govindkrjoshi: Moving authentication from SyncController to ApiController

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -50,20 +54,35 @@
...
     end
   end
 
   describe '#find' do
     let(:phone_number) { Faker::PhoneNumber.phone_number }
-    let!(:db_users) { FactoryBot.create_list(:user, 10) }
-    let!(:user) { FactoryBot.create(:user, phone_number: phone_number) }
+    let(:facility) { FactoryBot.create(:facility) }
+    let!(:db_users) { FactoryBot.create_list(:user, 10, facility_ids: [facility.id]) }
+    let!(:user) { FactoryBot.create(:user, phone_number: phone_number, facility_ids: [facility.id]) }
```
timcheadle: Please add an empty line between your `let` calls and the first example

spec/models/user_spec.rb
```diff
@@ -2,7 +2,8 @@
 
 RSpec.describe User, type: :model do
   describe 'Associations' do
-    it { should belong_to(:facility) }
+    it { should have_many(:user_facilities) }
```
timcheadle: Add a test for `dependent: :destroy`

app/views/admin/users/index.html.erb
```diff
@@ -1,5 +1,32 @@
...
+      <% if policy(user).destroy? %>
+        <td><%= link_to 'Delete', [:admin, user], method: :delete, data: { confirm: 'Are you sure?' } %></td>
+      <% end %>
+    </tr>
+  <% end %>
+  </tbody>
+</table>
+
+<% if policy(User).new? %>
+  <%= link_to 'New User', new_admin_user_path, class: "btn btn-sm btn-primary" %>
```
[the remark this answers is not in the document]
govindkrjoshi: I haven't removed this feature yet, because I'm not sure that we don't need it anymore. The user can be approved at the the time of creation, and then they can log in to use the app.
I was thinking we would remove this feature in a future PR.

PR #90, opened 2018-08-16
Add an endpoint for users to request otps for login

govindkrjoshi:


app/controllers/api/v1/users_controller.rb
```diff
@@ -16,6 +16,14 @@ def find
     render json: user_to_response(user), status: 200
   end
 
+  def request_otp
+    user = User.find(request_otp_id_param)
+    user.set_otp
+    user.save
+    SmsNotificationService.new(user).send_request_otp_sms
```
ssrihari: Since this is essentially the same functionality as reset_otp, should we move this to the User model?

app/payloads/api/v1/schema.rb
```diff
@@ -108,6 +108,13 @@ def self.user_registration_request
       required:   %i[user] }
   end
 
+  def self.user_request_otp_request
+    { type:       :object,
+      properties: {
+        id:         { '$ref' => '#/definitions/uuid' } },
```
ssrihari: Isn't this better as a route parameter? `/users/:id/request_otp`?

PR #91, opened 2018-08-17
Add access token to the registration response

govindkrjoshi:


app/payloads/api/v1/schema.rb
```diff
@@ -95,10 +95,7 @@ def self.user_login_success_response
   end
 
   def self.user_registration_response
-    { type:       :object,
-      properties: {
-        user:         { '$ref' => '#/definitions/user' } },
-      required:   %i[user] }
+    user_login_success_response
```
ssrihari: Since registration precedes login in the user flows, perhaps we should reference the registration response in the login response instead. This is very minor.

PR #98, opened 2018-08-23
Sync Appointments and Communications

govindkrjoshi:


app/models/appointment.rb
```diff
@@ -0,0 +1,22 @@
...
+    scheduled: 'scheduled',
+    cancelled: 'cancelled',
+    visited: 'visited'
+  }, _prefix: true
+
+  enum status_reason: {
+    not_called_yet: 'not_called_yet',
+    not_responding: 'not_responding',
+    moved: 'moved',
+    dead: 'dead'
```
ssrihari: probably add an 'other', since it's in the designs.

app/models/appointment.rb
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  validates :device_created_at, presence: true
+  validates :device_updated_at, presence: true
+end
```
ssrihari: Should also validate presence of `date`.

app/payloads/api/v1/schema/models.rb
```diff
@@ -184,6 +184,35 @@ def self.login_user
...
+  end
+
+  def self.communication
+    { type: :object,
+      properties: {
+        id: { '$ref' => '#/definitions/uuid' },
+        appointment_id: { '$ref' => '#/definitions/uuid' },
+        user_id: { '$ref' => '#/definitions/uuid' },
+        communication_type: { type: :string, enum: Communication.communication_types.keys },
+        communication_result: { type: :string, enum: Communication.communication_results.keys },
```
ssrihari: We could transform `type` and `result` with prefixes, so the API can be elegant.

db/schema.rb
```diff
@@ -63,6 +63,20 @@
     t.index ["unlock_token"], name: "index_admins_on_unlock_token", unique: true
   end
 
+  create_table "appointments", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "patient_id", null: false
+    t.uuid "facility_id", null: false
+    t.date "date", null: false
+    t.string "status"
+    t.string "status_reason"
```
ssrihari: Both `status` and `status_reason` should be `null: false`.

db/schema.rb
```diff
@@ -87,6 +101,19 @@
     t.index ["user_id"], name: "index_blood_pressures_on_user_id"
   end
 
+  create_table "communications", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "appointment_id", null: false
+    t.uuid "user_id", null: false
+    t.string "communication_type"
+    t.string "communication_result"
```
ssrihari: Similarly, type and result here shouldn't be nullable.

app/models/appointment.rb
```diff
@@ -0,0 +1,22 @@
+class Appointment < ApplicationRecord
+  include Mergeable
+
```
ssrihari: Should `have_many` communications.

PR #101, opened 2018-08-29
[WIP] Punjab bulk facility import

timcheadle:
**Story:** https://www.pivotaltracker.com/story/show/159645477

This PR should enable us to import the Punjab Master Facility Registry (as provided to us from the World Health Organization). The script should be idempotent, allowing to re-run it and not duplicate facilities. It should also allow us to import future versions of the MFR and only update what changed.


ssrihari:
Most of this is subsumed by work over time. Will revisit this when it becomes necessary.

PR #106, opened 2018-09-05
Sync patient medical history

govindkrjoshi:


saket:
@govindkrjoshi does this also include purging?

govindkrjoshi:
@saket https://github.com/simpledotorg/simple-server/pull/106/commits/15ac158e2e7346d8d1c6d581d195500070e8107c

saket:
@govindkrjoshi `MedicalHistory` is missing a field for diabetes

app/controllers/api/v1/medical_histories_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  def transform_to_response(medical_history)
+    Api::V1::Transformer.to_response(medical_history)
+  end
+
+  def medical_histories_params
+    params.require(:medical_histories).map do |medical_history_params|
+      medical_history_params.permit(
+        :id,
+        :patient_id,
+        :has_prior_heart_attack,
```
timcheadle: I personally prefer removing the `has` or `is` prefixes in Rails, as the tense is confusing. `patient.prior_heart_attack?` answers the question more clearly.

Can we change these to `prior_heart_attack`, `prior_stroke`, `has_chronic_kidney_disease`, and `on_hypertension_treatment`?

app/models/medical_history.rb
```diff
@@ -0,0 +1,7 @@
+class MedicalHistory < ApplicationRecord
+  include Mergeable
+  belongs_to :patient, optional: true
```
timcheadle: Is this optional to allow async record creation? I mean the patient association _isn't_ optional.

spec/api/appointments_spec.rb
```diff
@@ -4,7 +4,7 @@
   path '/appointments/sync' do
 
     post 'Syncs appointment data from device to server.' do
-      tags 'Follow Up Schedule'
+      tags 'Appointment'
```
timcheadle: Why is this in this PR?

spec/api/communications_spec.rb
```diff
@@ -4,7 +4,7 @@
   path '/communications/sync' do
 
     post 'Syncs communication data from device to server.' do
-      tags 'Follow Up'
+      tags 'Communication'
```
timcheadle: Why is this in this PR?

spec/api/appointments_spec.rb
govindkrjoshi: Missed renaming these tags in the PR adding `Appointment` and `Communication`. 

Wanted to fix these in the PR itself, instead of raising another one, since it's a small change.

app/models/medical_history.rb
govindkrjoshi: Yes, patient association is optional because of weak references between patient and medical history. 
Patients and Medical histories can be synced independent of each other.

spec/api/appointments_spec.rb
timcheadle: If you don't mind, can you add those to another PR? If it's small it should be easy to review & merge.

Putting unrelated changes in can make future reference confusing.

app/controllers/api/v1/medical_histories_controller.rb
timcheadle: I agree, and Active Record gives you that. Having a boolean field called `prior_stroke` gives you the function `patient.prior_stroke?` in Rails.

govindkrjoshi: Oh, I missed that. Thanks. I've made these changes.

PR #108, opened 2018-09-06
Add views for audit logs

govindkrjoshi:


app/views/admin/audit_logs/index.html.erb
```diff
@@ -0,0 +1,31 @@
+<p id="notice"><%= notice %></p>
+
+<h1>Audit Logs</h1>
+
+<%= render 'search_form' %>
```
timcheadle: No need for a partial here. We aren't going to reuse it, are we?

app/views/admin/audit_logs/index.html.erb
```diff
@@ -0,0 +1,31 @@
...
+<br/>
```
timcheadle: Why a break here after a `<div class="row">`?

app/views/admin/audit_logs/index.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+<table class="table">
+  <thead class="thead-light">
+  <tr>
```
timcheadle: Please indent the `<tr>`

app/views/admin/audit_logs/index.html.erb
```diff
@@ -0,0 +1,31 @@
...
+    <th>Action type</th>
+    <th>Model type</th>
+    <th>Model id</th>
+    <th>User Name</th>
+    <th colspan="2"></th>
+  </tr>
+  </thead>
+
+  <tbody>
+  <% @audit_logs.each do |audit_log| %>
```
timcheadle: Please indent the ERB block

app/views/admin/audit_logs/index.html.erb
```diff
@@ -0,0 +1,31 @@
...
+    <tr>
+      <td><%= link_to audit_log.created_at, [:admin, audit_log] %></td>
```
timcheadle: I'm not sure how useful a `show` page is here. Aren't we showing all of the same info in this table?

spec/controllers/admin/audit_logs_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Admin::AuditLogsController, type: :controller do
```
timcheadle: No need to add a blank test. Instead, I'd love to see a feature spec that tests the search.

app/views/admin/audit_logs/index.html.erb
govindkrjoshi: I wanted to separate out the form because it's likely this will change soon. I'll remove this for now.

app/views/admin/audit_logs/index.html.erb
govindkrjoshi: Without the break, the table sticks to the search button. This is an easy way to add a margin.

app/views/admin/audit_logs/index.html.erb
govindkrjoshi: Done.

app/views/admin/audit_logs/index.html.erb
govindkrjoshi: Removed the show page.

spec/controllers/admin/audit_logs_controller_spec.rb
govindkrjoshi: Added feature tests and deleted this file.

PR #116, opened 2018-09-17
Document deployment process

govindkrjoshi:
- Add changelog
- Add a list of steps, or a checklist one could follow
- Separate portal changes from API changes

-- Srihari and Govind

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,43 @@
+# Unreleased
+### Added 
+### Changed
+### Fixed
+### Deprecated 
+### Removed 
+### Security 
+
+# 2018-09-17
+
+### Changed 
+
+- User can belong to multiple facilities
+- Sort users by fullname in admin portal
+- Remove user otp message from admin portal
+- Add phone numbers to users index view   
+- Order facilities by name
+- Rename tags in swagger schema
+
+### Fixed
+
+- Ensure config for required keys exists before app starts
+- Reset user access token every time the user logs in (only occurs on a new device)
+- Don't reset access token when enabling access, only update sync status
+- Remove sms notification to user when access is enabled
+- Redirect to login when admin is not logged in
+- Add access token to the registration response
+- Remove authentication for ping checks
+- Only display links that the admin can access
+- Delete all appointments and communications on purge (for QA)
+
+### Added 
+ 
+- Sync APIs for Patient medical history
+- Sync APIs for Appointments and Communications (for overdue list)
+- API for users to request OTPs for login  
+- Views for audit logs
+- Feature toggles
+   - to auto approve users after registration (for QA)
+   - to have fixed OTPs (for QA)
+
+## 2018-08-13
+## 2018-07-02
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,11 @@
 ### Security 
 
 # 2018-09-17
+## API
 
 ### Changed 
 
 - User can belong to multiple facilities
-- Sort users by fullname in admin portal
-- Remove user otp message from admin portal
-- Add phone numbers to users index view   
-- Order facilities by name
 - Rename tags in swagger schema
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


doc/deployment.md
```diff
@@ -0,0 +1,51 @@
...
+- Create a tag off master with the same name as in the changelog. Typically with the format: yyyy-mm-dd-n.
+- Add a short description in the tag for the release
+- `git tag -a <yyyy-mm-dd-n> -m "<description>"`
+- `git push --tags`
+
+3. Understand migrations to be run as a part of this release
+- `git diff <previous-tag>..<current-tag> db/migrate`
+- lookout for data migrations, and irreversible migrations   
+
+Perform the following steps first on staging, QA it, and then perform them on production.
```
timcheadle: Rather than have an iterative instruction, I'd explicitly call out that the next steps should be done on **staging**. Then, below all of them, have a line that says repeat the above for **production** if everything went OK.

doc/deployment.md
```diff
@@ -0,0 +1,51 @@
...
+- Run the migration locally with the prod data to ensure it succeeds
+- Make a backup of the production DB before doing the data migration
+- Should it run before the server comes back up?
+  - If so, ensure cap deploy does not start the service before that  
+
+5. Setup configs, and feature flags
+- Find out what config has changed since the last release
+- `git diff -w <previous-tag>..<current-tag> .env.development`
+- Verify ansible vault has all the required keys and values
+- Run ansible deploy from deployment repository to ensure config is updated
```
timcheadle: Link to the deployment repo :)

doc/deployment.md
```diff
@@ -0,0 +1,51 @@
...
+5. Setup configs, and feature flags
+- Find out what config has changed since the last release
+- `git diff -w <previous-tag>..<current-tag> .env.development`
+- Verify ansible vault has all the required keys and values
+- Run ansible deploy from deployment repository to ensure config is updated 
+- `ansible-playbook -v  --vault-id ~/Projects/resolve/secrets/password_file deploy.yml -i hosts.<env>`
+ 
+6. Prepare for potential downtime
+- Figure out the right time of the day during which downtime is acceptable
+- Inform stakeholders of this downtime, and ensure this is alright.
```
timcheadle: Who are the stakeholders? We should not depend on devs just knowing what this step means

Instead, we could say to clearly message this on Slack, and/or institute maintenance windows for production.

CHANGELOG.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Portal
+
+- Add views for audit logs
+- Sort users by fullname 
+- Remove user otp message
+- Add phone numbers to users index view   
+- Order facilities by name
+
+## 2018-08-13
```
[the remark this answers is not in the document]
ssrihari: This should be `2018-08-13-1`, and it's the release tag. The date on which we create the tag and deploy to production.

doc/deployment.md
```diff
@@ -0,0 +1,51 @@
...
+- `git push --tags`
+
+3. Understand migrations to be run as a part of this release
+- `git diff <previous-tag>..<current-tag> db/migrate`
+- lookout for data migrations, and irreversible migrations   
+
+Perform the following steps first on staging, QA it, and then perform them on production.
+
+4. If there is a data migration
+- Check for changes in lib: `git diff -w 2018-08-13-1..2018-09-17-1 lib/`
```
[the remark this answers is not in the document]
ssrihari: This diff tells us if there are any data migrations, or important scripts/seeds that need to be run. The following steps just say "look out for" migrations, etc because what needs to be done depends on the context.

Sometimes, data migrations can be run after the deploy, and sometimes they need to be run before. The `cap deploy` script does a code deploy, runs schema migrations, and starts the service. We might need to modify it to not start the service until the data migrations are run.

We decided we won't detail this, or modify the scripts until we come across such a scenario. But it's probably worth adding the above description into the document.

Thanks for pointing this out, @timcheadle :)

doc/deployment.md
ssrihari: Cool, that sounds nicer. Will change.

doc/deployment.md
ssrihari: Makes sense. Will change.

PR #117, opened 2018-09-19
[WIP] Add api and schema specification for reset password api

govindkrjoshi:


app/controllers/admin/users_controller.rb
```diff
@@ -1,9 +1,16 @@
 class Admin::UsersController < AdminController
   before_action :set_user, except: [:index, :new, :create]
 
+  def ordered_sync_approval_statuses
```
govindkrjoshi: Make this a private method.

app/controllers/api/v1/users_controller.rb
```diff
@@ -20,19 +27,24 @@ def find
...
     user.save
     SmsNotificationService.new(user).send_request_otp_sms
     head :ok
   end
 
+  def reset_password
+    current_user.reset_password(reset_password_digest)
+    current_user.save
+    ApprovalNotifierMailer.with(user: current_user).reset_password_approval_email.deliver_later
+    head :ok
```
govindkrjoshi: Should return the user object and access token.

PR #143, opened 2018-10-19
Add patient visits to dashboard

timcheadle:
Show unique patients seen, and extend report back 9 months


app/controllers/admin/dashboard_controller.rb
```diff
@@ -35,12 +37,16 @@ def bps_by_facility_user_day
     bp_counts_by_day = BloodPressure.group(:facility_id, :user_id).group_by_day(:device_created_at, last: 7).count
```
govindkrjoshi: Numbers like 7 and 9 can be extracted as variables to make the context more clear.

PR #144, opened 2018-10-23
Add email prefixes in staging

timcheadle:
- [x] Add the ability to prefix email subjects (in staging)


app/controllers/api/v1/users_controller.rb
```diff
@@ -37,7 +37,7 @@ def request_otp
   def reset_password
     current_user.reset_password(reset_password_digest)
     current_user.save
-    ApprovalNotifierMailer.with(user: current_user).reset_password_approval_email.deliver_later
+    ApprovalNotifierMailer.with(user: current_user).reset_password_approval_email.deliver_now
```
ssrihari: I think we should keep this `deliver_later` for now. ActiveJobs are async by default since Rails 5. And we shouldn't block API calls from nurses on sending emails. Agree that we should move towards a persistent queue, with more control in the future.

ssrihari:
```suggestion
    ApprovalNotifierMailer.with(user: current_user).reset_password_approval_email.deliver_later
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -11,7 +11,7 @@ def register
     else
       user.sync_approval_requested(I18n.t('registration'))
       user.save
-      ApprovalNotifierMailer.with(user: user).registration_approval_email.deliver_later
+      ApprovalNotifierMailer.with(user: user).registration_approval_email.deliver_now
```
ssrihari:
```suggestion
      ApprovalNotifierMailer.with(user: user).registration_approval_email.deliver_later
```

ssrihari:
Apart from the `deliver_now` change, this PR looks good. Really needed the prefix!

timcheadle:
I'll switch back to `deliver_later`, but we _really_ should set up Sidekiq. ActiveJobs async handler only keeps jobs in memory, so if the app restarts we lose them.

PR #147, opened 2018-10-26
Respond with false if medical history question is nil

govindkrjoshi:
- Reverts simpledotorg/simple-server#145
- Transforms nils to falls in sync_to_user response


app/payloads/api/v1/medical_history_transformer.rb
```diff
@@ -0,0 +1,20 @@
...
+    :prior_stroke,
+    :chronic_kidney_disease,
+    :receiving_treatment_for_hypertension,
+    :diabetes,
+    :diagnosed_with_hypertension
+  ].freeze
+
+  def self.medical_history_questions
+    MEDICAL_HISTORY_QUESTIONS
+  end
```
ssrihari: This looks a bit weird. Can't we access module constants outside directly?

govindkrjoshi: I tried, but couldn't figure out a way to this.

PR #157, opened 2018-11-02
Masked phone numbers

govindkrjoshi:


ssrihari:
Todo:
1. Use consistent terminology => `user_phone_number`, `patient_phone_number`. Avoid `number`, `nurse`, etc.
2. Check from user should include sync approval check.
3. Need to add an ADR/sequence diagram for the handshake between app<>twilio<>server
4. Add API to reallocate another number
5. Create a call logs table, add to it during the call
6. Create a table for storing allocated numbers, add admin interface for managing these numbers

Optionally:
1. Listen to callbacks indicating calls ending
2. Only allow calls against scheduled appointments

PR #164, opened 2018-11-20
Update medical history questions from boolean to enum

govindkrjoshi:


app/tranformers/api/v1/medical_history_transformer.rb
```diff
@@ -1,2 +1,28 @@
...
+  MEDICAL_HISTORY_ANSWERS_MAP = {
+    yes: true,
+    no: false,
+    unknown: false
+  }.with_indifferent_access.freeze
+
+  INVERTED_MEDICAL_HISTORY_ANSWERS_MAP = {
+    true => :yes,
+    false => :unknown
+  }.with_indifferent_access.freeze
```
ssrihari: Use ruby's Hash#invert?

ssrihari: Also, avoid `_MAP` in the names.

PR #167, opened 2018-11-23
Soft deletes

ssrihari:


spec/payloads/api/current/patient_payload_validator_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+                                 'date_of_birth' => nil).valid?).to be false
+    end
+
+    it 'validates patient date_of_birth is less than today' do
+      payload = new_patient_payload('date_of_birth' => 3.days.from_now)
+      expect(payload.valid?).to be false
+      expect(payload.errors[:date_of_birth]).to be_present
+    end
+
+    it "Validates json spec for patient sync request" do
```
govindkrjoshi: Split these into separate `it` blocks?

spec/payloads/api/current/patient_payload_validator_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+  payload
+end
+
+describe Api::Current::PatientPayloadValidator, type: :model do
+  describe "Validations" do
+    it "Validates that either age or date of birth is present" do
+      expect(new_patient_payload('address'       => nil,
+                                 'phone_numbers' => nil,
+                                 'age'           => nil,
+                                 'date_of_birth' => Date.today).valid?).to be true
```
govindkrjoshi:
```suggestion
                                 'date_of_birth' => Date.today)).to be_valid
```

app/schema/api/v1/models.rb
```diff
@@ -27,6 +27,42 @@ def medical_history
       }
     end
 
+    def patient
+      super.tap { |d| d[:properties].delete(:deleted_at) }
```
govindkrjoshi: What happens when someone updates a deleted record using the V1 api?

ssrihari: The updates go through the system as though nothing changed.

Only difference is the deleted records should not show up on the UI.

spec/payloads/api/current/patient_payload_validator_spec.rb
ssrihari: This change seems unrelated to this PR. Don't know why it shows up here. I don't remember making it. Will check.

ssrihari: Ah, I see I versioned the spec file, because the v1 test was different. Fixed in 11a00c59fe892d3b6e893c5961ce7e49e3617f7e.

PR #169, opened 2018-11-28
Replace negation of present? with blank?

arun1595:
This is a minor refactor replacing `!present?` with `blank?`

Also introduces specs for `User#has_never_logged_in?` to add coverage.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -44,4 +44,20 @@ RSpec.describe User, type: :model do
       end
     end
   end
+
+  describe '#has_never_logged_in?' do
+    context 'user has never logged in' do
+      it 'returns true' do
+        user = User.new(logged_in_at: nil)
+        expect(user.has_never_logged_in?).to be true
+      end
+    end
+
+    context 'user has logged in atleast once' do
+      it 'returns false' do
+        user = User.new(logged_in_at: DateTime.yesterday)
+        expect(user.has_never_logged_in?).to be false
+      end
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


ssrihari:
Hey thanks @arun1595!

spec/models/user_spec.rb
```diff
@@ -44,4 +44,20 @@
       end
     end
   end
+
+  describe '#has_never_logged_in?' do
+    context 'user has never logged in' do
+      it 'returns true' do
+        user = User.new(logged_in_at: nil)
```
arun1595: We could use `FactoryBot` to build the `User` object. I used `User` to keep it relatively simple.

spec/models/user_spec.rb
```diff
@@ -44,4 +44,20 @@
...
+        expect(user.has_never_logged_in?).to be true
```
arun1595: This could also be written as `expect(user).to have_never_logged_in` and is more readable imo

But the expectation in the other context would read `expect(user),not_to have_never_logged_in` which sounds complicated.

arun1595:
@ssrihari Hey! :wave: 

Thanks for the quick reviews :slightly_smiling_face: 

Btw, the test suite fails locally but I'm not sure of the reason. Are you able to get all the tests to pass on your local machine?

ssrihari:
@arun1595 Yeah, the tests pass :) We also run on CI here: https://semaphoreci.com/resolvetosavelives/simple-server

spec/models/user_spec.rb
timcheadle: Agree; I'm not a fan of the `have_*` syntax

PR #173, opened 2018-11-30
[Bugfix] Report user id with extra args to sentry

govindkrjoshi:


app/controllers/api/v1/medical_histories_controller.rb
```diff
@@ -5,7 +5,7 @@ class Api::V1::MedicalHistoriesController < Api::Current::MedicalHistoriesContro
   def sync_from_user
     Raven.capture_message(
       'User is trying to sync medical histories using API V1.',
-      user: current_user.id
+      extra: { user: current_user.id }
```
timcheadle: There's also a way to explicitly set user context in the raven gem:

```ruby
Raven.user_context(id: current_user.id)
```

Should we do that on user login so it's auto-captured for all sentry events in the session?

govindkrjoshi: Agreed. Updated the PR.

app/controllers/api_controller.rb
```diff
@@ -38,4 +38,8 @@ def access_token_authorized?
       ActiveSupport::SecurityUtils.secure_compare(token, current_user.access_token)
     end
   end
+
+  def set_sentry_context
+    Raven.user_context(id: request.headers['HTTP_X_USER_ID'])
```
ssrihari: If sentry allows it, add `facility_id` also?

govindkrjoshi:
```suggestion
    Raven.user_context(
      id: request.headers['HTTP_X_USER_ID'],
      request_facility_id: request.headers['HTTP_X_USER_ID']
    )
```

PR #180, opened 2018-12-07
Update cancel reasons in appointments

ssrihari:
- Add the 3 new reasons to v2, exclude them from v1
- Coerce them to other in v1 response
- Do not accept 'other' from v1 if the appointment is cancelled


app/controllers/api/v1/appointments_controller.rb
```diff
@@ -1,4 +1,26 @@
 class Api::V1::AppointmentsController < Api::Current::AppointmentsController
   include Api::V1::ApiControllerOverrides
   include Api::V1::SyncControllerOverrides
+
+  private
+
+  def merge_if_valid(appointment_params)
+    validator = Api::Current::AppointmentPayloadValidator.new(appointment_params)
```
govindkrjoshi:
```suggestion
    validator = Api::V1::AppointmentPayloadValidator.new(appointment_params)
```

app/controllers/api/v1/appointments_controller.rb
```diff
@@ -1,4 +1,26 @@
...
+    logger.debug "Follow Up Schedule had errors: #{validator.errors_hash}" if validator.invalid?
+    record_params = Api::Current::Transformer.from_request(appointment_params)
```
govindkrjoshi:
```suggestion
    record_params = Api::V1::Transformer.from_request(appointment_params)
```

app/schema/api/v1/models.rb
```diff
@@ -27,6 +27,10 @@ def medical_history
       }
     end
 
+    def appointment
+      super.tap { |d| d[:properties][:cancel_reason][:enum] -= %w(invalid_phone_number public_hospital_transfer moved_to_private) }
```
govindkrjoshi: This might read better using `merge`?

app/tranformers/api/v1/appointment_transformer.rb
```diff
@@ -0,0 +1,15 @@
...
+  class << self
+    def new_keys_mapping
+      { invalid_phone_number: :other,
+        public_hospital_transfer: :other,
+        moved_to_private: :other }.with_indifferent_access
+    end
+
+    def to_response(model)
+      h =  model.attributes.with_indifferent_access
+      h[:cancel_reason] = new_keys_mapping[h[:cancel_reason]]
```
govindkrjoshi: Cancel reasons not part of new_keys_mapping will be transformed to nil.

app/controllers/api/v1/appointments_controller.rb
```diff
@@ -1,4 +1,26 @@
 class Api::V1::AppointmentsController < Api::Current::AppointmentsController
   include Api::V1::ApiControllerOverrides
   include Api::V1::SyncControllerOverrides
+
```
govindkrjoshi: We should also log to sentry when `other` cancel reason tries to overwrite a cancel reason. The chance of it happening is low, and we should capture it if it ever happens.

ssrihari: I'm explicitly catching cases where updates happen to cancelled appointments and returning errors. And we log to sentry when we return errors, ya?

app/tranformers/api/v1/appointment_transformer.rb
ssrihari: Oh ya. Will fix this.

app/schema/api/v1/models.rb
ssrihari: I'd appreciate a code suggestion here actually. I agree this isn't elegant, since `tap` is generic. But I'd like to express "I'm removing some values from a nested hash value".

PR #181, opened 2018-12-08
Associate admins to facility groups

govindkrjoshi:


app/controllers/admin/dashboard_controller.rb
```diff
@@ -4,43 +4,55 @@ def show
 
     Groupdate.time_zone = "New Delhi"
 
-    @users_requesting_approval = User.requested_sync_approval
-
-    @facilities = Facility.all.order(:name)
+    @users_requesting_approval = User.where(facility: current_admin.facility_groups.flat_map(&:facilities)).requested_sync_approval
```
divs1210: will requested_sync_approval work here?
it is a class method on User, but here it's being called on an array.

app/models/admin.rb
```diff
@@ -9,4 +9,7 @@ class Admin < ApplicationRecord
   ]
 
   validates :role, presence: true
+
+  has_many :admin_access_controls
+  has_many :facility_groups, through: :admin_access_controls
```
divs1210: make distinct

app/models/user.rb
```diff
@@ -106,4 +106,8 @@ def reset_password(password_digest)
   def registered_at_facility
     self.facility
   end
+
```
divs1210: use delegate instead

app/models/user.rb
```diff
@@ -106,4 +106,8 @@ def reset_password(password_digest)
...
+  def facility_group
```
divs1210: use delegate instead

app/controllers/admin/dashboard_controller.rb
```diff
@@ -4,43 +4,55 @@ def show
...
 
     @days_previous = 20
```
divs1210: duplicates, remove

PR #191, opened 2018-12-20
Fix: patient's registration_facility should not be nil

divs1210:


app/controllers/api_controller.rb
```diff
@@ -18,19 +18,22 @@ def current_user
   end
 
   def current_facility
-    @current_facility ||= Facility.find_by(id: request.headers['HTTP_X_FACILITY_ID'])
+    @current_facility = @current_facility ||
+                        Facility.find_by(id: request.headers['HTTP_X_FACILITY_ID']) ||
+                        (@current_user && Facility.find_by(id: @current_user.registration_facility_id))
```
govindkrjoshi: We should not do this. The fallback should only in patients controller.

app/controllers/api_controller.rb
```diff
@@ -18,19 +18,22 @@ def current_user
...
+    @current_facility&.facility_group
   end
 
   def validate_facility
     return head :bad_request unless current_facility.present?
   end
 
   def validate_current_facility_belongs_to_users_facility_group
-    return head :unauthorized unless current_user.present? && current_facility_group.facilities.where(id: current_facility.id).present?
+    return head :unauthorized unless current_user.present? &&
```
govindkrjoshi:
```suggestion
    unless current_user.present? && current_facility_group.facilities.where(id: current_facility.id).present?
      return  head :unauthorized
    end
```

PR #197, opened 2018-12-24
Associate facility groups with protocols

govindkrjoshi:
- Return protocol id with facility sync api

### DIFF
--- a/app/models/facility.rb
+++ b/app/models/facility.rb
@@ -18,7 +18,5 @@ class Facility < ApplicationRecord
   validates :country, presence: true
   validates :pin, numericality: true, allow_blank: true
 
-  def protocol
-    facility_group.protocol
-  end
+  delegate :protocol, to: :facility_group
 end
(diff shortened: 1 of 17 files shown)


app/models/facility.rb
```diff
@@ -17,4 +17,8 @@ class Facility < ApplicationRecord
   validates :state, presence: true
   validates :country, presence: true
   validates :pin, numericality: true, allow_blank: true
+
+  def protocol
+    facility_group.protocol
+  end
```
ssrihari: Can we use `delegate` here?

app/models/facility_group.rb
```diff
@@ -11,6 +11,8 @@ class FacilityGroup < ApplicationRecord
   has_many :medical_histories, through: :patients
   has_many :communications, through: :appointments
 
+  belongs_to :protocol
```
ssrihari: We should remember to move this to the admin notion of facility groups, when we separate those concerns.

PR #207, opened 2018-12-27
Sync only facilities with associated facility groups

govindkrjoshi:


app/controllers/api/v1/facilities_controller.rb
```diff
@@ -4,5 +4,6 @@ class Api::V1::FacilitiesController < Api::Current::FacilitiesController
 
   def find_records_to_sync(since, limit)
```
divs1210: should this not delegate to `Api::Current::FacilitiesController/records_to_sync`

divs1210: hmm, might need non-trivial refactor

govindkrjoshi: The timestamps are different. For current it’s `other_facilities_processed_since` and for v1 its `since`

PR #243, opened 2019-02-13
Follow ups v2

kitallis:
Continuing work for https://github.com/simpledotorg/simple-server/pull/229 here.

**TODO**

- [x] Rig the select button to submit the form on change
- [x] Hook up the select box to the controller to parse the input and save the correct fields to the appointment record
- [ ] Add a confirmation UI so the user knows their input was accepted and the patient is no longer in the list


divs1210:
new_ux

@timcheadle We have added a 'Remind in a Week' option to the call result dropdown.
- if that particular item is selected, `agreed_to_visit` is set to `true` with `remind_on` 1 week in the future, and will not be shown in this list till that date.
- If any other item is selected, the appointment is cancelled with that as the reason, and the appointment is removed from this list forever.

*NOTE* The page does not refresh automatically after the request is made, which leads to removed items still showing.

timcheadle:
I've added changes that reflect @dburka's designs and remove the update button. The form will submit on `<select>` field change

divs1210:
@timcheadle We had planned to not have the updte button initially, but then @kitallis pointed out that it might lead to mistakes when using on the phone if someone taps the dropdown list accidentally.

divs1210:
Also, an appointment can exist without a patient, and a patient can exist without a BP. We will add gaurds for this for now, but ideally they should be handled better.

divs1210:
Also, we are grouping by facility after paginating: might a facility show up multiple times on different pages?

PR #245, opened 2019-02-14
Add risk levels to the overdue list

kitallis:
* https://www.pivotaltracker.com/story/show/163861736
* https://www.pivotaltracker.com/story/show/163902317


app/models/appointment.rb
```diff
@@ -36,6 +36,13 @@ def self.overdue
     where(status: 'scheduled').where('scheduled_date <= ?', Date.today)
   end
 
+  def self.appointments_per_facility
```
timcheadle: I'd call this `group_by_facility`, as it's already in the `Appointment` class

app/models/appointment.rb
```diff
@@ -36,6 +36,13 @@ def self.overdue
...
+    includes(:facility,
+             :patient => [:address,
```
timcheadle: `includes(:patient)` (etc.) is irrelevant to this method. Let's let queries decide for themselves if they want to include patient/address/phone

app/views/layouts/application.html.erb
```diff
@@ -39,6 +39,7 @@
                   <%= link_to "Overdue Appointments", appointments_path, class: "nav-link #{active_controller?("patient_details")}" %>
                 </li>
               <% end %>
+              <% unless ['counsellor', 'analyst'].include?(current_admin.role)  %>
```
timcheadle: Let's add a helper method for this to `Admin`:

```ruby
  def has_role?(*roles)
    roles.map(&:to_sym).include?(self.role.to_sym)
  end

# example: Admin.first.has_role?(:owner, :supervisor)
```

timcheadle: I'll go ahead and add a commit that does this and includes specs

app/models/appointment.rb
kitallis: Ha, yeah. Just an artifact from before. Still a work in progress. I will change this name, it makes sense.

app/models/patient.rb
```diff
@@ -61,6 +92,13 @@ def current_age
     end
   end
 
+  private
+
+  def low_risk?
+    latest_scheduled_appointment.overdue_for_over_a_year? &&
```
timcheadle: Where did this definition come from?

app/views/layouts/application.html.erb
```diff
@@ -48,6 +48,7 @@
                 <% end %>
 
               <% end %>
+              <% unless ['counsellor', 'analyst'].include?(current_admin.role)  %>
```
timcheadle: You can use the new `has_role?` helper here instead

```ruby
<% unless current_admin.has_role?(:counsellor, :analyst) %>
```

spec/factories/blood_pressures.rb
```diff
@@ -9,6 +9,26 @@
     association :facility, strategy: :build
     association :patient, strategy: :build
     user
+
+    trait :critical do
+      systolic { rand(180..240) }
```
timcheadle: No need to make these random; just set them statically

spec/models/blood_pressure_spec.rb
```diff
@@ -53,6 +53,37 @@
       end
     end
 
+    describe "#critical?" do
+      [{ systolic: 181, diastolic: 111 },
+       { systolic: 181, diastolic: 109 },
+       { systolic: 179, diastolic: 111 }]
```
timcheadle: Style fix: put the `.each do |row|` on this line instead of below

app/models/medical_history.rb
```diff
@@ -26,4 +26,11 @@ class MedicalHistory < ApplicationRecord
   enum receiving_treatment_for_hypertension: MEDICAL_HISTORY_ANSWERS, _prefix: true
   enum diabetes: MEDICAL_HISTORY_ANSWERS, _prefix: true
   enum diagnosed_with_hypertension: MEDICAL_HISTORY_ANSWERS, _prefix: true
+
+  def risk_history?
```
timcheadle: This name confused me at first. What we're really describing is "does this patient have a history that make them at risk?"

Maybe `indicates_risk?` is better? What do you think?

kitallis: Yeah. That works. Had trouble naming this quite a bit, since `history` is already contextual.

spec/factories/blood_pressures.rb
kitallis: Makes sense.

app/models/patient.rb
kitallis: Roughly from here: https://github.com/simpledotorg/documentation/blob/master/simple.org%20major%20flows/high%20risk%20label.jpg

I'm now thinking that `low_priority` makes more sense?

Since, there's levels of priority which map onto risk. From "Very high" to "Low". They are computed by a combination of the patient's medical history, blood pressures and appointment information (specifically overdue appointments). The risk label is only shown when the priority is highest, otherwise the distinction is made only for sorting purposes.

app/models/appointment.rb
```diff
@@ -58,11 +58,21 @@ def days_overdue
     (Date.today - scheduled_date).to_i
   end
 
+  def scheduled?
+    status.to_sym == :scheduled
+  end
+
   def overdue?
-    status.to_sym == :scheduled && scheduled_date <= Date.today
+     scheduled? && scheduled_date <= Date.today
```
govindkrjoshi:
```suggestion
    scheduled? && scheduled_date <= Date.today
```

govindkrjoshi: Should this be a strict inequality?

app/models/patient.rb
```diff
@@ -3,11 +3,18 @@ class Patient < ApplicationRecord
 
   GENDERS = %w[male female transgender].freeze
   STATUSES = %w[active dead migrated unresponsive inactive].freeze
+  RISK_PRIORITIES = {
+    HIGHEST: 0,
```
govindkrjoshi: Why do we want the risk priorities to be numbers? Can they just be symbols?

app/models/patient.rb
```diff
@@ -49,7 +56,31 @@ def latest_scheduled_appointment
   end
 
   def latest_blood_pressure
-    latest_blood_pressures.first
+    blood_pressures.order(device_created_at: :desc).first
```
govindkrjoshi: `latest_blood_pressures` is used along with includes to eager load blood pressures. This will not be helpful in those cases no?

app/models/patient.rb
```diff
@@ -81,4 +106,17 @@ def call_result=(new_call_result)
 
     super(new_call_result)
   end
+
+  def self.not_contacted
```
govindkrjoshi: This can become a named scope.

app/models/patient.rb
kitallis: Having them as numbers makes it easier to sort them. Making them symbols will require specifying custom comparators.

app/models/appointment.rb
kitallis: I'm resolving this since it's unrelated to this PR.

app/models/patient.rb
kitallis: I'm resolving this since it's unrelated to this PR.

app/models/patient.rb
kitallis: I'm resolving this since it's unrelated to this PR.

govindkrjoshi:
Merging this since all the comments have been addressed.

PR #247, opened 2019-02-15
Make phone number comparision case insensitive for checking uniqness

govindkrjoshi:


kitallis:
I think adding a uniqueness migration for PG which is case-insensitive might be not straighforward.

PR #258, opened 2019-02-20
Format the way we display when last BP was recorded

kitallis:
https://www.pivotaltracker.com/story/show/164033384
```
Today if less than 24 hours
Yesterday if yesterday
X days ago if less than a year
30/2/2016 if more than a year
```

time


timcheadle:
This looks fine, but I wouldn't tie it to blood pressures or appointments. This is a generic helper for dates. Let's put it in `ApplicationHelper` and call it `rounded_time_ago_in_words` or something like that.

PR #261, opened 2019-02-25
A bin/setup script for easily setting up an environment. 

nas887:
### Problem 
Devs who are unfamiliar with rails environments often have difficulty spinning up new rails apps on their local machines. 

### Solution
This script fast forwards the basic problems of installing ruby (via both RBenv or RVM), installing postgres and installing any gems simple.


govindkrjoshi:
We use the gem [parallel_tests](https://github.com/grosser/parallel_tests) to run tests in parallel during development. Could you add setup instructions for this as well? 

On our dev machines, we run the following commands before running the migrations.
```
rake parallel:create
rake parallel:prepare
```

nas887:
@govindkrjoshi all set!

PR #264, opened 2019-02-28
[WIP] Correctly display BP counts for users who switch facilities

timcheadle:
**Story:** https://www.pivotaltracker.com/n/projects/2184102/stories/162240967

The dashboard currently has a bug where users are only displayed in the facility where they are registered, not where they record BPs. This means that if users switch facilities, their BP counts never appear on the dashboard.

This is fixed by looping through the list of all users who have **ever** recorded a BP in each facility, not just the users registered there.

Example of the fix in action:
image
(Notice how "Rosanna Ratke" appears in two places, as she has recorded BPs in two clinics)

### DIFF
--- a/app/controllers/facility_groups_controller.rb
+++ b/app/controllers/facility_groups_controller.rb
@@ -13,6 +13,7 @@ class FacilityGroupsController < AdminController
     @months_previous = 8
 
     @facilities = @facility_group.facilities
+    @users_by_facility = users_by_facility(@facility_group)
     @visits_by_facility = visits_by_facility(@facility_group)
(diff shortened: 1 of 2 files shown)


app/controllers/facility_groups_controller.rb
```diff
@@ -13,6 +13,7 @@ def show
     @months_previous = 8
 
     @facilities = @facility_group.facilities
+    @users_by_facility = users_by_facility(@facility_group)
```
divs1210:
```suggestion
    @users_by_facility_group = users_by_facility(@facility_group)
```

PR #266, opened 2019-02-28
Nurse reports styling

govindkrjoshi:
screenshot 2019-03-08 at 11 05 23


govindkrjoshi:
Updated the graphs to match styles from zeplin.

PR #276, opened 2019-03-05
Added rake task for anonymizing audit logs

timmyjose:
This is as part of https://www.pivotaltracker.com/n/projects/2184102/stories/164201136 to fix existing data which has appointments (created via Card Reader) incorrectly marked as "scheduled" instead of "visited".

### DIFF
--- a/lib/tasks/anonymize.rake
+++ b/lib/tasks/anonymize.rake
@@ -28,9 +28,13 @@ def whitelist_timestamps
   whitelist 'device_created_at', 'device_updated_at', 'created_at', 'updated_at'
 end
 
+def hashed_uuid(uuid)
+  UUIDTools::UUID.md5_create(UUIDTools::UUID_DNS_NAMESPACE, { uuid: uuid }.to_s).to_s
+end
+
 def anonymize_uuid(column)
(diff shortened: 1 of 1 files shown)


divs1210:
Why are we doing this?

timmyjose:
> Why are we doing this?

I have updated the description of the PR. This is as part of fixing the existing incorrectly marked Card Reader appointments. This task will anonymize the audit logs from production so that they can be used on the sandbox for testing the script that will update suitable appointments from "scheduled" to "visited".

The WIP part is that since we have a huge number of audit logs, we need to filter out the audit logs of type "fetch" since we are only interested in audit logs of type "create" and "update" as part of the clean-up process.

govindkrjoshi:
We want to test the rake task to update overdue cardreader appointments on sandbox before running it on production.

timmyjose:
The approach we have decided on is to use `skip` as provided by the `data-anonymization` gem to filter out the appointments of interest (action type `create` and `update`).

divs1210:
@govind We should probably rename the `:full_database` task at some point, or put everything under it, as our sandbox environment should have Users and AuditLogs to be fully testable.

PR #277, opened 2019-03-05
Handle pagination when 'All' is selected, but records.size is 0

divs1210:
https://www.pivotaltracker.com/story/show/164397979

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,6 @@ gem 'groupdate'
 gem 'data-anonymization', require: false
 gem 'uuidtools', require: false
 gem 'discard', '~> 1.0'
-gem 'chartkick'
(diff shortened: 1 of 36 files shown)


divs1210:
Though the error message seems to indicate that the `per` method should be passed a non-zero value, the actual failure seems to happen at the `<%= paginate @appointments %>` call site in the view.

govindkrjoshi:
@divs1210 We should show a message to the user when there are no appointments or patients.

divs1210:
> @divs1210 We should show a message to the user when there are no appointments or patients.

@govindkrjoshi that's already the case. This error happens when there are no results (and the UI says something like No Overdue Patients Found), and the user clicks on `All` under the pagination dropdown.

PR #280, opened 2019-03-07
Fixing case sensitivity spec failures

timmyjose:
This fixes the case sensitivity spec failures that were being observed in Semaphore.


divs1210:
please link to the story on pivotal in the description.

divs1210:
Will be good if we could add a test around this, as we merged a fix earlier but it didn't work.

PR #282, opened 2019-03-08
Exotel session creation endpoint for phone number masking

kitallis:
Relevant story: https://www.pivotaltracker.com/story/show/164462828

- [x] Expose a User-Patient session creation endpoint
- [x] Authorize the session creation by checking User / Patient registration
- [x] Authorize the session creation if the User isn't calling themselves
- [x] Create a session in-memory
- [x] Put this feature behind a feature flag

**What's missing and will come next in subsequent PRs:**

* Adding a connection attempt endpoint to respond with the Patient's phone number to Exotel
* Format the Patient phone number to a normalized schema
* Expire the User-Patient session when the call is complete


app/controllers/api/current/exotel_sessions_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Api::Current::ExotelSessionsController < ApplicationController
+  def passthru
+    session = ExotelSession.new(params[:From], parse_digits)
+
+    if session.save(params[:CallSid])
+      render plain: "OK", status: 200
```
divs1210: The more idiomatic way of doing this seems to be to use the [head](https://guides.rubyonrails.org/layouts_and_rendering.html#using-head-to-build-header-only-responses) method.

kitallis: Head appears to default to `text/html`. Probably `head` along with a default `:text` format in `routes.rb` is the cleanest way to go about.

divs1210:
Let's note down what's pending in this PR, and then start actually working.

app/controllers/api/current/exotel_sessions_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+  end
+
+  private
+
+  def parse_patient_phone_number
+    params[:digits].tr('"', '')
+  end
+
+  def valid_patient_phone_number?
```
divs1210: remove unused

app/models/exotel_session.rb
```diff
@@ -0,0 +1,30 @@
+class ExotelSession
+  KEY = 'EXOTEL-SESSION'
+
+  def initialize(user_phone_number, patient_phone_number)
+    @user = User.find_by_phone_number(user_phone_number)
+    @patient_phone_number = PatientPhoneNumber.find_by_number(patient_phone_number)
+  end
+
+  def save(call_id)
+    Rails.cache.write(session_key(call_id), record)
```
divs1210:
```suggestion
    Rails.cache.write(session_key(call_id), record, expires_in: 5.minutes)
```

config/environments/development.rb
```diff
@@ -23,7 +23,7 @@
   else
     config.action_controller.perform_caching = false
 
-    config.cache_store = :null_store
+    config.cache_store = :memory_store
```
govindkrjoshi: Let's make this an environment variable?

config/routes.rb
```diff
@@ -67,6 +67,12 @@
       get 'ping', to: 'pings#show'
       post 'login', to: 'logins#login_user'
 
+      if FeatureToggle.enabled?('PHONE_NUMBER_MASKING')
+        scope :exotel_sessions do
+          get 'create', to: 'exotel_sessions#create'
```
govindkrjoshi: Seeing `get` with a `create` action is weird. I think we should leave a comment here say it's because of exotel calls using get.

app/controllers/api/current/exotel_sessions_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    parse_patient_phone_number.scan(/\D/).empty?
+  end
+
+  def report_http_status(api_name, status)
+    NewRelic::Agent.increment_metric("ExotelSessions/#{api_name}/#{status.to_s}")
+  end
+
+  def respond_in_plain_text(api_name, status)
+    report_http_status(api_name, status)
```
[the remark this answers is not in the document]
kitallis: Had no idea about `action_name`: thanks!

app/models/exotel_session.rb
```diff
@@ -0,0 +1,31 @@
+class ExotelSession
```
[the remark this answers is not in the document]
divs1210: I remember having a discussion around this, and deciding to keep this very explicitly specific to Exotel.

config/environments/development.rb
kitallis: We'll make this an environment variable once we introduce redis to this.

kitallis:
@timcheadle Could you help review this? This is a first in the series of this implementation and we would like it to be reviewed and merged before we commit further.

PR #290, opened 2019-03-15
Phone number masking connect and terminate endpoints

divs1210:
https://www.pivotaltracker.com/story/show/164462839
https://www.pivotaltracker.com/story/show/164657741


divs1210:
A known (unrelated) flaky test seems to be failing over and over.

app/controllers/api/current/exotel_call_sessions_controller.rb
```diff
@@ -25,11 +47,16 @@ def valid_patient_phone_number?
     parse_patient_phone_number.scan(/\D/).empty?
   end
 
-  def respond_in_plain_text(status)
-    head status, content_type: 'text/plain'
+  def respond_in_plain_text(status, text = '')
```
govindkrjoshi: Do we need this method anymore? Can we inline it?

app/controllers/api/current/exotel_call_sessions_controller.rb
```diff
@@ -6,15 +6,37 @@ def create
...
+      respond_in_plain_text(:not_found)
+    end
+  end
+
+  def terminate
+    session = CallSession.fetch(params[:CallSid])
+
+    if session.present?
+      session.kill
+      report_call_info
```
govindkrjoshi: This can be an after_action hook

app/models/call_session.rb
```diff
@@ -1,33 +1,53 @@
 class CallSession
-  KEY = 'CALL-SESSION'
   EXPIRE_CALL_SESSION_IN = 24.hours
+  COUNTRY = 'IN'
```
govindkrjoshi: This can be a config variable, maybe?

app/models/call_session.rb
```diff
@@ -1,33 +1,53 @@
...
+  def initialize(call_id, user_phone_number, patient_phone_number)
+    @call_id = call_id
+    @user = User.find_by_phone_number(sanitized_phone_number(user_phone_number))
     @patient_phone_number = PatientPhoneNumber.find_by_number(patient_phone_number)
   end
 
-  def save(call_id)
-    Rails.cache.write(session_key(call_id),
+  def authorized?
+    @user.present? &&
```
govindkrjoshi: If a user is not approved to sync, can they still make calls? 
`authorized?` will currently allow that, we should check if this is okay.

app/models/call_session.rb
```diff
@@ -1,33 +1,53 @@
...
+      patient_phone_number.present? &&
+      @user.phone_number != patient_phone_number.number
+  end
+
+  def save
+    Rails.cache.write(CallSession.session_key(@call_id),
```
govindkrjoshi: We should keep this outside the rails cache. A better option would be to use a redis store.

app/models/call_session.rb
```diff
@@ -1,33 +1,53 @@
...
+    def fetch(call_id)
+      data = Rails.cache.fetch(session_key(call_id))
+
+      CallSession.new(call_id,
+                      data[:user_phone_number],
+                      data[:patient_phone_number]) if data.present?
+    end
+
+    def session_key(call_id)
+      [name, call_id].join('/')
```
govindkrjoshi: What is `name`?

spec/models/call_session_spec.rb
```diff
@@ -1,36 +1,101 @@
...
+
+    it 'should return a falsey value if a session does not exist against the call id' do
+      call_id = SecureRandom.uuid
+      expected_session = CallSession.new(call_id, user.phone_number, patient.phone_numbers.first.number)
+
+      expect(expected_session.kill).to be_falsey
+    end
+  end
+
+  describe '.fetch' do
```
govindkrjoshi:
```suggestion
  describe '#fetch' do
```

app/models/call_session.rb
kitallis: Yeah, we want to do this too. We'll keep this separate from `Rails.cache` when we run the Redis PR.

app/models/call_session.rb
kitallis: Hm, maybe this should be `self.name` for clarity. This is a class method you can call for any class that gets you the name of the class in String.

spec/models/call_session_spec.rb
kitallis: This is a class method so I think we do `.`s for those?

app/controllers/api/current/exotel_call_sessions_controller.rb
kitallis: I think this method is nice because it sort of weakly encapsulates the whole ExotelController to ensure that it is always in 'text/plain'.

app/controllers/api/current/exotel_call_sessions_controller.rb
divs1210: We had this as an after_action, but ended up changing it because:
* only used by one method
* used only if `session.present?` is `true`

app/models/call_session.rb
divs1210: We will move to Redis in the future. Will need changes to infra, terraform, etc. before we can do that.

spec/models/call_session_spec.rb
divs1210: `fetch` is a class method.

app/models/call_session.rb
kitallis: This is a good point. We'll add this in.

PR #291, opened 2019-03-15
Allow downloading the overdue list as a CSV

timcheadle:
**Story:** https://www.pivotaltracker.com/story/show/164245733

Allow supervisors to download the overdue list as a CSV.

- [x] Display a "you're about to download PHI" warning before downloading


app/views/appointments/index.csv.erb
```diff
@@ -0,0 +1,30 @@
...
+  "Age",
+  "Days overdue",
+  "Last BP",
+  "Last BP taken at",
+  "Last BP date",
+  "Risk level",
+  "Patient address",
+  "Patient village or colony",
+  "Patient phone"
+].to_csv -%>
```
divs1210: Looks like `to_csv` adds a `\n` at the end of the line by default. This should be changed to `to_csv( row_sep: nil )`.

app/views/appointments/index.csv.erb
```diff
@@ -0,0 +1,30 @@
...
+<%- @appointments.group_by { |a| a.patient.latest_blood_pressure.facility }.each do |facility, facility_appointments| %>
```
divs1210: There are a couple of `<%-` and `-%>` ligatures.
Should these `-`s be `=`s instead?

config/locales/en.yml
```diff
@@ -62,6 +62,7 @@ en:
     allowed_access_to_user: '%{admin_name} allowed access'
     denied_access_to_user: '%{admin_name} denied access'
     dashboard_title: "Simple Server"
+    phi_access_alert: "You are about to view confidental patient data. Are you sure you want to proceed?"
```
divs1210: I think `about to view confidential data` should be changed to `about to download confidential data`, as they are already viewing the data when they click the button. This will make it clearer that they are going to get a copy of confidential data on their machine.

app/views/appointments/index.csv.erb
timcheadle: These suppress printing of newlines and whitespace in ERB:
https://stackoverflow.com/questions/7996695/what-is-the-difference-between-and-in-erb-in-rails

app/models/patient_phone_number.rb
```diff
@@ -7,4 +7,12 @@ class PatientPhoneNumber < ApplicationRecord
 
   validates :device_created_at, presence: true
   validates :device_updated_at, presence: true
+
+  def self.first
+    order(device_created_at: :asc).first
+  end
+
+  def self.last
+    order(device_created_at: :asc).last
```
divs1210: `order(device_created_at: :desc).first`

divs1210:
Error when attempting to download data for 'All Facilities'
Screenshot 2019-03-29 at 5 23 58 PM

app/models/patient_phone_number.rb
timcheadle: Having a method `last` call `first` was confusing to me, so made it match.

timcheadle:
@divs1210 that seems to indicate you have an overdue patient without a most recent BP... How does that happen?

app/models/patient.rb
```diff
@@ -78,6 +78,15 @@ def risk_priority
     end
   end
 
+  def risk_priority_label
+    case risk_priority
```
govindkrjoshi: This can be rewritten as a hash look up instead.

divs1210:
> The download button currently downloads all the overdue patients for the facility. 

Had the same question, but it also makes sense to just download the entire CSV.
This is mergeable in the current state.

divs1210:
Not merging this at the moment. Waiting for:
- [] disable for counsellors
- [] handle pagination

PR #292, opened 2019-03-18
Add patient already visited option to overdue list

timmyjose:
For the story: [Add Patient has already visited option to the overdue list](https://www.pivotaltracker.com/n/projects/2184102/stories/164569839)


timmyjose:
Specific change(s) that I would like to have particularly reviewed:
 - in the case of `patient_already_visited`, I am also updating the `agreed_to_visit` field to `nil` (since it might have been previously set to `true`, and setting that to `false` is not technically correct), and for extra measure, I also update `remind_on` to `nil`.

app/models/appointment.rb
```diff
@@ -35,10 +35,14 @@ def self.overdue
   end
 
   def call_result=(new_call_result)
-    if new_call_result == "agreed_to_visit"
+    if new_call_result == 'agreed_to_visit'
```
govindkrjoshi: Let's pull this out into proper methods. Using attribute setters way is a code smell.

timmyjose: Extracted logic for each path into separate private methods.

app/models/appointment.rb
```diff
@@ -76,7 +54,34 @@ def overdue_for_under_a_month?
...
+
+  def mark_patient_already_visited
+    self.status = :visited
+    self.agreed_to_visit = nil
+    self.remind_on = nil
+  end
+
+  def mark_patient_as_dead
+    self.patient.status = :dead
+    self.patient.save
```
divs1210: do we need this call to `save`?

app/controllers/appointments_controller.rb
```diff
@@ -34,4 +36,22 @@ def set_appointment
...
+    elsif call_result == :patient_has_already_visited
+      appointment.mark_patient_already_visited
+    elsif call_result == :remind_to_call_later
+      appointment.mark_remind_to_call_later
+    elsif Appointment.cancel_reasons.values.include? call_result.to_s
+      appointment.mark_appointment_cancelled(call_result)
+    end
+
+    if call_result == :dead
+      appointment.patient.status = appointment.mark_patient_as_dead
```
divs1210: `appointment.patient.status = 'dead'`

PR #295, opened 2019-03-19
Update nurse reports in progress tab

govindkrjoshi:
Screenshot 2019-03-21 at 15 58 40

Screenshot 2019-03-21 at 15 18 55


spec/controllers/api/current/analytics/user_analytics_controller_spec.rb
```diff
@@ -1,45 +1,85 @@
...
     request.env['HTTP_AUTHORIZATION'] = "Bearer #{request_user.access_token}"
   end
 
-  describe 'GET: send data from server to device;' do
-    it 'gets data for 4 weeks as a hashmap' do
-      get :show, format: :json
+  describe '#show' do
+    context 'json_api' do
+      describe 'facility has no patients registered' do
+        it 'gets data for as a hashmap' do
```
kitallis: This asserts for `nil` and not a hashmap

spec/controllers/api/current/analytics/user_analytics_controller_spec.rb
```diff
@@ -1,45 +1,85 @@
...
 
-    it 'has the facility name' do
-      get :show, format: :html
+    context 'html api' do
+      render_views
 
-      expect(response.body).to match(Regexp.new(request_facility.name))
-    end
+      describe 'facility has no patients registered' do
+        it 'returns an empty graph' do
```
kitallis: These tests can maybe just be if the correct templates / partials are being rendered instead of a piece of some content from each view?

app/controllers/api/current/analytics/user_analytics_controller.rb
```diff
@@ -1,58 +1,57 @@
...
+  def unique_patients_recorded_per_month
+    BloodPressure.where(facility: current_facility)
+      .group_by_month(:device_created_at, last: MONTHS_TO_REPORT, reverse: true)
+      .count('distinct patient_id')
+      .select { |k, v| k >= first_patient_at_facility.device_created_at.at_beginning_of_month }
   end
 
-  def graph_label(value, from_date_string, to_date_string)
-    "<div class='graph-label'><p>#{from_date_string}</p><p>#{to_date_string}</p>".html_safe
+  def patients_enrolled_per_month
```
kitallis: Can maybe move these to `UserAnalytics` model and test the queries separately.

app/views/api/current/analytics/user_analytics/show.html.erb
```diff
@@ -7,150 +7,39 @@
...
+<% else %>
+  <div class="card">
+    <%= inline_svg('icon_new_bp.svg', classname: 'icon-float') %>
+    <h3>How many patients have a BP recorded each month?</h3>
+    <h5><%= @current_facility.name %></h5>
+    <%= render 'graph',
+               id: 'return-patients',
+               statistics: @statistics[:unique_patients_per_month],
+               first_of_current_month: @statistics[:first_of_current_month],
+               max_value: @statistics[:unique_patients_per_month].values.max %>
```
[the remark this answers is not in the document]
govindkrjoshi: :+1: Updated.

spec/controllers/api/current/analytics/user_analytics_controller_spec.rb
govindkrjoshi: Fixed

spec/controllers/api/current/analytics/user_analytics_controller_spec.rb
govindkrjoshi: Added https://github.com/rails/rails-controller-testing for this. 
Updated the spec.

app/controllers/api/current/analytics/user_analytics_controller.rb
govindkrjoshi: This doesn't seem very straight forward right now. Will do this as a separate PR.

PR #297, opened 2019-03-20
Improve admin navigation

timcheadle:
**Story:** https://www.pivotaltracker.com/story/show/164758499

These changes are intended to make administration easier and more intuitive.

- Reduce org/group/facility management down to "Facilities" UI
- Improve user management, group by facility groups
- Remove things like facility and user deletion that aren't well supported today (or would break things)


divs1210:
As these are individual tables, the columns are misaligned and look off.
It might be better to explicitly set the spacing for these tables.

Screenshot 2019-04-05 at 6 14 52 PM

PR #298, opened 2019-03-20
Added help docs and API endpoint

timcheadle:
**Story:** https://www.pivotaltracker.com/story/show/164760139

Add basic help documentation HTML endpoint

To do:
- [x] Simplify this by publishing it as a basic view, not requiring authentication
- [x] Add a basic feature test to make sure it renders as expected


govindkrjoshi:
We should move the svgs to assets and use inline_svg helper from AssetHelper

govindkrjoshi:
We should add internationalizing support to the help screen HTML. It’ll much easier to translate to users preferred language when we start supporting it.

timcheadle:
We can use localized views in this case, as adding these giant strings to our locales is messy. We can do that later. I've moved the SVGs to the asset pipeline.

PR #303, opened 2019-03-22
Added column for appointment_type; defined enum with same name in model

timmyjose:
Story: https://www.pivotaltracker.com/n/projects/2184102/stories/164570825
Task: Task 1 (Add a column to mark appointment as "manual" or "automatic")


app/models/appointment.rb
```diff
@@ -22,6 +22,11 @@ class Appointment < ApplicationRecord
     other: 'other'
   }
 
+  enum appointment_type: {
+    manual: 'manual',
+    automatic: 'automatic'
+  }
```
govindkrjoshi:
```suggestion
  }, _prefiix: true
```

timmyjose: Added `_prefix: true` to the `appointment_type` enum.

PR #308, opened 2019-03-22
Add views for facility analytics

govindkrjoshi:


app/controllers/analytics_controller.rb
```diff
@@ -0,0 +1,28 @@
+class AnalyticsController < AdminController
```
kitallis: Do you think a better name might be `AnalyticsBaseController`? Since it's only purpose is to be a base class?

app/views/analytics/facility_groups/show.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    in the past 3 months.</p>
+  <p class="footnote">
+    <strong><%= t('analytics.control_rate') %></strong>
+    <%= t('analytics.control_rate_explanation',
+          hypertensive_patients_in_cohort: @facility_group_analytics[:control_rate][:hypertensive_patients_in_cohort],
+          patients_under_control: @facility_group_analytics[:control_rate][:patients_under_control_in_period],
+          control_rate_for_month: @facility_group_analytics[:control_rate][:control_rate],
+          number_of_months: pluralize(1, 'month')) %>
+  </p>
+  <p class="footnote"><strong>Top performers:</strong> > 100 enrolled patients and highest BP control.</p>
```
kitallis: Minor space issue: before ` >` after `</strong>`

app/views/shared/analytics/_patients_count_panel.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="featured-section">
+  <strong class="featured-title" style="color: #007A31;">
+    <%= t('analytics.newly_enrolled') %>
+  </strong>
+  <span id='newly-enrolled-patients-count' class="featured-number text-green">
+      <%= newly_enrolled_patients_count %>
+    </span>
```
kitallis: Some strange indentation here.

app/views/shared/analytics/_patients_count_panel.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <span class="featured-label">New patients</span>
+
+  <strong class="featured-title" style="color: #007A31;">
+    <%= t('analytics.return_patients') %>
+  </strong>
+  <span id='returning-patients-count' class="featured-number text-green">
+      <%= returning_patients_count %>
+    </span>
```
kitallis: Some strange indentation here.

spec/models/analytics/patient_set_analytics_spec.rb
```diff
@@ -119,6 +118,7 @@
         Date.new(2019, 3, 03) => 0,
         Date.new(2019, 3, 10) => 0,
         Date.new(2019, 3, 17) => 0,
+        Date.new(2019, 3, 24) => 0,
```
kitallis: These tests have been failing on `master` since last week. I don't think they correctly respecting time right now. Can we nest them under having them frozen in time for the time they span?

app/controllers/analytics_controller.rb
govindkrjoshi: For consistency with ApiController and AdminController, we should keep it as AnalyticsController

app/views/analytics/facility_groups/show.html.erb
govindkrjoshi: Changed `>` to `&gt;` to be more explicit.

PR #310, opened 2019-03-24
Add connection pooling and redis store for CallSessions

kitallis:
[Story on Pivotal](https://www.pivotaltracker.com/story/show/164567768)


app/models/exotel_api.rb
```diff
@@ -0,0 +1,51 @@
+class ExotelAPI
```
govindkrjoshi: This belongs to `app/services/`

divs1210: done

app/models/call_session.rb
```diff
@@ -17,18 +17,28 @@ def authorized?
...
+        .hmset_with_expiry(CallSession.session_key(@call_id),
+                           session_data,
+                           EXPIRE_CALL_SESSION_IN_SECONDS)
+    end
   end
 
   def kill
-    Rails.cache.delete(CallSession.session_key(@call_id))
+    CALL_SESSION_STORE_POOL.with do |connection|
+      connection.del(CallSession.session_key(@call_id))
```
govindkrjoshi: This can also be moved to `RedisService`

config/initializers/call_session_store.rb
```diff
@@ -0,0 +1,13 @@
+require 'connection_pool'
+
+DEFAULT_CALL_SESSION_REDIS_POOL_SIZE = 12
+DEFAULT_CALL_SESSION_REDIS_TIMEOUT_MS = 1000
+
+connection_parameters = {
+  size: ENV['CALL_SESSION_REDIS_POOL_SIZE'].to_i || DEFAULT_CALL_SESSION_REDIS_POOL_SIZE,
+  timeout: (ENV['CALL_SESSION_REDIS_TIMEOUT_MS'].to_i || DEFAULT_CALL_SESSION_REDIS_TIMEOUT_MS) / 1000
```
govindkrjoshi: We can just mention the seconds to be simpler no? `/ 1000` will anyways return a whole number.

app/models/call_session.rb
kitallis: Done.

config/initializers/call_session_store.rb
kitallis: Good point.

PR #278, opened 2019-03-06
Add Patient Set Analytics

govindkrjoshi:
Queries and models for calculating Patient Analytics

PR #312, opened 2019-03-26
Fix time-sensitive analytics (and/or specs)

timcheadle:
The current `patient_set_analytics` specs are incomplete and broken. The class requires a `from_time` and `to_time`, but the methods that report analytics for months and weeks previous don't pay attention to these times.

As a result, the specs only work on a given date. We can fix the specs by time traveling to that date, but the better answer is to update the analytics classes to respect the `to_time` or at least be more explicit in what they're returning.

Also, there is no spec for `control_rate_per_month` so we should add that.

cc @govindkrjoshi (this was introduced in #278)

### DIFF
--- a/spec/models/analytics/patient_set_analytics_spec.rb
+++ b/spec/models/analytics/patient_set_analytics_spec.rb
@@ -104,8 +104,9 @@ RSpec.describe Analytics::PatientSetAnalytics do
     end
   end
 
-  describe '#blood_pressure_recored_per_week' do
+  describe '#blood_pressure_recorded_per_week' do
     it 'returns the number of blood pressures recorded per week for a group of patients' do
+      Timecop.travel(Date.new(2019, 3, 21))
       expected_counts = {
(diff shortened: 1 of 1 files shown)


timcheadle:
I've included an example commit of "fixing" the `blood_pressure_recorded_per_week` spec, but it really just shows how it's broken.

kitallis:
> I've included an example commit of "fixing" the `blood_pressure_recorded_per_week` spec, but it really just shows how it's broken.

@timcheadle We've fixed this (and a couple of other similar instances) in this commit: https://github.com/simpledotorg/simple-server/blob/0b84cf55ca66111389aff8a7039cb57b0c5c9962/app/models/analytics/patient_set_analytics.rb#L61

Thanks for pointing this out, closing this PR.

kitallis:
@timcheadle There were some related tests that would fail eventually in a similar way. @govindkrjoshi has fixed those here: https://github.com/simpledotorg/simple-server/pull/316

PR #316, opened 2019-03-26
Fix range queries grouping data by weeks and months

govindkrjoshi:


govindkrjoshi:
The specs needed some fixing as well. I've updated the specs.

PR #320, opened 2019-03-30
Add a job to warmup analytics cache

govindkrjoshi:


app/jobs/warm_up_facility_group_analytics_cache_job.rb
```diff
@@ -0,0 +1,12 @@
+class WarmUpFacilityGroupAnalyticsCacheJob < ApplicationJob
+  queue_as :default
+
+  def perform(facility_group, from_time_string, to_time_string)
```
kitallis: It's better to pass in IDs rather than full ruby objects to Jobs typically [1]. The ID then can be queried from within the job itself.

[1] https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple

spec/jobs/warm_up_facility_group_analytics_cache_job_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    let(:facility_group) { create(:facility_group) }
+    let!(:facilities) { create_list(:facility, 2, facility_group: facility_group) }
+
+    let(:from_time) { Date.new(2019, 1, 1).strftime('%Y-%m-%d') }
+    let(:to_time) { Date.new(2019, 3, 31).strftime('%Y-%m-%d') }
+
+    let(:job) { WarmUpFacilityGroupAnalyticsCacheJob.perform_later(facility_group, from_time, to_time) }
+
+    it 'queues the job' do
+      expect { job }.to change(ActiveJob::Base.queue_adapter.enqueued_jobs, :size).by(1)
```
kitallis: This helper can be used here instead: https://apidock.com/rails/v4.2.7/ActiveJob/TestHelper/assert_enqueued_jobs

config/environments/production.rb
```diff
@@ -58,7 +58,7 @@
   # config.cache_store = :mem_cache_store
 
   # Use a real queuing backend for Active Job (and separate queues per environment)
-  # config.active_job.queue_adapter     = :resque
+  config.active_job.queue_adapter     = :sidekiq
```
kitallis: Does sidekiq work without a configured redis?

app/jobs/warm_up_quarterly_analytics_cache_job.rb
```diff
@@ -0,0 +1,15 @@
+class WarmUpQuarterlyAnalyticsCacheJob < ApplicationJob
+  queue_as :default
+
+  def perform
```
kitallis: Do you think it's better to write this as `[-2, -3, -4, -5].each`? I'm not sure if `-1 * (n + 1)` is obvious when reading.

spec/jobs/warm_up_analytics_cache_job_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  let!(:facility_groups) { create_list(:facility_group, 2) }
+  let(:to_time) { Time.new(2019, 3, 31).strftime('%Y-%m-%d') }
+  let(:from_time) { (to_time.to_time - 90.days).strftime('%Y-%m-%d') }
+
+  describe '.perform_later' do
+    let(:job) { WarmUpAnalyticsCacheJob.perform_later }
+
+    it 'queues the job' do
+      expect { job }.to change(ActiveJob::Base.queue_adapter.enqueued_jobs, :size).by(1)
```
kitallis: This helper can be used here instead: https://apidock.com/rails/v4.2.7/ActiveJob/TestHelper/assert_enqueued_jobs

app/jobs/warm_up_facility_group_analytics_cache_job.rb
govindkrjoshi: The activejob guide says it's safe to pass ActiveRecord objects to jobs, since they are passed as GlobalIds https://guides.rubyonrails.org/active_job_basics.html#globalid 

I can change it simply accept an ID if you think that's more explicit.

timcheadle: Yes! GlobalID is one of those nice things in Rails 5 :)

app/jobs/warm_up_quarterly_analytics_cache_job.rb
timcheadle: Agreed; just put the explicit values in the loop.

config/environments/production.rb
```diff
@@ -59,7 +59,7 @@
     { host: ENV['RAILS_CACHE_REDIS_URL'] }
 
   # Use a real queuing backend for Active Job (and separate queues per environment)
-  config.active_job.queue_adapter      = :sidekiq
+  config.active_job.queue_adapter     = :inline
```
timcheadle: Why is this not sidekiq in production?

config/schedule.rb
```diff
@@ -0,0 +1,28 @@
...
+#   rake "some:great:rake:task"
+# end
+#
+# every 4.days do
+#   runner "AnotherModel.prune_old_records"
+# end
+
+# Learn more: http://github.com/javan/whenever
+
+every :day, at: ['1:00 am'] do
```
timcheadle: This will run at 1am UTC. Is that desired? Maybe we should put timezone in here.

PR #321, opened 2019-03-31
Fixes to new analytics dashboard

govindkrjoshi:


Gemfile
```diff
@@ -36,6 +36,7 @@ gem 'http'
 gem 'sidekiq'
 gem 'connection_pool'
 gem 'redis'
+gem 'redis-rails'
```
govindkrjoshi: Rails 5.1 does not have an inbuilt redis-cache-store. 

https://github.com/redis-store/redis-rails#a-quick-note-about-rails-52

PR #325, opened 2019-04-01
Adding patient record creation date as enrolment date for adherence tab

timmyjose:
Story: https://www.pivotaltracker.com/n/projects/2184102/stories/164885867


timmyjose:
We can simply get rid of the `show_enrolment_date` field and have the enrolment date show up on both the Overdue and Adherence tabs, but I feel that showing it in the Overdue tab as a side-effect of showing it in the Adherence tab is incorrect. This will ensure that the registration date is shown only in the Adherence tab.

PR #315, opened 2019-03-26
Adding Rake task for creating automatic appointments for defaulters

timmyjose:
Link to story: https://www.pivotaltracker.com/n/projects/2184102/stories/164570825
Task 4: Create a Rake task to create "automatic" appointments for defaulters with the `facility_id` set to the id of facility where the latest BP reading was taken.

PR #327, opened 2019-04-02
Adding non-null constraint on appointment_type

timmyjose:
Story: https://www.pivotaltracker.com/n/projects/2184102/stories/164570825
Task: 3 (Create a migration to mark the appointment_type column as non-nullable.)

Dependent on: #315


timmyjose:
Note: The failing tests will be resolved as soon as the dependencies are merged in and the Rake task (to mark `appointment_type` as `manual` is run).

app/controllers/api/current/appointments_controller.rb
```diff
@@ -18,6 +24,7 @@ def merge_if_valid(appointment_params)
       NewRelic::Agent.increment_metric('Merge/Appointment/schema_invalid')
       { errors_hash: validator.errors_hash }
     else
+      set_default_appointment_type(appointment_params)
       appointment = Appointment.merge(Api::Current::Transformer.from_request(appointment_params))
```
divs1210: this should happen only if the record is new

app/controllers/api/v1/appointments_controller.rb
```diff
@@ -15,6 +15,7 @@ def merge_if_valid(appointment_params)
       NewRelic::Agent.increment_metric('Merge/Appointment/invalid_request')
       { errors_hash: { updated_at: 'Cancelled appointment cannot be updated', id: appointment_params[:id] } }
     else
+      set_default_appointment_type(record_params)
       appointment = Appointment.merge(record_params)
```
divs1210: this should happen only if the record is new

timmyjose:
This merge should go along with the API version update so that appointments which are created in the interim do not fail. Will sync merge accordingly.

app/controllers/api/v1/appointments_controller.rb
timmyjose: Implemented change.

app/controllers/api/current/appointments_controller.rb
timmyjose: Implemented change.

timmyjose:
> we want to set default type only if the record is new

Updated - now we check if the payload does not contain `appointment_type` and also if the record in `new`, and only then update the payload hash before merging.

PR #328, opened 2019-04-02
Added Appointment Transformer for current (v2) API version for `appointment_type` 

timmyjose:
Story (Chore): https://www.pivotaltracker.com/n/projects/2184102/stories/165055118

### DIFF
--- a/app/tranformers/api/v1/appointment_transformer.rb
+++ b/app/tranformers/api/v1/appointment_transformer.rb
@@ -1,4 +1,4 @@
-class Api::V1::AppointmentTransformer < Api::Current::Transformer
+class Api::V1::AppointmentTransformer < Api::Current::AppointmentTransformer
   class << self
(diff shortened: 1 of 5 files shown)


app/tranformers/api/v1/appointment_transformer.rb
```diff
@@ -7,7 +7,7 @@ def new_keys_mapping
     end
 
     def to_response(model)
-      h = rename_attributes(model.attributes, inverted_key_mapping)
+      h =  super(model).with_indifferent_access
```
divs1210:
```suggestion
      h =  model.with_indifferent_access
```
If I' understanding what `super(model)` returns correctly

divs1210: should we move `.with_indifferent_access` to the super method instead

PR #335, opened 2019-04-04
Improved the formatting of the approval list on the home screen

dburka:
Some simple css and text formatting changes to make the approvals list less ass-like.


timcheadle:
Let's wait until #297 is merged before putting this in as I think they conflict.

PR #341, opened 2019-04-08
Fix recorded patients count on the dashboard

govindkrjoshi:
The column currently show the all time patients recorded count. This is incorrect.

The column is meant to show the count of people who were recorded in that period. 
Which is the number of returning patients + number of newly enrolled patients


app/views/analytics/facility_groups/_facility_table.html.erb
```diff
@@ -51,7 +53,9 @@
 
         <td class="row-control"><%= @facility_group_analytics[:control_rate][:control_rate] %></td>
         <td class="row-spacer">&nbsp;</td>
-        <td class="row-unique-date"><%= @facility_group_analytics[:unique_patients_enrolled] %></td>
+        <td class="row-unique-date">
+          <%= @facility_group_analytics[:returning_patients] + @facility_group_analytics[:newly_enrolled_patients] %>
```
arku: Nit. Can we pass the data to the partial explicitly via local variables and remove the instance variable references?

This is definitely not a blocker for this bugfix but would be nice to get done in a separate PR.

PR #342, opened 2019-04-08
Soft delete for facility groups (if empty)

timmyjose:
Story: https://www.pivotaltracker.com/n/projects/2184102/stories/164461475
Depends on: https://github.com/simpledotorg/simple-server/pull/297


app/models/facility_group.rb
```diff
@@ -1,5 +1,10 @@
 class FacilityGroup < ApplicationRecord
   include PatientSetAnalyticsReportable
+
+  include Discard::Model
```
divs1210: this seems to already be coming from ApplicationRecord

timmyjose: You are right. I have added the `scope` to `ApplicationRecord` and removed the other changes from `FacilityGroup`.

app/views/admin/facilities/index.html.erb
```diff
@@ -23,7 +23,7 @@
     </div>
 
     <p>
-        Protocol: <%= facility_group.protocol.name %>
+        Protocol: <%= facility_group.protocol&.name %>
```
govindkrjoshi: `&.` is unnecessary. 

We should also avoid using `&.` and instead handle the nil case as a condition.

timmyjose: Removed the `nil` punning since it is logically impossible for the protocol to be `nil`.

app/controllers/admin/facility_groups_controller.rb
```diff
@@ -41,7 +41,7 @@ def update
   end
 
   def destroy
-    @facility_group.destroy
+    @facility_group.discard
```
divs1210: Should there be a check to see if the group is empty (has no facilities) ?

timmyjose: @divs1210 Added destroyability checks for FG in the controller.

PR #343, opened 2019-04-08
Dashboard graphs improvements

timcheadle:
Tweaks & fixes:

- [x] Removed bar charts
- [x] Use dashes instead of zeroes in tabular data
- [x] Remove unused metrics in footnotes ("Top performer")
- [x] Make sure current range is always underlined in links
- [x] Show current time range under group & facility names


govindkrjoshi:
Most of these changes are already on master. Shall we close this PR?

timcheadle:
@govindkrjoshi @kitallis I think some of this is still necessary, no?

For example, it's preferable to comment out the bar charts at the top, instead of rendering it then hiding it with CSS.

PR #344, opened 2019-04-09
[Bugfix] Fix cron log file path and sidekiq concurrnecy settings

govindkrjoshi:


config/sidekiq.yml
```diff
@@ -1,4 +1,4 @@
-:concurrency: <%= ENV['SIDEKIQ_CONCURRENCY'] %>
```
govindkrjoshi: Dotenv doesn't load config before the sidekiq.yml file is read. This results in `ENV['SIDEKIQ_CONCURRENCY']` evaluating to nil here. 

https://github.com/mperham/sidekiq/issues/3090

PR #345, opened 2019-04-09
User time range in user analytics

govindkrjoshi:
Currently UserAnalytics does not use `from_time` and `to_time` to calculate some of the statistics. 

This PR fixes that.


divs1210:
Hi @govindkrjoshi, what is this PR for?

PR #348, opened 2019-04-10
Add generator for migrating to newer API version

divs1210:
**Spec directories to copy**
- [x] Api
- [x] Controllers
- [x] Payloads
- [x] Requests

**App directories to copy**
- [x] View/api

**Create templatized versions of**
- [x] Controller/api
- [x] Schema
- [x] Transformers
- [x] Validators

**Files to update**
 - [x] routes.rb - It does not modify the routes file, instead asks the user to update it.


divs1210:
We should probably add the USAGE stuff to Readme, and put in a warning about things like `current_month` changing to `v2_month`

kitallis:
@govindkrjoshi @divs1210 Could you add a section describing how to use these generators and what they do?

govindkrjoshi:
@kitallis We've added an USAGE file for the generator.

PR #350, opened 2019-04-11
Handle the case from Exotel where terminate endpoint sends ‘null’ as the call status

kitallis:
This causes the the CallLog to not be created because the result enum does not handle 'null's. It now has a `unknown` parameter.


spec/controllers/api/current/exotel_call_sessions_controller_spec.rb
```diff
@@ -6,10 +6,6 @@
   let!(:unknown_phone_number) { '1234567890' }
   let!(:invalid_patient_phone_number) { '1800-SIMPLE' }
 
-  before(:each) do
```
divs1210: why remove this?

divs1210: because of removed dependency on Rails cache

app/models/call_log.rb
```diff
@@ -3,6 +3,8 @@ class CallLog < ApplicationRecord
 
   validates :callee_phone_number, presence: true
 
+  UNKNOWN_RESULT = 'unknown'
```
govindkrjoshi: Isn't this unnecessary?

divs1210: ^^ i think the same

kitallis: I added this so that I could use the string `unknown` as a first class value

kitallis: Simplified this.

PR #354, opened 2019-04-11
Add API support for Patient Business Identifiers

govindkrjoshi:
* Schema
   - [x] current
   - [x] v1 and v2
* Patient Sync Controllers
  * Sync from User
     - [x] current
     - [x] v1 and v2
  * Sync to User
     - [x] current
     - [x] v1 and v2

***NOTE:** metadata is a JSON encoded string in request/response, but saved as type JSON in postgres*


divs1210:
Add spec for sync to user response

divs1210:
The request specs are still failing. Can't figure out why.

app/controllers/api/v1/appointments_controller.rb
```diff
@@ -1,4 +1,4 @@
-class Api::V1::AppointmentsController < Api::Current::AppointmentsController
+class Api::V1::AppointmentsController < Api::V2::AppointmentsController
```
divs1210: The generator can be updated to do this automatically in the future.
Created [a story](https://www.pivotaltracker.com/story/show/165438812) for this.

app/models/patient_business_identifier.rb
```diff
@@ -0,0 +1,15 @@
...
+  enum identifier_type: {
+    simple_bp_passport: 'simple_bp_passport'
+  }
+
+  validates :identifier, presence: true
+  validates :identifier_type, presence: true
+
+  validates :device_created_at, presence: true
+  validates :device_updated_at, presence: true
+end
```
kitallis: Add a newline.

app/transformers/api/current/patient_transformer.rb
```diff
@@ -4,12 +4,15 @@ def from_nested_request(payload_attributes)
       payload_attributes = payload_attributes.to_hash.with_indifferent_access
       address = payload_attributes[:address]
       phone_numbers = payload_attributes[:phone_numbers]
+      business_identifiers = payload_attributes[:business_identifiers]
       address_attributes = Api::Current::Transformer.from_request(address) if address.present?
       phone_numbers_attributes = phone_numbers.map { |phone_number| Api::Current::Transformer.from_request(phone_number) } if phone_numbers.present?
+      business_identifiers_attributes = business_identifiers.map { |business_identifier| Api::Current::Transformer.from_request(business_identifier) } if business_identifiers.present?
```
kitallis: Let's move the exceptions to the `current` namespace out into the inherited classes wherever the exceptions apply.

For example,

`business_identifiers` can be filtered away in the V2 transformer namespace.
`phone_numbers` can be filtered away in the V1 transformer namespace.
etc.

Kind of like `to_nested_response` in the Transformer inheritance hierarchy.

app/transformers/api/current/patient_transformer.rb
```diff
@@ -21,7 +24,12 @@ def to_nested_response(patient)
...
         .merge(
           'address' => Api::Current::Transformer.to_response(patient.address),
-          'phone_numbers' => patient.phone_numbers.map { |phone_number| Api::Current::Transformer.to_response(phone_number).except('patient_id') }
+          'phone_numbers' => patient.phone_numbers.map do |phone_number|
+            Api::Current::Transformer.to_response(phone_number).except('patient_id')
+          end,
+          'business_identifiers' => patient.business_identifiers.map do |business_identifier|
+            Api::Current::Transformer.to_response(business_identifier).except('patient_id')
+          end
         ).as_json
```
kitallis: Can get rid of this `as_json`.

app/transformers/api/current/patient_transformer.rb
kitallis: Nevermind, figured that `business_identifiers` are optional even in V3.

spec/controllers/api/current/patients_controller_spec.rb
```diff
@@ -149,6 +159,14 @@ def create_record_list(n, options = {})
           expect(db_phone_number.attributes.with_payload_keys.with_int_timestamps)
             .to eq(updated_phone_number)
         end
+
+        expect(PatientBusinessIdentifier.updated_on_server_since(sync_time).count).to eq 10
```
kitallis: This could potentially be refactored by moving nested patient attributes into different specs.

Maybe,
* Scalar values + address as a single test
* Patient Phone number as a separate test
* Business ID as a separate test

divs1210: refactored

app/transformers/api/current/patient_transformer.rb
```diff
@@ -28,7 +34,10 @@ def to_nested_response(patient)
             Api::Current::Transformer.to_response(phone_number).except('patient_id')
           end,
           'business_identifiers' => patient.business_identifiers.map do |business_identifier|
-            Api::Current::Transformer.to_response(business_identifier).except('patient_id')
+            Api::Current::Transformer
+              .to_response(business_identifier)
+              .except('patient_id')
+              .merge('metadata' => business_identifier.metadata&.to_json)
```
govindkrjoshi: Let's avoid the usage of `&.` 

If metadata is not present., we should not add the key to the response.

divs1210: I believe the key will already be present with a `nil` value

divs1210: refactored, added a `PatientBusinessIdentifierTransformer`

PR #358, opened 2019-04-15
Make sync_to_user audit log creation async

govindkrjoshi:
- Create audit logs during fetch asynchronously.
- Does not change the behavior of merge or login audit logs
- Disable audit logs for facility sync and protocol apis. They are public APIs, most cases there will be no user to create an audit log for.

### DIFF
--- a/app/jobs/create_audit_logs_job.rb
+++ b/app/jobs/create_audit_logs_job.rb
@@ -4,12 +4,12 @@ class CreateAuditLogsJob < ApplicationJob
 
   def perform(user_id, record_class, record_ids, action)
     user = User.find(user_id)
-    audit_logs_attributes = record_ids.map do |record_id|
-      { user: user,
-        auditable_type: record_class,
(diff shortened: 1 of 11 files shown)


app/jobs/create_audit_logs_job.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def perform(user_id, record_class, record_ids, action)
+    user = User.find(user_id)
+    audit_logs_attributes = record_ids.map do |record_id|
+      { user: user,
+        auditable_type: record_class,
+        auditable_id: record_id,
+        action: action }
+    end
+    AuditLog.create!(audit_logs_attributes)
```
kitallis: This won't do a bulk insert right? It will fire N insert queries?

kitallis: I've used this gem in the past to do this: https://github.com/jamis/bulk_insert

PR #360, opened 2019-04-15
Rake task to whitelist patient phone numbers on Exotel

kitallis:
This is to whitelist patient phone numbers that are on DND so that we can phone mask while calling them through Exotel.


lib/tasks/exotel_tasks.rake
```diff
@@ -0,0 +1,30 @@
+namespace :exotel_tasks do
+  desc 'Whitelist patient phone numbers on Exotel'
+  task whitelist_patient_phone_numbers: :environment do
+    require 'exotel_tasks/whitelist_phone_numbers'
+
+    data_file = ENV['JSON_FILE_PATH']
```
govindkrjoshi: Let's use `ENV.fetch` here so we can fail early if the params are missing.

PR #366, opened 2019-04-16
Adherence tab: show enrolment date for patients without BP

timmyjose:
Bug: https://www.pivotaltracker.com/n/projects/2184102/stories/165380393
Description: Enrolment date not shown for patients without BP readings in the Adherence Tab.


divs1210:
the link to story is incorrect

timmyjose:
> the link to pivotal story is incorrect

Thanks for catching that! Fixed with correct link to the bug in Pivotal.

divs1210:
The adherence tab should only show patients with BPs, no?
Why are patients without BPs showing up there in the first place?

app/views/shared/_followup_patient_card.html.erb
```diff
@@ -12,12 +12,14 @@
     <%= patient.gender.capitalize %>,
     <%= patient.current_age %>
   </div>
+  <% if show_enrolment_date %>
+   <div class="overdue-details">
```
govindkrjoshi: Indentation is screwed up.

timmyjose:
> Should the adherence tab should only show patients with BPs?

Based on the discussion with @dvenkatr yesterday, the expected behaviour is:
  * Show the enrolment date for all patients (with and without BP measurements), and
  * Show `Unclear` for data before the program inception date (2018-01-01).

I think the idea is to have the Adherence Tab list as is, and only fix displaying the enrolment date for all regardless of BP measurements existing or not.

app/views/shared/_followup_patient_card.html.erb
timmyjose: Fixed indentation in latest push.

PR #368, opened 2019-04-17
Schedule cronjobs in IST timezone

govindkrjoshi:


config/schedule.rb
```diff
@@ -1,28 +1,15 @@
...
-#   rake "some:great:rake:task"
-# end
-#
-# every 4.days do
-#   runner "AnotherModel.prune_old_records"
-# end
 
-# Learn more: http://github.com/javan/whenever
+def local(time)
+  TZInfo::Timezone.get("Asia/Kolkata").local_to_utc(Time.parse(time))
```
kitallis: Let's use ENV vars here.

PR #369, opened 2019-04-17
Fix capfile require

govindkrjoshi:


govindkrjoshi:
Merging this since the deploy is currently broken.

PR #372, opened 2019-04-22
Adding appointment type as part of the API sync payload; Swagger changes as well

timmyjose:
Story: https://www.pivotaltracker.com/n/projects/2184102/stories/164891328

Note: This completes the following story: https://www.pivotaltracker.com/n/projects/2184102/stories/164570825 (Add defaulters to overdue list).


divs1210:
Update `current` controller specs to see if this field is set/updated correctly.

swagger/v1/swagger.json
```diff
@@ -1669,6 +1669,13 @@
             "null"
           ]
         },
+        "appointment_type": {
```
govindkrjoshi: This should not be present in V1

swagger/v2/swagger.json
```diff
@@ -1853,6 +1853,13 @@
             "null"
           ]
         },
+        "appointment_type": {
```
govindkrjoshi: This should not be present in V2

app/controllers/api/current/appointments_controller.rb
```diff
@@ -46,6 +46,7 @@ def appointments_params
         :cancel_reason,
```
govindkrjoshi: Can we add an explicit check to see that the API marks appointment_type

divs1210:
Is the Android team ready to support this in V3?

swagger/v1/swagger.json
timmyjose: Updated V1 model to exclude this in V1.

swagger/v2/swagger.json
timmyjose: Updated V2 model to exclude V2 swagger json.

app/controllers/api/current/appointments_controller.rb
timmyjose: Will add some relevant specs for the same.

timmyjose: Added specs for `appointment_type` checks against the current API.

spec/factories/appointments.rb
```diff
@@ -10,7 +10,7 @@
     device_updated_at { Time.now }
     agreed_to_visit nil
     remind_on nil
-    appointment_type { Appointment.appointment_types[:manual] }
+    appointment_type { Appointment.appointment_types.keys.to_a.sample }
```
kitallis: I think `to_a` is unnecessary here.

spec/controllers/api/current/appointments_controller_spec.rb
```diff
@@ -34,6 +34,36 @@ def create_record_list(n, options = {})
...
     it_behaves_like 'a working sync controller updating records'
+
+    context 'handles appointment_type correctly' do
+      before :each do
+        set_authentication_headers
+      end
+
+      describe 'stores appointment_type correctly' do
+        let(:request_key) { model.to_s.underscore.pluralize }
+        let(:new_records) { (1..5).map { build_payload.call } }
```
divs1210: `5.times { ... }`

spec/controllers/api/current/appointments_controller_spec.rb
```diff
@@ -34,6 +34,36 @@ def create_record_list(n, options = {})
...
+          before_post_appointments_count = Appointment.all.count
+
+          post(:sync_from_user, params: new_records_payload, as: :json)
+
+          expect(model.count).to eq 5
+          expect(response).to have_http_status(200)
+
+          after_post_appointments = Appointment.all
+          after_post_appointments_ids = after_post_appointments.map(&:id)
+          expect(after_post_appointments.count()).to eq before_post_appointments_count + new_records.count()
```
divs1210: `.count()` => `.count`

spec/controllers/api/current/appointments_controller_spec.rb
```diff
@@ -90,5 +120,25 @@ def create_record_list(n, options = {})
         expect(response_facilities).not_to include(facility_in_another_group.id)
       end
     end
+
+    context 'handles appointment_type correctly' do
+      before :each do
+        set_authentication_headers
+        FactoryBot.create_list(:appointment, 10, facility: request_facility)
```
kitallis: Maybe just keep this list to `5`. Lot of our `create_list` tests build a lot of factories that slow the tests down.

spec/factories/appointments.rb
timmyjose: Done.

spec/controllers/api/current/appointments_controller_spec.rb
timmyjose: Done.

spec/controllers/api/current/appointments_controller_spec.rb
timmyjose: Removed sync count check as per @govindkrjoshi's comment - the shared specs should take care of checking the sync counts.

timmyjose:
> The test should just check if the appointment type has been set correctly. The shared specs will verify if count of the synced records is correct.

Done - removed the record count checks.

spec/controllers/api/current/appointments_controller_spec.rb
timmyjose: Actually reverted this since we are storing the created records for use in the sync payload. 5.times will return 5 as the value, so map is actually correct here.

PR #381, opened 2019-04-30
Fix `appointment type` syncing issues with API v1 and v2.

timmyjose:
Story for bug: https://github.com/simpledotorg/simple-server/pull/381

Changes:
- Moved `set_default_appointment_type` from `current` to `v2` to ensure transformers applied for v1 and v2 and not for current,
- Fixed `attr_accessors` in `v1` and `v2` to exclude `appointment_type`.

### DIFF
--- a/spec/controllers/api/v1/appointments_controller_spec.rb
+++ b/spec/controllers/api/v1/appointments_controller_spec.rb
@@ -32,6 +32,54 @@ RSpec.describe Api::V1::AppointmentsController, type: :controller do
   describe 'POST sync: send data from device to server;' do
     it_behaves_like 'a working sync controller creating records'
     it_behaves_like 'a working sync controller updating records'
+
+    context 'appointment_type' do
+      before :each do
+        set_authentication_headers
+      end
+
+      describe 'Appointments without appointment_type are compatible with v1' do
+        let(:request_key) { model.to_s.underscore.pluralize }
+
+        it 'sets appointment_type to manual if field is not set in the sync payload' do
+          new_records = (1..5).map { build_payload.call.except(:appointment_type) }
+          new_records_payload = Hash[request_key, new_records]
+
+          post(:sync_from_user, params: new_records_payload, as: :json)
+
+          expect(response).to have_http_status(200)
+          after_post_appointments = Appointment.all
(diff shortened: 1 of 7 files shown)


spec/controllers/api/v1/appointments_controller_spec.rb
```diff
@@ -32,6 +32,54 @@ def create_record_list(n, options = {})
...
+          new_records.each do |record|
+            expect(after_post_appointments_ids.include? record[:id]).to be true
+          end
+
+          after_post_appointments.each do |appointment|
+            expect(appointment.appointment_type).to eq Appointment.appointment_types[:manual]
+          end
+        end
+
+        it 'sets appointment_type to manual even if appointment_type explicitly provided' do
```
divs1210: why are we doing this?

spec/controllers/api/v2/appointments_controller_spec.rb
```diff
@@ -34,6 +34,54 @@ def create_record_list(n, options = {})
...
+          new_records.each do |record|
+            expect(after_post_appointments_ids.include? record[:id]).to be true
+          end
+
+          after_post_appointments.each do |appointment|
+            expect(appointment.appointment_type).to eq Appointment.appointment_types[:manual]
+          end
+        end
+
+        it 'sets appointment_type to manual even if appointment_type explicitly provided' do
```
divs1210: why are we doing this?

spec/controllers/api/v1/appointments_controller_spec.rb
timmyjose: @divs1210 This is just a negative test case to ensure that our transformer (only for v1 and v2) take care of ignoring any `appointment_type` that might be sent in the payload (which should not be the case and is likely impossible, but still), and automatically creates an appointment of type `manual` for that appointment.

spec/controllers/api/v2/appointments_controller_spec.rb
timmyjose: @divs1210 This is just a negative test case to ensure that our transformer (only for v1 and v2) take care of ignoring any `appointment_type` that might be sent in the payload (which should not be the case and is likely impossible, but still), and automatically creates an appointment of type `manual` for that appointment.

PR #383, opened 2019-05-02
Create audit log jobs only when records are present

govindkrjoshi:


app/models/audit_log.rb
```diff
@@ -46,7 +46,9 @@ def self.create_logs_async(user, records, action)
     records_by_class = records.group_by { |record| record.class.to_s }
 
     records_by_class.each do |record_class, records_for_class|
-      CreateAuditLogsJob.perform_later(user.id, record_class, records_for_class.map(&:id), action)
+      if records_by_class.present?
```
divs1210: did you mean `records_for_class` instead?

govindkrjoshi:
This change is actually unnecessary.

PR #390, opened 2019-05-07
Display facilities by district instead of Facility Group

timmyjose:
Story: https://www.pivotaltracker.com/n/projects/2184102/stories/165465958
Scope: View only (along with basic caching). 
Depends on: https://github.com/simpledotorg/deployment/pull/73


timcheadle:
Also, why are we putting this behind a feature flag? Seems unnecessary.

timmyjose:
> This all looks good locally. I didn't see a job to cache these values though, is that missing?
> 
> It is concerning that Mansa takes 41 seconds to render on my machine. We need to fix these queries in a later PR.

Hey @timcheadle,  yes, you're right about the caching part. In fact, I had split that into a separate story for which I have a draft PR - https://github.com/simpledotorg/simple-server/pull/393. 

I am planning to merge that into this PR itself (now that the full scope is done), and have further improvements in separate PRs.

timmyjose:
> Also, why are we putting this behind a feature flag? Seems unnecessary.

@timcheadle 
I was actually thinking of having a separate followup story for cleanup (get rid of the Facility Group Analytics part), so this is more like a safety net in the meantime (so that we could switch back to FG Analytics in case of any issues). The idea is to also get rid of the feature flag in the same followup PR once everything is confirmed to be working as anticipated. 

Does that make sense? If not, I can always get rid of the feature flag in the same PR. What do you think?

app/controllers/analytics/districts_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  private
+
+  def set_district
+    district_id = params[:id] || params[:district_id]
+    @district = District.new(district_id)
+    authorize(@district)
+  end
+
+  def set_organization
+    @district.organization_id = params[:organization_id]
```
govindkrjoshi: We can just make this an optional constructor argument.

govindkrjoshi: We can instead set `@organization` in the controller.

app/controllers/analytics/districts_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def set_facilities
+    organization = Organization.find(@district.organization_id)
+    facilities_by_district = policy_scope(organization.facility_groups).flat_map(&:facilities).group_by(&:district).sort.to_h
```
govindkrjoshi:
```suggestion
    facilities_by_district = policy_scope(organization.facilities).group_by(&:district).sort.to_h
```

app/policies/district_policy.rb
```diff
@@ -0,0 +1,52 @@
+class DistrictPolicy < ApplicationPolicy
+  def index?
+    user.owner? || user.organization_owner?
+  end
+
+  def show?
+    user.owner? || [:organization_owner, :supervisor, :analyst].map { |role| admin_can_access?(role) }.any?
+  end
+
+  def create?
```
govindkrjoshi: Unnecessary

app/controllers/analytics/districts_controller.rb
timmyjose: Fixed.

app/policies/district_policy.rb
timmyjose: Removed all except for `graphics` (which might be used later when Whatsapp graphics is re-enabled, if I read it correctly).

app/controllers/analytics/districts_controller.rb
timmyjose: Updated to inject `@organization`, got rid of references to `organization_id`, fixed failing specs.

app/controllers/analytics/facilities_controller.rb
```diff
@@ -22,7 +22,14 @@ def set_facility
   end
 
   def set_facility_group
-    @facility_group = @facility.facility_group
+    if FeatureToggle.enabled?('DASHBOARD_VIEW_BY_DISTRICT')
+      organization = Organization.find(FacilityGroup.find(@facility.facility_group_id).organization_id)
```
govindkrjoshi: Move this to a seperate function?
Also, we have set_organization. Can we create the district using 
```
OrganizationDistrict.new(@facility.district, @organization)
```

app/models/organization.rb
```diff
@@ -12,4 +12,8 @@ class Organization < ApplicationRecord
   validates :name, presence: true
 
   friendly_id :name, use: :slugged
+
+  def get_districts
+    facilities.map(&:district).uniq
```
govindkrjoshi:
```suggestion
    facilities.select(:district).distinct.pluck(:district)
```

app/controllers/analytics/facilities_controller.rb
timmyjose: Done - updated creation of `OrganizationDistrict`, created a new function for organization district.

app/models/organization.rb
timmyjose: Updated.

PR #391, opened 2019-05-08
Control rate summary bar charts for facilities

timcheadle:
Display a summary chart for control and default rates, similar to what the WHO uses in their reports. Something like this (as a reference):

image


timcheadle:
Moved to #394

PR #393, opened 2019-05-08
Display facilities by district - Analytics Cache Jobs

timmyjose:
Story: https://www.pivotaltracker.com/n/projects/2184102/stories/165465958
Scope: Analytics Cache jobs for Districts and separate Sidekiq jobs for each facility
Depends on: https://github.com/simpledotorg/simple-server/pull/390


app/controllers/analytics/districts_controller.rb
```diff
@@ -30,16 +31,20 @@ def set_organization
 
   def set_facilities
     facilities_by_district = policy_scope(@organization.facilities).group_by(&:district).sort.to_h
-    @facilities = facilities_by_district[@district.id].sort_by(&:name)
-    @district.facilities_ids = @facilities&.map(&:id)
+    @district.facilities = facilities_by_district[@district.id].sort_by(&:name)
```
govindkrjoshi:
```suggestion
    @district.facilities = policy_scope(@organization.facilities).where(district: @district.id)
```

timmyjose: Fixed.

app/jobs/concerns/cacheable.rb
```diff
@@ -0,0 +1,35 @@
+module Cacheable
```
kitallis: We can call this `AnalyticsCacheable`

timmyjose: Updated.

PR #394, opened 2019-05-08
Control rate summaries & bar charts for facilities

timcheadle:
(Moving over from #391)

Display a summary chart for control and default rates, similar to what the WHO uses in their reports. Something like this (as a reference):

image


kitallis:
@timcheadle Can we close this PR for now? If we need to access the implementation of bar charts we can simply refer to the branch later. WDYT?

timcheadle:
Sounds good

PR #401, opened 2019-05-10
Moving methods from current to v2 (original version)

timmyjose:
Story: https://www.pivotaltracker.com/n/projects/2184102/stories/165934056

### DIFF
--- a/app/transformers/api/current/appointment_transformer.rb
+++ b/app/transformers/api/current/appointment_transformer.rb
@@ -1,11 +1,2 @@
 class Api::Current::AppointmentTransformer
-  class << self
-    def to_response(appointment)
-      Api::Current::Transformer.to_response(appointment).except('appointment_type')
-    end
-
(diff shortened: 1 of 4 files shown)


app/transformers/api/current/appointment_transformer.rb
```diff
@@ -1,11 +1,2 @@
 class Api::Current::AppointmentTransformer
```
divs1210: This class should inherit the generic transformer, and the current controller should use this instead.

timmyjose: Fixed.

timmyjose:
> the current transformer should inherit the generic transformer,
> and the current controller should use this instead.
> 
> or this transformer shouldn't exist. the current implementation of the transformer is misleading.

Fixed. Keeping the `current` transformer for consistency with the api version. This can be removed (if so needed).

PR #406, opened 2019-05-14
Reduce the memory foot print of analytics queries 

govindkrjoshi:


govindkrjoshi:
These changes are not complete

PR #415, opened 2019-05-21
Cleanup Facility Group Analytics and Graphics 

timmyjose:
For chore: https://www.pivotaltracker.com/story/show/166151057


app/controllers/analytics/districts_controller.rb
```diff
@@ -32,6 +28,10 @@ def set_facilities
     @facilities = policy_scope(@organization_district.facilities).order(:name)
   end
 
+  def set_cache_key
+    @cache_key = ["view", @organization_district.analytics_cache_key(@from_time, @to_time)]
```
kitallis: Don't need `view` here, the view `cache do` already adds this.

app/controllers/analytics/facilities_controller.rb
```diff
@@ -34,6 +24,10 @@ def set_organization
     @organization = @facility.organization
   end
 
+  def set_cache_key
+    @cache_key = ["view", @facility.analytics_cache_key(@from_time, @to_time)]
```
kitallis: Don't need view here, the view cache do already adds this.

app/controllers/analytics/districts_controller.rb
timmyjose: Removed `set_cache_key`.

app/controllers/analytics/facilities_controller.rb
timmyjose: Removed `set_cache_key`.

timmyjose:
Removed the redundant `set_cache_key` methods. Ready for merge.

PR #416, opened 2019-05-21
Add View Caching for Districts

timmyjose:
Chore: https://www.pivotaltracker.com/story/show/166151728

### DIFF
--- a/spec/models/call_log_spec.rb
+++ b/spec/models/call_log_spec.rb
@@ -1,7 +1,8 @@
 require 'rails_helper'
 
 RSpec.describe CallLog, type: :model do
-  describe 'Associations' do
-    it { should belong_to(:user) }
+  describe 'Validations' do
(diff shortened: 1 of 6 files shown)


spec/models/call_log_spec.rb
```diff
@@ -1,7 +1,8 @@
 require 'rails_helper'
 
 RSpec.describe CallLog, type: :model do
-  describe 'Associations' do
-    it { should belong_to(:user) }
+  describe 'Validations' do
```
timcheadle: I think this is unrelated to this PR?

timmyjose: @timcheadle You are correct - this is unrelated, but these specs were failing so fixed this in the same PR.

timmyjose: @timcheadle As discussed on Slack, moved out the specs changes into a a separate PR, and keeping only relevant changes in this PR. 👍

PR #418, opened 2019-05-21
Fix warmup for quarterly analytics

govindkrjoshi:
We discovered two issues with warmup up quarterly caches for a facilities and facility groups. 

- The cache key for a quarterly period will not get refreshed if the key is already present
- There can be discrepancies when displaying all time patients counts between the different time periods depending on when the cache was updated. 
- Adds a `cache_updated_at` timestamp to track when a cache was updated.

### DIFF
--- a/app/views/analytics/facility_groups/show.html.erb
+++ b/app/views/analytics/facility_groups/show.html.erb
@@ -6,9 +6,8 @@
   <div class="dashboard">
     <h1><%= @facility_group.name %></h1>
     <%= string_for_range(@from_time, @to_time) %>
-    <p class="text-muted"><small>Last updated: <%= Time.now.strftime("%B %d, %Y") %></small></p>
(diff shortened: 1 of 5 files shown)


app/views/analytics/facility_groups/show.html.erb
```diff
@@ -6,10 +6,11 @@
   <div class="dashboard">
     <h1><%= @facility_group.name %></h1>
     <%= string_for_range(@from_time, @to_time) %>
-    <p class="text-muted"><small>Last updated: <%= Time.now.strftime("%B %d, %Y") %></small></p>
+    <% if @facility_group_analytics[:cached_updated_at].present? %>
```
kitallis: I think this should be `cache_updated_at`
```suggestion
    <% if @facility_group_analytics[:cache_updated_at].present? %>
```

PR #420, opened 2019-05-22
Added percentage symbol to BP header and values in Dashboard

timmyjose:
Story: https://www.pivotaltracker.com/story/show/165672235

### DIFF
--- a/app/helpers/dashboard_helper.rb
+++ b/app/helpers/dashboard_helper.rb
@@ -4,11 +4,11 @@ module DashboardHelper
   end
 
   def dash_if_zero(value)
-    is_zero(value) ? "-" : value
+    zero?(value) ? "-" : value
   end
 
-  def is_zero(value)
-    value == nil || value == 0
+  def zero?(value)
+    value.nil? || value.zero?
   end
 
   def analytics_date_format(time)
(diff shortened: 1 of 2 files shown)


app/helpers/dashboard_helper.rb
```diff
@@ -4,7 +4,11 @@ def lighten_if_zero(value)
   end
 
   def dash_if_zero(value)
-    (value == nil || value == 0) ? "-" : value
+    is_zero(value) ? "-" : value
+  end
+
+  def is_zero(value)
+    value == nil || value == 0
```
kitallis: Potentially,
```suggestion
    value.nil? || value.zero?
```

app/helpers/dashboard_helper.rb
```diff
@@ -4,7 +4,11 @@ def lighten_if_zero(value)
   end
 
   def dash_if_zero(value)
-    (value == nil || value == 0) ? "-" : value
+    is_zero(value) ? "-" : value
+  end
+
+  def is_zero(value)
```
kitallis:
```suggestion
  def zero?(value)
```

timmyjose: Updated the method name to be more idiomatic.

app/helpers/dashboard_helper.rb
timmyjose: Updated to use methods.

PR #425, opened 2019-05-27
Fixing formatting issues in overdue list dashboard view

timmyjose:
Story: https://www.pivotaltracker.com/story/show/166009799

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -55,7 +55,7 @@ module ApplicationHelper
     interaction_result
   end
 
-  def patient_has_street_address(patient_address)
+  def patient_has_street_address?(patient_address)
     patient_address.street_address.present?
   end
 end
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -54,5 +54,9 @@ def show_last_interaction_date_and_result(patient)
 
     interaction_result
   end
+
+  def patient_has_street_address(patient_address)
```
kitallis: Wondering whether this necessitates a helper. If we keep it, would be valuable to make a boolean function.
```suggestion
  def patient_has_street_address?(patient_address)
```

timmyjose: Yes, I did debate whether such a short function could be inlined. However, I feel like it would still be worth keeping it separate from the view (cleaner). I will make the predicate more explicit by adding the `?`.

timcheadle: You could add it as an instance method on `PatientAddress` itself, might be cleaner.

```
patient_address.has_street_address?
```

timmyjose: @timcheadle Yes, that does make more sense. Removed the helper method and added predicate to the `Address` model instead.

PR #428, opened 2019-05-28
Add caller_phone_number as an argument to SMSNotificationService

kitallis:
`SMSNotificationService` was incompatible with Twilio sub-accounts, this PR makes the service accept a `caller_phone_number` and hence we can use it to send SMSes from various accounts.


app/jobs/appointment_notification/worker.rb
```diff
@@ -26,9 +26,14 @@ def perform(user_id, appointments, communication_type)
...
   private
 
+  SMS_CLIENT = Twilio::REST::Client.new(ENV.fetch('TWILIO_REMINDERS_ACCOUNT_SID'),
+                                        ENV.fetch('TWILIO_REMINDERS_ACCOUNT_AUTH_TOKEN'))
+
   def send_sms(appointment, type)
     SmsNotificationService
-      .new(appointment.patient.latest_phone_number)
+      .new(appointment.patient.latest_phone_number,
+           ENV.fetch('TWILIO_REMINDERS_ACCOUNT_SID'),
```
govindkrjoshi:
```suggestion
           ENV.fetch('TWILIO_REMINDERS_ACCOUNT_PHONE_NUMBER'),
```

app/services/sms_notification_service.rb
```diff
@@ -2,7 +2,8 @@ class SmsNotificationService
   include SmsHelper
   DEFAULT_LOCALE = :en
 
-  def initialize(recipient_number, client = Twilio::REST::Client.new)
+  def initialize(recipient_number, caller_number, client = Twilio::REST::Client.new)
```
govindkrjoshi: Can we rename `caller_number` to `sms_number` or something else?

config/routes.rb
```diff
@@ -197,7 +197,7 @@
   end
 
   authenticate :admin, lambda(&:owner?) do
-    require 'sidekiq/web'
+    require 'sidekiq_unique_jobs/web'
```
govindkrjoshi: This should be reverted.

PR #431, opened 2019-05-29
Add non-breaking Sync API changes without frequent version bumps

kitallis:
https://www.pivotaltracker.com/story/show/166284929


app/controllers/concerns/api/current/sync_to_user.rb
```diff
@@ -29,25 +29,40 @@ def processed_until(records)
     def response_process_token
       { current_facility_id: current_facility.id,
```
govindkrjoshi: We can check if `sync_from_beginning?` is true here and return an empty token from there itself.

spec/requests/shared_examples/api/current/shared_spec_for_sync_requests.rb
```diff
@@ -87,4 +87,52 @@ def assert_sync_success(response, request_process_token)
...
+    let(:headers_with_resync_token) { headers.merge('HTTP_X_RESYNC_TOKEN' => resync_token) }
+    let(:process_token_without_resync) { make_process_token(current_facility_processed_since: Time.now,
+                                                            other_facilities_processed_since: Time.now) }
+
+    before do
+      post sync_route, params: many_valid_records.to_json, headers: headers
+    end
+
+    it 'syncs all records from beginning if resync_token in headers is different from the one in process_token' do
+      get sync_route, params: { process_token: process_token_without_resync }, headers: headers_with_resync_token
```
govindkrjoshi: This test can be split into two; one to check when then resync_token is nil, and another when there is a mismatch

kitallis: This is already done as a separate context down below.

app/controllers/concerns/api/current/sync_to_user.rb
```diff
@@ -29,25 +29,40 @@ def processed_until(records)
...
         current_facility_processed_since: processed_until(current_facility_records) || current_facility_processed_since,
-        other_facilities_processed_since: processed_until(other_facility_records) || other_facilities_processed_since }
+        other_facilities_processed_since: processed_until(other_facility_records) || other_facilities_processed_since,
+        resync_token: resync_token
+      }
     end
 
     def encode_process_token(process_token)
       Base64.encode64(process_token.to_json)
```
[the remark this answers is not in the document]
kitallis: We can do this as a separate PR.

app/controllers/concerns/api/current/sync_to_user.rb
kitallis: Makes sense.

PR #447, opened 2019-06-12
Simple Dashboard V2 (District + Facility View)

kitallis:
https://www.pivotaltracker.com/story/show/166497280
https://www.pivotaltracker.com/story/show/166499190


app/models/facility.rb
```diff
@@ -41,8 +40,12 @@ class Facility < ApplicationRecord
...
   friendly_id :name, use: :slugged
 
-  def report_on_patients
-    registered_patients
+  def dashboard_analytics
+    query = FacilityAnalyticsQuery.new(self)
+
+    [query.follow_up_patients_by_month,
+     query.registered_patients_by_month,
+     query.total_registered_patients].compact.inject(&:deep_merge)
```
govindkrjoshi: Do we need compact here?
Would any of the three elements be nil?

app/models/organization_district.rb
```diff
@@ -1,24 +1,13 @@
...
-    "analytics/organization/#{organization.id}/district/#{district_slug(district_name)}/#{time_cache_key(from_time)}/#{time_cache_key(to_time)}/#{cache_key}"
-  end
+  def dashboard_analytics
+    query = DistrictAnalyticsQuery.new(district_name)
 
-  def district_slug(district_name)
-    district_name.split(" ").select(&:present?).join("-").downcase
+    [query.follow_up_patients_by_month,
+     query.registered_patients_by_month,
+     query.total_registered_patients].compact.inject(&:deep_merge)
```
govindkrjoshi: Same here.

app/services/appointment_notification_service.rb
```diff
@@ -19,7 +19,9 @@ def fan_out_reminders(appointments, communication_type, schedule_at)
                                 communication_type)
     end
 
-    sampled_appointments = eligible_appointments.group_by(&:facility_id).flat_map do |_, facility_appointments|
+    sampled_appointments = eligible_appointments
```
govindkrjoshi: These changes are unnecessary in this PR.

spec/controllers/analytics/district_controller_spec.rb
```diff
@@ -1,82 +1,5 @@
...
-                                   :blood_pressures_recorded_per_week,
-                                   :cache_updated_at]
-
-        expect(assigns(:district_analytics).keys).to match_array(expected_analytics_keys)
-        expect(assigns(:district_analytics)[:cache_updated_at]).to eq(request_time)
-        expect(assigns(:facility_analytics)[facility].keys).to match_array(expected_analytics_keys)
-      end
-    end
-  end
+  pending '#show'
```
govindkrjoshi: Missing specs

spec/controllers/districts_controller_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'rails_helper'
```
govindkrjoshi: This file is in the wrong location. 
Should be spec/controllers/analytics/district_controller_spec.rb

app/models/facility.rb
kitallis: Yeah, the elements can be nil when there are no `follow_up_patients` for example. This is happening for a few districts in SBX data.

spec/controllers/districts_controller_spec.rb
kitallis: Good catch, fixing.

spec/controllers/analytics/district_controller_spec.rb
kitallis: It went into the wrong file.

PR #448, opened 2019-06-14
Bulk upload facilities V1

vkrmis:
- allow owners to create multiple facilities by uploading a csv file


app/controllers/admin/facilities_controller.rb
```diff
@@ -42,6 +42,24 @@ def destroy
     redirect_to admin_facilities_url, notice: 'Facility was successfully deleted.'
   end
 
+  def upload
+    authorize Facility
+    file = params[:upload_facilities_file]
+    @file_contents = ''
+    @errors = []
+    run_validations_and_read(file) if file.present?
```
govindkrjoshi: Split file type and size validation, reading the file and validating it content in to separate functions. You can also use before action to run the file validations.

app/controllers/admin/facilities_controller.rb
```diff
@@ -42,6 +42,24 @@ def destroy
     redirect_to admin_facilities_url, notice: 'Facility was successfully deleted.'
   end
 
+  def upload
+    authorize Facility
+    file = params[:upload_facilities_file]
```
govindkrjoshi: You can use `params.require` here to ensure the param is set.

app/controllers/admin/facilities_controller.rb
```diff
@@ -42,6 +42,24 @@ def destroy
...
+    if @errors.present?
+      @errors.prepend 'Please fix the errors below and try again:'
+      @errors << ['', 'Please contact team@simple.org if you need assistance']
+      flash.now[:alert] = @errors.join('<br/>').html_safe
+    elsif file.present?
+      facilities = Facility.parse_facilities(@file_contents)
+      ImportFacilitiesJob.perform_later(facilities)
+      flash.now[:notice] = 'File upload successful, your facilities will be created shortly.'
+    end
+    render :upload
```
govindkrjoshi: We should return a bad request if any of the params are missing, or the validations fail.

app/models/facility.rb
```diff
@@ -28,4 +40,58 @@ class Facility < ApplicationRecord
...
+                  district: row['district'],
+                  state: row['state'],
+                  country: row['country'],
+                  pin: row['pin'],
+                  latitude: row['latitude'],
+                  longitude: row['longitude'],
+                  import: true}
+      facilities << facility
+    end
+    facilities.delete_if { |facility| facility.except(:import).values.all?(&:blank?) }
```
govindkrjoshi: Instead of filtering them out later, you can skip adding them in the do block using `next`

app/models/facility.rb
```diff
@@ -28,4 +40,58 @@ class Facility < ApplicationRecord
...
+  end
+
+  def self.read_import_file(file)
```
govindkrjoshi: The model shoudn't be concerned with reading a file. 
Do you think it's better to create a FileUploadHelper?

app/models/facility.rb
```diff
@@ -20,6 +26,12 @@ class Facility < ApplicationRecord
   validates :country, presence: true
   validates :pin, numericality: true, allow_blank: true
 
+  with_options if: :is_import? do |facility|
```
govindkrjoshi: Why not just directly use `:import`

app/models/facility.rb
```diff
@@ -28,4 +40,58 @@ class Facility < ApplicationRecord
...
+    file_contents = file.read if file.content_type == 'text/csv'
+    file_contents
+  end
+
+  def is_import?
+    import
+  end
+
+  def unique_for_facility_group_and_organization
+    organization = Organization.find_by(name: organization_name)
```
govindkrjoshi: Won't model validations take care of these?

app/policies/facility_policy.rb
```diff
@@ -31,6 +31,10 @@ def graphics?
     show?
   end
 
+  def upload?
+    user.owner?
```
govindkrjoshi: Won't this be the same as facility_group_policy?

spec/controllers/admin/facilities_controller_spec.rb
```diff
@@ -0,0 +1,166 @@
+require 'rails_helper'
+
+RSpec.describe Admin::FacilitiesController, type: :controller do
+  let(:facility_id) { SecureRandom.uuid }
```
govindkrjoshi: This is unused.

spec/controllers/admin/facilities_controller_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+  let(:facility_group) { FactoryBot.create(:facility_group) }
+  let(:valid_attributes) {
+    FactoryBot.attributes_for(
+        :facility,
+        facility_group_id: facility_group.id,
```
govindkrjoshi:
```suggestion
        facility_group: facility_group,
```

app/views/admin/facilities/upload.html.erb
```diff
@@ -0,0 +1,22 @@
+<h1>Upload Facilities</h1>
+
+<h3>Instructions:</h3>
+<ol>
+  <li>Download and fill in <%= link_to 'this template', '/documents/upload_facilities.csv'%> to create multiple facilities via file upload.</li>
```
govindkrjoshi: Check if you can generate the asset path

app/views/admin/facilities/upload.html.erb
```diff
@@ -0,0 +1,22 @@
...
+      <%= link_to '+ Organization', new_admin_organization_path, class: "btn btn-sm btn-outline-primary" %>
+    <% end %></li>
+  <li>Facility Groups for the facilities must already exist.
+    <% if policy(FacilityGroup).new? %>
+      Click here to create a facility group:
+      <%= link_to '+ Facility group', new_admin_facility_group_path, class: "btn btn-sm btn-outline-primary" %>
+    <% end %></li>
+  <li>Ensure that the Organization (Column A) and Facility Group (Column B) are entered correctly.</li>
+</ol>
+<%= bootstrap_form_tag(url: '/admin/facilities/upload' , multipart: true) do |f| %>
```
govindkrjoshi: Use helper method instead of direct path

app/models/facility.rb
govindkrjoshi: Split the validations into individual validations

app/jobs/import_facilities_job.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def perform(facilities)
+
+    import_facilities = []
+    facilities.each do |facility|
+      organization = Organization.find_by(name: facility[:organization_name])
+      facility_group = FacilityGroup.find_by(name: facility[:facility_group_name],
+                                             organization_id: organization.id)
+      import_facility = Facility.new(facility.merge!(facility_group_id: facility_group.id))
+      puts import_facility
```
govindkrjoshi: unnecessary

app/jobs/import_facilities_job.rb
```diff
@@ -0,0 +1,18 @@
+class ImportFacilitiesJob < ApplicationJob
+  queue_as :default
+  self.queue_adapter = :sidekiq
+
+  def perform(facilities)
+
+    import_facilities = []
+    facilities.each do |facility|
```
govindkrjoshi: Maybe write it as a map

app/assets/javascripts/bs-file-input-init.js
```diff
@@ -0,0 +1,3 @@
+$(document).ready(function () {
+    bsCustomFileInput.init()
+})
```
govindkrjoshi: Missing semicolon

app/policies/facility_policy.rb
vkrmis: Removing the facility_group_policy for upload?, since I'm not currently using it.

spec/controllers/admin/facilities_controller_spec.rb
vkrmis: This doesn't seem to work. I'll come back to this later.

app/views/admin/facilities/upload.html.erb
vkrmis: Tried doing this, but it seems like overkill
Need to:
- Create a route
- Add a method under the facilities controller
- Add a policy for the method

Do you think its worth doing?

app/jobs/import_facilities_job.rb
vkrmis: Not doing this - side effects

spec/spec_helper.rb
```diff
@@ -3,7 +3,7 @@
 require 'webmock/rspec'
 require 'fakeredis/rspec'
 require 'sidekiq/testing'
-
+require 'pry'
```
govindkrjoshi:
```suggestion
```

PR #455, opened 2019-07-01
Only display flash messages of type String in the login page

vkrmis:


app/views/layouts/application.html.erb
```diff
@@ -112,6 +112,8 @@
 
     <main role="main" class="container">
       <% flash.each do |type, message| %>
+        <% next unless message.is_a? String %>
```
timcheadle:
```suggestion
        <% next unless message.is_a?(String) %>
```

PR #474, opened 2019-07-12
Add a confirmation step for deployment tasks (production & staging)

kitallis:
For any capistrano tasks.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -48,6 +48,8 @@ namespace :deploy do
       WARNING: You're about to run tasks on #{ENVS_FOR_CONFIRMATION_STEP.join('/')} server(s)
       Please confirm that all your intentions are kind and friendly.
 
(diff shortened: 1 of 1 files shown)


config/deploy.rb
```diff
@@ -38,4 +38,35 @@
...
+    puts <<-WARN
+
+    ===============================================================================
+
+      WARNING: You're about to run tasks on #{ENVS_FOR_CONFIRMATION_STEP.join('/')} server(s)
+      Please confirm that all your intentions are kind and friendly.
+
+      Check if:
+
+      * You are deploying the correct branch
```
timcheadle: Let's make this more explicit, such as:
```suggestion
      This will deploy `#{BRANCH}` to #{ENV})
      
      Check if:
      * You are deploying the correct branch
      * You are deploying to the correct environment
```

PR #476, opened 2019-07-12
Complete UI refresh

timcheadle:
Now that it's a first-class product, we need to make simple-server look as polished as our Android app. @dburka's put in a ton of work to get us there, so this PR is a bit of a monster.

We know it's going to break some of the automated UI tests for now. We'll comment the breaking tests out and fix them in a separate PR. That means this PR not only needs code review, but will need to manually QA everything the automated tests did before. I'll catalog them here as we comment them out.


timcheadle:
I've fixed all of the specs _except_ for those in `spec/features/admin`. Those I've added `skip` commands with a to do item of fixing them in another PR.

To review this, let's get it locally (or on sandbox) and manually QA the user flows.

PR #481, opened 2019-07-17
Adding Rails caching for analytics dashboard

timmyjose:
(Chore) Story: https://www.pivotaltracker.com/story/show/167342985

The idea is to enable view caching for the dashboard views for snappier user experience.

Simply caching the rendered partials without caching the query data does not seem to make much of an improvement since the major bottleneck is the queries. So we now,

* Use a low-level rails cache on the controller level (which also caches the query data)
* The cache invalidates after a day (the key is set to today's date)


timmyjose:
Currently ensuring cache invalidation through an `expires_in` value of `2.days` for both district and facility level views. Discussion on this required.

kitallis:
> This looks great. However, I think the `analytics_cache_key` methods belong in the controllers, not the models. I don't want the model class concerned with how it's cached in a view.

Do you think it's also just sufficient to have a `cache_time` instead of `from_time` and `to_time` since that currently doesn't apply that well.

timmyjose:
> I don't think we need caching tests. We can trust that Rails caching works as intended for now.

I just saw this comment after I had pushed a couple of caching specs! :-).

spec/controllers/analytics/facilities_controller_spec.rb
```diff
@@ -37,18 +37,55 @@
...
+      before do
+        Rails.cache.clear
+      end
+
+      let(:cohort_date1) { (Date.today - (0 * 3).months).beginning_of_quarter }
+      let(:cohort_date2) { (Date.today - (1 * 3).months).beginning_of_quarter }
+      let(:cohort_date3) { (Date.today - (2 * 3).months).beginning_of_quarter }
+
+      it 'caches the facility correctly' do
+        today = Date.today.strftime("%Y-%m-%d")
```
timcheadle: Instead of repeating `Date.today`, do this:

```ruby
let(:today) { Date.today }
```

Then use `today` in your other variables and specs. This prevents the race condition where a test flaps because it runs at midnight and spans two days.

timmyjose: @timcheadle Yes, that was a silly oversight on my part! Updated the specs.

PR #516, opened 2019-08-07
FIX: Sync approval status and reason

timmyjose:
Story: https://www.pivotaltracker.com/story/show/167505761

Index page:

* When user is in `allowed` state:

index_allowed

* When user is in `denied` state:

index_denied

Show page:

* When user is in `allowed` state:
show_allowed

* When user is in `denied` state:
show_denied


timmyjose:
> Can we do without the request_access?
> Can we not go from denied -> allowed directly?

We are done implementing this change. Removed `requested` state for the index and show pages.

PR #519, opened 2019-08-07
Paginate Recent Blood Pressure log

kitallis:
Screen Shot 2019-08-07 at 3 46 44 PM


timcheadle:
This is good, but there's a very good chance that this view will change to a "visit log" instead. Stay tuned.

kitallis:
> This is good, but there's a very good chance that this view will change to a "visit log" instead. Stay tuned.

I'll just merge this for now and we can modify as 'visit logs' come in

kitallis:
> Can you add an html anchor to the paginated div, and include it in the pagination link url?
> It currently jumps to the top of the page every time you click on a page



> Can you add an html anchor to the paginated div, and include it in the pagination link url?
> It currently jumps to the top of the page every time you click on a page

Done.

PR #529, opened 2019-08-14
Skip BP display in BP log (facilities dashboard) if its patient record is not present in the backend

timmyjose:
Story: https://www.pivotaltracker.com/story/show/167893720


timcheadle:
We can also just skip these records entirely in the BP query:

```
    @recent_blood_pressures = @facility.blood_pressures
                                .joins(:patient, :user)
                                .includes(:patient, :user)
                                .order("DATE(recorded_at) DESC, recorded_at ASC")
```

PR #531, opened 2019-08-17
Update ruby.yml

govindkrjoshi:


kitallis:
Closing this since semaphore continues to be a lot more faster and easier to setup. We can revisit this later.

PR #533, opened 2019-08-19
Add permissions for users

govindkrjoshi:


app/policies/appointment_policy.rb
```diff
@@ -21,7 +34,19 @@ def initialize(user, scope)
     end
 
     def resolve
-      scope.where(facility: @user.facilities)
+      if user.has_permission?(:can_access_appointment_information_for_all_organizations)
```
timmyjose: This is rather nice.

app/policies/appointment_policy.rb
```diff
@@ -21,7 +34,19 @@ def initialize(user, scope)
     end
 
     def resolve
```
timmyjose: This method reads rather nicely now.

app/policies/facility_policy.rb
```diff
@@ -48,7 +77,22 @@ def initialize(user, scope)
     end
 
     def resolve
-      scope.where(facility_group: user.facility_groups)
+      if user.has_permission?(:can_manage_all_organizations)
+        return scope.all
+      elsif user.has_permission?(:can_manage_an_organization)
+        facility_groups = resources_for_permission(:can_manage_an_organization).flat_map(&:facility_groups)
```
timmyjose: Perhaps this could also follow the same style as the other else branches? I mean like so: `return scope.where(facility_groups: ...)`

lib/tasks/scripts/create_master_user.rb
```diff
@@ -0,0 +1,77 @@
...
+  def self.create_user_for_admin!(admin)
+    admin_attributes = admin.attributes.with_indifferent_access
+
+    user_attributes =
+      admin_attributes
+        .slice(:role, :created_at, :updated_at, :deleted_at)
+        .merge(id: master_user_id(admin.email),
+               full_name: admin.email.split('@').first,
+               organization: get_organization(admin),
+               sync_approval_status: 'denied',
```
timmyjose: Wouldn't it be nicer to have a 'n/a' status in the `sync_approval_status` enum? Right now, the status and the status reason seem a bit disjointed.

app/policies/appointment_policy.rb
```diff
@@ -1,18 +1,29 @@
...
   end
 
   def download?
-    user.owner? || user.supervisor? &&
-      user.organizations.include?(Organization.find(ENV['IHCI_ORGANIZATION_UUID']))
+    update?
   end
 
-  class Scope
+  class Scope < Scope
```
kitallis:
```suggestion
  class Scope < ApplicationPolicy::Scope
```

PR #544, opened 2019-08-27
Add index on patient_id references

govindkrjoshi:
Adds index on patient id references to speed up lookups


db/migrate/20190827061512_add_index_for_patient_references.rb
```diff
@@ -0,0 +1,9 @@
+class AddIndexForPatientReferences < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
```
govindkrjoshi: https://thoughtbot.com/blog/how-to-create-postgres-indexes-concurrently-in

PR #546, opened 2019-08-28
Change 'enrollment' to 'registration' for consistency

timcheadle:


kitallis:
@timcheadle Thanks! This was my bad. I wanted to do this in the original Enrollment --> Registration PR but it slipped my mind.

PR #548, opened 2019-08-29
added attributes in html tags

BPrerana:


app/views/admins/index.html.erb
```diff
@@ -60,7 +60,7 @@
 
 
 <% @admins.each do |admin| %>
-    <div class="card">
+    <div class="card" alt="<%= admin.email %>">
```
govindkrjoshi: Do you want to use `title` here instead?

app/views/admins/show.html.erb
```diff
@@ -29,7 +29,7 @@
         <div class="card-row">    
             <h5><i class="fas fa-network-wired"></i> Facility Groups</h5>
 
-            <ul>
+            <ul alt="facility_groups">
```
govindkrjoshi: Same here.

app/views/admins/show.html.erb
```diff
@@ -39,7 +39,7 @@
         <div class="card-row">
             <h5><i class="fas fa-network-wired"></i> Organizations</h5>
 
-            <ul>
+            <ul alt="organizations">
```
govindkrjoshi: Same here.

PR #564, opened 2019-09-11
Add the Encounters model

kitallis:
https://www.pivotaltracker.com/story/show/168719092

What this does:

- [x] Add an Encounter sync API endpoint
- [x] Add support to back-fill Encounters for calls through the existing BloodPressure sync API 

What it does not do:

- Add a migration to create Encounters for existing BPs (see this https://github.com/simpledotorg/simple-server/pull/602)
- Move all the appropriate rails references to use the `through :encounters` relationship


app/models/blood_pressure.rb
```diff
@@ -5,9 +5,12 @@ class BloodPressure < ApplicationRecord
   ANONYMIZED_DATA_FIELDS = %w[id patient_id created_at bp_date registration_facility_name user_id
                               bp_systolic bp_diastolic]
 
-  belongs_to :facility, optional: true
   belongs_to :patient, optional: true
   belongs_to :user, optional: true
+  belongs_to :facility, optional: true
+
+  has_one :encounter_event, as: :encounterable
```
timcheadle: Perhaps we should call these `Observation` to match FHIR?

app/models/encounter.rb
```diff
@@ -0,0 +1,42 @@
+class Encounter < ApplicationRecord
+  belongs_to :patient
+  belongs_to :facility
+
+  has_many :encounter_events
+  has_many :blood_pressures, through: :encounter_events, source: :encounterable, source_type: 'BloodPressure'
+
+  def events
```
timcheadle: Again, these should probably be `observations`. The language in this method is not obvious.

app/models/encounter.rb
```diff
@@ -0,0 +1,42 @@
...
+    encounter_events.includes(:encounterable).map(&:encounterable)
+  end
+
+  def self.create_encounter_with_events!(params)
```
timcheadle: Can you outline the use case for this method?

In my mind, you'd simply create a `BloodPressure`. Perhaps that model can be "observable" and, before create, finds or creates an encounter for the patient. The logic wouldn't live in Encounter at all.

kitallis: This method and its uses are still under progress and it will support editing as well (find or create). But generally speaking, the purpose of this method would be to create an Encounter "as-is" (sent from the app). The app will send us an appropriate Encounter ID and its constituent `Observations` since the data will now be nested within the Encounter object.

The API will also represent a completely new sync resource called `Encounter` and that will simply linearly traverse the contents of the Encounter payload.

I don't think I understood why we need to approach creating Encounters from bottom-up (through BloodPressures / PrescriptionDrugs) on a modelling level. This would be true for backwards-compatability's for existing BP/PD APIs sake but the primary Encounter API needn't behave that way.

app/models/blood_pressure.rb
kitallis: `Observations` is a much better name than `Event`. Thanks!

app/models/encounter.rb
kitallis: Haven't fully thought through this method yet, subject to change, but the basic intention should be:

Transactionally create:
- The actual `BloodPressure`
- Associate `BloodPressure` against the `Observation` table
- Associate the Observation with the appropriate `Encounter`

Does not support partial creates for the Encounter (does not allow saving 5/10 BPs). It's either 100% updated or the whole thing is unsynced.

app/schema/api/current/models.rb
```diff
@@ -106,6 +106,32 @@ def nested_patient
...
+          updated_at: { '$ref' => '#/definitions/timestamp' },
+          patient_id: { '$ref' => '#/definitions/uuid' },
+          encountered_on: { type: :string, format: :date },
+          notes: { type: :string },
+          observations: {
+            type: :object,
+            properties: {
+              blood_pressures: { '$ref' => '#/definitions/blood_pressures' },
+            },
+            required: %w[blood_pressures]
```
govindkrjoshi: We can have empty observations, we should not have `blood_pressures` as a required field.

spec/factories/blood_pressures.rb
```diff
@@ -6,9 +6,10 @@
     device_created_at { Time.now }
     device_updated_at { Time.now }
     recorded_at { device_created_at }
-    association :facility, strategy: :build
-    association :patient, strategy: :build
-    user
+
+    association :facility, strategy: :create
```
govindkrjoshi: Why does the strategy need to change?

swagger/v2/swagger.json
```diff
@@ -1486,6 +1486,65 @@
         "$ref": "#/definitions/nested_patient"
       }
     },
+    "encounter": {
```
govindkrjoshi: V2 api schema should not have encounter definitions.

PR #568, opened 2019-09-13
Support switching between monthly and quarterly cohort charts

timcheadle:
We want to allow users to switch the dashboard pages from monthly to quarterly time periods. This PR adds a dropdown to do so and updates the cohort charts to use the new time periods.

The rest of the page will follow suit in subsequent PRs.

Monthly district view:

image

Quarterly facility view:

image


app/queries/cohort_analytics_query.rb
```diff
@@ -1,17 +1,33 @@
...
 
-  def patient_counts(year:, quarter:, quarters_previous: 1)
-    report_start = quarter_start(year, quarter)
-    report_end   = quarter_end(year, quarter)
+  def patient_counts_by_period(period, prev_periods)
+    results = {}
 
-    cohort_start = report_start - (quarters_previous * 3).months
-    cohort_end   = report_end   - (quarters_previous * 3).months
+    (0..(prev_periods)).each do |periods_back|
```
timmyjose: So in the case of period `:month`, this will run for 7 iterations. I presume that this is so that we can skip the current month in the view, and still be left with 6 months of cohort data in the view. Also that this approach is used so that we can have the processing for both `:month` and `:quarter` done in the same loop. Is my understanding correct?

timcheadle: That's right. The thinking is "current period + `prev_preiods` periods".

It's a little convoluted, all caused because we want to skip the current month in the view.

PR #576, opened 2019-09-18
Backfill user ids from audit logs

vkrmis:


lib/tasks/data_migration.rake
```diff
@@ -94,4 +94,40 @@ namespace :data_migration do
...
+  task backfill_user_ids_for_medical_histories: :environment do
+    batch_size = ENV.fetch('BACKFILL_USER_ID_FROM_AUDIT_LOGS_BATCH_SIZE').to_i
+    AuditLog.creation_logs_for_type('MedicalHistory').in_batches(of: batch_size) do |batch|
+      medical_history_log_ids = batch.map do |medical_history|
+        { id: medical_history.auditable_id,
+          user_id: medical_history.user_id }
+      end
+      UpdateUserIdsFromAuditLogsWorker.perform_async(MedicalHistory, medical_history_log_ids)
+    end
+  end
```
kitallis: Could make this into a function that does this generically to avoid having to change all three of them if there's ever a need.

app/models/audit_log.rb
```diff
@@ -16,37 +16,55 @@ class AuditLog < ApplicationRecord
...
+                                            action: action,
+                                            time: time }.to_json)
     end
   end
+
+  def self.write_audit_log(log)
+    AuditLogger.info(log.to_json)
+  end
+
+  def self.creation_logs_for_type(type)
```
kitallis: Could move this directly into the rake task itself.

PR #577, opened 2019-09-19
Adherence spec

BPrerana:
Added adherence flow spec and patient page


vkrmis:
Is chromedriver absolutely necessary here?

PR #582, opened 2019-09-30
Patient line list download

timcheadle:
Allow dashboard users (CVHOs primarily) to download patient line lists for the district or for a facility.

The link will trigger a background job that renders the CSV line list and emails it to the currently logged in user.

image


app/models/patient_business_identifier.rb
```diff
@@ -12,4 +12,13 @@ class PatientBusinessIdentifier < ApplicationRecord
 
   validates :device_created_at, presence: true
   validates :device_updated_at, presence: true
+
+  def shortcode
+    case identifier_type
+    when "simple_bp_passport"
```
kitallis: I think this can be written more simply as `when simple_bp_passport?`

app/models/patient.rb
```diff
@@ -69,6 +69,10 @@ def latest_phone_number
     phone_numbers.last&.number
   end
 
+  def latest_bp_passport
+    business_identifiers.where(identifier_type: "simple_bp_passport").order(:device_created_at).last
```
kitallis: This could be written as,

```suggestion
    business_identifiers.simple_bp_passport.order(:device_created_at).last
```

app/mailers/patient_list_download_mailer.rb
```diff
@@ -0,0 +1,18 @@
+class PatientListDownloadMailer < ApplicationMailer
+  default from: 'help@simple.org'
+
+  def patient_list(recipient_email, model_type, model_name, patients_csv)
+    @model_type = model_type
+    @model_name = model_name
```
kitallis: Is marking them as instance variables necessary here?

app/jobs/patient_list_download_job.rb
```diff
@@ -0,0 +1,21 @@
...
+  queue_as :default
+  self.queue_adapter = :sidekiq
+
+  def perform(recipient_email, model_type, params)
+    case model_type
+    when 'district' then
+      district_name = params[:district_name]
+      organization  = Organization.find(params[:organization_id])
+      model = OrganizationDistrict.new(district_name, organization)
+      model_name = district_name
```
kitallis: Can this be named something else? Got thrown off thinking that `model_name` was the name of the model itself like `Facility` or `District`. Don't really have a suggestion though.

app/models/patient_business_identifier.rb
timcheadle: I don't follow... is `simple_bp_passport` defined somewhere?

app/models/patient.rb
timcheadle: Yeah... this scope isn't defined? Are you asking me to make it?

app/mailers/patient_list_download_mailer.rb
timcheadle: Yes, so they make it into the view

app/jobs/patient_list_download_job.rb
timcheadle: You're right, this is all super confusing. Can I fix in a separate PR? I'd like to separate district from facility concerns, and just make them use a single partial, instead of having these conditionals all over.

app/models/patient.rb
kitallis: Rails enums add scopes to models, since `identifier_type` is an enum, `simple_bp_passport` should be an available scope on `business_identifiers` already.

app/mailers/patient_list_download_mailer.rb
kitallis: Makes sense.

app/models/patient_business_identifier.rb
kitallis: Rails enums should also give you booleans for asserting whether the `identifier_type` is of a particular kind like `simple_bp_passport?` automatically?

kitallis: Rails enums should also give you booleans for asserting whether the `identifier_type` is of a particular kind like `simple_bp_passport?` automatically.

PR #587, opened 2019-10-04
Update monthly cohort calculation

timcheadle:
Our monthly cohorts should calculate as follows:

In April, we see:
- Patients registered in January
- ... who followed up any time between February and March

image


app/queries/cohort_analytics_query.rb
```diff
@@ -5,17 +5,17 @@ def initialize(patients, period = :month)
...
 
-  def patient_counts_by_period(period, prev_periods)
+  def patient_counts_by_period(period, prev_periods, from_date: Date.today)
     results = {}
 
     (0..(prev_periods)).each do |periods_back|
       if period == :month
-        report_start = (Time.current - periods_back.months).beginning_of_month
-        report_end   = report_start.end_of_month
+        report_start = (from_date - (1 + periods_back).months).beginning_of_month
```
govindkrjoshi: If we aren't skipping the current quarter, I think we shouldn't skip the current month either. We will have to actually wait longer for the current quarter to have complete data, than for a month.

app/views/shared/_cohort_charts.html.erb
```diff
@@ -32,7 +32,7 @@
         report_period = quarter_string(report_start)
         cohort_period = quarter_string(cohort_start)
       else
-        report_period = report_start.strftime("%b %Y")
+        report_period = [report_start.strftime("%b"), (report_start + 1.month).strftime("%b")].join(" - ")
```
govindkrjoshi: This labeling can be confusing. For example, does `Jul-Aug` represented
- (start of july) to (start of august) OR
- (start of july) to (end of august) 

Also if I want to look at what happened in month in a quarter, The labeling suggest there are only two bars to look at, but there should be 3.

timcheadle:
It's clear that these charts are somewhat confusing. @dburka let's work on improving the labeling.

@govindkrjoshi you can't compare the monthly charts to quarterly charts; it's an entirely different cohort. The only numbers that add up are, for example, "Registered in [Jan, Feb, Mar]" should match "Registered in Q1"

timcheadle:
@govindkrjoshi @dburka I've updated the month labels to include the day of the month (see the pic in the PR description)

app/queries/cohort_analytics_query.rb
timcheadle: This is a design decision we made a while ago. I agree that the inconsistency is potentially confusing and makes the code more complex.

@dburka What do you think of only showing the most recently completed quarter or month, not showing the current quarter or month?

Showing the current quarter allows people to see how they're doing right now, as opposed to waiting 3 months. Of course, "didn't attend" for the current quarter includes patients that just haven't had time to show up yet, inflating the numbers a bit.

Showing the "current month" is tricky, as monthly cohort spans 2 months (e.g. followed up between Feb and March). This code skips the current month and only shows the last completed monthly cohort (from last month).

PR #590, opened 2019-10-04
Remove audit log routes, controller and links

vkrmis:


config/routes.rb
```diff
@@ -157,7 +157,6 @@
   get "admin", to: redirect("/")
 
   namespace :admin do
-    resources :audit_logs, only: [:index, :show]
```
govindkrjoshi:
```suggestion
```

PR #596, opened 2019-10-08
Add reminder_consent to patient model and APIs

vkrmis:
Co-authored-by: Govind Krishna Joshi <govind@nilenso.com>


db/migrate/20191007100156_add_reminder_consent_to_patients.rb
```diff
@@ -0,0 +1,5 @@
+class AddReminderConsentToPatients < ActiveRecord::Migration[5.1]
+  def change
+    add_column :patients, :reminder_consent, :string, index: true, null: false, default: Patient.reminder_consents[:denied]
```
kitallis: Maybe make this `concurrently` since this will re-write the whole table with the default value.

spec/transformers/api/current/patient_transformer_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.describe Api::Current::PatientTransformer do
+  describe 'from_nested_request' do
+    context 'request payload has reminder consent' do
+      let(:request_patient) { build_patient_payload.merge('reminder_consent' => 'denied') }
+
+      it "adds a default `granted` reminder_consent" do
```
kitallis: I think the `reminder_consent` should be `denied` here? (in the spec title)

db/migrate/20191007100156_add_reminder_consent_to_patients.rb
kitallis: Just test this on 140k patients and see if the default column value is updated throughout the table in a reasonable time.

spec/transformers/api/v2/patient_transformer_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.describe Api::V2::PatientTransformer do
+  describe 'from_nested_request' do
+    context 'request payload has reminder consent' do
+      let(:request_patient) { build_patient_payload.merge('reminder_consent' => 'denied') }
+
+      it "adds a default `granted` reminder_consent" do
```
kitallis: I think the `reminder_consent` should be `denied` here? (in the spec title)

PR #606, opened 2019-10-10
Configure capistrano for multi-country deployments

kitallis:
https://www.pivotaltracker.com/story/show/168945237

**PLEASE READ BEFORE MERGING**

* It's important to run `bundle exec cap india:sandbox sidekiq:install` and `bundle exec cap bangladesh:sandbox sidekiq:install` respectively before trying to deploy. This is because the environments are now namespaced under the country name and hence the sidekiq service file needs an updated name.
* Deploying to a particular country can be done using `bundle exec cap india:sandbox deploy`
* To deploy using the deploy script, do `bin/deploy.rb bangladesh sandbox <tag>`
* Run `bundle exec cap -T` to see all the other tasks
* Currently, there's no support to deploy to all countries at once


Capfile
```diff
@@ -13,4 +14,4 @@ require 'capistrano-db-tasks'
 
 require "whenever/capistrano"
 
-Dir.glob("lib/capistrano/tasks/*.rake").each { |r| import r }
+Dir.glob("lib/capistrano/tasks/*.rake").each(&method(:import))
```
govindkrjoshi: Why is this change required?

kitallis: It was just a rubucop suggestion

PR #625, opened 2019-10-22
Add support to easily translate API strings using transifex

govindkrjoshi:
- Moves api strings to namespaced yml
- Moves user analytics to locale yml
- Read help screen body from localized .html files instead of .erb


app/controllers/api/current/help_controller.rb
```diff
@@ -7,6 +7,21 @@ class Api::Current::HelpController < APIController
   skip_before_action :validate_facility, only: [:show]
   skip_before_action :validate_current_facility_belongs_to_users_facility_group, only: [:show]
 
+  HTML_TRANSLATIONS_DIR = 'config/locales/api/help/*.html'
```
vkrmis: Could this be HELP_TRANSLATIONS_DIR instead? 
Since its limited to the help htmls

app/views/api/current/analytics/user_analytics/show.html.erb
```diff
@@ -17,7 +17,7 @@
 <% else %>
   <div class="card">
     <%= inline_svg('icon_new_bp.svg', classname: 'icon-float') %>
-    <h3>How many follow-up patients visited <%= @current_facility.name %>?</h3>
+    <h3><%= t('analytics.bps_recorded', facility_name: @current_facility.name) %></h3>
```
vkrmis: analytics.followup_patients might be better here. Since we changed this graph to only show followups.

PR #644, opened 2019-10-28
Add data migration for existing BloodPressures to have Encounters

kitallis:
* This assumes the user to be the user that recorded the BloodPressure


lib/tasks/data_migration.rake
```diff
@@ -86,4 +86,31 @@ namespace :data_migration do
   task grant_reminder_consent_for_all_patients: :environment do
     Patient.update_all(reminder_consent: Patient.reminder_consents[:granted])
   end
+
+  desc 'Backport all BloodPressures to have Encounters and appropriate Observations'
+  task :add_encounters_to_existing_blood_pressures => :environment do |_t, _args|
+    batch_size = ENV['BACKFILL_ENCOUNTERS_FOR_BPS_BATCH_SIZE'].to_i || 1000
+    timezone_offset = 19800 # For 'Asia/Kolkata'
```
govindkrjoshi: This could be a config variable as well.

kitallis:
> for the different environments be raised as a dependent PR prior to this getting merged, right?

The configs are actually already merged in.

PR #653, opened 2019-10-31
Add a ManifestController to hold permanent country-specific details

kitallis:
* Points Bangladesh SBX and QA to a single default SBX/QA env
* Also adds a development manifest for local testing


app/controllers/manifests_controller.rb
```diff
@@ -0,0 +1,5 @@
+class ManifestsController < ApplicationController
+  def show
+    render file: "public/manifest/#{ENV['SIMPLE_SERVER_ENV']}.json", status: :ok
```
govindkrjoshi: There is also https://apidock.com/rails/ActionController/Streaming/send_file

spec/controllers/manifest_controller_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe ManifestsController, type: :controller do
+  describe 'GET #show' do
+    ['production', 'sandbox', 'demo', 'qa', 'development'].each do |env|
+      it "for #{env}" do
+        ENV['SIMPLE_SERVER_ENV'] = env
```
govindkrjoshi: Its better to stub it like this since it won't affect the other tests. 
```suggestion
        expect(ENV).to receive([]).with('SIMPLE_SERVER_ENV').and_return(env)
```

PR #669, opened 2019-11-07
Feature toggle encounters sync

vkrmis:
This is to artificially return successes while syncing encounters to smoothen out the app release process. This is largely to suppress the couple of issues we've found while testing Encounters and to avoid blocking the release around it.


app/controllers/api/current/encounters_controller.rb
```diff
@@ -2,11 +2,24 @@ class Api::Current::EncountersController < Api::Current::SyncController
   include Api::Current::PrioritisableByFacility
 
   def sync_from_user
-    __sync_from_user__(encounter_params)
+    if FeatureToggle.enabled?('SYNC_ENCOUNTERS')
+      __sync_from_user__(encounter_params)
+    else
+      render json: { errors: nil }, status: :ok
```
govindkrjoshi: Do you think this is better written as a before action?

app/controllers/api/current/encounters_controller.rb
```diff
@@ -2,11 +2,24 @@ class Api::Current::EncountersController < Api::Current::SyncController
...
   end
 
   def sync_to_user
-    __sync_to_user__('encounters')
+    response_key = 'encounters'
+
+    if FeatureToggle.enabled?('SYNC_ENCOUNTERS')
+      __sync_to_user__(response_key)
+    else
+      render(
```
govindkrjoshi: Same here

config/initializers/config.rb
```diff
@@ -38,5 +38,6 @@
   'APPOINTMENT_NOTIFICATION_FACILITY_IDS',
   'ENABLE_SMS_REMINDERS',
   'ANALYTICS_DASHBOARD_CACHE_TTL',
-  'IHCI_ORGANIZATION_UUID'
+  'IHCI_ORGANIZATION_UUID',
+  'ENABLE_SYNC_ENCOUNTERS'
```
govindkrjoshi: It's not necessary for this key to be required. 

Feature toggles are assumed to be off unless they are explicitly set to `true`

PR #678, opened 2019-11-15
Add zone to facility

kitallis:
- [x] Add zone (address field) to Facility model
- [x] Make facility edit/new forms locale-sensitive
- [x] Allow configuring a default locale for a particular region


app/controllers/admin_controller.rb
```diff
@@ -13,6 +13,15 @@ class AdminController < ApplicationController
     head :bad_request
   end
 
+  def switch_locale(&action)
+    locale =
+      ENV['DEFAULT_PREFERRED_DASHBOARD_LOCALE'].presence ||
```
govindkrjoshi: Here we are making an assumption that each country has their own servers. 
Should we leave a comment for the same?

app/views/admin/facilities/_form.html.erb
```diff
@@ -6,6 +6,7 @@
       <%= form.text_field :street_address, id: :facility_street_address %>
       <%= form.text_field :village_or_colony, id: :facility_village_or_colony %>
       <%= form.text_field :district, id: :facility_district, required: true %>
+      <%= form.text_field :zone, id: :facility_district, required: true unless I18n.locale.eql?(:en_IN) %>
```
govindkrjoshi:
```suggestion
      <%= form.text_field :zone, id: :facility_district, required: true if I18n.locale.eql?(:en_BD) || I18n.locale.eql?(:bn_BD)%>
```

config/locales/address/en_BD.yml
```diff
@@ -0,0 +1,11 @@
+en_BD:
+  helpers:
```
govindkrjoshi: Why not use https://guides.rubyonrails.org/i18n.html#translations-for-active-record-models?

kitallis: To avoid using `Model.human_attribute_name(attribute)`. Since our current requirement is literally only labels, we can add this when we start using the labels throughout the codebase.

app/controllers/admin_controller.rb
kitallis: It's not necessarily per country, it's locale-per-deployment, whatever that might be. We can later have a locale / language switcher in the dashboard, but for now we can set this for any of our deployments appropriately.

PR #680, opened 2019-11-18
Added schema and scaffolding for Diabetes Observations API

vkrmis:
Co-authored-by: Govind Krishna Joshi <govind@nilenso.com>


vkrmis:
> Why would we not call the model `BloodSugar`?

Makes sense, will rename it. I went with `observations` to stay in line with FHIR naming.

spec/api/current/blood_sugars_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    get 'Syncs blood sugar data from server to device.' do
+      tags 'Blood Sugar'
+      security [basic: []]
+      parameter name: 'HTTP_X_USER_ID', in: :header, type: :uuid
+      parameter name: 'HTTP_X_FACILITY_ID', in: :header, type: :uuid
+      Api::Current::Schema.sync_to_user_request.each do |param|
+        parameter param
+      end
+
+      # before :each do
```
kitallis: Is this used in subsequent PRs?

app/schema/api/current/models.rb
```diff
@@ -150,6 +150,23 @@ def blood_pressure
...
+          blood_sugar_type: { type: :string, enum: BloodSugar.blood_sugar_types.keys },
+          blood_sugar_value: { type: :integer },
+          deleted_at: { '$ref' => '#/definitions/nullable_timestamp' },
+          created_at: { '$ref' => '#/definitions/timestamp' },
+          updated_at: { '$ref' => '#/definitions/timestamp' },
+          recorded_at: { '$ref' => '#/definitions/timestamp' },
+          patient_id: { '$ref' => '#/definitions/uuid' },
+          facility_id: { '$ref' => '#/definitions/uuid' },
+          user_id: { '$ref' => '#/definitions/uuid' } },
+        required: %w[blood_sugar_type blood_sugar_value created_at updated_at patient_id facility_id user_id]
```
kitallis: I think `id` should be required as well.

spec/api/current/blood_sugars_spec.rb
kitallis: If not, we can get rid of it.

govindkrjoshi: This is a part of #681

PR #684, opened 2019-11-25
Add diabetes api

vkrmis:
https://www.pivotaltracker.com/story/show/169808882

This adds an API endpoint to sync Blood Sugars, and creates appropriate encounters.


app/controllers/api/current/blood_sugars_controller.rb
```diff
@@ -1,15 +1,49 @@
...
+  end
+
+  def merge_if_valid(blood_sugar_params)
+    validator = Api::Current::BloodSugarPayloadValidator.new(blood_sugar_params)
+    logger.debug "Blood Sugar payload had errors: #{validator.errors_hash}" if validator.invalid?
+    if validator.invalid?
+      NewRelic::Agent.increment_metric('Merge/BloodSugar/schema_invalid')
+      { errors_hash: validator.errors_hash }
+    else
+      set_patient_recorded_at(blood_sugar_params)
```
kitallis: I think we don't need this for `BloodSugars` since this is a fresh API, so we need not set the `recorded_at` in real-time as all new apps who have support for Diabetes will send `recorded_at`

app/controllers/api/current/encounters_controller.rb
```diff
@@ -24,14 +24,21 @@ def generate_id
 
   private
 
+  def encounter_facility_id(encounter_params)
+    if encounter_params['observations'].values.flatten.empty?
```
kitallis: [STYLE] We can inline this `if` check?

app/controllers/concerns/api/current/sync_encounter_observation.rb
```diff
@@ -0,0 +1,60 @@
...
+        if FeatureToggle.enabled?('SYNC_ENCOUNTERS')
+          # this will always return a single blood_sugar
+          add_encounter_and_merge_record(key, params)[:observations][key][0]
+        else
+          key.to_s.classify.constantize.merge(params)
+        end
+      end
+    end
+
+    def add_encounter_and_merge_record(key, params)
```
kitallis: Should we rename the `key` to `observation_name` or something similar?

app/controllers/concerns/api/current/sync_encounter_observation.rb
```diff
@@ -0,0 +1,60 @@
+module Api::Current::SyncEncounterObservation
+  extend ActiveSupport::Concern
+  included do
+
+    def merge_encounter_observation(key, params)
+      ActiveRecord::Base.transaction do
+        if FeatureToggle.enabled?('SYNC_ENCOUNTERS')
+          # this will always return a single blood_sugar
```
kitallis: Since this is a concern, this comment can probably mention that it will return a single observation rather than a single `blood_sugar`?

app/services/merge_encounter_service.rb
```diff
@@ -19,13 +19,16 @@ def merge
...
   def add_observations(encounter, observation_params)
-    {
-      :blood_pressures =>
-        observation_params[:blood_pressures].map do |params|
-          blood_pressure = BloodPressure.merge(params)
-          blood_pressure.find_or_update_observation!(encounter, user)
-          blood_pressure
-        end
-    }
+    n = [:blood_pressures, :blood_sugars].map do |key|
```
kitallis: Let's make this array a service-level constant.

kitallis: Let's call it something other than `n`?

spec/controllers/api/current/encounters_controller_spec.rb
```diff
@@ -97,6 +97,11 @@ def create_record_list(n, options = {})
         expect(response).to have_http_status(200)
         expect(Encounter.find(encounter[:id]).patient).to eq patient
       end
+
+      context 'encounter contains observations from more than one facility' do
```
kitallis: Empty spec?

app/controllers/concerns/api/current/sync_encounter_observation.rb
```diff
@@ -0,0 +1,60 @@
...
+          add_encounter_and_merge_record(key, params)[:observations][key][0]
```
kitallis: This is a relic from before, but maybe now `add_encounter_and_merge_record` can always return an encounter with a single observation so we don't have to do this `[0]` business? What do you think?

kitallis:
Can we put some description in the PR? At the very least a link to the tracker story?

app/controllers/api/current/blood_sugars_controller.rb
vkrmis: We still need to set_patient_recorded_at for retroactive BloodSugar records

PR #685, opened 2019-11-26
Notify only patients with reminder consent granted

kitallis:
https://www.pivotaltracker.com/story/show/169944334


app/services/appointment_notification_service.rb
```diff
@@ -14,6 +16,7 @@ def send_after_missed_visit(days_overdue: 3, schedule_at:)
   attr_reader :appointments, :schedule_at
 
   def fan_out_reminders(appointments, communication_type, schedule_at)
+
```
vkrmis:
```suggestion
```

PR #688, opened 2019-11-27
Return 404 if the Twilio delivery callback fails to find a session

kitallis:
So that Twilio knows that the delivery status update was unsuccessful.

### DIFF
--- a/app/controllers/api/current/twilio_sms_delivery_controller.rb
+++ b/app/controllers/api/current/twilio_sms_delivery_controller.rb
@@ -3,8 +3,14 @@ class Api::Current::TwilioSmsDeliveryController < ApplicationController
   skip_before_action :verify_authenticity_token
 
   def create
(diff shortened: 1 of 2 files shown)


app/controllers/api/current/twilio_sms_delivery_controller.rb
```diff
@@ -3,8 +3,14 @@ class Api::Current::TwilioSmsDeliveryController < ApplicationController
   skip_before_action :verify_authenticity_token
 
   def create
-    TwilioSmsDeliveryDetail.where(session_id: params['SmsSid']).first.update(update_params)
-    head :ok
+    existing_session = TwilioSmsDeliveryDetail.where(session_id: params['SmsSid'])
+
+    if existing_session.blank?
+      head :not_found
```
govindkrjoshi: We should update the session in a sidekiq job so that it can be retried.

kitallis:
Closing this because retries should be handled via sidekiq instead of relying on Twilio. That change is much bigger, hence we're closing this PR for now.

PR #699, opened 2019-12-06
SQL view for patient line lists

timcheadle:
Created a SQL view to back the patient line list results we often look at. I've also added a `PatientSummary` view-backed model to access this data more cleanly throughout the reporting code.

This view should represent everything we need from the current patient line list, including:

```
      "Registration Date",
      "Registration Quarter",
      "Patient Name",
      "Patient Age",
      "Patient Gender",
      "Patient Phone Number",
      "Patient Village/Colony",
      "Patient District",
      "Patient State",
      "Registration Facility Name",
      "Registration Facility Type",
      "Registration Facility District",
      "Registration Facility State",
      "Latest BP Systolic",
      "Latest BP Diastolic",
      "Latest BP Date",
      "Latest BP Quarter",
      "Latest BP Facility Name",
      "Latest BP Facility Type",
      "Latest BP Facility District",
      "Latest BP Facility State",
      "Days Overdue",
      "Risk Level",
      "BP Passport ID",
      "Simple Patient ID"
```


timcheadle:
There may be a bug with the way I'm calculating `current_age`. It takes the current year minus the year of `age_updated_at`, but this could add a year before it should.

e.g. age 50, updated Dec 31 2019, today is Jan 1 2020. Age should still be 50.

db/views/patient_summaries_v01.sql
```diff
@@ -0,0 +1,88 @@
...
+latest_bp_facility.state AS latest_bp_state,
+greatest(0, date_part('day', NOW() - next_appointment.scheduled_date)) AS days_overdue,
+next_appointment.scheduled_date AS next_appointment_scheduled_date,
+next_appointment_facility.name AS next_appointment_facility_name,
+next_appointment_facility.facility_type AS next_appointment_facility_type,
+next_appointment_facility.district AS next_appointment_district,
+next_appointment_facility.state AS next_appointment_state,
+
+(
+    CASE
```
kitallis: The risk level calculations don't take into account one of the scenarios mentioned [here](https://docs.simple.org/engineering/backend/major-flows#high-risk-label-in-overdue-list): when the patient's latest scheduled appointment is overdue under a month. Currently implemented in [this](https://github.com/simpledotorg/simple-server/blob/master/app/models/patient.rb#L99).

I believe another scenario where it's overdue > 365 days is moot since overdue itself now is capped to 365 days. Currently implemented [here](https://github.com/simpledotorg/simple-server/blob/master/app/models/patient.rb#L180).

db/views/patient_summaries_v01.sql
```diff
@@ -0,0 +1,88 @@
+SELECT
+p.recorded_at,
+CONCAT(date_part('year', p.recorded_at), ' Q', EXTRACT(QUARTER FROM p.recorded_at)) AS registration_quarter,
```
kitallis: All date calculations in the SQL-land will be in UTC. So we shouldn't directly use these dates as is in the views anywhere.

kitallis:
I think this generally looks good. However, we should actually create smaller more composable views for the sub-queries that are being `LEFT OUTER JOIN`d. For example:

* latest_bp_passport
* latest_bp
* latest_phone_number
* next_appointment

This will help us reuse and compose. That can be done as a separate PR.

PR #709, opened 2019-12-11
Send reminder notifications to mobile numbers only

prabhanshuguptagit:
Skips reminder notification SMS to landline and invalid numbers.

https://www.pivotaltracker.com/story/show/169837941


app/models/patient.rb
```diff
@@ -78,6 +78,10 @@ def latest_phone_number
     phone_numbers.last&.number
   end
 
+  def latest_mobile_number
+    phone_numbers.where(phone_type: 'mobile').last&.number
```
govindkrjoshi:
```suggestion
    phone_numbers.where(phone_type: 'mobile').order(device_created_at: :desc).first
```

spec/factories/patient_phone_numbers.rb
```diff
@@ -8,6 +8,15 @@
     device_updated_at { Time.current }
     dnd_status { true }
     patient
+
+    trait :landline do
+      phone_type { 'landline' }
```
govindkrjoshi: The trait is unnecessary. We can simple use 
```
create(:patient_phone_number, phone_type: :landline
```

spec/services/appointment_notification_service_spec.rb
```diff
@@ -76,5 +76,21 @@
       eligible_appointments = overdue_appointments.select { |a| a.communications.present? }
       expect(eligible_appointments.count).to eq(1)
     end
+
+    it 'should skip sending reminders to patients with just landline or invalid numbers' do
```
govindkrjoshi: You can wrap this in a context.

spec/services/appointment_notification_service_spec.rb
```diff
@@ -76,5 +76,21 @@
...
+        landline_number = create(:patient_phone_number, :landline)
```
govindkrjoshi: [STYLE] The assignments can become `let` block, will make the it block easier to read.

app/models/patient.rb
kitallis: Maybe also consider making the `where` a scope on PhoneNumber itself.

app/services/appointment_notification_service.rb
```diff
@@ -25,7 +25,7 @@ def eligible_appointments(appointments, communication_type)
     end
 
     def eligible_for_sending_sms?(appointment, communication_type)
-      (not appointment.previously_communicated_via?(communication_type)) && appointment.patient&.phone_number?
+      (not appointment.previously_communicated_via?(communication_type)) && appointment.patient&.mobile_number?
```
kitallis: Consider scoping the Appointments to also only be included if they patient has a mobile based phone number, in this service.

app/models/patient.rb
prabhanshuguptagit: @govindkrjoshi There's a default scope on PhoneNumber which orders asc. The query effectively looks like: 
```ruby
phone_numbers.unscope(:order).mobile.order(device_created_at: :desc).first&.number
```

Should I try and get rid of the default scope?

PR #710, opened 2019-12-11
Add My Facilities Dashboard 

vkrmis:
This adds the new "My Facilities" overview page. The "My Facilities" dashboards are still a work in progress, but the overview page is releasable as-is. UI navigation links to the new dashboards will remain hidden for the time being, controlled by a feature flag which can be flipped when more dashboards are available.

https://www.pivotaltracker.com/story/show/170275609
https://www.figma.com/file/kONxBLSXwcQDnV7ya4hIXA/Simple-App?node-id=2305%3A398  
https://docs.google.com/document/d/1VITcxxzaMU21D5NFYpQBe3Y9iYSvTAZb_tP-BuF20vc


app/views/my_facilities/blood_pressure_control.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="row">
+  <div class="col-lg-12 pr-lg-0">
+    <%= render "shared/my_facilities/blood_pressure_control" %>
```
harimohanraj89: Do we need a partial for a single render?

app/views/shared/my_facilities/_blood_pressure_control.html.erb
```diff
@@ -0,0 +1,151 @@
...
+      <td class="type-number">30</td>
+      <td class="type-percent"><em>50%</em></td>
+      <td class="type-number">50</td>
+      <td class="type-percent"><em>20%</em></td>
+      <td class="type-number">20</td>
+      <td class="type-percent"><em>18%</em></td>
+      <td class="type-number">456</td> 
+    </tr>
+    <tr>
+      <td class="type-title"><a href="#">CHC Cat Location</a></td>
```
harimohanraj89: Oh this is fake data for now. Hmm, should we be shipping this to production?

spec/features/admin/users_spec.rb
```diff
@@ -29,7 +29,7 @@
     context "recent BP log" do
       it "displays the user's recent BPs", :aggregate_failures do
         within("#recent-bps") do
-          expect(page).to have_selector("th", text: "Facility")
+          expect(page).to have_selector("th", text: "FACILITY")
```
harimohanraj89: Can we use a regex to avoid this annoying case problem? `/facility/i`?

spec/queries/my_facilities_query_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    let!(:blood_pressures_for_inactive_facility) { create_list(:blood_pressure, 9, facility: inactive_facility, recorded_at: Time.now) }
+    let!(:encounters) do
+      (blood_pressures_for_active_facility + blood_pressures_for_inactive_facility).each { |record|
+        create(:encounter, :with_observables, patient: record.patient, observable: record, facility: record.facility) }
+    end
+
+    describe ".inactive_facilities" do
+      it 'should return only inactive facilities' do
+        facility_ids = [active_facility.id, inactive_facility.id, inactive_facility_with_zero_bps.id]
+        expect(MyFacilitiesQuery.inactive_facilities(Facility.where(id: facility_ids))).to match_array([inactive_facility, inactive_facility_with_zero_bps])
```
harimohanraj89: Since this test case is covering all the facilities created in this context, can we just `expect(MyFacilitiesQuery.inactive_facilites).to match_array...`?

harimohanraj89: Depending on our decision here, could we add another test for the other scenario? Effectively, I'd love to have one test for what happens when we pass in a scope, and one test for what happens when we don't (defaulting to `.all`)

app/controllers/my_facilities_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    @users_requesting_approval = paginate(@users_requesting_approval)
+
+    @facilities = policy_scope([:manage, :facility, Facility])
+    @inactive_facilities = MyFacilitiesQuery.inactive_facilities(@facilities)
+
+    @facility_counts_by_size = { total: @facilities.group(:facility_size).count,
+                                 inactive: @inactive_facilities.group(:facility_size).count }
+
+    @bp_counts = { last_week: @inactive_facilities.bp_counts_in_period(1.week.ago, Time.current),
```
harimohanraj89: Are these BP counts for inactive facilities? If so, could we indicate that in this var name? eg.  `@inactive_facility_bp_counts`?

app/controllers/my_facilities_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+  include SizeFiltering
+
+  before_action :authorize_my_facilities, only: %i[index ranked_facilities blood_pressure_control]
+
+  def index
+    @users_requesting_approval = policy_scope([:manage, :user, User])
+                                 .requested_sync_approval
+                                 .order(updated_at: :desc)
+
+    @users_requesting_approval = paginate(@users_requesting_approval)
```
harimohanraj89: We could just wrap this paginate directly around the line above if you'd like. Not a big deal.

app/controllers/my_facilities_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    @facilities = policy_scope([:manage, :facility, Facility])
+    @inactive_facilities = MyFacilitiesQuery.inactive_facilities(@facilities)
+
+    @facility_counts_by_size = { total: @facilities.group(:facility_size).count,
+                                 inactive: @inactive_facilities.group(:facility_size).count }
```
harimohanraj89: Can we make _all_ of the queries in this controller through the `MyFacilitiesQuery` class? It seems like a great place to put them, since the purpose of that class is to encapsulate queries needed to populate the My Facilities dashboard.

app/models/facility.rb
```diff
@@ -99,6 +99,13 @@ def self.parse_facilities(file_contents)
     facilities
   end
 
+  def self.bp_counts_in_period(period_start, period_end)
```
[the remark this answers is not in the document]
harimohanraj89: Let's add test coverage for this method.

app/views/shared/my_facilities/_ranked_facilities.html.erb
```diff
@@ -0,0 +1,234 @@
...
+    </tbody>
+  </table>
+</div>
+<section class="footnotes">
+    <ol>
+        <li><strong>Ranking:</strong> is ranked by...</li>
+        <li><strong>BP control:</strong> Had at least one visit in last 90 days and BP was &lt;140/90 at last visit.</li>
+        <li><strong>OPD load:</strong> Outpatient department load is entered by surveillance officers for each facility. Monthly OPD load is an average. Where a facility has no OPD Load entered (see asterisk), the number is an average of other faciities of the same type in this district.</li>
+    </ol>
+</section>
```
harimohanraj89: A couple of sneaky missing newlines, hehe. I'll let you sort that out with CodeClimate. You could also run the following locally to auto-fix

```
$ gem install rubocop
$ rubocop --auto-correct --only 'Layout/TrailingEmptyLines'
```

app/controllers/my_facilities_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+  def ranked_facilities; end
+
+  def blood_pressure_control; end
+
+  private
+
+  def authorize_my_facilities
+    authorize(:dashboard, :show?)
+  end
+end
```
harimohanraj89: Let's add test coverage for this brand new controller ✨

app/models/blood_pressure.rb
```diff
@@ -18,6 +18,7 @@ class BloodPressure < ApplicationRecord
 
   scope :hypertensive, -> { where("systolic >= 140 OR diastolic >= 90") }
   scope :under_control, -> { where("systolic < 140 AND diastolic < 90") }
+  scope :with_encounters, -> { includes(:encounter).where.not(encounters: { id: nil }) }
```
harimohanraj89: Looks like this new scope isn't being used anywhere else in the PR. Can we remove?

app/controllers/concerns/size_filtering.rb
```diff
@@ -0,0 +1,25 @@
...
+      @size = params[:size].present? ? params[:size] : 'All'
+    end
+
+    def set_facility_sizes
+      @facility_sizes = Facility.facility_sizes.keys.reverse
+    end
+
+    def facilities_by_size(scope_namespace = [])
+      @facilities = @size == 'All' ? Facility.all : Facility.where(facility_size: @size)
+      policy_scope(scope_namespace.concat([@facilities]))
```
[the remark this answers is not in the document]
vkrmis: Yup, you would pass it the scope like this: https://github.com/simpledotorg/simple-server/blob/6587bf75b9b2a0e484b52e7529fa384a1f828bfb/app/controllers/my_facilities_controller.rb#L29

app/controllers/my_facilities_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    @facility_counts_by_size = { total: @facilities.group(:facility_size).count,
+                                 inactive: @inactive_facilities.group(:facility_size).count }
+
+    @bp_counts = { last_week: @inactive_facilities.bp_counts_in_period(1.week.ago, Time.current),
+                   last_month: @inactive_facilities.bp_counts_in_period(1.month.ago, Time.current) }
+  end
+
+  def ranked_facilities; end
+
+  def blood_pressure_control; end
```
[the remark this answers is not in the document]
vkrmis: We have placeholder views for these, probably worth keeping

app/models/facility.rb
```diff
@@ -99,6 +99,13 @@ def self.parse_facilities(file_contents)
...
+    left_outer_joins(:blood_pressures)
+      .where('recorded_at IS NULL OR (recorded_at > ? AND recorded_at < ?)', period_start, period_end)
```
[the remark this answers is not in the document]
vkrmis: This is to include facilities that have zero blood pressures in the period, recorded_at = NULL preserves them for the group_by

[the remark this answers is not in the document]
harimohanraj89: Ahhh annoying but nice approach 😄 Let's make sure we add a test case for `it "includes facilities with no measurements"` for good measure 👍

app/controllers/my_facilities_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  private
+
+  def authorize_my_facilities
+    authorize(:dashboard, :show?)
```
harimohanraj89: Should this be `authorize(:dashboard, :view_my_facilities?)` to make use of [this bit](https://github.com/simpledotorg/simple-server/pull/710/files#diff-5182f4f26eef1925711f3c6a19e91fcbR45-R47)?

app/controllers/my_facilities_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class MyFacilitiesController < AdminController
+  include DistrictFiltering
+  include Pagination
+  include SizeFiltering
+
+  before_action :authorize_my_facilities, only: %i[index ranked_facilities blood_pressure_control]
```
harimohanraj89: I'd say it's likely that any additional actions in this controller would also need to be authorized the same. Can we drop the `only` clause?

app/views/layouts/my_facilities.html.erb
```diff
@@ -0,0 +1,168 @@
...
+          </form>
+        </section>
+      <% end %>
+    </div>
+    <div class="col-lg-10 pr-lg-0">
+      <%= yield %>
+    </div>
+  </div>
+<% end %>
+<%= render template: "layouts/application" %>
```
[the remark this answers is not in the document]
vkrmis: This is a nested layout, is there another way to do this?

app/views/my_facilities/blood_pressure_control.html.erb
```diff
@@ -0,0 +1,5 @@
...
+  </div>
+</div>
```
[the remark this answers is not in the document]
vkrmis: We're shipping this as a 'hidden' feature, thought it might be useful to showcase these upcoming pages to specific users to get their feedback.

spec/queries/my_facilities_query_spec.rb
vkrmis: This isn't covering all the facilities, the factories create some facilities implicitly.

app/controllers/my_facilities_controller.rb
vkrmis: We could do this later. For now, we want people who navigate to '/my_facilities' to be able to see the page. We're using the :view_my_facilities permission to hide the link in the layout.

app/controllers/my_facilities_controller.rb
vkrmis: We're trying to follow the QueryObject pattern for the new query classes, so the idea is to only return relations from the QueryObject itself. This means we'll have to do grouping, counting etc in the controller. We might consider moving these things to a ViewObject if it gets too hard to keep track of.

app/queries/my_facilities_query.rb
```diff
@@ -0,0 +1,16 @@
...
+
+# Contains queries required for the My Facilities Dashboards
+module MyFacilitiesQuery
+  def self.inactive_facilities(facilities = Facility.all)
+    facility_ids = facilities.left_outer_joins(:blood_pressures)
+                             .where('blood_pressures.recorded_at IS NULL OR blood_pressures.recorded_at > ?',
+                                    1.week.ago)
+                             .group('facilities.id')
+                             .count(:blood_pressures)
+                             .select { |_, count| count < 10 }
```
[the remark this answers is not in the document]
vkrmis: Yup, found this tricky to do, I'll come back to it

app/assets/stylesheets/application.scss
```diff
@@ -47,7 +47,7 @@ body {
     background: $grey-lightest;
 }
 
-h1 {
+h1, h1.btn {
```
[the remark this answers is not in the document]
vkrmis: @dburka ^

app/assets/stylesheets/application.scss
```diff
@@ -582,55 +768,14 @@ Dashboard UI
...
 .table-compact .row-control {
     text-align: center;
 }
 
-.table-compact .row-total {
-    background: #fff;
-    border-top: 1px solid $grey-light;
-    border-bottom: 1px solid $grey-light;
+.table-compact .row-total td {
+    background: #FFFAE3;
```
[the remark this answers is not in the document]
vkrmis: @dburka ^

app/views/layouts/my_facilities.html.erb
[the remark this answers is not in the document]
harimohanraj89: Oh oops! This is a _layout_. I totally understand what's going on now. I missed the [yield](https://github.com/simpledotorg/simple-server/pull/710/files#diff-3f903a6f3c4c99132c9701723a59544eR164) in here.

Is this layout automatically used for the `my_facilities` controller through some Rails magic?

app/views/my_facilities/blood_pressure_control.html.erb
harimohanraj89: Ah I see how the partials are the "graphs", and the view controls how to lay them out on a page. I like it.

spec/queries/my_facilities_query_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe MyFacilitiesQuery do
+  context 'utility methods' do
```
harimohanraj89: We can remove this context block. I don't think it adds too much useful info to the spec.

PR #717, opened 2019-12-13
Add facility sizes

timcheadle:
**Story:** https://www.pivotaltracker.com/story/show/169441390

This change adds a basic facility size attribute to `Facility`, based loosely on `facility_type`. It also adds a data migration task to backfill sizes if not present already.


govindkrjoshi:
@timcheadle Should the size of a facility be a tag, since we are already planning to add tags?

timcheadle:
> @timcheadle Should the size of a facility be a tag, since we are already planning to add tags?

I'd like to totally rethink how we organize facilities, but I don't want that to block implementation of our dashboard. This is OK for now. We will likely wind up migrating lots of things to tags (or whatever structure we decide).

PR #730, opened 2019-12-18
Update overdue definition

vkrmis:
Changes to the overdue definition so that
- Appointments which are overdue today are not counted
- Appointments are excluded from the list after `365 days` instead of `12 months`


app/models/appointment.rb
```diff
@@ -42,7 +42,8 @@ class Appointment < ApplicationRecord
 
   def self.overdue
     where(status: 'scheduled')
-      .where('scheduled_date <= ?', Date.current)
+      .where('scheduled_date < ?', Date.current)
+      .where("scheduled_date >= ?", 365.days.ago)
```
timcheadle: I'd make this `1.year.ago` for clarity, otherwise LGTM.

prabhanshuguptagit: When using `12 months ago` or `1 year ago` appointments from 366 days ago are included if the current year is a leap year. It was suggested to change this to simply `365 days` for consistency. Missed adding this in the comments.

timcheadle: Hah! I didn't know that. This is fine.

PR #745, opened 2020-01-01
Expose zone address field as block for Indian facilities

vkrmis:
This PR un-hides the zone field in facility addresses, and exposes it with the label "Block" for the en-IN locale.

This has been done so we can capture block level address information for facilities and use it for filtering data in the new dashboards.


vkrmis:
> What's up @vkrmis, could you add a link to the story or a couple of lines of context to this PR description? I don't think I quite understand the goal of these changes from just the PR title.

I've added some context in the PR description above, also fixed the form field id, good catch @harimohanraj89

harimohanraj89:
Love it 👍 I'll let one of the other devs provide official approval on this one.

PR #747, opened 2020-01-02
Setup improvements

harimohanraj89:
The setup instructions in the README list some manual steps to get the app up and running (eg. install ruby, install gems, etc). However, I also stumbled upon a setup script `bin/setup` that seems to automate _most_ of the required tasks, and works quite nicely. This PR aims to improve this script so it can be used to set up the Simple server locally.

The commit trail lists the changes pretty well, and I've left questions inline below 👇


db/schema.rb
```diff
@@ -509,7 +509,9 @@
       master_users.device_created_at,
       master_users.created_at,
       master_users.updated_at,
-      master_users.deleted_at
+      master_users.deleted_at,
+      master_users.role,
+      master_users.organization_id
```
harimohanraj89: This diff has erupted for a few reasons put together.

1. The [view](https://github.com/simpledotorg/simple-server/blob/master/db/views/users_v01.sql) selects all fields from the `master_users` table at the time the view was generated.
1. A couple fields were added to `master_users` afterwards, and these fields were not added to the view at the time
1. The setup instructions and script advise the use of `db:migrate` instead of `db:setup`, so all migrations are rerun, rather than loading the database schema from `schema.rb`.

My question is: Do we want `role` and `organization_id` to be added to the database view? If yes, I will add another migration to refresh the view with the additional fields. In either case, I will switch the setup to use `db:setup` or `db:schema:load`.

README.md
```diff
@@ -5,23 +5,46 @@
...
-First, you need to install ruby: https://www.ruby-lang.org/en/documentation/installation
-It is recommended to use rbenv to manage ruby versions: https://github.com/rbenv/rbenv
+
+To set up the Simple server for local development, clone the git repository and
+run the setup script included.
+
+```
+$ git clone git@github.com:simpledotorg/simple-server.git
+$ cd simple-server
+$ bin/setup
```
harimohanraj89: Just double-checking, is the `bin/setup` script intended for use?

harimohanraj89:
Though attached to an outdated diff, the question [here](https://github.com/simpledotorg/simple-server/pull/747/files#r362425316) is still relevant.

bin/setup
```diff
@@ -120,35 +120,28 @@ verify_bundled_gems (){
...
+  bundle exec yarn install 2>&1
 }
 
-verify_database_configuration (){
-  if [ -f config/database.yml ]
-  then
-    bundle exec rake db:create:all
+verify_database_migrations (){
+  echo 'Setting up database...'
+  migrate_existing_database || setup_database
```
harimohanraj89: If a database already exists, the script will merely run any pending migrations. This should ensure we don't clobber our local data if we accidentally run `bin/setup` more than once.

If migration fails (most commonly in this scenario due to a missing database), then we assume this is a fresh installation and proceed with `db:setup`.

bin/setup
```diff
@@ -120,35 +120,28 @@ verify_bundled_gems (){
...
+}
 
-    if [ $? -ne 0 ]
-    then
-      echo 'Unable to create the databases for you. Please ensure your database.yml is configured for your system and run rake db:create:all'
-      exit 0
-    fi
-  else
-    echo 'The config/database.yml file does not exist. Please create it.'
```
harimohanraj89: I thought it's ok to stop accounting for a missing `config/database.yml` since the file's checked into the codebase.

kitallis:
Thanks for reviving the `bin/setup` script. It is indeed something we've used before. It came as an open-source contribution from outside.

README.md
kitallis: It is something we've used before. It came as an open-source contribution from outside.

PR #749, opened 2020-01-06
Add documentation for permissions

govindkrjoshi:


doc/adding-new-user-permissions.md
```diff
@@ -0,0 +1,48 @@
+All available permissions are listed in app/policies/permissions.rb. The file also describes a
+set of defualt access roles. These permissions are used in Pundit policies to authorize access 
+and scope models accessible by the user. Each feature requires it's own set of policies, for this
+we use pundits namespaced policy feature (https://github.com/varvet/pundit#policy-namespacing).
```
harimohanraj89: Since these are all Markdown anyway, could we take advantage of link formatting?

```
we use Pundit's [namespaced policy feature](https://github.com/varvet/pundit#policy-namespacing)
```

doc/arch/006-single-user-model.md
```diff
@@ -0,0 +1,21 @@
...
+Combine User and Admin into a single User model, and have different authentication mechanism for
+both these models.
+
+- Authenticate syncing api using phone number authentication
+- Authenticate dashboard access using email authentication
+
+## Status
+Accepted
+
+## Consequences
```
harimohanraj89: Are there consequences to combining (good or bad) that we should capture? What happens if a user needs to view dashboards? Can a single user authenticate both ways? Are there future scenarios that are made harder because of this change? Easier?

PR #750, opened 2020-01-06
Fix failing age spec due to year rollover

vkrmis:
We're not sure why the -10 days was required on age_updated_at in the patient factory, it was causing a spec to fail due to year rollover.

Example failure: https://semaphoreci.com/resolvetosavelives/simple-server/branches/soft-delete-duplicate-patients/builds/10

### DIFF
--- a/spec/factories/patients.rb
+++ b/spec/factories/patients.rb
@@ -14,7 +14,7 @@ FactoryBot.define do
     status { Patient::STATUSES[0] }
     date_of_birth { Date.current if has_date_of_birth? }
     age { rand(18..100) unless has_date_of_birth? }
-    age_updated_at { Time.current - rand(10).days if age.present? }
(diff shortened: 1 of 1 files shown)


harimohanraj89:
Could we add a reference to the failing spec - either a Github link or a Semaphore build link - that this PR addresses?

vkrmis:
> Does this spec failure point to an underlying bug where, for example, 375 days can be miscalculated as 2 years if it cross over 2 new years? Do we want to take a stab at addressing that underlying bug, or get this fix out and ticket up the underlying bug for future debt cleanup?

I don't think its an underlying bug as such, this is for deriving age when we don't have a date of birth, and rely on an integer age instead. So we do a (age + (years since we updated the age)) to derive current age. This is always going to be inexact, and have edge cases where we're off by ~1 year.

PR #753, opened 2020-01-07
Soft delete duplicate patients

vkrmis:
Undoes the previous revert, and removes the deleted_at column migration for admin_access_controls


harimohanraj89:
Hotdamn, why was that timestamp a problem in #736?

vkrmis:
> Hotdamn, why was that timestamp a problem in #736?

I'm not sure what timestamp you mean @harimohanraj89, If you mean the year rollover thing, that was fixed and merged in #750, so its no longer an issue.

We had to revert #736 because it tried to add a column to a table that got dropped in #742

harimohanraj89:
Ah yup, I meant the `deleted_at` timestamp that you removed [here](https://github.com/simpledotorg/simple-server/pull/753/commits/2e4daa3555426ef7e56d8e402f589ca7ba22db48). Ok that makes sense 👍

PR #754, opened 2020-01-07
Add Rubocop configuration

harimohanraj89:
Pivotal: https://www.pivotaltracker.com/story/show/170573130

After some [Slack discussion](https://simpledotorg.slack.com/archives/CFHMC60P5/p1578312764151000), we decided to add a basic Rubocop configuration to the codebase as a starting point to start addressing code style issues systematically, saving code review time.

This PR adds a `.rubocop.yml` file, and also modifies a couple of rules just to start. Additionally, I have run one round of auto-correction on the `spec` directory (as it's lower risk) as an example of how we can start auto-correcting large chunks of the database to quickly bring our violations under control.

**Question:** Do we want to auto-correct more? `app/` too?


harimohanraj89:
33 "new" issues, but also ~800 issues fixed. 😅

kitallis:
Configuraions  (will keep updating this list):

* Remove: Missing top-level class documentation comment
* Line-limit should be 120

harimohanraj89:
@kitallis I'd say `frozen_string_literal: true` is important, and should be kept around. It will enable us to upgrade very smoothly to Ruby 3 without introducing any string-related errors. It also has a positive impact on performance, as it mitigates the number of string literals that are initialized and de-initialized at runtime.

kitallis:
> @kitallis I'd say `frozen_string_literal: true` is important, and should be kept around. It will enable us to upgrade very smoothly to Ruby 3 without introducing any string-related errors. It also has a positive impact on performance, as it mitigates the number of string literals that are initialized and de-initialized at runtime.
> 
> Would you agree?

Yes. I agree with using it, I may have incorrectly assumed that this was a run-time check and can be done by Rails (which is silly since it's a comment). This is a pre-run-time check which means we will have to add it to all files? I think RubyMine has a way to auto-inject this, but I was wondering if we could do this as a compile[sic]-time flag during builds?

harimohanraj89:
I've found it's not difficult to live with it sitting directly in source code. The effort to add and maintain these comments is usually less than the effort to set up compile-time (or build-time?) injection.

PR #759, opened 2020-01-16
Consolidate high risk

harimohanraj89:
Story: https://www.pivotaltracker.com/story/show/170594701

## Overview

This PR updates risk levels of BP patients. In order to do so, it also updates the levels of hypertension that a BP reading can be categorized in. Read on to find out what definitions are proposed in this PR. The reference rules for these changes is shown below.

---
image

---

### Critical

A BP is critical if:

* SP >= 180, or
* DP >= 110

### Hypertensive

A BP is hypertensive if:

* SP >= 140, or
* DP >= 90

"Hypertensive" is a superset of "Critical" BPs.

### Under Control

A BP is under control if it is **not** hypertensive. In other words,

* SP < 140, and
* DP < 90

---

## Patient Risk Levels

Given these :point_up: definitions of hypertensive levels for a BP, patient risk levels are defined as follows:

### High

Patients are `HIGH` risk if they are 30-365 days overdue and:

* their latest BP is critical, or
* they have history of heart attack or stroke only and their latest BP is hypertensive

### Regular

Patients are `REGULAR` risk if they are 30-365 days overdue and:

* their latest BP is hypertensive

### Low

Patients are `LOW` risk if they are more than 365 days overdue and their latest BP is under control.
This is the same as the existing definition of "Low" risk.

### None

Patients are `NONE` risk in all other cases. For example:

* They are not >30 days overdue for a visit, or
* They are 30-365 days overdue for a visit, but their BP is under control

This is the same as the existing definition of "None" risk.

---

## Other Developer Notes

Since the definitions of blood pressure criticality and patient risk levels were very self-contained, I took this opportunity to simplify the definitions and methods in this part of the codebase. Notably:

* Blood pressures can be `critical`, `hypertensive`, or `under_control` - Check out the thresholds defined [here](https://github.com/simpledotorg/simple-server/pull/759/files#diff-63473e8edd8f883bb9da3172fb9d28d4R9)
* I suspect that the distinction between `REGULAR`, `LOW`, and `NONE` is still important because it affects the way patients are sorted in overdue lists. Is this true?


kitallis:
> I suspect that the distinction between REGULAR, LOW, and NONE is still important because it affects the way patients are sorted in overdue lists. Is this true?

This is true, however, I would like to point out that the sorting which is used in the [overdue list](https://api.simple.org/appointments) is currently broken as it happens in the [view](https://github.com/simpledotorg/simple-server/blob/master/app/views/appointments/index.html.erb#L34) which is paginated. This means that the sort by risk is local to a page – which is mostly meaningless. This is technically a defect (and lodged as such).

There was an attempt to do this as a part of a [database view](https://github.com/simpledotorg/simple-server/blob/master/db/views/patient_summaries_v01.sql#L35), which is currently unused. An attempt has been made previously to sort in-memory before pagination but that proved to be extremely slow to run in production. 

Our current thinking is to move this sort to the database (via the dbview) to power the overdue list as the majority of the work for it is already done. We can also consider controller-level caching as a slightly bigger than a PR-comment conversation.

Knowing well the flaw in the sort, I think we should still keep the risk priority numbers in the application-layer for now until we figure out how to hook up the database view to the overdue page. The high-risk label will continue to function correctly the sorting issues notwithstanding.

app/models/medical_history.rb
```diff
@@ -35,4 +35,8 @@ def indicates_risk?
       diabetes_boolean ||
       chronic_kidney_disease_boolean
   end
+
+  def indicates_hypertension_risk?
```
kitallis: Since we have this, I believe `indicates_risk?` can go away (along with tests).

spec/models/medical_history_spec.rb
```diff
@@ -47,4 +47,34 @@
       expect(patient.medical_history.indicates_risk?).to eq(true)
     end
   end
+
+  describe '#indicates_hypertension_risk?' do
+    it 'returns true if there was a prior heart attack' do
+      patient = create(:patient)
+      create(:medical_history, prior_heart_attack_boolean: true, patient: patient)
+
+      expect(patient.medical_history.indicates_hypertension_risk?).to eq(true)
```
kitallis: Did we not use the fancy rspec boolean helper to do this like we've done [here](https://github.com/simpledotorg/simple-server/pull/759/files#diff-fcb8e53799332de9bb6227ebcb0cde53R39) because it'd read weird like `be_indicates_hypertension_risk`?

spec/models/medical_history_spec.rb
```diff
@@ -18,7 +18,7 @@
     it_behaves_like 'a record that is deletable'
   end
 
-  describe '#risk_history?' do
+  describe '#indicates_risk?' do
```
kitallis: If we remove the `indicates_risk?` method, we should also get rid of these tests.

spec/models/medical_history_spec.rb
```diff
@@ -47,4 +47,34 @@
...
+    end
+
+    it 'returns true if there was a prior stroke' do
+      patient = create(:patient)
+      create(:medical_history, prior_stroke_boolean: true, patient: patient)
+
+      expect(patient.medical_history.indicates_hypertension_risk?).to eq(true)
+    end
+
+    it 'returns false if there is diabetes history' do
```
kitallis: Since the method can return `nil`, which I presume why we've chosen `falsey` in the assertion, we can be clear about that in the `it` description as well.

app/models/patient.rb
```diff
@@ -115,10 +111,6 @@ def risk_priority
 
   def risk_priority_label
```
kitallis: A thought on this – this method is used in two places currently. One in the [CSV overdue list](https://github.com/simpledotorg/simple-server/blob/master/app/models/appointment.rb#L165) and in the [patient line list](https://github.com/simpledotorg/simple-server/blob/master/app/exporters/patients_exporter.rb#L78l). I was under the impression that it was used in the dashboard as well, but the [dashboard](https://github.com/simpledotorg/simple-server/blob/master/app/views/appointments/index.html.erb#L37) simply shows `High` based on the `high_risk?` method instead of calling this label function.

Since this method now just returns `High` shall we just rely on showing the `High` label in both CSV export + line list by reading off the `high_risk?` boolean? This will also relieve the model of seemingly view-related concerns.

What do you think?

kitallis:
> > I suspect that the distinction between REGULAR, LOW, and NONE is still important because it affects the way patients are sorted in overdue lists. Is this true?
> 
> This is true, however, I would like to point out that the sorting which is used in the [overdue list](https://api.simple.org/appointments) is currently broken as it happens in the [view](https://github.com/simpledotorg/simple-server/blob/master/app/views/appointments/index.html.erb#L34) which is paginated. This means that the sort by risk is local to a page – which is mostly meaningless. This is technically a defect (and lodged as such).
> 
> There was an attempt to do this as a part of a [database view](https://github.com/simpledotorg/simple-server/blob/master/db/views/patient_summaries_v01.sql#L35), which is currently unused. An attempt has been made previously to sort in-memory before pagination but that proved to be extremely slow to run in production.
> 
> Our current thinking is to move this sort to the database (via the dbview) to power the overdue list as the majority of the work for it is already done. We can also consider controller-level caching as a slightly bigger than a PR-comment conversation.
> 
> Knowing well the flaw in the sort, I think we should still keep the risk priority numbers in the application-layer for now until we figure out how to hook up the database view to the overdue page. The high-risk label will continue to function correctly the sorting issues notwithstanding.

Additionally, as another PR, we should revive the database view and start using it for displaying and sorting purposes. I'll check if there's a bug raised for this, if not, I'll make one and assign it to either you or Tim.

app/models/patient.rb
harimohanraj89: Ooh good eye, I'll take a look at this method again.

spec/models/medical_history_spec.rb
harimohanraj89: Whoops, forgot to update the spec description.

spec/models/medical_history_spec.rb
harimohanraj89: Haha yup exactly 😄 I like the magic helper because the tests read easier. In this case, since the test reads worse, I've avoided it.

app/models/medical_history.rb
harimohanraj89: Ah, coolbeans. I thought there might be merit in keeping it around, but you're right. If it's currently unused, then we should axe it (and tests)

spec/models/medical_history_spec.rb
harimohanraj89: Whoops, forgot to update the spec description.

PR #761, opened 2020-01-17
Use soft-deleted records while merging as well

govindkrjoshi:
Merge doesn't currently use soft-deleted records when merging. This will result in updates on discarded records to be interpreted as new inserts. To avoid this remove the kept scope when syncing. 

This bug is a side-effect of this issue: 
https://www.pivotaltracker.com/n/projects/2184102/stories/170671775


app/models/observation.rb
```diff
@@ -1,5 +1,5 @@
 class Observation < ApplicationRecord
   belongs_to :encounter
   belongs_to :user
-  belongs_to :observable, polymorphic: true
+  belongs_to :observable, polymorphic: true, optional: true
```
harimohanraj89: Is this change related to merging soft-deleted records?

kitallis: Thanks for catching this, removed!

doc/arch/006-no-op-updates-on-discarded-records.md
```diff
@@ -0,0 +1,38 @@
...
+## Context
+
+The addition of soft-deletes means that we need a way to deal with updates 
+that can happen on top of the soft-deleted records.
+
+An example could be follows,
+
+* Nurse A from Facility X soft-deletes BP#1 and syncs to the server
+* Nurse B from the same Facility X updates BP#1 without syncing changes from Nurse A
+* When Nurse B syncs, her update to BP#1 will take place, although on a technically discarded record
```
harimohanraj89: Why is this a problem?

Originally, I thought the problem was that Nurse B's sync would un-delete the user. Is that not the case?

kitallis: This is a problem because it works that way because of how the client behaves. The client behaves in a way *currently* where un-delete is not possible since for updates it doesn't send `deleted_at: nil` (so it stays deleted).

But relying on this behavior is not ideal since that behaviour is not intentional and not very easy to capture in an API.

So it's better to explicitly `no-op` on the server-side.

harimohanraj89: As you mention in your ADR, no-opping introduces some complexity when it comes to associated data like addresses, phone numbers, maybe even patient name? It raises the question:

If Nurse B in your example was updating the patient's address (and not recording another BP), do we want that update to persist against the patient or not? 

Is it simpler for us to update merging logic in such a way that `deleted_at` is always merged as `deleted_at_A || deleted_at_B`, and all other attributes are merged as last-one-wins.

kitallis: If Nurse B was updating the patient's address, we should no-op it (imo). Because it's both meaningless to update something that needn't be updated + the reliance on the client behavior as we discussed before.

Riffing off of your suggestion, it might be easiest, if cascading deletes were handled by the app instead. So, if the Nurse deletes a `patient`, the `address` and `phone_number` also gets deleted, in which case the added complexity for dependent resource needn't be handled explicitly since we already have a generic way to do per-resource no-op (we currently don't have a generic way to do dependent no-op).

cc @govindkrjoshi

kitallis: Since patient and its dependencies currently sync atomically anyway (patient + address + number + business identifier), it may make sense to do that.

harimohanraj89: I like the cascading deletes for a couple of reasons:
* It simplifies the code to achieve the same end-result, ie, attempts to edit info related to a deleted patient should no-op
* Soft-deletion of addresses, etc, seems like a logical forward-thinking move. Imagine a world where a patient can manage multiple addresses or phone numbers (like an Amazon.in account). In other words, controlling the explicit deletion of addresses, instead of inferring their deletion from their parent patient record, seems like a wise move.

kitallis: @harimohanraj89 Do you think we should add that as a proposition for later in the ADR and proceed with the code changes as in the PR?

We can punt on handling cascading deletes right now because it'd be a very rare scenario where that would be necessary and at the moment, there's only currently one instance of that possibility. (patient + dependents). We've added a logger to track how much that would be happening and we can take a call on cascading deletes (or dependent no-ops).

What do you think?

harimohanraj89: Yup I think that's totally fine. Let's make sure to craft a calendar invite or scheduled story or something to remind us to check back on this instrumentation in, say, 30-90 days.

harimohanraj89:
:shipit:

PR #765, opened 2020-01-21
ADR: Robust data migrations

harimohanraj89:
Story: https://www.pivotaltracker.com/story/show/170814447

This PR adds an ADR for robust data migrations using background jobs, after the events of [Jan 20, 2020](https://simpledotorg.slack.com/archives/CFHMC60P5/p1579524825012500).

The numbering anticipates #749 is merged first. If not, I'll update the numbering before merging, NBD.


doc/arch/008-robust-data-migrations.md
```diff
@@ -0,0 +1,93 @@
...
+* We avoid the use of very-long-running processes on the server (which could also result in very-long-running database
+  connections)
+* It becomes easier to monitor the progress of the migration through the Sidekiq dashboard
+* The execution of the data migration can be easily throttled, paused, or halted
+
+We will need to take care of our Sidekiq queues. [See below](#consequences).
+
+## Status
+
+Accepted
```
harimohanraj89: I guess we keep this "Accepted" and use PR review instead to actually review it? 😅

harimohanraj89:
@govindkrjoshi I'll let you pull in a second reviewer if you'd like.

doc/arch/008-robust-data-migrations.md
kitallis: Pretty much, yeah.

timcheadle:
To be honest, putting data migrations into `db/migrate` still makes me itch.

The classic problem is when you have to roll back, and the migration rollback references code that may have been deleted. This is often the case when you delete or rename a model.

I've never used it before, but it looks like the [data_migrate](https://github.com/ilyakatz/data-migrate) gem may be useful here?

https://www.ombulabs.com/blog/rails/data-migrations/three-useful-data-migrations-patterns-in-rails.html

harimohanraj89:
`data_migrate` does look promising. The key advantage I'm seeing is that we can run _only_ DB migrations forwards or backwards if we need to, avoiding the issue of data migrations referencing deleted code.

However, I still think there's a good case to be made that if a data migration uses a one-off service class, the service class should be kept around in the codebase, just like the data migration is. I'll update accordingly.

PR #770, opened 2020-01-23
My facilities bp control

vkrmis:
Adds the MyFacilities > BP Control Dashboard

This PR adds three new materialized views for computing the latest blood pressures per patient. The `BloodPressureControlQuery` class implements query methods to calculate controlled bps, uncontrolled bps and all time control for facilities.

We might change some semantics around the period selection based on feedback in the definitions document.

Definitions: https://docs.google.com/document/d/1Mo83rTwR_tuuUNnJnWKWlZ49nTzlLxqlrbKYnDj1kLQ/edit#

https://www.pivotaltracker.com/story/show/170179305
https://www.figma.com/file/kONxBLSXwcQDnV7ya4hIXA/Simple-Exploration?node-id=2915%3A57


db/views/latest_blood_pressures_per_patient_per_months_v01.sql
```diff
@@ -0,0 +1,15 @@
+SELECT DISTINCT ON (patient_id, year, month)
```
harimohanraj89: This `DISTINCT` assures that a patient only shows up once in each month, right? If we didn't use `DISTINCT`, I believe a patient would show up multiple times in a month if they had multiple BP recordings in the month. Now that we're using `DISTINCT`, which reading wins? Do we have control over that? Does the most recent reading win because of the ordering clause?

db/views/latest_blood_pressures_per_patient_per_months_v01.sql
```diff
@@ -0,0 +1,15 @@
...
+    blood_pressures.id AS bp_id,
+    patient_id,
```
harimohanraj89: For ease of reading through this file, can we include the appropriate table prefixes on _all_ fields, instead of just the ambiguous ones?

lib/tasks/refresh_materialized_db_views.rake
```diff
@@ -0,0 +1,16 @@
...
+  # LatestBloodPressuresPerPatientPerQuarter and LatestBloodPressuresPerPatient
+
+  Rails.logger.info 'Refreshing LatestBloodPressuresPerPatientPerMonth'
+  LatestBloodPressuresPerPatientPerMonth.refresh
+
+  Rails.logger.info 'Refreshing LatestBloodPressuresPerPatient'
+  LatestBloodPressuresPerPatient.refresh
+
+  Rails.logger.info 'Refreshing LatestBloodPressuresPerPatientPerQuarter'
+  LatestBloodPressuresPerPatientPerQuarter.refresh
```
harimohanraj89: From a dev's UX perspective, it's nice to have a `Rails.logger.info "Refresh complete"` at the very end. That way, when the task takes a long time, I don't have to wonder whether the last step takes a while, or whether the script hung or failed to terminate correctly after execution.

db/views/latest_blood_pressures_per_patient_per_quarters_v01.sql
```diff
@@ -0,0 +1,3 @@
+SELECT DISTINCT ON (patient_id, year, quarter) *
+FROM latest_blood_pressures_per_patient_per_months
+ORDER BY patient_id, year, quarter, bp_recorded_at DESC, bp_id;
```
harimohanraj89: Is it worth introducing the complexity of a view derived from another view? Does the performance gain justify this? How terrible would it be for us to craft this view the same way we do the per-month table above, and simply `DISTINCT` on quarter instead? That would avoid the need to be careful about ordering when refreshing these views, and reduce the coupling between them.

As a sanity check, do we _need_ a per-quarter view, if we're already adding a `quarter` column to the per-month view?

package.json
```diff
@@ -2,6 +2,6 @@
   "name": "simple-server",
   "private": true,
   "dependencies": {
-    "tablesort": "^5.1.0"
+    "tablesort": "^5.2.0"
```
harimohanraj89: 😎

spec/helpers/dashboard_helper_spec.rb
```diff
@@ -1,4 +1,4 @@
-require 'rails_helper'
+  require 'rails_helper'
```
harimohanraj89: 👾

config/schedule.rb
```diff
@@ -24,3 +24,7 @@ def local(time)
 every :month, at: local('02:00 am'), roles: [:seed] do
   rake 'generate:seed[1]'
 end
+
+every :day, at: local('03:00 am'), roles: [:cron] do
+  rake 'refresh_materialized_db_views'
```
harimohanraj89: Do we need to refresh the tables to get updated schemas, or to even get updated rows?

app/queries/my_facilities/overview_query.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def inactive_facilities
+    facilities = @facilities.left_outer_joins(:blood_pressures)
+                            .where('blood_pressures.recorded_at IS NULL OR blood_pressures.recorded_at > ?',
+                                   INACTIVITY_THRESHOLD_PERIOD)
+                            .having('COUNT(blood_pressures) < ? ', INACTIVITY_THRESHOLD_BPS)
+                            .group('facilities.id')
+
+    Facility.where(id: facilities.pluck(:id))
```
harimohanraj89: Can a view make this query easier? We don't have to address in this PR, but if so, let's add a story to tackle it later.

app/queries/my_facilities/blood_pressure_control_query.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  def all_time_bps
+    LatestBloodPressuresPerPatient
+      .where(bp_facility_id: @facilities)
+  end
+
+  def all_time_controlled_bps
+    all_time_bps
+      .where('bp_recorded_at > ?', Date.current - 90.days)
+      .where('systolic < 140 AND diastolic < 90')
```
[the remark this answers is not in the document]
vkrmis: We were waiting for #759 to be merged. All of this can now be replaced with a scope in the scenic model and use the `BloodPressure` threshold constants. 😎

app/queries/my_facilities/blood_pressure_control_query.rb
```diff
@@ -0,0 +1,105 @@
...
+    cohort_registrations = monthly_registrations
+    LatestBloodPressuresPerPatientPerMonth
+      .select("distinct on (patient_id)
+       bp_id, patient_id, bp_facility_id, bp_recorded_at, deleted_at, systolic, diastolic, quarter, year")
+      .order('patient_id, bp_recorded_at DESC, bp_id')
+      .where(patient: cohort_registrations)
+      .where('(year = ? AND month = ?) OR (year = ? AND month = ?)',
+             @year.to_s, @month.to_s,
+             *(previous_year_and_month(@year, @month).map(&:to_s)))
+  end
```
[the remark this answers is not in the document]
vkrmis: The cohort definition for months is inherently more complex: 
- Quarterly Cohorts look at registrations in one quarter, and the visits in the next quarter. 
- Monthly cohorts look at registrations in a month and visits in the next **two** months.

db/views/latest_blood_pressures_per_patient_per_months_v01.sql
vkrmis: Yes, the ordering + distinct ensures we get the last BP for each month

config/schedule.rb
vkrmis: We need to refresh the views to get updated rows. Since these are materialized, they only update when refreshed.

app/queries/my_facilities/overview_query.rb
vkrmis: We could create another view for computing `facility_activity_per_day`, but we'll do this as a separate PR. Adding a story for this.

app/queries/my_facilities/blood_pressure_control_query.rb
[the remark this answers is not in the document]
harimohanraj89: Ugh, beautiful 😎

app/queries/my_facilities/overview_query.rb
harimohanraj89: Neat, thanks. Go ahead and add a link to that story in this comment thread or PR description.

db/views/latest_blood_pressures_per_patient_per_quarters_v01.sql
vkrmis: We do need the per quarter view, primarily to get the latest bp per quarter.

As for the derived views, we'll need to benchmark the view refresh against prod sized data, before we can conclusively decide. We'd also be extracting the month, quarter and year again for each view.

Maybe we can test this at some point and make the views independent then.

harimohanraj89: Can we test with a large enough dataset _before_ we ship this? Is it hard to backtrack if we ship this as is and refresh the views in the wrong order?

app/queries/my_facilities/overview_query.rb
vkrmis: https://www.pivotaltracker.com/story/show/170919252

app/queries/my_facilities/blood_pressure_control_query.rb
```diff
@@ -0,0 +1,107 @@
...
+
+  def quarterly_bps
+    visited_in_quarter = next_year_and_quarter(@registration_year, @registration_quarter)
+    @quarterly_bps ||= LatestBloodPressuresPerPatientPerQuarter
+                       .where(patient: quarterly_registrations)
+                       .where(year: visited_in_quarter.first, quarter: visited_in_quarter.second)
+  end
+
+  def quarterly_controlled_bps
+    quarterly_bps.where('systolic < 140 AND diastolic < 90')
```
harimohanraj89: These derived methods can also be cached.

db/views/latest_blood_pressures_per_patient_per_quarters_v01.sql
harimohanraj89: I thought about this a little more. I think the regular use of a the carefully crafted rake task (with associated comment) should cover us well. Let's punt on any performance testing for now.

app/models/latest_blood_pressures_per_patient.rb
```diff
@@ -2,4 +2,18 @@ class LatestBloodPressuresPerPatient < ApplicationRecord
   def self.refresh
     Scenic.database.refresh_materialized_view(table_name, concurrently: false, cascade: false)
   end
+
+  THRESHOLDS = BloodPressure::THRESHOLDS
```
harimohanraj89: Do you think it's worth extracting these "BloodPressureable" into a module?

harimohanraj89:
One unresolved comment, one new comment, and one failing spec. This is looking great, can't wait to polish it up.

harimohanraj89:
One high-level question. Due to the nature of the query object interfaces, we do a lot of data assembly in the controller. How can we add test coverage for this behavior?

app/queries/my_facilities/blood_pressure_control_query.rb
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+  def quarterly_uncontrolled_bps
+    quarterly_bps.where('systolic >= 140 OR diastolic >= 90')
+  end
+
+  def monthly_registrations
+    patients = Patient.where(registration_facility: @facilities)
```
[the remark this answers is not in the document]
vkrmis: Not sure if this helps much, but why would we do this for `facilities` and not for the other variables we're instantiating?

app/views/layouts/my_facilities.html.erb
```diff
@@ -24,14 +24,14 @@
     </div>
     <% unless request.path == my_facilities_overview_path %>
       <div class="mt-0 mb-4" style="text-align: center; display: block; border-radius: 3px; background: rgba(0,0,0,0.02); padding: 10px 16px;">
-        <a data-toggle="modal" data-target="#facility-filter-modal" href="#" style="display: block; text-transform: uppercase; letter-spacing: 0.3em; font-size: 12px; font-weight: bold;">
+        <a data-toggle="modal" data-target="#facility-filter-modal" style="display: block; text-transform: uppercase; letter-spacing: 0.3em; font-size: 12px; font-weight: bold; cursor: pointer">
```
harimohanraj89: Let's pull these styles out into CSS.

app/queries/my_facilities/blood_pressure_control_query.rb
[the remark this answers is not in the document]
harimohanraj89: Because we access this one in several places, it's idiomatic Ruby practice to access through a method. The other vars being instantiated are also never accessed directly, but only through the query methods in this class 😉 

Not a huge deal, I'll let you address/resolve as you see fit 👍

vkrmis:
> One high-level question. Due to the nature of the query object interfaces, we do a lot of data assembly in the controller. How can we add test coverage for this behavior?

Since we're only aggregating data from the queries, which themselves have tests, I'm not sure this is strictly necessary. The only way I can think of to test this is to mock up some facilities and test the `.group` and `.count` behaviour. Doesn't seem very useful.

PR #771, opened 2020-01-23
Add facility help tips

dburka:
This adds a few helpful tips for adding Facility Groups and Facilities. Also has clearer labels for form fields during adding both Facility Groups and Facilities. And, Tim and I moved facility selection out of the Facility Group form and moved it into the Edit Facility form.

We also removed "description" from facility groups as they are not necessary.


app/views/admin/facilities/index.html.erb
```diff
@@ -1,13 +1,21 @@
 <div class="page-header">
   <h1 class="page-title">All facilities</h1>
+
   <nav class="page-nav">
     <% if policy([:manage, :facility, Facility]).upload? %>
       <%= link_to '+ Upload facilities CSV', upload_admin_facilities_path, class: "btn btn-sm btn-default" %>
     <% end %>
     <% if policy([:manage, FacilityGroup]).new? %>
-      <%= link_to 'New facility group', new_admin_facility_group_path, class: "btn btn-sm btn-primary" %>
+      <%= link_to '<i class="fas fa-plus mr-1"></i> New facility group' .html_safe, new_admin_facility_group_path, class: "btn btn-sm btn-primary" %>
```
harimohanraj89: Let's avoid the use of `html_safe` with:
```
<%= link_to new_admin_facility_group_path, class: "btn btn-sm btn-primary" do %>
  <i class="fas fa-plus mr-1"></i>
  New facility group
<% end %>
```

app/views/admin/facilities/index.html.erb
```diff
@@ -22,10 +30,10 @@
         </div>
         <div class="col-6 text-right">
           <% if policy([:manage, facility_group]).edit? %>
-            <%= link_to 'Edit', edit_admin_facility_group_path(facility_group), class: "btn btn-sm btn-outline-primary" %>
+            <%= link_to 'Edit group', edit_admin_facility_group_path(facility_group), class: "btn btn-sm btn-outline-primary" %>
           <% end %>
           <% if policy([:manage, :facility, facility_group.facilities.new]).new? %>
-            <%= link_to "New Facility", new_admin_facility_group_facility_path(facility_group), class: "btn btn-sm btn-outline-success" %>
+            <%= link_to '<i class="fas fa-plus mr-1"></i> Add a facility' .html_safe, new_admin_facility_group_facility_path(facility_group), class: "btn btn-sm btn-primary" %>
```
harimohanraj89: Let's avoid the use of `html_safe` with:
```
<%= link_to new_admin_facility_group_facility_path(facility_group), class: "btn btn-sm btn-primary" do %>
  <i class="fas fa-plus mr-1"></i>
  Add a facility
<% end %>
```

app/controllers/admin/facilities_controller.rb
```diff
@@ -70,6 +70,7 @@ def set_facility_group
 
   def facility_params
     params.require(:facility).permit(
+      :facility_group_id,
```
harimohanraj89: Users can modify their form to send any facility group ID. Are we checking that the facility group ID received here is allowed for the current user?

timcheadle: Good catch. I'll fix this. We limit the group ID on the form, but you're right that it can be modified.

harimohanraj89:
Updated facility group authorization in e3c58e6

Now anyone tampering with the form is presented with this helpful tip.

Screenshot 2020-02-11 at 4 55 02 PM

harimohanraj89:
Closing this for now to maintain focus while we learn more about permissioning. Just the tips have been decoupled to #822 and can be shipped indepenedently.

PR #772, opened 2020-01-28
Bangladesh import

harimohanraj89:
Story: https://www.pivotaltracker.com/story/show/168945561

### Overview

This PR introduces scripting to import Bangladesh patient starter data provided by the NHF into the Bangladesh Simple instance.

### Parts

* [x] Scripts to convert XLSX format to more slurpable CSV
  * [x] `to_csv.rb`
* [x] Scripts to import encrypted slurpable CSV into database
  * [x] `to_patients.rb`
    * [x] Parse patient info from CSV
    * [x] Persist to database
    * [x] Read encrypted CSV from a remote URL
* [x] Import user needs to show up in the facility analytics dashboard
* [x] Why are Golapganj Q1 2020 follow-ups missing?
* [x] What do we do about future BP dates?


timcheadle:
I don't think we need any of the Excel to CSV scripts; you can save directly to CSV from Excel itself. Doing so requires converting the file to unix file format, which is easiest with `dos2unix` (available on homebrew).

We also need to think about how to import the medications; I don't see those fields matched yet (I know it's a work in progress).

I'm not sure I know what `to_database.rb` is meant to do... Won't we create the records directly in the parsing script? Might be easier than trying to reparse another format.

Lastly, I don't want to put this data in our repo at all. Instead we can throw it on S3; the script can use `open-uri` to read from a URL.

harimohanraj89:
Excel doesn't support more than 255 columns in a sheet, and this one has 662. Hence I've added the script to convert to CSV.

You are right, I need to tackle medications today. It will be the most tricky, I think.

We can definitely write to DB in the parsing script. The intermediate file that's generated and re-parsed was mainly to give me a good idea of what the data looked like in a more readable fashion. We can basically just tack the two together and avoid writing to and reading from a temp file.

For now, the script reads from a local file to speed up development. I'll update to read an encrypted file from S3 and decrypt using some configured key.

harimohanraj89:
@timcheadle  I’ve pushed more work to the import script:
* Eliminated an intermediate step and the `to_database` script. `to_patients` will be the only one we need to run on the Bangladesh server. 
* Captured medicines prescribed at the most recent visit

Last steps are 
* Persisting all this parsed info to the DB
* Reading an encrypted CSV from a URI

I'm going to pair with a server dev tomorrow to translate this hash of info into valid models.

harimohanraj89:
@timcheadle This PR now does the following:
* Imports all patient data including medications
* Reads an encrypted csv from a remote URL (eg. S3)

I've got the encrypted CSV ready to go. Let me know if I can upload to S3 and test the import on my local machine.

lib/bangladesh/to_patients.rb
```diff
@@ -0,0 +1,268 @@
...
+
+$registration_facility = Facility.find_by(name: 'UHC Golapganj')
+$registration_user = User.create_with(
+  sync_approval_status: "denied",
+  sync_approval_status_reason: "User is a robot",
+  device_created_at: DateTime.current,
+  device_updated_at: DateTime.current
+).find_or_create_by!(full_name: 'bangladesh-import-user')
+
+def nid(value)
```
kitallis: Should we just call this `national_id`?

PR #773, opened 2020-01-28
CV History Update - Diabetes Management

harimohanraj89:
Story: https://www.pivotaltracker.com/story/show/169808996

### Overview

This PR adds support for the new "Diagnosis" section of the app that is presented in clinics that treat oth Hypertension (HTN) and Diabetes (DM). Check out [the PRD](https://docs.google.com/document/d/1YA8juemDvIqI2J_8-JIB-xNlwZUxiHT7AP726Z-7qs0/edit#heading=h.q8zm89ru7j26) for full context of decisions made in this PR.

### Strategy

Here are the key points on the approach.

* Diabetes answers contain valid data, and will be preserved across this change
* Current hypertension answers are not useful, so we introduce a new `hypertension` field on medical histories.
* This field will be force-set to `yes` for all current patients, and for all HTN-only patients moving forward
* This field will capture the "Hypertension" diagnosis question in the new app for HTN+DM patients
* The old hypertension question will match the answers of the new question moving forward.

### The Server's Role

In this strategy, the server will be responsible for:

* Defaulting `hypertension` and `diagnosed_with_hypertension` to `yes` for all users of the current app
* One-time setting `hypertension: yes, diagnosed_with_hypertension: yes` for all existing patients


kitallis:
@hari, I did a sweep of the PR, I just have some high-level questions (everything else will be inline):

* As I understand it, `receiving_treatment_for_hypertension` is a discontinued field from the app for the new questions and is a required field in the `current` API, is our strategy changed in this regard?
* I missed fully grasping the what `data_migrate` entails, and just to ensure, does this force run the data migration *during* deploy?
  * Can we test the migration for approximately 250,000 medical histories?

swagger/v2/swagger.json
```diff
@@ -2089,6 +2089,13 @@
             "unknown"
           ]
         },
+        "hypertension": {
```
kitallis: `v2` should not have this field in the API. This is just an artifact of how swagger functions, not how you've implemented it.

It needs to be elided from the v2 [payload](https://github.com/simpledotorg/simple-server/blob/master/spec/factories/medical_histories.rb#L57) in the factory which is used in this [spec](https://github.com/simpledotorg/simple-server/blob/master/spec/api/v2/medical_histories_spec.rb#L19) (that which auto-generates the swagger schema)

app/transformers/api/current/medical_history_transformer.rb
```diff
@@ -8,7 +8,12 @@ def to_response(medical_history)
     end
 
     def from_request(medical_history_payload)
-      Api::Current::Transformer.from_request(medical_history_payload)
+      Api::Current::Transformer.from_request(medical_history_payload).tap do |params|
+        if params[:hypertension].blank?
```
kitallis: This is fine, but keep in mind that `v2` ([which inherits from Current](https://github.com/simpledotorg/simple-server/blob/master/app/transformers/api/v2/medical_history_transformer.rb)) will also behave this way. Since this is an accretion `v3` already has versions of itself that don't handle `hypertension`, so this check is valuable and good.

However, we should check with the mobile team, in the `to_response` [case](https://github.com/simpledotorg/simple-server/blob/master/app/transformers/api/v2/appointment_transformer.rb#L4), whether we need to do an `except` for the `hypertension` field for `v2` or will the old API continue to work correctly (as a sanity check).

harimohanraj89: Ooh I did not know that V2 inherits from Current. This is good to know. I'll check with the mobile team and update the necessary transformer accordingly. 👍

swagger/v2/swagger.json
harimohanraj89: Interesting. I will go exclude this field from that payload for now.

Are we concerned that updates to the current API have these trickledown effects to previous versions like v2? I'd assume that v2 should remain fixed in the face of changes to the current API.

harimohanraj89:
* The mobile will send `receiving_treatment_for_hypertension: unknown` after the update. I am confirming this with Vinay.
* Regarding `data_migrate` - Data migrations are executed with the command `rails db:migrate:with_data`. Do we auto-run regular migrations through scripts? If so, we will have to update the script's migration command to include the `:with_data` suffix. As for order of operations, I think we first merge this PR, then make the change to deployment scripts, then deploy this PR using the updated deployment scripts. Adding the suffix right away will block current deployments, since `:with_data` won't be supported yet.
* Thanks for the number (I guess it makes sense that it's close to the number of patients 😅). Could I do a test with 250000 records locally? Can we assume that our AWS infra is more powerful? Or would you prefer we do a test in a prod-like environment (and how should I get my hands on one of those)

swagger/v2/swagger.json
harimohanraj89: I _think_ the Swagger definitions are being updated because `Api::V2::Models` inherit from `Api::Current::Models`. I think the factory usage you linked to can remain unchanged, but I'll need to delete some schema keys in the V2 models [like we do for appointments](https://github.com/simpledotorg/simple-server/blob/master/app/schema/api/v2/models.rb#L30-L35).

harimohanraj89:
@kitallis 7 seconds on my local machine for 250,000 records - locking 1000 rows at a time.

```
== Data =======================================================================
== 20200129084509 SetHypertensionToTrue: migrating ============================
== 20200129084509 SetHypertensionToTrue: migrated (7.1023s) ===================
```

Does that sounds reasonable? We can tune down the time by utilizing bigger batches.

kitallis:
@harimohanraj89 

>     * The mobile will send `receiving_treatment_for_hypertension: unknown` after the update. [Confirmed with Vinay here](https://simpledotorg.slack.com/archives/CLS8YD22H/p1580362080041400).
> 

👍 

>     * Regarding `data_migrate` - Data migrations are executed with the command `rails db:migrate:with_data`. Do we auto-run regular migrations through scripts? If so, we will have to update the script's migration command to include the `:with_data` suffix. As for order of operations, I think we first merge this PR, then make the change to deployment scripts, then deploy this PR using the updated deployment scripts. Adding the suffix right away will block current deployments, since `:with_data` won't be supported yet.
> 

The current capistrano deploy only does `db:migrate` at the moment, we'll have to modify it to do `with_data` and keep that as the default operation.

>     * Thanks for the number (I guess it makes sense that it's close to the number of patients 😅). Could I do a test with 250000 records locally? Can we assume that our AWS infra is more powerful? Or would you prefer we do a test in a prod-like environment (and how should I get my hands on one of those)

I think a local test is sufficiently realistic, and 7 seconds is perfectly reasonable. Our deploys currently are parallel and not rolling (it's hard to do rolling deploys with cap) – so I'm a bit nervous about this on that front. But we have a hack to do rolling deploys currently where we'd just keep one box in the deploy script and comment out the rest.

Alternatively, I believe generating 250,000 records on staging should not be that hard because MH are pretty invisible to the dashboard and we can just create Nearly Garbage™ data in the absence of a proper data generation script.

kitallis:
> @kitallis 7 seconds on my local machine for 250,000 records - locking 1000 rows at a time.
> 
> ```
> == Data =======================================================================
> == 20200129084509 SetHypertensionToTrue: migrating ============================
> == 20200129084509 SetHypertensionToTrue: migrated (7.1023s) ===================
> ```
> 
> Does that sounds reasonable? ~We can tune down the time by utilizing bigger batches.~
> 
> Increasing batch size did not bring down the migration time significantly (batches of 10,000 -> 6sec), indicating that the bottleneck is the actual database operations, and not the overhead of each query. Given that the time to run is fairly constant, I'm in favor of smaller batch sizes to ensure we're not locking too many patient records at once. Would you agree?

Sounds perfect 👍

swagger/v2/swagger.json
kitallis: Ah, yep, you figured it out. That's the correct way. Apologies I missed that part in my comment.

spec/transformers/api/current/medical_history_transformer_spec.rb
```diff
@@ -3,17 +3,50 @@
...
       expect(transformed_medical_history).not_to include('user_id')
     end
   end
+
+  describe 'from_request' do
+    let(:transformed_params) { described_class.from_request(medical_history_params) }
+    context "when hypertension is provided" do
+      let(:medical_history_params) do
+        {
+          id: 'uuuuuuuu-uuuu-uuuu-iiii-ddddddddddd1',
```
kitallis:
```suggestion
          id: SecureRandom.uuid,
```

spec/transformers/api/current/medical_history_transformer_spec.rb
```diff
@@ -3,17 +3,50 @@
...
+          patient_id: 'uuuuuuuu-uuuu-uuuu-iiii-ddddddddddd2',
```
kitallis:
```suggestion
          patient_id: SecureRandom.uuid,
```

spec/transformers/api/current/medical_history_transformer_spec.rb
```diff
@@ -3,17 +3,50 @@
...
+      it 'passes along the received hypertension values' do
+        expect(transformed_params[:hypertension]).to eq('no')
+        expect(transformed_params[:diagnosed_with_hypertension]).to eq('unknown')
+      end
+    end
+
+    context "when hypertension is not provided" do
+      let(:medical_history_params) do
+        {
+          id: 'uuuuuuuu-uuuu-uuuu-iiii-ddddddddddd1',
```
kitallis:
```suggestion
          id: SecureRandom.uuid,
```

spec/transformers/api/current/medical_history_transformer_spec.rb
```diff
@@ -3,17 +3,50 @@
...
+          patient_id: 'uuuuuuuu-uuuu-uuuu-iiii-ddddddddddd2',
```
kitallis:
```suggestion
          patient_id: SecureRandom.uuid,
```

harimohanraj89:
@kitallis test UUIDs are more realistic now 👍

PR #775, opened 2020-01-28
Create the Resources section.

dburka:
I added a new section to the admin with useful resources for a CVHO to answer user questions and to access training videos. This can be used by ANY user of Simple and does not need access controls.


app/controllers/resources_controller.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class ResourcesController < AdminController
+  skip_after_action :verify_authorized, only: [:index]
+  skip_after_action :verify_policy_scoped, only: [:index]
```
harimohanraj89: I have deliberately skipped these actions instead of adding a Resources policy which would always return true. The page is still protected behind authentication. If the rules to access resources grow more complex, we can add a policy.

harimohanraj89:
I've cleaned up the controller and extracted the CSS. This looks good IMO, but I'm no longer an objective reviewer 😅

app/views/resources/index.html.erb
```diff
@@ -0,0 +1,179 @@
...
+      <div class="row">
+
+          <div class="col-sm-12 col-lg-7">
+              <div class="card">
+                <h2>Common user questions</h2>
+
+                <div class="question"><h6 class="question">User can't find a patient</h6>
+                    <ol type="a">
+                        <li>Try searching with a different spelling, e.g <span class="badge badge-info">Aakash</span> instead of <span class="badge badge-info">Akash</span></li>
+                        <li>Make sure to match patient details - <span class="badge badge-info">full name</span> <span class="badge badge-info">Age</span> <span class="badge badge-info">Address</span> <span class="badge badge-info">Last visit date</span> to check if patient is already registered.</li>
```
vkrmis:
```suggestion
                        <li>Make sure to match patient details - <span class="badge badge-info">Full name</span> <span class="badge badge-info">Age</span> <span class="badge badge-info">Address</span> <span class="badge badge-info">Last visit date</span> to check if patient is already registered.</li>
```

PR #776, opened 2020-01-29
Add My Facilities - New Registrations dashboard

vkrmis:
Adds the My Facilities > Registrations page:

This adds a `PatientRegistrationsPerDayPerFacility` materialised view and a `RegistrationsQuery` class that implements the queries for the new page.

Registrations definitions: https://docs.google.com/document/d/1Mo83rTwR_tuuUNnJnWKWlZ49nTzlLxqlrbKYnDj1kLQ/edit#heading=h.kxmq12han0io

Figma: https://www.figma.com/file/vmZlmlFSCOrEYzTMkqsNB8/Simple(Production)?node-id=3684%3A0

-- 
Note to self: The PR is based off `my-facilities-bp-control` for review-ability, change base to master before merge.


spec/models/patient_registrations_per_day_per_facility_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe PatientRegistrationsPerDayPerFacility, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
harimohanraj89: Let's make sure we test at least the `belongs_to :facility` association here.

spec/queries/my_facilities/registrations_query_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        end
+
+        before do
+          ActiveRecord::Base.transaction do
+            ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{ENV['ANALYTICS_TIME_ZONE']}'")
+            PatientRegistrationsPerDayPerFacility.refresh
+          end
+        end
+
+        specify { pp LatestBloodPressuresPerPatient.all
```
harimohanraj89: Oh hi there. Let's remove this `pp` :smile:

spec/queries/my_facilities/registrations_query_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        end
+
+        before do
+          ActiveRecord::Base.transaction do
+            ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{ENV['ANALYTICS_TIME_ZONE']}'")
+            PatientRegistrationsPerDayPerFacility.refresh
+          end
+        end
+
+        specify { expect(registrations_query.registrations.count).to eq(2) }
```
harimohanraj89: Could we introduce some semantic arrays of patients or facilities, so we can write things like `expect(this_count_thing).to eq(included_patients.count)`? That makes it a whole lot easier for future us to understand which of the test patients are _supposed_ to be included, and which aren't.

spec/queries/my_facilities/registrations_query_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        end
+
+        specify { pp LatestBloodPressuresPerPatient.all
+          expect(registrations_query.registrations.count).to eq(2) }
+      end
+
+      context 'monthly' do
+        let!(:registrations_query) { described_class.new(period: :month, include_months: 3) }
+
+        let!(:patient_registration_timestamps) { [1.month.ago, 2.months.ago, 1.year.ago] }
```
harimohanraj89: `12.months` for consistency? 😅

db/views/patient_registrations_per_day_per_facilities_v01.sql
```diff
@@ -0,0 +1,11 @@
+SELECT COUNT(patients.id) AS registration_count,
+       patients.registration_facility_id AS facility_id,
+       patients.deleted_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE')) AS deleted_at,
```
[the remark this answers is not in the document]
prabhanshuguptagit: Thanks for catching, that's quite an oversight. Including `patients.deleted_at` is not only meaningless, it also means the aggregation is broken. Fixing this to have the `facilities.deleted_at` instead.

kitallis:
@prabhanshuguptagit @vkrmis Mind merging master into this PR?

db/views/patient_registrations_per_day_per_facilities_v01.sql
```diff
@@ -1,11 +1,11 @@
...
+       facilities.deleted_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE')) AS deleted_at,
 cast(EXTRACT(DOY FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS day,
 cast(EXTRACT(MONTH FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS month,
 cast(EXTRACT(QUARTER FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS quarter,
 cast(EXTRACT(YEAR FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS year
 FROM patients
+INNER JOIN facilities ON patients.registration_facility_id = facilities.id
 WHERE patients.deleted_at IS NULL
-GROUP BY day, month, quarter, year, facility_id, patients.deleted_at, patients.recorded_at;
+GROUP BY day, month, quarter, year, facility_id, facilities.deleted_at;
```
harimohanraj89: I don't think we need to group by `facilities.deleted_at`

app/controllers/concerns/cohort_period_selection.rb
```diff
@@ -28,7 +23,7 @@ def sanitize_registration_quarter(quarter_param)
 
     def sanitize_registration_year(year)
       return year.to_i if year.to_i.positive?
-      return (Time.current.beginning_of_month - 1.month).year if @period == :month
+      return (Time.current.beginning_of_month - 1.month).year if @selected_cohort_period == :month
```
harimohanraj89: I think there is still a bug here. @selected_cohort_period is the whole hash, not a single symbol like `:month`

app/transformers/api/current/facility_transformer.rb
```diff
@@ -2,9 +2,9 @@ class Api::Current::FacilityTransformer
   class << self
     def to_response(facility)
       facility.as_json
-        .except('enable_diabetes_management')
+        .except('enable_diabetes_management', 'monthly_opd_load')
```
harimohanraj89: I believe extra fields in the API response are gracefully ignored by the app. You can confirm with the mobile devs and remove this change if so.

harimohanraj89:
Another round of review complete.
* Critical - [this bug](https://github.com/simpledotorg/simple-server/pull/776#discussion_r374558269)
* Critical - Semaphore - looks like a syntax error in the SQL view?
* Non-critical - comments above
* Non-critical - CodeClimate - Feel free to address the easy ones :smile:

db/views/patient_registrations_per_day_per_facilities_v01.sql
prabhanshuguptagit: The `deleted_at` needs to be grouped by since it's part of the `SELECT` clause. We need the `deleted_at` for activerecord to apply the `kept` scope.

app/transformers/api/current/facility_transformer.rb
prabhanshuguptagit: The idea here is to not let users access this field, since this is admin-only and we should maybe exclude it in the API response.

app/views/shared/my_facilities/_registrations.html.erb
```diff
@@ -0,0 +1,111 @@
...
+        <% unless @selected_period == :day %>
+          <col>
+        <% end %>
+      <% end %>
+      <col class="table-divider">
+      <col class="mobile">
+    </colgroup>
+    <thead>
+    <tr>
+      <% case @selected_period %>
```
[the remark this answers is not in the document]
prabhanshuguptagit: `Registrations` is the only page with this block currently, although some pages to come soon will have this. We could maybe refactor this as the next page is added.

app/assets/stylesheets/application.scss
```diff
@@ -744,6 +744,7 @@ th sup {
     td.type-number { text-align: right; padding-right: 16px; }
     td.type-percent { text-align: center; }
     td.type-percent em { font-style: normal; background: #F1DBFF; color: #BE72E3; border-radius: 100px; padding: 2px 10px; }
+    .row-total td.type-percent em { background: rgba(0,0,0,0.1); color: rgba(0,0,0,0.6); }
```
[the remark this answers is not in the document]
prabhanshuguptagit: We seem to have a preference for multi-line CSS. These should definitely be multi-line, for consistency's sake.

app/controllers/concerns/period_selection.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
prabhanshuguptagit: CohortPeriod and Period are fundamentally different data. For example,
- The `CohortPeriod` period is the cohort’s duration. It can be a quarter/month.
- The `Period` is the query’s duration. It can be a quarter, month or day. 

This becomes more apparent when we switch pages between cohort reports and regular reports and want to preserve these filters.

Also, the concerns deal with defaulting each of them in a separate manner. It might be worth the overhead to keep them separate.

app/queries/my_facilities/registrations_query.rb
```diff
@@ -0,0 +1,59 @@
...
+      .where(facility: @facilities)
+      .where("(year, quarter) IN (#{year_quarter_tuples})")
+  end
+
+  def monthly_registrations
+    year_month_tuples = @periods.map { |(year, month)| "('#{year}', '#{month}')" }.join(',')
+
+    PatientRegistrationsPerDayPerFacility
+      .where(facility: @facilities)
+      .where("(year, month) IN (#{year_month_tuples})")
```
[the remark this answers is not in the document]
prabhanshuguptagit: The other (non-postgres, activerecord/arel) way to say this would be:
`SELECT * WHERE (year = year1 AND month = month1) OR (year = year2 AND month = month2)`

Couldn't figure out a way to generate this sanely with arel/activerecord, mostly because they aren't good at composing OR's. 

```ruby
# Activerecord
periods.reduce(PatientRegistrationsPerDayPerFacility) do
  |query, (year, period)| query.or(PatientRegistrationsPerDayPerFacility.where(year: year).where(@period => period))
end

# Arel
t = PatientRegistrationsPerDayPerFacility.arel_table
periods.reduce(PatientRegistrationsPerDayPerFacility) do |query, (year, a)| 
  query.or(table.where(t[:year].eq(year)).where(t[period].eq(a))) 
end
```
Both of these generate queries that return good data, but the queries themselves look quite horrid.

app/queries/my_facilities/registrations_query.rb
```diff
@@ -0,0 +1,59 @@
+class MyFacilities::RegistrationsQuery
+  include QuarterHelper
+  include MonthHelper
+  include DayHelper
+
+  attr_reader :periods
+
+  def initialize(period: :quarter, facilities: Facility.all, include_quarters: 3, include_months: 3, include_days: 7)
```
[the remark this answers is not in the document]
prabhanshuguptagit: Keeping a default around, so one can neatly get `all_time_registrations` without having to pass a period (which don't matter if we only want to look at all_time_registrations).
```ruby
MyFacilities::RegistrationsQuery.new.all_time_registrations
```

app/transformers/api/current/facility_transformer.rb
harimohanraj89: Herpderp, makes perfect sense.

app/queries/my_facilities/blood_pressure_control_query.rb
```diff
@@ -8,14 +8,15 @@ class MyFacilities::BloodPressureControlQuery
...
-                            previous_year_and_quarter(Time.current.year, quarter(Time.current)).second
-    @registration_month = cohort_period[:registration_month] ||
-                          (Time.current.beginning_of_month - 1.month).month
-    @registration_year = cohort_period[:registration_year] ||
-                         previous_year_and_quarter(Time.current.year, quarter(Time.current)).first
+  def initialize(facilities: Facility.all, cohort_period: {})
+    # cohort_period is map that contains
+    # - :cohort_period (:quarter/:month),
+    # - :registration_quarter/:registration_month
+    # - :registration_year
```
harimohanraj89: I feel this is evident from the code itself, is the comment necessary?

app/views/my_facilities/registrations.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="row">
+  <div class="col-lg-12 pr-lg-0">
+    <%= render "shared/my_facilities/registrations" %>
```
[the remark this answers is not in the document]
harimohanraj89: I had the same thought, but this approach separates the dashboard itself from the grid in which it's placed. You could argue that this'll allow for us to craft pages in the future that are combos of these pages. Eg. we could overhaul the overview page to contain registrations and bp control side-by-side. I'm not _too_ worried about the indirection, but my preference would be discarding the partial too.

app/queries/my_facilities/blood_pressure_control_query.rb
prabhanshuguptagit: It seems like a good to have, especially when someone needs to quickly reference usage (like when consoling in). Also only one of `registration_quarter`/` registration_month` need to be supplied in the hash, felt like a good place to add that information in.

PR #781, opened 2020-02-03
CV history lenience

harimohanraj89:
After [this Slack conversation](https://simpledotorg.slack.com/archives/CLS8YD22H/p1580709471072400), we decided that even though the answers to the new `hypertension` question must be `yes` or `no`, the API should continue to support `unknown` as a valid answer to ease the app update transition, and avoid messy race conditions between updating the app and triggering a resync.

This PR adds `unknown` as a valid option, and removes the `MEDICAL_HISTORY_DEFINITIVE_ANSWERS` constant, as it's no longer needed.

### DIFF
--- a/app/models/medical_history.rb
+++ b/app/models/medical_history.rb
@@ -22,11 +22,6 @@ class MedicalHistory < ApplicationRecord
     unknown: 'unknown'
   }.freeze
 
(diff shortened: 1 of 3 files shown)


kitallis:
I followed the conversation on Slack, this change makes sense 👍 .

PR #783, opened 2020-02-04
Default Country Bug

harimohanraj89:
Story: https://www.pivotaltracker.com/story/show/170972119

This PR addresses a bug where the edit facility screen would always pre-populate the "Country" field to be "India". The following changes are included:

* Only pre-populate a country value if none exists
* Pre-populate a country based on configuration, instead of always populating "India"
* Create a `Rails.application.config.country` hash that stores country-specific configuration in memory
* Consolidate all time-zone configurations into a single env var
* Add support for using a gitignored `.env.development.local` to override `.env.development` configurations locally.


.env.test
```diff
@@ -93,11 +93,7 @@ SIMPLE_SERVER_HOST_PROTOCOL='https'
 #
 # Country specific configuration
 #
-ANALYTICS_TIME_ZONE='Asia/Kolkata'
-CRON_TIME_ZONE='Asia/Kolkata'
 DEFAULT_COUNTRY=IN
```
govindkrjoshi: Shall we make this full country name to avoid ambiguity?

config/initializers/countries.rb
```diff
@@ -0,0 +1,16 @@
...
+  },
+  'BD': {
+    abbreviation: 'BD',
+    name: 'Bangladesh',
+    dashboard_locale: ENV['DEFAULT_PREFERRED_DASHBOARD_LOCALE'] || 'en_BD',
+    time_zone: ENV['DEFAULT_TIME_ZONE'] || 'Asia/Dhaka'
+  }
+}.with_indifferent_access
+
+Rails.application.config.country = country_config[ENV.fetch('DEFAULT_COUNTRY')]
```
govindkrjoshi:
```suggestion
Rails.application.config.default_country = country_config[ENV.fetch('DEFAULT_COUNTRY')]
```

config/initializers/countries.rb
```diff
@@ -0,0 +1,16 @@
+country_config = {
+  'IN': {
+    abbreviation: 'IN',
+    name: 'India',
+    dashboard_locale: ENV['DEFAULT_PREFERRED_DASHBOARD_LOCALE'] || 'en_IN',
```
govindkrjoshi: These values should be country specific constants, right?

.env.test
harimohanraj89: I'm ok leaving this as a country code. The code is still being used in some places too (specifically phone number parsing, I think).

config/initializers/countries.rb
harimohanraj89: Why do we use the `default` terminology? Can an instance of Simple switch between countries effectively? In other words, can a single instance of Simple show reports in IST _and_ also in Bangla time?

config/initializers/countries.rb
harimohanraj89: Yes. The second values in these statements are the country-specific ones. The env var just allows us to override the country-specific values _if_ we need to. I suspect that we won't need to do this any time soon.

PR #785, opened 2020-02-05
Dashboard CSV downloads

harimohanraj89:
Story: https://www.pivotaltracker.com/story/show/171077165

This PR adds a "Download" link to the Adherence Patients tab also. This link follows the same visibility rules as the overdue list download link, ie, it is not available for "All facilities". This PR implements this with the following bits:

* A new `download_adherence_follow_up_list` permission, which mimics the way `download_overdue_list` is implemented
* A new `index.csv.erb` template to handle generation of CSV, instead of model or controller logic.
* The existing overdue list download is also updated to use an ERB template
* The CSV logic in the Appointment model is removed

### Outstanding questions

* [ ] To which users should we automatically add this new permission? To all users with `download_overdue_list`, or to all users with `view_adherence_follow_up_list`?
* [ ] Do we like this ERB-based approach to generating these CSVs? I'm a big fan, since it removes this logic from the model, and the template is not very hard to grok
* [ ] What other pages did we want to add download links to? @timcheadle


harimohanraj89:
This PR is in the wrong direction with respect to both which page to download and permissioning. Will re-create another PR after updating the story.

PR #786, opened 2020-02-05
Generate sensible seed data

kitallis:
**Philosophy:**

* Use it in tandem with the `db:seed` script. This adds a level of repeatability in what is seen on SBX and what is generated on your local DB. This can be used in both places.
* `db:seed` has been improved to incorporate more recent schema changes.
* The generate rake task relies on making API calls to the `sync` APIs, this ensures all the internal checks and data creation is as close as possible to how the app would generate (trade-off: speed).
* Define "traits" of data in a map full of lambdas: like creating: `retroactive_bps`, `overdue_appointments`, `completed_phone_calls` etc.
* Add randomization to create a normal distribution.
* Add time-ranges so that the script can be run on an on-going basis.

**Tasks:**

* `db:seed`: generates Orgs → FacilityGroups → Facilities → Users. 
This also adds a default admin user.
* `generate:fake_data`: generates patient data and the rest for all users.
* `purge:users_data`: deletes patient data and the rest for all users. 

The `db:seed` script only really needs to be run once. After it has finished running, we can simply keep using a combination `purge:users_data` + `generate:fake_data` to keep refreshing.

**TODO**

- [x] Make all data for a given user a sidekiq job
- [x] Tweak the numbers to be more uniform across months
- [x] Make it a `whenever` job
- [x] Run on SBX
- [x] Make it so that it can never be run on production (cron role check)
- [x] Update README

**Cron**

There's a cron job that generates fake data periodically (currently configured to a week). It also takes care of purging all data from SBX before it regens new stuff. The purge task deletes only patient level data and the Users, Orgs, Facilities etc. are preserved.


app/jobs/populate_fake_data_job.rb
```diff
@@ -0,0 +1,160 @@
...
+    },
+    completed_phone_call: {
+      time_fn: -> { Faker::Time.between(6.months.ago, Date.today) },
+      size_fn: -> { rand(1..10) },
+      patient_sample_size: 0.5
+    }
+  }.freeze
+
+  def perform(user_id)
+    @user = User.find(user_id)
```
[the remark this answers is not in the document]
kitallis: `Rails.env.production?` won't work since all non-dev environments run in "production" mode. We've added a check against the env var `SIMPLE_SERVER_ENV` to be extra careful. As far as the job is concerned, it's stuck against the role `fake_data` which is only given to the sandbox boxes, so the cron job will never run this elsewhere.

app/jobs/populate_fake_data_job.rb
```diff
@@ -0,0 +1,160 @@
...
+    puts "#{path} failed with status: #{output.status}" unless output.status.ok?
+  end
+
+  def sample_patients(percentage)
+    user.registered_patients
+        .sample([percentage * user.registered_patients.size, 1].max)
+  end
+
+  def generate(trait, args)
+    (1..args[:size_fn].call).flat_map { send(trait, args.slice(:patient, :time_fn)) }
```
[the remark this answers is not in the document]
kitallis: That's not the same thing right? `times` doesn't return the generated array which is then passed over to make API calls with in `create_resources`.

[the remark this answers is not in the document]
harimohanraj89: Ah the return value is important. Yup, you're right 👍

spec/factories/patients.rb
```diff
@@ -26,11 +26,23 @@
...
+      medical_history {
+        build(:medical_history, :hypertension, patient_id: id, user: registration_user)
+      }
+    end
+
+    trait :diabetes do
+      medical_history {
+        build(:medical_history, :diabetes, patient_id: id, user: registration_user)
+      }
+    end
```
prabhanshuguptagit, line 44: Remove unused traits

app/jobs/populate_fake_data_job.rb
```diff
@@ -0,0 +1,277 @@
...
+    trait_args[:build_fn].call(trait_args)
+  end
+
+  def api_post(path, data)
+    output = HTTP
+               .auth("Bearer #{user.access_token}")
+               .headers(api_headers)
+               .post(api_url(path), json: data)
+
+    raise HTTP::Error unless output.status.ok?
```
kitallis, line 264: Log these errors better, maybe report to Sentry?

app/jobs/populate_fake_data_job.rb
```diff
@@ -0,0 +1,277 @@
...
+    puts "Creating #{trait_name} for #{user.full_name}..."
+
+    request_key = trait_args[:request_key]
+    api_version = trait_args[:api_version]
+    return if request_key.blank?
+
+    logger.info("Creating #{trait_name} for #{user.full_name}" +
+                  "with #{data.size} #{request_key} – facility: #{user.facility.name}")
+
+    data.each_slice(20) do |data_slice|
```
prabhanshuguptagit, line 237: Make 20 a constant.

app/jobs/populate_fake_data_job.rb
```diff
@@ -0,0 +1,160 @@
...
+  HOST = URI.parse("#{ENV['SIMPLE_SERVER_HOST_PROTOCOL']}://#{ENV['SIMPLE_SERVER_HOST']}").to_s
+  DEFAULT_HEADERS = { 'Content-Type' => 'application/json', 'ACCEPT' => 'application/json' }.freeze
+
+  TRAITS = {
+    newly_registered_patient: {
+      time_fn: -> { Faker::Time.between(1.month.ago, Time.now) },
+      size_fn: -> { rand(50..200) },
+      request_key: :patients
+    },
+    ongoing_bp: {
```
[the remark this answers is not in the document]
kitallis: Not really, but we should naturally get distributions like that along with distributions like, multiple BPs in a month, multiple BPs in a day and so on.

app/jobs/populate_fake_data_job.rb
```diff
@@ -0,0 +1,160 @@
...
+  sidekiq_throttle threshold: { limit: 20, period: 1.minute }
+
+  attr_reader :user
+
+  FAKE_DATA_USER_ROLE = 'Seeded'.freeze
+  HOST = URI.parse("#{ENV['SIMPLE_SERVER_HOST_PROTOCOL']}://#{ENV['SIMPLE_SERVER_HOST']}").to_s
+  DEFAULT_HEADERS = { 'Content-Type' => 'application/json', 'ACCEPT' => 'application/json' }.freeze
+
+  TRAITS = {
+    newly_registered_patient: {
```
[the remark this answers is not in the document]
kitallis: We could do that as an incremental improvement. It'll just be Yet Another Trait.

.env.development
```diff
@@ -93,8 +93,8 @@ SECRET_KEY_BASE=b90c991f1987f5be558af41e1fdc73741adbef2285f30a5bd500dcd7804b2ef1
 
 SIMPLE_APP_SIGNATURE=
 SIMPLE_SERVER_ENV=development
-SIMPLE_SERVER_HOST='localhost'
-SIMPLE_SERVER_HOST_PROTOCOL='https'
+SIMPLE_SERVER_HOST='localhost:3000'
```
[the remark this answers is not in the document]
kitallis: Closing this, since it's not relevant anymore.

kitallis:
> If the regular monthly cadence of BP measurements for a patient is important, then I have an idea to throw your way.
> 
> What if we modeled each patient as "regular" or "flaky"? For each patient, we start from their reg date, pick a random day in each month, and either create a BP measurement or not depending on a probability. Probability for regular patients may look like this
> 
> [1, 0.90, 0.89, 0.88, 0.87, 0.86]
> 
> Probability for flaky patients may look like this
> 
> [ 1, 0.5, 0.4, 0.3, 0.2, 0.1].
> 
> We may naturally get a good mix of patients that are regular vs overdue, and also new vs old if we just distribute registration date evenly across `[1.year.ago...1.day.ago]`.
> 
> Additionally, are we doing anything to generate realistic BP control trends? Or are we randomly creating high/low BPs for each patient? What if we had a "under_control" probability for each patient that looked like the arrays above?

These are all good suggestions. I think if we were going for a simulation of production data, then probabilistic generation would make sense. We're going for roughly acceptable data in this script to play around with on SBX and Dev.

We can always incrementally improve this script to add realistic trends via the Traits. Just make new traits with appropriate `time_fn`s that can potentially take probabilities to space out the generation.

timcheadle:
Be careful not to allow creating a default admin user in actual production.

app/jobs/populate_fake_data_job.rb
```diff
@@ -0,0 +1,285 @@
...
+
+    request_key = trait_args[:request_key]
+    api_version = trait_args[:api_version]
+    return if request_key.blank?
+
+    logger.info("Creating #{trait_name} for #{user.full_name} " +
+                  "with #{data.size} #{request_key} – facility: #{user.facility.name}")
+
+    data.each_slice(SYNC_PAYLOAD_SIZE) do |data_slice|
+      api_post("/api/#{api_version}/#{request_key}/sync", request_key => data_slice) if data_slice.present?
```
rsanheim, line 241: Can you say more about why we are going through our API to generate seed data? Are there service objects or other business constraints enforced in the API (I haven't really looked at our API yet, so I'm ignorant here =)) 

Our model layer _should_ have all the data consistency / validations enforced, so using the models directly would be fine for this I think. My preference for seed generation would be to keep it as simple as possible via FactoryBot calls or plain old `Model.create` calls, so we can change it and maintain it easily over time.

lib/tasks/generate.rake
```diff
@@ -0,0 +1,7 @@
+namespace :generate do
+  desc 'Generate fake data; Example: rake "generate:fake_data'
+  task fake_data: :environment do
```
rsanheim, line 3: Proposed name change: `db:seed_users`.

lib/tasks/purge.rake
```diff
@@ -0,0 +1,8 @@
+namespace :purge do
+  require 'tasks/scripts/purge_users_data'
+
+  desc 'Purge all user data; Example: rake "purge:users_data'
+  task users_data: :environment do
```
rsanheim, line 5: Proposed name change: `db:purge_users`

lib/tasks/generate.rake
```diff
@@ -0,0 +1,7 @@
+namespace :generate do
```
rsanheim, line 1: You can wrap this task definition in something like `unless Rails.env.production?` to ensure we could never run these tasks in a prod environment.

lib/tasks/purge.rake
```diff
@@ -0,0 +1,8 @@
+namespace :purge do
```
rsanheim, line 1: Same as above, we should failsafe to make sure it never runs where it shouldn't

lib/tasks/scripts/purge_users_data.rb
```diff
@@ -0,0 +1,16 @@
...
+    return if ENV['SIMPLE_SERVER_ENV'] == 'production'
+
+    models = [BloodPressure, BloodSugar, Appointment, CallLog, Communication, PrescriptionDrug,
+              Observation, Encounter, ExotelPhoneNumberDetail, TwilioSmsDeliveryDetail, MedicalHistory,
+              PatientBusinessIdentifier, PatientPhoneNumber, Patient, Address]
+
+    ActiveRecord::Base.transaction do
+      models.each do |model|
+        puts "Deleting #{model} data"
+        model.delete_all
```
rsanheim, line 12: Note that this won't invoke `destroy` based callbacks. Not sure how much we are using them.  

Do we have dependent destruction setup, either via AR or at the DB level?

lib/tasks/scripts/purge_users_data.rb
```diff
@@ -0,0 +1,16 @@
+module PurgeUsersData
+  def self.perform
+    return if ENV['SIMPLE_SERVER_ENV'] == 'production'
+
+    models = [BloodPressure, BloodSugar, Appointment, CallLog, Communication, PrescriptionDrug,
```
rsanheim, line 5: I'm guessing the order of models here is important to avoid FK violations. If so, can you leave a comment stating as such?

Also, for formatting long lists of Arrays like this, I would recommend doing one model per line:

```
models = [
  BloodPressure,
  BloodSuguar,
  Appointment,
... etc
]
```

It will make diffs and changing this list much easier over time.

app/jobs/populate_fake_data_job.rb
```diff
@@ -0,0 +1,285 @@
...
+            end_time: args[:time_fn].call
+          )
+        },
+        patient_sample_size: 0.20
+      }
+    }
+  end
+
+  def perform(user_id)
+    return if ENV['SIMPLE_SERVER_ENV'] == 'production'
```
rsanheim, line 200: We should raise a custom error here, this will fail silently which would be bad.

Something like


```
class InvalidSeedOperation < RuntimeError; end

def perform(user_id)
  if Rails.env.production?
    raise InvalidSeedOperation, "Cannot generate seed data in Rails.env!"
  end
...
```

This should generate a flurry of Sentry errors, which would hopefully hit someone's pager if we have alerting on exception spikes. =). Or would at least make it very clear what the error spike was.

app/jobs/populate_fake_data_job.rb
kitallis: Yeah, ideally our data layer should not be different from what you'd do through the API. That's however, currently not the case. 

We do a few things that are special in APIs:

1. Handling some tricky edge-cases around API compatibility across versions and sanitizing data accordingly in the pre-Model layers (Transformers, Validators)
2. Creating Encounters behind the scenes as and when BloodPressures or BloodSugars are synced. This could be done by simply creating an Encounter factory in here and linking the children over, but given 1. and our past experiences with datagen not going so well because stuff is happening behind the scenes, I'd chosen an ostensibly less worrisome path albeit a costlier one (in terms of speed and potentially maintenance). Though I'd argue it keeps us closer to how a real user would generate data, as it stands.

lib/tasks/scripts/purge_users_data.rb
kitallis: We only have [1 dependent destroy](https://github.com/simpledotorg/simple-server/blob/master/app/models/organization.rb#L4) in the system which is a resource that is intentionally not a part of this purge.

We've mostly never needed dependent destroys (or destroy callbacks) throughout because very few important things are actually deletable by users.

app/jobs/populate_fake_data_job.rb
rsanheim: > * Handling some tricky edge-cases around API compatibility across versions and sanitizing data accordingly in the pre-Model layers (Transformers, Validators)

This wouldn't impact seed data generation, right?

> * Creating Encounters behind the scenes as and when BloodPressures or BloodSugars are synced. This could be done by simply creating an Encounter factory in here and linking the children over, but given 1. and our past experiences with datagen not going so well because stuff is happening behind the scenes, I'd chosen an ostensibly less worrisome path albeit a costlier one (in terms of speed and potentially maintenance). Though I'd argue it keeps us closer to how a real user would generate data, as it stands.

Can you point me to the relevant Encounter code in the API? Or maybe give me a code walkthrough tomorrow ? 😄 I'm pretty free all morning eastern US time. 

It sounds like getting Encounter syncing into a proper service object decoupled from the API (possibly fired via a callback) is a pretty high priority to make our data model easier to work with and test.

I totally agree we want seed data to be correct and in the same state as when users sync it. My main point is that if we have to use the API to do it, something is not good and we gotta unravel that before it gets worse.

lib/tasks/scripts/purge_users_data.rb
rsanheim: Cool. Totally not necessary right now, but long term we should probably start to add dependent destroys in a sensible way throughout the code base. Given we are dealing with sensitive data, we are going to have to handle safely and cleanly deleting patient data at some point I'm sure.

lib/tasks/generate.rake
kitallis: Changed it to `db:seed_users_data` instead of `seed_users` cuz that might make it sound like you're creating users.

lib/tasks/purge.rake
kitallis: Changed to `db:purge_users_data`

lib/tasks/generate.rake
kitallis: We had these checks in the inner-layers, but also aggressively added checks to all boundary layers.

lib/tasks/scripts/purge_users_data.rb
kitallis: [Addressed here.](https://github.com/simpledotorg/simple-server/pull/786/commits/8c30c3878505e0a30003dfc87ab64c870b2bc448)

lib/tasks/purge.rake
kitallis: [Addressed here.](https://github.com/simpledotorg/simple-server/pull/786/commits/8c30c3878505e0a30003dfc87ab64c870b2bc448)

app/jobs/populate_fake_data_job.rb
kitallis: [Addressed here.](https://github.com/simpledotorg/simple-server/pull/786/commits/8c30c3878505e0a30003dfc87ab64c870b2bc448)

.env.development
```diff
@@ -112,3 +112,12 @@ IHCI_ORGANIZATION_UUID=89008920-021e-4116-b74c-8edf86180dd6
 OWNER_EMAILS='Owner <owner@test.com>, Another Owner <another-owner@test.com>'
 PROGRAM_INCEPTION_DATE=2018-01-01
 SUPERVISOR_EMAILS='Supervisor <supervisor@test.com>'
+
+#
+# Generate fake data
+#
+GENERATED_ADMIN_PASSWORD='password'
+ACTIVE_GENERATED_USER_ROLE="Seed User | Active"
+INACTIVE_GENERATED_USER_ROLE="Seed User | Inactive"
```
harimohanraj89, line 121: Can we prefix these env vars with `SEED_` to avoid the need for comment?

README.md
```diff
@@ -115,11 +115,28 @@ Alternatively, you can start these services locally _without_ foreman by using t
...
-from scratch.
-* Please refer to `config/seed.yml` to set the multiplier values to control the volume of seed data generated.
+To generate seed data, execute the following command from the project root:
+
+```bash
+bundle exec rake db:seed db:seed_users_data
+```
+
+**Note**: This spins up sidekiq jobs to generate data and expects the server+sidekiq to be running. 
+Ensure that you are running:
```
harimohanraj89, line 127: I'd just put this step right before the seed command to minimize frustration.

> To generate seed data, run the following commands to start up the server's web and sidekiq processes,
> and generate the seed data.

```
$ foreman start -f Procfile.dev
$ bundle exec rake db:seed db:seed_users_data
```

README.md
```diff
@@ -115,11 +115,28 @@ Alternatively, you can start these services locally _without_ foreman by using t
...
-Note:
+### Generating seed data
 
-* For the `development` environment, this will truncate existing data and seed the database
-from scratch.
-* Please refer to `config/seed.yml` to set the multiplier values to control the volume of seed data generated.
+To generate seed data, execute the following command from the project root:
+
+```bash
+bundle exec rake db:seed db:seed_users_data
```
harimohanraj89, line 123: Consider s/rake/rails/ to keep up with Rails 5 convention.

config/application.rb
```diff
@@ -35,7 +34,7 @@ class Application < Rails::Application
 
     # Locale configuration
     config.i18n.load_path += Dir[Rails.root.join('config', 'locales', '**', '*.{rb,yml}')]
-    config.i18n.available_locales = %w[en mr-IN pa-Guru-IN bn-BD en_BD en_IN bn-IN hi-IN ta-IN te-IN or-ET ti-ET]
+    config.i18n.available_locales = %w[en mr-IN pa-Guru-IN bn-BD en_BD en_IN en-IND bn-IN hi-IN ta-IN te-IN or-ET ti-ET]
```
harimohanraj89, line 37: If we're afforded flexibility, can we name the new locale `en-SEED`?

spec/factories/facilities.rb
```diff
@@ -1,22 +1,22 @@
...
-    zone 'Block ABC'
-    facility_type 'PHC'
+    name { "#{facility_type} #{village_or_colony}" + rand(1..100).to_s }
+    street_address { Faker::Address.street_address }
+    village_or_colony { Faker::Address.village }
+    district { Faker::Address.district }
+    state { Faker::Address.state }
+    country { 'India' }
+    pin { Faker::Address.zip_code }
+    zone { Faker::Address.block }
```
harimohanraj89, line 11: These changes might really slow down our regular test suite. Could we add a trait for this - eg. `realistic` or `seed`?

spec/factories/medical_histories.rb
```diff
@@ -23,6 +23,14 @@
       diagnosed_with_hypertension { MedicalHistory::MEDICAL_HISTORY_ANSWERS[:unknown] }
     end
 
+    trait :hypertension do
```
harimohanraj89, line 26: Consider adjectives, `hypertensive`, `diabetic`?

kitallis: I considered that. I thought it might get confusing because `hypertensive` and `diabetic` means something else in our system: high BP, high BS etc.

What do you think?

config/application.rb
kitallis: Can't, this unforuntately is the locale used by Fake for Indian things. It uses `en-IND` instead of `en-IN` for some reason. This allows us to extend Faker's corpus.

spec/factories/medical_histories.rb
kitallis: However, on second-thought, Diagnosis is pretty canonical at this point.

prabhanshuguptagit: How about `hypertension_yes` and `diabetes_yes`?

kitallis:
@rsanheim @harimohanraj89 I think we can merge this in. I think the seed generation being driven through APIs and a related walkthrough for @rsanheim for context needn't block the merging for this PR since it's functional as it stands.

Separately, regardless of what the outcome of that conversation is, it's pretty easy to not call APIs why this script as there's already a model that **[doesn't](https://github.com/simpledotorg/simple-server/pull/786/files#diff-3c9164570647a7d8ec597460b975cb73R167)** go through APIs.

spec/factories/medical_histories.rb
harimohanraj89: @prabhanshuguptagit I'm not a big fan of the `yes` suffix. I'd say the original `hypertension` `diabetes` naming was better. I'll let you take a final call either way 👍

app/jobs/populate_fake_data_job.rb
harimohanraj89: @rsanheim you're absolutely right. Our model layer doesn't represent all the constraints necessary for coherent data, so we're forced to use our API layer. As Kitty mentions below, I don't think this needs to block this PR in particular, as fixing this model issue is a larger effort. Let's walk through it together soon!

config/schedule.rb
```diff
@@ -25,3 +25,8 @@ def local(time)
 every 3.hours, roles: [:cron] do
   rake 'refresh_materialized_db_views'
 end
+
+every :week, at: local('02:00 am'), roles: [:cron] do
```
harimohanraj89, line 29: Let's do this every month instead while we figure it out. If we do it every week, I feel that by the time we identify and fix issues from one seed round, the next seed round will run again.

harimohanraj89: This will run on every environment right?? I'd rather not rely on protection in our script against running in production. Can we somehow add this task to a separate role that is only spun up on Sandbox, etc, and deliberately omitted from our production environments?

PR #787, opened 2020-02-05
Remove config that allows selectively disabling apis

govindkrjoshi:
This config was initially used during MVP launch to selectively turn on APIs for syncing. 
This is no longer needed and causes syncing issues when an api is not added.

https://www.pivotaltracker.com/story/show/171126343


harimohanraj89:
@govindkrjoshi can you please create a story (bug I guess?) for this PR and link to this PR?

PR #791, opened 2020-02-10
My facilities missed visits

vkrmis:
Adds the My Facilities - Missed Visits dashboard

Figma - https://www.figma.com/file/kONxBLSXwcQDnV7ya4hIXA/Simple(Exploration)?node-id=2381%3A74431

Pivotal - https://www.pivotaltracker.com/epic/show/4476152

Definitions as used in the dashboard - https://docs.google.com/document/d/1N0wwRILLuMVcga53KCK-WF8KLmcycGbpMJP1-It9nrA/edit#heading=h.erre6mb66bg0

Missed Visits uses the notion of a 'responsible facility' for each patient at a point in time, for now this is defined as the facility that recorded the latest blood pressure in the previous time period (quarter/month). This is captured in the Materialized Views for Blood Pressures. The MissedVisitQuery builds on this to figure out how many patients a facility is responsible for in a given period, and how many of them have visited in the following period. 

image


harimohanraj89:
@vkrmis can you add a screenshot of this page to the PR description? It'll help provide context for PR review (eg. How are we rendering call logs?)

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -0,0 +1,86 @@
...
+    # last_n is the number of quarters/months data to be returned
+    @facilities = facilities
+    @period = period
+    @periods = period_list(period, last_n)
+    @latest_period = @periods.first
+  end
+
+  def patients
+    @patients ||=
+      Patient
```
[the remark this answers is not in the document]
vkrmis: Hmm, I think we can leave it as-is. The latest_bp from the views will be updated only once a day, so it can be stale. 
The view also returns a LatestBloodPressurePerPatient object instead of a BloodPressure object, this might not be what we want from Patient.latest_blood_pressures. We'll have to join this with the BloodPressures table to return a BP object

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -0,0 +1,86 @@
...
+      .joins('INNER JOIN latest_blood_pressures_per_patients
+              ON patients.id = latest_blood_pressures_per_patients.patient_id')
+      .where(latest_blood_pressures_per_patients: { bp_facility_id: @facilities })
+      .where('recorded_at < ?', Time.current.beginning_of_day - 2.months)
```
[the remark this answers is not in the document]
vkrmis: Constantizing rather than adding to config. This is a definition related number, don't want it to get lost in the configs.

app/controllers/my_facilities_controller.rb
```diff
@@ -89,7 +103,33 @@ def authorize_my_facilities
     authorize(:dashboard, :show?)
   end
 
-  def missed_visits(registered_patients, controlled_bps, uncontrolled_bps)
+  def calculate_missed_visits(registered_patients, controlled_bps, uncontrolled_bps)
     registered_patients - controlled_bps - uncontrolled_bps
   end
+
+  def missed_visits_by_facility(query)
```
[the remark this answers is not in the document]
vkrmis: Decided to move this to the query namespace

app/controllers/my_facilities_controller.rb
```diff
@@ -89,7 +103,33 @@ def authorize_my_facilities
...
+
+      responsible_patients.map do |facility_id, patient_count|
+        [[facility_id, year, period],
+         { patients: patient_count.to_i,
+           missed: patient_count.to_i - visited_patients[facility_id].to_i }]
+      end.to_h
+    end.reduce(:merge)
+  end
+
+  def missed_visit_totals(query)
```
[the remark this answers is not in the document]
vkrmis: Decided to move this to the query namespace

app/controllers/my_facilities_controller.rb
```diff
@@ -77,6 +77,20 @@ def registrations
...
+    missed_visits_query = MyFacilities::MissedVisitsQuery.new(facilities: @facilities,
+                                                              period: @selected_period,
+                                                              last_n: PERIODS_TO_DISPLAY[@selected_period])
+
+    @display_periods = missed_visits_query.periods
+    @missed_visits_by_facility = missed_visits_query.missed_visits_by_facility
+    @calls_made = missed_visits_query.calls_made.count
+    @all_time_registrations = missed_visits_query.patients.group(:bp_facility_id).count
+    @totals_by_period = missed_visits_query.missed_visit_totals
+  end
```
harimohanraj89: This controller action looks really neat!

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -0,0 +1,116 @@
...
+    @latest_period = @periods.first
+  end
+
+  def patients
+    @patients ||=
+      Patient
+      .joins('INNER JOIN latest_blood_pressures_per_patients
+              ON patients.id = latest_blood_pressures_per_patients.patient_id')
+      .where(latest_blood_pressures_per_patients: { bp_facility_id: @facilities })
+      .where('recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
```
harimohanraj89: This clause excludes recently registered patients?

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -0,0 +1,116 @@
...
+  end
+
+  def patients_by_period
+    @patients_by_period ||=
+      @periods.map do |year, period|
+        period_start = (@period == :quarter ? local_quarter_start(year, period) : local_month_start(year, period))
+        [[year, period], patients.where('patient_recorded_at < ?', period_start - REGISTRATION_BUFFER)]
```
harimohanraj89: Is this assembling patients from, say, Jan-Mar, Apr-Jun, Jul-Sept, or is it doing (beginning of time - Jan), (beginning of time - Apr), (beginning of time - Sept)?

app/views/layouts/my_facilities.html.erb
```diff
@@ -79,7 +84,7 @@
             </div>
           </div>
           <div class="modal-footer">
-            <button type="button" class="btn btn-secondary" data-dismiss="modal">Clear all</button>
+            <a href="?" type="button" class="btn btn-secondary">Clear all</a>
```
harimohanraj89: Why a `?` href? Can we use `#` instead?

app/views/layouts/my_facilities.html.erb
```diff
@@ -95,20 +100,25 @@
...
+                    class: "nav-link #{active_action?("missed_visits")}" %>
       </nav>
       <% unless request.path == my_facilities_overview_path %>
         <section id="facility-filters" class="secondary-nav-filters">
           <h4>Facility Filters</h4>
           <div class="dropdown show">
             <a class="btn btn-sm dropdown-toggle" href="#" role="button" id="dropdownMenuLink" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-              All blocks
+              <% if @selected_zones.size == 1 %>
+                <%= @selected_zones.first&.capitalize %>
```
harimohanraj89: If we have one selected zone, why do we need to use the safe navigation `&.capitalize`?

app/views/my_facilities/missed_visits.html.erb
```diff
@@ -0,0 +1,130 @@
...
+            <% @display_periods.reverse.each do |(year, quarter)| %>
+              <th colspan="2"><%= "Q#{quarter}-#{year}" %></th>
+            <% end %>
+          <% when :month %>
+            <% @display_periods.reverse.each do |(year, month)| %>
+              <th colspan="2"><%= month_short_name_and_year(local_month_start(year, month)) %></th>
+            <% end %>
+          <% end %>
+          <th>Calls made <sup>1</sup></th>
+          <th>ALL-TIME</th>
```
harimohanraj89: I recommend upcasing this through CSS, what do you think?

app/views/my_facilities/missed_visits.html.erb
```diff
@@ -0,0 +1,130 @@
...
+        </tr>
+        </thead>
+        <tbody>
+        <tr class="row-total" data-sort-method="none">
+          <td class="type-title">All</td>
+
+          <% @display_periods.reverse.each do |period| %>
+            <td class="type-percent" data-sort-column-key="<%= period.join %>">
+              <em><%=
+                percentage(@totals_by_period.dig([period.first, period.second], :missed),
```
harimohanraj89: Is `period` an array with two elements? Do you think a hash would make our code much more readable?

`...dig([period[:start], period[:finish])`

harimohanraj89: Actually, I may not be understanding the data structure of `@totals_by_period`, could you help me understand?

db/migrate/20200210074711_update_latest_blood_pressures_per_patient_per_months_to_version_2.rb
```diff
@@ -0,0 +1,12 @@
...
+    drop_view :latest_blood_pressures_per_patient_per_quarters, materialized: true, revert_to_version: 1
+    drop_view :latest_blood_pressures_per_patients, materialized: true, revert_to_version: 1
+    update_view :latest_blood_pressures_per_patient_per_months,
+      version: 2,
+      revert_to_version: 1,
+      materialized: true
+    create_view :latest_blood_pressures_per_patient_per_quarters, materialized: true
+    create_view :latest_blood_pressures_per_patients, materialized: true
+  end
+end
```
harimohanraj89: Can you describe briefly what schema changes are being made to the views?

app/queries/my_facilities/missed_visits_query.rb
vkrmis: Yes, this is intentional and part of the definition. The idea is that we don't want recently registered patients to skew these numbers too much.

app/queries/my_facilities/missed_visits_query.rb
vkrmis: The second -> Beginning of time up to the point the period starts

app/views/layouts/my_facilities.html.erb
vkrmis: The `?` clears the query params, and therefore clears out the filters. The `#` doesn't do that.

app/views/my_facilities/missed_visits.html.erb
vkrmis: Hmm, not sure that's necessary, given that its just a static string. Does upcasing this through CSS do something else I'm missing?

db/migrate/20200210074711_update_latest_blood_pressures_per_patient_per_months_to_version_2.rb
vkrmis: We added a new column in v2 called `responsible_facility` This is the `facility_id` of the last bp in the previous period, and is used to calculate missed visits

app/views/my_facilities/missed_visits.html.erb
vkrmis: Yes, `period` is an array with two elements, the first element is the year, and the second is either month or quarter, all of these are integers.

totals_by_period is a hash map keyed by period, so it looks like this:
`{[2020, 2]=>{:patients=>14498, :missed=>7474}, 
[2020, 1]=>{:patients=>13222, :missed=>3868}, 
[2019, 12]=>{:patients=>11845, :missed=>4478}}`

The `period.first` and `period.second` stuff is unnecessary, I'll refactor it to just be 
`.dig(period, :missed)`.

app/controllers/concerns/period_selection.rb
```diff
@@ -4,10 +4,18 @@ module PeriodSelection
...
     def set_selected_period
       @selected_period = params[:period].blank? ? :quarter : params[:period].to_sym
     end
+
+    def populate_all_periods
+      @periods = PERIODS
+    end
+
+    def populate_periods(periods)
+      @periods = PERIODS.select { |period, _| periods.include?(period) }
```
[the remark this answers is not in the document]
vkrmis: We've removed this method entirely, available periods are now declaratively defined in a map in the concern

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -0,0 +1,90 @@
...
+    when :quarter then
+      last_n_quarters(n: last_n, inclusive: true)
+    when :month then
+      last_n_months(n: last_n, inclusive: true)
+        .map { |month| [month.year, month.month] }
+    end
+  end
+
+  def periods_as_sql_list
+    @periods.map { |(year, period)| "('#{year}', '#{period}')" }.join(',')
```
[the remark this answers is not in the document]
vkrmis: We've removed this method now, we're generating this list elsewhere using a reduce.

I thought we did need the `()` to destructure, but it works without them as well.

app/queries/my_facilities/missed_visits_query.rb
harimohanraj89: Cool. Wanted to confirm that that's the intention.

app/views/layouts/my_facilities.html.erb
harimohanraj89: Oooooh 👍

app/views/my_facilities/missed_visits.html.erb
harimohanraj89: Just separates the content from the "presentation". I think it's a good practice to follow.

app/views/layouts/my_facilities.html.erb
vkrmis: removing the `&`, think it got left behind from a refactor

app/views/my_facilities/missed_visits.html.erb
vkrmis: Turns out this was already being uppercased by some table header css, just changed it to `All-time` in the html.

app/queries/my_facilities/missed_visits_query.rb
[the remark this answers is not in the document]
govindkrjoshi: We should try to have a single source of truth for all our definitions. If we are using the view in our reports, we should use it else where in the dashboard as well.

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -0,0 +1,116 @@
...
+          [key, visits_in_month(*key, patients)]
+        end
+      end.to_h
+  end
+
+  def calls_made
+    period_start = (@period == :quarter ? local_quarter_start(*@latest_period) : local_month_start(*@latest_period))
+    period_end = @period == :quarter ? period_start.end_of_quarter : period_start.end_of_month
+
+    CallLog
```
harimohanraj89: Let's cache this one in an instance var too, yeah?

app/queries/my_facilities/registrations_query.rb
```diff
@@ -44,6 +44,6 @@ def period_list(period, last_n)
   end
 
   def periods_as_sql_list
-    @periods.map { |(year, day)| "('#{year}', '#{day}')" }.join(',')
+    @periods.map { |(year, period)| "('#{year}', '#{period}')" }.join(',')
```
harimohanraj89: I think you mentioned in an earlier comment that we don't need to destructure the block args with `()`.

spec/queries/my_facilities/missed_visits_query_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+    context '#patients_by_period' do
+      it 'should bucket patients by the period they registered in' do
+        expect(query.patients_by_period[query.periods.first].map(&:id).sort).to eq(patients.map(&:id).sort)
+        expect(query.patients_by_period[query.periods.second].map(&:id).sort).to eq([patients.first.id])
+        expect(query.patients_by_period[query.periods.third].map(&:id)).to eq([patients.first.id])
+      end
+    end
+
+    context '#visits_by_period' do
+      it "should bucket visits by the period they were recorded in" do
```
harimohanraj89: Some quotation issues in the spec descriptions. Additionally, it's generally good practice to stop using "should" in the spec description as well. Instead, consider something like `it "buckets visits by the period they were recorded in"`. I think this can be applied in a few places across the specs. 👍

PR #803, opened 2020-02-14
Update bangladesh medical import script to ensure safe imports

govindkrjoshi:
The PR fixes the following issues while importing data for bangladesh:

- Fix patient gender to `male`, `female` or `transgender`
- Ensure either `date_of_birth` or (`age` and `age_updated_at`) are present
- Add patient status
- Add patient phone number status
- Assign `unknown` to nil medical history questions
- Add metadata and metadata_version to patient business identifiers
- Add an empty rxnorm_code to prescription drugs

### DIFF
--- a/lib/bangladesh/to_patients.rb
+++ b/lib/bangladesh/to_patients.rb
@@ -257,16 +257,6 @@ def create_patient(params)
       )
     end
 
-    if Appointment.group(:creation_facility_id).count.keys.any? { |k, _| k.nil? } ||
-      Patient.group(:recorded_at).count.keys.any? { |k, _| k.nil? }
-
-      puts  $registration_facility.name
-      puts "#{Appointment.group(:creation_facility_id).count}"
-      puts "#{Patient.group(:recorded_at).count}"
-      raise "HOLA"
-
-    end
-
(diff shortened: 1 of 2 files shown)


harimohanraj89:
We need to add `receiving_treatment_for_hypertension` to [this list](https://github.com/simpledotorg/simple-server/blob/master/lib/bangladesh/to_patients.rb#L137)

lib/bangladesh/to_patients.rb
```diff
@@ -56,35 +60,69 @@
...
 end
 
+def patient_recorded_at(blood_pressures)
+  date = blood_pressures.map { |bp| bp[:recorded_at] }.min
+  return DateTime.new(2019, 6, 1) if date.nil?
+  date
+end
+
 def create_patient(params)
+  if params[:gender].blank?
```
kitallis: Any idea why this doesn't work? Does the sheet have spaced strings?

timcheadle: We moved the gender check to where the CSV is parsed to fix this

harimohanraj89:
Closing in favor of #809

PR #806, opened 2020-02-18
Updated failed login error message

timcheadle:
The previous message indicated you needed to sign up, which of course we don't allow.


timcheadle:
Good catch @harimohanraj89. Will fix.

PR #807, opened 2020-02-19
Add api validation skip toggle

vkrmis:
This PR adds a toggle `SKIP_API_VALIDATION` to allow us to skip schema validations selectively.
It will skip schema validations for the PatientsController merge_if_valid method if the toggle is enabled


app/schema/api/current/models.rb
```diff
@@ -87,14 +87,14 @@ def patient_business_identifier
...
-          identifier: { '$ref' => '#/definitions/non_empty_string' },
+          identifier: { type: [:string, 'null'] },
           identifier_type: { type: :string, enum: PatientBusinessIdentifier.identifier_types.keys },
           metadata_version: { type: :string },
           metadata: { type: :string },
           deleted_at: { '$ref' => '#/definitions/nullable_timestamp' },
           created_at: { '$ref' => '#/definitions/timestamp' },
           updated_at: { '$ref' => '#/definitions/timestamp' } },
-        required: %w[id created_at updated_at identifier identifier_type] }
+        required: %w[id created_at updated_at identifier_type] }
```
timcheadle: `identifier` is still required, no? We're just allowing it to be blank instead?

harimohanraj89: Probably. Since we're turning off JSON validation anyway, and we need to revisit this stuff, we're not gonna try addressing this right now for today's hotfix.

PR #810, opened 2020-02-20
Coldfixes 2020-02-20

harimohanraj89:
Story: https://www.pivotaltracker.com/story/show/171380089

This PR includes cleaned-up versions of several hotfixes and additional instrumentation that has been deployed to IHCI and Bangladesh environments separately.

Notably, it includes:
* Blank identifier fixes for both environments
* SMS Country Code fixes, intended for use in BD
* BD import scripting, to keep around if we need to re-use it. We'll remove once the dust settles after the BD launch
* Ability to skip API schema validations, intended for IHCI
* NewRelic Instrumentation to key sync methods


app/controllers/api/current/patients_controller.rb
```diff
@@ -46,17 +46,25 @@ def unscope_associations
   end
 
   def merge_if_valid(single_patient_params)
-    validator = Api::Current::PatientPayloadValidator.new(single_patient_params)
-    logger.debug "Patient had errors: #{validator.errors_hash}" if validator.invalid?
-    if validator.invalid?
-      NewRelic::Agent.increment_metric('Merge/Patient/schema_invalid')
-      { errors_hash: validator.errors_hash }
-    else
+    if FeatureToggle.enabled?('SKIP_API_VALIDATION')
```
kitallis: Let's maybe remove this code-change in favor of disabling it from this [place](https://github.com/simpledotorg/simple-server/blob/a4ea71b23b1881b0efabfdd3407ef3f30558d446/app/validators/api/current/patient_payload_validator.rb#L24) for a much cleaner disable?

app/schema/api/current/models.rb
```diff
@@ -87,14 +87,14 @@ def patient_business_identifier
...
-          identifier: { '$ref' => '#/definitions/non_empty_string' },
+          identifier: { type: [:string, 'null'] },
           identifier_type: { type: :string, enum: PatientBusinessIdentifier.identifier_types.keys },
           metadata_version: { type: :string },
           metadata: { type: :string },
           deleted_at: { '$ref' => '#/definitions/nullable_timestamp' },
           created_at: { '$ref' => '#/definitions/timestamp' },
           updated_at: { '$ref' => '#/definitions/timestamp' } },
-        required: %w[id created_at updated_at identifier identifier_type] }
+        required: %w[id created_at updated_at identifier_type] }
```
kitallis: We can make `identifier` required but `null`able.

app/models/patient_business_identifier.rb
```diff
@@ -8,7 +8,8 @@ class PatientBusinessIdentifier < ApplicationRecord
     bangladesh_national_id: 'bangladesh_national_id'
   }
 
-  validates :identifier, presence: true
+  validates :identifier, presence: true, if: -> { identifier_type == 'simple_bp_passport' }
```
timcheadle: What happens when it doesn't match Simple BP passport NOR Bangladesh ID?

This should be `identifier_type != 'bangladesh_national_id`

spec/controllers/api/current/patients_controller_spec.rb
```diff
@@ -229,6 +229,39 @@ def create_record_list(n, options = {})
...
+          payload_with_missing_biz_id = patients_payload.deep_merge('business_identifiers' => [business_identifier])
+
+          post :sync_from_user, params: { patients: [payload_with_missing_biz_id] }, as: :json
+
+          expect(response).to have_http_status(200)
+          # Nested errors are not currently reported, so the response errors map doesn't contain an error
+          # Checking if the PatientBusinessIdentifier got created instead
+          expect(JSON.parse(response.body)['errors']).to be_empty
+          expect(PatientBusinessIdentifier.where(id: business_identifier['id']).count).to eq 0
+        end
```
timcheadle: Also need a test case for when the `identifier_type` is neither Simple nor BD

app/services/sms_notification_service.rb
```diff
@@ -29,7 +29,7 @@ def send_reminder_sms(reminder_type, appointment, callback_url, locale = DEFAULT
   def send_sms(body, callback_url = '')
     client.messages.create(
       from: sender_phone_number,
-      to: recipient_number.insert(0, I18n.t('sms.country_code')),
+      to: recipient_number.insert(0, ENV['SMS_COUNTRY_CODE'] || I18n.t('sms.country_code')),
```
kitallis: What do we think about moving this information to the `countries.rb` initializer?

harimohanraj89: Great idea. We should do that now, as we're cold fixing 👍

kitallis:
@harimohanraj89 Does it make sense to move the `lib/bangladesh/` files to `lib/bangladesh/import/` directory?

PR #811, opened 2020-02-21
My facilities Overview: Performance improvements

prabhanshuguptagit:
This adds a SQL view `blood_pressures_per_facility_per_day` to back the My Facilities overview page. The inactive facilities calculation used a join before to compute the number of BPs in the last `n` days. That calculation has been moved to the SQL view.

https://www.pivotaltracker.com/story/show/170919252

-- 
Note to self: Change base to master before merge.


app/views/shared/my_facilities/_inactive_facilities.html.erb
```diff
@@ -19,11 +19,11 @@
       <% @inactive_facilities.sort_by(&:name).each do |facility| %>
         <tr>
           <td><%= link_to facility.name, analytics_facility_path(facility) %></td>
-          <td class="text-right"><%= @inactive_facilities_bp_counts[:last_week][facility.id] %></td>
-          <td class="text-right"><%= @inactive_facilities_bp_counts[:last_month][facility.id] %></td>
+          <td class="text-right"><%= @inactive_facilities_bp_counts[:last_week][facility.id].to_i %></td>
+          <td class="text-right"><%= @inactive_facilities_bp_counts[:last_month][facility.id].to_i %></td>
```
harimohanraj89: Why did we add `.to_i` here? To ensure we show `0` if nil?

spec/models/blood_pressures_per_facility_per_day_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      days.map do |day|
+          create_list(:blood_pressure, 2, facility: facility_with_bp, recorded_at: day, patient: patient)
+      end + [create(:blood_pressure, facility: facility_with_bp, recorded_at: days.first, patient: patient_2)]
+    end
+
+    before do
+      described_class.refresh
+    end
+
+    it 'should return a row per facility per day' do
```
harimohanraj89: Can we remove the "should"s from the spec descriptions? It's generally better (and fewer keystrokes) to have the descriptions read something like `it 'returns a row per facility per day'`, etc.

app/models/facility.rb
```diff
@@ -100,13 +100,6 @@ def self.parse_facilities(file_contents)
     facilities
   end
 
-  def self.bp_counts_in_period(start:, finish:)
```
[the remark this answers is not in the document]
prabhanshuguptagit: This method was added specifically for the inactive facilities page. It's good to remove.

app/views/shared/my_facilities/_inactive_facilities.html.erb
prabhanshuguptagit: Yes, the `to_i` makes the `nil` a zero.

PR #815, opened 2020-02-24
Fixed the path to the favicon so it looks correct on production. Yay!

dburka:

PR #816, opened 2020-02-24
Fixed the path to the favicon so it is correct on production. Yay!

harimohanraj89:
This PR replaces #815 , since #815 was not based off of master

Favicon `link` tags were not using Rails asset helpers to reference the favicon assets. and hence favicon assets were broken. This PR fixes this so favicons work in production-like environments again. Screenshot of the PR on QA environment below:

Screenshot 2020-02-24 at 2 52 12 PM

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -21,7 +21,7 @@
     <%= stylesheet_link_tag    'application', media: 'all' %>
     <%= javascript_include_tag 'application' %>
 
-    <link rel="icon" type="image/png" href="/assets/<%= favicon_for_environment %>">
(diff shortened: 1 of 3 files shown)


app/views/layouts/application.html.erb
```diff
@@ -21,7 +21,7 @@
     <%= stylesheet_link_tag    'application', media: 'all' %>
     <%= javascript_include_tag 'application' %>
 
-    <link rel="icon" type="image/png" href="/assets/<%= favicon_for_environment %>">
+    <link rel="icon" type="image/png" href="<%= image_path(favicon_for_environment) %>">
```
govindkrjoshi: Can we add a test for this?

PR #823, opened 2020-02-26
Fix issues with caching on the [OLD] dashboards 

kitallis:
https://www.pivotaltracker.com/story/show/171490643

It seems like we were calling the dashboard queries before caching, at all times, eagerly, defeating the purpose of caching in the first place. 

The cache itself is set and retrieved correctly, but the query is still made needlessly.

**Fix**

The query data is now yielded lazily as per a cache hit or miss.

This was discovered due to the huge amount of data created on SBX.

### DIFF
--- a/app/controllers/analytics/districts_controller.rb
+++ b/app/controllers/analytics/districts_controller.rb
@@ -60,17 +60,19 @@ class Analytics::DistrictsController < AnalyticsController
   end
 
   def set_cohort_analytics(period, prev_periods)
-    @cohort_analytics = set_analytics_cache(
-      analytics_cache_key_cohort(period),
-      @organization_district.cohort_analytics(period, prev_periods))
+    @cohort_analytics =
+      set_analytics_cache(analytics_cache_key_cohort(period)) do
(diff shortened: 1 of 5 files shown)


app/controllers/analytics/districts_controller.rb
```diff
@@ -60,17 +60,19 @@ def set_organization_district
   end
 
   def set_cohort_analytics(period, prev_periods)
-    @cohort_analytics = set_analytics_cache(
-      analytics_cache_key_cohort(period),
-      @organization_district.cohort_analytics(period, prev_periods))
+    @cohort_analytics =
+      set_analytics_cache(analytics_cache_key_cohort(period)) do
+        @organization_district.cohort_analytics(period, prev_periods)
+      end
```
harimohanraj89: Oh... oh dear... I see what happened here 😅

kitallis: In Haskell, everything is lazy by default.

kitallis:
This branch was tested on SBX and it works as expected 🦀.

PR #824, opened 2020-02-27
Refresh materialized views concurrently

prabhanshuguptagit:
The materialized view refreshes for My facilities are non-concurrent right now. This means that the views are locked while they are being refreshed. 

Postgres 9.4+ (and scenic) supports concurrent view refreshes [[link]](https://github.com/scenic-views/scenic#what-about-materialized-views). This PR adds concurrent refreshability for all  materialized views. This will allow us to continue reading even while the refreshes happen. 

https://www.pivotaltracker.com/story/show/171532526


db/schema.rb
```diff
@@ -746,6 +746,21 @@
...
+      (date_part('year'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('utc'::text, patients.recorded_at))))::text AS year
+     FROM (patients
+       JOIN facilities ON ((patients.registration_facility_id = facilities.id)))
+    WHERE (patients.deleted_at IS NULL)
+    GROUP BY (date_part('doy'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('utc'::text, patients.recorded_at))))::text, (date_part('month'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('utc'::text, patients.recorded_at))))::text, (date_part('quarter'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('utc'::text, patients.recorded_at))))::text, (date_part('year'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('utc'::text, patients.recorded_at))))::text, patients.registration_facility_id, facilities.deleted_at;
+  SQL
+  add_index "patient_registrations_per_day_per_facilities", ["facility_id", "day", "year"], name: "index_patient_registrations_per_day_per_facilities", unique: true
+
```
harimohanraj89: Looks like Rails isn't great at preserving the order of these views yet? 😄

PR #829, opened 2020-03-01
Complete rewrite of the HELP section

dburka:
Please make sure I can see this in Sandbox before it goes out. It looks fine in Chrome on mobile but really needs to be checked on an Android device in the webview before going live.


kitallis:
@harimohanraj89 This looks good. Let's do a run through on how to get `en.html` translated on transifex? 👍

PR #831, opened 2020-03-02
BD phone number correction

harimohanraj89:
Story: https://www.pivotaltracker.com/story/show/171538058

This PR adds a rake task and associated services to correct phone numbers for Bangladesh patients.

Imported Bangladesh patient phone numbers have two issues that make it difficult to search for patients by phone number:
* Missing leading zero
* Presence of dashes

This PR adds leading zeros and removes dashes to improve searchability.


kitallis:
This PR looks good. I'm just not certain whether a `bangladesh.rake` makes sense though. If it's a one-off, then maybe we could stick it in a one-off namespace. Otherwise, `data_migration.rake` makes more sense since it already has a check for Bangladesh so it's safe.

harimohanraj89:
I like it. `rake data_migration:correct_bangladesh_phone_numbers` sounds good.

PR #832, opened 2020-03-02
Use registration_facility_id instead of responsible_facility_id to calculate missed visits

vkrmis:
This PR will switch `missed_visit` calculations to use patients' `registration_facility_id` instead of the derived `responsible_facility_id`. 

It doesn't remove the `responsible_facility_id` column in the materialized views or the associated specs, I've added a story to do that here: https://www.pivotaltracker.com/story/show/171576689

### DIFF
--- a/app/queries/my_facilities/missed_visits_query.rb
+++ b/app/queries/my_facilities/missed_visits_query.rb
@@ -12,7 +12,7 @@ class MyFacilities::MissedVisitsQuery
 
   def initialize(facilities: Facility.all, period: :quarter, last_n: 3)
     # period can be :quarter, :month.
-    # last_n is the number of quarters/months data to be returned
+    # last_n is the number of quarters/months for which data is to be returned
     @facilities = facilities
     @period = period
     @periods = period_list(period, last_n)
@@ -22,9 +22,7 @@ class MyFacilities::MissedVisitsQuery
   def patients
     @patients ||=
       Patient
-      .joins('INNER JOIN latest_blood_pressures_per_patients
-              ON patients.id = latest_blood_pressures_per_patients.patient_id')
-      .where(latest_blood_pressures_per_patients: { bp_facility_id: @facilities })
+      .where(registration_facility_id: @facilities)
       .where('recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
   end
 
@@ -32,7 +30,7 @@ class MyFacilities::MissedVisitsQuery
     @patients_by_period ||=
       @periods.map do |year, period|
         period_start = (@period == :quarter ? local_quarter_start(year, period) : local_month_start(year, period))
-        [[year, period], patients.where('patient_recorded_at < ?', period_start - REGISTRATION_BUFFER)]
+        [[year, period], patients.where('recorded_at < ?', period_start - REGISTRATION_BUFFER)]
       end.to_h
   end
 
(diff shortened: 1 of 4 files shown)


app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -22,17 +22,15 @@ def initialize(facilities: Facility.all, period: :quarter, last_n: 3)
...
+      .where(registration_facility_id: @facilities)
       .where('recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
   end
 
   def patients_by_period
     @patients_by_period ||=
       @periods.map do |year, period|
         period_start = (@period == :quarter ? local_quarter_start(year, period) : local_month_start(year, period))
-        [[year, period], patients.where('patient_recorded_at < ?', period_start - REGISTRATION_BUFFER)]
+        [[year, period], patients.where('recorded_at < ?', period_start - REGISTRATION_BUFFER)]
```
harimohanraj89: Has this changed because we no longer join with another table?

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -95,14 +93,12 @@ def missed_visit_totals
 
   def visits_in_quarter(year, quarter, patients)
     LatestBloodPressuresPerPatientPerQuarter
-      .where(year: year, quarter: quarter)
-      .where(patient_id: patients)
+      .where(year: year, quarter: quarter, patient: patients)
```
harimohanraj89: This clause used to be `patient_id: patients`, and now it is `patient: patients`. Can we confirm that both work?

app/queries/my_facilities/missed_visits_query.rb
vkrmis: Yup, the joined table had `patient_recorded_at` and `bp_recorded_at`. Need to use `recorded_at` now since we're just looking up Patient.

app/queries/my_facilities/missed_visits_query.rb
vkrmis: Yup, this works because `LatestBloodPressuresPerPatientPerQuarter` and `LatestBloodPressuresPerPatientPerMonth` both have a `belongs_to :patient` relationship. This needed to use `patient_id` earlier because the `patients` arg we were passing in was actually the result of the join up top, and not just a `Patient` relation.

harimohanraj89: Perfect 👌

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -66,8 +64,8 @@ def calls_made
   def missed_visits_by_facility
     @missed_visits_by_facility ||=
       patients_by_period.map do |(year, period), patients|
-        responsible_patients = patients.group(:bp_facility_id).count
-        visited_patients = visits_by_period[[year, period]].group(:responsible_facility_id).count
+        responsible_patients = patients.group(:registration_facility_id).count
```
prabhanshuguptagit: Should we rename this to `registered_patients`?

harimohanraj89: I like keeping it as `responsible_patients`. This solidifies the idea that we will eternally care about responsible patients, and our definition of what "responsible" means may shift over time.

vkrmis:
> Looks great. Let's test this against the "negative missed visits" problem to ensure we've fixed it.

I've done this manually to some extent in a console session against the same data. Unsure of how to replicate the situation to verify. Since we actually changed the definition slightly rather than just fixing the bug with the previous definition, I haven't been able to simulate the negative missed visits situation.

harimohanraj89:
@vkrmis Can we look at the problematic facility in production and recreate similar conditions? Do we understand _how_ this bug occurred? Eg. "Register and record BPs for 2 patients in facility A and and 2 patients in facility B, record a BP for all 4 in facility B the next month.

PR #833, opened 2020-03-02
Add hba1c blood sugar support

vkrmis:
This PR adds support for float `blood_sugar_value`s, and adds support for the `hba1c` `blood_sugar_type`

https://www.pivotaltracker.com/story/show/171377632


app/schema/api/current/models.rb
```diff
@@ -155,7 +155,7 @@ def blood_sugar
         properties: {
           id: { '$ref' => '#/definitions/uuid' },
           blood_sugar_type: { type: :string, enum: BloodSugar.blood_sugar_types.keys },
-          blood_sugar_value: { type: :integer },
+          blood_sugar_value: { type: :decimal },
```
kitallis: There appears to be also be a `number` type in JSON schema. Is that any different from `decimal`?

db/schema.rb
```diff
@@ -89,7 +89,7 @@
 
   create_table "blood_sugars", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.string "blood_sugar_type", null: false
-    t.integer "blood_sugar_value", null: false
+    t.decimal "blood_sugar_value", null: false
```
kitallis: The migration says `numeric`. Does rails change that to `decimal` in the schema?

app/schema/api/current/models.rb
vkrmis: The `number` type was stale, re-swaggerized to update it

db/schema.rb
vkrmis: Yup, rails is changing this to `decimal`. This should be fine, `numeric` and `decimal` are equivalent from Postgresql's perspective. 

> The types decimal and numeric are equivalent. Both types are part of the SQL standard.

[source](https://www.postgresql.org/docs/current/datatype-numeric.html)

harimohanraj89:
@vkrmis @kitallis we can close this in favor of v4 right?

vkrmis:
> @vkrmis @kitallis we can close this in favor of v4 right?

Yup, closing this

PR #839, opened 2020-03-05
PR template

harimohanraj89:
Story: https://www.pivotaltracker.com/story/show/171644871

This PR adds a PR template to encourage adding story links and descriptions to our PRs.


kitallis:
We briefly explored commit messages like this: https://github.com/simpledotorg/simple-server/commit/2a7b1dee7a42e01e55539b75d80f5992c3e16a3e which I personally found useful (it was inspired by [this](https://thoughtbot.com/blog/better-commit-messages-with-a-gitmessage-template)). 

Do you think the `Description` template can be more specific like above? It can be roughly duplicated text from the Pivotal story as well. Let me know what you think.

harimohanraj89:
I was keeping this light to begin with. I'm happy to drop a **Because** and a **This addresses** in here if you'd like #nbd

PR #840, opened 2020-03-06
Dashboard downloads

harimohanraj89:
**Story:** https://www.pivotaltracker.com/story/show/171077165

This PR adds the ability to download district and facility cohort report charts that are seen on the dashboard as CSV files. Notable features:

* The filename and file itself contain the timestamps when the report was generated
* This PR intentionally does not introduce Excel support yet
* An existing download (overdue list) has been refactored to replace CSV generation logic in the appointment model with CSV generation logic in a view. I think this is a more fitting place for this work.


kitallis:
This looks fine to me, we should check with @timcheadle if this supersedes the [rake task](https://github.com/simpledotorg/simple-server/blob/master/lib/tasks/cohort_reports.rake)

timcheadle:
> This looks fine to me, we should check with @timcheadle if this supersedes the [rake task](https://github.com/simpledotorg/simple-server/blob/master/lib/tasks/cohort_reports.rake)

It definitely may replace that. Let's keep the rake task for now and see how this shakes out.

PR #842, opened 2020-03-06
Incorporate BloodSugar values when calculating patient risk

vkrmis:
https://www.pivotaltracker.com/story/show/164723564

The server side code base still has multiple risk categories (`High`, `Regular`, `Low`, `None`).
This PR only categorizes people with high blood sugar as `High` risk, and doesn't modify the other risk level categorization.
 
Adding a section to the tech debt document to evaluate and remove unnecessary risk levels from the server side code base [here](https://docs.google.com/document/d/15kcMseP5szCj-Bg36bPXF4lwhX_KLiW-mlt74j7adbk/edit#heading=h.drt3ksytu1jo)

high risk label in overdue


spec/models/blood_sugar_spec.rb
```diff
@@ -10,4 +10,28 @@
     it { should belong_to(:patient).optional }
     it { should belong_to(:user).optional }
   end
+
+  describe '#diabetic?' do
+    [{ blood_sugar_type: :random, blood_sugar_value: 300 },
+     { blood_sugar_type: :fasting, blood_sugar_value: 200 },
+     { blood_sugar_type: :post_prandial, blood_sugar_value: 301 }].each do |row|
```
harimohanraj89: How about `300`, to be consistent with the others?

spec/models/blood_sugar_spec.rb
```diff
@@ -10,4 +10,28 @@
...
+      it 'returns true if blood sugar is in a high state' do
+        blood_sugar = create(:blood_sugar,
+                             blood_sugar_type: row[:blood_sugar_type],
+                             blood_sugar_value: row[:blood_sugar_value])
+        expect(blood_sugar).to be_diabetic
+      end
+    end
+
+    [{ blood_sugar_type: :random, blood_sugar_value: 299 },
+     { blood_sugar_type: :fasting, blood_sugar_value: 150 },
```
harimohanraj89: How about 199 (and 299 below) to be consistent?

spec/models/blood_sugar_spec.rb
vkrmis: I thought it might better to have a slightly wider range to test, but it doesn't really matter, I'll update to the boundary values.

spec/models/blood_sugar_spec.rb
vkrmis: Same as above

PR #843, opened 2020-03-06
Progress Tab (new designs)

kitallis:
**Story card**: https://www.pivotaltracker.com/story/show/171672661

## Because

The previous progress tab was insufficient for the user in terms of: daily progress view, more interactive data lookup for the Facility, sense of progress and achievement.

## This addresses

The new progress tabs in these gifs should cover the various aspects of its usage.

### Facility with data updated today
* Achievements are visible and carousel works up until today
* In this Facility, Today's data is absent

update_today

### Facility with data updated earlier
* Achievements are visible and carousel works up until the next day after last update
* A nudge card to sync is visible

update-not-today

### Facility with no data

* Everything is zeroed out
* The achievements do not populate (needs a minimum of 10 follow-ups)

no-data

### Data

The current data population logic behaves the following way:

**Daily**
* Registrations: Both Hypertension and Diabetes patients
* Follow-ups: Only Hypertension follow-ups

**Monthly**
* Registrations: Both Hypertension and Diabetes patients filterable by gender
* Follow-ups: Only Hypertension follow-ups filterable by gender
* Control: Only Hypertension control rates

**Trophies**
* Based on Hypertension follow-ups

### Implementation notes

* All the data on the progress tab is dumped onto the page in one shot (including filters etc.). There are no AJAX calls being made.
* The data continues to be exposed as both HTML and JSON.
* The timezones for all the date work is based on the country selected, this is to ensure that the JS time in the WebView is in the same tz as the dates rendered from the server.
* There's a new materialized view: `LatestBloodPressuresPerPatientPerDay`, which refreshes a lot more often to speed up daily data population.

### Future

* ~~❗️Before merging this PR, the [translations PR](https://github.com/simpledotorg/simple-server/pull/860) should be merged in first.❗️~~ See this [note](https://github.com/simpledotorg/simple-server/pull/843#issuecomment-607138934).
* The current progress tab does not allow filtering Diabetes patients separately. This will come as a separate PR as a part of this [story](https://www.pivotaltracker.com/story/show/170179855).


app/assets/javascripts/user_analytics.js
```diff
@@ -0,0 +1,185 @@
...
+  return new Date();
+}
+
+function yesterday() {
+  var todayDate = today();
+  return new Date(todayDate.setDate(todayDate.getDate() - 1));
+}
+
+function formattedTomorrowDate() {
+  return statistics().metadata.formatted_next_date;
```
harimohanraj89: Is this also a string, like `formattedTodayString`? Can we consistent-ize the names if so?

app/assets/javascripts/user_analytics.js
```diff
@@ -0,0 +1,185 @@
...
+    dates.push(Date.parse(date));
+  }
+
+  return new Date(Math.max.apply(null, dates));
+}
+
+//
+// elements
+//
+function syncNudgeCardElement() {
```
harimohanraj89: If you'd like you can use a nifty convention of prefixing all these methods with `$` to avoid having to add `Element` to the end.

app/assets/javascripts/user_analytics.js
```diff
@@ -0,0 +1,185 @@
...
+  return document.getElementsByClassName("button-next")[0];
+}
+
+function prevSlideButtonElement() {
+  return document.getElementsByClassName("button-prev")[0];
+}
+
+function syncNudgeCardPosition() {
+  for (let [position, element] of Object.entries(allDaysInCarouselElements())) {
+    if (element === syncNudgeCardElement()) {
```
harimohanraj89: Maybe we should get the `syncNudgeCardElement` just once at the beginning of this method?

app/assets/javascripts/user_analytics.js
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  return -1
+}
+
+//
+// behaviour
+//
+function showSyncNudge(currentSlide) {
+  var weHaveDataForToday = false;
```
harimohanraj89: I appreciate the verbosity 😅 but perhaps we can go with something like `todaysDataPresent`?

app/assets/javascripts/user_analytics.js
```diff
@@ -0,0 +1,185 @@
...
+  var tableElements = document.getElementsByClassName('progress-table ' + tableName);
+  var tableFilterElement = document.getElementsByClassName('card-dropdown ' + tableName);
+  var selectedOption = tableFilterElement[0].selectedOptions[0].value;
+  var selectedTableElement = document.getElementsByClassName('progress-table ' + tableName + ' ' + selectedOption);
+
+  for (let i = 0; i < tableElements.length; i++) {
+    tableElements[i].style.display = 'none';
+  }
+
+  selectedTableElement[0].style.display = 'inline-table';
```
harimohanraj89: This hides all first, then shows the selected one, is that right?

app/controllers/api/v3/analytics/user_analytics_controller.rb
```diff
@@ -1,21 +1,34 @@
 class Api::V3::Analytics::UserAnalyticsController < Api::V3::AnalyticsController
+  include HashUtilities
+  include ApplicationHelper
   include DashboardHelper
+  include MonthHelper
+  include DayHelper
+  include PeriodHelper
```
harimohanraj89: I see a lot of great new helpers in here. Could we add the necessary test coverage for the new helpers, and for the modifications to existing helpers as well?

app/queries/my_facilities/follow_ups_query.rb
```diff
@@ -0,0 +1,32 @@
...
+  # Example: `Time.use_zone('timezone string') { bp_control_query_object.cohort_registrations }`
+
+  include QuarterHelper
+  include MonthHelper
+  include DayHelper
+  include PeriodHelper
+
+  attr_reader :periods
+
+  def self.total_follow_ups(facilities)
```
harimohanraj89: Even though this doesn't use all the initialization args, I think this should still be an instance method.

app/queries/my_facilities/follow_ups_query.rb
```diff
@@ -0,0 +1,32 @@
...
+  def initialize(facilities: Facility.all, period: :quarter, last_n: 3)
+    # period can be :quarter, :month, :day.
+    # last_n is the number of quarters/months/days data to be returned
+    @facilities = facilities
+    @period = period
+    @periods = period_list(period, last_n)
+  end
+
+  def follow_ups
+    MyFacilities::FollowUpsQuery
```
harimohanraj89: `self.class`?

app/views/api/v2/analytics/user_analytics/_monthly_gender_table.html.erb
```diff
@@ -0,0 +1,48 @@
...
+  </tr>
+  <% @monthly_period_list.each do |month_date| %>
+    <tr>
+      <td class="row-label no-padding"><%= month_date.strftime("%B") %></td>
+      <td>
+        <%= zero_if_unavailable(@statistics
+                                    .dig(:monthly, :grouped_by_gender_and_date, resource)
+                                    .values
+                                    .inject { |acc, next_item| acc.merge(next_item) { |_, v1, v2| v1 + v2 } }
+                                    &.dig(month_date)) %>
```
harimohanraj89: There's a lot going on here. How can we simplify this for fellow devs?

kitallis:
> This is a first round of review. I'm a bit saturated from review as well, so unfortunately I think I'll have to give you a second round next week as well.

Thanks @harimohanraj89! This is helpful. Second round next week works!

db/views/latest_blood_pressures_per_patient_per_days_v01.sql
```diff
@@ -0,0 +1,23 @@
...
+        cast(EXTRACT(QUARTER FROM blood_pressures.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS quarter,
+        cast(EXTRACT(YEAR FROM blood_pressures.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS year
+    FROM blood_pressures JOIN patients ON patients.id = blood_pressures.patient_id
+    WHERE blood_pressures.deleted_at IS NULL
+    ORDER BY patient_id, year, month, blood_pressures.recorded_at DESC, bp_id
+)
+SELECT latest_bp_per_patient_per_month.*,
+       LAG(bp_facility_id, 1) OVER (PARTITION BY patient_id ORDER BY bp_recorded_at ASC, bp_id) AS responsible_facility_id,
+       LAG(bp_id, 1) OVER (PARTITION BY patient_id ORDER BY bp_recorded_at ASC, bp_id) AS previous_bp_id
+FROM latest_bp_per_patient_per_month;
```
prabhanshuguptagit: Looks like this view is being calculated incorrectly, there should be a `DISTINCT ON day` as well.

app/queries/my_facilities/registrations_query.rb
```diff
@@ -22,26 +22,28 @@ def initialize(facilities: Facility.all, period: :quarter, last_n: 3)
   def registrations
     @registrations ||=
       PatientRegistrationsPerDayPerFacility
-      .where(facility: @facilities)
-      .where("(year, #{@period}) IN (#{periods_as_sql_list})")
+        .where(facility: @facilities)
+        .where("(year, #{@period}) IN (#{periods_as_sql_list})")
```
prabhanshuguptagit: Codeclimate would be angry.

app/queries/my_facilities/registrations_query.rb
```diff
@@ -22,26 +22,28 @@ def initialize(facilities: Facility.all, period: :quarter, last_n: 3)
...
       LatestBloodPressuresPerPatient
-      .where(registration_facility_id: @facilities)
-      .where('patient_recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
+        .where(registration_facility_id: @facilities)
+        .where('patient_recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
   end
 
   private
 
   def period_list(period, last_n)
```
prabhanshuguptagit: Can be removed now that `PeriodHelper` is a thing.

db/views/latest_blood_pressures_per_patient_per_days_v01.sql
kitallis: This a good catch, thanks @prabhanshuguptagit!

app/queries/my_facilities/registrations_query.rb
kitallis: I tried. I couldn't because the `PeriodHelper` in `RegistrationsQuery` is not inclusive for the `day`

app/assets/javascripts/user_analytics.js
kitallis: That's correct.

app/queries/my_facilities/follow_ups_query.rb
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/843/commits/df92264398c30997e9fe4571b9ac938becce08ff.

app/queries/my_facilities/follow_ups_query.rb
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/843/commits/df92264398c30997e9fe4571b9ac938becce08ff.

app/views/api/v2/analytics/user_analytics/_monthly_gender_table.html.erb
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/843/commits/df92264398c30997e9fe4571b9ac938becce08ff.

app/views/api/v2/analytics/user_analytics/_gender_filter.html.erb
```diff
@@ -0,0 +1,22 @@
+<form>
+  <select class="card-dropdown <%= data_table_name %>" onchange='filterDataByGender("<%= data_table_name %>")'>
+    <!--
+     - We don't need "Hypertension: All" unless DM is enabled.
```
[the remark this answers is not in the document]
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/843/commits/df92264398c30997e9fe4571b9ac938becce08ff.

app/views/api/v2/analytics/user_analytics/show.html.erb
```diff
@@ -1,48 +1,189 @@
...
 <% end %>
+
+<!-- footer -->
+<footer>
+  <h4 style="padding-top: 60px;">Notes</h4>
+  <p>Data is for all activity at <%= @current_facility.name %>. If data appears out of date, tap the "Sync
+    pending" link on the home screen.</p>
+  <p>A follow-up patient is only counted for one visit each month. If a patient follows up multiple times in a month,
+    he or she is
+    only counted once.</p>
```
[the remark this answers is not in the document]
kitallis: I think it's just there for limiting under 80 chars.

app/views/api/v3/analytics/user_analytics/show.html.erb
```diff
@@ -1,48 +1,189 @@
...
+2C. Patient counts are for 'unique' patients that month (e.g. if a patient visited 5 times, they are only counted once... unless this is hard, in which case discuss with Daniel)
+2D. Patients with controlled BP is if the patient had BP <140/90 at most recent visit
+
+3. I think registrations is pretty obvious, but if you have questions, let me know.
+
+4A. Achievements is based on data from 2C.
+4B. Don't show any achievements if < 10. Don't show the locked "upcoming" one either
+4C. When > 10 show first badge and locked "upcoming" badge for 25 follow-up visits
+4D. Follow the rules in the achievements html for how to count. After "10,000" just increment by 10,000 increments for infinity
+-->
```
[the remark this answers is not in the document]
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/843/commits/df92264398c30997e9fe4571b9ac938becce08ff.

app/views/api/v3/analytics/user_analytics/show.html.erb
```diff
@@ -1,48 +1,189 @@
...
+          <%= zero_if_unavailable(@statistics.dig(:monthly, :grouped_by_date, :total_visits, month_date)) %>
+        </td>
+      </tr>
+    <% end %>
+  </table>
+</div>
+
+<!-- all trophies view -->
+<!-- only show achievements if follow_ups > 10 -->
+<% unless @statistics.dig(:trophies, :locked_trophy_value) == 10 %>
```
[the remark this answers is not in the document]
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/843/commits/df92264398c30997e9fe4571b9ac938becce08ff.

app/controllers/api/v3/analytics/user_analytics_controller.rb
```diff
@@ -30,32 +43,132 @@ def respond_to_html_or_json(stats)
...
+  # { :grouped_by_gender =>
+  #    { "male" =>
+  #        { "Sun, 01 Mar 2020" => 21 },
+  #          "Tue, 01 Oct 2019" => 0 },
+  #      "female" =>
+  #        { "Sun, 01 Mar 2020" => 18,
+  #          "Tue, 01 Oct 2019" => 0 } } }
+  #
+  def group_by_gender_and_date(resource_data)
+    resource_data.inject(autovivified_hash) do |acc, (group, resource)|
```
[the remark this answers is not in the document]
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/843/commits/df92264398c30997e9fe4571b9ac938becce08ff.

app/assets/javascripts/user_analytics.js
kitallis: I don't like pluralizing in variable names. Regardless, I've changed it to another name.

Addressed JS things here: https://github.com/simpledotorg/simple-server/pull/843/commits/ad4d217691c22b825a7a76e43af3099e9be2fa03. This includes moving to ES6 best practices (no usage of `var`s anywhere).

app/assets/javascripts/user_analytics.js
kitallis: Addressed JS things here: https://github.com/simpledotorg/simple-server/pull/843/commits/ad4d217691c22b825a7a76e43af3099e9be2fa03. This includes moving to ES6 best practices (no usage of `var`s anywhere).

app/assets/javascripts/user_analytics.js
kitallis: Addressed JS things here: https://github.com/simpledotorg/simple-server/pull/843/commits/ad4d217691c22b825a7a76e43af3099e9be2fa03. This includes moving to ES6 best practices (no usage of `var`s anywhere).

app/assets/javascripts/user_analytics.js
kitallis: Addressed JS things here: https://github.com/simpledotorg/simple-server/pull/843/commits/ad4d217691c22b825a7a76e43af3099e9be2fa03. This includes moving to ES6 best practices (no usage of `var`s anywhere).

config/schedule.rb
```diff
@@ -26,6 +26,6 @@ def local(time)
   rake 'generate:seed[1]'
 end
 
-every :day, at: local('03:00 am'), roles: [:cron] do
+every 3.hours, roles: [:cron] do
```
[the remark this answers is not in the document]
kitallis: We did this so that we can get more frequent refreshes for data and the daily data for today especially can be populated a lot more frequently.

app/controllers/api/v3/analytics_controller.rb
```diff
@@ -1,14 +1,18 @@
...
 
-  def set_timezone_to_IST
-    Groupdate.time_zone = "Asia/Kolkata"
-  end
+  def set_time_zone
+    time_zone = Rails.application.config.country[:time_zone] || DEFAULT_ANALYTICS_TIME_ZONE
+
+    Groupdate.time_zone = time_zone
+
+    Time.use_zone(time_zone) do
```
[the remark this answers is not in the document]
kitallis: It is intentional. The reason why I've added this is because tz are more important now since dates are *also* being generated in JS. For a fair comparison of JS dates and dates generated server-side, it's valuable to generate server-side dates in the best guess timezone for the particular country where the progress report is being rendered.

Now of course, there's exceptions to this, like you could be a nurse from Punjab, entering patients in New York. But that should be rare and minimal.

app/controllers/api/v3/analytics/user_analytics_controller.rb
```diff
@@ -30,32 +43,132 @@ def respond_to_html_or_json(stats)
     end
   end
 
-  def first_of_current_month
-    Date.current.at_beginning_of_month
+  def prepare_daily_stats
```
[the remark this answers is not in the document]
kitallis: I thought about it, but it felt overkill since `UserAnalytics` is a relatively cordoned-off aspect of the app. We have a lot of "abstractions" now. Query objects, database models, materialized views – it felt too much to add another layer.

Happy to stick a `Presenter` class in if you feel the controller is too heavy.

app/controllers/api/v3/analytics_controller.rb
```diff
@@ -1,14 +1,18 @@
...
+      yield
+    end
 
-  def set_timezone_to_UTC
+    # Make sure we reset the timezone
     Groupdate.time_zone = "UTC"
```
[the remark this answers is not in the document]
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/843/commits/df92264398c30997e9fe4571b9ac938becce08ff.

app/queries/user_analytics_query.rb
```diff
@@ -0,0 +1,94 @@
+class UserAnalyticsQuery
+  include DayHelper
+  include MonthHelper
+  include PeriodHelper
+
+  attr_reader :current_facility, :months_ago, :days_ago
```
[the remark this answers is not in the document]
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/843/commits/ccac04adba0edc9b0a126ef04e62ed5033d97bd6.

app/assets/javascripts/user_analytics.js
```diff
@@ -0,0 +1,182 @@
...
+function formatDate(date) {
+  return date.toJSON().slice(0, 10);
+}
+
+function today() {
+  return new Date();
+}
+
+function yesterday() {
+  const todayDate = today();
```
harimohanraj89: Is the ES6 syntax ok? Is it transpiled to ES5? Will apps support this in their webview?

app/controllers/api/v3/analytics/user_analytics_controller.rb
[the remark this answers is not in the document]
harimohanraj89: Yeah the controller does seem quite heavy. I'm still leaning towards putting all this stuff into a separate presenter that can handle these methods, as well as any data massaging logic that's still in the view.

app/controllers/api/v3/analytics_controller.rb
```diff
@@ -1,14 +1,18 @@
...
+    time_zone = Rails.application.config.country[:time_zone] || DEFAULT_ANALYTICS_TIME_ZONE
+
+    Groupdate.time_zone = time_zone
+    Time.use_zone(time_zone) do
+      yield
+    end
 
-  def set_timezone_to_UTC
+  ensure
+    # Make sure we reset the timezone
```
harimohanraj89: We can omit this comment, it's pretty self-explanatory.

app/views/api/v3/analytics/user_analytics/_monthly_gender_table.html.erb
```diff
@@ -0,0 +1,49 @@
+<!-- the "all" table -->
+<table class="progress-table <%= data_table_name %> hypertension:all">
+  <tr class="total">
+    <td class="row-label">All-time</td>
+    <td>
+      <%= zero_if_unavailable(@statistics
+                                  .dig(:all_time, :grouped_by_gender, resource)
+                                  .values
+                                  .inject(:+)) %>
```
harimohanraj89: `.sum`? PS, this sort of logic can also be extracted to a presenter, keeping the views very simple:

`<%= all_time_count %>` or whatever.

app/assets/javascripts/user_analytics.js
```diff
@@ -0,0 +1,182 @@
...
+//
+window.onload = function () {
+  window.lastSlidePositionForProgressCards = 1;
+  refreshCarousel(window.lastSlidePositionForProgressCards)
+};
+
+//
+// on-click events
+//
+function nextSlide(increment) {
```
harimohanraj89: Is this function ever called with an increment other than +1 or -1? Could we just make a `function nextSlide()` and `function prevSilde()`?

harimohanraj89:
Additionally, could you point me to where we're generating and handing off the JSON blob of data to the JS-powered web page?

app/queries/my_facilities/registrations_query.rb
prabhanshuguptagit: Could we make the helper `period_list` take `inclusive` as a key argument (defaults to true)? That could make this work.

kitallis:
> Additionally, could you point me to where we're generating and handing off the JSON blob of data to the JS-powered web page?

It's in a [hidden content div](https://github.com/simpledotorg/simple-server/pull/843/files#diff-1695aa4fd0071ee42c25d34186e2a78aR29).

app/queries/my_facilities/registrations_query.rb
kitallis: I could, but then I think `RegistrationsQuery` wants inclusive to be true for `month` and `quarter` but not `day`. So that parameter will not be precise.

app/views/api/v2/analytics/user_analytics/_gender_filter.html.erb
```diff
@@ -0,0 +1,22 @@
...
+     - We don't need the word "Hypertension" at all unless DM is enabled.
+    -->
+
+    <% hypertension_option_text = statistics.dig(:metadata, :is_diabetes_enabled) ? 'Hypertension: ' : '' %>
+
+    <option value="hypertension:all"><%= "#{hypertension_option_text}All" %></option>
+    <option value="hypertension:female"><%= "#{hypertension_option_text}Female" %></option>
+    <option value="hypertension:male"><%= "#{hypertension_option_text}Male" %></option>
+    <option value="hypertension:transgender"><%= "#{hypertension_option_text}Transgender" %></option>
```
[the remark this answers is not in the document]
kitallis: Progress tab will go through a few iterations. Let's punt on this until we reach towards the end.

app/views/api/v3/analytics/user_analytics/_monthly_gender_table.html.erb
kitallis: I have made some adjustments to this. More consolidated notes coming up.

app/assets/javascripts/user_analytics.js
kitallis: I'm not sure I agree. I find `nextSlide` sufficiently simple and DRY.

app/assets/javascripts/user_analytics.js
kitallis: As I understand it `> API 21`  supports ES6. I will confirm this with the mobile team.

kitallis:
@harimohanraj89 @prabhanshuguptagit @vkrmis 

**New consolidated notes on a few implementation changes (both PR feedback and otherwise)**:
1. The controller bloat is now moved into a [Presenter](https://github.com/simpledotorg/simple-server/pull/843/commits/7ac1eef21d3fc0bb2f617e136c557dd4c2867c93) (along with [specs](https://github.com/simpledotorg/simple-server/pull/843/commits/dcaceb9e29351779a3f48d6a3756ff4813586b86))
2. Some existing Helper specs are [here](https://github.com/simpledotorg/simple-server/pull/843/commits/0299dd780e3d9c45f38d45b2ea9f43176a9bf190) and [here](https://github.com/simpledotorg/simple-server/pull/843/commits/928e4a06d6b536e742bdd0b5fc1ebdd2a454ed74).
3. The whole progress tab is now [feature-toggle'd](https://github.com/simpledotorg/simple-server/pull/843/commits/16c0a56f779b7e03586497985fd67581837396e5). We can safely merge this branch after review. It requires pushing this [config](https://github.com/simpledotorg/simple-server/pull/843/commits/f2c7be993db8617d693eceb64a5f10ab2689269a) in.
4. The HTN control rate is now correctly [calculated](https://github.com/simpledotorg/simple-server/pull/843/commits/c6b3dd845b018a61351c3f13fb0fb45e97574d59).
5. Some basic query [optimizations](https://github.com/simpledotorg/simple-server/pull/843/commits/c54d77fa5b776a36dd3635c226f0eefc42e935b1) have been made.
6. The [LatestBloodPressuresPerPatientPerDay](https://github.com/simpledotorg/simple-server/pull/843/commits/a2eb5bd14abab46a39cea33d10c3f4a87aef36fa) now properly unique's patients by day.
7. The monthly queries were not taking into account the new data for the current month, that fix has been made [here](https://github.com/simpledotorg/simple-server/pull/843/commits/f79b1f248d44f69ecb3bad61dd3363f999a036b3).
8. Moved some complex [calculations](https://github.com/simpledotorg/simple-server/pull/843/commits/7ac1eef21d3fc0bb2f617e136c557dd4c2867c93#diff-e9e2bbe1c12d40d2ea6f7a9d383a866aR37) from the view into the `UserAnalyticsPresenter`. However, a majority of the data is still parsed directly. This is an intentional choice as the data-structure is simple and obvious enough that I didn't feel like every single call needed to be abstracted over a function name. The remaining ones are simply picking values off of a map and no calculation work is done in the view.

**Misc**:
* The `last_updated_at` we show on the top of the progress tabs is not very reliable, since it tells you the truth about the page refresh rather than the data refresh. The challenge with this is that there are queries that come from Materialized Views (which have their own refresh cadence) and some directly from the table. I have a strategy on how we can handle this and make it more precise but I'd like to do that as another PR.
* We don't have to wait for the translations to arrive before we merge this in
   * We can incrementally release translations as they come and default to English
   * The whole thing can just be feature toggle'd off now, if required

app/presenters/user_analytics_presenter.rb
```diff
@@ -0,0 +1,163 @@
...
+      metadata: {
+        is_diabetes_enabled: false,
+        last_updated_at: I18n.l(Time.current),
+        formatted_next_date: display_date(Time.current + 1.day),
+        today_string: I18n.t(:today_str)
+      }
+    }
+  end
+
+  def stats_across_genders_for_month(resource, month_date)
```
harimohanraj89, line 37: This was my biggest concern in the view. Nice to see it here now 👍

app/assets/stylesheets/legacy/user_analytics.scss
```diff
@@ -0,0 +1,148 @@
+.text-muted {
+  color: #ADB2B8;
+}
+
+.bar-chart {
+  clear: both;
+  //overflow: scroll;
+  //margin-left: -16px;
```
harimohanraj89, line 8: We can remove these comments.

app/controllers/api/v3/analytics/user_analytics_controller.rb
```diff
@@ -1,23 +1,14 @@
 class Api::V3::Analytics::UserAnalyticsController < Api::V3::AnalyticsController
+  include ApplicationHelper
   include DashboardHelper
-  layout false
 
-  MONTHS_TO_REPORT = 6
```
harimohanraj89, line 5: This is no longer required for the legacy page?

app/helpers/application_helper.rb
```diff
@@ -21,6 +21,10 @@ def display_date(date)
     date.strftime(STANDARD_DATE_DISPLAY_FORMAT)
   end
 
+  def display_singular_or_plural_str(pluralized_string, count)
+    count == 1 ? pluralized_string.singularize : pluralized_string
+  end
```
harimohanraj89, line 26: Can we use [Rails's pluralization](https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize) instead?

app/queries/user_analytics_query.rb
```diff
@@ -0,0 +1,97 @@
...
+      .group(:year, :day)
+      .count
+      .map { |group, count| [doy_to_date(*group), count] }
+      .to_h
+  end
+
+  def daily_registrations
+    group_by_time_range = (days_ago.to_i - 1).days.ago.beginning_of_day..Date.current.end_of_day
+
+    current_facility
```
harimohanraj89, line 32: Do we have a `MyFacilities::Registrations` query that can give us this number? Should we use it for consistency?

harimohanraj89:
I think the last round of comments ☝️

app/helpers/application_helper.rb
kitallis: Wow, I swear I searched for this...

app/queries/user_analytics_query.rb
kitallis: Not quite. The backing view that it uses primarily has this issue:

It counts up (aggregates) the `registration_count` against the day/month/quarter and hence cannot expose the individual `patient_id`. We want a flatter access to the Patient data so that we can group by `gender` later. I considered making that itself a view, but that would be redundant since it'll be the same amount of rows as just making a normal query to the database; potentially even *less* performant that a normal query.

PR #844, opened 2020-03-09
Address feedback for Dashboards V1

vkrmis:
* Discount patients without BPs for missed visits totals calculation
* Fix broken sort on registrations all-time
* Add/Update footnotes
* Allow sorting for inactive facilities table
* Fix broken percentage rounding

https://www.pivotaltracker.com/story/show/171677168


app/helpers/dashboard_helper.rb
```diff
@@ -44,6 +44,6 @@ def percentage_string(percentage)
   def percentage(numerator, denominator)
     return percentage_string(0) if denominator.zero?
 
-    percentage_string((numerator * 100) / denominator)
+    percentage_string((numerator * 100.0) / denominator)
```
harimohanraj89: Now that this is a floating point number, do we need to specify a precision to ensure we don't generate strings like "8.33333333%"?

app/views/my_facilities/registrations.html.erb
```diff
@@ -94,15 +94,19 @@
...
           </tr>
         <% end %>
         </tbody>
       </table>
     </div>
     <section class="footnotes">
       <ol>
-        <li><strong>All registered patients:</strong> All registered patients with exclusions for: lost to follow-up, transferred to another public facility, and died.</li>
+        <li>
+          <strong>All time registered patients:</strong> All registered patients with at least one recorded blood pressure.
```
harimohanraj89: "All-time" or "All time"?

app/views/shared/my_facilities/_facilities_by_size.html.erb
```diff
@@ -43,5 +43,9 @@
       </tr>
       </tbody>
     </table>
+    <ol style="padding-left: 15px"><li>
+      <strong>Inactive:</strong>
+      Facilities with <10 patients with BPs recorded in the last 7 days.
```
harimohanraj89: We should escape this `<` with `&lt;`

app/views/shared/my_facilities/_facilities_by_size.html.erb
```diff
@@ -12,7 +12,7 @@
       <tr>
         <th></th>
         <th>Total</th>
-        <th>Inactive*</th>
+        <th>Inactive<sup>1</sup></th>
```
harimohanraj89: Sup

app/helpers/dashboard_helper.rb
vkrmis: The `percentage_string` function has a `.round(0)` that will handle this. It was getting an integer input earlier 🤦‍♂

app/views/my_facilities/registrations.html.erb
vkrmis: I'll change this to "All-time"

vkrmis:
> Minor feedback. Looks great overall. If you could add some screenshots to the PR description that describe your changes, that would be 💯

There aren't any major visual changes @harimohanraj89, I'm not sure which screenshots would be helpful here.

harimohanraj89:
Screenshots of the fixes you've made - aka the reordering columns, the footnote, the sort working, etc.

It's a nice way to provide context to reviewers, and a cool forcing function to make sure we test stuff :) This PR is just about good to go, though, so don't sweat it on this one.

app/helpers/dashboard_helper.rb
harimohanraj89: Understood. Was the "off-by-1-percent" a real issue?

PR #845, opened 2020-03-10
Create a V4 api for Blood Sugars

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/171377632

## Because

Adding support for `hba1c` blood sugars is a breaking change for older clients.
The `blood_sugar_value` field will now need to support decimal values.

## This addresses

This PR adds a new v4 api version for Blood Sugars. This ensures that older app versions will continue to work without disruption, and new app versions will be able to record and view `hba1c` Blood Sugars.

## Notes

This PR also does away with the notion of a `current` API version, and therefore breaks the `ApiVersionGenerator`. This generator only works on a full API basis, and doesn't support upgrading a single endpoint at a time. 

~~This PR patches `BigDecimal#as_json` to return a float due to this: https://github.com/rails/rails/issues/25017. 
Is there a better way to handle this?~~


app/controllers/api/v4/blood_sugars_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Api::V4::BloodSugarsController < Api::V3::BloodSugarsController
```
harimohanraj89: I understand this is pulling a lot of existing functionality from its v3 counterpart. However, for consistency, it appears this should inherit from an `Api::V4::SyncController`. How should we handle this?

app/controllers/api/v4/blood_sugars_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def merge_if_valid(blood_sugar_params)
+    validator = Api::V4::BloodSugarPayloadValidator.new(blood_sugar_params)
+    logger.debug "Blood Sugar payload had errors: #{validator.errors_hash}" if validator.invalid?
+    if validator.invalid?
+      NewRelic::Agent.increment_metric('Merge/BloodSugar/schema_invalid')
+      { errors_hash: validator.errors_hash }
+    else
+      set_patient_recorded_at(blood_sugar_params)
+      transformed_params = Api::V3::Transformer.from_request(blood_sugar_params)
```
harimohanraj89: Can we make a v4 version of this transformer for use here?

harimohanraj89:
PS, do we have to explicitly enforce blood sugar values as integers in the V3 API?

app/models/blood_sugar.rb
```diff
@@ -15,13 +15,17 @@ class BloodSugar < ApplicationRecord
   enum blood_sugar_type: {
     random: 'random',
     post_prandial: 'post_prandial',
-    fasting: 'fasting'
+    fasting: 'fasting',
+    hba1c: 'hba1c'
   }, _prefix: true
 
+  scope :without_hba1c, -> { where.not(blood_sugar_type: :hba1c) }
```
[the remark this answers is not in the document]
vkrmis: We're unlikely to add a fifth BP type anytime soon afaik, but we can probably switch this to a whitelist if/when that happens.

vkrmis:
> PS, do we have to explicitly enforce blood sugar values as integers in the V3 API?

Yes, I'll do this in the transformer as well

app/controllers/api/v4/blood_sugars_controller.rb
vkrmis: I can add a V4 SyncController that just inherits from the V3 SyncController, we'll end up with some duplication between the V3 and V4 BloodSugarsControllers

app/models/blood_sugar.rb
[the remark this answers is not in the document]
harimohanraj89: My concern is that we'll forget. Let's make this future-proof now, while this stuff is fresh in our mind. What do you think?

app/transformers/api/current/prescription_drug_transformer.rb
```diff
@@ -1,11 +0,0 @@
...
-  class << self
-    def to_response(prescription_drug)
-      Api::Current::Transformer.to_response(prescription_drug).except('user_id')
-    end
-
-    def from_request(prescription_drug_payload)
-      Api::Current::Transformer.from_request(prescription_drug_payload)
-    end
-  end
-end
```
harimohanraj89: Why did we remove this guy?

config/initializers/rswag-ui.rb
```diff
@@ -5,6 +5,7 @@
   # JSON endpoint and the second is a title that will be displayed in the document selector
   # NOTE: If you're using rspec-api to expose Swagger files (under swagger_root) as JSON endpoints,
   # then the list below should correspond to the relative paths for those endpoints
-  c.swagger_endpoint '/api-docs/current/swagger.json', 'API V2 Docs'
```
harimohanraj89: Why remove V2?

config/routes.rb
```diff
@@ -129,6 +129,13 @@
         resource :user_analytics, only: [:show]
       end
     end
+
+    namespace :v4, path: 'v4' do
+      scope '/blood_sugars' do
```
harimohanraj89: We should be able to do `scope :blood_sugars` right?

spec/controllers/api/v4/blood_sugars_controller_spec.rb
```diff
@@ -103,6 +103,19 @@ def create_record_list(n, options = {})
...
+      context 'hba1c blood sugars' do
+        it 'successfully records hba1c blood sugars' do
+          blood_sugar = build_blood_sugar_payload(build(:blood_sugar, blood_sugar_type: :hba1c))
+
+          post(:sync_from_user, params: { blood_sugars: [blood_sugar] }, as: :json)
+          errors = JSON(response.body)['errors']
+
+          expect(response).to have_http_status(200)
+          expect(errors).to eq([])
+          expect(BloodSugar.where(id: blood_sugar['id']).count).to eq 1
```
harimohanraj89: Let's assert the right value got in too, since we're dealing with a lot of types.

spec/lib/generators/api_version/api_version_generator_spec.rb
```diff
@@ -26,7 +26,7 @@ def files_in_directory(directory, extension: '*.rb')
       .map { |file| file.to_s.remove(Rails.root.to_s) }
   end
 
-  describe 'generates the scaffold required to migrate to a new API version' do
+  xdescribe 'generates the scaffold required to migrate to a new API version' do
```
harimohanraj89: 😄

spec/requests/shared_examples/api/v4/shared_spec_for_sync_requests.rb
```diff
@@ -0,0 +1,161 @@
+require 'rails_helper'
```
harimohanraj89: Are these the legendary shared sync specs?

harimohanraj89:
Left another round of review on your PR ☝️Overall, the structural changes you made (setting V4 base classes, etc) look great :ok_hand:

app/controllers/api/v4/blood_sugars_controller.rb
harimohanraj89: I think that's fine 👌 If this becomes a problem tomorrow, we can extract the commonalities into modules as appropriate 🤷‍♂

config/initializers/rswag-ui.rb
vkrmis: I think the earlier text was a typo, it actually has the `current` link which I updated to V3

app/models/blood_sugar.rb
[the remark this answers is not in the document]
vkrmis: Cool, I'll call the scope `api_v3_types`?

spec/lib/generators/api_version/api_version_generator_spec.rb
vkrmis: Yup, this thing is totally broken now 😄

spec/requests/shared_examples/api/v4/shared_spec_for_sync_requests.rb
vkrmis: _some_ of them

config/initializers/rswag-ui.rb
vkrmis: Also, I'm not sure what this doing exactly, the V2 docs do still show up under /api-docs

harimohanraj89: Meh, ok 😅

spec/requests/shared_examples/api/v4/shared_spec_for_sync_requests.rb
harimohanraj89: My god

spec/models/blood_sugar_spec.rb
```diff
@@ -36,4 +36,25 @@
...
+    let!(:hba1c) { create(:blood_sugar, blood_sugar_type: :hba1c) }
+
+    context '#for_v3' do
+      it 'only includes non hba1c blood sugars' do
+        expect(BloodSugar.for_v3).not_to include(hba1c)
+        expect(BloodSugar.for_v3.count).to eq 3
+      end
+    end
+
+    context '#all' do
```
harimohanraj89: I'd be ok skipping this context. We're basically testing ActiveRecord in this block. I'll leave it up to you though! 👍

app/schema/api/v3/models.rb
```diff
@@ -154,8 +154,8 @@ def blood_sugar
       { type: :object,
         properties: {
           id: { '$ref' => '#/definitions/uuid' },
-          blood_sugar_type: { type: :string, enum: BloodSugar.blood_sugar_types.keys },
-          blood_sugar_value: { type: :integer },
+          blood_sugar_type: { type: :string, enum: BloodSugar.blood_sugar_types.keys.reject { |type| type == 'hba1c' } },
+          blood_sugar_value: { type: :number },
```
harimohanraj89: Should this go back to `integer` now?  Technically a datatype change could be a breaking change, right?

vkrmis: Yes, adding a spec for this

app/schema/api/v3/models.rb
```diff
@@ -154,8 +154,8 @@ def blood_sugar
       { type: :object,
         properties: {
           id: { '$ref' => '#/definitions/uuid' },
-          blood_sugar_type: { type: :string, enum: BloodSugar.blood_sugar_types.keys.reject { |type| type == 'hba1c' } },
-          blood_sugar_value: { type: :number },
+          blood_sugar_type: { type: :string, enum: BloodSugar::V3_TYPES },
```
harimohanraj89: ❤️

PR #846, opened 2020-03-11
Upgrade the move_user_recorded_data script to handle BusinessIdentifier metadata, Blood Sugars and Encounters

vkrmis:
**Story card:** https://www.pivotaltracker.com/n/projects/2184102/stories/171743446

## Because

The existing `move_user_recorded_data_to_registration_facility` is now stale, and doesn't handle `BloodSugars` and `Encounters`. Additionally, the metadata for `PatientBusinessIdentifier` also has a facility field that was missed out previously. 

## This addresses

This PR adds `PatientBusinessIdentifier`. `BloodSugar`, and `Encounter` handling to the `move_user_recorded_data_to_registration_facility` script

### DIFF
--- a/lib/tasks/scripts/move_user_recorded_data_to_registration_facility.rb
+++ b/lib/tasks/scripts/move_user_recorded_data_to_registration_facility.rb
@@ -9,15 +9,11 @@ class MoveUserRecordedDataToRegistrationFacility
   end
 
   def fix_patient_data
+    patients = Patient.where(registration_user: user, registration_facility: source_facility)
+    fix_pbi_metadata(patients)
     fix_data_for_relation(
-      Patient.where(registration_user: user, registration_facility: source_facility),
+      patients,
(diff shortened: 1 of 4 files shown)


lib/tasks/scripts/move_user_recorded_data_to_registration_facility.rb
```diff
@@ -9,17 +9,13 @@ def initialize(user, source_facility, destination_facility)
   end
 
   def fix_patient_data
+    patients = Patient.where(registration_user: user, registration_facility: source_facility)
+    fix_pbi_metadata(patients)
```
prabhanshuguptagit: Should we call `fix_pbi_metadata` from the rake task itself instead of coupling it with `fix_patient_data`?

prabhanshuguptagit: No

PR #848, opened 2020-03-12
My facilities permissioning

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/171752073

## Because

We want to grant CVHOs access to My Facilities one by one.

## This addresses

Removes the use of the My Facilities feature flag, and instead relies solely on the provisioning of the `view_my_facilities` permission. No users will be pre-populated with this permission.

The permission will also _not_ be added to any of the role presets to ensure that all existing users' roles are preserved in the Permissioning UI. Eg. Super admins will remain super admins, and will remain able to invite new super admins to Simple.

This PR also addresses a couple of bugs found along the way:
* Inactive facilities were not scoped correctly on the overview page - a523025
* Facilities without zones were not scoped correctly on the BP Control page - a665565

**Post-deploy**

- [ ] Remove `ENABLE_VIEW_MY_FACILITIES` from all environment configs.


app/policies/permissions.rb
```diff
@@ -121,7 +121,6 @@ module Permissions
         view_health_worker_activity
         download_patient_line_list
         manage_admins
-        view_my_facilities
```
vkrmis: Shouldn't these permissions remain for STSs, CVHOs and Owners?
Is the idea that we'll do this later, after piloting the first few CVHOs?

harimohanraj89: Yup exactly. I think it's best to add these in when we decide to launch My Facilities 100%. Otherwise, existing CVHOs won't be able to invite other CVHOs, for example (same issue as superadmins, just on a different level).

vkrmis: Right, this is less broken, but it could still be confusing to the CVHOs who are piloting this. Might need to do some hand-holding.

vkrmis:
We should remove `ENABLE_VIEW_MY_FACILITIES` from `.env.development` in this PR.

app/policies/permissions.rb
harimohanraj89: I agree. We're planning to send this to three CVHOs to start, so we can speak with them directly to explain why their role is no longer "CVHO" for a while, if that's something they notice or care about. 👍

harimohanraj89: Some thoughts for how we can mitigate this confusion for all future "dark launches". 
* We could start tracking a static snapshot of a user's role against the user record. This will also help us do things like adding a new permission to CVHOs, and then refreshing all CVHO users to auto-add the new permission
* We could add an "optional" list of permissions to these roles. Optional permission will not be added to a new invitee when you select a role from the dropdown, but their presence on an existing user won't affect the role displayed for them. Eg. CVHOs with and without the My Facilities permission would show as "CVHO" in our UI.

Just planting some longer-term thoughts here while they're in my brain.

PR #850, opened 2020-03-17
Fix percentage rounding

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/171831776

## Because

There are two `percentage` helper methods, one in DashboardHelper, and the other in MyFacilitiesHelper. Both of these were rounding incorrectly, however the [previous fix](https://github.com/simpledotorg/simple-server/pull/844) only addressed one of them

## This addresses

This removes the `percentage` helper in DashboardHelper, and fixes the one in MyFacilitiesHelper

### DIFF
--- a/app/helpers/dashboard_helper.rb
+++ b/app/helpers/dashboard_helper.rb
@@ -40,10 +40,4 @@ module DashboardHelper
 
     "#{percentage.round(0)}%"
   end
-
-  def percentage(numerator, denominator)
-    return percentage_string(0) if denominator.zero?
-
-    percentage_string((numerator * 100.0) / denominator)
-  end
 end
--- a/app/helpers/my_facilities_helper.rb
+++ b/app/helpers/my_facilities_helper.rb
@@ -6,6 +6,6 @@ module MyFacilitiesHelper
   def percentage(numerator, denominator)
     return 'NA' if denominator.nil? || denominator.zero? || numerator.nil?
 
-    percentage_string(((numerator * 100) / denominator).to_i)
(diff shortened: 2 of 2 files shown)


app/helpers/dashboard_helper.rb
```diff
@@ -40,10 +40,4 @@ def percentage_string(percentage)
 
     "#{percentage.round(0)}%"
   end
-
-  def percentage(numerator, denominator)
-    return percentage_string(0) if denominator.zero?
-
-    percentage_string((numerator * 100.0) / denominator)
-  end
```
harimohanraj89: I see controllers like [this one](https://github.com/simpledotorg/simple-server/blob/master/app/controllers/api/current/analytics/user_analytics_controller.rb#L2) that only use the `DashboardHelper`. If we remove this method, won't we break those controllers?

harimohanraj89:
Could you link to the PR or commit that you mean by "the previous fix"?

app/helpers/dashboard_helper.rb
vkrmis: No, the `percentage` method is only used within MyFacilities.

vkrmis:
> Could you link to the PR or commit that you mean by "the previous fix"?

It was done in this PR, linked it in the description as well.

harimohanraj89:
PS, looks like we don't have the best test coverage for these two modules (since the code changes didn't require any spec changes 😅 ) Let's punt on that for now. @vkrmis could I ask you to craft a dashboard v2 or techdebt ticket to add the necessary test coverage?

PR #851, opened 2020-03-17
Add the “web” cap role for all standalone sidekiq boxes

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/171832155
## Because

This is due to a bug addressed [here](https://simpledotorg.slack.com/archives/CFHKJ5WJY/p1584088859022800). Standalone sidekiq boxes don't have the `web` capistrano role and hence don't run the `assets:precompile` step. This causes the mailers to just break because the new e-mails wouldn't have the the appropriate assets available off of the sidekiq box (from where async emails are sent).

## This addresses

Adds a `web` role to all envs with standalone sidekiq boxes.


kitallis:
> Love it. I guess the other environments have `web app sidekiq db` all on a single machine, so no action required?

Yup.

PR #853, opened 2020-03-18
Remove staging artifacts

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/171597578

## Because

Now that our IHCI demo environment uses `SIMPLE_SERVER_ENV=demo` (https://github.com/simpledotorg/deployment/pull/173 has been deployed), we can remove the artifacts in our code that were only in use when `SIMPLE_SERVER_ENV=staging`.

## This addresses

* Removal of `staging` icon set
* Removal of `staging` manifest
* Updation of some styles and specs to drop support for `staging` and add support for `demo`.


harimohanraj89:
Tested to verify that the fixes for CSS classes on nav links works 👍

PR #854, opened 2020-03-19
Add the facility info to the district CSV download

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/171852791

## Because

CVHOs gave us feedback that they'd like to see the "second half" of district cohort reports present in the CSV download. That is, the per-facility breakdown of registration and followup numbers for the district.

## This addresses

* Adding the per-facility information requested to the Cohort CSV Download


app/views/analytics/districts/show.csv.erb
```diff
@@ -53,3 +53,44 @@
...
+<%=
+  [
+    "All",
+    @dashboard_analytics.sum { |_, row| row[:total_registered_patients] || 0 },
+    *dates.map { |period| analytics_totals(@dashboard_analytics, :follow_up_patients_by_period, period) },
+    *dates.map { |period| analytics_totals(@dashboard_analytics, :registered_patients_by_period, period) }
+  ].to_csv
+-%>
+<%
+  line = nil
```
kitallis: Unused?

app/views/analytics/districts/show.csv.erb
```diff
@@ -53,3 +53,44 @@
...
+  row_resource = Facility.order(:name)
+  row_resource_description = "Facilities"
```
kitallis: Unused?

harimohanraj89:
Good catch. Cruft removed in 4513281

PR #855, opened 2020-03-20
Replace unsafe html_safe use with safe code

timcheadle:
**Story card:** https://www.pivotaltracker.com/story/show/171901425

## Because

`html_safe` is not safe when using variable data. Let's replace it where possible, either with Rails built-in locale support or with `link_to` blocks.


app/views/patient_list_download_mailer/patient_list.html.erb
```diff
@@ -17,4 +17,4 @@
 <div>&nbsp;</div>
 <div>&nbsp;</div>
 
-<div style="font-size:110%;"><%= t("patient_list_email.unrequested").html_safe %></div>
+<div style="font-size:110%;"><%= t("patient_list_email.unrequested_html") %></div>
```
harimohanraj89: This will require the locale file to be updated too. In the interest of time, I've pushed the change to your branch already 👍 

**Before**
Screenshot 2020-03-23 at 12 36 05 PM


**After**
Screenshot 2020-03-23 at 12 36 08 PM

harimohanraj89:
@timcheadle Yesterday I mentioned that this PR might be missing one of the key user-input-html-safe-ifications that can cause real problems. I discovered today that your PR does indeed cover it [here](https://github.com/simpledotorg/simple-server/pull/855/files#diff-87d343a70b50e6c45f9219462cab9294R18-R21). This PR is good to go 🚀

PR #859, opened 2020-03-26
Disallow multiple OTP uses

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/171813295

## Because

An OTP can currently be used multiple times to generate as many access tokens as you'd like.

## This addresses

* Making the use of an OTP a one-time affair by setting the validity of the OTP to epoch upon successful use.


kitallis:
I think it might be valuable to add a controller spec that simulates using an OTP multiple times and disallows it from happening. What do you think?

harimohanraj89:
Added in 8e3e7fa

PR #860, opened 2020-03-27
Locale strings for the new progress tabs

kitallis:

## Because

The [new progress tab](https://github.com/simpledotorg/simple-server/pull/843) have a lot of strings that need to be localized.

## This addresses

* Left out all locales other an `en` as blank to be populated through Transifex later. 
* The new progress tab locales also sit under a different namespace: `user_analytics` so as to not affect the current progress tab in any way even if this PR gets deployed to Production.

### DIFF
--- a/config/locales/api/en.yml
+++ b/config/locales/api/en.yml
@@ -29,7 +29,7 @@ en:
     hypertension_control: 'Hypertension control'
     hypertension_control_desc: 'How many patients with hypertension had BP below 140/90 at their last follow-up visit in the month?'
     notes: 'Notes'
-    footer_note_1: "Data is for all activity at %{facility_name}. If data appears out of date, tap the \"Sync pending\" link on the home screen."
+    footer_note_1: 'Data is for all activity at %{facility_name}. If data appears out of date, tap the "Sync pending" link on the home screen.'
     footer_note_2: 'A follow-up patient is only counted for one visit each month. If a patient follows up multiple times in a month, he or she is only counted once.'
     thank_you: 'Thank You'
     thank_you_note: 'Thank you for all of your hard work treating patients with hypertension. Your work saves lives from heart attacks and strokes.'
--- a/config/locales/api/en.yml
+++ b/config/locales/api/en.yml
@@ -19,6 +19,22 @@ en:
     patients_enrolled: "How many new patients are enrolled at %{facility_name}?"
     all_time_patients: "All time: %{count} patients"
     updated_at: "Updated at %{time}"
+  user_analytics:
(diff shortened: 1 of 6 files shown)


config/locales/api/en.yml
```diff
@@ -19,6 +19,22 @@ en:
...
+    hypertension: 'Hypertension'
+    diabetes: 'Diabetes'
+    genders: ['all', male', 'female', 'transgender']
+    all_time: 'All-time'
+    follow_up_patients: 'Follow-up hypertension patients'
+    registered_patients: 'Registered Patients'
+    hypertension_control: 'Hypertension control'
+    hypertension_control_desc: 'How many patients with hypertension had BP below 140/90 at their last follow-up visit in the month?'
+    notes: 'Notes'
+    footer_note_1: "Data is for all activity at %{facility_name}. If data appears out of date, tap the \"Sync pending\" link on the home screen."
```
harimohanraj89: Easier to wrap this in single quotes? I think the interpolation still works right?

config/locales/api/en.yml
```diff
@@ -19,6 +19,22 @@ en:
     patients_enrolled: "How many new patients are enrolled at %{facility_name}?"
     all_time_patients: "All time: %{count} patients"
     updated_at: "Updated at %{time}"
+  user_analytics:
+    hypertension: 'Hypertension'
+    diabetes: 'Diabetes'
+    genders: ['all', male', 'female', 'transgender']
```
harimohanraj89: Will we be able to exclude `transgender` from, say, Ethiopia's progress tab?

kitallis:
> When will these locales be _applied_ into the new progress tab view?

I think we'll just go back to https://github.com/simpledotorg/simple-server/pull/843 and read off of the locales. Pretty straightforward.

config/locales/api/en.yml
kitallis: I don't exactly know how that will play out. But I believe the valid gender list will be a model level concern and thereby the presence of the gender translation should not be relevant to what data needs to be displayed.

harimohanraj89: Great point.

config/locales/api/en.yml
kitallis: Resolved. It works!

kitallis:
> Code changes look fine. Question: What happens if we merge this and #843 and a user views the progress tab in a non-English language? What will they see? Will their progress tab break?

Nope. Sorry, I mentioned this in the description. I've moved the new translations into a completely new locale namespace, so old things will not brea,

PR #863, opened 2020-03-31
My facilities fixes

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/171938138

## Because

We had some overlapping/confusing definitions for counting patients across various dashboards

## This addresses

This PR simplifies some of these definitions to make them more uniform. More detail on the changes is available [here](https://docs.google.com/document/d/1ISqqfbctlvFHuJ3mPgAiztdD1gCVP3__LOa6gCwNy0c/edit#heading=h.q8zm89ru7j26)

### DIFF
--- a/app/queries/my_facilities/blood_pressure_control_query.rb
+++ b/app/queries/my_facilities/blood_pressure_control_query.rb
@@ -8,6 +8,8 @@ class MyFacilities::BloodPressureControlQuery
   include QuarterHelper
   include MonthHelper
 
+  REGISTRATION_BUFFER = 2.months
+
   def initialize(facilities: Facility.all, cohort_period: {})
     # cohort_period is map that contains
     # - :cohort_period (:quarter/:month),
@@ -36,14 +38,27 @@ class MyFacilities::BloodPressureControlQuery
     @cohort_period == :month ? monthly_bps_cte : quarterly_bps
   end
 
-  def all_time_patients
-    @all_time_patients ||= Patient
-                      .where('recorded_at < ?', Time.current.beginning_of_day - 2.months)
+  def cohort_missed_visits_count
+    cohort_registrations.count - cohort_bps.count
+  end
+
+  def cohort_missed_visits_count_by_facility
+    registrations = cohort_registrations.group(:registration_facility_id).count
+    bps = cohort_bps.group(:registration_facility_id).count
+
+    @facilities.map do |f|
+      [f.id, (registrations[f.id].to_i - bps[f.id].to_i)]
+    end.to_h
+  end
+
+  def overall_patients
+    @overall_patients ||= Patient
+                      .where('recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
                       .where(registration_facility: facilities)
   end
 
-  def all_time_controlled_bps
-    @all_time_controlled_bps ||=
+  def overall_controlled_bps
+    @overall_controlled_bps ||=
       LatestBloodPressuresPerPatient
       .where(registration_facility: facilities)
       .where('bp_recorded_at > ?', Time.current.beginning_of_day - 90.days)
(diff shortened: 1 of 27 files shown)


spec/models/latest_blood_pressures_per_patient_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 RSpec.describe LatestBloodPressuresPerPatient, type: :model do
   describe 'Associations' do
-    it { should belong_to(:facility) }
+    it { should belong_to(:bp_facility) }
```
harimohanraj89: Let's use the `it { is_expected.to belong_to }` syntax [[ref](http://www.betterspecs.org/#expect)]

app/queries/my_facilities/blood_pressure_control_query.rb
```diff
@@ -32,15 +32,20 @@ def cohort_uncontrolled_bps
...
-    @all_time_bps ||= LatestBloodPressuresPerPatient
-                      .where('patient_recorded_at < ?', Time.current.beginning_of_day - 2.months)
-                      .where(bp_facility_id: facilities)
+  def cohort_bps
+    @cohort_period == :month ? monthly_bps_cte : quarterly_bps
+  end
+
+  def all_time_patients
+    @all_time_patients ||= Patient
+                      .where('recorded_at < ?', Time.current.beginning_of_day - 2.months)
```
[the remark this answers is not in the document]
harimohanraj89: Just clarifying, since we're using the `Patient` table directly now, we now include patients with 0 BPs in the denominator (as intended), right?

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -63,16 +33,34 @@ def calls_made
       .group('facilities.id::uuid')
   end
 
+  def bp_query_by_cohort
```
harimohanraj89: Neat. This generates a hash of BP queries, one for each time period, right?

app/queries/my_facilities/blood_pressure_control_query.rb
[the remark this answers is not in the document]
vkrmis: I removed the 2.months constant since we've stopped using it, but I can add it back for this query.

`patients.recorded_at` is either the time at which the patient record was created on the app or the `recorded_at` of their earliest BP, whichever is earlier (this is to support retroactive data entry).

Yes, we are now including patients with 0 BPs in the denominator.

app/queries/my_facilities/missed_visits_query.rb
vkrmis: That's correct, I've tried to minimize the changes here so I can return more or less the same data structures from the query. This is so controller and view code stays _mostly_ the same.

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -55,28 +55,25 @@ def missed_visits_by_facility
...
-
-        total_missed_visit_data[period][:patients] ||= 0
-        total_missed_visit_data[period][:patients] += missed_visit_data[:patients]
-
-        total_missed_visit_data[period][:missed] ||= 0
-        total_missed_visit_data[period][:missed] += missed_visit_data[:missed]
-      end
+      bp_query_by_cohort.map do |(year, period), bp_query|
+        cohort_registrations = bp_query.cohort_registrations.count
+        cohort_missed_visits_count = bp_query.cohort_missed_visits_count
```
harimohanraj89: Meh, I'd be ok not using these local vars, and just directly operating on the bp_query below. 🤷‍♂

app/queries/my_facilities/blood_pressure_control_query.rb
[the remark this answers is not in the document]
harimohanraj89: Ah, `recorded_at` is the _earliest_ activity. Nice.

PR #864, opened 2020-03-31
Turn the OTP SMS into a sidekiq job

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/172074271

## Because

When `users/activate` API sends an SMS asynchronously along with the user deets, the app can get into a place where the SMS arrives before the user details are saved.

To gracefully handle the SMS arriving early on the app is a larger change, so for now we just delay sending the SMS from the server-side so that situation is effectively pushed away.

## This addresses

Make the OTP SMS an async job and delay it by a configurable amount (defaulted to 5 seconds).

**Note**

The related deployment repo changes are here: https://github.com/simpledotorg/deployment/pull/180


app/controllers/api/v4/users_controller.rb
```diff
@@ -5,6 +5,8 @@ class Api::V4::UsersController < APIController
   skip_before_action :validate_facility, only: [:find, :activate]
   skip_before_action :validate_current_facility_belongs_to_users_facility_group, only: [:find, :activate]
 
+  DEFAULT_USER_OTP_DELAY_IN_SECS = 5
```
harimohanraj89: Let's call this `SECONDS` for consistency 😄

spec/controllers/api/v4/users_controller_spec.rb
```diff
@@ -42,6 +42,9 @@
 
       post :activate, params: { user: { id: user.id, password: '1234' } }
 
+      # execute the RequestOtpSmsJob
+      Sidekiq::Worker.drain_all
```
harimohanraj89: Neat!

PR #866, opened 2020-04-01
Follow-up and medication info in line list

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172097532

## Because

Adding follow-up and medication info will greatly aid health workers in following up with patients, especially with a heightened emphasis on community-level and tele-medicine.

## This addresses

* Adds the following fields to the patient line list download
  * Followup facility name
  * Followup date
  * List of medication columns in the format `Medication 1, Dosage 1, Medication 2, Dosage 2...`
* Adds an `includes` statement to the patient exporter to remove N+1 queries
* Adds several associations to the `Patient` model to power the `includes` ☝️


app/exporters/patients_exporter.rb
```diff
@@ -74,10 +96,17 @@ def self.csv_fields(patient)
...
       ('High' if patient.high_risk?),
       patient.latest_bp_passport&.shortcode,
-      patient.id
+      patient.id,
+      *medications_for(patient)
     ]
   end
+
+  def self.medications_for(patient)
+    patient.prescription_drugs.map { |drug| [drug.name, drug.dosage] }.flatten
```
kitallis, line 110: I think we can do a `flat_map` here?

spec/models/patient_spec.rb
```diff
@@ -5,14 +5,14 @@
...
   describe 'Associations' do
-    it { should belong_to(:address).optional }
-    it { should have_many(:phone_numbers) }
-    it { should have_many(:blood_pressures) }
-    it { should have_many(:blood_sugars) }
-    it { should have_many(:prescription_drugs) }
-    it { should have_many(:facilities).through(:blood_pressures) }
-    it { should have_many(:appointments) }
-    it { should have_one(:medical_history) }
+    it { is_expected.to belong_to(:address).optional }
```
kitallis, line 8: Is this some newfangled rspec syntax?

app/exporters/patients_exporter.rb
```diff
@@ -10,7 +10,17 @@ def self.csv(patients)
...
       patients.in_batches(of: BATCH_SIZE).each do |batch|
-        batch.each do |patient|
+        batch.includes(
+          :registration_facility,
+          :phone_numbers,
+          :address,
+          :medical_history,
+          :prescription_drugs,
+          :latest_bp_passports,
+          { latest_scheduled_appointments: :facility },
```
kitallis, line 20: I assume we're cascadingly including `facility` with `latest_scheduled_appointments` and `latest_blood_pressures` because of patient transfer and/or just patient visiting elsewhere?

spec/models/patient_spec.rb
harimohanraj89: It's the equivalent of replacing `object.should eq(something)` with `expect(object).to eq(something)`. The `expect` syntax is the new recommended way, and `should` is frowned upon (it does a bunch of `Object` monkeypatching I think to power its DSL)

app/exporters/patients_exporter.rb
harimohanraj89: YES!

app/exporters/patients_exporter.rb
harimohanraj89: The latest scheduled appointments pull in their facility to power the new requirement in this PR: Showing the follow-up facility name.

The latest BPs pull in their facility because the latest BP's facility info is included in the CSV as well.

PR #869, opened 2020-04-02
The Updated timestamp on the Progress Tab reflects the data refresh time

kitallis:

## Because

This is an addendum to the Progress Tabs PR: https://github.com/simpledotorg/simple-server/pull/843. The `Updated` timestamp on the progress tabs don't reflect the data refresh time, instead they reflect the time at which the page was refreshed. This would be confusing to the users who are frequently updating and adding data and checking the progress tab.

## This addresses

The progress tab is now cached against the last refresh date for the materialized view and until that date moves ahead, the same data is served off of the cache.

This date is also used to cap the upper-bound of time for the non-materialized view queries that come straight from the tables.

This date is what is now displayed as the `Updated` timestamp on the progress tab.


kitallis:
Closed this in favor of the cleaner: https://github.com/simpledotorg/simple-server/pull/872

PR #870, opened 2020-04-03
Add app.json

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172145441

## Because

We want to set up Heroku Review Apps

## This addresses

The first step to review app configuration is the addition of an `app.json` file that governs how to set up a new Heroku app. This is very similar to, say, a terraform config. It dictates environment variables, Heroku buildpacks, required addons, processes.

[Heroku reference](https://devcenter.heroku.com/articles/app-json-schema)

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,0 +1,233 @@
+{
+  "name": "Simple Server",
+  "description": "The web app behind Simple.org",
+  "keywords": [
+    "healthcare"
+  ],
+  "website": "https://simple.org",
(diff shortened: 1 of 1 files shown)


harimohanraj89:
Turns out Heroku is smart enough for us to test #871 on its own. Closing this.

harimohanraj89:
Never mind, Heroku is _not_ smart enough to test #871 on its own. We'll have to first merge these `app.json` changes before we can test review app creation.

PR #871, opened 2020-04-03
Review apps

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172145441

## Because

We'd like to use Heroku Reviews Apps to spin up feature environments for easy testing.

## This addresses

The setup for launching on Heroku, and to power review apps.

* Adds a Ruby version to the `Gemfile`
* Adds a `Procfile`
* Adds a Heroku-specific `sidekiq-heroku.yml` configuration. It avoids using pid and log files in favor of Heroku's automated process management and logging
* Allows Simple to fall back to using `REDIS_URL` for all Redis instances
* Removes the conflicting `package-lock.json` in favor of `yarn.lock`
* Fixes an asset reference while we're at it

## To Do

* [x] Merge #870 first
* [x] Set `SIMPLE_SERVER_HOST` dynamically
* [x] Create a super admin on spin-up


kitallis:
Looks good. Should we drop a note in the README that specifies that we fallback to `REDIS_URL` for heroku envs?

harimohanraj89:
I'd love to put that in a "Configuration" section in our README, that details the different env vars necessary to run Simple. We don't have that yet, should we add now?

harimohanraj89:
As discussed offline, we will mark these docs requirements in our tech debt list and handle later.

PR #872, opened 2020-04-06
The Updated timestamp on the Progress Tab reflects the data refresh time

kitallis:

## Because

This is an addendum to the Progress Tabs PR: https://github.com/simpledotorg/simple-server/pull/843. The `Updated` timestamp on the progress tabs don't reflect the data refresh time, instead they reflect the time at which the page was refreshed. This would be confusing to the users who are frequently updating and adding data and checking the progress tab.

## This addresses

- [x] The progress tab is now cached against the last refresh date for the materialized view and until that date moves ahead, the same data is served off of the cache.

- [x] This date is also used to cap the upper-bound of time for the non-materialized view queries that come straight from the tables.

- [x] This date is what is now displayed as the `Updated` timestamp on the progress tab.

**Note**

Since the `Updated` timestamp is acquired off of the last refresh, the first time this is deployed, it will default to `Time.current` and eventually correct itself. We can trigger a manual refresh just to get things on level.


spec/lib/tasks/refresh_materialized_db_views_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    expect(Rails.cache.fetch(matview_refresh_time_key)).to eq(refresh_begin_time)
+  end
+
+  it 'does not modify the refresh time if any view fails to refresh' do
+    Timecop.freeze(refresh_begin_time) do
+      invoke_task('refresh_materialized_db_views')
+    end
+
+    Timecop.freeze(refresh_begin_time + 1.day) do
+      expect(LatestBloodPressuresPerPatientPerDay).to receive(:refresh).and_raise(StandardError)
```
prabhanshuguptagit, line 28: Would an `allow` instead of `except` here be idiomatic?

kitallis: This is spot-on. It should be `allow`.

app/queries/user_analytics_query.rb
```diff
@@ -3,33 +3,34 @@ class UserAnalyticsQuery
   include MonthHelper
   include PeriodHelper
 
-  attr_reader :current_facility, :days_ago, :months_ago
```
harimohanraj89, line 6: Why did we axe the attr_readers? Can we reinstate them?

config/initializers/constants.rb
```diff
@@ -0,0 +1,3 @@
+Rails.application.config.app_constants = {
+  MATVIEW_REFRESH_TIME_KEY: 'last_materialized_view_refresh_time'.freeze
+}
```
harimohanraj89, line 3: Neat.

spec/controllers/api/v2/analytics/user_analytics_controller_spec.rb
```diff
@@ -11,6 +11,10 @@
   end
 
   describe '#show' do
+    before do
+      Rails.cache.clear
+    end
```
harimohanraj89, line 16: Can this go into our spec helper?

app/queries/user_analytics_query.rb
```diff
@@ -3,33 +3,34 @@ class UserAnalyticsQuery
...
-
-  def initialize(facility, days_ago: 30, months_ago: 6)
+  def initialize(facility, days_ago: 30, months_ago: 6, fetch_until: Date.current)
```
harimohanraj89, line 6: The default for this in the controller is `Time.current`, but the default here is `Date.current`. Should this be a Date or a Time? Can we consistent-ize across both places?

app/presenters/user_analytics_presenter.rb
```diff
@@ -179,4 +189,8 @@ def group_by_gender(resource_data)
   def data_for_unavailable_dates(period_list)
     period_list.map { |date| [date, 0] }.to_h
   end
+
+  def statistics_cache_key
+    ['user_analytics', @last_refreshed_at].join('/')
```
harimohanraj89, line 194: Consider `"users_analytics/#{@last_refreshed_at}"`

app/presenters/user_analytics_presenter.rb
```diff
@@ -16,22 +16,32 @@ def initialize(current_facility)
...
-      trophies: trophy_stats,
-      metadata: {
-        is_diabetes_enabled: false,
-        last_updated_at: I18n.l(Time.current),
-        formatted_next_date: display_date(Time.current + 1.day),
-        today_string: I18n.t(:today_str)
-      }
-    }
+    @statistics ||=
+      Rails.cache.fetch(statistics_cache_key) do
```
harimohanraj89, line 31: If I understand correctly, we are introducing caching here because we now can, since the data is up to the last matview refresh, and not up to the present time when the page is fetched.

app/queries/user_analytics_query.rb
kitallis: I axed them because I have become disillusioned.

I have become disillusioned by the best-practice OOisms from the Sandi Metzes and the Uncle Bobs of the world.

Lot of the semantic best practices of OO, I feel, come as a protection against dubious programmers and programming. As I understand it, the common reason for sticking this piece of *data* in an `attr_reader` is to make it *behavior*. This ostensibly allows for two things:

1. You can change the meaning of this `attr_reader` by re-defining it later, as a proper public method, and modifying its *behavior* without changing any code that depends on it. For example: `days_ago` can later become: `def days_ago; @days_ago - 1; end` and have it be propagated everywhere.
2. You deal with the piece of *data* in an effectively immutable way.

I find this form of I̧̳͇̦n͍̞̥̙̹fo̟͕r̡̧̗͇ma̡̫͇͔t̢̙̣͚͟i̞̥̺̥on̼ H̫͓͍̰įd͍͉̫̻i̡͕n̞g͇̬͇̠̘ a local-maxima optimization.

1., has two problems in my head,

a. **Behavior**: This conflates data with behavior. There's now no distinction between them. Classes are nothing but data + behavior, and we've made all data behavior just for refactoring ease for the future, that RubyMine can do in 5 seconds.
b. **Visibility**: This class now needlessly exposes pieces of data as behavior (get-ers) when the interface shouldn't really care about reading `days_ago` off of the object at all, further complicating what the public API of this class is.

2., is faking immutability and a protection against dubious programming. If we're worried about typos causing us state issues, we should invest in languages that are typed and support immutability as first-class citizens.

I see another general issue with this pattern (an extension of 1a.), that is of defining the boundaries of what should be considered data. Data is data. It's never anything else. As soon as it becomes non-static, it's a transformation on the data. It's important to know what is clearly static and what isn't, when looking at a class, imo.

Similarly (an extension of 1b.), it's important to clearly know what the public API of the class is by just looking at. It is the public methods + `attr_*`.

**Caveats**

* None of this is a stone-tablet. All of it is just my opinion. I'm happy to change it if this is a pattern we'd like to adopt.
* Some of the problems of 1b. can be solved by a pattern that I've [used in the past](https://github.com/simpledotorg/simple-server/blob/master/app/services/sms_notification_service.rb#L27). This allows the reader to be non-public, but still act as *behavior* and be immutable. This is not always feasible though since the boundaries of access become unclear when you start using private `attr_readers` in public methods of the class. I'm tentatively open to adopting this pattern though, if we like it.
* This entire comment is intentionally written in a lightly-frustrated tone. I'm not actually salty about any of this 👾

app/presenters/user_analytics_presenter.rb
kitallis: Yeah, just stuck in caching as a natural extension to capping data until the last refresh.

app/queries/user_analytics_query.rb
harimohanraj89: I definitely do not feel as strongly about this one way or the other 😅 Let's axe em! Be sure to catch these in my PRs as well 👍

PR #875, opened 2020-04-07
Remove V2 APIs and related code

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/172199846

## Because

Most users have upgraded to an app version that uses V3 APIs, we should deprecate V2 when its safe to do so.

## This addresses

- Removes V2 API code
- Removes V2 API specs
- Removes the API Version Generator
- Returns an [`HTTP 410`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410) response for any requests to `/api/v2/` paths


vkrmis:
Re-opening this PR to deprecate the V2 API

spec/controllers/api/v3/analytics/user_analytics_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-RSpec.describe Api::V2::Analytics::UserAnalyticsController, type: :controller do
+RSpec.describe Api::V3::Analytics::UserAnalyticsController, type: :controller do
```
vkrmis, line 3: @kitallis I'm guessing this was an accident. Let me know if this was done intentionally. This change was introduced in the HTN only dashboards PR.

kitallis: The state prior to that was incorrect at some point, the `v3` spec was calling into `v2`. This is the correct state.

vkrmis:
> Can we remove the V2 files from the `.standardrb-todo.yml`?
Done 
> Can we remove the `Api::V2::LogApiUsageByUsers` too?
Moved it to V3 concern, could be useful in the future.

PR #878, opened 2020-04-10
Sort overdue list

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172197925

## Because

The overdue dashboard page did not sort patients correctly

## This addresses

Correctly sorting overdue patients by tweaking and using the `patient_summaries` SQL view to order all overdue patients by risk before pagination.

Implementation notes:

* Turns out we shouldn’t use a materialized view because the page allows for actions that should immediately change the page’s contents! The overdue page allows you to mark patients call results, after which the patient should be removed from the list. We’ll have to take the cost of joining tables on every request and maybe slower ordering to support this real-time behavior. I think it’ll be ok, not too bad.
* I've duplicated the patient card partial so that the "Adherence Patients" page can continue to use the existing one.
* I've added a couple of associations to the `PatientSummary` model to power the `includes` statements.
* I was able to test this new approach side-by-side against the existing approach, and confirmed parity.
  * The only difference is that the old view was showing `blood_pressures.device_created_at` for the "Recorded at..." timestamp. I've replaced it with `blood_pressures.recorded_at`.


app/models/patient_summary.rb
```diff
@@ -1,7 +1,17 @@
...
   def readonly?
     true
   end
+
+  def self.overdue
+    where(next_appointment_status: 'scheduled')
+      .where('next_appointment_scheduled_date < ?', Date.current)
+      .where('next_appointment_scheduled_date >= ?', 365.days.ago)
+      .where('next_appointment_remind_on IS NULL OR next_appointment_remind_on <= ?', Date.current)
+  end
```
prabhanshuguptagit, line 16: `.overdue` could be a scope.

Just a thought, if we named `next_appointment_scheduled_date` as `scheduled_date` and so on, this scope could be re-used across `Appointment` and `PatientSummary`. Might be worth doing to avoid duplication.

harimohanraj89: I considered using a scope, but avoided it because of some silly multi-line lambda code style issues. On second thought, I think a scope is the way to go.

I'm hesitant to rename this column since the prefix is basically a namespace to avoid clashes with joined tables. Imagine a scenario where we include the latest scheduled and latest visited appointment in this table 🤔 I'll give some thought to avoiding this duplication somehow.

kitallis:
I think this looks good. I'm wondering whether it's not such a good idea to duplicate the Overdue and Adherence view partials. They are intentionally made to look identical, cutting them up might cause drift that we have to maintain.

The fundamental difference between the two is the `patient_summary` versus `patient` presenter object. Could we just de-structure it in the parent view and send these over as partial locals and keep the two in sync?

Let me know if I'm missing any edge-case here.

harimohanraj89:
The `PatientSummary` is a whole different data source, so swapping it in wouldn't be a trivial "wrap something around patients". I chose to split up the partials for the following reasons:
* I thought the risk of flipping the Adherence page was greater than the pay-off
* The Adherence page is primarily created for PATH, and not really used by IHCI
* These are old dashboards anyway, we're not going to continue putting love into them.
* Drift in definitions is very problematic, but drift in view logic is less critical

For these reasons, I split the partial and made the smaller change to the controllers. Does that logic seem agreeable?

app/models/patient_summary.rb
harimohanraj89: @prabhanshuguptagit I think I've got it. I've reworked the `PatientSummary.overdue` scope to simply re-use the `Appointment.overdue` scope through a `#merge()`. Check it out in 63cea05

This avoids duplication and also doesn't require a bunch of column name juggling.

prabhanshuguptagit: Neat 👍

PR #883, opened 2020-04-13
Log V2 api usage

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/172286787

## Because

We need to know how many users are still using the V2 APIs

## This addresses

Adds a custom logger that captures data about which users are using the V2 APIs


app/controllers/concerns/api/v2/log_api_usage_by_users.rb
```diff
@@ -0,0 +1,10 @@
+module Api::V2::LogApiUsageByUsers
+  extend ActiveSupport::Concern
+  included do
+    before_action :log_api_usage
+
+    def log_api_usage
+      Loggers::ApiUsageLogger.tagged(params[:controller], params[:action], current_user&.id || "no-user-id") { Loggers::ApiUsageLogger.info(1) }
```
harimohanraj89, line 7: What's an example output of this? How are we indicating that the third param is the user id?

kitallis: We could stick a `user:<uuid>` in the tag, that should help?


(although the file is quite specific to users, but the above change should be tiny)

harimohanraj89: Ah, so these logs are going to be written to a separate file? I'm guessing then we'll parse the file to get a count of how many unique users have used the v2 API in the last 7 days ,or 30 days, or whatever time frame.

All of this seems a bit wheel-re-inventy to me. Can we instead stick a `::NewRelic::Agent.add_custom_attributes({ user_id: current_user&.id || "" })` as an `after_action` on our V2 APIs, or even on all our APIs? We can then use NR to ask these questions.

We can do both as well. No harm in capturing this info two ways.

kitallis: I think we can try that. However, we should make sure that we can actually aggregate these attributes on NR over long a period of time.

PR #885, opened 2020-04-13
Patient app API authentication

timcheadle:
**Story cards:**
- https://www.pivotaltracker.com/story/show/172298399
- https://www.pivotaltracker.com/story/show/172298486

## Because

We're enabling patients to download their own BP Passport record from Simple, and we need to authenticate and authorize them to do so. This adds patient app APIs as a separate namespace to our current API endpoints.

See the PRD:
https://docs.google.com/document/d/1nrdTxR0nyv8aKnPTo32Lnem9YiddusbfjNgKW7uKxIk/edit#

## This addresses

This adds patient activation (via OTP) and authentication via access token.

## Implementation Details

* A new `PassportAuthentication` model is introduced, referencing a BP passport and a Patient
* Two API endpoints are introduced
  * `POST /api/v4/patient/activate` - Send a BP Passport UUID and trigger an OTP
  * `POST /api/v4/patient/login` - Send a BP Passport UUID and OTP and receive an access token
* A new `PatientAPIController` base controller class is introduced to separate patient app concerns from the Simple app API

## To Do

- [x] Sending the OTP via SMS
- [x] Handling access token expiration - New tokens are issued on every sign-in, like the Simple app, for now. Expiration of each token should likely be a subsequent PR.
- [x] Sending errors if needed - We will establish a contract with the A2 team, but don't need to implement it here.
- [x] Any necessary integration with swagger/schema/docs
- [x] Integration specs to ensure the thing actually works :)
- [x] Wrap responses in a `patient: {}` key
- [x] Stop inheriting from `ApiController`
- [x] Pick a naming convention for the API endpoints. The most consistent option would be `activate` and `login`, instead of `request_otp` and `activate`.

### DIFF
--- a/app/controllers/api/v4/patient_controller.rb
+++ b/app/controllers/api/v4/patient_controller.rb
@@ -53,8 +53,10 @@ class Api::V4::PatientController < PatientAPIController
 
   def access_token_response(authentication)
     {
-      access_token: authentication.access_token,
-      patient_id: authentication.patient.id
+      patient: {
+        access_token: authentication.access_token,
+        patient_id: authentication.patient.id
+      }
     }
   end
 end
--- a/app/controllers/api/v4/patient_controller.rb
+++ b/app/controllers/api/v4/patient_controller.rb
@@ -20,7 +20,9 @@ class Api::V4::PatientController < APIController
       patient_business_identifier: passport
     )
 
-    # TODO: Send OTP
+    unless FeatureToggle.enabled?('FIXED_OTP_ON_REQUEST_FOR_QA')
+      SendPatientOtpSmsJob.set(wait: otp_delay_seconds).perform_later(authentication)
+    end
 
     head :ok
   end
@@ -54,5 +56,9 @@ class Api::V4::PatientController < APIController
 
   def access_token_response(authentication)
     { access_token: authentication.access_token }
+  end
+
+  def otp_delay_seconds
+    (ENV['USER_OTP_SMS_DELAY_IN_SECONDS'] || DEFAULT_USER_OTP_DELAY_IN_SECONDS).to_i.seconds
   end
 end
--- a/app/controllers/api/v4/patient_controller.rb
+++ b/app/controllers/api/v4/patient_controller.rb
@@ -20,6 +20,11 @@ class Api::V4::PatientController < APIController
       patient_business_identifier: passport
     )
 
+    unless authentication.otp_valid?
+      authentication.reset_otp
+      authentication.save!
+    end
+
     unless FeatureToggle.enabled?('FIXED_OTP_ON_REQUEST_FOR_QA')
       SendPatientOtpSmsJob.set(wait: otp_delay_seconds).perform_later(authentication)
     end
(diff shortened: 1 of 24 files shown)


timcheadle:
This still needs work:
- Sending the OTP via SMS
- Handling access token expiration
- Sending errors if needed
- Any necessary integration with swagger/schema/docs

app/jobs/send_patient_otp_sms_job.rb
```diff
@@ -0,0 +1,14 @@
+class SendPatientOtpSmsJob < ApplicationJob
+  queue_as :default
+  self.queue_adapter = :sidekiq
+
+  def perform(patient_authentication)
```
timcheadle, line 5: Typically you want to pass model IDs to jobs and do a `find`. Passing them like this serializes the object in Redis. For this one it's fine for now, just a heads up.

harimohanraj89: From Rails 5, ActiveJob now handles ActiveRecord objects seamlessly using GlobalID ([check it out](https://guides.rubyonrails.org/v5.0/active_job_basics.html#globalid)), so we should be totally fine to pass in the record directly!

app/controllers/api/v4/patient_controller.rb
```diff
@@ -32,7 +32,7 @@ def activate
     head :ok
   end
 
-  def authenticate
+  def activate
```
harimohanraj89, line 35: If we want to be consistent with the v3 naming, should we call this `login`?

harimohanraj89:
@timcheadle I've added more specs, performed some cleanup, and commented as well.

I think the test coverage is good now, and this PR is ready for review.

app/jobs/send_patient_otp_sms_job.rb
timcheadle: Nice! TIL

app/controllers/api/v4/patient_controller.rb
vkrmis: This should be `login`, and `request_otp` should be `activate` to be consistent with v4 user APIs

harimohanraj89: That sounds right to me.

@timcheadle since that change would cause the endpoints to change on Automation Squared again, I'm going to hold off on making it. When you get in, if you agree, can you make the change and notify A2?

app/controllers/api/v4/patient_controller.rb
```diff
@@ -0,0 +1,66 @@
+class Api::V4::PatientController < APIController
```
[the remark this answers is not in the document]
harimohanraj89: That's a great idea.

app/services/sms_notification_service.rb
```diff
@@ -22,6 +22,10 @@ def send_reminder_sms(reminder_type, appointment, callback_url, locale = DEFAULT
     send_sms(body, callback_url)
   end
 
+  def send_patient_request_otp_sms(otp)
```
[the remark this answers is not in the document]
harimohanraj89: We don't want to just yet. We can add that later as needed.

harimohanraj89:
@timcheadle I've updated the PR description with our remaining to-dos based on feedback.

timcheadle:
I've attempted to resolve all outstanding issues. Please review and fix whatever is necessary to merge this in.

PR #886, opened 2020-04-14
Fix missing exclusion in MyFacilities BloodPressureControlQuery

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/172306738

## Because

An exclusion for recently registered patients is missing from the numerator for the BloodPressureControlQuery

## This addresses

Re-adds the exclusion for recent patients

### DIFF
--- a/app/queries/my_facilities/blood_pressure_control_query.rb
+++ b/app/queries/my_facilities/blood_pressure_control_query.rb
@@ -53,14 +53,13 @@ class MyFacilities::BloodPressureControlQuery
 
   def overall_patients
     @overall_patients ||= Patient
-                      .where('recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
-                      .where(registration_facility: facilities)
+                          .where('recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
+                          .where(registration_facility: facilities)
   end
 
   def overall_controlled_bps
     @overall_controlled_bps ||=
       LatestBloodPressuresPerPatient
-      .where(registration_facility: facilities)
       .where(patient: overall_patients)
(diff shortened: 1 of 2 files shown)


app/queries/my_facilities/blood_pressure_control_query.rb
```diff
@@ -61,6 +61,7 @@ def overall_controlled_bps
     @overall_controlled_bps ||=
       LatestBloodPressuresPerPatient
       .where(registration_facility: facilities)
+      .where('patient_recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
```
harimohanraj89, line 64: What's the corresponding denominator method? We should make this method something like

```
def overall_controlled_bps
  denominator.under_control
end
```

to ensure that the numerator is always a subset of the denominator as intended.

vkrmis: The denominator is `overall_patients` which is now a query on `Patient` instead of `LatestBloodPressurePerPatient`.
We had something along the lines of what you recommended earlier, but lost it when we changed the denominator to be a query on `Patients` directly.
I've updated this condition to be `.where(patients: overall_patients)` instead.

app/queries/my_facilities/blood_pressure_control_query.rb
```diff
@@ -61,6 +61,7 @@ def overall_controlled_bps
     @overall_controlled_bps ||=
       LatestBloodPressuresPerPatient
       .where(registration_facility: facilities)
+      .where(patient: overall_patients)
```
harimohanraj89, line 64: Neat. We can remove the `.where(registration_facility: facilities)` now too, right?

PR #888, opened 2020-04-16
Upgrade our Follow-ups queries across dashboards

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/172264306

## Because

We've seen issues of follow-ups not lining up between the old dashboards, within the old dashboards and across Progress Tabs. This is primarily due to different ways of calculating follow-ups and not accounting for all edge-cases properly.

## This addresses

This attempts to simplify the ergonomics of calculating follow-ups and the queries themselves. Since most user-facing dashboards are cached, the queries are intentionally not optimized to be very fast. However, they *are* reasonably fast for most cases (< 2-3s, worst-case).

**Note**: Some inline comments in the PR are left for ease of review since this PR re-formats and deletes a lot of code.


app/models/facility.rb
```diff
@@ -51,6 +48,7 @@ class Facility < ApplicationRecord
 
   delegate :protocol, to: :facility_group, allow_nil: true
   delegate :organization, to: :facility_group, allow_nil: true
+  delegate :follow_ups, to: :patients, allow_nil: true, prefix: :patient
```
kitallis, line 51: This is largely the only change made to this file, the rest is space changes.

app/presenters/user_analytics_presenter.rb
```diff
@@ -160,37 +123,64 @@ def trophy_stats
...
-  def data_for_unavailable_dates(period_list)
-    period_list.map { |date| [date, 0] }.to_h
+  def all_time_registrations_by_gender
+    @current_facility
+      .registered_patients
+      .group(:gender)
   end
 
   def statistics_cache_key
-    "user_analytics/#{@current_facility.id}/#{@last_refreshed_at}"
```
kitallis, line 194: Since there's no matview business, we don't need to cache by time, we just expire every `15.minutes`

app/queries/facility_analytics_query.rb
```diff
@@ -37,28 +37,19 @@ def registered_patients_by_period
...
-        .joins(:facility)
-        .where(facility: @facility)
-        .where(deleted_at: nil)
-        .group('users.id')
-        .group_by_period(@period, 'blood_pressures.recorded_at')
-        .where("patients.recorded_at < #{date_truncate_string}")
-        .order('users.id')
-        .distinct
-        .count('patients.id')
+      Patient
```
kitallis, line 41: This is an ugly hacky query, but we have intentionally not spent too much effort improving it. It however, produces the correct results.

app/views/api/v2/analytics/user_analytics/show.html.erb
```diff
@@ -13,178 +13,180 @@
...
+          <td class="row-label no-padding"><%= month_date.strftime("%B") %></td>
+          <td><span class="percent">
+                  <%=
+                    @user_analytics.display_percentage(@statistics.dig(:monthly,
+                                                                       :grouped_by_date,
+                                                                       :controlled_visits,
+                                                                       month_date),
+                                                       @statistics.dig(:monthly,
+                                                                       :grouped_by_date,
+                                                                       :follow_ups,
```
kitallis, line 127: This used to be called `total_visits` before, it's now called `follow_ups`. This is the only logical change to this file. The rest is just reformatting.

app/views/api/v3/analytics/user_analytics/_monthly_gender_table.html.erb
```diff
@@ -14,7 +14,10 @@
     <tr>
       <td class="row-label no-padding"><%= month_date.strftime("%B") %></td>
       <td>
-        <%= zero_if_unavailable(@user_analytics.stats_across_genders_for_month(resource, month_date)) %>
```
kitallis, line 17: We've refactored the data-structure to remove the need for the presenter method `stats_across_genders_for_month` entirely. It's a lot simpler this way.

app/models/facility.rb
```diff
@@ -17,12 +17,10 @@ class Facility < ApplicationRecord
   has_many :encounters
   has_many :blood_pressures, through: :encounters, source: :blood_pressures
   has_many :blood_sugars, through: :encounters, source: :blood_sugars
-  has_many :patients, -> { distinct }, through: :encounters
+  has_many :patients, -> { distinct }, through: :blood_pressures
```
kitallis, line 20: This is the first change made to this file.

app/models/facility.rb
```diff
@@ -51,6 +49,7 @@ class Facility < ApplicationRecord
 
   delegate :protocol, to: :facility_group, allow_nil: true
   delegate :organization, to: :facility_group, allow_nil: true
+  delegate :follow_ups, to: :patients, prefix: :patient
```
kitallis, line 52: This is the second relevant change made to this file. The rest is reformatting.

app/models/patient.rb
```diff
@@ -53,6 +53,11 @@ class Patient < ApplicationRecord
 
   attribute :call_result, :string
 
+  scope :follow_ups, -> (period, last: nil) {
```
kitallis, line 56: This scope is intentionally partially applied in two ways:

1. It doesn't join over `blood_pressures`, even though it expects it to exist
2. It ends with a group clause, which means the caller must either add more groups or run an aggregate themselves.

This is done for 2 reasons:

1. Rails doesn't properly resolve queries when, say, this scope is run through [facilities](https://github.com/simpledotorg/simple-server/pull/888/files#diff-a0e971b921f56b979cc5aa07fee7066fR52) like:

`Facility.first.patient_follow_ups.count`

Now, if we had a an extra `joins(:blood_pressures)` in `Patient.follow_ups`. It would end up sticking two `INNER JOINs` with `BloodPressures` because patients in Facility are already [through](https://github.com/simpledotorg/simple-server/pull/888/files#diff-a0e971b921f56b979cc5aa07fee7066fR20) `blood_pressures`.

Double joins still result in correct data, but make the query performance quadratic. The usual difference (on SBX) that we were seeing was like: 1.5s vs 12s.

The unsavoury result of this is if you're using this method from Patient class, you'd have to do `Patient.joins(:blood_pressures).follow_ups`

2. The group-by clause is not technically meaningless. This scope still allows you to chain more things in front. It just won't provide you with a realized result immediately. It'd expect some aggregate fn, more group-bys, or wheres and other SQLs.

spec/queries/district_analytics_query_spec.rb
```diff
@@ -109,7 +110,7 @@
     end
 
     describe '#registered_patients_by_period' do
-      it "should'nt count discarded patients" do
+      it "shouldn't count discarded patients" do
```
harimohanraj89, line 113: Consider "doesn't" or "excludes" to avoid the use of "should"

app/queries/district_analytics_query.rb
```diff
@@ -41,26 +41,12 @@ def registered_patients_by_period
...
-        .where(deleted_at: nil)
-        .group('facilities.id')
-        .group_by_period(@period, 'blood_pressures.recorded_at')
-        .where("patients.recorded_at < #{date_truncate_string}")
-        .order('facilities.id')
+      Patient
+        .joins(:blood_pressures)
+        .group('blood_pressures.facility_id')
+        .follow_ups(@period, last: @prev_periods)
+        .where(blood_pressures: { facility: facilities })
```
harimohanraj89, line 49: This is the dashboard page that shows district cohort info, right?

app/models/facility.rb
harimohanraj89: Why are we modifying this relationship? Since we now deal with BPs and BSs, should we instead add a second association like `facility.blood_pressure_patients`?

app/queries/facility_analytics_query.rb
```diff
@@ -37,28 +37,20 @@ def registered_patients_by_period
...
+                .follow_ups(@period, last: @prev_periods)
+                .group('user_id', 'blood_pressures.patient_id', BloodPressure.date_to_period_sql(@period), 'blood_pressures.recorded_at', 'patients.deleted_at')
+                .where(blood_pressures: { facility: @facility })
+                .select("DISTINCT ON (blood_pressures.patient_id, #{BloodPressure.date_to_period_sql(@period)}) blood_pressures.user_id AS user_id, patients.deleted_at")
+                .select("blood_pressures.recorded_at AS bp_recorded_at")
+                .order('blood_pressures.patient_id', BloodPressure.date_to_period_sql(@period), 'blood_pressures.recorded_at'),
+              'patients')
+        .group('user_id')
+        .group_by_period(:month, 'bp_recorded_at')
+        .count
```
harimohanraj89, line 53: This query looks more complex than what was here before. Do we need to change it at all? Could you shed some light on what's going on here?

app/queries/my_facilities/follow_ups_query.rb
```diff
@@ -1,62 +0,0 @@
-class MyFacilities::FollowUpsQuery
```
harimohanraj89, line 1: This isn't used anywhere else?

app/queries/user_analytics_query.rb
```diff
@@ -1,98 +0,0 @@
-class UserAnalyticsQuery
```
harimohanraj89, line 1: If I understand correctly, this stuff has moved to the presenter, which leverages simple aggregations on the `follow_ups` relation?

app/queries/district_analytics_query.rb
kitallis: Yup.

app/queries/my_facilities/follow_ups_query.rb
kitallis: Nope.

app/queries/user_analytics_query.rb
kitallis: Yeah, the Query abstraction was useful before because we were doing more complex queries. Not they are a lot simpler.

app/queries/facility_analytics_query.rb
kitallis: If we don't change it then we'll get that weird _multiple users in the same facility causing overcounted_ follow-ups data issue. I wouldn't want to keep that data issue around even if that page is not used by CVHOs because it's another point of confusion if someone clicks on it and might mistrust the supposedly correct data because now there's two sources of data.

I'll explain how this query works in another comment.

app/presenters/user_analytics_presenter.rb
```diff
@@ -9,52 +9,34 @@ class UserAnalyticsPresenter
   MONTHS_AGO = 6
   TROPHY_MILESTONES = [10, 25, 50, 100, 250, 500, 1_000, 2_000, 3_000, 4_000, 5_000]
   TROPHY_MILESTONE_INCR = 10_000
+  EXPIRE_STATISTICS_CACHE_IN = 30.minutes
```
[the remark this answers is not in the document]
kitallis: Fine with either. I'll change it to `15.minutes`

app/queries/district_analytics_query.rb
harimohanraj89: Ah yup, should have been clearer. Yes the tabular data that shown below cohort bars. 👍

app/models/blood_pressure.rb
```diff
@@ -64,4 +64,9 @@ def anonymized_data
       bp_diastolic: diastolic
     }
   end
+
+  def self.date_to_period_sql(period)
+    tz = Rails.application.config.country[:time_zone]
+    "(DATE_TRUNC('#{period}', (blood_pressures.recorded_at::timestamptz) AT TIME ZONE '#{tz}')) AT TIME ZONE '#{tz}'"
```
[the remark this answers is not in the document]
kitallis: This deserves a comment in the file I think, will push that in shortly.

app/views/api/v3/analytics/user_analytics/_monthly_gender_table.html.erb
kitallis: Yeah, we've exposed `follow_ups` under `grouped_by_date` which "un-segments" the gender-segmentation and pulls out [date-only segmentation](https://github.com/simpledotorg/simple-server/pull/888/files#diff-e9e2bbe1c12d40d2ea6f7a9d383a866aR145).

The mapping has also been inverted, previously we had a map that was like:

```ruby
{ male: { date_1: 2, date_2: 3} } 
```

Now we have, 

```ruby
{ [date, gender_1] => 3,
  [date_2, gender_1] => 4 } 
```

This can be used if you're showing gender + date split **and** just date-split by "reducing" over the genders.

app/presenters/user_analytics_presenter.rb
```diff
@@ -160,37 +123,75 @@ def trophy_stats
...
+        .group(:gender)
+        .count
+  end
+
+  def monthly_follow_ups
+    monthly_follow_ups_by_gender
+      .each_with_object({}) do |((date, _), count), by_date|
+        by_date[date] ||= 0
+        by_date[date] += count
+      end
```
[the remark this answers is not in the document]
kitallis: I know about `Hash#merge` block syntax (a presenter method was using it before). But don't think that applies here that well? We're reducing an existing map to an even smaller map in this case. See this [comment](https://github.com/simpledotorg/simple-server/pull/888/files#r414443159) for details.

It is possible to use it, if I merge it against itself. Don't think that's any more nicer though.

app/queries/facility_analytics_query.rb
kitallis: The `DISTINCT ON` accounts for the case where a patient visits a facility twice in a period and is seen by different users. Without it, the visit is counted twice and the total increases in comparison to facility follow ups. 

The notable clause being that we're `DISTINCT`ing `ON` the `BloodPressure.date_to_period_sql` **and** the user_id together.

However, we've turned this into a nested subquery using `.from` as a workaround for `ActiveRecord`'s inability to compose a `COUNT` with a `DISTINCT ON` – it ends up with garble like `COUNT DISTINCT DISTINCT ON` which is an invalid query since AR doesn't know how to handle `DISTINCT ON`.

This is a query I'm not happy with either. Though, I also don't want to spend additional time in making it easier to read. I've pushed this as a [comment](https://github.com/simpledotorg/simple-server/pull/888/commits/df832116799d90a81be106b5749927d7d02fbe73) in the code as well for some clarity.

app/presenters/user_analytics_presenter.rb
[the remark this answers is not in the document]
kitallis: Addressed [here](https://github.com/simpledotorg/simple-server/pull/888/commits/8abca6e2ad04e2b11778ae36777db676215f4f06).

app/models/blood_pressure.rb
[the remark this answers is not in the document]
kitallis: Addressed [here](https://github.com/simpledotorg/simple-server/pull/888/commits/8abca6e2ad04e2b11778ae36777db676215f4f06).

spec/queries/district_analytics_query_spec.rb
kitallis: Addressed [here](https://github.com/simpledotorg/simple-server/pull/888/commits/8abca6e2ad04e2b11778ae36777db676215f4f06).

app/models/facility.rb
kitallis: This is a good point.

Added a new association for [this](https://github.com/simpledotorg/simple-server/pull/888/commits/14fb637fd5b072dcce41e97afc5184c20ca63d8d).

The reason why I originally did it because I thought we'd naturally want to assume hypertensive as the default case. But it makes sense to be explicit instead.

There's a `diabetes_patients` scope in the upcoming [DM/HTN split in progress tab](https://github.com/simpledotorg/simple-server/pull/897).

app/models/patient.rb
harimohanraj89: I did a lot of digging to see if there's a way to say "please join this if not already joined", but looks like that concept is frowned upon because it messes up some internal table alias assignment process.

I think this is fine, good balance of simplicity and performance.

app/queries/facility_analytics_query.rb
harimohanraj89: Yikes indeed. The comment is very descriptive. 🚀

PR #895, opened 2020-04-21
Drop Audit Logs and related export code

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/172433826

## Because

Old AuditLogs make up a big chunk of our database. These logs are not exposed anywhere and are very slow to query. Current audit logs are written to files and exported to S3. 

## This addresses

Drop the `audit_logs` table and related code.

The work to import these old audit logs into our current logs system, and how to do so, is captured in [this follow-up story](https://www.pivotaltracker.com/story/show/168746200)


app/models/audit_log.rb
```diff
@@ -1,4 +1,6 @@
-class AuditLog < ApplicationRecord
+class AuditLog
```
harimohanraj89, line 1: Why is this sticking around?

db/migrate/20200421132003_drop_audit_logs.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.timestamps
+      t.datetime :deleted_at
+
+    end
+
+    add_index :audit_logs, [:auditable_type, :auditable_id]
+    add_index :audit_logs, [:action, :auditable_type], algorithm: :concurrently
+    add_reference :audit_logs, :user, type: :uuid, null: false
+  end
+end
```
harimohanraj89, line 22: Love it ❤️

app/models/audit_log.rb
vkrmis: Our logging functions for AuditLogs are all class methods on this model(`fetch_log, merge_log, and login_log`), and keeping this as a Model lets us continue to use the validations.

harimohanraj89: Never mind I understand now.

PR #897, opened 2020-04-23
Add Diabetes support to Progress Tabs

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/172471609

## Because

An improvement over the new progress tabs that allows one to see a HTN/DM split in their progress if the Facility has diabetes management enabled.

## This addresses

Adds additional drop-downs and filters to show exclusively hypertension or exclusively diabetes registered/follow-up patients.

ptab

## Technical notes

* Add scopes on `Patient` to calculate `follow_ups`
* Add delegates on `Facility` to calculate `follow_ups` scoped by Facility
* Add scopes on `Patient` to distinguish between hypertensive/diabetic patients
* Modifies the `DistrictAnalyticsQuery` and `FacilityAnalyticsQuery` to be hypertension only

## Data definitions

Refer to the filters in the screenshot above.

#### `All` for registered patients

Patients who were registered in that Facility (regardless of diagnosis)

#### `Diabetes:All` for registered patients

Patients who had diagnosis as `diabetes=yes` registered in that Facility

#### `Hypertension:All` for registered patients

Patients who had diagnosis as `hypertension=yes` registered in that Facility

#### `All` for follow up patients

Patients who, regardless of their diagnosis followed up in that Facility

#### `Diabetes:All` for follow up patients

Patients who had diagnosis as `diabetes=yes` and followed up in that Facility

#### `Hypertension:All` for follow up patients

Patients who had diagnosis as `hypertension=yes` and followed up in that Facility


app/models/encounter.rb
```diff
@@ -1,12 +1,13 @@
...
   include Mergeable
+  include SQLHelpers
 
   belongs_to :patient, optional: true
   belongs_to :facility
 
   has_many :observations
-  has_many :blood_pressures, through: :observations, source: :observable, source_type: 'BloodPressure'
   has_many :blood_sugars, through: :observations, source: :observable, source_type: 'BloodSugar'
+  has_many :blood_pressures, through: :observations, source: :observable, source_type: 'BloodPressure'
```
timcheadle, line 10: Why this change?

app/models/facility.rb
```diff
@@ -3,25 +3,28 @@
...
   include Mergeable
   include QuarterHelper
-  extend FriendlyId
 
-  attribute :import, :boolean, default: false
-  attribute :organization_name, :string
-  attribute :facility_group_name, :string
-
-  belongs_to :facility_group, optional: true
+  extend FriendlyId
```
timcheadle, line 7: This entire file looks like it was just reorganized? The diff makes it hard to see if there were any logic changes. Let's try and keep reformatting to different PRs.

app/models/patient.rb
```diff
@@ -17,52 +17,71 @@ class Patient < ApplicationRecord
...
+      .diabetes_only
+  }
+
+  scope :hypertension_follow_ups, -> (period, last: nil) {
+    follow_ups_with(BloodPressure, period, last: last)
+      .hypertension_only
+  }
+
+  def self.follow_ups_with(model_name, period, time_col: 'recorded_at', last: nil)
+    model_name_sym = model_name.name.underscore.pluralize.to_sym
```
timcheadle, line 74: You can just use AcitveRecord's `model_name.table_name` here

app/models/patient.rb
```diff
@@ -17,52 +17,71 @@ class Patient < ApplicationRecord
...
+    follow_ups_with(BloodSugar, period, last: last)
+      .diabetes_only
+  }
+
+  scope :hypertension_follow_ups, -> (period, last: nil) {
+    follow_ups_with(BloodPressure, period, last: last)
+      .hypertension_only
+  }
+
+  def self.follow_ups_with(model_name, period, time_col: 'recorded_at', last: nil)
```
timcheadle, line 73: Can we make this attribute `time_column`? No need to abbreviate.

app/models/encounter.rb
kitallis: Reverted. I was using BloodSugars before BloodPressures everywhere so I changed it about. It's not a useful change.

app/models/facility.rb
kitallis: Reverted. The formatting in this file was not sensible in any way, but it makes sense to keep this re-organization separate. 👍

app/views/api/v2/analytics/user_analytics/_monthly_gender_table.html.erb
```diff
@@ -1,48 +1,137 @@
...
+  </table>
+<% end %>
+
+<% Patient::GENDERS.each do |gender| %>
+  <!-- gender-wise tables -->
+  <table class="progress-table <%= data_table_name %> diabetes:<%= gender %>" style="display: none">
+    <tr class="total">
+      <td class="row-label">All-time</td>
+      <td>
+        <%= zero_if_unavailable(statistics.dig(:all_time,
```
rsanheim, line 63: All of these deeply nested statistics calls are hard to read and follow and also couple us to the exact structure of the data.

We have a presenter available already for all these views I believe. Lets use that and make the views way simpler and a bit less coupled:

Something like

```ruby
class UserAnalyticsPresenter

  def all_time_stats_grouped_by_gender(htn_or_dm, resource, gender)
  # this can handle the zero logic and the dig logic...and whatever else
...
  def monthly_grouped_by_gender(htn_or_dm, resource, gender)
...
```

Then in the views:

```
  <table class="progress-table <%= data_table_name %> diabetes:<%= gender %>" style="display: none">
    <tr class="total">
      <td class="row-label">All-time</td>
      <td>
             <%= user_analytics.monthly_grouped_by_gender(:hypertension, resource, gender) %>
       </td>
```

And so on down the line for all the views.

app/views/api/v3/analytics/user_analytics/show.html.erb
```diff
@@ -14,178 +14,184 @@
...
+      <button onclick="nextSlide(-1)" class="button-prev">
+        <span><%= inline_svg('icon_arrow_forward.svg') %></span>
+      </button>
+    </div>
 
-      <!-- dump all statistics in a hidden div so that we can parse it in JS -->
-      <%= content_tag :div, id: "statistics", data: { statistics: @statistics }, style: "display: none" do %>
-      <% end %>
+    <!-- dump all statistics in a hidden div so that we can parse it in JS -->
+    <%= content_tag :div, id: "statistics", data: { statistics: @statistics }, style: "display: none" do %>
```
rsanheim, line 31: whoa, I see this was here already...but how much data can this be? how big will the page size get?

app/views/api/v3/analytics/user_analytics/show.html.erb
```diff
@@ -14,178 +14,184 @@
...
-                                                      :grouped_by_date,
-                                                      :follow_ups,
-                                                      month_date)) %>
-            </td>
-          </tr>
-        <% end %>
-      </table>
-    </div>
+          <td>
+            <%= zero_if_unavailable(@statistics.dig(:monthly,
```
rsanheim, line 136: Same comments as above, given we are five layers deep in a nested structure, these should be moved to the presenter.

db/schema.rb
```diff
@@ -570,6 +595,138 @@
      FROM (appointments ap
        JOIN facilities f ON ((ap.facility_id = f.id)));
   SQL
+  create_view "latest_blood_pressures_per_patient_per_days", materialized: true, sql_definition: <<-SQL
```
rsanheim, line 598: Should all these views be in the diff? /cc https://github.com/simpledotorg/simple-server/pull/939

spec/controllers/api/v3/analytics/user_analytics_controller_spec.rb
```diff
@@ -25,87 +25,123 @@
...
 
-        it 'has the hypertension control card' do
-          get :show, format: :html
+          it 'has the follow-ups card' do
+            get :show, format: :html
 
-          expect(response.body).to match(/Hypertension control/)
-        end
+            expect(response.body).to match(/Follow-up patients/)
+          end
```
rsanheim, line 63: You can combine all of these expectations into one spec. Since we are rendering views here, these specs are going to be really slow. Its definitely worth it to combine simple assertions like this to one spec for real heavy things like this.

spec/models/facility_spec.rb
```diff
@@ -7,22 +7,177 @@
...
-    it 'has distinct patients' do
-      facility = create(:facility)
-      patient = create(:patient)
-      blood_pressures = create_list(:blood_pressure, 2, facility: facility, patient: patient)
-      blood_sugars = create_list(:blood_sugar, 2, facility: facility, patient: patient)
-      (blood_pressures + blood_sugars).each { |record| create(:encounter, :with_observables, patient: patient, observable: record, facility: facility) }
-      expect(facility.patients.count).to eq(1)
+    context 'patients' do
+      it 'has distinct patients' do
+        facility =
```
rsanheim, line 16: Whats with the formatting here? Why are these multiline?

spec/models/facility_spec.rb
```diff
@@ -7,22 +7,177 @@
...
+
+        create(:encounter, :with_observables, observable:
+          create(:blood_sugar, facility: facility, patient: dm_patient))
+        create(:encounter, :with_observables, observable:
+          create(:blood_sugar, facility: facility, patient: htn_patient))
+
+        create(:encounter, :with_observables, observable:
+          create(:blood_pressure, facility: facility, patient: htn_patient))
+        create(:encounter, :with_observables, observable:
+          create(:blood_pressure, facility: facility, patient: dm_patient))
```
rsanheim, line 32: I think its time for `create(:blood_pressure, :with_encounter)` ...these factories are getting really unwieldly like this.

spec/models/patient_spec.rb
```diff
@@ -76,49 +82,191 @@
...
+        expect(Patient.diabetes_only.count).to eq(2)
+      end
+    end
+
+    describe '.hypertension_only' do
+      it 'only  includes patients with diagnosis of hypertension' do
+        create_list(:patient, 2, :hypertension)
+        create(:patient, :diabetes)
+
+        expect(Patient.hypertension_only.count).to eq(2)
```
rsanheim, line 99: This would be a better spec if you asserted the records directly rather than a count with `match_array`.

spec/presenters/user_analytics_presenter_spec.rb
```diff
@@ -1,241 +1,663 @@
...
+
+            create(:encounter,
+                   :with_observables,
+                   observable: create(:blood_sugar,
+                                      patient: patient,
+                                      facility: current_facility,
+                                      user: current_user,
+                                      recorded_at: follow_up_date))
+
+            create(:encounter,
```
rsanheim, line 53: More factories where we should really let the blood_pressure drive, and let encounter be hidden behind the scene.

app/views/api/v2/analytics/user_analytics/_monthly_gender_table.html.erb
kitallis: I went back and forth on this. I'd wanted the statistics map to remain frozen and hence not require any further method abstractions. This "map" is contractual and should be a big deal if changed.

But I see readability concerns and I'll go ahead and stick them in methods.

spec/models/facility_spec.rb
kitallis: Just did it for readability, not an important change. I'll go and revert it.

spec/presenters/user_analytics_presenter_spec.rb
kitallis: I definitely want to do this, I punted on it since that factory is a little tricky to create.

I'll give it a shot and keep us posted here.

kitallis:
> This is a lot of code to add diabetes support, which I know our app really isn't focused on at all.

Ha, yeah. This was a lot more non-trivial work than I had imagined. Esp. making sure the diabetes code isn't leaked all over. I think it's fairly restricted at the moment to Progress Tabs which are a relatively isolated piece of code in the first place. I'm fairly happy with the scopes that allow this.

app/views/api/v3/analytics/user_analytics/show.html.erb
kitallis: It shouldn't be that big and would remain fairly constant, since the keys are the only real data. The data primarily has a bunch of dates and counts.

I did a quick check and it should almost always roughly be in the range of ~2kb.

kitallis: This page is rendered as a webview and has a bunch of JS that operates on the data. So it's important that it can work offline.

app/presenters/user_analytics_presenter.rb
```diff
@@ -45,41 +44,72 @@ def display_percentage(numerator, denominator)
...
+  end
+
   def daily_stats
+    diabetes_enabled? ? daily_total_stats : daily_htn_stats
+  end
+
+  def daily_htn_stats
+    follow_ups =
+      current_facility
+        .hypertension_follow_ups(:day, last: DAYS_AGO)
```
harimohanraj89, line 60: Should we rename these methods to `hypertension_follow_ups_by_period`? Contrasted against the registrations calculation below, it might be more intuitive?

app/presenters/user_analytics_presenter.rb
```diff
@@ -122,79 +154,181 @@ def trophy_stats
...
+      by_date[date] ||= 0
+      by_date[date] += count
+    end
+  end
+
+  def sum_by_gender(data)
+    data.each_with_object({}) do |((_, gender), count), by_gender|
+      by_gender[gender] ||= 0
+      by_gender[gender] += count
+    end
```
harimohanraj89, line 332: I'm having a hard time grokking the changes in this file. I see certain gender groupings moved around, I see some data structures (like `all_time_stats`) changing significantly. Could you help me understand what's going on here, and how it supports DM?

app/models/patient.rb
```diff
@@ -51,18 +51,35 @@ class Patient < ApplicationRecord
...
 
-  #
-  # Note: This scope expects a join(:blood_pressures) to exist.
-  # For eg, Patient.joins(:blood_pressures).follow_ups(:month).
-  #
-  # It doesn't include the join in this scope to play well with certain parent scopes (eg. Facility).
-  # Parent scopes might auto add this join and the final query would end up with unnecessary joins, affecting perf.
-  #
+  scope :diabetes_only, -> { joins(:medical_history).merge(MedicalHistory.diabetes_yes) }
+  scope :hypertension_only, -> { joins(:medical_history).merge(MedicalHistory.hypertension_yes) }
```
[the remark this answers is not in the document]
kitallis: Makes sense, I think `with_hypertension` works. I don't want to use `diagnosed_with_hypertension` because it can be confused with the field in MedicalHistory called `diagnosed_with_hypertension`.

app/presenters/user_analytics_presenter.rb
kitallis: Shall I walk you through it over a call?

app/models/facility.rb
```diff
@@ -20,10 +20,20 @@ class Facility < ApplicationRecord
...
-  has_many :hypertension_patients, -> { distinct }, through: :blood_pressures, source: :patient
   has_many :prescription_drugs
   has_many :appointments
-  has_many :registered_patients, class_name: "Patient", foreign_key: "registration_facility_id"
+
+  has_many :registered_patients,
+           class_name: "Patient",
+           foreign_key: "registration_facility_id"
+  has_many :registered_diabetes_patients,
+           -> { diabetes_only },
```
harimohanraj89, line 30: Using the `only` scope here ensures that a patient can't show up in both `registered_diabetes_patients` _and_ `registered_hypertension_patients`?

This asymmetry seems a bit odd to me. Should this just be `-> { with_diabetes }`?

kitallis: This is just a bad merge. It's supposed to be `with_diabetes`. Hopefully the test should break any time now...

app/models/patient.rb
```diff
@@ -51,18 +51,35 @@ class Patient < ApplicationRecord
...
+      .with_diabetes
+  }
+
+  scope :hypertension_follow_ups_by_period, -> (period, last: nil) {
+    follow_ups_with(BloodPressure, period, last: last)
+      .with_hypertension
   }
 
+  def self.follow_ups_with(model_name, period, time_column: 'recorded_at', last: nil)
+    table_name = model_name.table_name.to_sym
```
harimohanraj89, line 72: Do we need `to_sym`? Is it just being `to_s`'d below anyway?

app/presenters/user_analytics_presenter.rb
```diff
@@ -1,6 +1,5 @@
-class UserAnalyticsPresenter
+class UserAnalyticsPresenter < Struct.new(:current_facility)
```
harimohanraj89, line 1: Can we put this back in an `initialize`? It's a bit plainer to read, and scalable too.

app/queries/district_analytics_query.rb
```diff
@@ -43,12 +43,10 @@ def registered_patients_by_period
   def follow_up_patients_by_period
     @follow_up_patients_by_period ||=
       Patient
-        .joins(:blood_pressures)
         .group('blood_pressures.facility_id')
-        .follow_ups(@period, last: @prev_periods)
+        .hypertension_follow_ups_by_period(@period, last: @prev_periods)
```
harimohanraj89, line 47: Is a corresponding change required in a cohort analytics query? Even if that query uses BPs to do stuff, we should explicitly add a `with_hypertension` filter on it.

app/queries/facility_analytics_query.rb
```diff
@@ -38,28 +38,36 @@ def registered_patients_by_period
...
+                .group('bp_user_id',
+                       'blood_pressures.patient_id',
+                       BloodPressure.date_to_period_sql('blood_pressures.recorded_at', @period),
+                       'blood_pressures.recorded_at',
+                       'patients.deleted_at')
                 .where(blood_pressures: { facility: @facility })
-                .select("DISTINCT ON (blood_pressures.patient_id, #{BloodPressure.date_to_period_sql(@period)}) blood_pressures.user_id AS user_id, patients.deleted_at")
+                .select(%Q(DISTINCT ON (blood_pressures.patient_id,
+                                        #{BloodPressure.date_to_period_sql('blood_pressures.recorded_at', @period)})
+                           blood_pressures.user_id AS bp_user_id, patients.deleted_at))
```
harimohanraj89, line 64: This is a bit hard to read. Can we reformat for ease, since we're using %Q anyway?

```
.select(%Q(
  DISTINCT ON
    (blood_pressures.patient_id, #{BloodPressure.date_to_period_sql(@period)}),
    blood_pressures.user_id,
    patients.deleted_at
))

```

harimohanraj89: Additionally, why is the patient ID and date in sub-parens? `()`? What does that do?

app/views/api/v2/analytics/user_analytics/_gender_filter.html.erb
```diff
@@ -1,13 +1,19 @@
 <form>
   <select class="card-dropdown <%= data_table_name %>" onchange='filterDataByGender("<%= data_table_name %>")'>
-    <% hypertension_option_text = statistics.dig(:metadata, :is_diabetes_enabled) ? 'Hypertension: ' : '' %>
+    <% if is_diabetes_enabled %>
+      <option value="all">
+        All
+      </option>
```
harimohanraj89, line 6: Let's single line this for consistency with the other options. Makes it easier to parse.

app/views/api/v2/analytics/user_analytics/show.html.erb
```diff
@@ -82,27 +82,33 @@
...
+                         statistics: @statistics,
+                         data_table_name: registrations_table_name,
+                         monthly_period_list: @user_analytics.monthly_period_list,
+                         is_diabetes_enabled: @statistics.dig(:metadata, :is_diabetes_enabled) } %>
   </div>
 
   <!-- monthly follow-up patients view -->
   <div class="card">
-    <h3>Follow-up hypertension patients</h3>
+    <h3>Follow-up<%= @statistics.dig(:metadata, :is_diabetes_enabled) ? ' ' : ' hypertension ' %>patients</h3>
```
harimohanraj89, line 98: For readability, consider:

```
<% if @statistics.dig(:metadata, :is_diabetes_enabled) %>
  <h3>Follow-up hypertension patients</h3.>
<% else %>
  <h3>Follow-up patients</h3.>
<% end %>
```

app/models/patient.rb
kitallis: It's for the `joins`. If you pass in a string to the `joins` it doesn't scope it against the class correctly. I can `to_sym` in the `joins` if that's better? Don't have a particular preference. It's easier to begin with symbols since they seem to have a special meaning instead of remembering to `to_sym` everywhere.

app/presenters/user_analytics_presenter.rb
kitallis: These can scale with the number of arguments as well. I think a struct works here quite well since it's a decorator with only one object you are decorating over. 

This is a pretty standard pattern?

app/queries/district_analytics_query.rb
kitallis: Are you talking about something different from what is already covered in https://github.com/simpledotorg/simple-server/pull/940?

app/presenters/user_analytics_presenter.rb
```diff
@@ -43,43 +34,80 @@ def display_percentage(numerator, denominator)
     "#{percentage.round(0)}%"
   end
 
+  def daily_period_list
```
harimohanraj89, line 37: This class and these methods make a lot of sense to me 👍 The diff is ugly, but that's ok 😄

app/presenters/user_analytics_presenter.rb
harimohanraj89: I haven't used it much but it seems pretty neat. I don't feel strongly about it, let's go with `Struct` 👍

app/queries/district_analytics_query.rb
harimohanraj89: Nope, that'll do it. Why do we address the district analytics query in this PR instead of #940?

app/queries/facility_analytics_query.rb
kitallis: Ugh, I was hoping I didn't need to explain this query. Can I do this separately? I just really want to delete this code.

app/queries/district_analytics_query.rb
kitallis: That's a good question.

The trouble is that it was originally using what is now the `encounters` version of the query. If we leave it be then various tests will break because of the strong coupling between the methods used in `DistrictAnalytics` and `Patient`. Instead of spending that effort trying to fix something that needs to get reverted anyway, I made the change right here.

The CohortAnalytics don't have this problem because patients are passed in.

kitallis:
@rsanheim @harimohanraj89 Thanks for the review.

I have not addressed:

* Adding `with_encounter` trait in this PR. This is being done in https://github.com/simpledotorg/simple-server/pull/940.
* Pulling the `dig` from the `statistics` map up into presenter function. I attempted it and I think it adds very little to the readability and a lot of verbosity to the presenter. I find the `dig` a lot simpler to reason about because:

a. `statistics` **is** the presenter layer data-structure, it is meant for the views, wrapping it in other one-off methods adds unnecessary indirection to what is a very explicit tree that one can traverse.

b. The presenter methods are all one-off traversals, they don't repeat, hence you end up with 10-12 presenter functions that are harder to describe than a pretty descriptive data-structure.

c. It shaves off very little in terms of code, you still have to pass in 3-4 arguments to all these presenter functions.

d. I don't think moving them into presenter functions reduces touch-points in the view. You'd most likely have to rename methods, pass in different arguments, or iterate over dates differently if you meddle with the data-structure.

What do you think?

app/queries/district_analytics_query.rb
harimohanraj89: Ooh neat. Yup, we're gonna put this all out together anyway, so let's do what's smoothest. 👍

harimohanraj89:
Sup @kitallis 

Re factory cleanup: If we're doing that in #940, let's remember to clean up the test data we're introducing here. If it's too tedious to remember and revisit, then let's go ahead and cherry-pick the changes from #940 into here right away.

Re "dig": There's one advantage that I think semantically named presenter methods will introduce. Right now, the view deals with markup complexity _and_ data structure traversal. This stresses my brain out a bit, if I imagine myself coming back here and making changes later. If extracted into well-named methods (we can replace arguments with methods - eg. `monthly_stats(:htn) -> monthy_htn_stats`), then the view _only_ deals with markup. While the traversal complexity is just moved to the presenter, that also means it lives right next to the _generation_ of the data structure as well. It seems more cognitively coherent to me.

What do you think? Of course, my suggestion is theoretical. You've tried it out, so if this looks ugly in practice, then we can avoid it.

kitallis:
> Sup @kitallis
> 
> Re factory cleanup: If we're doing that in #940, let's remember to clean up the test data we're introducing here. If it's too tedious to remember and revisit, then let's go ahead and cherry-pick the changes from #940 into here right away.
> 
> Re "dig": There's one advantage that I think semantically named presenter methods will introduce. Right now, the view deals with markup complexity _and_ data structure traversal. This stresses my brain out a bit, if I imagine myself coming back here and making changes later. If extracted into well-named methods (we can replace arguments with methods - eg. `monthly_stats(:htn) -> monthy_htn_stats`), then the view _only_ deals with markup. While the traversal complexity is just moved to the presenter, that also means it lives right next to the _generation_ of the data structure as well. It seems more cognitively coherent to me.
> 
> What do you think? Of course, my suggestion is theoretical. You've tried it out, so if this looks ugly in practice, then we can avoid it.

@harimohanraj89 For reference, I've partially migrated some calls to these extractor functions in this commit https://github.com/simpledotorg/simple-server/pull/897/commits/2b03c960dbd60fb2cb8c9bf0a50fd0496e9b6831 – I haven't fully migrated anything to just get a taste of what it would look like.

An immediate thing that emerges is that these can't really be purely "view" functions since some operation like _calculate percentage_ or _pluralize based on counts_ need raw integers to work with. View functions aren't composable.

Otherwise, they shave off some nested digging as expected.

app/models/patient.rb
harimohanraj89: Meh, was just looking for cruft. This is used 👍

harimohanraj89:
@kitallis I think we can push this approach to the extreme. I tried to explain my thoughts in words but I failed miserably, so I pushed an example commit instead.

It shifts complexity to the presenter, yes, but I _love_ how clean the view code gets.

I know this is only monthly HTN, and there are several more permutations. The `UserAnalyticsPresenter` class will have a bunch of new methods, for each datapoint we're displaying, but it's a presenter, so I think it's appropriate. What do you think?

kitallis:
> @kitallis I think we can push this approach to the extreme. I tried to explain my thoughts in words but I failed miserably, so I pushed an [example commit](https://github.com/simpledotorg/simple-server/commit/19cec2624a16d9def938c8a6f318a3f156bb24e7) instead.
> 
> It shifts complexity to the presenter, yes, but I _love_ how clean the view code gets.
> 
> I know this is only monthly HTN, and there are several more permutations. The `UserAnalyticsPresenter` class will have a bunch of new methods, for each datapoint we're displaying, but it's a presenter, so I think it's appropriate. What do you think?
> 
> Is the "need raw integers" remark influenced by the behavior of `zero_if_unavailable`? We can simply return an integer there to solve that, as the display percentage method is smart enough to handle zeros.

Yeah, I think this can be done wholesale, my commit was just representative. However, if we were to do this, I would remove the `statistics` method entirely, you'd just make methods out of the data and not have the data structure in the first place: _data --> behaviour_.

I'll make this change.

harimohanraj89:
Great point about the usefulness of `statistics`. We could also privatize `statistics` and cache it. I'll let you feel that out 👍

kitallis:
> Great point about the usefulness of `statistics`. We could also privatize `statistics` and cache it. I'll let you feel that out 👍

Cool. 

Generally, I think we should try to avoid large design-changes late in the PRs because that causes a lot of attrition – something to keep in mind for all of us for later.

app/views/api/v2/analytics/user_analytics/show.html.erb
```diff
@@ -14,7 +14,9 @@
 
 <body>
 <div class="progress-contents">
-  <div class="updated-date">Updated: <span><%= @statistics.dig(:metadata, :last_updated_at) %></span></div>
+  <div class="updated-date">
+    Updated: <span><%= @user_analytics.statistics.dig(:metadata, :last_updated_at) %></span>
```
harimohanraj89, line 18: One last extraction for consistency?

db/schema.rb
```diff
@@ -178,6 +178,7 @@
     t.datetime "deleted_at"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.index ["deleted_at"], name: "e_d_at"
```
harimohanraj89, line 181: Oh hello there. Is this an unintentional addition?

PR #901, opened 2020-04-24
Facility breakdowns

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172096574

## Because

CVHOs would like a per-facility breakdown of district cohort reports when downloaded.

## This addresses

This PR makes a few major changes to achieve this:
* Updates the data structure of the cohort analytics query
* Flattens the district cohort report into a single row in the CSV download
* Adds additional rows to the download for every facility in the district.

**Note:** The output formatting of the CSV may change a bit.

**Deployment checklist**
* [ ] Clear dashboard caches


app/views/analytics/districts/show.csv.erb
```diff
@@ -3,55 +3,61 @@
...
+    followed_up,
+    controlled,
+    uncontrolled,
+    defaulted,
+    "#{controlled_percent}%",
+    "#{uncontrolled_percent}%",
+    "#{default_percent}%",
+    nil
+  ]
+end).to_csv -%>
```
harimohanraj89, line 60: Changes up to this point are adding per-facility rows for the cohort report

app/views/analytics/districts/show.csv.erb
```diff
@@ -63,7 +69,7 @@
   dates = dates_for_periods(@period, 3, include_current_period: @show_current_period)
   ["Follow up patients", "Registered patients"].each do |metric|
     dates.each do |date|
-      headers << "#{metric} #{date}"
+      headers << "#{metric} #{format_period(@period, date)}"
```
harimohanraj89, line 72: `Registrations 2020-04-01` -> `Registrations Apr 2020`

app/controllers/analytics/districts_controller.rb
```diff
@@ -82,6 +83,12 @@ def set_dashboard_analytics(period, prev_periods)
       end
   end
 
+  def set_facility_keys
+    facilities = @organization_district.facilities.order(:name).pluck(:id, :name).to_h
+
+    @facility_keys = { total: "Total" }.merge(facilities).with_indifferent_access
+  end
```
harimohanraj89, line 90: We pre-fetch a list of facilities in alphabetical order to help generate the rows of data in the CSV in order.

app/views/analytics/facilities/show.csv.erb
```diff
@@ -18,11 +18,11 @@
...
-    registered = analytics[:registered]
-    followed_up = analytics[:followed_up]
-    defaulted = analytics[:defaulted]
-    controlled = analytics[:controlled]
-    uncontrolled = analytics[:uncontrolled]
+    registered = analytics.dig(:registered, :total) || analytics[:registered]
+    followed_up = analytics.dig(:followed_up, :total) || analytics[:followed_up]
+    defaulted = analytics.dig(:defaulted, :total) || analytics[:defaulted]
+    controlled = analytics.dig(:controlled, :total) || analytics[:controlled]
+    uncontrolled = analytics.dig(:uncontrolled, :total) || analytics[:uncontrolled]
```
harimohanraj89, line 25: We leave the old `analytics[:registered]` method in here to ensure that the code still works with the older data structure that'll be in cache when we deploy this change. 48 hours after we deploy this change, we can clean up the second clause.

app/queries/cohort_analytics_query.rb
```diff
@@ -41,13 +38,24 @@ def patient_counts(cohort_start, cohort_end, report_start, report_end)
...
-      defaulted: registered_patients.size - followed_up_patients.size,
-      controlled: controlled_patients.size,
-      uncontrolled: uncontrolled_patients.size
-    }
+      registered: { total: registered_patients.size, **registered_counts },
+      followed_up: { total: followed_up_patients.size, **followed_up_counts },
+      defaulted: { total: registered_patients.size - followed_up_patients.size, **defaulted_counts },
+      controlled: { total: controlled_patients.size, **controlled_counts },
+      uncontrolled: { total: uncontrolled_patients.size, **uncontrolled_counts }
+    }.with_indifferent_access
```
kitallis, line 58: Consider removing the individual `symbolize_keys` and do a `deep_symbolize_keys` at the very end.

kitallis:
I've only partially reviewed this, will take another pass at it on 30.04.2020.

app/queries/cohort_analytics_query.rb
```diff
@@ -74,6 +82,6 @@ def followed_up(registered_patients, report_start, report_end)
   end
 
   def controlled(patients)
-    patients.select { |p| p.bp_systolic < 140 && p.bp_diastolic < 90 }
+    patients.where('newest_bps.systolic < 140 AND newest_bps.diastolic < 90')
```
prabhanshuguptagit, line 85: 👌

app/queries/cohort_analytics_query.rb
harimohanraj89: I tried making this change, and remembered why it's this way. The splat (`**registered_counts`) requires the keys to be symbols, and I find the splat operator a lot more readable and scalable than merging hashes together.

kitallis: Fair enough.

harimohanraj89:
The cohort analytics query object accepts a relation of patients to perform its operations on. I think the best spot to apply HTN filtering would be right before we pass a relation of patients into this object (somewhere in [this vicinity](https://github.com/simpledotorg/simple-server/blob/master/app/models/organization_district.rb#L17-L20) to start). That's a single place to make the modification that would reflect on all cohort reporting, both district and facility level.

PR #902, opened 2020-04-24
Add the new English SMS reminders

kitallis:
## Because

COVID-19 and,

The previous [attempt at translating them](https://github.com/simpledotorg/simple-server/pull/889) failed because we eventually [reverted](https://github.com/simpledotorg/simple-server/pull/891) it subsequently, thereby re-writing the original SMSes onto transifex back again. So the new translations didn't actually get translated at all.

Transifex only picks up changes from `master`.

## This addresses

Updates the English SMS reminder, all other languages defaulted to English. The reminders are currently disabled on production, so this can be merged to `master` safely.


harimohanraj89:
@kitallis Could you explain to me why our previous PR didn't work, and why we're doing this again, and what's different this time around? Coudl we also link to the previous PR in the description?

kitallis:
> @kitallis Could you explain to me why our previous PR didn't work, and why we're doing this again, and what's different this time around? Coudl we also link to the previous PR in the description?

Don.e

PR #906, opened 2020-04-27
Add guard

rsanheim:
[no story]

## Because

[Guard](https://github.com/guard/guard) makes testing and developing quicker!

Note that this also required and update of Pry, as there was a conflict with the previously locked older version and Guard.

## This addresses

Adds guard-rspec and the generated Guardfile.


Guardfile
```diff
@@ -0,0 +1,70 @@
...
+  # Rails config changes
+  watch(rails.spec_helper)     { rspec.spec_dir }
+  watch(rails.routes)          { "#{rspec.spec_dir}/routing" }
+  watch(rails.app_controller)  { "#{rspec.spec_dir}/controllers" }
+
+  # Capybara features specs
+  watch(rails.view_dirs)     { |m| rspec.spec.call("features/#{m[1]}") }
+  watch(rails.layouts)       { |m| rspec.spec.call("features/#{m[1]}") }
+
+  # Turnip features and steps
```
timcheadle, line 65: What is Turnip? Do we need this stanza?

rsanheim: We don't, I removed it.

PR #907, opened 2020-04-27
Search filters for overdue patients

rsanheim:
**Story card:** https://www.pivotaltracker.com/n/projects/2184102/stories/172537522/

## Because

We want more filtering abilities on the overdue list.

## This addresses

This adds search filters to the overdue listing. I ended up extracting to a small query class to keep things a bit cleaner.

Because there are two different filters w/ different forms on the page, we needed to add hidden field tags to maintain state.  See the inline code comments for details.  It isn't ideal, but combining the forms was beyond my bootstrap / layout abilities.

## TODO

* [x] Extract the filtering piece out of the controller
* [x] ask about auto applying filters vs having a button
* [x] Use `form` attribute to resolve two forms problem
* [x] Fix up mobile view filter display
* [x] Extract the filters to a helper or some other better place
* [x] Handle the 'has phone number' and 'no phone number' case


app/controllers/appointments_controller.rb
```diff
@@ -2,22 +2,25 @@ class AppointmentsController < AdminController
...
 
     if current_facility
       @patient_summaries = @patient_summaries.where(next_appointment_facility_id: current_facility.id)
     end
+    @patient_summaries = @patient_summaries.order(risk_level: :desc, next_appointment_scheduled_date: :desc)
 
     respond_to do |format|
       format.html do
         @patient_summaries = paginate(@patient_summaries).includes(:next_appointment, patient: :appointments)
+        render layout: "overdue"
```
rsanheim, line 23: Need a new layout for the new sidebar.

app/controllers/appointments_controller.rb
```diff
@@ -38,6 +41,37 @@ def update
 
   private
 
+  def apply_search_filters
```
rsanheim, line 44: Here is the query logic - eventually I'd like to have this in a `PatientSummarySearch` object or something, but want to know where else we are going to be searching patients like this w/ other dashboards.

app/controllers/appointments_controller.rb
```diff
@@ -38,6 +41,37 @@ def update
...
+    end
+    if @search_filters.include?("high_risk")
+      @patient_summaries = @patient_summaries.where("risk_level == 1")
+    end
+    @patient_summaries
+  end
+
+  def set_patient_search_filters
+    @patient_search_filters = {
+      "only_less_than_year_overdue" => "< 365 days overdue",
```
rsanheim, line 64: This is awkward, need to ask about doing the inverse of this filter.

app/views/layouts/overdue.html.erb
```diff
@@ -0,0 +1,99 @@
+<% content_for(:content) do %>
```
rsanheim, line 1: This new layout was mostly modeled after the my facilities layout, which has a similar sidebar.

app/views/layouts/overdue.html.erb
```diff
@@ -0,0 +1,99 @@
...
+  DESKTOP NAV
+  --------------------->
+  <div class="row">
+    <div class="col-lg-2 pl-lg-0 desktop">
+      <section class="secondary-nav-filters">
+        <h4>Patients</h4>
+
+        <%= form_tag appointments_path, method: :get do %>
+          <%= hidden_field_tag :facility_id, @facility_id %>
+          <%= hidden_field_tag :per_page, @per_page %>
```
rsanheim, line 65: We needed to do this because of our two forms and how they are using GET requests.  See https://github.com/simpledotorg/simple-server/pull/907/commits/0275fc5be68b7a0d10e5725b3b591f9a0fe1031e for details.

app/views/shared/_paginate_and_filter_by_facility.html.erb
```diff
@@ -23,5 +23,8 @@
       %>
     </div>
   </div>
+  <% (@search_filters || []).each do |value| %>
+    <%= form.hidden_field "search_filters[]", value: value %>
```
rsanheim, line 27: Here is the other side of the state madness: we need to include search fields as hidden fields on the facilities  / pagination filter. Two forms using GET requests and managing state is nasty. 😢

spec/spec_helper.rb
```diff
@@ -11,6 +11,9 @@
 RSpec.configure do |config|
   SimpleCov.start
 
+  config.filter_run focus: true
+  config.run_all_when_everything_filtered = true
```
rsanheim, line 15: Just making rspec smarter so it respects focused specs.  Makes guard much nicer to use.

app/views/layouts/overdue.html.erb
```diff
@@ -0,0 +1,100 @@
...
+                        <%= label %>
+                      </label>
+                    </div>
+                  <% end %>
+                <% end %>
+              </div>
+            </div>
+          </div>
+          <div class="modal-footer">
+            <a href="?" type="button" class="btn btn-secondary">Clear all</a>
```
harimohanraj89, line 46: Is there a Clear All link in the desktop version? Does this link trigger a page refresh with empty query parameters? Will that end up populating the default search params? We could use `?submitted=true` if we want to avoid that.

spec/controllers/appointments_controller_spec.rb
```diff
@@ -63,6 +63,46 @@
       end
     end
 
+    describe "filtering by days overdue" do
+      it "displays patients only less than one year overdue when checked" do
+        really_overdue_appointment = create(:appointment,
```
harimohanraj89, line 68: `really_overdue` 😅 Feel free to use `lost_to_followup` terminology.

spec/controllers/appointments_controller_spec.rb
```diff
@@ -63,6 +63,46 @@
...
+          facility: facility_2,
+          scheduled_date: 380.days.ago,
+          status: 'scheduled'
+        )
+        create(:blood_pressure, patient: really_overdue_appointment.patient, facility: facility_2)
+        really_overdue_patient_ids = really_overdue_appointment.patient_id
```
harimohanraj89, line 74: There's only one, can we rename the var to be singular?

spec/controllers/appointments_controller_spec.rb
```diff
@@ -63,6 +63,46 @@
...
+
+        get :index, params: {
+          search_filters: ["only_less_than_year_overdue"],
+          per_page: 'All'
+        }
+
+        patient_ids = (overdue_appointments_in_facility_1 + overdue_appointments_in_facility_2).map(&:patient_id)
+        expect(assigns(:patient_summaries).map(&:id)).to match_array(patient_ids)
+        expect(assigns(:patient_summaries).map(&:id)).to_not match_array(really_overdue_patient_ids)
```
harimohanraj89, line 83: Rather than matching an array, can we assert `.to_not include(really_overdue_patient_id)`? The `match_array` will pass even if the ID _is_ in the array, but doesn't exactly match, right?

spec/controllers/appointments_controller_spec.rb
```diff
@@ -63,6 +63,46 @@
...
+      end
+
+      it "displays patients with all overdue date when unchecked" do
+        really_overdue_appointment = create(:appointment,
+          facility: facility_2,
+          scheduled_date: 380.days.ago,
+          status: 'scheduled'
+        )
+        create(:blood_pressure, patient: really_overdue_appointment.patient, facility: facility_2)
+        really_overdue_patient_ids = really_overdue_appointment.patient_id
```
harimohanraj89, line 93: There's only one, can we rename the var to be singular?

spec/controllers/appointments_controller_spec.rb
```diff
@@ -63,6 +63,46 @@
       end
     end
 
+    describe "filtering by days overdue" do
+      it "displays patients only less than one year overdue when checked" do
```
harimohanraj89, line 67: Can use contexts to capture `when checked` and `when unchecked`? Furthermore, can we then set up our test data using `let!`s outside of the example itself?

spec/queries/patient_summary_query_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  let(:facility) do
+    create(:facility)
+  end
+
+  let(:really_overdue_appointments) do
+    create_list(:appointment, 2, facility: facility, scheduled_date: 380.days.ago, status: "scheduled")
+  end
+
+  it "returns nothing for no data" do
+    expect(PatientSummaryQuery.call).to eq([])
```
harimohanraj89, line 13: Using `described_class` instead of `PatientSummaryQuery` is quite handy.

spec/queries/patient_summary_query_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+RSpec.describe PatientSummaryQuery do
+  let(:facility) do
+    create(:facility)
+  end
+
+  let(:really_overdue_appointments) do
+    create_list(:appointment, 2, facility: facility, scheduled_date: 380.days.ago, status: "scheduled")
+  end
+
+  it "returns nothing for no data" do
```
harimohanraj89, line 12: Could we use contexts to capture these different scenarios? That'll let us set up the necessary test data in each context in isolation:

```ruby
subject(:patients) { described_class.call(relation: relation, filters: filters) }

let(:filters) { [] }
let(:relation) { PatientSummary.all }

context "when filtering for less than one year overdue"
  let(:filters) { ["only_less_than_year_overdue"] }
  before do
    # make some patients
  end
  it "does some stuff" do
    # assertions
  end  
end

context "when filtering for no phone numbers"
  let(:filters) { ["no_phone_number"] }
  before do
    # make some patients
  end
  it "does some stuff" do
    # assertions
  end  
end
```

app/controllers/appointments_controller.rb
rsanheim: > Does the existing layout introduce a bunch of grid things that make it hard to incorporate the new design?

The existing default layout doesn't really have any notion of a sidebar at all, so it would take an overhaul if we wanted to build in sidebar ability.

> PS. can we re-use whatever layout the My Facilities pages are using?

That is basically what this is, except copy / pasted =).  My Facilities had a bunch of code that was facilities specific, so we couldn't direclty reuse.  If all dashboards will all have some sort of sidebar in the future (/cc @dburka ?), then we could do a subsequent PR to get a common layout for dashboards that supports a sidebar in a cleaner way.

app/controllers/appointments_controller.rb
rsanheim: Sure - I think we should wait till that terminology is encoded into the data dictionary to make that change, though. What do you think @harimohanraj89 @timcheadle ?

app/views/layouts/overdue.html.erb
rsanheim: After thinking about this a bit, I'd really rather do this without javascript, which would require the layout changes we discussed while pairing.

This really feels to me like it should be one form that lets you filter on facilities, search filters, or whatever else we come up with. Adding JS for a pretty straightforward 'filter' endpoint would add a lot of unnecessary complexity.

app/controllers/appointments_controller.rb
```diff
@@ -2,22 +2,36 @@ class AppointmentsController < AdminController
...
-                           .order(risk_level: :desc, next_appointment_scheduled_date: :desc)
+    @search_filters = index_params[:search_filters]
+    # We have to check for a couple cases here to handle the default, page load with no parameters as well as the
+    # case where a user cleared out all search filters and then submitted the search form.
+    # This is because the default page load has one search filter enabled.
+    if @search_filters.blank? && index_params[:submitted] # user cleared out all filters and submitted
+      @search_filters = []
+    elsif @search_filters.blank? # initial page load
+      @search_filters = DEFAULT_SEARCH_FILTERS
+    end
```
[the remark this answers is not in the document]
rsanheim: Yup, fixed in 4f058af8b7c72edd679886e43b2af405ff4c5f06.

app/queries/patient_summary_query.rb
```diff
@@ -0,0 +1,36 @@
+class PatientSummaryQuery
+  def self.call(*args)
+    new(*args).call
+  end
+
+  # Create the query object with optional base relation and search filters
+  #
+  # relation: A base AR Relation to scope things off of - this would typically be used for authorization
+  # filters: An Array of search filters
```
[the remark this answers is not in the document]
rsanheim: Done.

spec/controllers/appointments_controller_spec.rb
rsanheim: Fixed.

spec/controllers/appointments_controller_spec.rb
rsanheim: I'd really rather not extract `let`s for situations like this...will explain further on a different reply.

spec/controllers/appointments_controller_spec.rb
rsanheim: Fixed

spec/controllers/appointments_controller_spec.rb
rsanheim: Fixed.

spec/queries/patient_summary_query_spec.rb
rsanheim: I find this type of over-extracted RSpec style to be detrimental to reading and learning from the specs, especially when future me or other future devs are trying to read a spec months or years after shipping the original feature.

The guidelines laid out in [An opinionated guide to readable RSpec](https://medium.com/selleo/an-opinionated-guide-to-readable-rspec-part-1-of-2-fe1dce79a478) largely align with what I've settled upon when writing readable specs, in particular this bit:

> Keep everything in “it”

> This is where the readability oriented approach differs the most from the optimization oriented one. The key is to keep most, if not all code required to execute test within individual it block. This means that we should drop all usages of let, subject and described_class and significantly limit the usage of both filters (before, around and after) and shared contexts / examples. This means more code will be duplicated with tests setup growing. However, the key advantage is, that while reading a test, developer will be fully aware of the context and will immediately see how test setup affects assertions. With heavily optimized specs code it might require a lot of navigation within spec file and external files to fully understand the environment in which assertions are checked.

A tldr version of this: the more I have to jump between different abstractions to grok a spec, the less useful that spec is long term for the team.  I like to keep the relevant factories, calling code, and expectations inside the spec itself.  If that gets very difficult to do, that is often a sign that the code under test is doing too much or not the right abstractions, and needs to be rethought and refactored.

Granted, there are certainly cases where a common `before` or `let` is used across 95% of a spec, and is worth extracting out. For the most _relevant_ pieces of the test, though, I like to avoid extraction or DRYing things up.

I'll now get off my soapbox here 😏 . Would be happy to discuss on a zoom call too with ya to see what you think.

app/controllers/appointments_controller.rb
rsanheim: I started a broader convo in slack about this https://simpledotorg.slack.com/archives/CK95QFJA2/p1588698688024800.

spec/queries/patient_summary_query_spec.rb
timcheadle: Agree totally with this.

harimohanraj89: Thanks for the soapbox sesh 😅 My rspec views are certainly a bit traditional (I bet they're 3-4 years old). I've never tried this approach while writing specs. I think the other server devs favor this approach as well, so I'm definitely down to try it.

In that case, I have only one request: Please point this out in my PRs too so I can get into this habit 😄

app/controllers/appointments_controller.rb
harimohanraj89: I'll let you plant the results of that convo into this PR 👍

app/controllers/appointments_controller.rb
harimohanraj89: Sounds like the sensible thing to do 👍 I'll resolve my other comments around this.

app/views/layouts/overdue.html.erb
prabhanshuguptagit: Could setting these inputs as `<input form="sidebar-form-id">` solve this? It allows you to have an input field belonging to a form outside the form. 

This is being done in My Facilities > BP control period selection, for example: There are "two forms with shared state" but effectively they're the same form thanks to the `form` attribute.

app/views/shared/_paginate_and_filter_by_facility.html.erb
prabhanshuguptagit: Oh yes, the `form=` will fix this, for sure :)

app/views/layouts/overdue.html.erb
harimohanraj89: This is really cool, is it a vanilla HTML feature?

prabhanshuguptagit: Yes! 😄

app/views/layouts/overdue.html.erb
```diff
@@ -0,0 +1,100 @@
...
+          </div>
+          <div class="modal-body">
+            <div class="row">
+              <div class="col-6 mb-3">
+                <h4>Patients</h4>
+                <%= form_tag appointments_path, method: :get do %>
+                  <% @search_filters.each do |filter, label| %>
+                    <div class="checkbox-row">
+                      <label for=<%= "search-#{filter}-mobile" %>>
+                        <% checked = @search_filters.include?(filter) %>
```
prabhanshuguptagit, line 35: Should be `@patient_search_filters`.

app/views/layouts/overdue.html.erb
```diff
@@ -0,0 +1,100 @@
...
+                      </label>
+                    </div>
+                  <% end %>
+                <% end %>
+              </div>
+            </div>
+          </div>
+          <div class="modal-footer">
+            <a href="?" type="button" class="btn btn-secondary">Clear all</a>
+            <button type="submit" form="query-filters-mobile" class="btn btn-primary">DONE</button>
```
prabhanshuguptagit, line 47: This button is not hooked to the modal form currently.

app/queries/patient_summary_query.rb
```diff
@@ -0,0 +1,32 @@
+class PatientSummaryQuery
+  def self.call(*args)
+    new(*args).call
```
prabhanshuguptagit, line 3: I've found this `.call` abstraction to be a bit unnecessary, assuming this is here only because it allows us to:
1. make neat looking scopes: [reference](http://craftingruby.com/posts/2015/06/29/query-objects-through-scopes.html). 
2. do things like `PatientSummaryQuery.where(...)`. This works only if we want to use the default parameters though, which isn't often the case.

My thoughts on 1: This query object is a presentation level concern (filtering etc.) and will never be/shouldn't be used in the `PatientSummary` model. Providing for it to be used as a scope seems a bit circular. Wdyt?

harimohanraj89: I've seen this pattern used _outside_ the context you've referenced, @prabhanshuguptagit . The crux being: we should operate with an instance of the `PatientSummaryQuery` to leverage advantages of OOP. This class method is then created as a convenience method so that callers don't have to worry about instantiating this object themselves. They can say "Hey, class, take care of this", while maintaining an OO implementation under the hood.

I quite like this.

spec/queries/patient_summary_query_spec.rb
rsanheim: > In that case, I have only one request: Please point this out in my PRs too so I can get into this habit 😄

👍 will do 😀

app/views/layouts/overdue.html.erb
rsanheim: > Could setting the `facility` and `per_page` filters as `<select form="sidebar-form-id">` solve this? It allows you to have an input field belonging to a form outside the form.

Cool, I'm doing this with everything going through the form in the sidebar when on the overdue page. Thanks for the suggestion!

app/views/layouts/overdue.html.erb
rsanheim: There is not, this looks like a mobile only thing that I took from my facilities.  I'll test this out to make sure it resets to the defaults.

app/views/layouts/overdue.html.erb
```diff
@@ -0,0 +1,100 @@
...
+        <h4>Patients</h4>
+
+        <%= form_tag appointments_path, method: :get do %>
+          <%= hidden_field_tag :facility_id, @facility_id %>
+          <%= hidden_field_tag :per_page, @per_page %>
+          <%= hidden_field_tag :submitted, true %>
+
+          <% Array(@patient_search_filters).each do |filter, label| %>
+            <div>
+              <label for=<%= "search-#{filter}" %>>
```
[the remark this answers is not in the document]
rsanheim: We could, but for cases like this where the form control logic gets a little more complicated I like dropping into raw html to avoid the burden of the view helper abstraction.

app/views/layouts/overdue.html.erb
```diff
@@ -0,0 +1,100 @@
...
+                <% checked = @search_filters.include?(filter) %>
+                <%= check_box_tag("search_filters[]", filter, checked, { id: "search-#{filter}" } ) %>
+                <%= label %>
+              </label>
+            </div>
+          <% end %>
+          <label>
```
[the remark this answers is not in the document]
rsanheim: This label wraps the 'apply filters' button, another thing taken straight from my facilities search filters. We should clean this up to wrap an input when we get to the "consolidate and clean up sidebar" work.

app/queries/patient_summary_query.rb
```diff
@@ -0,0 +1,36 @@
...
+      relation.overdue
+    else
+      relation.all_overdue
+    end
+    if filters.include?("phone_number")
+      result = result.where("latest_phone_number is not null")
+    end
+    if filters.include?("no_phone_number")
+      result = result.where("latest_phone_number is null")
+    end
```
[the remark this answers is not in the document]
rsanheim: Agreed, adding to my todo list in the PR body.

app/views/layouts/overdue.html.erb
rsanheim: 👍

app/views/layouts/overdue.html.erb
rsanheim: 👍 Thanks, fixed that up.

app/models/appointment.rb
```diff
@@ -47,6 +47,12 @@ def self.overdue
       .where(arel_table[:remind_on].eq(nil).or(arel_table[:remind_on].lteq(Date.current)))
   end
 
+  def self.all_overdue
+    where(status: 'scheduled')
+      .where(arel_table[:scheduled_date].lt(Date.current))
+      .where(arel_table[:remind_on].eq(nil).or(arel_table[:remind_on].lteq(Date.current)))
+  end
```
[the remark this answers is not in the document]
harimohanraj89: @rsanheim, <bump>, should we update the `overdue` scope to use this method?

rsanheim:
> Only [one outstanding comment](https://github.com/simpledotorg/simple-server/pull/907#discussion_r424206795) from me. Pre-emptive approval for unblockage 🚀

Ah yup, took care of that.  👍

PR #911, opened 2020-04-29
Propagate new permissions to users

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/172585739

## Because

We don't currently have a good way to propagate new permissions to users

## This addresses

This PR adds a new rake task to give user permissions. The class that provides this functionality can also be called from a data migration to make adding permissions easier.


lib/tasks/scripts/add_permission_to_access_level.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def create
+    users.each do |user|
+      permission_resources(user).each do |resource|
+        UserPermission
+          .where(user: user,
+                 permission_slug: permission[:slug],
+                 resource_type: resource[:resource_type],
+                 resource_id: resource[:resource_id])
+          .first_or_create
```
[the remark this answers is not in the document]
vkrmis: > I realized that _technically_, this method will always only `create`, since the `user_is_a_match?` method will filter out users that already have the permission, but we should keep this in here to be safe, and safeguard against parallel processing.

That's not necessarily true. For example, a CVHO could have permissions of the same type for multiple facility groups.

lib/tasks/scripts/add_permission_to_access_level.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+    end
+  end
+
+  private
+
+  def users
+    users = User.includes(:user_permissions).where.not(user_permissions: { id: nil })
```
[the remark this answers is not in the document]
vkrmis: It wouldn't, so this `where` excludes users who have no permissions i.e., app users who don't use the dashboard. This is the vast majority of our users (>98%).

lib/tasks/scripts/add_permission_to_access_level.rb
```diff
@@ -0,0 +1,70 @@
...
+    users.select{ |user| user_is_a_match?(user) }
```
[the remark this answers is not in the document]
vkrmis: MIght be a bit premature, this number is likely to be < 1000 for a very long time. It is currently < 100.

lib/tasks/scripts/add_permission_to_access_level.rb
```diff
@@ -0,0 +1,70 @@
...
+    when permission[:resource_priority].include?(:organization) && existing_resource_types.include?('Organization')
+      user.user_permissions.where(resource_type: 'Organization').map do |resource|
+        { resource_type: resource.resource_type, resource_id: resource.resource_id }
+      end.uniq
+    when permission[:resource_priority].include?(:global) && existing_resource_types.include?(nil)
+      [{ resource_type: nil, resource_id: nil }]
+    end
+  end
+
+  def user_is_a_match?(user)
```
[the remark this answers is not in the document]
vkrmis: Yup, I couldn't think of a good name

lib/tasks/scripts/add_permission_to_access_level.rb
```diff
@@ -0,0 +1,70 @@
...
+      user.user_permissions.where(resource_type: 'FacilityGroup').map do |resource|
+        { resource_type: resource.resource_type, resource_id: resource.resource_id }
+      end.uniq
+    when permission[:resource_priority].include?(:organization) && existing_resource_types.include?('Organization')
+      user.user_permissions.where(resource_type: 'Organization').map do |resource|
+        { resource_type: resource.resource_type, resource_id: resource.resource_id }
+      end.uniq
+    when permission[:resource_priority].include?(:global) && existing_resource_types.include?(nil)
+      [{ resource_type: nil, resource_id: nil }]
+    end
```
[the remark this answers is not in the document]
vkrmis: Hmm, I get that, but 
- Its possible for a user to have say `view_overdue_list` on multiple facility groups, I've seen this in our data. The user would still have the same `access_level/role`.
- Some existing roles create permissions with different scopes. Eg: The `supervisor` role (CVHOs) have mostly `facility_group` permissions, but one `organization` permission. In this case its so that they can `manage_admins` and invite other people to the dashboard.

lib/tasks/scripts/add_permission_to_access_level.rb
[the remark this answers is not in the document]
harimohanraj89: Oh, because users covers both app user and dashboard users. That seems fair enough 👍

lib/tasks/scripts/add_permission_to_access_level.rb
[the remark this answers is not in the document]
harimohanraj89: Ah, then let's keep as is, thanks.

lib/tasks/scripts/add_permission_to_access_level.rb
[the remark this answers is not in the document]
harimohanraj89: After our Slack discussion, this seems to make great sense 👌 Nice.

lib/tasks/scripts/add_permission_to_access_level.rb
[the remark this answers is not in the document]
harimohanraj89: Personally, I think `eligible?` is pretty good. `users.select { |u| eligible?(u) }`

spec/lib/tasks/scripts/add_permission_to_access_level_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    it "creates permissions for users who don't have them" do
+      UserPermission.where(user_id: supervisor_1.id, permission_slug: 'view_cohort_reports').each(&:delete)
+
+      expect { described_class.new(:view_cohort_reports, :supervisor).create }
+        .to change { UserPermission.where(user_id: supervisor_1.id, permission_slug: 'view_cohort_reports', user: [supervisor_1]).count }.by(2)
+    end
+
+    it 'creates permissions of the appropriate resource_priority' do
+      UserPermission.where(permission_slug: 'view_cohort_reports').each(&:delete)
```
harimohanraj89, line 44: Let's use `destroy!` so that if the deletion is unsuccessful for any reason, we'll know very clearly.

spec/lib/tasks/scripts/add_permission_to_access_level_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      expect { described_class.new(:view_cohort_reports, :supervisor).create }
+        .to change { UserPermission.where(permission_slug: 'view_cohort_reports', resource_type: 'FacilityGroup', user: [supervisor_1, supervisor_2]).count }.by(3)
```
harimohanraj89, line 47: Instead of asserting counts of very specific queries, can we instead do something like:

```ruby
described_class.new(:view_cohort_reports, :supervisor).create

expect(UserPermission.exists?(params)).to eq(true)
expect(UserPermission.exists?(other_params)).to eq(true)
expect(UserPermission.exists?(more_params)).to eq(true)
```

spec/lib/tasks/scripts/add_permission_to_access_level_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      expect { described_class.new(:view_cohort_reports, :organization_owner).create }
+        .to change { UserPermission.where(permission_slug: 'view_cohort_reports', resource_type: 'Organization', user: organization_owner).count }.by(1)
+
+      expect { described_class.new(:view_cohort_reports, :owner).create }
+        .to change { UserPermission.where(permission_slug: 'view_cohort_reports', resource_type: nil, user: owner).count }.by(1)
```
harimohanraj89, line 53: Can we split these expects into separate examples that describe FG permissions, Org permissions, and Global permissions respectively?

spec/lib/tasks/scripts/add_permission_to_access_level_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  let!(:owner) { create(:admin, :owner) }
+  let!(:organization_owner) { create(:admin, :organization_owner, organization: organization) }
+  let!(:supervisor_1) { create(:admin, :supervisor, facility_group: facility_group_1) }
+  let!(:supervisor_2) { create(:admin, :supervisor, facility_group: facility_group_2) }
+  let!(:access_levels) { Permissions::ACCESS_LEVELS }
+  let!(:all_permissions) { Permissions::ALL_PERMISSIONS }
+
+  before do
+    create(:user_permission, user: supervisor_1, permission_slug: :manage_facilities, resource_type: 'FacilityGroup', resource_id: facility_group_2.id)
+  end
```
harimohanraj89, line 17: Can we move setup data to the section of the specs that it's used in? For example, which parts of this is only relevant to the `create` section, and not the `valid` section?

spec/lib/tasks/scripts/add_permission_to_access_level_spec.rb
harimohanraj89: This will also allow the specs to better document the nuance involved, as we can replace 

```
creates permissions of the appropriate resource_priority
```
with

```
when the permission has global priority
  global permission is created
...
```

lib/tasks/scripts/add_permission_to_access_level.rb
```diff
@@ -0,0 +1,81 @@
...
+    @permission = Permissions::ALL_PERMISSIONS[permission_name]
+    @access_level = Permissions::ACCESS_LEVELS.select { |level| level[:name] == access_level_name }.first
+  end
+
+  def valid?
+    permission && valid_permission_for_access_level?
+  end
+
+  def create
+    return unless valid?
```
harimohanraj89, line 20: It would be nice to return `false` in this case, similar to ActiveRecord.

spec/lib/tasks/scripts/add_permission_to_access_level_spec.rb
```diff
@@ -2,55 +2,159 @@
...
-  before do
-    create(:user_permission, user: supervisor_1, permission_slug: :manage_facilities, resource_type: 'FacilityGroup', resource_id: facility_group_2.id)
-  end
-
   describe '#valid?' do
     it 'validates that the permission exists' do
-      expect(described_class.new(:fake_permission, :owner).valid?).to be_falsey
-      expect(described_class.new(:manage_organizations, :owner).valid?).to be_truthy
+      expect(described_class.new(:fake_permission, :owner).valid?).to eq(false)
+      expect(described_class.new(:manage_organizations, :owner).valid?).to eq(true)
```
harimohanraj89, line 8: Sweeeet. Feel free to use the swanky rspec magic:

```ruby
expect(described_class.new(:fake_permission, :owner)).to be_valid
```

app/policies/permissions.rb
```diff
@@ -121,7 +121,6 @@ module Permissions
         view_health_worker_activity
         download_patient_line_list
         manage_admins
-        view_my_facilities
```
harimohanraj89: Why are we reverting this change?

lib/tasks/scripts/add_permission_to_access_level.rb
```diff
@@ -80,4 +82,12 @@ def eligible?(user)
   def valid_permission_for_access_level?
     access_level && access_level[:default_permissions].include?(permission_name)
   end
+
+  def log_string(permission, user, resource)
```
harimohanraj89, line 86: Can we use named arguments? Also, the `string` suffix is redundant. I'd be comfortable with just `def log`

lib/tasks/add_permission_to_access_level.rake
```diff
@@ -18,5 +18,9 @@ task :add_permission_to_access_level,
           ' this task'
   end
 
-  new_permissions.create
+  created_permissions = new_permissions.create
+  users_with_new_permissions = created_permissions.flatten.map(&:user).uniq.sort_by(&:full_name)
+
+  puts "Created #{created_permissions.count} permission records for the following #{users_with_new_permissions.count} "\
+       "users: #{users_with_new_permissions.map(&:full_name)}"
```
harimohanraj89, line 25: Handling these permissions and flattening things and stuff seems awkward. Can we just print the log inside the permissioning script directly? It seems simpler. Additionally, should we use `Rails.logger` instead of `puts`?

vkrmis: We're already printing these using Rails.logger in the script, the point of doing this via a `puts` here would be to print this message on a remote console where the rake task was invoked. I'll drop it in favour of the inlined Rails log messages.

harimohanraj89: Ahhh I missed that. Neat!

lib/tasks/scripts/add_permission_to_access_level.rb
vkrmis: added some specs via `#create`

app/policies/permissions.rb
vkrmis: This wasn't intentional, I think this PR itself shouldn't add any permissions. I'll add that separately.

harimohanraj89: Ah I see. This one will set up the script, and a subsequent PR will add this permission, and after that we'll run the script. 👍

harimohanraj89:
This PR is lookin slick ✨

PR #915, opened 2020-04-29
Upgrade to rails 5.2

rsanheim:
**Story card:** https://www.pivotaltracker.com/story/show/172593017

* [Rails 5.2 Release notes](https://guides.rubyonrails.org/5_2_release_notes.html)
* [5.1 to 5.2 Upgrade Guide](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-1-to-rails-5-2)

## Because

We want to be on the latest Rails, so starting upgrade w/ 5.1 to 5.2.

## This addresses

Our version of Rails.


kitallis:
Rails 5.2 has introduced the new [credentials](https://guides.rubyonrails.org/5_2_release_notes.html#credentials) system, would that break our existing way of just sticking the secret key via `SECRET_KEY_BASE`? Or is it backwards-compatible?

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -0,0 +1,38 @@
...
+# Rails.application.config.active_record.cache_versioning = true
+
+# Use AES-256-GCM authenticated encryption for encrypted cookies.
+# Also, embed cookie expiry in signed or encrypted cookies for increased security.
+#
+# This option is not backwards compatible with earlier Rails versions.
+# It's best enabled when your entire app is migrated and stable on 5.2.
+#
+# Existing cookies will be converted on read then written with the new scheme.
+# Rails.application.config.action_dispatch.use_authenticated_cookie_encryption = true
```
harimohanraj89, line 20: Looks like we should set this to false to start, until we're confident that the upgrade is stable. Otherwise, we won't be able to roll back without possibly invalidating all user cookies.

harimohanraj89:
According to the [upgrade guide](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-1-to-rails-5-2), looks like it's backwards compatible. We can always test to confirm.

config/initializers/new_framework_defaults_5_2.rb
rsanheim: Yeah, I don't know enough about our clients and how long we want their sessions to last...though it feels like forcing folks to re-login wouldn't be a bad thing to do with this upgrade?
In any case, I set this to false for now.

harimohanraj89: The risk is that during clinic hours, if a nurse is logged out and has to sign in again to continue treating patients, it could be a very disruptive experience. Since we have the option of avoiding this friction, let's do it 👌

timcheadle: For what it's worth, nurses don't deal with cookies. That should only affect the web side of things for CVHOs, and it looks like Rails auto-migrates them.

Let's keep this as `false` for now which allows us to roll back, but change it once we're confident the app works.

PR #917, opened 2020-04-30
Add new SMS reminders – Punjabi, Marathi, Kannada

kitallis:
**Story card:** https://www.pivotaltracker.com/n/projects/2184102/stories/172243161

## Because

Due to COVID-19, we're modifying the SMS reminders a bit. Related [PR](https://github.com/simpledotorg/simple-server/pull/902).

## This addresses

* Modifies the reminders for: Punjabi, Marathi
* Adds a reminder in: Kannada

**Note**: SMS reminders must be re-enabled for these to actually start sending SMSes. However, this PR and the re-enabling can happen independently.


spec/services/sms_notification_service_spec.rb
```diff
@@ -68,6 +68,18 @@
         sms.send_reminder_sms('missed_visit_sms_reminder', appointment, '', 'pa-Guru-IN')
       end
 
+      it 'should have the SMS body in Kannada' do
+        sms = SmsNotificationService.new(recipient_phone_number, sender_phone_number, twilio_client)
+
+        expected_msg_punjabi = 'Simple Facility ನಲ್ಲಿರುವ ನಮ್ಮ ಸಿಬ್ಬಂದಿ ನಿಮ್ಮ ಮತ್ತು ನಿಮ್ಮ ಹೃದಯದ ಆರೋಗ್ಯದ ಬಗ್ಗೆ ಯೋಚಿಸುತ್ತಿದ್ದಾರೆ. ದಯವಿಟ್ಟು ನಿಮ್ಮ ಬ್ಲಡ್ ಪ್ರೆಶರ್ ಔಷಧಗಳನ್ನು ಮುಂದುವರೆಸಿ. ನಿಮ್ಮ ಹತ್ತಿರದ ಉಪ ಕೇಂದ್ರದಲ್ಲಿ ನಿಮ್ಮ ಔಷಧಗಳನ್ನು ಸಂಗ್ರಹಿಸಿ. ನಿಮ್ಮ ANM ಅಥವಾ ASHA ಅವರನ್ನು ಸಂಪರ್ಕಿಸಿ.'
```
harimohanraj89, line 74: `expected_msg_kannada` 🍝

kitallis: Argh 🙄

app/models/address.rb
```diff
@@ -4,6 +4,7 @@ class Address < ApplicationRecord
   STATE_TO_LOCALE = {
     punjab: 'pa-Guru-IN',
     maharashtra: 'mr-IN',
+    karnataka: 'kn-IN'
```
harimohanraj89, line 7: Iiinteresting, so the language chosen for the reminder is not based on the patient's registration facility, but based on their address's "state"?

kitallis: Yeah, we took this decision [here](https://docs.google.com/document/d/1r0lfpK6--PlJEZbnGVx1-KnnMwMBfCk5jErpnJopZzY/edit#heading=h.com9rkcwqvda). But I can't remember exactly why. Best guess would be that we uncovered that patients move around and their specified address is most reliable.

This is not very multi-country sensitive though since it sits in the model. But it can be if we later migrate it to a config file or something. Should not be very hard.

harimohanraj89: Neato 👍 that clarifies why we need to update this map, and why you were [looking at state distribution](https://simpledotorg.slack.com/archives/CFHMC60P5/p1588240632089300) 😄

PR #919, opened 2020-04-30
User analytics specs failing - Freeze time to try to fix

rsanheim:
I'm not entirely sure what is going on here, but it appears since we are
on the last day of the month these specs started failing that are
setting up fixture data on one month intervals.

In the interest of getting CI green, it looks like freezing time to the
morning here fixes things, at least locally.  So lets see if that helps CI.

**Story card:** n/a - fixing the build

## Because

Trying to fix the build

## This addresses

Build is broken on last day of month

### DIFF
--- a/app/presenters/user_analytics_presenter.rb
+++ b/app/presenters/user_analytics_presenter.rb
@@ -105,11 +105,11 @@ class UserAnalyticsPresenter
   def trophy_stats
     follow_ups = all_time_follow_ups_by_gender.values.sum
 
-    all_trophies =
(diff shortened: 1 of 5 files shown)


rsanheim:
> LGTM. Let's add a bug to Pivotal to make sure we don't forget about this issue; Timecop seems like a workaround, not a fix.

Yup, opened https://www.pivotaltracker.com/story/show/172618243. I'l make sure it stays open after this merges....not sure if GH / Pivotal auto closes things at all on merge.

PR #923, opened 2020-05-01
Change AppointmentNotification to one notification per job

rsanheim:
In support of https://www.pivotaltracker.com/story/show/168072307

## Because

Changing the job to send one notification per job. This will make it easier to manage things via Sidekiq, because Sidekiq jobs are cheap. We also don't need to worry about large batches failing halfway through and having to recover from that.

## This addresses

Preparing for whats app reminders.


timcheadle:
Let's just remove all of the fan out function since we're going with a worker per appointment reminder anyway. We can add this back in later if necessary.

rsanheim:
> Let's just remove all of the fan out function since we're going with a worker per appointment reminder anyway. We can add this back in later if necessary.

Yup, done.

harimohanraj89:
Shall we test that this new per-appointment job queuing will not strain our Redis memory usage?

rsanheim:
> Shall we test that this new per-appointment job queuing will not strain our Redis memory usage?

Sure - would that require a deploy to sandbox and then some sort of mass notification send via Twilio's test mode?  I'm not sure what would all be involved in that.

timcheadle:
How much RAM does our production elasticache instance have?

Sandbox has 2GB and around 1.7GB is almost always free.

[image: image.png]

rsanheim:
> Sandbox has 2GB and around 1.7GB is almost always free.

If we have 1.7 free there, I'm pretty confident in our sandbox env.
We would really need some visibility into our prod env to know if there is a risk there at all. 
A bigger concern for prod would be making sure Sidekiq is using a different Redis database than the cache.

harimohanraj89:
Cache and Sidekiq currently share a Redis instance in production. Could we queue 1000 of these jobs and see how much space that takes? We may find that we'll need to queue 10 million jobs before this becomes a problem, in which case we can confidently ship.

rsanheim:
I threw this out to sandbox and queued up 2000 jobs. It basically didn't make a dent in free memory at all. Below is from watching the Sidekiq admin while the jobs processed, the memory never went about 4.5 M (thats megabytes). I was also watching via SSH and `free`, and we always had over 

Screen Shot 2020-05-05 at 3 35 54 PM

Before running the batch

```
deploy@ec2-15-206-123-187:~/apps/simple-server/shared/log$ free -h
              total        used        free      shared  buff/cache   available
Mem:            31G        1.7G         27G         17M        2.0G         29G
Swap:            0B          0B          0B
```

After running the batch:
```
deploy@ec2-15-206-123-187:~/apps/simple-server/current$ free -mh
              total        used        free      shared  buff/cache   available
Mem:            31G        1.1G         28G         17M        2.0G         29G
Swap:            0B          0B          0B

```

So apparently we had _more_ free memory after? 😄  I imagine that was a blip, but basically, we could queue many many thousands of these w/o worrying about memory.

PR #924, opened 2020-05-01
Rename otp_valid_until to otp_expires_at for consistency

timcheadle:
## Because

I'm cleaning up some authentication behavior and this field was confusing.

## This addresses

This renames `otp_valid_until` to `otp_expires_at` to be consistent with other DateTime fields (and match Rails convention). It also better indicates that this field is for expiration, not validity.


db/migrate/20180614061014_add_columns_for_login_to_users.rb
```diff
@@ -1,7 +1,7 @@
 class AddColumnsForLoginToUsers < ActiveRecord::Migration[5.1]
   def change
     add_column :users, :otp, :string, null: false
-    add_column :users, :otp_valid_until, :datetime, null: false
```
rsanheim, line 4: Oops, these renames should not have landed.

db/migrate/20200501174342_rename_otp_valid_until_to_otp_expires_at.rb
```diff
@@ -0,0 +1,6 @@
+class RenameOtpValidUntilToOtpExpiresAt < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :passport_authentications, :otp_expires_at, :otp_expires_at
```
rsanheim, line 3: This is renaming the same column name here.

PR #926, opened 2020-05-04
Upgrade to Ruby 2.7.1

rsanheim:
**Story card:** ...

**NOTE**: This upgrade will require at least one PR in the deployment repo, as well as various test deploys to make sure we can upgrade to Ruby 2.7.1. We should also ensure we can downgrade back to Ruby 2.5.x if needed in the case something doesn't work during the transition.

I've added a checklist to get effort shipped below. 

## Checklist

* [ ] Update Semaphore so we can run builds in 2.5 and 2.7
* [ ] Create a deployment PR to install Ruby 2.7.1 on all environments
* [ ] Ship that deployment PR
* [ ] Verify Ruby 2.7.1 exists in all enviroments
* [ ] Ensure Bundler + rubygems works in new ruby
* [ ] Open Question - do we need to update Bundler along with this?
* [ ] Test deploy - SBX
* [ ] Test deploy - QA
* [ ] Test deploy - staging
* [ ] Test deploy - Herkou
* [ ] Test deploy - India prod
* [ ] Test deploy - Bangladesh prod
* [ ] Merge this PR!
* [ ] Monitor production environments

## This addresses


kitallis:
@harimohanraj89 @rsanheim If we're just running `2.7.1` off of Semaphore 1.0, should we kick out `semaphore.yml` if/until we switch over to Semaphore 2.0.

harimohanraj89:
@kitallis Installing Ruby 2.7.1 manually on Semaphore is painful (it adds 3 minutes of time per build). I did that just to validate this PR, we should continue the Semaphore 2.0 upgrade asap (I'm assuming Ruby 2.7.1 is available on Semaphore 2.0?)

kitallis:
@harimohanraj89 Ah, that makes sense. I assumed that it would cache the env artifacts, but it sounds like it's a patch to just get it to work.

rsanheim:
@harimohanraj89 @kitallis I've updated the PR body with what we need to get this over the line. There are a bunch of steps, but most of them should be pretty straightforward (hopefully).

I've also set this back to draft, since it isn't really close yet.

.semaphore/semaphore.yml
```diff
@@ -0,0 +1,87 @@
+# Use the latest stable version of Semaphore 2.0 YML syntax:
+  version: v1.0
```
rsanheim, line 2: It would make life easier if we backported this to master I think, so we could make sure Semaphore 2 works fine with all our branches.

.semaphore/semaphore.yml
```diff
@@ -0,0 +1,87 @@
...
+            - bundle exec rake db:test:prepare
+
+        jobs:
+        - name: RSpec - model tests
+          commands:
+            - bundle exec rspec spec/models
+
+        - name: RSpec - controller tests
+          commands:
+            - bundle exec rspec spec/controllers
```
rsanheim, line 67: I think we are missing a bunch of our specs with this approach, like api / service / queries.

kitallis:
The checklist seems good 👍

PR #932, opened 2020-05-05
Add view_my_facilites permissions to access levels

vkrmis:
**Story card:** https://www.pivotaltracker.com/n/projects/2184102/stories/172585739

## Because

Supervisors and Owners should have the 'view_my_facilities' permission

## This addresses

Adds the 'view_my_facilities' permission as a default permission for Supervisors and Owners

### DIFF
--- a/app/policies/permissions.rb
+++ b/app/policies/permissions.rb
@@ -121,6 +121,7 @@ module Permissions
         view_health_worker_activity
         download_patient_line_list
         manage_admins
+        view_my_facilities
       ]
     },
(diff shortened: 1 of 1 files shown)


vkrmis:
> Nice. Do we have specs for this config or no? Additionally, after shipping this PR, we'll run the rake task twice - once for superadmins and once for CVHOs, is that correct?

Nope, there aren't any specs on the config. We have specs for the various policies.
Yes, we will need to run it once per `ACCESS_LEVEL`

PR #933, opened 2020-05-05
Added a nowrap to the bar charts

dburka:
**Story card:** NONE

## Because

Just adds a `no-wrap` to the cohort bar charts because <1% was wrapping awfully.

## This addresses

Wrapping


harimohanraj89:
@dburka This means the <1% bar will sometimes be bigger than, say, a 3% bar. Confirming, are we ok with that tradeoff?

Screenshot 2020-05-05 at 5 27 13 PM

PR #935, opened 2020-05-05
Add teleconsultation support

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/172671449

## Because
We need to add a few fields to the Facility model to support Teleconsultation
[PRD](https://docs.google.com/document/d/1FI3lxWb79zNBgU5walrCbJIo19iT4RpJiuura4R4gQI/edit#heading=h.n3eua32nsj8y)

## This addresses
Adds the `enable_teleconsultation` and `teleconsultation_phone_number` fields


app/assets/javascripts/application.js
```diff
@@ -22,6 +22,7 @@
 //= require components
 //= require tablesort
 //= require tablesort/dist/sorts/tablesort.number.min
+//= require hide-teleconsultation-fields.js
```
harimohanraj89, line 25: Can we drop the `js` extension? Also, could we call it just `teleconsultation-fields`, or `facility-teleconsultation-fields`, as it hides _and_ shows the fields?

app/controllers/api/v4/facility_teleconsultations_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def show
+    render json: { teleconsultation_phone_number: teleconsultation_phone_number }
+  end
+
+  private
+
+  def teleconsultation_phone_number
+    "#{facility.teleconsultation_isd_code} #{facility.teleconsultation_phone_number}" if
+      facility.teleconsultation_isd_code && facility.teleconsultation_phone_number && facility.enable_teleconsultation
```
harimohanraj89, line 15: We should provide the number even if teleconsultation is disabled, let the client decide how to manage that.

app/controllers/api/v4/facility_teleconsultations_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  attr_reader :facility
+
+  def show
+    render json: { teleconsultation_phone_number: teleconsultation_phone_number }
+  end
+
+  private
+
+  def teleconsultation_phone_number
+    "#{facility.teleconsultation_isd_code} #{facility.teleconsultation_phone_number}" if
```
harimohanraj89, line 14: Is the space ok? Should we smush it together? I instinctively think we should mush it together, but let's check with Sasikanth on what format the client expects (you may have already done so).

app/controllers/api/v4/facility_teleconsultations_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+      facility.teleconsultation_isd_code && facility.teleconsultation_phone_number && facility.enable_teleconsultation
+  end
+
+  def set_facility
+    @facility ||= Facility.find(params[:facility_id])
+  end
+
+  def validate_teleconsultation_facility_belongs_to_users_facility_group
+    head :unauthorized unless current_facility_group.facilities.where(id: facility.id).present?
```
harimohanraj89, line 23: Do we use pundit policies for authorization in APIs? Or is this hand-rolled approach ok?

app/schema/api/v3/models.rb
```diff
@@ -189,9 +189,11 @@ def facility
           config: {
             type: :object,
             properties: {
-              enable_diabetes_management: { type: :boolean } }
+              enable_diabetes_management: { type: :boolean },
+              enable_teleconsultation: { type: :boolean }
+            }
           },
-          required: %w[enable_diabetes_management]
+          required: %w[enable_diabetes_management enable_teleconsultation]
```
harimohanraj89, line 196: If we expect this key to be present, can we add a `in: [true, false]` inclusion validation to the model?

app/controllers/admin/facilities_controller.rb
```diff
@@ -89,7 +89,10 @@ def facility_params
       :longitude,
       :enable_diabetes_management,
       :monthly_estimated_opd_load,
-      :zone
+      :zone,
+      :enable_teleconsultation,
+      :teleconsultation_phone_number,
+      :teleconsultation_isd_code
```
harimohanraj89, line 95: If the `teleconsultation_phone_number` is blank, should we also clear out the `teleconsultation_isd_code`? Otherwise every update to a facility will persist an ISD code behind the scenes.

spec/api/v4/facility_teleconsultations_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      let!(:request_user) { create(:user) }
+      let!(:request_facility) { create(:facility, facility_group: request_user.facility.facility_group) }
+      let!(:HTTP_X_USER_ID) { request_user.id }
+      let!(:HTTP_X_FACILITY_ID) { request_facility.id }
+      let!(:Authorization) { "Bearer #{request_user.access_token}" }
+      let(:facility_id) { request_facility.id }
+
+      response '200', 'Teleconsultation phone number is returned' do
+        schema Api::V4::Schema.facility_teleconsultations_response
+        run_test!
```
harimohanraj89, line 21: Should we assert anything about the response in this spec?

app/controllers/api/v4/facility_teleconsultations_controller.rb
vkrmis: I've mushed them together for now, will check with Sasikanth if there's a preference either way.

app/controllers/api/v4/facility_teleconsultations_controller.rb
vkrmis: No, we only apply pundit policies to Admin (i.e., Dashboard) users. API user authorization is handled like this. See APIController for reference.

spec/api/v4/facility_teleconsultations_spec.rb
vkrmis: This is a swagger spec, we're asserting that the response matches the declared `schema`

harimohanraj89: Neat. I like the separation of schema assertions here and payload assertions in the controller spec.

app/models/facility.rb
```diff
@@ -130,5 +135,15 @@ def diabetes_enabled?
...
+  def teleconsultation_enabled?
+    enable_teleconsultation.present?
+  end
+
   CSV::Converters[:strip_whitespace] = ->(value) { value.strip rescue value }
+
+  private
+
+  def clear_isd_code
+    self.teleconsultation_isd_code = ""
```
harimohanraj89, line 147: Any difference between setting to `""` and `nil`? How does the validation messaging behave when we try to save a facility with isd code but no number?

app/views/admin/facilities/_form.html.erb
```diff
@@ -24,12 +24,23 @@
...
+            <%= form.text_field :teleconsultation_phone_number, id: :facility_teleconsultation_phone_number, label: 'Mobile Number' %>
+          </div>
+        </div>
       </div>
-
       <%= form.primary %>
     <% end %>
   </div>
 </div>
+
```
harimohanraj89, line 46: 👾

app/assets/javascripts/teleconsultation-fields.js
```diff
@@ -0,0 +1,11 @@
+$(document).ready(function () {
+    $("#facility_enable_teleconsultation").click(function () {
```
harimohanraj89, line 2: I would set some data attribute on the checkbox that we can use to find this. It may be identical to the ID, but it's a good idea to use something we control, rather than an autogenerated Rails ID.

app/models/facility.rb
vkrmis: Not a big difference, since `nil.present?` and `"".present?` are both false. Just sticking to the defaults, all the other text fields in our forms get saved with a `""` value when they're empty.

app/assets/javascripts/teleconsultation-fields.js
vkrmis: This ID is being set explicitly in the form, wouldn't that suffice?

harimohanraj89: Oh I completely missed that! Perf.

app/controllers/api/v4/facility_teleconsultations_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    render json: { teleconsultation_phone_number: teleconsultation_phone_number }
+  end
+
+  private
+
+  def teleconsultation_phone_number
+    phone_number = Phonelib.parse(facility.teleconsultation_isd_code.to_s + facility.teleconsultation_phone_number.to_s)
+
+    phone_number.full_e164 if
+      facility.teleconsultation_isd_code.present? && facility.teleconsultation_phone_number.present?
```
harimohanraj89, line 17: Can we `return unless...` at the top of the method?

app/models/facility.rb
vkrmis: Screenshot 2020-05-11 at 3 28 04 PM

app/views/admin/facilities/_form.html.erb
```diff
@@ -32,7 +32,7 @@
         <div class="row mt-2" id="teleconsultation_fields" style=<%= "#{'display:none' unless @facility[:enable_teleconsultation]}" %> >
           <div class="col col-md-4">
             <%= form.text_field :teleconsultation_isd_code, id: :facility_teleconsultation_isd_code, label: 'Country Code',
-                                value: @facility[:teleconsultation_isd_code] || Rails.application.config.country['sms_country_code'] %>
+                                value: @facility[:teleconsultation_isd_code].present? ? @facility[:teleconsultation_isd_code] : Rails.application.config.country['sms_country_code'] %>
```
harimohanraj89, line 35: Nice catch `@facility[:teleconsultation_isd_code].presence || Rails.application.config.country['sms_country_code']` also works 👍

vkrmis: TIL

PR #937, opened 2020-05-06
Clean Biswanath Duplicate Patients

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/172693472

## Because

We accidentally ended up syncing patients to Bangladesh production after a trial import which caused the patients (and their data) to be duplicated. Relevant [Slack](https://simpledotorg.slack.com/archives/CFHMC60P5/p1588676903213000) thread.

## This addresses

Adds a rake task to delete and port patient data sensibly. Strategy is as follows:

Duplicate patients are identified as those registered by "Akshay Gupta". We attempt to match these dupes with real patients using different criteria increasing in specificity until a single match is found. Once found, any patient activity or modifications (BPs, address changes, etc) are ported from the dupe to the real patient, and the dupe is discarded.

Patients with ambiguous matches: We have verified that the ambiguous patients have no activity against them, so they will be discarded

Patients with no matches: If a "real" patient doesn't exist, then the dupe patient _is_ the real patient, now. They must have been skipped in the final Biswanath import due to incomplete data. Their registration user is updated to the import bot user.

## Summary of impact

* There are 2322 dupe patients created by the "Akshay Gupta" user. Out of these:
  * For 2224 patients, we can identify an exact matching real patient, based on search by (name and age), (name, age, and address), or (name, age, address, and phone).
    * 8 out of 2224 dupe patients have legitimate activity against them (recorded BP, scheduled appointment, etc). We will port this activity to the real patient record.
    * After porting activity, all 2224 will be soft-deleted
  * 76 patients have no matching "real" patient, so these 76 are the real patients now. We will reassign their registration user to the bot user `bishwanath-import-user`
  * 22 patients more than one potential matching real patient. However, we're fortunate that these 22 patients don't have any legitimate activity against them. We will soft-delete them.


lib/tasks/scripts/clean_biswanath_dupes.rb
```diff
@@ -0,0 +1,248 @@
...
+      real_patient_id = exact_matches[patient.id]
+
+      blood_pressures = patient.blood_pressures.where.not(id: bp_denylist).each { |record| record.update!(patient_id: real_patient_id) }
+      blood_sugars = patient.blood_sugars.each { |record| record.update!(patient_id: real_patient_id) }
+
+      port_bp_encounters(real_patient_id, blood_pressures)
+      port_bs_encounters(real_patient_id, blood_sugars)
+      patient.appointments.each { |record| record.update!(patient_id: real_patient_id) }
+      patient.prescription_drugs.each { |record| record.update!(patient_id: real_patient_id) }
+      patient.medical_history&.update!(patient_id: real_patient_id)
```
kitallis, line 114: I assume we don't need to add `Address` or `PatientPhoneNumber` here since we didn't find any updates to this?

rsanheim:
From skimming the slack thread, it looks like this impacts a small amount of patients, correct? Could you summarize the data that _should_ be merged / removed in this PR to make things easier to track?

My other question is how did you test this? This is a lot of code modifying production data, and it would make me nervous to run this sort of thing w/o some amount of automated tests + manual verification.

harimohanraj89:
I've added a summary of impact to the PR desc 👍

lib/tasks/scripts/clean_biswanath_dupes.rb
```diff
@@ -0,0 +1,223 @@
...
+      matches = match_by_name_age(patient)
+      if matches.count == 0
+        (no_matches << patient) && next
+      end
+
+      if matches.count == 1
+        (exact_matches[patient.id] = matches.first.id) && next
+      end
+
+      matches = match_by_name_age_address(patient)
```
vkrmis, line 44: This method should just be `match_by_address`, since the `match_by_name_age` method does the rest.

lib/tasks/scripts/clean_biswanath_dupes.rb
```diff
@@ -0,0 +1,223 @@
...
+  end
+
+  def match_by_name_age(patient)
+    Patient.where(age: patient.age, full_name: patient.full_name).where.not(id: patient.id)
+  end
+
+  def match_by_name_age_address(patient)
+    address_attrs = patient.address.slice(:street_address, :village_or_colony, :district, :state, :country, :pin)
+
+    match_by_name_age(patient).joins(:address).where(addresses: address_attrs)
```
vkrmis, line 194: Why is this necessary? We already do the name and age match separately

lib/tasks/scripts/clean_biswanath_dupes.rb
harimohanraj89: Ah, I see. Unfortunately, we don't re-use the `matches` from the previous method call, we start a fresh search for "by name, age, and address". It's just a design choice we didn't think much about for a one-off task.

Given that this starts a fresh search, do you think that this naming makes sense? Does this also explain your other comment?

PR #938, opened 2020-05-06
Add strong, entropy-based password checks

rsanheim:
Uses https://github.com/bdmac/strong_password to add entropy-based
password strength checking. This gem follows the advice laid out by the
NIST, which is summarized in this helpful Stack Overflow post:

https://stackoverflow.com/questions/48345922/reference-password-validation

The basic idea is to calculate the complexity of the password, in
addition to restricting the top 500 most commonly re-used passwords
across the internet. This means there aren't arbitrary restrictions like
"use 1 digit, 1 upper case, 1 lower case, and 1 special character"...
so more memorable passphrase based approaches will work.

https://www.useapassphrase.com/ was also useful to see how easily some "complex" passwords can be guessed, and how much stronger a passphrase is.

**Story card:** https://www.pivotaltracker.com/n/projects/2184102

Here is the UI as it stands with the updated error messaging:

image

## Because

We want simple.org to be secure

## This addresses

Password restrictions


rsanheim:
Obligatory https://xkcd.com/936/

image

config/locales/en.yml
```diff
@@ -36,6 +36,10 @@ en:
   time:
     formats:
       default: "%d-%^b-%Y %I:%M %p"
+  errors:
+    messages:
+      password:
+        password_strength: "is too weak. Please avoid common words and reptitive / sequential characters."
```
claudiovallejo, line 42: Copy suggestion: `This password is too easy to guess. Change your password to something stronger, like "<passphrase>"`

.env.development
```diff
@@ -112,7 +112,7 @@ OWNER_EMAILS='Owner <owner@test.com>, Another Owner <another-owner@test.com>'
 PROGRAM_INCEPTION_DATE=2018-01-01
 SUPERVISOR_EMAILS='Supervisor <supervisor@test.com>'
 
-SEED_GENERATED_ADMIN_PASSWORD='password'
+SEED_GENERATED_ADMIN_PASSWORD='baby monkey honey badger star'
```
harimohanraj89, line 115: Ah, my favorite type of star.

spec/controllers/email_authentications/invitations_controller_spec.rb
```diff
@@ -5,6 +5,7 @@
     @request.env['devise.mapping'] = Devise.mappings[:email_authentication]
     admin = create(:admin, :owner)
     sign_in(admin.email_authentication)
+    ActionMailer::Base.deliveries.clear
```
harimohanraj89, line 8: Looks like the spec on line 63 is failing because it expects a delivery to be available 🤔

app/models/email_authentication.rb
```diff
@@ -1,9 +1,24 @@
...
+
+  private
+
+  def cannot_use_example_password
+    return unless password.present?
+    if password == EXAMPLE_PASSPHRASE
+      errors.add(:password, "cannot match the example password")
+    end
+  end
+
```
harimohanraj89, line 23: 👾

app/models/email_authentication.rb
```diff
@@ -1,9 +1,24 @@
...
   devise :database_authenticatable, :invitable, :lockable, :recoverable,
          :rememberable, :timeoutable, :trackable, :validatable, validate_on_invite: true
 
   has_one :user_authentication, as: :authenticatable
   has_one :user, through: :user_authentication
 
   delegate :full_name, :resources, :role, :organization, to: :user, allow_nil: true
+
+  validates :password, password_strength: {use_dictionary: true}, allow_nil: true
```
harimohanraj89, line 12: Spaces inside the hash?

PS we talked about a styleguide tool that's friendlier than Rubocop. Might be a good idea to get that in sooner rather than later 😄 

Feel free to ignore my style-related comments until then if you'd like.

spec/controllers/email_authentications/invitations_controller_spec.rb
```diff
@@ -102,22 +103,22 @@
       end
 
       it 'responds with bad request email already exists' do
-        EmailAuthentication.create!(email: email, password: Faker::Internet.password)
+        EmailAuthentication.create!(email: email, password: generate(:strong_password))
```
harimohanraj89, line 106: Curious, where does `generate` come from? Does `strong_password` register some sort of generator with Rspec or FactoryBot?

spec/factories/users.rb
```diff
@@ -51,10 +51,14 @@
     factory :user_created_on_device, traits: [:with_phone_number_authentication]
   end
 
+  sequence(:strong_password) do |n|
```
rsanheim, line 54: @harimohanraj89 this creates the generator w/ the name `strong_password`

config/locales/en.yml
rsanheim: Updated with that.

spec/models/email_authentication_spec.rb
```diff
@@ -1,8 +1,46 @@
...
 
 RSpec.describe EmailAuthentication, type: :model do
+  WEAK_PASSWORD_ERROR = "Please choose a stronger password with at least 8 characters. Try a mix of letters, numbers, and symbols."
+
   describe 'Associations' do
     it { should have_one(:user_authentication) }
     it { should have_one(:user).through(:user_authentication) }
   end
+
+  it "should require a non blank password" do
```
harimohanraj89, line 11: Consider dropping the "should" - `it "requires...", it "allows..."`

rsanheim: Updated. 👍

rsanheim:
> If I understand correctly, this change will affect new admins, and admins that trigger a password reset.

Yup, that is correct!

PR #939, opened 2020-05-06
Added schema changes caused by db:migrate

timcheadle:
**Story card:** NONE

## Because

Because of recent changes, every time I run `rails db:migrate`, even on `master`, I get this diff of changes. @rsanheim reported the same issue.

## This addresses

This covers only the changes caused to the schema by running migrations. I'm not sure why these aren't checked in, or if they even should be.


timcheadle:
@kitallis can you run migrations on your side and see if these same changes show up? I suspect there might be a software version mismatch or something that's causing some reordering. The removed index is what worries me the most.

db/schema.rb
```diff
@@ -570,6 +594,138 @@
      FROM (appointments ap
        JOIN facilities f ON ((ap.facility_id = f.id)));
   SQL
+  create_view "latest_blood_pressures_per_patient_per_days", materialized: true, sql_definition: <<-SQL
+      WITH latest_bp_per_patient_per_month AS (
```
harimohanraj89, line 598: I regenerated the schema locally on this branch to see if there were any diffs. The only issue I found is that this view is wrong. I reckon you ran the migration that created this view while it was still in development. During its development we found a bug in the view and [fixed it here](https://github.com/simpledotorg/simple-server/commit/a2eb5bd14abab46a39cea33d10c3f4a87aef36fa#diff-46826eac74e5c6efde6ce2c953d2ca03). 

I will push an update.

The other changes in here look legitimate. Looks like our schema was out of date?

db/schema.rb
```diff
@@ -295,7 +295,6 @@
     t.uuid "patient_business_identifier_id", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["patient_business_identifier_id"], name: "index_passport_auth_on_business_id"
```
harimohanraj89, line 298: This is a recent model we added. Our migrations on `master` don't show us adding this index. The addition of the index appears to be a mistake I made [here](https://github.com/simpledotorg/simple-server/commit/b3f7aff519d7c3690e530abd9d8095aa35a60a89). It's correct for it to be removed.

Now, if we _do_ want to index this field, we can add a separate migration to do so :)

db/schema.rb
```diff
@@ -570,6 +594,138 @@
...
+      latest_bp_per_patient_per_quarter.deleted_at,
+      latest_bp_per_patient_per_quarter.month,
+      latest_bp_per_patient_per_quarter.quarter,
+      latest_bp_per_patient_per_quarter.year,
+      lag(latest_bp_per_patient_per_quarter.bp_facility_id, 1) OVER (PARTITION BY latest_bp_per_patient_per_quarter.patient_id ORDER BY latest_bp_per_patient_per_quarter.bp_recorded_at, latest_bp_per_patient_per_quarter.bp_id) AS responsible_facility_id
+     FROM latest_bp_per_patient_per_quarter;
+  SQL
+  add_index "latest_blood_pressures_per_patient_per_quarters", ["bp_id"], name: "index_latest_blood_pressures_per_patient_per_quarters", unique: true
+
+  create_view "latest_blood_pressures_per_patients", materialized: true, sql_definition: <<-SQL
```
harimohanraj89, line 709: This view is new.

db/schema.rb
```diff
@@ -468,6 +467,31 @@
   add_foreign_key "patients", "addresses"
   add_foreign_key "protocol_drugs", "protocols"
 
+  create_view "blood_pressures_per_facility_per_days", materialized: true, sql_definition: <<-SQL
```
harimohanraj89, line 470: Double-checked. All the view stuff is reordering. I believe Rob also found that this is a known issue with `scenic`

kitallis:
I just applied the same things as @harimohanraj89 – found identical issues on my machine. The fixes look good to me.

PR #940, opened 2020-05-07
Dashboard reports should be HTN-only

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/170179855

## Because

Simple is primarily a hypertension-tracking app. All dashboard reports must ensure that patients are diagnosed with HTN.

## This addresses

This branch is based off of https://github.com/simpledotorg/simple-server/pull/897 which already converts the `follow_up_patients` numbers in `DistrictAnalyticsQuery` & `FacilityAnalyticsQuery` to be HTN-only.

Hence, this PR only attempts to migrate: 

* `DistrictAnalyticsQuery` registrations
* `CohortAnalyticsQuery`
* and `MyFacility` dashboards.

Adds scopes and joins to all dashboards to account for the Patient's diagnosis to be HTN = yes (through `MedicalHistory`).

**Reports to be changed:**

- [x] District-level cohort charts
- [x] Facility-level cohort charts
- [x] `DistrictAnalyticsQuery#total_registered_patients`
- [x] `DistrictAnalyticsQuery#registered_patients_by_period`
- [x] `FacilityAnalyticsQuery#total_registered_patients`
- [x] `FacilityAnalyticsQuery#registered_patients_by_period`
- [x] `MyFacilities::OverviewQuery`
- [x] `MyFacilities::BloodPressureControlQuery`
- [x] `MyFacilities::RegistrationsQuery`

This also adds a couple of supplementary changes:
- Removes the `:hypertension` trait from patients, builds a `hypertension_yes` medical history by default (de7760c)
- Adds a `:with_encounter` trait to `blood_sugars` and `blood_pressures` factories (f5bcf1a)
- Clear up encounters created in #897 using `with_encounter` (7e156e7)


spec/factories/patients.rb
```diff
@@ -30,10 +30,7 @@
                              assigning_user_id: registration_user.id })
     end
     reminder_consent { Patient.reminder_consents[:granted] }
-
-    trait :hypertension do
-      medical_history { build(:medical_history, :hypertension_yes, patient_id: id) }
-    end
```
harimohanraj89, line 36: Good call 😄 @prabhanshuguptagit heads up, that the default patient factory is "hypertension_yes" with this change.

spec/models/facility_spec.rb
```diff
@@ -191,4 +191,17 @@
   describe 'Behavior' do
     it_behaves_like 'a record that is deletable'
   end
+
+  describe '#cohort_analytics' do
+    it 'should only consider registered hypertensive patients' do
```
harimohanraj89, line 196: I suspect we'll need to add a bunch of contexts like this across the different touchpoints?

app/queries/my_facilities/blood_pressure_control_query.rb
```diff
@@ -53,24 +53,27 @@ def cohort_missed_visits_count_by_facility
...
   end
 
   private
 
   attr_reader :facilities
 
   def quarterly_registrations
-    patients = Patient.where(registration_facility: facilities)
+    patients = Patient
+                 .hypertension_only
```
harimohanraj89, line 75: I love how easy this is. As mentioned in my review on #897, I'd love for this scope to be named something like `with_hypertension` or `diagnosed_with_hypertension`, since `hypertension_only` makes it seem like patients with both HTN and DM are excluded from this scope (but they're not).

spec/factories/blood_pressures.rb
```diff
@@ -11,6 +11,16 @@
     association :user, strategy: :create
     association :patient, strategy: :create
 
+    trait(:with_encounter) do
```
harimohanraj89, line 14: ❤️

app/queries/my_facilities/blood_pressure_control_query.rb
```diff
@@ -53,24 +53,27 @@ def cohort_missed_visits_count_by_facility
...
-                          .where('recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
-                          .where(registration_facility: facilities)
+                            .hypertension_only
+                            .where('recorded_at < ?', Time.current.beginning_of_day - REGISTRATION_BUFFER)
+                            .where(registration_facility: facilities)
   end
 
   def overall_controlled_bps
     @overall_controlled_bps ||=
       LatestBloodPressuresPerPatient
```
harimohanraj89, line 63: PS, what's our plan for these views? Do we plan to filter the results returned by them, or filter the data that goes into them in the first place?

kitallis: So from my understanding so far, only 2 view-implementations actually need to change, the rest can be done using scopes alone:

* The Inactive facilities (populated by `MyFacilities::OverviewQuery`)

This is because this view doesn't join on `Patient` at all. If we joined with patients, we can scope this query up in application code.

*  `MyFacilities::RegistrationsQuery` (period-wise registrations)

This is a weird one. This does join on patients but returns an aggregate. So we might have to stick the htn-only clause in the view for this to work correctly.

spec/models/facility_spec.rb
kitallis: We'll need a few in `MyFacilities`, the Cohort ones are covered in this PR. The Dashboard ones are covered in https://github.com/simpledotorg/simple-server/pull/897.

kitallis:
> Some early review, I think this the right direction, and looks like what I expected as well. Gotta love that we have a controlled number of "doors into the data", and we're able to apply a hypertension scope on all of them.
> 
> @kitallis you had mentioned that there might be a place that was executing some raw SQL, did you find those? Did they need to be addressed as well?

There were two places in particular, one was `CohortAnalytics` which turned out to be very easy. And the other was the dreaded [User-level follow ups query](https://github.com/simpledotorg/simple-server/blob/master/app/queries/facility_analytics_query.rb#L55). The scoping on both turned out to be quite straightforward.

I'll definitely dig more to see if we aren't missing anything.

spec/factories/patients.rb
prabhanshuguptagit: Is this wrt seed data gen? I checked, that is not being affected (since we use payloads and medical histories are a separate payload). Thanks for the heads up.

db/views/patient_registrations_per_day_per_facilities_v02.sql
```diff
@@ -0,0 +1,13 @@
+SELECT COUNT(patients.id) AS registration_count,
+       patients.registration_facility_id AS facility_id,
+       facilities.deleted_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE')) AS deleted_at,
+cast(EXTRACT(DOY FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS day,
+cast(EXTRACT(MONTH FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS month,
+cast(EXTRACT(QUARTER FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS quarter,
+cast(EXTRACT(YEAR FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS year
+FROM patients
+INNER JOIN facilities ON patients.registration_facility_id = facilities.id
+INNER JOIN medical_histories ON patients.id = medical_histories.patient_id
```
kitallis, line 10: @prabhanshuguptagit Quick minor thing: we can do a `ON (patients.id = medical_histories.patient_id AND medical_histories.hypertension = 'yes')` style `join` filter. It'll keep the definition in a single line.

db/views/blood_pressures_per_facility_per_days_v02.sql
```diff
@@ -0,0 +1,19 @@
+WITH latest_bp_per_patient_per_day AS
+ (SELECT DISTINCT ON (facility_id, blood_pressures.patient_id, year, day)
+      blood_pressures.id AS bp_id,
+      blood_pressures.facility_id AS facility_id,
+      cast(EXTRACT(DOY FROM blood_pressures.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS day,
+      cast(EXTRACT(MONTH FROM blood_pressures.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS month,
+      cast(EXTRACT(QUARTER FROM blood_pressures.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS quarter,
+      cast(EXTRACT(YEAR FROM blood_pressures.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS year
+  FROM blood_pressures
+  INNER JOIN medical_histories ON (blood_pressures.patient_id = medical_histories.patient_id AND medical_histories.hypertension = 'yes')
```
kitallis, line 10: @prabhanshuguptagit I was wondering if we can achieve this by simply merging patients to this view and scoping via `hypertension_only` in the application code and not bother with merging MH + join filter?

prabhanshuguptagit: We can't do that since the `count` aggregation happens in the view. The only way to make it happen is if we do the count also in AR.

kitallis: Ugh, yeah, missed the second part of this view.

spec/factories/blood_sugars.rb
```diff
@@ -6,9 +6,20 @@
     device_created_at { Time.current }
     device_updated_at { Time.current }
     recorded_at { device_created_at }
-    facility
-    user
-    patient
+
+    association :facility, strategy: :create
```
kitallis, line 10: @prabhanshuguptagit Why these changes?

prabhanshuguptagit: Just to keep things consistent across `blood_sugars` and `blood_pressures`.

spec/queries/my_facilities/registrations_query_spec.rb
```diff
@@ -72,20 +58,19 @@
           end
         end
 
-        specify { expect(registrations_query.registrations.count).to eq(included_timestamps.count) }
+        it 'counts registrations diagnosed with hypertension in the last_n months' do
```
kitallis, line 61: I think the `specify` is clean enough, no need to break it up and claim HTN explicitly

spec/queries/my_facilities/registrations_query_spec.rb
```diff
@@ -48,22 +35,21 @@
           end
         end
 
-        specify { expect(registrations_query.registrations.count).to eq(included_timestamps.count) }
+        it 'counts registrations diagnosed with hypertension in the last_n quarters' do
```
kitallis, line 38: I think the specify is clean enough, no need to break it up and claim HTN explicitly

spec/queries/my_facilities/blood_pressure_control_query_spec.rb
```diff
@@ -209,25 +246,25 @@
       end
 
       describe '#cohort_registrations' do
-        specify do
+        it 'counts registrations diagnosed with hypertension' do
```
kitallis, line 249: We can wrap the `specify` under a context of `with htn` or some such and leave the `specify` intact.

db/views/blood_pressures_per_facility_per_days_v02.sql
kitallis, line 10: This is the new clause being added to support HTN-only.

db/views/patient_registrations_per_day_per_facilities_v02.sql
```diff
@@ -0,0 +1,12 @@
+SELECT COUNT(patients.id) AS registration_count,
+       patients.registration_facility_id AS facility_id,
+       facilities.deleted_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE')) AS deleted_at,
+cast(EXTRACT(DOY FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS day,
+cast(EXTRACT(MONTH FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS month,
+cast(EXTRACT(QUARTER FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS quarter,
+cast(EXTRACT(YEAR FROM patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE'))) as text) AS year
+FROM patients
+INNER JOIN facilities ON patients.registration_facility_id = facilities.id
+INNER JOIN medical_histories ON (patients.id = medical_histories.patient_id AND medical_histories.hypertension = 'yes')
```
kitallis, line 10: This is the new clause being added to support HTN-only.

spec/controllers/api/v2/medical_histories_controller_spec.rb
```diff
@@ -45,9 +45,9 @@ def create_record_list(n, options = {})
     let(:facility_in_same_group) { FactoryBot.create(:facility, facility_group: request_user.facility.facility_group) }
     let(:facility_in_another_group) { FactoryBot.create(:facility) }
 
-    let(:patient_in_request_facility) { FactoryBot.create(:patient, registration_facility: request_facility) }
-    let(:patient_in_same_group) { FactoryBot.create(:patient, registration_facility: facility_in_same_group) }
-    let(:patient_in_another_group) { FactoryBot.create(:patient, registration_facility: facility_in_another_group) }
+    let(:patient_in_request_facility) { FactoryBot.build(:patient, registration_facility: request_facility) }
+    let(:patient_in_same_group) { FactoryBot.build(:patient, registration_facility: facility_in_same_group) }
+    let(:patient_in_another_group) { FactoryBot.build(:patient, registration_facility: facility_in_another_group) }
```
prabhanshuguptagit, line 50: This is a build since the patient factory creates a `medical_history` by default now, the controller spec create one separately and so there are two.

harimohanraj89: This seems like a problem with the way we're asserting things in the spec. Can we stop asserting counts and start asserting the presence of specific medical history IDs instead? 

`build` is much faster, and that's great, but I'm not sure we should use it to avoid this fragile situation.

spec/factories/patients.rb
```diff
@@ -30,9 +30,10 @@
                              assigning_user_id: registration_user.id })
     end
     reminder_consent { Patient.reminder_consents[:granted] }
+    medical_history { build(:medical_history, :hypertension_yes, patient_id: id) }
 
-    trait :hypertension do
-      medical_history { build(:medical_history, :hypertension_yes, patient_id: id) }
+    trait :hypertension_no do
```
harimohanraj89, line 35: Could we use `without_hypetension` for readability? 😄

spec/models/organization_district_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+RSpec.describe OrganizationDistrict, type: :model do
+  describe '#cohort_analytics' do
+    it 'should only consider registered hypertensive patients' do
```
harimohanraj89, line 5: Consider `it 'only considers...'`

spec/models/patient_registrations_per_day_per_facility_spec.rb
```diff
@@ -31,5 +37,9 @@
     it 'should return at least one row per facility' do
       expect(query_results.pluck(:facility_id).uniq).to match_array(facilities.map(&:id))
     end
+
+    it 'counts only patients who are diagnosed hypertensive' do
+      expect(query_results.pluck(:registration_count).uniq).to eq([1])
```
harimohanraj89, line 42: We should assert that specific patient IDs are included and excluded in this spec. This spec right now would pass if we flipped to `hypertension=no` patients by accident.

spec/queries/my_facilities/blood_pressure_control_query_spec.rb
```diff
@@ -87,27 +105,29 @@
         end
       end
 
-      describe '#cohort_registrations' do
-        specify do
-          expect(query.cohort_registrations.count).to eq(7)
+      context 'considers only htn diagnosed patients' do
+        describe '#cohort_registrations' do
+          specify do
+            expect(query.cohort_registrations.count).to eq(7)
```
harimohanraj89, line 111: Again, the counts here are iffy. Of course, this spec has always been like this, so I'll let you decide whether you wanna take a stab at updating some of these.

db/views/blood_pressures_per_facility_per_days_v02.sql
prabhanshuguptagit: This is used in `overview_query` to calculate total BPs and Inactive facilities.

db/views/patient_registrations_per_day_per_facilities_v02.sql
prabhanshuguptagit: This is used in `registrations_query` for the registrations page in my facilities.

db/views/blood_pressures_per_facility_per_days_v02.sql
prabhanshuguptagit: @harimohanraj89 we can't do that since the count aggregation happens in the view. The only way to make it happen is if we do the count also in AR.

spec/models/patient_registrations_per_day_per_facility_spec.rb
prabhanshuguptagit: This view returns the counts of patients, so there is no way to do ID matches. Also, the count check works since facilities would have `2 bps` on `1.day.ago` if `hypertension=no` was accidentally flipped; the spec will fail in that case.

db/views/blood_pressures_per_facility_per_days_v02.sql
harimohanraj89: Ha, this is exactly what Kitty asked you a while ago, isn't it. Thanks for answering me again 😄

prabhanshuguptagit: Yes😛

spec/models/patient_registrations_per_day_per_facility_spec.rb
harimohanraj89: Ah, so this is also a consequence of the view aggregating the counts. In that case, yup, asymmetric data should do the trick 👍

spec/controllers/api/v2/medical_histories_controller_spec.rb
kitallis: I think we should avoid touching this territory for the scope of this PR to not explore.

prabhanshuguptagit: This change touches the shared specs and changing them is quite tricky. I would like to avoid going down that road unless really necessary.

I think it's ok to keep the `build`, since it's just an artefact of how our factories work now. It _will_ create a medical history in the background, which we don't want in this case.

harimohanraj89: Yup, that sounds sensible. 👍

PR #941, opened 2020-05-07
Seed data improvements

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/story/show/172734259

## Because

We need some improvements to the seed data quality.

## This addresses

- Randomize Facility enable_diabetes_management
- Randomize medical histories better


app/jobs/seed_users_data_job.rb
```diff
@@ -47,7 +47,8 @@ def patient_traits
         build_fn: -> (args) {
           return if args[:patient].medical_history
           build_medical_history_payload_current(FactoryBot.build(:medical_history,
-                                                                 [:hypertension_yes, :diabetes_yes].sample,
+                                                                 [:hypertension_yes, :hypertension_no].sample,
+                                                                 [:diabetes_yes, :diabetes_no, :diabetes_unknown].sample,
```
harimohanraj89, line 51: Neat. Were you able to confirm with Daniel that this is the desired change?

prabhanshuguptagit: I was able to confirm this with prod data. There are no `nils` in any of these fields. Since the `enable_diabetes_management` doesn't imposing any hard constraints, we should be good to generate it this way. Trying to exactly match prod data patterns will be overkill anyways.

PR #950, opened 2020-05-07
Fix patient_business_identifier spec

rsanheim:
Having two validations of the same field and type probably doesn't do
what we want it too, especially in Rails 5.2 and newer. I don't know if
they have any sort of 'merge' behavior.

I unrolled to a custom validator, and also unrolled the specs a bit for
clarity.

**Story card:** n/a

Cherry picked from Rails 5.2 upgrade - #915 .


rsanheim:
This ended up going out with Rails 5.2, which got merged last week. So no longer need this as a separate PR.

PR #960, opened 2020-05-12
Change setup and readme to set bundler version

rsanheim:
If you don't set this, you'll get bundler 2, which is not yet setup for
our deployments and will also conflict with our Gemfile.lock.

**Story card:** card for the eventual upgrade: https://www.pivotaltracker.com/story/show/172819142

## Because

Tried to setup simple-server on a different machine and ran into this problem.

## This addresses

Making our dev environment easy to setup.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -4,6 +4,8 @@
   RUBY_VERSION=$(cat .ruby-version)
 } > /dev/null 2>&1
 
+BUNDLER_VERSION=1.17.3
+
 init (){
   verify_ruby_version && verify_bundled_gems && verify_yarn_packages
   verify_foreman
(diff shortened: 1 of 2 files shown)


bin/setup
```diff
@@ -117,7 +119,7 @@ verify_ruby_version (){
 verify_bundled_gems (){
   echo 'Updating gems...'
   {
-    (gem list -i bundler || gem install bundler) && bundle install --binstubs=./bundle/bin
+    (gem list -i -v ${BUNDLER_VERSION} bundler || gem install bundler -v ${BUNDLER_VERSION}) && bundle _${BUNDLER_VERSION}_ install --binstubs=./bundle/bin
```
harimohanraj89, line 122: TIL `bundle _version_ install`

rsanheim: yeah it is some weird syntax, but has saved me in some tough spots before

rsanheim:
> Looks great. Can I ask you to craft a maintenance story for us to upgrade to Bundler 2 and slap a link in this PR for reference?

Yup: https://www.pivotaltracker.com/story/show/172819142

PR #957, opened 2020-05-12
Help translations

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172783902

This PR aggregates several transifex translations of the Help page into a single PR.

* #943 
* #944 
* #945 
* #948 
* #955

PR #961, opened 2020-05-13
Add copy of en.html to the new directory

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172783902

## Because

#957 didn't work, since Transifex HTML translations are not intelligent enough to update the structure of the target HTML files.

## This addresses

This PR adds a copy of the new-design `en.html` file to a subdirectory so it can be independently translated accurately. Once translation is finished, and the `config/locales/api/help/new/` directory is fully populated, we can replace all existing help files with the new ones.

The steps to follow this PR are:
* Update Transifex configuration to include this new file as a source
* Translate the new source into all target languages
* Add the translated files into the new directory
* Replace the old help files with the new translated ones


harimohanraj89:
Yup, a copy 👍 Sure, `upcoming` sounds good.

harimohanraj89:
Yup, a copy 👍 Sure, `upcoming` sounds good.

PR #963, opened 2020-05-13
Clean up deprecation warnings

rsanheim:
Fixing deprecation warnings to prep for Rails 6.0, and also so we don't have tons of noise in our build or logs.

Fixes warning: DEPRECATION WARNING: The success? predicate is deprecated and will be removed in Rails 6.0. Please use successful? as provided by Rack::Response::Helpers.

**Story card:**  https://www.pivotaltracker.com/story/show/172821415

## Because

We want useful logs and test output

## This addresses

Deprecations that our Rails 5.2 upgrade revealed


lib/tasks/scripts/clean_biswanath_dupes.rb
```diff
@@ -166,10 +167,15 @@ def deactivate_ambiguous_patients
     ambiguous_matches.each(&:discard_data)
   end
 
+  def log(msg)
+    puts msg if verbose
+  end
```
harimohanraj89, line 172: Yup, I like this verbose pattern. This file has been axed 😅 but I will leverage this idea in #969 👍

app/queries/facility_analytics_query.rb
```diff
@@ -57,7 +57,7 @@ def follow_up_patients_by_period
                 .where(blood_pressures: { facility: @facility })
                 .select("DISTINCT ON (blood_pressures.patient_id, #{BloodPressure.date_to_period_sql(@period)}) blood_pressures.user_id AS user_id, patients.deleted_at")
                 .select("blood_pressures.recorded_at AS bp_recorded_at")
-                .order('blood_pressures.patient_id', BloodPressure.date_to_period_sql(@period), 'blood_pressures.recorded_at'),
+                .order('blood_pressures.patient_id', Arel.sql(BloodPressure.date_to_period_sql(@period)), 'blood_pressures.recorded_at'),
```
harimohanraj89, line 60: Ah, I see. `Arel.sql` assumes its contents are safe, and this is why you added protections against unknown values for `@period` 👍 

Can the whole list of columns here be wrapped in an `Arel.sql`?

rsanheim:
> Looks good. My only question is: Why use `Arel.sql` in the first place?

If we don't for these call sites, we get a flurry of deprecation notices in our test output.  You can read a long saga around this deprecation here: https://github.com/rails/rails/issues/32995

PR #964, opened 2020-05-13
Fast-follow fixes for overdue listing

rsanheim:
**Story card:** quick fixes for overdue list: https://www.pivotaltracker.com/story/show/172537522

* apply the filters to the download links by merging params
* ensure consistent order between the web UI and downloads by adding `id` to the order by clause

## Because

We weren't applying the search filters for CSV downloads, as params weren't getting merged.

## This addresses

This fixes that issue


app/controllers/appointments_controller.rb
```diff
@@ -22,7 +22,7 @@ def index
     if current_facility
       @patient_summaries = @patient_summaries.where(next_appointment_facility_id: current_facility.id)
     end
-    @patient_summaries = @patient_summaries.order(risk_level: :desc, next_appointment_scheduled_date: :desc)
+    @patient_summaries = @patient_summaries.order(risk_level: :desc, next_appointment_scheduled_date: :desc, id: :asc)
```
harimohanraj89, line 25: Wowza. Patients with the same risk level and appointment date were being ordered randomly each time? Fun 😅 

This is a good callout too. Since we use UUIDs heavily across the board, often not auto-generated, we can't rely on the default-by-ID ordering that we'd usually have.

PR #968, opened 2020-05-14
Seed Review Apps

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172833585

Seed Heroku Review Apps using the new `rake db:seed_users_data` task.


bin/heroku-setup
```diff
@@ -2,7 +2,4 @@
 
 bundle exec rails db:schema:load:with_data
 bundle exec rails db:seed
-
-if [ ! -z "${SUPERADMIN_PASSWORD}" ]; then
-  bundle exec rake create_admin_user[admin,admin@simple.org,${SUPERADMIN_PASSWORD}]
-fi
+bundle exec rails db:seed_user_data
```
prabhanshuguptagit, line 5: 🐞 typo: `seed_users_data`

Not sure if this is being done somewhere else in the setup but `db:seed_users_data` expects sidekiq and the app to be running.

harimohanraj89: I'm guessing that Heroku will provision these resources before running this script. We'll find out 😄

harimohanraj89:
Needed to add the `low` queue to the Heroku sidekiq config as well, and the seeds are now working 👍

bin/heroku-setup
harimohanraj89: It works!

PR #969, opened 2020-05-14
Correct BD medication dosages

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/171377703

## Because

Several protocol dosages of drugs were imported as "0 mg" in our initial import.

## This addresses

Setting all "0 mg" records for the three protocol drugs to their respective protocol dosages.

### Dosages

* Amlodipine: 5 mg
* Losartan Potassium: 50 mg
* Hydrochlorothiazide: 12.5 mg

### Estimated Impact

* Amlodipine: 938 records
* Losartan Potassium: 1524 records
* Hydrochlorothiazide: 19 records


lib/tasks/scripts/correct_bangladesh_medication_dosages.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  attr_reader :dryrun, :verbose
+
+  def initialize(dryrun: false, verbose: true)
+    @dryrun = dryrun
+    @verbose = verbose
```
kitallis, line 18: This is to prevent it from polluting the specs I assume? Nice.

harimohanraj89: Correct, inspired by Rob's work in https://github.com/simpledotorg/simple-server/pull/963

PR #971, opened 2020-05-14
Use WhatsApp for appointment notifications

timcheadle:
**Story card:** https://www.pivotaltracker.com/story/show/168072307

## Because

We want to send WhatsApp messages instead of SMS in India.

## This addresses

I originally sought to just use the `send_reminder_whatsapp` method in the `SmsNotificationSerice`. However, with a little TLC I was able to clean up the entire notification service and worker structure for clarity and having each object address a singular concern.

The new `NotificationService` is effectively a Twilio interface, knowing how to send SMS and WhatsApp but being agnostic as to who receives the message and what the body contains.

The revised `ApppointmentNotification::Worker` now knows very little about Twilio, and simply formats and sends a WhatsApp message to the right patient. 

I also added a new `sms.missed_visit_whatsapp_reminder` locale string and copied the same message formats we were using for SMS.

I've wrapped all of this in a feature flag, `WHATSAPP_APPOINTMENT_REMINDERS`. I've also renamed the `SMS_REMINDERS` feature flag to `APPOINTMENT_REMINDERS` since that's what it actually controls, enabling the entire reminder feature at all.

**Note:** In order to get this to work, I had to modify everything that was talking to the old `SmsNotificationService`. This touched a LOT of code, even in places (like specs) that should not even have knowledge of how SMS were being sent, just testing that they were sent at all.


app/services/notification_service.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def default_country_code
+    Rails.application.config.country[:sms_country_code]
+  end
+
+  def send_twilio_message(sender_number, recipient_number, message)
+    client.messages.create(
+      from: sender_number,
+      to: recipient_number,
+      status_callback: twilio_callback_url,
```
kitallis, line 41: We should probably not have `status_callback` for OTP SMSes, the Twilio callback is currently quite tied to `Communication` – so it should just be for reminders.

timcheadle: Agree; fixed this so that callback URLs are optional.

spec/services/appointment_notification_service_spec.rb
```diff
@@ -36,25 +33,26 @@
     end
 
     it 'should skip sending reminders for appointments for which reminders are already sent' do
-      AppointmentNotificationService.send_after_missed_visit(appointments: overdue_appointments, schedule_at: Time.current)
-      AppointmentNotification::Worker.drain
+      #AppointmentNotificationService.send_after_missed_visit(appointments: overdue_appointments, schedule_at: Time.current)
+      #AppointmentNotification::Worker.drain
```
rsanheim, line 37: Looks like this commented code can be removed?

spec/jobs/request_otp_sms_job_spec.rb
```diff
@@ -2,24 +2,17 @@
...
-
   let!(:user) { create(:user, phone_number: '1234567890') }
-  let!(:sms_notification_service) { double(SmsNotificationService.new(nil, nil)) }
-
-  it 'calls off to the SMSNotificationService to deliver the otp SMS' do
-    expect(SmsNotificationService)
-      .to receive(:new)
-            .with(user.phone_number, ENV['TWILIO_PHONE_NUMBER'])
-            .and_return(sms_notification_service)
+  let(:app_signature) { ENV["SIMPLE_APP_SIGNATURE"] }
```
rsanheim, line 6: Is it okay for this to be blank? I see its empty in our .env.test file.

app/services/notification_service.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def default_country_code
+    Rails.application.config.country[:sms_country_code]
+  end
+
+  def send_twilio_message(sender_number, recipient_number, message, callback_url=nil)
+    client.messages.create(
+      from: sender_number,
+      to: recipient_number,
+      status_callback: callback_url,
```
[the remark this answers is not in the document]
timcheadle: Just confirmed it works fine.

```
21:20:09 in simple-server on  replace-sms-with-whatsapp [$]
➜  curl -X POST https://api.twilio.com/2010-04-01/Accounts/AC009716268a00cb71364038c500d17ff2/Messages.json \
--data-urlencode "Body=McAvoy or Stewart? These timelines can get so confusing." \
--data-urlencode "From=+15005550006" \
--data-urlencode "StatusCallback=" \
--data-urlencode "To=+17041234567" \
-u AC009716268a00cb71364038c500d17ff2:23bb19d5f546d76202f72fad82ae1cce
{"sid": "SM10fbdecc784b46e79e8ba5881e554142", "date_created": "Wed, 20 May 2020 01:20:20 +0000", "date_updated": "Wed, 20 May 2020 01:20:20 +0000", "date_sent": null, "account_sid": "AC009716268a00cb71364038c500d17ff2", "to": "+17043009287", "from": "+15005550006", "messaging_service_sid": null, "body": "McAvoy or Stewart? These timelines can get so confusing.", "status": "queued", "num_segments": "1", "num_media": "0", "direction": "outbound-api", "api_version": "2010-04-01", "price": null, "price_unit": "USD", "error_code": null, "error_message": null, "uri": "/2010-04-01/Accounts/AC009716268a00cb71364038c500d17ff2/Messages/SM10fbdecc784b46e79e8ba5881e554142.json", "subresource_uris": {"media": "/2010-04-01/Accounts/AC009716268a00cb71364038c500d17ff2/Messages/SM10fbdecc784b46e79e8ba5881e554142/Media.json"}}%
```

spec/jobs/request_otp_sms_job_spec.rb
timcheadle: Yep, perfectly fine for it to be blank.

app/jobs/appointment_notification/worker.rb
```diff
@@ -4,52 +4,57 @@ class AppointmentNotification::Worker
...
       },
-      tags: { type: 'appointment-notification-job' })
+      tags: {
+        type: 'appointment-notification-job'
+      }
+    )
+  end
+
+  def callback_url
+    api_v3_twilio_sms_delivery_url(
```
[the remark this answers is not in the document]
timcheadle: It's a route helper:

```
➜  rails routes | ag api_v3_twilio
                  api_v3_twilio_sms_delivery POST                      /api/v3/twilio_sms_delivery(.:format)                                                   api/v3/twilio_sms_delivery#create {:format=>"json"}
```

app/services/notification_service.rb
```diff
@@ -0,0 +1,55 @@
+class NotificationService
```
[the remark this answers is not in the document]
timcheadle: You got it. This also opens up the possibility down the line of further abstracting this so we support not just SMS/WhatsApp via Twilio, but also Telegram, Line, WeChat, etc.

app/jobs/send_patient_otp_sms_job.rb
```diff
@@ -1,14 +1,17 @@
...
-    phone_number = patient_authentication.patient&.latest_mobile_number
 
+  def perform(passport_authentication)
+    phone_number = passport_authentication.patient&.latest_mobile_number
     return unless phone_number.present?
 
-    SmsNotificationService
-      .new(phone_number, ENV['TWILIO_PHONE_NUMBER'])
-      .send_patient_request_otp_sms(patient_authentication.otp)
+    NotificationService.new.send_sms(phone_number, otp_message(passport_authentication))
```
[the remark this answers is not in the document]
timcheadle: I didn't do that because I saw some value in the future of being able to access the instance's twilio client, perhaps for debugging or logging.

Also, I don't personally like handing method-specific attributes to the initializer, when you can theoretically use the service instance to send a bunch of messages at once.

app/services/notification_service.rb
```diff
@@ -0,0 +1,55 @@
...
+  DEFAULT_LOCALE = :en
+
+  attr_reader :client
+
+  def initialize
+    @client = Twilio::REST::Client.new(twilio_account_sid, twilio_auth_token)
+  end
```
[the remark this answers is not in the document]
timcheadle: That's much nicer.

app/services/appointment_notification_service.rb
```diff
@@ -7,7 +7,7 @@ def initialize(appointments:, days_overdue: 3, schedule_at:)
     @appointments = appointments
     @days_overdue = days_overdue
     @schedule_at = schedule_at
-    @communication_type = Communication.communication_types[:missed_visit_sms_reminder]
+    @communication_type = Communication.communication_types[:missed_visit_whatsapp_reminder]
```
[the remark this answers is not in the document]
timcheadle: I've tried to do this but may have missed something. Please have a look :)

app/controllers/api/v3/users_controller.rb
```diff
@@ -1,45 +1,46 @@
...
-
-  def register
-    user = User.build_with_phone_number_authentication(user_from_request)
-    return head :not_found unless user.registration_facility.present?
-
-    if user.phone_number_authentication.invalid?
-      return render json: {
-        errors: user.phone_number_authentication.errors
-      }, status: :bad_request
+    skip_before_action :validate_sync_approval_status_allowed, only: [:register, :find, :request_otp]
```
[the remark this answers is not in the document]
timcheadle: Fixed!

.env.development
```diff
@@ -52,10 +52,11 @@ ENABLE_FIXED_OTP_ON_REQUEST_FOR_QA=false
 ENABLE_GENERATE_ENCOUNTER_ID_ENDPOINT=true
 ENABLE_PATIENT_FOLLOWUPS=true
 ENABLE_PURGE_ENDPOINT_FOR_QA=false
-ENABLE_SMS_REMINDERS=true
+ENABLE_APPOINTMENT_REMINDERS=true
 ENABLE_SYNC_ENCOUNTERS=false
 ENABLE_SKIP_API_VALIDATION=false
 ENABLE_DIABETES_SUPPORT_IN_PROGRESS_TAB=true
+ENABLE_WHATSAPP_REMINDERS=false
```
harimohanraj89, line 59: `ENABLE_WHATSAPP_APPOINTMENT_REMINDERS`?

.env.development
```diff
@@ -52,10 +52,11 @@ ENABLE_FIXED_OTP_ON_REQUEST_FOR_QA=false
 ENABLE_GENERATE_ENCOUNTER_ID_ENDPOINT=true
 ENABLE_PATIENT_FOLLOWUPS=true
 ENABLE_PURGE_ENDPOINT_FOR_QA=false
-ENABLE_SMS_REMINDERS=true
+ENABLE_APPOINTMENT_REMINDERS=true
```
harimohanraj89, line 55: I guess we'll have to _add_ `ENABLE_APPOINTMENT_REMINDERS`, release this PR, then _remove_ `ENABLE_SMS_REMINDERS`, right? Just calling it out 👍

app/jobs/send_patient_otp_sms_job.rb
```diff
@@ -1,14 +1,17 @@
 class SendPatientOtpSmsJob < ApplicationJob
   queue_as :default
-  self.queue_adapter = :sidekiq
```
harimohanraj89, line 3: Have we configured Sidekiq as the default adapter for `ApplicationJob`?

.env.development
timcheadle: You're right :)

app/jobs/send_patient_otp_sms_job.rb
timcheadle: Good catch; fixed.

Our jobs are pretty inconsistent and I wanted to separate that cleanup in a different PR.

app/services/appointment_notification_service.rb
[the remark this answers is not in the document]
rsanheim: Hmm, the problem is this will always be hardcoded to whatsapp for the `communication_type`, right? So this piece specifically needs to take into account the feature flag as well.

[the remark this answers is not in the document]
rsanheim: I think the communication type needs to be the piece that checks the feature flag and switches appropriately...then the Worker uses the communication_type only to decide how to send reminders. In other words, this service should be the driver which looks at FeatureFlags, and the worker just consumes that data and does as its told.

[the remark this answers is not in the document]
harimohanraj89: Yes, makes sense! 🙌

app/services/notification_service.rb
```diff
@@ -0,0 +1,53 @@
+class NotificationService
+  DEFAULT_LOCALE = :en
+
+  def client
+    @client ||= Twilio::REST::Client.new(twilio_account_sid, twilio_auth_token)
+  end
+
+  def send_sms(recipient_number, message, callback_url=nil)
```
kitallis, line 8: This seems incorrect. The OTP SMSes (from `SendPatientOtpSmsJob#send_sms`) use a different twilio account than the reminder messages since that's the only way to segregate the short-codes (`BPCALL` etc.). The original `SmsNotification` class would optionally take in the client itself so you could segregate accounts cleanly (and/or pass in a different client which is not Twilio-related)

We should also do that here in some fashion, a dumb way to do it would be to create two clients. But I prefer passing in the Client because it makes the `NotificationService` mechanism-agnostic (a bit).

kitallis: Another way to do this could be to create a `MessagingClient` abstraction that returns the client that you need without having to worry about `sid`, `auth_token` details.

app/services/appointment_notification_service.rb
```diff
@@ -7,7 +7,12 @@ def initialize(appointments:, days_overdue: 3, schedule_at:)
...
-    @communication_type = Communication.communication_types[:missed_visit_sms_reminder]
+
+    if FeatureToggle.enabled?("WHATSAPP_APPOINTMENT_REMINDERS")
+      @communication_type = Communication.communication_types[:missed_visit_whatsapp_reminder]
+    else
+      @communication_type = Communication.communication_types[:missed_visit_sms_reminder]
+    end
   end
 
   def send_after_missed_visit
```
kitallis, line 18: We could extract out `overdue_with_patient_consent` in its own scope on `Appointment`.

kitallis: I asked @prabhanshuguptagit to do it in https://github.com/simpledotorg/simple-server/pull/993 instead (might be a simpler change there)

kitallis:
I think generally with the addition of WhatsApp, refactoring would involve these primary things:

* Remove/reduce dependence on the naked `communication_type` string and use `Communication` objects more freely to "decide" stuff.
* Allow notifications to be client agnostic (this is a spectrum).

app/services/appointment_notification_service.rb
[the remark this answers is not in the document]
harimohanraj89: That sounds nifty. I'd like to defer that coupling until _after_ we put the whatsapp -> sms fallback mechanism in place. Once that code's up, it may change our mind on whether coupling comms type to country is a good idea or not. What do you think @kitallis?

app/services/notification_service.rb
harimohanraj89: Ah interesting. So, the notification service can't limit its awareness to just recipient number and message content, because different messages need to come from different senders? Is the sender number sufficient info, or is there a completely different set of credentials for each sender?

kitallis: Yeah, as far as I know. A single `TwilioClient` holds 1 account at a time which involves two things: the auth token and the account id, we have 2 different accounts (currently).

app/services/appointment_notification_service.rb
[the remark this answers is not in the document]
kitallis: Yeah, I think it's ok for now. However, even if we don't use country-specific configs right now, keep in mind that all our feature toggles are _already_ implicitly country-specific.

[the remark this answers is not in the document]
harimohanraj89: With the caveat that they can be _more_ flexible than per-country. We can easily trial Whatsapp in Demo before launching in Prod this way :smile:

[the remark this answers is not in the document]
kitallis: We can do that with country configs too. But I understand your point, tying it to countries is a very specific _product-ish_ decision we'd be making.

My larger point is that, this naked `communication_type` string/sym is making decisions now, so we should think about encapsulating it, related point: https://github.com/simpledotorg/simple-server/pull/971#discussion_r428531475. This is a difficult refactor though, so we can talk about it separately.

app/services/notification_service.rb
timcheadle: We could just optionally pass these credentials into the `NotificationService` initializer. I really, really don't want to put more refactoring into this PR if we can avoid it.

timcheadle:
Regarding the separate Twilio credentials, Indian routes for SMS as of 6-Apr-2020 now prevent us from sending alphanumeric shortcodes, so our `BPCARD` shortcode no longer works. This means we don't need to worry about separate credentials or clients.

See https://support.twilio.com/hc/en-us/articles/223134167-Limitations-sending-SMS-messages-to-Indian-mobile-devices

app/jobs/appointment_notification/worker.rb
```diff
@@ -4,52 +4,63 @@ class AppointmentNotification::Worker
...
+
+    return if appointment.previously_communicated_via?(communication_type)
+
+    patient_phone_number = appointment.patient.latest_mobile_number
+    message = appointment_message(appointment, communication_type, locale)
+
+    begin
+      notification_service = NotificationService.new
+
+      if communication_type == "missed_visit_whatsapp_reminder"
```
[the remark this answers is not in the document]
rsanheim: Agree we should refactor this in follow up PR(s)...Tim and I had discussed that on some pair sessions to get this over the line.

PR #973, opened 2020-05-15
Mark dead patients in line list download

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/story/show/172817001

## This addresses

- Updates the `patient_summary` view to include patient status
- Adds the column to the exporter


db/views/patient_summaries_v03.sql
```diff
@@ -0,0 +1,114 @@
...
+CONCAT(date_part('year', p.recorded_at), ' Q', EXTRACT(QUARTER FROM p.recorded_at)) AS registration_quarter,
+p.full_name,
+(
+    CASE
+      WHEN p.date_of_birth IS NOT NULL THEN date_part('year', age(p.date_of_birth))
+      ELSE p.age + date_part('years', age(NOW(), p.age_updated_at))
+    END
+) AS current_age,
+p.gender,
+p.status,
```
prabhanshuguptagit, line 12: This is the only change in this view.

spec/models/patient_summary_spec.rb
```diff
@@ -58,6 +58,10 @@
         expect(patient_summary.district).to eq(patient.address.district)
         expect(patient_summary.state).to eq(patient.address.state)
       end
+
+      it "includes patient status" do
+        expect(patient_summary.status).to eq(patient.status)
+      end
```
harimohanraj89, line 64: Non-blocking comment: Hmm, what do you think of lumping this together with name and gender, as `it "includes patient attributes"`?

PR #975, opened 2020-05-15
Lockout a user after five failed attempts at logging in

rsanheim:
**Story card:** https://www.pivotaltracker.com/story/show/172823699

## Because

We want Simple.org to be secure

## This addresses

Brute force attacks - this will lock out a user for 20 minutes if they have five failed login attempts.

## Checklist

* [x] add any AR level constraints around new fields
* [x] extract `unlock` method
* [x] extract the math logic to its own method.
* [x] extract constants to ENV vars for lockout time and max failed attempts
* [x] get design to offer suggestions for the error message(s)


Guardfile
```diff
@@ -24,7 +24,7 @@
 #  * zeus: 'zeus rspec' (requires the server to be started separately)
 #  * 'just' rspec: 'rspec'
 
-guard :rspec, cmd: "bundle exec rspec" do
+guard :rspec, cmd: "bin/rspec" do
```
harimohanraj89, line 27: Leveraging Spring?

app/models/phone_number_authentication/authenticate.rb
```diff
@@ -40,19 +44,42 @@ def verify_auth
       case
       when authentication.nil?
         failure('login.error_messages.unknown_user')
+      when authentication.locked_at
+        if authentication.locked_at >= LOCKOUT_TIME.ago
+          minutes_left = (LOCKOUT_TIME - (Time.current - authentication.locked_at)) / 1.minute
+          minutes_left = minutes_left.round
```
harimohanraj89, line 50: Shall we put this math into its own method?

app/models/phone_number_authentication/authenticate.rb
```diff
@@ -40,19 +44,42 @@ def verify_auth
...
     end
 
-    def failure(message_key)
-      error_string = I18n.t(message_key)
+    def unlock
+      authentication.update!(locked_at: nil, failed_attempts: 0)
+    end
+
+    def track_failed_attempt
+      authentication.increment!(:failed_attempts)
```
harimohanraj89, line 75: Whoa, TIL `increment`! Looks like `increment!` returns a `true` or `false`. Do we need to check for `false`s, or can we trust this will work?

app/models/phone_number_authentication/authenticate.rb
```diff
@@ -1,5 +1,8 @@
 class PhoneNumberAuthentication
   class Authenticate
+    MAX_FAILED_ATTEMPTS = 5
+    LOCKOUT_TIME = 20.minutes
```
harimohanraj89, line 4: Can we make these constants configurable through environment variables?

spec/models/phone_number_authentication/authenticate_spec.rb
```diff
@@ -67,4 +67,86 @@
...
+        Timecop.travel(2.minutes.from_now) do
+          result = PhoneNumberAuthentication::Authenticate.call(otp: user.otp, password: "5489", phone_number: user.phone_number)
+          expect(result.success).to be false
+          expect(result.error_message).to eq("Your account is locked out for 18 minutes. Please wait and try again later.")
+        end
+        Timecop.travel(21.minutes.from_now) do
+          phone_number_authentication.set_otp # handled by the mobile app I think?
+          phone_number_authentication.save
+          result = PhoneNumberAuthentication::Authenticate.call(otp: user.otp, password: "5489", phone_number: user.phone_number)
+          expect(result.success).to be true
```
harimohanraj89, line 145: What do you think of `expect(result).to be_success`?

app/models/phone_number_authentication/authenticate.rb
```diff
@@ -40,19 +44,42 @@ def verify_auth
...
         failure('login.error_messages.invalid_password')
       else
         success
       end
     end
 
-    def failure(message_key)
-      error_string = I18n.t(message_key)
+    def unlock
+      authentication.update!(locked_at: nil, failed_attempts: 0)
```
harimohanraj89, line 71: Could we add an `#unlock` method to the PhoneNumberAuthentication model? I figure this would be a good way for us to cleanly unlock an account if we need to do so manually.

config/locales/api/en.yml
```diff
@@ -11,6 +11,7 @@ en:
       invalid_otp: 'Your OTP does not match. Try again?'
       expired_otp: 'You need a fresh OTP. Request a new one.'
       invalid_password: 'Your password does not match. Try again?'
+      account_locked: "Your account is locked out for %{minutes} minutes. Please wait and try again later."
```
harimohanraj89, line 14: Have we run the messaging by design 👀 ?

db/migrate/20200514193029_add_lockouts_to_phone_number_authentication.rb
```diff
@@ -0,0 +1,6 @@
+class AddLockoutsToPhoneNumberAuthentication < ActiveRecord::Migration[5.2]
+  def change
+    add_column :phone_number_authentications, :failed_attempts, :integer, default: 0, null: false
```
harimohanraj89, line 3: Even though this is for internal use, should we back these constraints up with model validations? Fragmented constraints have bitten us before (eg. this is why our data generation is done through the API, as you pointed out).

spec/models/phone_number_authentication/authenticate_spec.rb
rsanheim: i dig it

.env.development
```diff
@@ -92,6 +92,8 @@ DEFAULT_NUMBER_OF_RECORDS=10
 EMAIL_SUBJECT_PREFIX="[Simple DEVELOPMENT]"
 RAILS_MAX_THREADS=20
 SECRET_KEY_BASE=b90c991f1987f5be558af41e1fdc73741adbef2285f30a5bd500dcd7804b2ef1b49e8b07e03bd7154dcecd5050102f0f59cd182c8b40e48a66d416228dbd33fe
+USER_AUTH_LOCKOUT_IN_MINUTES=20
+USER_AUTH_MAX_FAILED_ATTEMPTS=5
```
rsanheim, line 96: Normally I would want to define these in a single `.env` file and be done with it for all environments. I don't think there is a situation where these will vary, at least to start.

I notice we have our `.env` gitignored for this repo, though...is there a way to define these in one file that will apply everywhere?

app/models/phone_number_authentication/authenticate.rb
harimohanraj89: Yup. The overall process for effecting config changes is:
1. Make the updates in the deployment repo and open a PR - [docs](https://github.com/simpledotorg/deployment/tree/master/ansible#changing-secrets)
1. Merge the PR after review (this autudeploys config changes to [five environments](https://semaphoreci.com/resolvetosavelives/deployment)
1. Manually propagate the config changes to other environments - [docs](https://github.com/simpledotorg/deployment/tree/master/ansible#updating-configs)

For steps 1 and 3, You'll need to navigate to the `ansible/` directory of the deployment repo and run the stated commands. The magic "ansible password file" is stored in 1Password. @timcheadle should be able to hook you up with 1Password access.. Look for the "Ansible vault password" entry once you're in. You can stick that anywhere on your computer and use it. Probably not the _best_ way to do that, but it works for now 🤷

.env.development
rsanheim: @harimohanraj89 or anyone - have any thoughts on this?

config/locales/api/en.yml
dburka: Thanks for asking. Am suggesting a slightly tighter language. Also, can we update the expired OTP one too?

Your OTP is expired. Request a new one. 

Your account has been locked for the next X minutes. Please wait and try again.

rsanheim: Thanks, updated!

rsanheim:
I think all the feedback has been resolved -- this is ready for a final round of review.

.env.development
harimohanraj89: Whoops, missed this during my last review, apologies. We don't have a global config file that's shared across environments. My go-to in this case would be to let these config vars be optional, and provide the "20" and "5" as defaults in the application layer. For example

```
class SomeModel
  LOCKOUT_PERIOD = Integer(ENV['USER_AUTH_LOCKOUT_IN_MINUTES'] || 20)
end
```

app/models/phone_number_authentication.rb
```diff
@@ -32,6 +36,30 @@ def mark_as_logged_in
...
+    increment!(:failed_attempts)
+    if failed_attempts >= USER_AUTH_MAX_FAILED_ATTEMPTS
+      update!(locked_at: Time.current)
+    end
+  end
+
+  def minutes_left_on_lockout
+    minutes_left = (lockout_time - (Time.current - locked_at)) / 1.minute
+    minutes_left = minutes_left.round
+  end
```
harimohanraj89, line 61: Now that we've moved several pieces of this to the model, can we add the necessary specs to the corresponding model spec, so we know they work in isolation? The `phone_number_authentication/authenticate_spec` could either remain unchanged, or stub its interactions with the `authentication` object. I'll let ya take a judgment call on that.

config/initializers/config.rb
```diff
@@ -24,6 +24,8 @@
   'RAILS_CACHE_REDIS_PASSWORD',
   'SIDEKIQ_CONCURRENCY',
   'ANALYTICS_DASHBOARD_CACHE_TTL',
+  "USER_AUTH_LOCKOUT_IN_MINUTES",
+  "USER_AUTH_MAX_FAILED_ATTEMPTS"
```
harimohanraj89, line 28: If you go with the ["optional configuration" approach](https://github.com/simpledotorg/simple-server/pull/975/files#r428483384), we can remove these two from here.

.env.development
harimohanraj89: PS If you go with the ["optional configuration" approach](https://github.com/simpledotorg/simple-server/pull/975/files#r428483384), we can remove these two from here.

.env.test
```diff
@@ -85,6 +85,8 @@ DEFAULT_NUMBER_OF_RECORDS=100
 EMAIL_SUBJECT_PREFIX="[Simple TEST]"
 RAILS_MAX_THREADS=20
 SECRET_KEY_BASE=b90c991f1987f5be558af41e1fdc73741adbef2285f30a5bd500dcd7804b2ef1b49e8b07e03bd7154dcecd5050102f0f59cd182c8b40e48a66d416228dbd33fe
+USER_AUTH_LOCKOUT_IN_MINUTES=20
+USER_AUTH_MAX_FAILED_ATTEMPTS=5
```
harimohanraj89, line 89: If you go with the ["optional configuration" approach](https://github.com/simpledotorg/simple-server/pull/975/files#r428483384), we can remove these two from here.

.env.development
rsanheim: Got it, I'll definitely go with that. Easier to just set these in one spot for now.

app/models/phone_number_authentication.rb
rsanheim: 👍 added specs

rsanheim:
This is all approved, but I want to wait on being able to run through this flow locally, end-to-end, before merging. Hopefully @kitallis and I can do that tomorrow.

PR #976, opened 2020-05-15
Add spring; cuts about ~3 seconds off every local test run

rsanheim:
Before **5.3 seconds for a single spec run**

```
[~/src/simpledotorg/simple-server (add-spring)]> time bin/rspec spec/models/phone_number_authentication_spec.rb
Running via Spring preloader in process 98807
Run options: include {:focus=>true}

All examples were filtered out; ignoring {:focus=>true}
[rspec-sidekiq] WARNING! Sidekiq will *NOT* process jobs in this environment. See https://github.com/philostler/rspec-sidekiq/wiki/FAQ-&-Troubleshooting
.........

Finished in 1.08 seconds (files took 0.51137 seconds to load)
9 examples, 0 failures

real	0m5.307s
user	0m0.149s
sys	0m0.058s
```

After **2.1 seconds for the same spec run**

```
[~/src/simpledotorg/simple-server (add-spring)]> time bin/rspec spec/models/phone_number_authentication_spec.rb
Running via Spring preloader in process 98747
Run options: include {:focus=>true}

All examples were filtered out; ignoring {:focus=>true}
[rspec-sidekiq] WARNING! Sidekiq will *NOT* process jobs in this environment. See https://github.com/philostler/rspec-sidekiq/wiki/FAQ-&-Troubleshooting
.........

Finished in 1.06 seconds (files took 0.55638 seconds to load)
9 examples, 0 failures

real	0m2.137s
user	0m0.148s
sys	0m0.055s
```

**Story card:** n/a - dev speed improvement

## Because

We want faster tests locally

## This addresses

adding spring

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -76,6 +76,8 @@ group :development do
   gem 'guard-rspec', require: false
   gem 'listen', '>= 3.0.5', '< 3.2'
   gem 'rails-erd'
+  gem "spring"
+  gem "spring-commands-rspec"
   gem 'web-console', '>= 3.3.0'
 end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -428,6 +428,9 @@ GEM
       docile (~> 1.1)
       simplecov-html (~> 0.11)
     simplecov-html (0.12.2)
+    spring (2.1.0)
+    spring-commands-rspec (1.0.4)
+      spring (>= 0.9.1)
     sprockets (4.0.0)
       concurrent-ruby (~> 1.0)
       rack (> 1, < 3)
@@ -548,6 +551,8 @@ DEPENDENCIES
   sidekiq
   sidekiq-throttled
   simplecov
+  spring
+  spring-commands-rspec
   timecop (~> 0.9.0)
   twilio-ruby (~> 5.10, >= 5.10.3)
   tzinfo-data
--- a/bin/rake
+++ b/bin/rake
@@ -1,4 +1,9 @@
 #!/usr/bin/env ruby
+begin
+  load File.expand_path('../spring', __FILE__)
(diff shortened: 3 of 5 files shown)


harimohanraj89:
I thought we used `bootsnap` instead? I guess `bootsnap` caches some loadtime info, and `spring` keeps a dormant server running behind the scenes?

rsanheim:
> I thought we used `bootsnap` instead? I guess `bootsnap` caches some loadtime info, and `spring` keeps a dormant server running behind the scenes?

Yeah they are complementary. Spring keeps the server running in the background for very fast test speed. Spring _can_ cause some confusing issues where your changes won't be picked up for things like initializers.  For that sort of case. you just need to do:

```
spring stop
```

from the command line to force a full reload.

rsanheim:
Oh, and you can always bypass spring by using `bundle exec rspec`...only the binstubs will use spring.

harimohanraj89:
Aah nice. So only `bin/rails server` would use Spring, while `bundle exec rails server` would not. This is very nifty.

Hehe, I recall some negative emotions wrt Spring, and your comment about initializers is exactly what I'm thinking about.

Since we know exactly how to fix those issues, and the Spring setup is very opt-in, I think this is very cool 👍

rsanheim:
> Hehe, I recall some negative emotions wrt Spring, and your comment about initializers is exactly what I'm thinking about.
> 
> Since we know exactly how to fix those issues, and the Spring setup is very opt-in, I think this is very cool 👍

Yeah it can be a bit surprising if you forget that spring is being used for `bin/rspec`. I find the time speed up is worth it when used in conjunction with `guard`, but its worth knowing the pitfalls.

PR #980, opened 2020-05-18
Prevent more deprecations (hopefully)

rsanheim:
**Story card:** n/a ; some minor cleanup

## Because

We want a useful, clean build and log output

## This addresses

Removing one deprecation and preventing new ones from entering the log.

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -78,8 +78,6 @@ class Patient < ApplicationRecord
       .distinct
   end
 
-  private_class_method :follow_ups_with
-
   enum could_not_contact_reasons: {
     not_responding: 'not_responding',
     moved: 'moved',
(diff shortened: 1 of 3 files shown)


app/models/patient.rb
```diff
@@ -78,8 +78,6 @@ def self.follow_ups_with(model_name, period, time_column: 'recorded_at', last: n
       .distinct
   end
 
-  private_class_method :follow_ups_with
```
rsanheim, line 81: @kitallis Not sure the risk of making this public? I think its worth exposing it for now to silence the deprecation warning though.

kitallis: Nope, it's perfectly ok for this to be public, it's just private for clean-public-interface's sake.

PR #983, opened 2020-05-19
Add standardrb

kitallis:
**Story card:** TBD

## Because

We want a a consistent, robust way of linting our code.

## This addresses

* Adds [standard](https://github.com/testdouble/standard) rb gem
* Fixes a few non-critical files
* Adds a `.standard_todo.yml` to keep a track of offenses that we can slowly correct over time

## Note

We don't need to fix all files in this PR right away. I've added a _few_ to just get a sense of what the changes might look like – these won't be comprehensive though.

Along with this PR, I'd be adding this [Action](https://github.com/simpledotorg/simple-server/pull/983) as well.

## Todo

- [x] Add RubyMine instructions to README


.standard_todo.yml
```diff
@@ -0,0 +1,500 @@
...
+- spec/transformers/api/v3/appointment_transformer_spec.rb
+- spec/transformers/api/v3/blood_sugar_transformer_spec.rb
+- spec/transformers/api/v3/medical_history_transformer_spec.rb
+- spec/transformers/api/v3/patient_business_identifier_transformer_spec.rb
+- spec/transformers/api/v3/patient_phone_number_transformer_spec.rb
+- spec/transformers/api/v3/patient_transformer_spec.rb
+- spec/transformers/api/v3/prescription_drug_transformer_spec.rb
+- spec/transformers/api/v3/user_transformer_spec.rb
+- spec/transformers/api/v4/blood_sugar_transformer_spec.rb
+- spec/transformers/api/v4/user_transformer_spec.rb
```
harimohanraj89, line 500: HAHA _adds standardrb, adds entire repo to ignorelist_

JK, this is a great way to chip away at refactors.

kitallis:
> All the corrections are quotes and space, nice.
> 
> How do we hook this into our development flow?

Ha, good question, I added a couple of TODOs to the PR so we can start using it in dev.

harimohanraj89:
Neato. 🚀

rsanheim:
> Fixes a few non-critical files

Suggestion: list out what standardrb commands you are running to fix what files, so that future PRs can then target other sub directories. For example, we could have a PR for all controllers and specs, then all models and their specs, and so-on. If we wanted to be extra safe that is. I find the autofixes are _usually_ pretty safe.

rsanheim:
> Suggestion: list out what standardrb commands you are running to fix what files

Hmm, now I realize we can kinda use the todo file here to address that. Though will the GH action tell us when a file in the `todo` list is in compliance and can be removed? Or is that up to the author?

kitallis:
> > Suggestion: list out what standardrb commands you are running to fix what files
> 
> Hmm, now I realize we can kinda use the todo file here to address that. Though will the GH action tell us when a file in the `todo` list is in compliance and can be removed? Or is that up to the author?

The GH Action will not tell us that, it's up to the author to remove stuff after the file is compliant. I'll document the commands in the README so they are visible outside the PR 👍

PR #984, opened 2020-05-19
Fix overview query spec

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/story/show/172909765

## Because

A flaky spec was added in #940 which uncovered an underlying issue in the way we use time zones in the my facilities query specs.

## This addresses

Ideally we should be running all our specs in UTC. Currently the My facilities query specs do a view refresh in `Asia/Kolkata` and run expectations in UTC, which is incorrect. This was causing a `daily` view spec for example to fail from midnight to 5:30 AM IST. This removes the time zoning around view refreshes so everything runs in UTC.

This fixes the flaky test marked pending in #982.

### DIFF
--- a/spec/queries/my_facilities/blood_pressure_control_query_spec.rb
+++ b/spec/queries/my_facilities/blood_pressure_control_query_spec.rb
@@ -99,7 +99,6 @@ RSpec.describe MyFacilities::BloodPressureControlQuery do
 
       before do
         ActiveRecord::Base.transaction do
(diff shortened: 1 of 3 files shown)


prabhanshuguptagit:
> We, however, are no longer testing that the dashboards work correctly if we use a different timezone, right? Could you take a peek to see if that's true, and craft any necessary follow-up stories for us to fill in those holes?

I think we _could_ add some controller tests to check for this. However the controller mostly does mock tests and doing data level tests would be slightly duplicative. Not sure what the best place to test this behaviour would be, added a story anyways: https://www.pivotaltracker.com/story/show/172911921

PR #985, opened 2020-05-19
Create mobile app builds for review aps

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172912962

This PR adds a custom Github Action to create an APK that points to the corresponding Heroku review app. Check out the "Review App Build" action below on this PR. It contains a downloadable Artifact that contains the APK.

These APKs will not be useful until we teach review apps to serve meaningful manifest files (#986).


.github/workflows/main.yml
```diff
@@ -0,0 +1,27 @@
+name: Review App Build
```
kitallis, line 1: Shall we call this workflow file `review-mobile-app.yml` or something?

harimohanraj89: Addressed in 34b8308

PR #986, opened 2020-05-19
Dynamic manifests

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172912962

## Because

We need to teach review apps to serve meaningful manifests, because #985 will start creating working app builds to test against review apps.

## This addresses

Serving dynamic manifests in the `development` and `review` environments. The only dynamic bit is the API endpoint listed against each country in the manifest.

The inclusion of the `development` env in this PR is to facilitate local server + emulator setup. The relevant details have been added to the README as well.


app/controllers/api/manifests_controller.rb
```diff
@@ -1,8 +1,43 @@
...
+      manifest_file = "public/manifest/#{ENV['SIMPLE_SERVER_ENV']}.json"
+      return head :not_found unless File.exists?(manifest_file)
+
+      render json: File.read(manifest_file)
+    end
+  end
+
+  private
+
+  def dynamic_manifest
```
kitallis, line 15: This fragments the source of truth for the manifest. Can we instead just move the manifest to an ERB template (like `manifest.json.erb`) and always populate the `api_endpoint` from `ENV['SIMPLE_SERVER_HOST_PROTOCOL']` and `ENV['SIMPLE_SERVER_HOST']` and make them required config keys?

We can source the rest of the data in the manifest from `Rails.application.config.country`.

README.md
```diff
@@ -43,17 +43,28 @@ rails db:setup
...
 ```
 
-The output of the ngrok command will have an url that can be used to access
-local-server. This url should be set as `qaApiEndpoint` in gradle.properties.
+The output of the ngrok command is HTTP and HTTPS URLs that can be used to access your local server. Configure the
+following places with the HTTPS URL.
+
+In the `gradle.properties` file in the `simple-android` repository,
+```
+qaApiEndpoint=<HTTPS URL>
```
rsanheim, line 60: This property does not currently exist in the [properties](https://github.com/simpledotorg/simple-android/blob/master/gradle.properties) file in simple-android. There are some other properties that kind of look like it, though 😄 .  Is this the right property and developers just need to add it? Do devs need to change `qaManifestEndpoint` or `qaFallbackApiEndpoint` ?

README.md
```diff
@@ -43,17 +43,28 @@ rails db:setup
...
 
 ```bash
 brew cask install ngrok
 rails server
 ngrok http 3000
 ```
 
-The output of the ngrok command will have an url that can be used to access
-local-server. This url should be set as `qaApiEndpoint` in gradle.properties.
+The output of the ngrok command is HTTP and HTTPS URLs that can be used to access your local server. Configure the
```
rsanheim, line 55: Might be good to mention that the HTTP URL should be ignored, as it won't work due to security policies ( I figured this out the hard way 😏 ).

README.md
kitallis: I think the right property is indeed `qaManifestEndpoint` (it used to be called `qaApiEndpoint`) 👍 

For testing purposes, we shouldn't need to modify `qaFallbackApiEndpoint`, so we can leave that be.

app/controllers/api/manifests_controller.rb
harimohanraj89: We unfortunately can't because each of the "real" manifests (prod, demo) need to know of _every_ country's endpoints. I think these static files for them make sense. In the case of dev and review (and even SBX, I guess), all country endpoints point to the same box, but the endpoint of the box itself is context-dependent.

That being said, you're right about `erb` extraction. I'll do that.

harimohanraj89: Extracted to a jbuilder in e19e10a

spec/controllers/api/manifests_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      environments.each do |env|
+        it "returns a dynamic manifest for #{env}" do
+          original_env = ENV["SIMPLE_SERVER_ENV"]
+          original_host = ENV["SIMPLE_SERVER_HOST"]
+          original_protocol = ENV["SIMPLE_SERVER_HOST_PROTOCOL"]
+
+          ENV["SIMPLE_SERVER_ENV"] = env
+          ENV["SIMPLE_SERVER_HOST"] = "simple.example.com"
+          ENV["SIMPLE_SERVER_HOST_PROTOCOL"] = "https"
```
prabhanshuguptagit, line 52: Would it be simpler to stub `ENV` for these specs? Something like [this thing](https://github.com/simpledotorg/simple-server/blob/bce5cbd14cd13d562ff4974e668c309eb47185d6/spec/jobs/appointment_notification/missed_visit_job_spec.rb#L22).

app/controllers/api/manifests_controller.rb
kitallis: Had no idea about `jbuilder`, pretty neat. I'm still a little concerned about the split, but I understand the limitation.

Do you think we can write a simple regression test that ensures that review-env keys in the manifest match the real-env keys?

harimohanraj89: @kitallis Yeah it kinda sucks that they're conceptually different now.

By regression test, do you mean like this one here? https://github.com/simpledotorg/simple-server/pull/986/files#diff-858391f632a627cb72d50e9884e1d8e1R41

Or did you have something else in mind?

spec/controllers/api/manifests_controller_spec.rb
harimohanraj89: Ah, it is fewer lines of code, and guaranteed to be contained to the scope of the spec. I'll rework it.

PR #987, opened 2020-05-19
Add fields to facility bulk upload

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/story/show/171114696

## Because

The facility bulk upload template needs to support some newly added fields.

## This addresses

Adds the following fields to bulk upload:
- `facility_size`
- `enable_teleconsultation`
- `teleconsultation_phone_number`
- `teleconsultation_isd_code`
- `enable_diabetes_management`

**Supplementary change:**

This adds a `LiberalEnum` type to allow for model validations on enums: [ref](https://github.com/simpledotorg/simple-server/pull/987#discussion_r427321639)

Rails throws an `ArgumentError` when a value outside an `enum` is provided to an object. This prevents us from doing rails validations on `enums`. The `LiberalEnum` patch bypasses the exception so we can do validations like other normal attributes.


app/controllers/admin/facilities_controller.rb
```diff
@@ -119,13 +119,21 @@ def validate_duplicate_rows
...
     @facilities.each do |facility|
-      import_facility = Facility.new(facility)
-      @row_errors << [row_num, import_facility.errors.full_messages.to_sentence] if
-          import_facility.invalid?
+      if valid_facility_size?(facility)
+        import_facility = Facility.new(facility)
+        @row_errors << [row_num, import_facility.errors.full_messages.to_sentence] if import_facility.invalid?
+      else
+        @errors << "Facility size on row #{row_num} is invalid"
+      end
```
prabhanshuguptagit, line 136: Is there a better way to validate `facility_size`? Looks like rails does not support `enum` inclusion validations and just throws an `ArgumentError` when `Facility.new` is created. From https://github.com/rails/rails/issues/13971 it seems like doing a separate check in the controller is the most reasonable

rsanheim: Yeah I've noticed our usage of enum fields doesn't work well w/ our primarily API driven workflows. 

Want to give this [LiberalEnum](https://github.com/rails/rails/issues/13971#issuecomment-542757073) patch a try ? That would be useful for several of our enum types we want to be able to validate.

app/models/concerns/liberal_enum.rb
```diff
@@ -0,0 +1,11 @@
+module LiberalEnum
+  extend ActiveSupport::Concern
```
prabhanshuguptagit, line 2: ref: https://github.com/simpledotorg/simple-server/pull/987#discussion_r427321639

public/documents/upload_facilities.csv
```diff
@@ -1 +1 @@
-organization,facility_group,facility_name,facility_type,street_address (optional),village_or_colony (optional),zone_or_block (optional),district,state,country,pin (optional),latitude (optional),longitude (optional)
+organization,facility_group,facility_name,facility_type,street_address (optional),village_or_colony (optional),zone_or_block (optional),district,state,country,pin (optional),latitude (optional),longitude (optional),size (optional),enable_diabetes_management (true/false),enable_teleconsultation (true/false),teleconsultation_phone_number,teleconsultation_isd_code
```
harimohanraj89, line 1: Reckon we should confirm the ordering of these new fields with Deepa?

harimohanraj89: Not sure where to leave this comment 😅 Why don't we default the "enable" columns to false if blank? That way, as a user, I can stick "true" in the 5 facilities I want DM in, without having to fill in "false" in all the other rows.

app/controllers/admin/facilities_controller.rb
```diff
@@ -108,46 +107,11 @@ def parse_file
...
-      import_facility = Facility.new(facility)
-      @row_errors << [row_num, import_facility.errors.full_messages.to_sentence] if
-          import_facility.invalid?
-      row_num += 1
-    end
-    if @row_errors.present?
-      group_row_errors.each { |error| @errors << error }
-    end
+  def validate_facility_rows
+    @errors = Admin::FacilityCSVValidator.validate(@facilities).errors
```
harimohanraj89, line 111: Curious, do we reject the entire CSV if any errors are found?

prabhanshuguptagit: Yes and we show the user something like:

image

public/documents/upload_facilities.csv
prabhanshuguptagit: I thought about this but this will make the import slightly more verbose. These fields will need to be dealt separately before the import since our model doesn't allow nils. If setting all cells is something we can live with, we should stick to that imho and keep it simple.

harimohanraj89: We could provide defaulting quite elegantly by inserting ourselves [here](https://github.com/simpledotorg/simple-server/blob/master/app/jobs/import_facilities_job.rb#L12) with something like:

```
import_facility = Facility.new(facility_params)

def facility_params
  facility.merge(
    enable_diabetes: facility[:enable_diabetes] || false
end
```

or even at the CSV parsing level [here](https://github.com/simpledotorg/simple-server/blob/master/app/models/facility.rb#L98-L112).

Speaking of CSV parsing, I think we need to do some work in this method to ensure we're parsing booleans correctly.

app/models/facility.rb
```diff
@@ -109,6 +116,11 @@ def self.parse_facilities(file_contents)
                    pin: row['pin (optional)'],
                    latitude: row['latitude (optional)'],
                    longitude: row['longitude (optional)'],
+                   facility_size: row['size (optional)'],
+                   enable_diabetes_management: row['enable_diabetes_management (true/false)'],
```
harimohanraj89, line 120: Will this be the string `"true"/"false"` or the boolean? Do we need to coerce to boolean here? Otherwise, `"false"` is a truthy value. 

As mentioned in my earlier comment, this is a great place to take care of defaulting behavior. Something like:

```ruby
enable_diabetes_management: (row['enable_diabetes_management'] == "true")
# or
enable_diabetes_management: row['enable_diabetes_management'].downcase.in?('t', 'true')
```

spec/fixtures/files/upload_facilities_test.csv
```diff
@@ -1,3 +1,3 @@
-organization,facility_group,facility_name,facility_type,street_address (optional),village_or_colony (optional),zone_or_block (optional),district,state,country,pin (optional),longitude (optional),latitude (optional)
-OrgOne,FGTwo,Test Facility,CHC,,,,Bhatinda,Punjab,India,,,
-OrgOne,FGTwo,Test Facility 2,CHC,,,,Bhatinda,Punjab,India,,,
\ No newline at end of file
+organization,facility_group,facility_name,facility_type,street_address (optional),village_or_colony (optional),zone_or_block (optional),district,state,country,pin (optional),latitude (optional),longitude (optional),size (optional),enable_diabetes_management (true/false),enable_teleconsultation (true/false),teleconsultation_phone_number,teleconsultation_isd_code
+OrgOne,FGTwo,Test Facility,CHC,,,,Bhatinda,Punjab,India,,,,,false,false,,
+OrgOne,FGTwo,Test Facility 2,CHC,,,,Bhatinda,Punjab,India,,,,,false,true,9999999999,91
```
harimohanraj89, line 3: Do these fixture updates automatically add test coverage for the new fields to the facility model's `#parse_facilities` method?

app/validators/admin/facility_csv_validator.rb
```diff
@@ -0,0 +1,48 @@
+class Admin::FacilityCSVValidator
```
kitallis, line 1: Feel like I prefer `Admin::CSV::FacilityValidator` over `FacilityCSVValidator`.

app/validators/admin/facility_csv_validator.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def initialize(facilities)
+    @facilities = facilities
+    @errors = []
+  end
+
+  attr_reader :errors
+
+  def validate
```
kitallis, line 14: Could we move some validation specs off the controller and move them against this class?

app/models/facility.rb
prabhanshuguptagit: These boolean fields are type casted appropriately according to the rule set mentioned [here](https://api.rubyonrails.org/classes/ActiveModel/Type/Boolean.html). So `"false"` would be a `false` as expected. The only defaulting we really need to do is the `nil`, which I've poked deepa about. It's not much to do at all, it's just more stuff happening in the import iykwim.

public/documents/upload_facilities.csv
prabhanshuguptagit: `parse_facilities` is definitely **the** place to put this in. Rest addressed in comment below.

app/validators/admin/facility_csv_validator.rb
prabhanshuguptagit: Addressed in 8bd1abe and 6fbff40.

app/validators/admin/facility_csv_validator.rb
prabhanshuguptagit: Addressed in dc5d307.

public/documents/upload_facilities.csv
prabhanshuguptagit: Addressed in 0f9d452. ([Slack ref](https://simpledotorg.slack.com/archives/CFW5F0GFQ/p1590069415021900?thread_ts=1589901630.002500&cid=CFW5F0GFQ))

app/models/facility.rb
prabhanshuguptagit: Addressed in 0f9d452. (Slack ref)

prabhanshuguptagit:
> I like the next if facility.values.all?(&:blank?) style.

I don't have a preference, so why not 😄

PR #988, opened 2020-05-20
Bangladesh medical history cleanup

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/n/projects/2184102/stories/172931517

## Because

Large swathes of imported patients in BD aren't marked "hypertension=true"

## This addresses

A rake task to execute in BD to correct these records, marking them "hypertension=true". It also marks the deprecated "diagnosed_with_hypertension=true" to be consistent with API behavior.

### DIFF
--- a/lib/tasks/scripts/correct_bangladesh_medical_histories.rb
+++ b/lib/tasks/scripts/correct_bangladesh_medical_histories.rb
@@ -17,7 +17,7 @@ class CorrectBangladeshMedicalHistories
 
     return if dryrun
 
-    log 'Updating dosages...'
+    log 'Updating medical histories...'
     update_medical_histories
 
     log 'Complete. Goodbye.'
--- a/lib/tasks/scripts/correct_bangladesh_medical_histories.rb
+++ b/lib/tasks/scripts/correct_bangladesh_medical_histories.rb
@@ -13,14 +13,19 @@ class CorrectBangladeshMedicalHistories
   end
 
   def call
+    unless Rails.application.config.country[:abbreviation] == "BD"
+      log "Cannot run this cleanup outside of Bangladesh"
+      return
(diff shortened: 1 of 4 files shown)


lib/tasks/scripts/correct_bangladesh_medical_histories.rb
```diff
@@ -0,0 +1,49 @@
...
+  def initialize(dryrun: false, verbose: true)
+    @dryrun = dryrun
+    @verbose = verbose
+  end
+
+  def call
+    unless Rails.application.config.country[:abbreviation] == "BD"
+      log "Cannot run this cleanup outside of Bangladesh"
+      return
+    end
```
harimohanraj89, line 19: This is important for this cleanup script, since it would clobber other environments.

spec/lib/tasks/scripts/correct_bangladesh_medical_histories.rb
```diff
@@ -0,0 +1,71 @@
...
+    context "outside of Bangladesh" do
+      it "does not modify any drugs" do
+        original_country = Rails.application.config.country[:abbreviation]
+        Rails.application.config.country[:abbreviation] = "IN"
+
+        yes = create(:medical_history, :hypertension_yes)
+        no = create(:medical_history, :hypertension_no)
+        unknown = create(:medical_history, :hypertension_unknown)
+
+        CorrectBangladeshMedicalHistories.call(dryrun: true, verbose: false)
```
kitallis, line 42: If we `dryrun` this, it won't actually test that it didn't run the updates?

harimohanraj89: Absolutely right.  🍝

PR #990, opened 2020-05-20
Add searchability to various models

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/167993535

## Because

We want to be able to search for users/admins/facilities on the dashboards.

## This addresses

This PR should only contain search scopes on relevant models, the actual functionality from the UI is in #1000.

- [x] Add `search_by_name_or_phone` on `User`
- [x] Add `search_by_name_or_email` on `User`
- [x] Add `search_by_name` on `Facility`

We've chosen separate scopes on `User` to optimize for the use-case of searching `Users` and `Admins` separately and reducing the search-space in each.

**Note:** In later PRs, I may tweak around the search strategies here and there to see what fits best.

### Searching across associations

It is possible to do a cross-association search with `pg_search`, but I've chosen to put explicit scopes on individual models and then make a composite scope in `User` because: the `associated_against` version makes for a less flexible query and is [impossible to index](https://github.com/Casecommons/pg_search#searching-through-associations).

Adding indexes and searching via the custom composite scope is a lot faster than the `associated_against` version of the same query. The composite scope simply `union`s the result over for effectively an `OR` query (you'd do a `merge` for an `AND` query)

#### Benchmarks

I loaded up my local db with 14k users and ran both the `pg_search` version of the scope and the custom composite scope, results down below:

```
> User.count
=> 14603
```

```
#
# test data / utils
#
$fnames = User.all.map(&:full_name)

$phones = User
            .includes(:phone_number_authentications)
            .all
            .map(&:phone_number)
            .compact

def time(&block)
  puts Benchmark.measure(&block)
end

#
# example scope using associated_against on User
#
 pg_search_scope :search_by_name_or_number,
   against: [:full_name],
   associated_against: {phone_number_authentications: [:phone_number]},
   using: {tsearch: {prefix: true, any_word: true}}

#
# time with pg_search scope
#
time do
  ($fnames + $phones)
    .shuffle
    .sample(5000)
    .map { |t| User
                 .joins(phone_number_authentications: :facility)
                 .search_by_name_or_number(t)
                 .load }
end
# => 12.166408   0.524870  12.691278 (190.312482)

#
# time with composite, indexed scope
#
time do
  ($fnames + $phones)
    .shuffle
    .sample(5000)
    .map { |t| User
                 .joins(phone_number_authentications: :facility)
                 .search_by_name_or_phone(t)
                 .load }
end
# => 14.253197   0.699315  14.952512 ( 47.055815)

```

The composite, union fully-realized query (`search_by_name_or_phone`) is 4x faster than the normal `pg_search` query (`search_by_name_or_number`).

### Indexes

Running similar tests like the above shows that the queries do end up using the indexes. 

Sample [explain](https://explain.depesz.com/s/wYTR).

Here you can see that the gin indexes: `index_gin_users_on_full_name` and `index_gin_phone_number_authentications_on_phone_number` are being used. 

We also see that there's an `INDEX SCAN` on the pkey to filter out `deleted_at`. This has some cost and hence I've also added an index on `deleted_at` to speed up that filtering: [improvements]().

**Note:** The `GIN` index uses the [Simple](https://www.postgresql.org/docs/10/textsearch-dictionaries.html#TEXTSEARCH-SIMPLE-DICTIONARY) dictionary and the two-argument [tsvector](https://www.postgresql.org/docs/9.5/textsearch-tables.html#TEXTSEARCH-TABLES-INDEX) form.

### Union

One more problem with the `associated_against` version of the `pg_search` query is its compositional issues. It breaks in `Admin::UsersController#index` when used along with the other `where` clauses due to naming conflicts. The `union` doesn't have any such problems and composes very well. A working version of this can be seen in [#1000](https://github.com/simpledotorg/simple-server/pull/1000/files#diff-1a862d212011bf5ea20a178c639c9f1aR16). 

### What the table-driven specs read like

Screenshot 2020-05-20 at 5 26 22 PM


spec/models/user_spec.rb
```diff
@@ -63,6 +63,85 @@
...
 
-  xdescribe '.update_with_phone_number_authentication' do
+  describe "Search" do
+    let!(:user_1) { create(:user, full_name: "Sri Priyanka John") }
+    let!(:user_2) { create(:user, full_name: "Priya Sri Gupta") }
+
+    shared_examples "searches whole words against full_names" do |search_method|
+      ["Sri", "sri", "SRi", "sRi", "SRI", "sRI"].each do |term|
+        it "returns results for case-insensitive searches: #{term.inspect}" do
+          expect(User.send(search_method, term)).to match_array([user_1, user_2])
```
rsanheim, line 73: Can you change all these `sends` to `public_send` ? `send` should be avoided unless necessary, as it completely bypasses method visibility.

rsanheim:
Sorry for dupe comments. GitHub is having issues and its hard to tell what is posting correctly.

app/models/facility.rb
```diff
@@ -35,6 +36,8 @@ class Facility < ApplicationRecord
            class_name: "Patient",
            foreign_key: "registration_facility_id"
 
+  pg_search_scope :search_by_name, against: {name: "A", slug: "B"}, using: {tsearch: {prefix: true, any_word: true}}
```
harimohanraj89, line 39: What a curious weighting DSL 😅

spec/models/facility_spec.rb
```diff
@@ -126,4 +126,44 @@
...
   end
+
+  describe "Search" do
+    let(:facility_1) { create(:facility, name: "HWC Bahadurgarh") }
+    let(:facility_2) { create(:facility, name: "CHC Docomo", slug: "chc-docomo") }
+    let(:facility_3) { create(:facility, name: "CHC Porla Docomo") }
+
+    ["CHC", "DoCoMo", "DOCOMO", "docomo", "cHc", "chc"].each do |term|
+      it "returns results for case-insensitive searches: #{term.inspect}" do
+        expect(Facility.search_by_name(term)).to match_array(facility_2)
```
harimohanraj89, line 137: Why do these searches not return facility 3?

kitallis: Great catch, my poor usage for `let` vs `let!` was the culprit.

harimohanraj89: Aaaah gotcha gotcha. Neato, in that case this behaves the way I'd expect 👍

kitallis: I fixed the specs to be clearer and not leak `let` state or be confusing.

db/migrate/20200528113844_add_deleted_at_indexes_to_user_related_models.rb
```diff
@@ -0,0 +1,6 @@
+class AddDeletedAtIndexesToUserRelatedModels < ActiveRecord::Migration[5.2]
+  def change
+    add_index :users, :deleted_at
+    add_index :phone_number_authentications, :deleted_at
```
harimohanraj89, line 4: Ah, I see we missed a few. Do we need to be safe and run this migration concurrently?

db/migrate/20200528113707_add_search_index_to_facility.rb
```diff
@@ -0,0 +1,11 @@
+class AddSearchIndexToFacility < ActiveRecord::Migration[5.2]
+  def up
+    execute "CREATE INDEX index_gin_facilities_on_name ON facilities USING GIN ((to_tsvector('simple'::regconfig, COALESCE((name)::text, ''::text))));"
+    execute "CREATE INDEX index_gin_facilities_on_slug ON facilities USING GIN ((to_tsvector('simple'::regconfig, COALESCE((slug)::text, ''::text))));"
```
harimohanraj89, line 4: These search indexes, do we need to worry about them affecting DB performance when run?

db/migrate/20200528113844_add_deleted_at_indexes_to_user_related_models.rb
kitallis: Made them all concurrent.

db/migrate/20200528113707_add_search_index_to_facility.rb
kitallis: Not yet, the users/facility tables are tiny, we needn't worry about this for now. The `GIN` indexes can get big, but don't have to worry about that now either, the `users` and `facilities` tables are both under 1mb right now on production.

PR #992, opened 2020-05-21
Add an allowlist for parameter logging

rsanheim:
This ensures we only log keys that we explicitly want, and everything
else is filtered. This lets us log just what we want for tracing /
debugging (so things like UUIDs, request times, HTTP info) and all
personal information will be dropped.

Also adding a request spec that is basically the patient sync spec duplicated, but it
also verifies that patient attributes were _not_ logged after the requests.
Getting a working spec on the logging was tricky, and it still needs
some work to properly verify things.

**Story card:** https://www.pivotaltracker.com/story/show/172949096

## Because

We don't want to accidentally log any sensitive data

## This addresses

a stricter parameter log filtering approach


spec/requests/filter_parameter_logging_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  let(:build_payload) { -> { build_patient_payload(FactoryBot.build(:patient, registration_facility: request_user.facility)) } }
+  let(:build_invalid_payload) { -> { build_invalid_patient_payload } }
+  let(:update_payload) { ->(record) { updated_patient_payload record } }
+
+  def to_response(patient)
+    Api::V3::PatientTransformer.to_nested_response(patient)
+  end
+
+  include_examples 'v3 API sync requests'
```
rsanheim, line 17: This brings in a bunch of specs we don't care about, but breaking apart the shared example pieces can wait ...

spec/requests/filter_parameter_logging_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    process_token = JSON(response.body)['process_token']
+
+    created_patients         = Patient.find(first_patients_payload.map { |patient| patient['id'] })
+    updated_patients_payload = created_patients.map do |patient|
+      updated_patient_payload(patient)
+        .except(%w[address phone_numbers business_identifiers].sample)
+    end
+
+    post sync_route, params: { patients: updated_patients_payload }.to_json, headers: headers
+    get sync_route, params: { process_token: process_token }, headers: headers
```
rsanheim, line 36: I'll refactor this spec to simplify things and inline the things we need from the shared code.

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,16 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += [:password, :age]
+
+WHITELISTED_KEYS_MATCHER = /((^|_)ids?|action|controller|code|created_at|updated_at|recorded_at|deleted_at|limit|process_token$)/.freeze
```
kitallis, line 6: Shall we call it an ALLOWLIST or SAFELIST?

rsanheim: Its now in a module and the naming is improved.

spec/requests/filter_parameter_logging_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+  let(:headers) do
+    {"ACCEPT" => "application/json", "CONTENT_TYPE" => "application/json"}.merge(auth_headers)
+  end
+
+  it "does not log sensitive patient data" do
+    stringio = StringIO.new
+    test_logger = Logger.new(stringio)
+    allow(ActionController::Base).to receive(:logger).and_return(test_logger)
+
+    patient_payloads = (1..3).map { build_payload.call }
```
timcheadle, line 22: I've always written this as `3.times.map { build_payload.call }` so it's one less thing to track/break

spec/requests/filter_parameter_logging_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    post sync_route, params: {patients: patient_payloads}.to_json, headers: headers
+    get sync_route, params: {}, headers: headers
+
+    patients = patient_payloads.map { |payload| Patient.find(payload[:id]) }
+    output = stringio.string
+    patients.each do |patient|
+      expect(output).to_not match(/\b#{patient.full_name}\b/)
+      expect(output).to_not match(/\b#{patient.age}\b/) if patient.age
+      expect(output).to_not match(/\b#{patient.date_of_birth}\b/) if patient.date_of_birth
+      if patient.address
```
timcheadle, line 32: We should filter all address fields here, including district and state.

We should also filter phone numbers.

spec/requests/filter_parameter_logging_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+  end
+
+  def check_payload(output, payload)
+    payload.each do |attr, value|
+      next if value.blank?
+
+      regex = nil
+      case value
+      when Hash
```
timcheadle, line 62: TIL that `case` statements use `===`!

```
[12] pry(main)> Hash === {foo: :bar}
=> true
[13] pry(main)> {foo: :bar} === Hash
=> false
```

Weird.

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,32 @@
 # Be sure to restart your server when you modify this file.
+module ParameterFiltering
+  ATTRIBUTE_WHITELIST = %w[
```
timcheadle, line 3: Shall we comment where we expect these attributes to come from? Are they params?

config/initializers/filter_parameter_logging.rb
harimohanraj89: Hey @rsanheim, I think @kitallis was referring to moving away from "whitelist/blacklist" terminology to use "allowlist/denylist" instead. Similar to moving away from "master/slave" towards "primary/follower", etc.

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,32 @@
...
+    metadata_version
+    process_token
+    recorded_at
+    status
+    updated_at
+  ].freeze
 
-# Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+  WHITELIST_REGEX = /(^|_)ids?|#{Regexp.union(ATTRIBUTE_WHITELIST)}/.freeze
```
harimohanraj89, line 18: I think we can include `/(^|_)ids?/` into the constant array above, and this regex can simply be 

`WHITELIST_REGEX = Regexp.union(ATTRIBUTE_WHITELIST).freeze`

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,32 @@
...
+Rails.application.config.filter_parameters += [:password, :age]
+Rails.application.config.filter_parameters << lambda do |key, value|
+  unless key.match(ParameterFiltering::WHITELIST_REGEX)
+    case value
+    when String
+      value.replace(ParameterFiltering::SANITIZED_VALUE)
+    else
+      value.to_s.replace(ParameterFiltering::SANITIZED_VALUE)
+    end
+  end
```
harimohanraj89, line 31: Hmm, rather than using the module as a bucket of constants, could we have the module generate this lambda for us as well?

```ruby
module ParameterFiletering
  # the constants...

  def self.filter
    lambda do |key, value|
      # the magic...
    end
  end
end

Rails.application.config.filter_parameters += [:password, :age, ParameterFiltering.filter]
```

We could also unit test this module effectively.

config/initializers/filter_parameter_logging.rb
rsanheim: Sorry, I missed that completely. Yup, I'll change it to allow-list or something similar.

config/initializers/filter_parameter_logging.rb
rsanheim: Sure I'll leave a comment explaining this approach, as its a bit different from most rails apps.

config/initializers/filter_parameter_logging.rb
rsanheim: Done. Yeah its nicer having it all in the module.

I'm not too concerned with unit testing this for now, as the request spec covers that this works end-to-end w/ the logging system.

config/initializers/filter_parameter_logging.rb
rsanheim: I want to keep the plain attributes separate from any regex magic, so we can add/remove from that list over time.

spec/requests/filter_parameter_logging_spec.rb
rsanheim: done

PR #993, opened 2020-05-21
Don’t send reminders to dead patients

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/story/show/172817005

## Because

We don't want to send missed appointment reminders to patients marked as `dead`.

## This addresses

This skips patients marked as `dead` in the notification service.


app/services/appointment_notification_service.rb
```diff
@@ -12,9 +12,10 @@ def initialize(appointments:, days_overdue: 3, schedule_at:)
 
   def send_after_missed_visit
     overdue_with_patient_consent = appointments.overdue_by(days_overdue)
```
kitallis, line 14: We could extract out `overdue_with_patient_consent` in its own scope on `Appointment`.

prabhanshuguptagit: Would `eligible_for_reminder` be a better name?

kitallis: Yup.

app/models/appointment.rb
```diff
@@ -54,6 +54,13 @@ def self.overdue_by(number_of_days)
     overdue.where('scheduled_date <= ?', Date.current - number_of_days.days)
   end
 
+  def self.eligible_for_reminders(days_overdue)
```
harimohanraj89, line 57: Can we use a named argument? The plain method signature `Appointment.eligible_for_reminders(3)` looks strange.

spec/models/appointment_spec.rb
```diff
@@ -46,6 +46,17 @@
         expect(Appointment.overdue).not_to include(upcoming_appointment)
       end
     end
+
+    describe '.eligible_for_reminders' do
+      let!(:dead_patient) { create(:patient, status: 'dead') }
+      let!(:patient_with_consent_denied) { create(:patient, :denied) }
+      let!(:overdue_appointment_for_dead_patient) { create(:appointment, :overdue, patient: dead_patient) }
+      let!(:overdue_appointment_for_denied_patient) { create(:appointment, :overdue, patient: patient_with_consent_denied) }
```
harimohanraj89, line 54: Let's explicitly schedule these appointments for 3 days ago. It's a lot clearer than inferring that the current default in the factory is for 3 days.

app/services/appointment_notification_service.rb
```diff
@@ -11,12 +11,12 @@ def initialize(appointments:, days_overdue: 3, schedule_at:)
   end
 
   def send_after_missed_visit
-    overdue_with_patient_consent = appointments.overdue_by(days_overdue)
-                        .includes(patient: [:phone_numbers])
-                        .where(patients: { reminder_consent: 'granted' })
-                        .merge(PatientPhoneNumber.phone_type_mobile)
+    eligible_appointments = appointments
+                             .eligible_for_reminders(days_overdue: days_overdue)
```
kitallis, line 15: @prabhanshuguptagit Any reason why you didn't squeeze up 

```
.includes(patient: [:phone_numbers]).merge(PatientPhoneNumber.phone_type_mobile)
```

into the `eligible_for_reminders` definition?

I feel like freezing the complete definition as we currently understand it is valuable in one place. We can't send  reminders to folks who don't have mobile numbers anyway (as per our current understanding).

wdyt?

prabhanshuguptagit: > We can't send reminders to folks who don't have mobile numbers anyway (as per our current understanding).

That makes sense. The reason to leave it out is to let the call site decide what number it wants to use (or if it even wants to include the number) but it definitely belongs in there given this.

prabhanshuguptagit: Added in 9c512a4

PR #994, opened 2020-05-21
Apply standardrb to models, mailers, and controllers

rsanheim:
Going to apply this to some of our core files and see how it plays out. I'll leave a review with thoughts -- if it looks good we can merge away to avoid bad conflicts, and iterate on any rule overrides in subsequent PRs.

```
  standardrb --fix app/models/**.rb
  standardrb --fix spec/models/**.rb
  standardrb --fix spec/mailers/**.rb
  standardrb --fix app/mailers/**.rb
  standardrb --fix app/controllers/**.rb
  standardrb --fix spec/controllers/**.rb
```

**Story card:** n/a just applying linting

/cc #983 

## Because

We want a consistent format for code style

## This addresses

applying standardrb to models, mailers, and controllers, and all of their specs


app/models/concerns/mergeable.rb
```diff
@@ -8,16 +8,15 @@ module Mergeable
   class_methods do
     def compute_merge_status(attributes)
       new_record = new(attributes)
-      existing_record = with_discarded.find_by(id: attributes['id'])
+      existing_record = with_discarded.find_by(id: attributes["id"])
 
-      case
```
rsanheim, line 13: I'm really not happy about standardrb deciding to change all case statements to if / elsif statemnts. How would the team feel about overriding that one rule globally?

FWIW there was a lot of push back on that rule when it [landed in Rubocop](https://github.com/rubocop-hq/rubocop/issues/3019).

app/models/appointment.rb
```diff
@@ -103,22 +103,21 @@ def mark_patient_already_visited
   end
 
   def mark_patient_as_dead
-    self.patient.status = :dead
-    self.patient.save
+    patient.status = :dead
+    patient.save
```
rsanheim, line 107: This type of change makes sense to me.  The `self` is redundant for method calls in ruby, _unless_ you operating only on `self`, for example:

```ruby
class Patient
  def assign=(val)
     @attributes[val] = val
  end

  def do_stuff
    self.assign = "value"
  end
```

app/models/appointment.rb
```diff
@@ -103,22 +103,21 @@ def mark_patient_already_visited
...
-      created_at: created_at,
-      registration_facility_name: facility.name,
-      user_id: hash_uuid(patient&.registration_user&.id),
-      scheduled_date: scheduled_date,
-      overdue: days_overdue > 0 ? 'Yes' : 'No',
-      status: status,
-      agreed_to_visit: agreed_to_visit,
-      remind_on: remind_on
-    }
+    {id: hash_uuid(id),
```
rsanheim, line 111: This always takes some getting used to for me, but I understand the reasoning. You can read some of the discussion here: https://github.com/testdouble/standard/issues/26.

Totally open to trying this out for a couple weeks, I imagine the distinction will be worth it.

app/models/communication.rb
```diff
@@ -48,12 +48,11 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
...
   def communication_result
-    case
-    when successful? then
+    if successful?
       COMMUNICATION_RESULTS[:successful]
-    when unsuccessful? then
+    elsif unsuccessful?
       COMMUNICATION_RESULTS[:unsuccessful]
-    when in_progress? then
+    elsif in_progress?
```
rsanheim, line 55: This one bothers me, and I don't see this as anything really functional.  Will bring this one up for possible override elsewhere.

app/models/facility.rb
```diff
@@ -124,20 +124,23 @@ def organization_exists
 
   def facility_group_exists
     organization = Organization.find_by(name: organization_name)
-    facility_group = FacilityGroup.find_by(name: facility_group_name,
-                                           organization_id: organization.id) if organization.present?
```
rsanheim, line 128: 👍 I rarely use the suffix `if / unless`, and I know it can be hard to catch sometimes as line length creeps up.

app/controllers/admin_controller.rb
```diff
@@ -17,30 +17,31 @@ class AdminController < ApplicationController
...
-      :manage_admins? => admins_path,
-      :manage_users? => admin_users_path }
+    {show?: organizations_path,
+     adherence_follow_up?: patients_path,
+     overdue_list?: appointments_path,
+     manage_organizations?: admin_organizations_path,
+     manage_facilities?: admin_facilities_path,
+     manage_protocols?: admin_protocols_path,
+     manage_admins?: admins_path,
+     manage_users?: admin_users_path}
```
rsanheim, line 44: This is a nice improvement to get for free.

app/controllers/admin_controller.rb
```diff
@@ -17,30 +17,31 @@ class AdminController < ApplicationController
...
+      http_accept_language.compatible_language_from(I18n.available_locales) ||
+      I18n.default_locale
 
     I18n.with_locale(locale, &action)
   end
 
   def root
     redirect_to default_root_paths.find { |policy, _path|
-      DashboardPolicy.new(pundit_user, :dashboard).send(policy) }.second
+                  DashboardPolicy.new(pundit_user, :dashboard).send(policy)
```
rsanheim, line 28: This is weird, but the fact that the underlying code is using `send` like this is actually the bigger warning flag. 😨

app/controllers/resources_controller.rb
```diff
@@ -4,5 +4,6 @@ class ResourcesController < AdminController
   skip_after_action :verify_authorized, only: [:index]
   skip_after_action :verify_policy_scoped, only: [:index]
 
-  def index; end
+  def index
+  end
```
rsanheim, line 8: 👍

app/models/concerns/hashable.rb
```diff
@@ -1,11 +1,7 @@
 module Hashable
-  extend ActiveSupport::Concern
+  def hash_uuid(original_uuid)
+    return nil if original_uuid.blank?
 
-  included do
```
rsanheim, line 4: I did this by hand -- there is no reason to use AS::Concern where a plain ole' module will do.

app/models/facility.rb
```diff
@@ -160,7 +163,13 @@ def teleconsultation_phone_number_with_isd
...
   end
 
-  CSV::Converters[:strip_whitespace] = ->(value) { value.strip rescue value }
+  CSV::Converters[:strip_whitespace] = ->(value) {
+    begin
+                value.strip
+    rescue
+      value
+              end
+  }
```
rsanheim, line 172: Not sure what happened here.

spec/rails_helper.rb
```diff
@@ -17,6 +17,7 @@
 ActiveRecord::Migration.maintain_test_schema!
 
 RSpec.configure do |config|
+  config.include Hashable
```
rsanheim, line 20: This brings Hashable to specs, as it seems a few specs needed it. They were including it at top level, which is really dangerous, as it includes it in **every object** in ruby via the top level object.

app/controllers/admin/facilities_controller.rb
```diff
@@ -114,8 +114,9 @@ def validate_at_least_one_facility
 
   def validate_duplicate_rows
     facilities_slice = @facilities.map { |facility|
-      facility.slice(:organization_name, :facility_group_name, :name) }
-    @errors << 'Uploaded file has duplicate facilities' if
+      facility.slice(:organization_name, :facility_group_name, :name)
+    }
```
harimohanraj89, line 118: Semantic approach to `{}` vs `do end`, nice. On reading more about this, my favorite reason to do this so far is:

> As Nick Morgan points out below, chaining method calls onto an end looks kinda weird (do ... end.foo)

app/controllers/admin_controller.rb
harimohanraj89: This indentation is wonky.

The `send` issue, I think it's just some metaprogramming, and we could very well use `public_send` instead. Looks like we're not accessing private methods, just invoking a bunch of Pundit policy actions.

app/controllers/admins_controller.rb
```diff
@@ -57,9 +58,9 @@ def permission_params
...
 
   def user_params
-    { full_name: params[:full_name],
-      role: params[:role],
-      organization_id: params[:organization_id],
-      device_updated_at: Time.current }.compact
+    {full_name: params[:full_name],
+     role: params[:role],
+     organization_id: params[:organization_id],
+     device_updated_at: Time.current}.compact
```
harimohanraj89, line 64: I'm not a fan of this indentation style yet. I much prefer:

```
{
  full_name: params[:full_name],
}
```

It adds lines, but I feel is so much more readable.

app/controllers/analytics/districts_controller.rb
```diff
@@ -15,39 +15,40 @@ def show
...
 
     redirect_to(
       analytics_organization_district_path(id: @organization_district.district_name),
-      notice: I18n.t('patient_list_email.notice',
-                     model_type: "district",
-                     model_name: @organization_district.district_name))
+      notice: I18n.t("patient_list_email.notice",
+        model_type: "district",
+        model_name: @organization_district.district_name)
+    )
```
harimohanraj89, line 51: Related to my hash formatting comment, this inconsistency makes me raise an eyebrow as well. The `redirect_to` lists its arguments on new lines, but the `I18n` includes its first argument on the same line as the method. Is this just an artefact of the autocorrect, or does standardrb have rules explaining this behavior? I'd love to nail this down, as I think it's one of the more problematic style areas.

Here's what I'd love to see:

```ruby
redirect_to(
  analytics_organization_district_path(id: @organization_district.district_name),
  notice: I18n.t(
    "patient_list_email.notice",
    model_type: "district",
    model_name: @organization_district.district_name
  )
)
```

allowing, of course, one-lining the entire method if it isn't too long.

app/controllers/analytics/districts_controller.rb
```diff
@@ -69,28 +71,28 @@ def set_organization_district
...
                                                    include_current_period: @show_current_period)
-      end
+      }
   end
 
   def set_facility_keys
     facilities = @organization_district.facilities.order(:name).pluck(:id, :name).to_h
 
-    @facility_keys = { total: "Total" }.merge(facilities).with_indifferent_access
+    @facility_keys = {total: "Total"}.merge(facilities).with_indifferent_access
```
harimohanraj89, line 91: I can dig the lack of spaces between hash brackets 👍

harimohanraj89:
@rsanheim this PR is quite large indeed 😅 Reckon we can make round 1 a lot smaller so we can find things we like/don't like and adjust them before we apply to a larger chunk?

app/models/concerns/mergeable.rb
kitallis: I really much prefer `case` statements over long if/else. I'm 👍 for overriding that rule.

rsanheim:
> this PR is quite large indeed 😅 Reckon we can make round 1 a lot smaller so we can find things we like/don't like and adjust them before we apply to a larger chunk?

Sure, I'll close this out and open up a new, smaller PR.

I'd like to propose we land the first PR with standardrb defaults intact, for a few reasons:

a) we can avoid merge hell that way
b) its more in-line with the spirit of standardrb, which is to just have a sane set of defaults that you avoid modifying if at all possible
c) its easy to override any rules in the future and re-apply them via autofix w/ the CLI tool

WDYT?

PR #999, opened 2020-05-25
Enable US+UK for Review apps

kitallis:
## Because

A quick stop-gap solution to testing Simple SMSes on non-Indian phone numbers on Review apps.

## This addresses

Adds US (and Canada by extension) and UK as countries to the review manifest with appropriate ISD codes so we can easily tweak `DEFAULT_COUNTRY` config in the review envs and receive messages.

This is to help simplify testing full app flows with OTPs for our distributed team in the US, Canada and London on the review apps.

**Note:** Sending SMSes to these countries must be enabled on Twilio console. It can be easily enabled from [here](https://www.twilio.com/console/sms/settings/geo-permissions) (I'm uncertain whether this has extra charges other than the SMS cost).


kitallis:
> After merging dynamic manifests, I realized that we'll have to add these ISD codes to the country config initializer too, right? The ISD code is pulled from there before constructing the full phone number to send an SMS to.
> 
> PS Dynamic manifests are merged, so you should be all set to add these countries to the [dynamic manifest](https://github.com/simpledotorg/simple-server/blob/master/app/views/api/manifests/show.json.jbuilder) too. Let me know if you wanna pair.

Country configs might be hard to change this way. Wondering if we should just store the ISD code in the DB along with the phone number instead...

public/manifest/development.json
```diff
@@ -0,0 +1,34 @@
...
+      "isd_code": "1"
+    },
+    {
+      "country_code": "UK",
+      "endpoint": "http://localhost:3000/api/",
+      "display_name": "United Kingdom",
+      "isd_code": "44"
+    }
+  ]
+}
```
harimohanraj89, line 34: This file isn't necessary anymore. I think it got re-added through some merge-fu.

harimohanraj89:
Derp, we have manifest controller specs now, so they'll need updating too 😄

PR #1009, opened 2020-05-27
Add proposed ADR for rollup summary tables for dashboards

rsanheim:
**Story card:** TBD

## Because

We want to document architecture decisions

## This addresses

a proposed approach for getting data for dashboard reports

### DIFF
--- a/doc/arch/011-rollup-summary-tables-for-dashboards.md
+++ b/doc/arch/011-rollup-summary-tables-for-dashboards.md
@@ -1,0 +1,65 @@
+# ADR 11 Rollup summary tables for Dashboards
+
+## Context
+Earlier we went with a [materialized view approach](https://github.com/simpledotorg/simple-server/blob/master/doc/arch/008-materialized-views-for-dashboards.md) for dashboard reporting data. This approach has worked out well for the current reports but has limitations in terms of scaling and complexity.
+
+We currently have a need for a new set of dashboard reports that rely on similar sets of data, and anticipate more needs for summary data for monthly and quarterly reports. We'd like to try to find a simpler, more stable, and more scalable approach to making that data easily accessible.
+
+## Decision
+
+We will create rollup summary tables for to capture the "most recent blood pressure per patient per period". These rollup tables will be updated in real-time as blood pressures are captured, removing the need for recurring batch jobs. We will denormalize as much data as we need to to make reporting queries simple and cheap to execute, to prevent the need for any complicated SQL queries or additional data layer caching.
+
+Once the current period has completed, we will treat the data in these rollup tables for that period as immutable. This will let us and our users have more confidence in reports from past periods, as they will not change due to patients moving to different facilities or changes in how we calculate things.
+
+We will focus on the District controlled blood pressure trend report as a first test case for this sort of approach, as it is a key report and one we'd like to gather feedback on as quickly as possible.
+
+## Approach
+
+### Defining the tables
+
+* We will start with the district based reports for BP controlled numbers over time. We will create a rollup table with a schema like the following (consider this a draft, it will change as we get into implementation):
+```
+id
+patient_id
+registration_facility_id
+blood_pressure_at_facility_id
+blood_pressure_id
+diastolic
+systolic
+recorded_at
+period_type
+period_name
+year
+```
+* `period_type` will be either `M` or `Q` (month or quarter), `period_name` will be either the month name or quarter number, and year will be year.
+* We will also start rollup tables to store the total number of registrations per period:
+```
+id
+facility_id
+registrations_count
+new_registrations_count
+period_type
+period_name
+year
+```
+
+* We will add logic to at the lowest level of our model layer to update the above rollup tables immediately after the inserts/updates. We will use [UPSERT](https://wiki.postgresql.org/wiki/UPSERT#.22UPSERT.22_definition) to atomically update or insert the 'most recent record' for these tables wherever possible.
+* We will build out a backfill job to fill out these tables for all previous periods.
+
+## Status
+Proposed
+
+## Consequences
+
+### Neutral
+* We need to be careful to avoid denormalizing calculated values representing domain logic -- for example, we should store the raw blood pressure values, and not a boolean for 'controlled'.
+* We should err on the side of storing more data than we think we may need, as we know reporting changes are often fluid and can change frequently. We want to minimize the need for schema changes for small reporting changes
+* The notion of an 'assigned facility' is something coming soon, and we should be mindful of how we will take that into account
+### Advantages
+* Updates to these tables should be relatively easy to implement without database contention.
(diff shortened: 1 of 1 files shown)


doc/arch/011-rollup-summary-tables-for-dashboards.md
```diff
@@ -0,0 +1,65 @@
+# ADR 11 Rollup summary tables for Dashboards
+
+## Context
+Earlier we went with a [materialized view approach](https://github.com/simpledotorg/simple-server/blob/master/doc/arch/008-materialized-views-for-dashboards.md) for dashboard reporting data. This approach has worked out well for the current reports but has limitations in terms of scaling and complexity.
+
+We currently have a need for a new set of dashboard reports that rely on similar sets of data, and anticipate more needs for summary data for monthly and quarterly reports. We'd like to try to find a simpler, more stable, and more scalable approach to making that data easily accessible.
+
+## Decision
+
+We will create rollup summary tables for to capture the "most recent blood pressure per patient per period". These rollup tables will be updated in real-time as blood pressures are captured, removing the need for recurring batch jobs. We will denormalize as much data as we need to to make reporting queries simple and cheap to execute, to prevent the need for any complicated SQL queries or additional data layer caching.
```
kitallis, line 10: > most recent blood pressure per patient per period

This sounds like the views that we have. 

Could we better establish some actual bottlenecks on why a daily batch job that refreshes views is problematic and a real-time job over every sync call helps mitigate those problems?

doc/arch/011-rollup-summary-tables-for-dashboards.md
```diff
@@ -0,0 +1,65 @@
...
+
+Once the current period has completed, we will treat the data in these rollup tables for that period as immutable. This will let us and our users have more confidence in reports from past periods, as they will not change due to patients moving to different facilities or changes in how we calculate things.
```
kitallis, line 12: The immutability makes sense, but does this also mean that a current period report is not possible until the period is over? What if the period is current month and we're in the middle of the month?

doc/arch/011-rollup-summary-tables-for-dashboards.md
```diff
@@ -0,0 +1,65 @@
...
+## Consequences
+
+### Neutral
+* We need to be careful to avoid denormalizing calculated values representing domain logic -- for example, we should store the raw blood pressure values, and not a boolean for 'controlled'.
+* We should err on the side of storing more data than we think we may need, as we know reporting changes are often fluid and can change frequently. We want to minimize the need for schema changes for small reporting changes
+* The notion of an 'assigned facility' is something coming soon, and we should be mindful of how we will take that into account
+### Advantages
+* Updates to these tables should be relatively easy to implement without database contention.
+* Retrieving data from these tables should be fast and simple. Queries should be relatively simple and inexpensive.  We will denormalize whatever data is needed to make queries fast.
+* If this approach is successful, we can slowly phase out the materialized views over time with corresponding rollup tables. This can be an incremental rollout based on new reports and scaling demands. This should remove the stress on our database of continual materialized view refreshes.
```
kitallis, line 61: I don't think the views cause any stress to the database, they are concurrent refreshes that run in the background with no exclusive locks on any tables. Could we provide better evidence here?

doc/arch/011-rollup-summary-tables-for-dashboards.md
```diff
@@ -0,0 +1,65 @@
...
+* We will build out a backfill job to fill out these tables for all previous periods.
+
+## Status
+Proposed
+
+## Consequences
+
+### Neutral
+* We need to be careful to avoid denormalizing calculated values representing domain logic -- for example, we should store the raw blood pressure values, and not a boolean for 'controlled'.
+* We should err on the side of storing more data than we think we may need, as we know reporting changes are often fluid and can change frequently. We want to minimize the need for schema changes for small reporting changes
```
kitallis, line 56: 👍

doc/arch/011-rollup-summary-tables-for-dashboards.md
rsanheim: > This sounds awfully similar to the views that we have 😉

Agreed 😄 . I think the fundamental difference is that the rollup table will be updated once for each new blood pressure that comes into the system. Once a BP is entered on May 29th for a patient, that row for that patient won't have to be updated or touched again (assuming the patient doesn't come back for more BP readings in May). 

With the materialized views, we are taking up CPU / DB time writing that May 29th BP, _all other_ BPs every time the views get refreshed. So we are doing way more work than we need to with the views vs a rollup summary approach.

> Could we better establish some actual bottlenecks on why a daily batch job that refreshes views is problematic and a real-time job over every sync call helps mitigate those problems?

I would love to, but we need better metrics and better visibility in the system first to do that. Something like grafana or prometheus, or maybe we could use custom new relic tracing.  

If its not a problem right now, I can be pretty sure that recomputing the entire blood pressure table into views will be a problem when we have tens of millions of blood pressures in the system.

doc/arch/011-rollup-summary-tables-for-dashboards.md
rsanheim: Yup exactly.

doc/arch/011-rollup-summary-tables-for-dashboards.md
rsanheim: I've added a very basic level of logging here: https://github.com/simpledotorg/simple-server/pull/1026

doc/arch/011-rollup-summary-tables-for-dashboards.md
kitallis: One thing to be aware of is that period-level locks will only allow back-dated BP / registrations to show up in the appropriate period report that hasn't been fully rolled up yet. 

So a monthly report from May will not show a back-dated BP if a registration is backdated to May, but entered in June. But it will show up in the quarterly / yearly reports. And so on. Basically, backdates will be only reported within currently active periods.

harimohanraj89:
Hey @rsanheim , I gave my early feedback to you directly, I dig this approach.

kitallis:
@rsanheim @harimohanraj89 

Just so that this isn't left hanging, I think the approach sounds good to me as well 👍. 

My only remaining 2c would be to put the notes in this [comment](https://github.com/simpledotorg/simple-server/pull/1009#discussion_r432742159) back into the ADR because they provide a stronger argumen. And some basic projected numbers on a linear scale; like _1mil BPs refresh in 1 minutes, 1.3 billion (population of India) will refresh in 10 hours_ or something like that.

rsanheim:
One thing I discovered in the spike PR (and we've discussed a bit in Slack) is that handling `updated` blood pressures is going to be tricky with the rollup approach, at least in the case where a BP has its date back-dated to a different month or quarter period.  I'll add a note about this as a drawback to the ADR text.

I think this is an acceptable drawback because these types of updates should be very rare, and if we have to we can do the engineering to properly handle those cases.  But if we have some sort of major sync / data issues where we need to do large edits of BPs, its worth considering this drawback.

PR #1011, opened 2020-05-27
Run standardrb on files related to blood pressure sync API

rsanheim:
**Story card:** https://www.pivotaltracker.com/story/show/173035489

## Because

I'm working on these files for the rollup story and we want to try out standardrb w/ a smaller set of files.  The first standard PR #994 got too big, and this one is more manageable.

I'd like to propose we land this first PR with standardrb defaults intact, for a few reasons:

a) we can avoid merge hell that way
b) its more in-line with the spirit of standardrb, which is to just have a sane set of defaults that you avoid modifying if at all possible
c) its easy to override any rules in the future and re-apply them via autofix w/ the CLI tool, as autofix PR's are cheap and easy

## This addresses

running standardrb --fix on a set of files related to BP sync


kitallis:
👍

PR #1013, opened 2020-05-28
Support user initiated soft deletes

vkrmis:
**Story card:**  [1](https://www.pivotaltracker.com/story/show/172930521), [2](https://www.pivotaltracker.com/story/show/172930747)

## Because

We don't currently have a way for Users to soft-delete patients. This has lead to requests to do de-duplication manually on the server.

## This addresses

Adds soft-deletion functionality to the V3 Patients sync API, so Users can soft-delete patients and their data
 
[PRD](https://docs.google.com/document/d/1QLEAPAkqBSJNvtLSimI60keiLrdnhuctot2_oWiGoKg/edit#)


app/services/merge_patient_service.rb
```diff
@@ -4,6 +4,7 @@ def initialize(payload)
   end
 
   def merge
+    existing_patient_attributes = Patient.with_discarded.find_by(id: payload['id'])&.attributes
```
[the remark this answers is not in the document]
vkrmis: Yes. Since the patient payload contains patients as well as their address, phone numbers etc., this service handles pushing updates to the individual models

app/services/merge_patient_service.rb
```diff
@@ -30,6 +31,13 @@ def merge
       log_update_discarded_patient(merged_patient)
     end
 
+    if (merged_patient.deleted_at.present? && existing_patient_attributes&.dig('deleted_at').nil?)
+      # Patient has been soft-deleted by the client, server should soft-delete the patient and their associated data
+      # patient_attributes[:metadata][:registration_user_id] contains the current user's id
+      merged_patient.update(deleted_by_user_id: patient_attributes[:metadata][:registration_user_id])
+      merged_patient.discard_data
+    end
```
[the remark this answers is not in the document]
vkrmis: I will do this separately, I will also need to refactor Patient metadata since that has poorly named fields.

spec/controllers/api/v3/patients_controller_spec.rb
```diff
@@ -268,6 +270,136 @@ def create_record_list(n, options = {})
...
+
+      it "sets the patient's deleted_reason and deleted_user_id" do
+        expect(Patient.find(existing_patient.id).present?).to eq(true)
+        post :sync_from_user, params: { patients: [delete_patient_payload] }, as: :json
+
+        expect(Patient.with_discarded.find_by(id: existing_patient.id).deleted_reason).to eq('duplicate')
+        expect(Patient.with_discarded.find_by(id: existing_patient.id).deleted_by_user_id).to eq(request_user.id)
+      end
+
+      it "soft deletes the patient's encounters" do
```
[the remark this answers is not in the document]
vkrmis: We didn't, I rewrote the specs for `Patient#discard_data` now, and refactored these specs.

Also discovered a small bug in discard_data in the process, fix + data_migration are included.

vkrmis:
> Regarding your caveat around undeleted BPs added to deleted patients, what's the reporting impact of those BPs being persisted? Will they affect our follow-up numbers, etc?

This is hard to answer definitively. It unfortunately depends on how each query is constructed. Queries that use the `Patient` model are safe, but queries on say `BloodPressure` directly that don't join `Patient` or check for `Patient.deleted_at` will include these. 

I'd say its worth investing the time to refactor `Mergeable` so this data never ends up in the database. An intermediate fix would be to periodically run a task that can identify these records and soft-delete them as well.

spec/models/patient_spec.rb
```diff
@@ -459,32 +459,61 @@
...
+    end
+
+    it "soft deletes the patient's medical history" do
+      patient.discard_data
+      expect(MedicalHistory.where(patient: patient)).to be_empty
+    end
+
+    it "soft deletes the patient's address" do
+      patient.discard_data
+      expect(Address.where(id: patient.address_id)).to be_empty
```
harimohanraj89, line 516: These are much better than asserting specific counts. Love it.

spec/controllers/api/v3/patients_controller_spec.rb
```diff
@@ -268,6 +270,51 @@ def create_record_list(n, options = {})
...
+
+        existing_patient.discard_data
+        update_payload_for_discarded_patient = build_patient_payload(existing_patient)
+                                                 .merge(full_name: "Test Patient Name Xcad7asd")
+
+        post :sync_from_user, params: { patients: [update_payload_for_discarded_patient] }, as: :json
+
+        expect(Patient.with_discarded.find_by(id: existing_patient.id).full_name).to eq(existing_patient_name)
+      end
+    end
```
harimohanraj89, line 317: We still need a spec here that says something like `expect(patient).to receive(:discard_data)`, right?

harimohanraj89:
Let's look through our queries now so we understand the impact right away, rather than having to discover inconsistencies if our caveat happens a lot.

spec/controllers/api/v3/patients_controller_spec.rb
vkrmis: Done, TIL how to mock this, thanks!

spec/controllers/api/v3/patients_controller_spec.rb
```diff
@@ -287,6 +287,29 @@ def create_record_list(n, options = {})
...
+          allow(patient).to receive(:slice)
+                              .with("registration_user_id", "registration_facility_id")
+                              .and_return(existing_patient.slice("registration_user_id",
+                                                                 "registration_facility_id"))
+          allow(patient).to receive_messages(id: existing_patient.id,
+                                             merge_status: :updated,
+                                             deleted_at: deleted_time,
+                                             update: patient)
+
+          allow(patient).to receive(:address=).and_return(existing_patient.address)
```
harimohanraj89, line 307: Could we put this in the `receive_messages` above? `"address=": existing_patient.address`?

PR #1014, opened 2020-05-28
Create multiple sidekiq processes

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/story/show/173008596

## Because

_from the pivotal story_: Sidekiq periodically dies because of single-core per-process affinity (as an effect of running MRI) when the CPU utilization goes too high. We should run multiple sidekiq processes and fully utilize our ec2 boxes so one long running job don't kill processing.

## This addresses

This adds support to spawn multiple sidekiq processes. This ditches the `capistrano-sidekiq` gem and adds our own systemd files for the following reasons:

- `capistrano-sidekiq` supports a `sidekiq_processes => n` flag but it only works if we use monit, not systemd
- monit requires root access for sidekiq setup/restart/stop. This means giving the `deploy` user root access (which is not great), _or_
- make sidekiq work through ansible. That will cause a lot of fragmentation in our deployments.

The new setup mirrors the existing flow, so this does not introduce any changes in the deployment process.

To get the multiple processes going on an env after a deploy we need to:
- Run `bundle exec cap <stage> sidekiq:install`
- Stop and disable the old sidekiq service (`sidekiq-<env>.service`)
- Remove the old systemd file (`~/.config/systemd/user/sidekiq-<env>.service`)


prabhanshuguptagit:
@rsanheim Here's a screencap https://drive.google.com/file/d/1mV-0SOaBkIGHk6nIx08UDpoq3nH1iqfV/view. The lone service running is the existing one.

prabhanshuguptagit:
The new service files are already on SBX, so we should be good. We will need to get rid of the old ones. The hooks work on `cap deploy`, this is tested on sandbox.👍

PR #1015, opened 2020-05-28
Rollup summary spike

rsanheim:
**Story card:** https://www.pivotaltracker.com/story/show/173035489

## Because

## This addresses


app/controllers/api/v3/blood_pressures_controller.rb
```diff
@@ -26,6 +26,17 @@ def merge_if_valid(bp_params)
     end
   end
 
+  def after_merge_completed(results)
+    records = results.flat_map { |r| r[:record] }.compact
+    send_to_job(records)
+  end
+
+  def send_to_job(records)
```
harimohanraj89, line 34: Naming nitpick. Can we pick a name that's more indicative of the "what" and not the "how"?

app/controllers/api/v3/sync_controller.rb
```diff
@@ -31,6 +30,20 @@ def __sync_to_user__(response_key)
...
+      AuditLog.merge_log(current_user, result[:record]) if result[:record].present?
+      result
+    }
+  end
+
+  # This is a hook method for subclasses that need to operate on the results after the merge but before the response
+  # is sent back to the client.
+  # results will be an array of Hashes, with the key values of either { :record => [Model] } or { :errors_hash => Hash }
+  def after_merge_completed(results)
+  end
```
harimohanraj89, line 45: Love this. No-op for most controllers, and we can start inserting roll-up logic tactically.

app/jobs/blood_pressure_rollup_backfill.rb
```diff
@@ -0,0 +1,14 @@
+class BloodPressureRollupBackfill < ApplicationJob
```
harimohanraj89, line 1: Wanna slap a `Job` suffix on there for consistency?

app/models/blood_pressure_rollup.rb
```diff
@@ -0,0 +1,74 @@
...
+      recorded_at: blood_pressure.recorded_at,
+      systolic: blood_pressure.systolic,
+      year: blood_pressure.recorded_at.year
+    }
+
+    month = blood_pressure.recorded_at.month
+    quarter = quarter_for_month(month)
+
+    month_attrs = attrs.merge(period_type: :month, period_number: month)
+    quarter_attrs = attrs.merge(period_type: :quarter, period_number: quarter)
```
harimohanraj89, line 51: Think we could split this up into `base_attrs`, `month_attrs`, and `quarter_attrs`?

```ruby
month_rollup=BloodPressureRollup.new(**base_attrs, **month_attrs)
quarter_rollup=BloodPressureRollup.new(**base_attrs, **quarter_attrs)

def base_attrs
  # stuff
end

def month_attrs
  # stuff
end

def quarter_attrs
  # stuff
end
```

app/jobs/blood_pressure_rollup_backfill.rb
rsanheim: Yup

app/models/blood_pressure_rollup.rb
rsanheim: yeah I plan on refactoring all of this at some point, its pretty ugly right now.

app/controllers/api/v3/blood_pressures_controller.rb
rsanheim: Yup renamed. 👍

app/models/blood_pressure_rollup.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    month = blood_pressure.recorded_at.month
+    quarter = quarter_for_month(month)
+
+    month_attrs = attrs.merge(period_type: :month, period_number: month)
+    quarter_attrs = attrs.merge(period_type: :quarter, period_number: quarter)
+    month_rollup = BloodPressureRollup.new(month_attrs)
+    result = month_rollup.upsert(attributes: [:blood_pressure_id, :diastolic, :period_number, :systolic, :recorded_at],
+                        arel_condition: BloodPressureRollup.arel_table[:recorded_at].lt(blood_pressure.recorded_at))
```
harimohanraj89, line 58: Nice. Upsert the record matching the `attributes`, as long as this BP is the more recent one. Do we need to add `year` to the list of `attributes`?

harimohanraj89: And isn't the BP ID changeable? We'd want to omit it from this list?

app/models/blood_pressure_rollup.rb
```diff
@@ -0,0 +1,86 @@
...
+      SELECT count(1)
+      FROM (
+        SELECT DISTINCT ON (patient_id) *
+        FROM #{table_name}
+        WHERE period_number in (?)
+        AND period_type = 0
+        ORDER BY patient_id ASC, (year, period_number) DESC
+      ) AS counts
+      WHERE diastolic >= 90
+      and systolic >= 140
```
harimohanraj89, line 29: Nitpick: These should be `<` to represent controlled BPs right?

app/models/blood_pressure_rollup.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  validates :diastolic, presence: true
+  validates :systolic, presence: true
+  validates :period_number, presence: true, numericality: {only_integer: true}
+  validates :period_type, presence: true
+  validates :year, presence: true, numericality: {greater_than_or_equal_to: 2000, only_integer: true}
+
+  def self.controlled_in_month(time)
+    range = [time.month, time.advance(months: -1).month, time.advance(months: -2).month]
+    sql = <<-SQL
```
harimohanraj89, line 19: If I understand correctly, grab the most recent BP of each patient over the last three months, and see how many of those are under control. If so, can we name that subselect something other than `counts`? That may help future-me understand this a bit quicker 😅

harimohanraj89:
One high-level thought. We chatted briefly about our data models not being consistent with our API validations, which causes data consistency problems. Eg. we're forced to do data imports through our API because of this.

Given we wanted our data model to be more robust, is the API layer the right place to do the rollup? Should it be at the model layer? What would that look like? Any disadvantages to putting this in the model?

rsanheim:
Closing this in favor of https://github.com/simpledotorg/simple-server/pull/1039, focusing on the dashboard and getting it live right now.

PR #1018, opened 2020-05-29
Data fixes: Bad dates

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/172820945

## Because

Some patients registered in 16AD, and that breaks our reports.

## This addresses

This PR adds a data cleanup rake task that:
* Identifies patients with non-sensical registration dates (`patients.recorded_at`)
* Soft-deletes their offending BPs and BSs
* Updates their registration date based on their first _valid_ measurement.

We choose Jan 1, 2000 as the cut-off, as that covers _most_ invalid readings, and definitely fixes our reporting.

This task is primarily being added to run in BD production, but can be run in IHCI production as well, to take care of [reports like these](https://simpledotorg.slack.com/archives/CFHKJ5WJY/p1590559770006400).

## Estimate Impact

* 49 patients in BD


lib/tasks/scripts/clean_ancient_dates.rb
```diff
@@ -0,0 +1,62 @@
...
+    ancient_blood_sugars.each(&:discard!)
+  end
+
+  def update_registration_for(patient)
+    patient.update!(recorded_at: registration_date_for(patient))
+  end
+
+  def registration_date_for(patient)
+    [
+      *patient.reload.blood_pressures.map(&:recorded_at),
```
kitallis, line 49: Is it possible that this patient does not have any BPs or BSs? Does this still compact the date list correctly?

harimohanraj89: Yup, correctly sets `recorded_at` to patient's `device_created_at` 👍 Added a spec for this in 80b7273 as well.

PR #1026, opened 2020-05-29
Add logging to refresh materialized views

rsanheim:
**Story card:** n/a for traceability

## Because

We want to know how long it takes to refresh materialized views

## This addresses

adding logging to show how long they take

FWIW refreshing locally takes about 32 seconds, this is with ~818k blood pressures. Our data set size is still low enough where I wouldn't expect this to be a problem, but its good to track to see how it grows over time.


rsanheim:
> Amaze. Do these times just get spit out into application logs?

Yup.

PR #1032, opened 2020-06-02
Line list medication fix

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/173123723

## Because

We should only show patient's current medications in the line list

### DIFF
--- a/app/exporters/patients_exporter.rb
+++ b/app/exporters/patients_exporter.rb
@@ -119,6 +119,6 @@ module PatientsExporter
   end
 
   def self.medications_for(patient)
-    patient.prescription_drugs.flat_map { |drug| [drug.name, drug.dosage] }
+    patient.current_prescription_drugs.flat_map { |drug| [drug.name, drug.dosage] }
   end
 end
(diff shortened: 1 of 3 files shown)


app/exporters/patients_exporter.rb
```diff
@@ -119,6 +119,6 @@ def self.csv_fields(patient)
   end
 
   def self.medications_for(patient)
-    patient.prescription_drugs.flat_map { |drug| [drug.name, drug.dosage] }
+    patient.current_prescription_drugs.flat_map { |drug| [drug.name, drug.dosage] }
```
kitallis, line 122: This scope, of course.

harimohanraj89: I know. I think we assumed initially that the `deleted_at` flag was being set on deleted prescription drugs, in which case the default scope would exclude deleted drugs. Now we know this isn't the case.

kitallis: Yeah, maybe during the medication history stuff, we could see if we can move this to be `deleted_at` and not this special `is_deleted` flag.

harimohanraj89: Yup, we've listed it as a "nice-to-do-while-we're-here" 👍

PR #1033, opened 2020-06-02
Add address fields to patient line list

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/story/show/173045369

## Because

We want to show the full patient address in the patient line list. Currently the `street_address` and  `zone` fields are missing. We also only want to show `zone` in Bangladesh only.

## This addresses

This adds the fields to the line list and a country config to show/hide the column.


app/exporters/patients_exporter.rb
```diff
@@ -116,9 +120,18 @@ def self.csv_fields(patient)
...
   end
 
   def self.medications_for(patient)
     patient.prescription_drugs.flat_map { |drug| [drug.name, drug.dosage] }
   end
+
+  private
+
+  def self.zone_column
+    "Patient #{Address.human_attribute_name :zone}"
```
harimohanraj89, line 135: Do we need to jump through AR hoops because "zone" is called different things in different countries?

app/exporters/patients_exporter.rb
```diff
@@ -116,9 +120,18 @@ def self.csv_fields(patient)
       patient.id,
       *medications_for(patient)
     ]
+
+    csv_fields.insert(zone_column_index, patient.address.zone) if zone_column_index
```
harimohanraj89, line 124: Ugh, this makes me so uncomfortable, but I think it's the simplest way to do this 😅

prabhanshuguptagit: Same, it was a bit painful to write this 😅

app/exporters/patients_exporter.rb
prabhanshuguptagit: Yeah, actually we would be doing this for all the columns but it doesn't matter there really.

app/exporters/patients_exporter.rb
harimohanraj89: I imagined reworking this exporter to use some sort of smart data structure like:

```ruby
{
  zone: {
    header: "Patient Zone",
    extractor: ->(patient) { patient.address.zone }
  },
  ...
}
```

but seemed like too much indirection. Sticking the zone in there manually is very KISS.

prabhanshuguptagit: (don't tell anyone but I tried out a whole refactor using a hash map 😬  ended up deleting and came back to this)

PR #1036, opened 2020-06-03
Remove calls column in cohort report

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/story/show/172887815

## Because

We want to remove the `Patients called` column from cohort reports and use that space to add a couple of extra months to the registrations and follow ups columns.

## This addresses

This removes the column and shows 6 periods in registrations and followups instead of 3.

Old:

image

image

New:

image

image


app/views/shared/_analytics_table.html.erb
```diff
@@ -103,29 +92,22 @@
             </td>
 
             <!--Follow-up patients-->
-            <% dates_for_periods(@period, 3, include_current_period: @show_current_period).each do |date| %>
+            <% dates_for_periods(@period, 6, include_current_period: @show_current_period).each do |date| %>
```
harimohanraj89, line 95: We're replacing 9 columns with 12. Can we make sure the new layout isn't too squished on desktop _and_ mobile? PS could you add some screenshots to the PR description?

app/views/shared/_analytics_table.html.erb
```diff
@@ -9,44 +9,40 @@
...
+        <col>
         <col>
         <col>
         <col class="table-divider">
         <col>
         <col>
+        <col>
+        <col>
+        <col>
+        <col class="table-divider">
```
harimohanraj89, line 21: Is this one extra? We should have 3 new cols right?

app/views/shared/_analytics_table.html.erb
```diff
@@ -9,44 +9,40 @@
...
-        <th class="row-label" colspan="3"><strong>Follow-up patients</strong></th>
-        <th class="row-label" colspan="3"><strong>New registrations</strong></th>
-        <th class="row-label" colspan="3"><strong>Patients called</strong></th>
+        <th class="row-label" colspan="6"><strong>Follow-up patients</strong></th>
+        <th class="row-label" colspan="6"><strong>New registrations</strong></th>
       </tr>
 
       <tr class="sorts" data-sort-method="thead">
-        <th class="row-label sort-label" style="text-align: left; padding: 8px 4px;" data-sort-default>Name</th>
+        <th class="row-label sort-label" style="text-align: left; padding: 8px 4px; min-width: 40px;" data-sort-default>Name</th>
```
harimohanraj89, line 32: It may be time to extract these styles to a CSS file.

app/models/facility.rb
```diff
@@ -95,8 +95,7 @@ def dashboard_analytics(period: :month, prev_periods: 3, include_current_period:
     results = [
       query.registered_patients_by_period,
       query.total_registered_patients,
-      query.follow_up_patients_by_period,
-      query.total_calls_made_by_period
```
harimohanraj89, line 99: Can the `total_calls_made_by_period` method and associate specs be removed now?

app/views/shared/_analytics_table.html.erb
prabhanshuguptagit: Added. The table is horizontal scrollable so the columns don't get squished. I did add a small `min-width` to make them look a little nicer.

app/models/facility.rb
prabhanshuguptagit: The column is still being displayed in the whatsapp graphics. This method is being called there.

app/views/shared/_analytics_table.html.erb
harimohanraj89: Cool. Can we run this by the design team?

app/models/facility.rb
harimohanraj89: Should it be removed from the whatsapp graphics as well? Let's check with T/D.

app/views/shared/_analytics_table.html.erb
prabhanshuguptagit: Thanks for catching!

prabhanshuguptagit:
Got the design approved @harimohanraj89. Arigato for the⚡ quick review.

PR #1037, opened 2020-06-03
Search patients by Address

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/173154679

## Because

We want to validate how clean and searchable our village data is.

## This addresses

Add a hidden link with ~`view_adherence_follow_up_list`~ `view_overdue_list` permission at `/patients/lookup`. I really wanted to avoid creating a new permission for this because the fate of this page is unknown and is likely to disappear.

#### Note

Since this is relatively throwaway work, I've avoided writing tests and just click-tested it all around and kept it all fairly segregated for removal if necessary.

There's a couple of unrelated whitespace-only changes which are difficult to revert given my editor settings. Please apply the `Hide whitespace changes` setting whilst reviewing.


app/models/patient.rb
```diff
@@ -50,7 +50,11 @@ class Patient < ApplicationRecord
 
   has_many :current_prescription_drugs, -> { where(is_deleted: false) }, class_name: 'PrescriptionDrug'
 
+  scope :search_by_address,
+    ->(term) { joins(:address).merge(Address.search_by_street_or_village(term)) }
+
```
harimohanraj89, line 55: I know it's throwaway, but can we put this with the other scopes? 😄

app/controllers/patients_controller.rb
```diff
@@ -18,10 +19,33 @@ def index
...
+  # view_overdue_list permission to avoid making a hyper-specific permission for this change.
+  def lookup
+    set_page
+    set_per_page
+    set_facility_id
+    authorize([:overdue_list, Patient])
+
+    if current_facility
+      @patients =
+        paginate(policy_scope([:overdue_list, Patient])
```
harimohanraj89, line 35: Does this policy scope point to the new policy scope you've created inside `OverdueList::PatientPolicy`?

app/views/patients/lookup.html.erb
```diff
@@ -0,0 +1,68 @@
+<h1>Lookup Patients</h1>
+
+<p class="info">
+  <i class="fas fa-question-circle"></i>
+  This is a hidden page and is only visible to you because you have the
+  <em><%= Permissions::ALL_PERMISSIONS.dig(:view_overdue_list, :slug) %></em>
+  permission.
```
harimohanraj89, line 7: Who is this messaging for? If it's for our alpha users (a few CVHOs), should it instead describe something like "Alpha", " You can search for your patients by village and street here", "Work in Progress", etc? Let's ask someone about the right message for this bit.

app/models/patient.rb
kitallis: Ah, yes, this wasn't intentional. Thanks.

app/controllers/patients_controller.rb
kitallis: It is, it's some pundit black-magic that I don't quite understand, but that's how it's supposed to work.

kitallis: It magically picks up the `class` and method name and looks it up in the `Policy` class, probably some metaprogramming magic behind-the-scenes.

harimohanraj89: Hehe I'm fine with the magic, just making sure it was grabbing the scope I _thought_ it was. 👍

app/views/patients/lookup.html.erb
kitallis: Consulted with Vineet and modified it a bit.

PR #1038, opened 2020-06-04
Squish / capitalize name fields editable from the dashboard

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/173285287

## Because

Districts are not first-class in Simple. Sometimes users accidentally _create new_ districts by touching the district name while editing a facility.

This could also happen with our other name fields that are editable from the dashboard and are visible to many users.

## This addresses

* [Squish](https://apidock.com/rails/String/squish) and [upcase_first](https://apidock.com/rails/v5.2.3/String/upcase_first) name fields for: `Facility`, `FacilityGroup`, `Protocol` and `Organization` and `district` on `Facility`

#### Note

`upcase_first` is a conservative way of capitalizing stuff without affecting other intentional case-changes like,

```
> "CH Bhandara".upcase_first
=> "CH Bhandara"

> "cH Bhandara".upcase_first
=> "CH Bhandara"

> "ch Bhandara".upcase_first
=> "Ch Bhandara"
```


Gemfile
```diff
@@ -9,6 +9,7 @@ end
 
 gem "active_record_union"
 gem "activerecord-import"
+gem "auto_strip_attributes"
```
harimohanraj89, line 12: Do we need a gem for this? Can we simply define `def facility=()`?

kitallis: I think so, I'd also tried that and it works. The only difference between this gem and setters is that:

* It does a `before_save` hook rather than an attribute set.
* You can easily write a whole bunch of generic filters across many attributes (like facility name, org name, protocol name etc.).

I'm totally fine with doing it either way. Just felt that there was some value to this really tiny gem.

harimohanraj89: I'm actually a fan of overriding the attribute set 😅 But I can see how we can apply the "titleize" filter across several attributes easily. To avoid YAGNI though, I'd like to set up the simple attr method, and come straight back to this gem if we need to do this across several attributes. What do you think? Less tooling for us to wrap our heads around :)

kitallis: @harimohanraj89 I added some more basic squishes to more editable fields within this PR, because why not...

harimohanraj89:
Oops, I preemptively pressed the wrong button. Apologies.

The test failures in this PR are annoying. They make me wonder if we should introduce this sanitization in the controller instead. If the problem is human error only, then we should target the human input?

kitallis:
> Oops, I preemptively pressed the wrong button. Apologies.
> 
> The test failures in this PR are annoying. They make me wonder if we should introduce this sanitization in the controller instead. If the problem is human error only, then we should target the human input?

@harimohanraj89 

I considered that, generally I would do these kinds of transformations in the handler layer. But I figured the rails-way™ is to move all general-purpose validations and data transformations into the model layer (virtual attributes, setters, validations etc.). So this didn't seem so bad.

This is totally doable by some kind of module that transforms input params and gets injected into controllers before being passed to the model-layer. But I'm torn. This is another one of those things where the API behaves differently from the data-model / factory-layer. I'm ok with doing this in the Model layer, but willing to change my mind.

Re: tests – they are all feature / UI specs, so even if I did this in the controller, they'd still break because they call into the controller and then render the views.

harimohanraj89:
@kitallis You're absolutely right. The API vs Model inconsistency etc has been too big of a problem, and you've convinced me that the model's the right place to put this 👍 I guess some strategic `squish/titleize` bits in our factories will solve the spec problems holistically?

kitallis:
> @kitallis You're absolutely right. The API vs Model inconsistency etc has been too big of a problem, and you've convinced me that the model's the right place to put this 👍 I guess some strategic `squish/titleize` bits in our factories will solve the spec problems holistically?

Factories just call model lifecycles normally, so don't need to do anything special.

PR #1039, opened 2020-06-04
Wire new dashboard reports with real data

rsanheim:
This wires up the new district dashboard views with live queries.

I decided to not go with the rollup summaries right now, because I want to get these reports live in production so that folks w/ access can review them to see how they look and feel with real data. I've been finding hard to test these with sandbox / seed data, because those datasets vary quite a bit from production data.

NOTE: I realize the code here isn't super great, but I think its a fair trade off since none of will be publicly available yet.  The URLs are all hidden and will only be distributed to select users for early-access testing.

**Definitions Doc:** https://docs.google.com/document/d/1MFRYAuy7pgXFZSH_p0YMLAIyGUldXB0KFYZg2jt8KiY/
**Story card:** https://www.pivotaltracker.com/story/show/171110822

## Because

We want great reports for CVHOs

## This addresses

Wiring up a district level report for metrics and benchmarks for CVHOSs


app/services/district_report_service.rb
```diff
@@ -0,0 +1,90 @@
...
+      formatted_period = time.strftime("%b %Y")
+      key = [@facility.id, time.year.to_s, time.month.to_s]
+
+      period = {cohort_period: :month,
+                registration_month: time.month,
+                registration_year: time.year}
+      monthly_bps = MyFacilities::BloodPressureControlQuery.new(facilities: @facilities, cohort_period: period)
+
+      p [formatted_period, monthly_bps.cohort_controlled_bps.count]
+      controlled_bps = Integer(monthly_bps.cohort_controlled_bps.count || 0)
```
rsanheim, line 38: Does it look like I'm calling this query correctly here to get the previous 12 months of controlled patient counts? I was debugging this for quite a bit on Wednesday, and the only data I got back in my spec was for two months prior from the current date in the spec. I was not able to get any counts returned for months further back from that, even though I'm iterating through previous months and re-running the query for each month. 

I may be missing something obvious -- any ideas?

spec/services/district_report_service_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    facility_group = FactoryBot.create(:facility_group, name: "Darrang")
+    facilities = FactoryBot.create_list(:facility, 5, facility_group: facility_group)
+
+    Timecop.freeze(4.months.ago) do
+      other_patients = create_list(:patient, 2, recorded_at: Time.current, registration_facility: facility, registration_user: user)
+      other_patients.map do |patient|
+        create(:blood_pressure, :under_control, facility: facility, patient: patient, recorded_at: Time.current)
+        create(:blood_pressure, :under_control, facility: facility, patient: patient, recorded_at: Time.current)
+      end
+    end
```
rsanheim, line 74: I would expect to see the above BPs returned as counts for four months ago in the results I was printing below, but I only got April counts, not March or Feb.

app/services/district_report_service.rb
prabhanshuguptagit: @rsanheim The bp control query returns bps recorded in the next two months for patients registered in the `:registration_month`. https://github.com/simpledotorg/simple-server/blob/de0ff1dfb830d00c9a7dad11038d5cfd3afb2d9d/app/queries/my_facilities/blood_pressure_control_query.rb#L111

I tried
```
Timecop.freeze(5.months.ago) do
      other_patients = create_list(:patient, 2, recorded_at: Time.current, registration_facility: facility, registration_user: user)
      other_patients.map do |patient|
        create(:blood_pressure, :under_control, facility: facility, patient: patient, recorded_at: Time.current + 1.month)
      end
    end
```
and got

```
["Jul 2019", 0]
0
["Aug 2019", 0]
0
["Sep 2019", 0]
0
["Oct 2019", 0]
0
["Nov 2019", 0]
0
["Dec 2019", 0]
0
["Jan 2020", 2]
2
["Feb 2020", 0]
0
["Mar 2020", 0]
0
["Apr 2020", 2]
2
["May 2020", 0]
0
["Jun 2020", 0]
0
```
Is that what we expect?

rsanheim: Ohhh I think that was it. So the BP `recorded_at` needs to be _after_ the patient `recorded_at`? What does the patient `recorded_at` mean ?

vkrmis: patient `recorded_at` represents when the patient was first registered in the app.
It's expected that this date would be earlier or the same as when BPs are recorded. When backdated BPs are recorded, we also backdate the patient `recorded_at` if necessary.

kitallis: [C1.1](https://docs.google.com/document/d/1Mo83rTwR_tuuUNnJnWKWlZ49nTzlLxqlrbKYnDj1kLQ/edit#heading=h.acbewxs3lz9z) and C2.1 in this document might help clear some definitions out as well.

app/jobs/blood_pressure_rollup_backfill_job.rb
```diff
@@ -0,0 +1,14 @@
+class BloodPressureRollupBackfillJob < ApplicationJob
```
rsanheim, line 1: This isn't wired up anywhere, and it can go away if we decide these rollup tables aren't worth it.

spec/services/district_report_service_spec.rb
rsanheim: Nevermind, got this figured out.

app/views/dashboard/districts/preview.html.erb
```diff
@@ -1,270 +1,9 @@
...
-            <line x1="21" y1="12" x2="9" y2="12"></line>
-          </svg>
-        </div>
-        <p class="font-sans text-md font-semibold text-red-200">
-          Logout
-        </p>
-      </div>
-    </a>
-  </div>
-</div>
```
rsanheim, line 261: @claudiovallejo just a heads up, I extracted the nav bar to a partial so I can use it from the new layout.

app/models/blood_pressure.rb
```diff
@@ -52,6 +60,50 @@ def to_s
...
+    when 10, 11, 12
+      4
+    else
+      raise ArgumentError
+    end
+  end
+
+  def create_or_update_rollup
+    month = recorded_at.month
+    quarter = quarter_for_month(month)
```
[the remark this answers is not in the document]
rsanheim: 👍

app/services/district_report_service.rb
```diff
@@ -0,0 +1,70 @@
...
+      formatted_period = time.strftime("%b %Y")
+      key = [time.year.to_s, time.month.to_s]
+
+      period = {cohort_period: :month,
+                registration_month: time.month,
+                registration_year: time.year}
+      monthly_bps = MyFacilities::BloodPressureControlQuery.new(facilities: @facilities, cohort_period: period)
+
+      controlled_bps = Integer(monthly_bps.cohort_controlled_bps.count || 0)
+      @data[:controlled_patients][formatted_period] = controlled_bps
```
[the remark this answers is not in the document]
rsanheim: Yeah that seems right, I'll update this.  Thanks.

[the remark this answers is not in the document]
rsanheim: I've updated this to use a query against the latest BP per month view, can you take another look?

rsanheim:
I know this PR has sprawled a bit - if y'all could focus on the District controller and the DistrictReportService, that would be great. Please ignore the Rollup related code for now, my plan is to rip it out before this lands in master. The monthly views are working okay for right now, and we want to see how things perform in production with the the queries we are grabbing for the dashboard.

app/views/dashboard/districts/index.html.erb
```diff
@@ -0,0 +1,14 @@
...
+      <div class="mr-2">
+      <% @districts.each do |district| %>
+        <p class="font-roboto font-bold text-md text-blue-200">
+          <%= link_to district.name, dashboard_district_path(district.slug) %>
+        </p>
+      <% end %>
+      </div>
+    </div>
+  </div>
+</div>
```
harimohanraj89, line 14: Nice.

app/services/district_report_service.rb
```diff
@@ -0,0 +1,83 @@
...
+  def controlled_patients(time)
+    end_range = time.end_of_month
+    mid_range = time.advance(months: -1).end_of_month
+    beg_range = time.advance(months: -2).end_of_month
+    sub_query = LatestBloodPressuresPerPatientPerMonth
+      .select("distinct on (patient_id) *")
+      .under_control
+      .order("patient_id, bp_recorded_at DESC, bp_id")
+      .where(registration_facility_id: facilities)
+      .where("(year = ? AND month = ?) OR (year = ? AND month = ?) OR (year = ? AND month = ?)",
```
harimohanraj89, line 32: You could use `(year, month) IN (?, ?), (?, ?), (?, ?)` if you'd like. Not too big a deal either way.

app/services/district_report_service.rb
```diff
@@ -0,0 +1,83 @@
...
+        beg_range.year.to_s, beg_range.month.to_s,
+        mid_range.year.to_s, mid_range.month.to_s,
```
prabhanshuguptagit, line 34: This query looks good. Doing
```
.where("bp_recorded_at > ?", Time.current.beginning_of_day - 3.months)
```
might be easier here.

app/jobs/blood_pressure_rollup_backfill_job.rb
rsanheim: Yup I closed that out.

app/services/district_report_service.rb
```diff
@@ -0,0 +1,83 @@
...
+        end_range.year.to_s, end_range.month.to_s)
+    LatestBloodPressuresPerPatientPerMonth.from(sub_query, "latest_blood_pressures_per_patient_per_months")
```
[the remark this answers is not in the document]
rsanheim: Because we need to apply a `count` on the `distinct on` query, and those can't be combined into one query using ActiveRecord. 

Its the same pattern we use in the [blood pressure query](https://github.com/simpledotorg/simple-server/blob/faf5afd3232a3bc80e5a9a7829959e71031c8958/app/queries/my_facilities/blood_pressure_control_query.rb#L111-L132) for My Facilities.

app/services/district_report_service.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      @data[:controlled_patients][formatted_period] = controlled_patients(time).count
+      @data[:cumulative_registrations] += registrations.fetch(key, 0).to_i.round
+      @data[:registrations][formatted_period] = @data[:cumulative_registrations]
+    end
+  end
+
+  def compile_cohort_trend_data
+    -1.downto(-4).each do |quarter|
+      date = selected_date.advance(months: quarter * 3)
```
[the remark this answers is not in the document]
rsanheim: Hmm yeah, could be an issue.  I'll look into this.

app/services/district_report_service.rb
rsanheim: I had tried this earlier and was getting some incorrect results.  It may have been an error on my side, but the year / month selection is a bit more clear for what we are doing IMO.

app/services/district_report_service.rb
```diff
@@ -0,0 +1,83 @@
...
+        no_bp: query.cohort_missed_visits_count,
+        uncontrolled: query.cohort_uncontrolled_bps.count
+      }.with_indifferent_access
+    end
+  end
+
+  def registrations
+    registrations_query ||= MyFacilities::RegistrationsQuery.new(facilities: @facilities, period: :month, last_n: 12)
+    registrations_query.registrations.group(:year, :month).sum(:registration_count)
+  end
```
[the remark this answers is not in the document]
rsanheim: sure, done.

app/services/district_report_service.rb
```diff
@@ -0,0 +1,83 @@
...
+      @data[:registrations][formatted_period] = @data[:cumulative_registrations]
+    end
+  end
+
+  def compile_cohort_trend_data
+    -1.downto(-4).each do |quarter|
+      date = selected_date.advance(months: quarter * 3)
+      quarter = QuarterHelper.quarter(date)
+      year = date.year
+      next_quarter = QuarterHelper.next_year_and_quarter(year, quarter)
```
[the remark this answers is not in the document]
rsanheim: 👍

app/services/district_report_service.rb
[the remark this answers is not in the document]
harimohanraj89: Ah this. 😂 😭 As you were...

app/services/district_report_service.rb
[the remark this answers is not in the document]
rsanheim: This is all cleaned up now, and will select the quarters correctly. I believe it was also correct before, but it wasn't clear at all from the code 😁 . I introduced a `Quarter` object to make things much more clear, and overtime we can refactor to use that object instead of the static methods on QuarterHelper.

app/services/district_report_service.rb
```diff
@@ -0,0 +1,83 @@
...
+      formatted_period = time.strftime("%b %Y")
+      key = [time.year.to_s, time.month.to_s]
+
+      period = {cohort_period: :month,
+                registration_month: time.month,
+                registration_year: time.year}
+
+      @data[:controlled_patients][formatted_period] = controlled_patients(time).count
+      @data[:cumulative_registrations] += registrations.fetch(key, 0).to_i.round
+      @data[:registrations][formatted_period] = @data[:cumulative_registrations]
```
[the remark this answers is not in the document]
rsanheim: Need to look into this on Thursday, was fighting some test data things around this.

rsanheim:
This is ready for another pass. There is one remaining TODO from @harimohanraj89 that I'll look into Thursady, but beyond that I think this is ready to go.

app/controllers/api/v3/sync_controller.rb
```diff
@@ -31,6 +30,20 @@ def __sync_to_user__(response_key)
...
+      result
+    }
+  end
+
+  # This is a hook method for subclasses that need to operate on the results after the merge but before the response
+  # is sent back to the client.
+  # results will be an array of Hashes, with the key values of either { :record => [Model] } or { :errors_hash => Hash }
+  def after_merge_completed(results)
+  end
+
```
harimohanraj89, line 46: We can remove all the changes in this controller and the `after_merge_completed` template method as well, right? That was only needed for rollups?

app/models/quarter.rb
```diff
@@ -0,0 +1,39 @@
+class Quarter
+  def self.create(date:)
+    new(date: date)
+  end
```
harimohanraj89, line 4: Do we need this `create` method? Could we just use `new`?

app/models/quarter.rb
```diff
@@ -0,0 +1,39 @@
...
+    (1..number).inject([self]) do |quarters, number|
+      quarters << quarters.last.previous_quarter
+    end
+  end
+
+  def upto(number)
+    (1..number).inject([self]) do |quarters, number|
+      quarters << quarters.last.next_quarter
+    end
+  end
```
harimohanraj89, line 38: So, `Quarter.current.upto(5)` will give me `[Q2 2020, Q3 2020, Q4 2020, Q1 2021, Q2 2021]`? Neato.

app/services/district_report_service.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def format_quarter(quarter)
+    "Q#{quarter.number}-#{quarter.year}"
+  end
+
+  # We want to return cohort data for the current quarter for the selected date, and then
+  # the previous three quarters. Each quarter cohort is made up of patients registered
+  # in the previous quarter who has had a follow up visit in the current quarter.
+  def compile_cohort_trend_data
+    Quarter.create(date: selected_date).downto(3).each do |results_quarter|
```
harimohanraj89, line 43: Love it, so semantic.

harimohanraj89:
Screenshot 2020-06-18 at 2 43 06 PM

Looks awesome! Couple notes:
* I tore down and re-setup your review app (it was a little old so db write access was revoked because of row limits)
* I had to run `heroku run rake refresh_materialized_db_views` on your review app. Not sure if that's related to some of the test data issues you were mentioning.
* Non-blocking: We could consider how to handle empty cohorts in upcoming PRs
* Non-blocking: The hard-coded benchmark numbers - I think we're doing the subtraction wrong? If your number is higher than the reference (Bathinda), you should get a positive diff in green, yes?

app/models/quarter.rb
rsanheim: Yup, I kept it inclusive of the quarter your are stepping from, which is how [downto](https://apidock.com/ruby/Integer/downto) and [upto](https://apidock.com/ruby/Integer/upto) work in Ruby.

app/models/quarter.rb
rsanheim: Yup removed that unnecessary create method.

app/services/district_report_service.rb
```diff
@@ -0,0 +1,98 @@
+class DistrictReportService
+  def initialize(facilities:, selected_date:)
+    @facilities = Array(facilities)
+    @selected_date = selected_date
+    @data = {
+      controlled_patients: {},
+      registrations: {},
+      cumulative_registrations: 0,
```
rsanheim, line 8: TODO fix this, need to grab cumulative from before.

app/assets/stylesheets/reports.scss
```diff
@@ -0,0 +1,7 @@
+@import "tailwindcss/base";
+@import "tailwindcss/components";
+@import "tailwindcss/utilities";
+
+// @tailwind base;
```
timcheadle, line 5: Can we ditch these comments?

app/controllers/dashboard/districts_controller.rb
```diff
@@ -5,22 +5,43 @@ class Dashboard::DistrictsController < AdminController
 
   EXAMPLE_DATA_FILE = "db/data/example_dashboard_data.json"
 
+  def index
+    authorize :dashboard, :view_my_facilities?
+    @districts = FacilityGroup.all.order(:name)
```
timcheadle, line 10: This should use some sort of policy scope to only show what a user has access to

app/helpers/application_helper.rb
```diff
@@ -35,8 +35,12 @@ def rounded_time_ago_in_words(date)
   # Defaults to precision of 0.
   def compute_percentage(numerator, denominator, options = {})
     options = options.with_defaults(precision: 0)
```
timcheadle, line 37: Can move this inside the `else` block

app/models/quarter.rb
timcheadle: LOVE this

app/services/district_report_service.rb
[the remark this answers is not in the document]
timcheadle: As mentioned, I think we can do something easier here, e.g.:
```ruby
Patient.where(registration_facility: @facilities).where('recorded_at <= ?', date).count
```

app/services/district_report_service.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def format_quarter(quarter)
+    "Q#{quarter.number}-#{quarter.year}"
+  end
+
+  # We want to return cohort data for the current quarter for the selected date, and then
+  # the previous three quarters. Each quarter cohort is made up of patients registered
+  # in the previous quarter who has had a follow up visit in the current quarter.
+  def compile_cohort_trend_data
```
timcheadle, line 52: Not sure this code is necessary; see `CohortAnalyticsQuery` and `Analytics::DistrictsController` for how we compute these in the old dashboards

app/services/district_report_service.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    compile_cohort_trend_data
+
+    data
+  end
+
+  def compile_control_and_registration_data
+    beginning_of_period = selected_date.advance(months: -11)
+    previous_registrations = 0
+    registrations.each do |(period, count)|
```
kitallis, line 26: This a bit complicated to understand, from what I gather from the logic, could this be simply achieved by the following:

```
previous_registrations =
  PatientRegistrationsPerDayPerFacility
    .where(facility: @facilities)
    .group(:year, :month)
    .where('year < ?', beginning_of_period.year)
    .or(PatientRegistrationsPerDayPerFacility
          .where('month <= ?', beginning_of_period.month))
    .sum(:registration_count)
```

config/initializers/time_and_date_formats.rb
```diff
@@ -0,0 +1,2 @@
+Time::DATE_FORMATS[:month_year] = "%b %Y"
```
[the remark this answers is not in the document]
rsanheim: This adds a new format we can use for `to_s` with time / date via `time.to_s(:month_year)`.  It doesn't change the default `to_s` behavior.  Its a handy Rails thing for custom formats, check out https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/time/conversions.rb.

app/services/district_report_service.rb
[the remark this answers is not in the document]
rsanheim: Yup, refactored to this query. Much simpler now.

app/controllers/dashboard/districts_controller.rb
rsanheim: I cargo culted a scope from another controller, I think its good for this first PR. I'll need to work with someone w/ more experience w/ Pundit to make sure we have this nailed down later.

PR #1041, opened 2020-06-05
Cohort download tweaks

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/173188739

## Because

Some missing and clunky bits of info in our district cohort report downloads were making analysis difficult

## This addresses

Tweaks the district cohort report downloads to:
* Add facility type as a new column
* Remove the % signs from values, and add them to the headers instead.


app/views/analytics/districts/show.csv.erb
```diff
@@ -13,21 +13,21 @@
     'Patients controlled',
     'Patients uncontrolled',
     "Patients didn't attend",
-    'Control rate',
-    'Uncontrolled rate',
-    'Default rate',
+    '% BP Control %',
+    '% BP Uncontrolled',
+    '% Defaulted %',
```
prabhanshuguptagit, line 18: `%` typo 👾

app/views/analytics/districts/show.csv.erb
```diff
@@ -88,6 +89,7 @@ end).to_csv -%>
     <%=
```
prabhanshuguptagit, line 89: image

The `All` row is missing a `nil` value under facility type.

app/views/analytics/districts/show.csv.erb
```diff
@@ -65,6 +65,7 @@ end).to_csv -%>
 <%
   headers = [
     "Facility name",
+    "Facility type",
```
prabhanshuguptagit, line 68: nitpick of the day: `Facility Type`?

app/controllers/analytics/districts_controller.rb
```diff
@@ -84,9 +84,21 @@ def set_dashboard_analytics(period, prev_periods)
   end
 
   def set_facility_keys
-    facilities = @organization_district.facilities.order(:name).pluck(:id, :name).to_h
-
-    @facility_keys = { total: "Total" }.merge(facilities).with_indifferent_access
+    district = {
+      id: :total,
+      name: "Total",
+      type: "District"
```
prabhanshuguptagit, line 90: image

The `Facility Type` : `District` column feels slightly confusing, can we maybe say **`Total - District`** and not have that column in the totals?

app/views/analytics/districts/show.csv.erb
harimohanraj89: In fact, changed the other one to `Facility type` for consistency instead.

app/controllers/analytics/districts_controller.rb
harimohanraj89: I'll just stick a `nil` in there. The fact that it says `<District> Cohort Report` right above makes it clear what the total is 👍

harimohanraj89:
Hey @prabhanshuguptagit, thanks for the feedback. Addressed in 241dbef. Here's what the download looks like now:

Screenshot 2020-06-08 at 2 55 57 PM

app/controllers/analytics/districts_controller.rb
harimohanraj89: There was also a bug where the facilities' types weren't being printed in there. Fixed now.

app/views/analytics/districts/show.csv.erb
```diff
@@ -13,21 +13,21 @@
     'Patients controlled',
     'Patients uncontrolled',
     "Patients didn't attend",
-    'Control rate',
-    'Uncontrolled rate',
-    'Default rate',
+    '% BP Control %',
```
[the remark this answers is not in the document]
harimohanraj89: I have forgotten how to read 🤦

app/controllers/analytics/districts_controller.rb
prabhanshuguptagit: Ah, ok. That's what the empty cells were about. :shipit:

PR #1042, opened 2020-06-05
Add Flipper

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/173192671

## Because

We want to be able to manage toggling of features in a more convenient and granular fashion.

## This addresses

This PR adds the Flipper gem and associated tooling to start managing feature flags.

### Key Decisions

* **Adapter:** I've chosen to use the ActiveRecord adapter, which creates and manages a couple of `flipper_` database tables.
* **UI:** I've mounted the Flipper UI at `/flipper` behind a new `view_flipper_ui` permission. Like Sidekiq, I have not added this to the superadmin role, and plan to provision this permission manually to users via console.
* **Permissions:** I've chosen not to run our `add_permission_to_role` script because all of these system permissions will be taken care of when we add that `superadmin: true` flag to users that short-circuits our permissions.
* **Navigation:** I've added Sidekiq and Flipper items to the "Manage" section of the dashboard navigation.

#### Features Dashboard
Screenshot 2020-06-05 at 4 32 02 PM

#### Feature Editor
Screenshot 2020-06-05 at 4 31 55 PM

#### New Navigation Items
Screenshot 2020-06-05 at 4 45 15 PM

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,7 @@ gem "factory_bot_rails", "~> 4.8", require: false
 gem "faker", require: false
 gem "flipper"
 gem "flipper-active_record"
+gem "flipper-ui"
 gem "friendly_id", "~> 5.2.4"
 gem "groupdate"
 gem "http"
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ gem "dotenv-rails"
 gem "factory_bot_rails", "~> 4.8", require: false
 gem "faker", require: false
 gem "flipper"
+gem "flipper-active_record"
 gem "friendly_id", "~> 5.2.4"
 gem "groupdate"
 gem "http"
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem "discard", "~> 1.0"
 gem "dotenv-rails"
 gem "factory_bot_rails", "~> 4.8", require: false
 gem "faker", require: false
+gem "flipper"
 gem "friendly_id", "~> 5.2.4"
 gem "groupdate"
 gem "http"
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -105,6 +105,12 @@
                   <% if policy([:dashboard]).manage_users? %>
                     <%= link_to "Users", admin_users_path, class: "dropdown-item #{active_controller?("users")}" %>
                   <% end %>
+                  <% if policy([:dashboard]).view_sidekiq_ui? %>
+                    <%= link_to "Sidekiq", "/sidekiq", class: "dropdown-item #{active_controller?("users")}" %>
+                  <% end %>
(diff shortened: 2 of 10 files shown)


app/views/layouts/application.html.erb
```diff
@@ -105,6 +105,12 @@
                   <% if policy([:dashboard]).manage_users? %>
                     <%= link_to "Users", admin_users_path, class: "dropdown-item #{active_controller?("users")}" %>
                   <% end %>
+                  <% if policy([:dashboard]).view_sidekiq_ui? %>
+                    <%= link_to "Sidekiq", "/sidekiq", class: "dropdown-item #{active_controller?("users")}" %>
```
prabhanshuguptagit, line 109: 🙌

Gemfile
```diff
@@ -22,6 +22,9 @@ gem "discard", "~> 1.0"
 gem "dotenv-rails"
 gem "factory_bot_rails", "~> 4.8", require: false
 gem "faker", require: false
+gem "flipper"
```
prabhanshuguptagit, line 25: If I get this right, `FeatureToggled.enabled?` becomes `Flipper.enabled?` everywhere?

harimohanraj89: Yup. Once we ship this, we can go create Flipper flags for all our feature flags, and then open a PR that migrates to using `Flipper.enabled?`. Any new feature flags we introduce in the meantime can start using Flipper straight away.

We can also start doing things like `Flipper.enabled?(feature, current_user)`, ie, target specific user groups to soft-launch features to.

kitallis:
@harimohanraj89 I assume with this, we can slowly phase out the original `FeatureFlag` module right? We don't need to migrate wholesale?

In which case, let's keep an eye out for the mechanism we employ in `standalone` to merge feature flags into the env config, because as we migrate a feature, we'd want to deprecate its usage from the config file.

harimohanraj89:
Yup, we can phase them out slowly. The env var feature flags will just lie around unused, and we can catch em in a periodic cleanup. Once every flag is cleaned up, we can remove the "merge feature flags into config" functionality in `standalone` altogether.

rsanheim:
> * **UI:** I've mounted the Flipper UI at `/flipper` behind a new `view_flipper_ui` permission. Like Sidekiq, I have not added this to the superadmin role, and plan to provision this permission manually to users via console.
> 
> * **Permissions:** I've chosen not to run our `add_permission_to_role` script because all of these system permissions will be taken care of when we add that `superadmin: true` flag to users that short-circuits our permissions.

I'm confused about how this is supposed to work for all our non-production environments? Developers acting as super-admins will need to be able to manage flipper in dev as well as sandbox/staging, but it seems like there isn't an easy way to add flipper rights to my account in dev without running a console script.

Also, I may be out of the loop around the `superadmin` flag, can you point me towards anything explaining that?

PR #1071, opened 2020-06-11
API translations

harimohanraj89:
https://www.pivotaltracker.com/story/show/173350287

Another batch of API translations.

**To Do**

* [x] Fix top-level language keys


prabhanshuguptagit:
@harimohanraj89 There's a `pa_Guru_IN` in `locales/` just like the `mr_IN` translation, shall we get rid of it as well while we're at this?

PR #1074, opened 2020-06-15
standardrb --fix for models and specs

rsanheim:
**Story card:** n/a, just a quick standardrb autofix

## Because

We want standardized formatting / linting

## This addresses

```
standardrb --fix app/models/**.rb
standardrb --fix spec/models/**.rb
```


app/models/concerns/hashable.rb
```diff
@@ -1,11 +1,9 @@
 module Hashable
-  extend ActiveSupport::Concern
+  module_function
```
harimohanraj89, line 2: TIL `module_function`. Nice.

rsanheim: Nope, it applies to all the methods defined after it when used this way.

PR #1082, opened 2020-06-16
Bulk enable Diabetes from FacilityGroups

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/173374470

## Because

Users want to be able to enable diabetes management for all facilities in their facility group at once. Currently they need to check the flag on each facility's edit page.

## This addresses

This adds a checkbox on the FacilityGroup edit page that enables diabetes management for all associated facilities.

image


app/models/facility_group.rb
```diff
@@ -24,8 +24,26 @@ class FacilityGroup < ApplicationRecord
...
   end
+
+  def diabetes_enabled?
+    facilities.where(enable_diabetes_management: false).count.zero?
+  end
+
+  private
+
+  def enable_diabetes_management!
+    facilities.update(enable_diabetes_management: true)
```
harimohanraj89, line 43: Do any validation errors here propagate back up to the FacilityGroup? If this update fails, how does the user know it failed on the FG edit screen?

app/models/facility_group.rb
```diff
@@ -24,8 +24,26 @@ class FacilityGroup < ApplicationRecord
   friendly_id :name, use: :slugged
 
   auto_strip_attributes :name, squish: true, upcase_first: true
+  attribute :enable_diabetes_management
+
+  before_save :enable_diabetes_management!, if: -> { enable_diabetes_management }
+  before_save :disable_diabetes_management!, if: -> { !enable_diabetes_management && diabetes_enabled? }
```
harimohanraj89, line 30: Putting this in the model layer may be risky. What happens when a facility group that is fully DM-enabled, is, say, edited through the console or a data migration? Would `facility_group.touch` disable DM in all of its facilities?

prabhanshuguptagit: That is correct, doing this on console for example:
```
facility_group.update(name: "name")
```
will set all facilities to `false`. Replacing this with a regular method 😕 , easier to wrap heads around.

harimohanraj89: Neat. Yup, a couple extra lines in the controller should be ok 👍

app/models/facility_group.rb
prabhanshuguptagit: Validation errors won't propagate from here but since we are explicitly setting it to `true` or `false`, validation shouldn't be an issue. If the update fails for other reasons it will throw an exception, which should be ok.

app/views/admin/facility_groups/_form.html.erb
```diff
@@ -19,9 +19,23 @@
...
 
       <%= form.primary("Save facility group") %>
     <% end %>
   </div>
 </div>
+<script>
+  function confirmDisableDM(el) {
+    if(!el.checked)
+        confirm("Are you sure? This will disable diabetes management for all associated facilities.")
+  }
```
harimohanraj89, line 40: We should only fire this confirmation if `facility_group.diabetes_enabled?`, right? If only a few facilities have DM enabled, the box will be unchecked on the edit screen, and when I save, I shouldn't see any confirmation because we won't change the DM status of any facilities.

app/models/facility_group.rb
kitallis: Looking into wrapping this up now. Seeing if I can propagate child errors to the parent. This kind of stuff is very annoying in rails :|

app/views/admin/facility_groups/_form.html.erb
kitallis: We might be ditching this JS in favor of not allowing to disable DM at all. Letting Vineet and Praveen discuss it until then. Regardless, this fix has been made even if we keep it.

app/models/facility_group.rb
kitallis: This is done:

image

app/models/facility_group.rb
kitallis: This is done:

image

PR #1083, opened 2020-06-17
Allow multiple teleconsultation numbers per facility

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/173372046

## Because

Facilities need backup Medical Officers for teleconsultation

## This addresses

Adds a JSON column to facilities to store multiple phone numbers.
Adds a `teleconsultation_phone_numbers` field that can contain multiple phone numbers to the `GET /api/v4/facility_teleconsultations/{facility_id}` api endpoint response.

## Hacks
Some hacks were involved in getting this to work:

- Explicitly modeled TeleconsultationPhoneNumber in Facility. This can probably become a real, generic model eventually. Creating accessors for the json field `teleconsultation_phone_number` allows this to work seamlessly.

- Couldn't get form errors for the phone number fields to propagate correctly. Added a facility level validator that validates all the phone number fields, and manually handled error-rendering for this part of the form.

- Used js cloneNode to add form fields for extra numbers


app/assets/javascripts/teleconsultation-fields.js
```diff
@@ -11,3 +17,50 @@ function toggleTeleconsultationFields(checkbox) {
...
+
+
+    button.removeAttribute('disabled')
+
+    $MORows().appendChild(clone)
+}
+
+function deletePhoneNumber(button) {
+    // Sets the _destroy field's value to true
+    // Disables form inputs for the deleted row
```
harimohanraj89, line 56: If we can't undo the deletion, then we should just hide the whole row, instead of disabling.

app/schema/api/v4/schema.rb
```diff
@@ -141,7 +141,10 @@ def user_me_response
 
     def facility_teleconsultations_response
       { type: :object,
-        properties: { teleconsultation_phone_number: { type: [:string, 'null'] } },
+        properties: { teleconsultation_phone_number: { type: [:string, 'null'] },
+                      teleconsultation_phone_numbers: { type: :array,
+                                                        items: { type: :object,
+                                                                 properties: { phone_number: { type: :string}}}}},
```
harimohanraj89, line 147: The `teleconsultation_phone_numbers` key should be required, right? It will always be present. Same for the `phone_number` property inside. When no telemed info is available, the API should return an empty array.

`teleconsultation_phone_numbers: []`

app/models/facility.rb
```diff
@@ -179,9 +179,16 @@ def teleconsultation_enabled?
...
+    teleconsultation_phone_number = teleconsultation_phone_numbers.first
+    return if teleconsultation_phone_number.blank?
+
+    Phonelib.parse(teleconsultation_phone_number.isd_code + teleconsultation_phone_number.phone_number).full_e164
+  end
 
-    Phonelib.parse(teleconsultation_isd_code + teleconsultation_phone_number).full_e164
+  def teleconsultation_phone_numbers_with_isd
+    teleconsultation_phone_numbers.map do |phone_number|
+      { phone_number: Phonelib.parse(phone_number.isd_code + phone_number.phone_number).full_e164 }
```
harimohanraj89, line 190: Love the `phone_number` key. Makes it easy to extend this API payload with more info later like Doctor name, serial numbers, other things.

app/models/facility.rb
```diff
@@ -179,9 +179,16 @@ def teleconsultation_enabled?
...
   def teleconsultation_phone_number_with_isd
-    return if teleconsultation_isd_code.blank? || teleconsultation_phone_number.blank?
+    teleconsultation_phone_number = teleconsultation_phone_numbers.first
+    return if teleconsultation_phone_number.blank?
+
+    Phonelib.parse(teleconsultation_phone_number.isd_code + teleconsultation_phone_number.phone_number).full_e164
+  end
 
-    Phonelib.parse(teleconsultation_isd_code + teleconsultation_phone_number).full_e164
+  def teleconsultation_phone_numbers_with_isd
```
harimohanraj89, line 188: There are several new behaviors on the Facility model, could we add test coverage for them?

app/assets/javascripts/teleconsultation-fields.js
```diff
@@ -2,6 +2,12 @@ function $teleconsultationFields() {
     return document.getElementById('teleconsultation_fields');
 }
 
+function $MORows() {
```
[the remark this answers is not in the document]
vkrmis: We do have jquery available, but we've been trying to use vanilla js where possible

app/controllers/admin/facilities_controller.rb
```diff
@@ -110,7 +110,8 @@ def facility_params
       :zone,
       :enable_teleconsultation,
       :teleconsultation_phone_number,
-      :teleconsultation_isd_code
+      :teleconsultation_isd_code,
+      teleconsultation_phone_numbers_attributes: [:isd_code, :phone_number, :_destroy]
```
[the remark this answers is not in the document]
vkrmis: Yes, the `_destroy` field is checked inside `teleconsultation_phone_numbers_attributes` to determine whether or not it should be saved.

app/assets/javascripts/teleconsultation-fields.js
vkrmis: I didn't want to do that because the row isn't actually deleted until the form is submitted. 
Don't want people to assume the phone number has been deleted because it disappeared.

app/models/facility.rb
```diff
@@ -192,9 +199,59 @@ def teleconsultation_phone_number_with_isd
...
+    numbers << TeleconsultationPhoneNumber.new({})
+    self.teleconsultation_phone_numbers = numbers
+  end
+
+  class TeleconsultationPhoneNumber
+    include ActiveModel::Validations
+    attr_accessor :isd_code, :phone_number
+
+    def initialize(hash)
+      @isd_code = hash['isd_code'] || Rails.application.config.country['sms_country_code']
```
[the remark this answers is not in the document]
vkrmis: No, the form validations will catch this. I'll try to find a different place to put this default.

app/assets/javascripts/teleconsultation-fields.js
```diff
@@ -11,3 +17,50 @@ function toggleTeleconsultationFields(checkbox) {
...
+    let clone = elem.cloneNode(true)
+    let msec = Date.now()
+    let isd_field = clone.querySelector('input[name="facility[teleconsultation_phone_numbers_attributes][0][isd_code]"]')
+    let phone_number_field = clone.querySelector('input[name="facility[teleconsultation_phone_numbers_attributes][0][phone_number]"]')
+    let delete_field = clone.querySelector('input[name="facility[teleconsultation_phone_numbers_attributes][0][_destroy]"]')
+    let button = clone.querySelector('button[name="delete_button"]')
+
+    clone.id = "mo_row_" + msec
+
+    isd_field.id = "facility[teleconsultation_phone_numbers_attributes][" + msec + "][isd_code]"
```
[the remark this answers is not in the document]
vkrmis: clone.id has a "mo_row_" string prefix. I think this needs to be a numeric index for the payload to be assembled properly.

app/models/facility.rb
```diff
@@ -192,9 +199,59 @@ def teleconsultation_phone_number_with_isd
...
+  def teleconsultation_phone_numbers_valid?
+    teleconsultation_phone_numbers.each_with_index do |mo, index|
+      if mo.isd_code.blank?
+        errors.add("teleconsultation_phone_numbers_attributes_#{index}_isd_code", "can't be blank")
+      end
+      if mo.phone_number.blank?
+        errors.add("teleconsultation_phone_numbers_attributes_#{index}_phone_number", "can't be blank")
+      end
+    end
+  end
```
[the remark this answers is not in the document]
vkrmis: The model validations are useless, couldn't get that to work. Will get rid of them. Still need the TeleconsultationPhoneNumber class so we can do fields_for and accept input for the isd_code and phone_number fields. It isn't an active model though.

app/assets/javascripts/teleconsultation-fields.js
```diff
@@ -11,3 +17,50 @@ function toggleTeleconsultationFields(checkbox) {
         teleconsultationFields.style.display = 'none';
     }
 }
+
+function addPhoneNumberFields() {
+    // Clones the first phone number fieldset, and resets it to an empty form state.
```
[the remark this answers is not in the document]
vkrmis: Considered doing that but decided this was cleaner. Will still have to reset the index for the hidden fieldset anyway, and make sure it doesn't get submitted with the rest of the form.

app/assets/javascripts/teleconsultation-fields.js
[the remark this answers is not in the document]
harimohanraj89: Oh I see, it uses the weird HTML array thingy 👍

app/assets/javascripts/teleconsultation-fields.js
harimohanraj89: Great point 👍

app/controllers/admin/facilities_controller.rb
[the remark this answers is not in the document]
vkrmis: Since we aren't using any magic here, shall we leave this as is? It's just a hidden field that lets us skip deleted rows when saving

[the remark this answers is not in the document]
harimohanraj89: Since we cannot remove the fieldst from the page on deletion, as you mentioned below, yes we should keep this as is 👍

app/views/admin/facilities/_form.html.erb
```diff
@@ -1,6 +1,11 @@
 <div class="row">
   <div class="col-md-6">
     <%= bootstrap_form_with(model: [:admin, facility_group, facility], local: true, autocomplete: "off", label_errors: true) do |form| %>
+      <% if @facility.errors.any? %>
+        <div id="errors">
+          <%= form.label "Please fix the indicated errors", class: "text-danger" %>
```
harimohanraj89, line 6: Let's add "Update failed." to the beginning to be super clear.

app/views/admin/facilities/show.html.erb
```diff
@@ -36,18 +36,25 @@
...
     <div class="card">
       <h3>Facility type &amp; size</h3>
       <%= @facility.facility_type || "(No type set)" %> (<%= @facility.facility_size&.capitalize || "No size set" %>)
     </div>
   </div>
+  <div class="col-md-3">
+    <div class="card">
+      <h3>Features</h3>
+      Diabetes Management: <%= @facility.diabetes_enabled? ? "Enabled" : "Disabled"  %>
+      Teleconsultation:    <%= @facility.teleconsultation_enabled? ? "Enabled" : "Disabled" %>
```
harimohanraj89, line 55: Can we add the teleconsult phone numbers to the card too?

app/models/facility.rb
```diff
@@ -179,22 +179,93 @@ def teleconsultation_enabled?
...
+
+  def build_teleconsultation_phone_number
+    numbers = teleconsultation_phone_numbers.dup
+    numbers << TeleconsultationPhoneNumber.new({isd_code: Rails.application.config.country["sms_country_code"],
+                                                phone_number: ""})
+    self.teleconsultation_phone_numbers = numbers
+  end
+
+  class TeleconsultationPhoneNumber
```
harimohanraj89, line 224: This class and its uses are now _just_ for feeding _into_ form elements on the page using `form_for`, right? Do you think we can remove all the custom attribute reading and writing, and add a single method to the facility class, or controller, that's like `TeleconsultationPhoneNumbersForForm` or something? Overall, I'm wondering if we can avoid so much model behavior to power one form in the view layer.

harimohanraj89:
@vkrmis a few more thoughts from me. Most major thought - Can we simplify the virtual model usage like crazy?

app/views/admin/facilities/show.html.erb
vkrmis: Wouldn't look good. This card is already too big. I'll line these cards up evenly, but if we want numbers we should make this a larger card in a different row.

Screenshot 2020-06-17 at 6 28 45 PM

harimohanraj89: Oh, annoying 😓 Ok, let' just stick to the two flags for now, awesome.

Is there a quick flex trick to make the four cards even? 😅 If not, no big deal.

app/models/facility.rb
vkrmis: I've refactored this class to be a struct instead, but I haven't been able to figure out a way to use this without the custom accessors.

The only other way I can think of is to do some client-side form hacking to make this a single field and serialize and deserialize its contents as needed. This will remove the need for the accessors in Facility, and move what they're doing to a separate serializer class.

This will make the form more complex though, since it will have to parse and assemble json for the single field, and we'd still want something like the struct to keep the field clean.

app/validators/admin/csv/facility_validator.rb
```diff
@@ -29,14 +29,22 @@ def duplicate_rows
...
-      import_facility = Facility.new(facility)
+      import_facility = Facility.new(sanitize_facility(facility))
       row_errors << [row_num, import_facility.errors.full_messages.to_sentence] if import_facility.invalid?
     end
     group_row_errors(row_errors).each { |error| @errors << error } if row_errors.present?
   end
 
   private
 
+  def sanitize_facility(facility)
```
harimohanraj89, line 40: This isn't really sanitizing as much as parameter massaging. Can we name this `facility_params` or `facility_attrs` or something?

I'm guessing this is the first case where columns in the import spreadsheet don't map exactly to columns in our db?

vkrmis: Yup

PR #1084, opened 2020-06-17
Add profiling + indexes to a few sync resources

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/173396573

## Because

`sync_to_user` can be demanding on the DB because some queries on NR clearly suggest that some big sync resources take a lot of time querying the DB. [Ref](https://rpm.newrelic.com/accounts/2000380/applications/180429547/transactions#id=5b22436f6e74726f6c6c65722f6170692f76332f626c6f6f645f7072657373757265732f73796e635f746f5f75736572222c22225d).

## This addresses

Adds an index to `updated_at` to various sync resources since it's the primary way to seek the cursor while fetching records. The following most important sync resources will be updated:

* `BloodPressure`
* `Patient`
* `Appointments`
* `BloodSugar`

## Index

They are created concurrently, and on a production-sized database of 6gb, it took ~6s to finish the migration task:

```
± bin/rails db:migrate
== 20200616111304 AddUpdatedAtToSyncResources: migrating ======================
-- add_index(:blood_pressures, :updated_at, {:algorithm=>:concurrently})
   -> 2.7192s
-- add_index(:blood_sugars, :updated_at, {:algorithm=>:concurrently})
   -> 0.0316s
-- add_index(:patients, :updated_at, {:algorithm=>:concurrently})
   -> 0.6786s
-- add_index(:appointments, :updated_at, {:algorithm=>:concurrently})
   -> 2.9340s
== 20200616111304 AddUpdatedAtToSyncResources: migrated (6.3637s) =============
```

And roughly 8-10 additional seconds for the actual concurrent indexes to finish in the background. This is of course with zero traffic locally. But it should be alright on production.

## Profiling

Added a way to profile some API requests using vegeta. Here's some benchmarks on before v. after index creation. Note that these benchmarks were not run by seeking to different sync cursors, all syncs were done from the beginning.

We ran the requests for 5 minutes at 5 requests per second (our production average is 3rps).

**Data-set**

* 1 huge FG (80k patients)
* 1 normally sized FG (2k patients)

**Before index**

I couldn't even get these requests to finish, some of them took so long that the load-balancer was killing the requests off. I removed the large FG (not pictured) and I could get a stable 200-300ms average response time.

Screenshot 2020-06-18 at 4 26 27 PM
Screenshot 2020-06-18 at 4 26 39 PM

**After index**

Even with the big FG, we finish even faster than the best-case scenario without the index (<200ms).

Screenshot 2020-06-18 at 4 25 01 PM

Here we see a spike, but it stabilizes eventually (even the spike is relatively small).
Screenshot 2020-06-18 at 4 25 38 PM

I think we should achieve a marked improvement in `sync_to_user` with this. My estimate would be at least 20-25% speed-up.


profiling/sync_to_user
```diff
@@ -0,0 +1,13 @@
+GET https://api-sandbox.simple.org/api/v3/blood_pressures/sync
+Accept: application/json
+Content-Type: application/json
+Authorization: Bearer aa9fdd9072e8c1726e5382170b2ae815c607d8d6718b223d1b0a9a8c2e7f8f55
+X-USER-ID: 18fa33d3-4f18-476a-8d63-51406352ba13
+X-FACILITY-ID: 36266c5b-490f-490f-8472-b8759aac0ace
```
harimohanraj89, line 6: Are these valid, longstanding SBX creds? Should we perhaps not include em in the repo? If handling these secrets is awkward, we can just XXXXXXX em out and add instructions to populate em into the readme.

profiling/README.md
```diff
@@ -0,0 +1,19 @@
+# Profiling
+
+The performance tests are designed to hit the APIs using the [Vegeta utility](https://github.com/tsenart/vegeta)
+
+## Running the perf test
```
harimohanraj89, line 5: *performance? :)

profiling/README.md
```diff
@@ -0,0 +1,19 @@
+# Profiling
```
harimohanraj89, line 1: Let's add a link to our top-level README to this, in the # Development section.

profiling/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The performance tests are designed to hit the APIs using the [Vegeta utility](https://github.com/tsenart/vegeta)
```
harimohanraj89, line 3: *Sandbox API?

harimohanraj89: I think it's safe to add a note here to use `brew install vegeta` to install it.

kitallis: You can change it to whatever API endpoint you want, the checked in one happens to be sandbox.

profiling/sync_to_user
harimohanraj89: I thought about this a bit. Shall we change this to `profiling/sync_to_user.example` with all the credentials XXXXX'd out, and .gitignore `profiling/sync_to_user`?

kitallis: Makes sense.

kitallis:
@harimohanraj89 Improved the README based on your comments.

profiling/sync_to_user.example
```diff
@@ -0,0 +1,13 @@
+GET https://api-sandbox.simple.org/api/v3/blood_pressures/sync
```
harimohanraj89, line 1: Love it, we should def keep the URL in here 👍

profiling/README.md
```diff
@@ -2,18 +2,33 @@
...
+
+The performance test itself can be run as follows:
 
-The perf test itself can be run as follows:
 ```
 make plot request=sync_to_user duration=20 rate=1
 ```
-This will run the perf test at the breakneck speed of 1 request per second for 20 seconds and open up (this will only work on OS X) a latency report page for you.
+
+This will run the test at the breakneck speed of 1 request per second for 20 seconds and open up (this will only work on OS X) a latency report page for you.
```
harimohanraj89, line 22: Your wit is undeniably entertaining, but perhaps we could omit it from the readme? 😅 (re: breakneck speed)

PR #1088, opened 2020-06-19
Fix broken data migration

vkrmis:
**Story card:** none
## Because

The Data migration to copy numbers from telemed v1 to telemed v2 was broken

### DIFF
--- a/db/data/20200612054906_copy_teleconsultation_phone_numbers_to_json.rb
+++ b/db/data/20200612054906_copy_teleconsultation_phone_numbers_to_json.rb
@@ -1,13 +1,18 @@
 class CopyTeleconsultationPhoneNumbersToJson < ActiveRecord::Migration[5.2]
   def up
-    Facility.where(enable_teleconsultation: true).each do |facility|
-      facility.teleconsultation_phone_numbers = [{"isd_code" => facility.teleconsultation_isd_code,
-                                                  "phone_number" => facility.teleconsultation_phone_number}]
+    Facility.all.each do |facility|
+      next if facility.teleconsultation_phone_number.blank? || facility.teleconsultation_isd_code.blank?
+
+      facility.teleconsultation_phone_numbers = [{isd_code: facility.teleconsultation_isd_code,
+                                                  phone_number: facility.teleconsultation_phone_number}]
       facility.save
     end
   end
 
   def down
-    Rails.logger.info "This data migration cannot be reversed. Skipping."
+    Facility.all.each do |facility|
+      facility.teleconsultation_phone_numbers = []
+      facility.save
+    end
   end
 end
--- a/db/data/20200612054906_copy_teleconsultation_phone_numbers_to_json.rb
+++ b/db/data/20200612054906_copy_teleconsultation_phone_numbers_to_json.rb
@@ -1,18 +1,14 @@
 class CopyTeleconsultationPhoneNumbersToJson < ActiveRecord::Migration[5.2]
   def up
-    Facility.all.each do |facility|
-      next if facility.teleconsultation_phone_number.blank? || facility.teleconsultation_isd_code.blank?
-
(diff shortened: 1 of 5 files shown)


db/data/20200612054906_copy_teleconsultation_phone_numbers_to_json.rb
```diff
@@ -1,13 +1,18 @@
 class CopyTeleconsultationPhoneNumbersToJson < ActiveRecord::Migration[5.2]
   def up
-    Facility.where(enable_teleconsultation: true).each do |facility|
-      facility.teleconsultation_phone_numbers = [{"isd_code" => facility.teleconsultation_isd_code,
-                                                  "phone_number" => facility.teleconsultation_phone_number}]
+    Facility.all.each do |facility|
```
harimohanraj89, line 3: What's the bug? We had facilities where teleconsult was disabled, but they had phone numbers?

harimohanraj89:
PS is it possible to re-run this migration correctly in SBX/QA etc, or is that a lost cause now?

db/data/20200612054906_copy_teleconsultation_phone_numbers_to_json.rb
vkrmis: That's also a possibility, but the bug here was that `teleconsultation_phone_numbers=` used to simply accept some JSON and set it, but now its a defined setter that failed to pick up string keys we're passing in here.

vkrmis:
> PS is it possible to re-run this migration correctly in SBX/QA etc, or is that a lost cause now?

Should be doable, the old columns were untouched.

harimohanraj89:
Neat. If the originals numbers were left untouched, I'd lean heavily towards writing a _new_ migration to do this correctly, and leaving a comment in this one. Editing migrations that have been run in some envs is a slippery slope. For example, if any other migrations were run after this, we'll have to roll them back as well to be able to rerun this correctly.

Shall we extract into a new migration?

vkrmis:
> Neat. If the originals numbers were left untouched, I'd lean heavily towards writing a _new_ migration to do this correctly, and leaving a comment in this one. Editing migrations that have been run in some envs is a slippery slope. For example, if any other migrations were run after this, we'll have to roll them back as well to be able to rerun this correctly.
> 
> Shall we extract into a new migration?

I'd be _okay_ to manually re-run this since its only run in SBX/QA, but generally I'd agree that new migration > edit migration

harimohanraj89:
Yup, it's simple enough. Let's do a new one, avoid the confusion. Makes it super straightforward to deploy 👍

vkrmis:
@harimohanraj89 Wdyt about sticking a `.with_indifferent_access` on each phone number in the 
setters so we don't have this problem?

harimohanraj89:
Love it 😄 I was considering doing that too, I'll let you decide whether you wanna include in this PR or a separate one. Either way works.

db/data/20200612054906_copy_teleconsultation_phone_numbers_to_json.rb
```diff
@@ -1,5 +1,6 @@
 class CopyTeleconsultationPhoneNumbersToJson < ActiveRecord::Migration[5.2]
   def up
+    # This data migration did not work correctly, the `teleconsultation_phone_numbers=` method didn't handle string keys
```
harimohanraj89, line 3: Neat. Let's add the filename of the working migration to this comment too for good measure.

db/data/20200619062218_copy_teleconsultation_phone_numbers_to_json_fixed.rb
```diff
@@ -0,0 +1,18 @@
+class CopyTeleconsultationPhoneNumbersToJsonFixed < ActiveRecord::Migration[5.2]
+  def up
+    Facility.all.each do |facility|
```
harimohanraj89, line 3: Consider `#find_each`.

PR #1091, opened 2020-06-19
Convert to Bootstrap

claudiovallejo:
### This addresses
* Creating new styles in `_reports.scss`
* Built responsive district report cards
* Removed Tailwind CSS
* Moved Chart.js script to `districts.js`
* Added Chart.js `script` to end of `preview.html.erb` (feels wrong)


app/assets/javascripts/districts.js
```diff
@@ -0,0 +1,244 @@
+const controlledPatientData = JSON.parse('<%= raw(@controlled_patients.to_json) %>');
```
claudiovallejo, line 1: @rsanheim is this the right place to save the `js` for reporting views?

rsanheim: Yup this should be right. You may have to include it in the layout via a JavaScript include tag, I don’t remember if Rails does that for you via convention or not.

rsanheim:
Closing in favor of https://github.com/simpledotorg/simple-server/pull/1105.

PR #1092, opened 2020-06-22
Small fixes to Patient Lookup

kitallis:
**Story card:** N/A

## Because

1. The page breaks because some patients don't have a registration_facility attached (probably card-reader patients). [Ref](https://sentry.io/organizations/resolve-to-save-lives/issues/1731043039/?environment=production&project=1217715&query=is%3Aunresolved)
2. Users who are not admins could not access the page because finer-grained resolution on `PatientPolicy` for overdue permission was broken. [Ref](https://sentry.io/organizations/resolve-to-save-lives/issues/1735134150/?environment=production&project=1217715&query=is%3Aunresolved).

## This addresses

1. Safe navigate.
2. The scope on Patient for facility is called `registration_facility` not `facility`.

### DIFF
--- a/app/policies/overdue_list/patient_policy.rb
+++ b/app/policies/overdue_list/patient_policy.rb
@@ -17,7 +17,7 @@ class OverdueList::PatientPolicy < ApplicationPolicy
       facility_ids = facility_ids_for_permission(:view_overdue_list)
       return scope.all unless facility_ids.present?
 
-      scope.where(facility_id: facility_ids)
+      scope.where(registration_facility_id: facility_ids)
     end
   end
 end
--- a/app/views/patients/lookup.html.erb
+++ b/app/views/patients/lookup.html.erb
@@ -16,7 +16,7 @@
               selected: @facility_id,
               wrapper: false
           },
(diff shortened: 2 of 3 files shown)


kitallis:
> Oh yeah, I recall seeing two different Sentry issues related to this. Could you link the sentry issues to this PR, or make a story to link to all?
> 
> PS any spec updates required for the policy change?

@harimohanraj89 

Done and done.

I know that's not what you asked, but I wrote a test for `OverdueList::PatientPolicy` anyway because I wanted to see what it takes to test permissions and... I tripped over a structure.

The tests I wrote refuse the fail even without the fix. It turns out a whole bunch of `OverdueList` scope tests are actually incorrect, for eg;

https://github.com/simpledotorg/simple-server/blob/master/spec/policies/overdue_list/appointment_policy_spec.rb#L114

Does not use `let!` and hence all data is always 0 in all these tests. I faced the same issue since I copied over some of the existing Policy scoping tests. Since they are `let` they are never evaluated inside the tests because we don't directly call them.

There's at least 3 examples of these scope tests that are wrong as far as I could tell. I'll fix them separately.

PR #1093, opened 2020-06-23
Add the missing constant with test coverage

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/173351363

This PR adds a missing constant and appropriate test coverage.

### DIFF
--- a/spec/controllers/api/v3/users_controller_spec.rb
+++ b/spec/controllers/api/v3/users_controller_spec.rb
@@ -148,7 +148,6 @@ RSpec.describe Api::V3::UsersController, type: :controller do
     it "updates the user otp and sends an sms to the user's phone number with the new otp" do
       existing_otp = user.otp
 
-      allow(RequestOtpSmsJob).to receive_message_chain("set.perform_later")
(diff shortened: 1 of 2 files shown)


spec/controllers/api/v3/users_controller_spec.rb
```diff
@@ -155,6 +155,19 @@
...
       expect(user.otp).not_to eq(existing_otp)
     end
+
+    it "uses a sensible default when the OTP delay is not configured" do
+      ENV["USER_OTP_SMS_DELAY_IN_SECONDS"] = nil
+
+      existing_otp = user.otp
+
+      allow(RequestOtpSmsJob).to receive_message_chain("set.perform_later")
+      expect(RequestOtpSmsJob).to receive_message_chain("set.perform_later").with(user)
```
kitallis, line 165: Why do we need the `allow` if there's already an `expect`?

harimohanraj89: Yup you're right. We don't really need em. Axed.

PR #1094, opened 2020-06-23
Support medication history in line lists

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/story/show/173464679

## Because

We want to expose medication histories for patients to our users. [PRD: Medication History](https://docs.google.com/document/d/1lJo9jLlPPDjBFJ12tBGgdF-22rCGhNiVkquG7u9hBbo). 

## This addresses

This adds a method to `PrescriptionDrug` that returns the `active` prescription drugs for a patient on a given date. This can be used to construct what the medications looked like for each visit in the line list.

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -188,7 +188,7 @@ class Patient < ApplicationRecord
     super(new_call_result)
   end
 
-  def prescribed_drugs(date: Date.today)
+  def prescribed_drugs(date: Date.current)
     prescription_drugs.prescribed_as_of(date)
   end
 
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -188,6 +188,10 @@ class Patient < ApplicationRecord
     super(new_call_result)
   end
 
+  def prescribed_drugs(date: Date.current)
+    prescription_drugs.prescribed_as_on(date: date)
+  end
+
   def self.not_contacted
     where(contacted_by_counsellor: false)
       .where(could_not_contact_reason: nil)
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -189,7 +189,7 @@ class Patient < ApplicationRecord
   end
 
   def prescribed_drugs(date: Date.today)
-    prescription_drugs.prescribed_as_on(date: date)
+    prescription_drugs.prescribed_as_of(date)
   end
 
   def self.not_contacted
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -188,7 +188,7 @@ class Patient < ApplicationRecord
     super(new_call_result)
   end
 
-  def prescribed_drugs(date: Date.current)
+  def prescribed_drugs(date: Date.today)
     prescription_drugs.prescribed_as_on(date: date)
   end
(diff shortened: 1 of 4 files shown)


app/models/patient.rb
```diff
@@ -188,6 +188,10 @@ def call_result=(new_call_result)
     super(new_call_result)
   end
 
+  def prescribed_drugs(date: Date.today)
+    prescription_drugs.prescribed_as_on(date: date)
```
kitallis, line 192:
```suggestion
    prescription_drugs.prescribed_as_of(date: date)
```

app/models/prescription_drug.rb
```diff
@@ -13,6 +13,13 @@ class PrescriptionDrug < ApplicationRecord
   validates :is_protocol_drug, inclusion: {in: [true, false]}
   validates :is_deleted, inclusion: {in: [true, false]}
 
+  def self.prescribed_as_on(date: Date.today)
```
kitallis, line 16:
```suggestion
  def self.prescribed_as_on(date: Date.current)
```

app/models/patient.rb
```diff
@@ -188,6 +188,10 @@ def call_result=(new_call_result)
     super(new_call_result)
   end
 
+  def prescribed_drugs(date: Date.today)
```
kitallis, line 191:
```suggestion
  def prescribed_drugs(date: Date.current)
```

app/models/patient.rb
harimohanraj89: Could `active_on` work too? Just throwing it into the mix, I'll let you pick the one you think reads best.

app/models/prescription_drug.rb
harimohanraj89: Given the method naming, I don't think it makes sense to provide a default argument. I'd argue this method should be passed a date always.

PS, does the method behave gracefully if we pass it a DateTime as well?

app/models/patient.rb
harimohanraj89: Could we add some trivial test coverage for this method too? Ensure that the `patient` invokes the right method on its `prescription_drugs` with the right arguments?

spec/models/prescription_drug_spec.rb
```diff
@@ -14,6 +14,52 @@
...
+    end
+
+    let!(:facility) { create(:facility) }
+    let!(:patient) { create(:patient) }
+    let!(:user) { create(:user) }
+    let!(:initial_visit_time) { Time.parse "01 Jan 2020 14:30" }
+    let!(:latest_visit_time) { Time.parse "15 Jan 2020 14:30" }
+    let!(:prescription_drugs) do
+      Timecop.freeze(initial_visit_time) do
+        create_list(:prescription_drug, 2, is_protocol_drug: true, patient: patient, facility: facility, user: user)
```
harimohanraj89, line 29: Why do we have to create and assign a facility and user? Ideally, the factory should create em if missing. If we _have_ to assign em, could we just use `patient.registration_*` to save on some db operations?

PR #1095, opened 2020-06-24
Mitigate uniqueness exception noise

harimohanraj89:
**Story card:** TBD

## Because

We get a lot of `ActiveRecord::RecordNotUnique` exceptions raised by our mergeable concern. This is likely due to a new object being processed in two web requests in parallel. Both processes attempt to `create` a record, and the slower one fails. 

## This addresses

This PR updates the mergeable concern to rescue `ActiveRecord::RecordNotUnique` exceptions and retry _once_. On the second try, the mergable concern should identify that another record exists (the one that would have clashed with this thing in the first place), and will update the existing record instead.


app/models/concerns/mergeable.rb
```diff
@@ -24,20 +24,28 @@ def compute_merge_status(attributes)
...
-        discarded_record(existing_record)
-      when :invalid
-        invalid_record(new_record)
-      when :new
-        create_new_record(attributes)
-      when :updated
-        update_existing_record(existing_record, attributes)
-      when :old
-        return_old_record(existing_record)
+        retries <= 1 ? retry : raise
```
kitallis, line 48: My natural design-for-high-throughput-brain flags a synchronous retry immediately, but I believe we won't run into any particular backpressure issues for the moment.

How can we test this?

kitallis, line 48: Should we also raise the original exception back instead of blank top-level one?

harimohanraj89: I believe `raise` with no args reraises the same exception. I will confirm.

harimohanraj89: Well, to start, we can look at the frequency of the problem. The number of extra retries shouldn't exceed that for sure. I'm going to see if any non-production environments have the same exceptions firing organically. If so, we can deploy to one of them and observe drop-offs in exceptions and successful updates.

Ooh, we could instrument some logs or NR events to identify specific records that are affected, so we can follow up on them one-by-one in, say, Sandbox. I'll do this too.

harimohanraj89:
@kitallis Thanks for the early review. Sounds like we could see this through to completion. I'll hook up the necessary instrumentation and test coverage, as well as scope out the field for best ways to test this.

harimohanraj89:
I'm not going to have time for this now. So long, fond PR.

harimohanraj89:
We used to have thousands of these, but now we have handfuls. Hence, I"m going to close this.

PR #1096, opened 2020-06-24
Fix the scoping-related tests around User Permissions

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/173486012

## Because

Many of the permission tests around scoping were incorrect because they weren't properly creating the necessary test data.

## This addresses

* Change the test data required to be `let!` instead of `let` so they are actually force-evaluated during the test
* Minor aesthetic changes

## Notes

All tests but one worked properly even after correcting the test data. The `spec/policies/overdue_list/patient_summary_policy_spec.rb` needed some modification because its assertions were wrong.

### DIFF
--- a/spec/policies/overdue_list/patient_summary_policy_spec.rb
+++ b/spec/policies/overdue_list/patient_summary_policy_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe OverdueList::PatientSummaryPolicy do
   end
 end
 
-RSpec.describe OverdueList::AppointmentPolicy::Scope do
+RSpec.describe OverdueList::PatientSummaryPolicy::Scope do
   let(:subject) { described_class }
   let(:organization) { create(:organization) }
   let(:facility_group) { create(:facility_group, organization: organization) }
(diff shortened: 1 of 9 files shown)


spec/policies/overdue_list/patient_summary_policy_spec.rb
```diff
@@ -24,25 +23,23 @@
...
-
   let(:facility1) { create(:facility, facility_group: facility_group) }
   let(:facility2) { create(:facility) }
-  let(:appointment1) { create(:appointment, :overdue, facility: facility1) }
-  let(:appointment2) { create(:appointment, :overdue, facility: facility2) }
+  let!(:appointment1) { create(:appointment, :overdue, facility: facility1) }
+  let!(:appointment2) { create(:appointment, :overdue, facility: facility2) }
 
-  context "user with permission to access appointment information for all organizations" do
+  context "user with permission to access patient summaries for all organizations" do
```
harimohanraj89, line 33: It appears this spec had some copypasta but we're correcting the wrong bits. This is meant to test the `OverdueList::PatientSummaryPolicy::Scope` ([check it out](https://github.com/simpledotorg/simple-server/blob/master/app/policies/overdue_list/patient_summary_policy.rb#L6-L22)). It's supposed to return a relation of `PatientSummary` records ([check it out](https://github.com/simpledotorg/simple-server/blob/master/app/controllers/appointments_controller.rb#L19)).

Could we update the classname above, and pass in `PatientSummary.all` below?

kitallis: @harimohanraj89 Yeah, I saw that (missed the scope name entirely), but I thought the assertions were done that way intentionally. Now that I look at it again, it's all completely wrong.

Could you verify if [this](https://github.com/simpledotorg/simple-server/pull/1096/commits/ea351d4a03e3ea6771c8722ac58df71c0ba42240) looks correct now?

I validated that the number of records from `PatientSummary` are non-zero (except for `scope.none` case), so it looks accurate to me.

harimohanraj89: Hehe yeah this looks a LOT better now. 👍

PR #1098, opened 2020-06-24
Telemed reporting

vkrmis:
**Story card:** https://www.pivotaltracker.com/story/show/173298496

## Because

We need reporting around telemedicine usage in this [format](https://docs.google.com/spreadsheets/d/14KWjr4JjVfQyrARcV2IPJYu5tytZIFBI6ivVuNqg4Ro/edit#gid=0)

## This addresses

Adds a rake task that can take mixpanel data as input, combine it with information from the database, and generate the required report.

Report example: 
Screenshot 2020-08-10 at 6 35 58 PM

Input example:
Screenshot 2020-08-10 at 6 37 04 PM


lib/tasks/telemedicine_reports.rake
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require "tasks/scripts/telemedicine_reports"
+
+desc "Generates the telemedicine report, takes the mixpanel report as input"
+task :telemedicine_reports, [:mixpanel_report, :p1_start, :p1_end, :p2_start, :p2_end] => :environment do |_t, args|
```
harimohanraj89, line 6: How about `bundle exec rake reports:telemedicine`?

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,143 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module TelemedicineReports
+  class << self
+    def parse_mixpanel(mixpanel_csv_path)
+      mixpanel_csv = File.read(mixpanel_csv_path)
+      mixpanel = CSV.parse(mixpanel_csv, headers: false)
```
harimohanraj89, line 9: Instead of `headers: false` and dropping 1 below, can we use `headers: true`, and use headers names to access data instead of `row[1]`, etc?

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,143 @@
...
+    def format_facility_data(facility_data)
+      facility_data.group_by { |row| row[:state] }.map { |state, districts|
+        {state: state, count: districts.count, hwc_and_sc: hwc_and_sc_count(districts),
+         p1: aggregate_period(:p1, districts), p2: aggregate_period(:p2, districts),
+         users: sum_rows(districts, :users),
+         districts: districts.group_by { |row| row[:district] }.map { |district, facilities|
+           {district: district, state: state, count: facilities.count,
+            hwc_and_sc: hwc_and_sc_count(facilities), p1: aggregate_period(:p1, facilities),
+            p2: aggregate_period(:p2, facilities), users: sum_rows(facilities, :users)}
+         }.sort_by { |district| district[:district] }}
```
harimohanraj89, line 92: Can we format this hash please?

```
{
  state: state, 
  count: districts.count, 
  hwc_and_sc: hwc_and_sc_count(districts),
  p1: aggregate_period(:p1, districts), 
  p2: aggregate_period(:p2, districts),
  users: sum_rows(districts, :users),
  districts: districts.group_by { |row| row[:district] }.map { |district, facilities|
    {
      district: district, 
      state: state, 
      count: facilities.count,
      hwc_and_sc: hwc_and_sc_count(facilities), 
      p1: aggregate_period(:p1, facilities),
      p2: aggregate_period(:p2, facilities), 
      users: sum_rows(facilities, :users)
    }
  }.sort_by { |district| district[:district] }}
}
```

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,143 @@
...
+      CSV.open("telemedicine_report.csv", "w") do |csv|
+        csv << ["", "", "", "", "", "", "Between #{p1_start.strftime("%d %b %Y")} and #{p1_end.strftime("%d %b %Y")}", "", "",
+          "", "", "", "Between #{p2_start.strftime("%d %b %Y")} and #{p2_end.strftime("%d %b %Y")}"]
+        csv << ["State", "District", "Facilities with TM", "HWCs & SCs with TM", "Users at HWCs & SCs", "",
+          "Patients who visited", "Patients with High BP", "Patients with High Blood Sugar", "Patients with High BP or Sugar", "Teleconsult Button Clicks", "",
+          "Patients who visited", "Patients with High BP", "Patients with High Blood Sugar", "Patients with High BP or Sugar", "Teleconsult Button Clicks"]
+        facilities_data.each do |state|
+          csv << [state[:state], "", state[:count], state[:hwc_and_sc], state[:users], "",
+            state[:p1][:visits], state[:p1][:high_bp], state[:p1][:high_bs], state[:p1][:high_bp_or_bs], fetch_clicks(p1_mixpanel_data, state, true), "",
+            state[:p2][:visits], state[:p2][:high_bp], state[:p2][:high_bs], state[:p2][:high_bp_or_bs], fetch_clicks(p2_mixpanel_data, state, true)]
```
harimohanraj89, line 49: Can we format these inserts please?

```
csv << [
  "State", 
  "District", 
  "Facilities with TM", 
  "HWCs & SCs with TM", 
  "Users at HWCs & SCs", 
  "",
  "Patients who visited", 
  "Patients with High BP", 
  "Patients with High Blood Sugar", 
  "Patients with High BP or Sugar", 
  "Teleconsult Button Clicks", 
  "",
  "Patients who visited", 
  "Patients with High BP", 
  "Patients with High Blood Sugar", 
  "Patients with High BP or Sugar", 
  "Teleconsult Button Clicks"
]

facilities_data.each do |state|
  csv << [
    state[:state], 
    "", 
    state[:count], 
    state[:hwc_and_sc], 
    state[:users], 
    "",
    state[:p1][:visits], 
    state[:p1][:high_bp], 
    state[:p1][:high_bs], 
    state[:p1][:high_bp_or_bs], 
    fetch_clicks(p1_mixpanel_data, state, true), 
    "",
    state[:p2][:visits], 
    state[:p2][:high_bp], 
    state[:p2][:high_bs], 
    state[:p2][:high_bp_or_bs], 
    fetch_clicks(p2_mixpanel_data, state, true)
  ]
end
```

lib/tasks/scripts/telemedicine_reports.rb
vkrmis: The last two headers are dynamic unfortunately, can't predict what they're going to be.
Screenshot 2020-06-25 at 4 41 20 PM

harimohanraj89: Fuuuuun, coolbeans 👍

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,322 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module TelemedicineReports
```
vkrmis, line 5: This module has two primary methods:
`parse_mixpanel` - reads the mixpanel csv and hydrates it with some facility level info
`generate_report` - uses the mixpanel info, and fetches data from the db to generate the report

All the other methods here are helpers to put the data structures together in the format that the report needs.

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,322 @@
...
+      }
+    end
+
+    def generate_report(mixpanel_data, period_start, period_end)
+      raw_mixpanel_data = mixpanel_data.select { |row|
+        (row[:date] >= period_start) && (row[:date] <= period_end)
+      }
+      mixpanel_data = format_mixpanel_data(raw_mixpanel_data)
+
+      facilities = Facility.where(enable_teleconsultation: true).map { |facility|
```
vkrmis, line 29: We only want detailed reporting for HWCs and SCs where teleconsultation is enabled

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,322 @@
...
+          "",
+          "Between #{period_start.strftime("%d %b %Y")} and #{period_end.strftime("%d %b %Y")}",
+          "",
+          "",
+          "",
+          "",
+          ""
+        ]
+
+        csv << [
```
vkrmis, line 66: This is the state level overview

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,322 @@
...
```
vkrmis, line 139: This is the district level view, with per-facility data

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,322 @@
...
+                ""
+              ]
+            end
+          end
+        end
+
+        csv << []
+        csv << []
+
+        daily_activity_data = raw_mixpanel_data.group_by { |row| row[:date] }.sort_by { |date, _rows| date }.map { |date, rows|
```
vkrmis, line 214: This prints the daily call log

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,322 @@
...
+        {state: state,
+         clicks: sum_values(districts, :clicks),
+         districts: districts.group_by { |row| row[:district] }.map { |district, users|
+                      {district: district,
+                       clicks: sum_values(users, :clicks)}
+                    }}
+      }
+    end
+
+    def format_facility_data(facility_data)
```
vkrmis, line 238: This creates the nested data structure that we iterate over to generate the report

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,322 @@
...
+
+    def high_bps(bps)
+      bps.hypertensive.uniq { |bp| bp[:patient_id] }
+    end
+
+    def high_sugars(blood_sugars)
+      blood_sugars.select(&:diabetic?).uniq { |bs| bs[:patient_id] }
+    end
+
+    def facility_measures(facility, p_start, p_end)
```
vkrmis, line 267: Fetches per-facility information that we roll up for district and state level numbers

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -0,0 +1,322 @@
...
+         date: Date.parse(row[2]),
+         clicks: row[3].to_i,
+         facility_id: facility.id,
+         district: facility.district,
+         state: facility.state,
+         type: facility.facility_type}
+      }
+    end
+
+    def generate_report(mixpanel_data, period_start, period_end)
```
harimohanraj89, line 23: Is this supposed to be used like `TelemedicineReports.generate_report(TelemedicineReports.parse_mixpanel('path'), ...)`?

spec/lib/tasks/scripts/telemedicine_reports_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  context ".generate_report" do
+    it "generates a report file" do
+      mixpanel_data = TelemedicineReports.parse_mixpanel(file_path)
+      period_start = Date.parse("2020-08-03").beginning_of_day
+      period_end = Date.parse("2020-08-09").end_of_day
+
+      report_file_path = "telemedicine_report_03_Aug_to_09_Aug.csv"
+
+      allow(CSV).to receive(:open).with(report_file_path, "w")
+      TelemedicineReports.generate_report(mixpanel_data, period_start, period_end)
```
harimohanraj89, line 29: There's no assertion in this spec. Can we assert that this creates the right data? We could either allow `CSV` to return a generic IOStream above, or we could split the report data generation and the report _writing to file_ into separate methods and test the former.

lib/tasks/scripts/telemedicine_reports.rb
vkrmis: There's a usage here: https://github.com/simpledotorg/simple-server/blob/242176467e371b39261c88f21f8761573dde6bc0/lib/tasks/reports.rake#L24

But, essentially yes

spec/lib/tasks/scripts/telemedicine_reports_spec.rb
vkrmis: Significantly refactored this to make it easier to test.
Caveat: the output numbers are implicitly reliant on current factory behaviour (especially facilities - it creates all facilities in the same State and District). Should be okay in the short term, but we should drop these specs once Telemed MVP is ready.

spec/lib/tasks/scripts/telemedicine_reports_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      report.generate
+      expect(report.mixpanel_data[:hydrated].size).to eq(6)
+    end
+  end
+
+  context ".generate_report" do
+    it "generates a report file" do
+      filename = "telemedicine_report_03_Aug_to_09_Aug.csv"
+      # This result is dependent on the factories, could become flaky if the factories change significantly.
+      # We should drop these specs once we switch over to the Telemed MVP
```
harimohanraj89, line 33: Could we take care of just the riskiest part of this you mentioned, ie, explicitly set a district and state in the two factory facilities above? I'd definitely prefer being verbose but explicit about traits of the test data that are required, vs relying on factory behavior.

PR #1099, opened 2020-06-24
Send one-off messages to patients

kitallis:
**Story card:** https://www.pivotaltracker.com/story/show/173300953

## Because

We want to be able to send arbitrary messages to an arbitrary set of patients

## This addresses

Adds a script to ease sending messages.


lib/tasks/scripts/message_patients.rb
```diff
@@ -0,0 +1,85 @@
+class MessagePatients
+  def self.call(*args)
+    new(*args).call
+  end
+
+  attr_reader :patients, :message, :whatsapp, :dryrun, :verbose, :report
+
+  def initialize(patients, message, whatsapp: true, dryrun: false, verbose: true)
```
harimohanraj89, line 8: Let's change this instead to `channel: :whatsaap`, or `message_type: :whatsapp`?

lib/tasks/scripts/message_patients.rb
```diff
@@ -0,0 +1,85 @@
...
+    return if dryrun
+
+    send_messages!
+    print_report
+    self
+  end
+
+  private
+
+  def send_messages!
```
harimohanraj89, line 33: The `! = mutation/danger` convention is nice when you've got both versions of a method. In this case, I think we can drop the bangs, the methods are pretty self-explanatory. What do you think?

lib/tasks/scripts/message_patients.rb
```diff
@@ -0,0 +1,85 @@
...
+    valid_patients.each do |patient|
+      phone_number = patient.latest_phone_number
```
harimohanraj89, line 35: `latest_mobile_number`?

lib/tasks/scripts/message_patients.rb
```diff
@@ -0,0 +1,85 @@
...
+          if whatsapp
+            NotificationService
+              .new
+              .send_whatsapp(phone_number, message)
+          else
+            NotificationService
+              .new
+              .send_sms(phone_number, message)
+          end
+      rescue Twilio::REST::TwilioError => _
```
harimohanraj89, line 48: You can leave out the `=> _` if we're not using it.

lib/tasks/scripts/message_patients.rb
```diff
@@ -0,0 +1,85 @@
...
+        update_report!(:exception, patient: patient)
+      else
+        update_report!(:responses, response: response, patient: patient)
+      end
+    end
+  end
+
+  def valid_patients
```
harimohanraj89, line 56: `contactable_patients`?

lib/tasks/scripts/message_patients.rb
kitallis: Ha, that's what I [started](https://github.com/simpledotorg/simple-server/pull/1099/commits/2d55c532f121b64ae54e4b715117bf750c684da3#diff-4b9785d5cf8fa1feff9c07408e1ce001R33) with and then reverted it to be this way because there were only two states so it made the conditional simpler. Fine with either way.

harimohanraj89: Since this isn't a background job, where maintaining interface is super important, let's leave it as is 😄

kitallis:
@harimohanraj89 I have address the feedback [here](https://github.com/simpledotorg/simple-server/pull/1099/commits/4fdeb8a785307d5b5139d6ede5794f9532b09fe6).

lib/tasks/scripts/message_patients.rb
```diff
@@ -0,0 +1,98 @@
...
+  def send_messages
+    contactable_patients.each do |patient|
+      phone_number = patient.latest_mobile_number
+
+      begin
+        response =
+          if whatsapp?
+            NotificationService
+              .new
+              .send_whatsapp(phone_number, message)
```
harimohanraj89, line 47: One-line em if you'd like 😄 `NotificationService.new.send_whatsapp(phone_number, message)` isn't too long.

PR #1100, opened 2020-06-24
Queue SMS fallback job if WhatsApp message fails

timcheadle:
**Story cards:**
- https://app.clubhouse.io/simpledotorg/story/61/fallback-to-sms-if-whatsapp-reminders-don-t-work
- https://www.pivotaltracker.com/story/show/172893687

## Because

Not every patient has WhatsApp, and we want appointment reminders to fall back to SMS if WhatsApp fails to deliver.

## This addresses

In order to know if a message delivered successfully, Twilio sends us a webhook callback to `/api/v3/twilio_sms_delivery`. If this payload is a WhatsApp message and is unsuccessful, we schedule a worker to send the same message via SMS.

## Implementation Notes

* The window of time when patients can be messaged, and logic to obtain the next available message timeslot has been moved to the `Communication` model
* This timeslot logic is leveraged in the Twilio callback controller, scheduling an SMS reminder for the next timeslot if a whatsapp message fails


app/models/appointment.rb
```diff
@@ -60,6 +60,27 @@ def self.eligible_for_reminders(days_overdue: 3)
       .merge(Patient.contactable)
   end
 
+  def self.reminder_start_hour
```
kitallis, line 63: **Non-blocking:**

I read your comment about moving this `Communication`, which sounds better.

What happens when we need to stagger the time intervals at which we send these messages for different kinds of notifications or communications to avoid spamming? (which say, the notification framework suggests)?

Ideally we shouldn't tie relatively generic model methods against specific env variables, and instead pass them in.

harimohanraj89: I think we can easily update the behavior of the `Communication.next_messaging_time` to spice up its returns with some entropy. The big win I see is one accessible place to ask this question, which eliminates the need to read and pass a time window through several layers of application code as arguments.

As for _where_ this logic should exist, that'll likely evolve as our requirements evolve. For now, `Communication` seems fine to me.

app/controllers/api/v3/twilio_sms_delivery_controller.rb
```diff
@@ -3,15 +3,24 @@ class Api::V3::TwilioSmsDeliveryController < ApplicationController
...
-    TwilioSmsDeliveryDetail.where(session_id: message_session_id).first.update(update_params)
+    twilio_message = TwilioSmsDeliveryDetail.find_by(session_id: message_session_id)
+
+    twilio_message.update(update_params)
+    communication_type = twilio_message.communication.communication_type
+    appointment_id = twilio_message.communication.appointment_id
+
+    if communication_type == "missed_visit_whatsapp_reminder" && twilio_message.unsuccessful?
+      AppointmentNotification::Worker.perform_at(Communication.next_messaging_time, appointment_id, "missed_visit_sms_reminder")
+    end
```
harimohanraj89, line 14: This ☝️ schedules the SMS job if we learn that a whatsapp just failed.

app/jobs/appointment_notification/missed_visit_job.rb
```diff
@@ -1,18 +1,19 @@
 class AppointmentNotification::MissedVisitJob
-  include Notifiable
   include Sidekiq::Worker
 
   sidekiq_options queue: "high"
 
-  def perform(schedule_hour_start, schedule_hour_finish)
+  def perform
```
harimohanraj89, line 6: This job no longer accepts the time window, since the `AppointmentNotificationService` invoked below can find that out itself.

app/services/appointment_notification_service.rb
```diff
@@ -21,11 +20,13 @@ def send_after_missed_visit
     eligible_appointments.each do |appointment|
       next if appointment.previously_communicated_via?(communication_type)
 
-      AppointmentNotification::Worker.perform_at(schedule_at, appointment.id, communication_type)
+      next_messaging_time = Communication.next_messaging_time
```
harimohanraj89, line 23: This is the meat of the refactoring done in this PR. This service fetches the "next message time" from the Communication model directly, instead of receiving it as an arg through a few layers of application code.

app/models/communication.rb
```diff
@@ -48,6 +48,26 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
     end
   end
 
+  def self.messaging_start_hour
+    @messaging_start_hour ||= ENV.fetch("APPOINTMENT_NOTIFICATION_HOUR_OF_DAY_START", 14).to_i
```
prabhanshuguptagit, line 52: `14` and `16` could be moved to `DEFAULT_..` constants.

app/models/communication.rb
```diff
@@ -48,6 +48,26 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
...
+  end
+
+  def self.messaging_end_hour
+    @messaging_end_hour ||= ENV.fetch("APPOINTMENT_NOTIFICATION_HOUR_OF_DAY_FINISH", 16).to_i
+  end
+
+  def self.next_messaging_time
```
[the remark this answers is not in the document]
prabhanshuguptagit: > wondering if we should move these to the worker itself

On second thought that doesn't seem like a great idea, we'll still want to do `perform_at` and pass in the time to the worker. Maybe we could call the methods `next_appointment_notification_time` and so on.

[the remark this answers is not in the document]
harimohanraj89: @prabhanshuguptagit I'd argue that any scheduled messaging we send to patients will need to adhere to "sensible hours", so it's better to keep these hours generic. What do you think?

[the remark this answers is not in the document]
prabhanshuguptagit: Just saw @kitallis's thread above about this, makes sense 👍

spec/models/communication_spec.rb
```diff
@@ -15,21 +15,54 @@
...
+  context "Utilities" do
+    describe ".next_appointment_time" do
+      before do
+        @original_time_zone = Time.zone
+        Time.zone = Rails.application.config.country[:time_zone]
+        allow(Communication).to receive(:messaging_start_hour).and_return(10)
+        allow(Communication).to receive(:messaging_end_hour).and_return(16)
+      end
+
+      after do
```
prabhanshuguptagit, line 27: This could be an `around` do block with a `Time.use_zone` instead. That'll be quite neat.

spec/models/communication_spec.rb
```diff
@@ -15,21 +15,54 @@
...
+        Time.zone = @original_time_zone
+      end
+
+      it "returns the current time if currently ithin the messaging window" do
```
prabhanshuguptagit, line 31: 👾  
```suggestion
      it "returns the current time if currently within the messaging window" do
```

harimohanraj89:
Old semaphore check is a flapper. Ignoring.

PR #1101, opened 2020-06-25
Patient line list with medication history

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/story/show/173464679

## Because

We want to give users a richer line list with some historical data for each patient.

**Starter CSV template**
[patient-line-list-with-history-example.txt](https://github.com/simpledotorg/simple-server/files/4830072/patient-line-list-with-history-example.txt)

## This addresses

Adds a new line list download that displays visit and medication history data. Compared to the current line list this:
- Adds last 3 BPs (_visits_).
- Adds prescription drugs for each visit.
- Adds a `Medication Updated` column for each visit which tells if medication was changed from the last visit.
- Adds a column to each visit for the number of days till the next appointment date ( `Follow up Days`).
- Re-orders some columns.

image

**Demo CSV**
[csv-demo-1.txt](https://github.com/simpledotorg/simple-server/files/4845589/csv-demo-1.txt)(Github doesn't allow attaching .csv's :|)

**Demo CSV 2**
[csv-demo-2.txt](https://github.com/simpledotorg/simple-server/files/4857875/csv-demo2.txt)


app/controllers/analytics/districts_controller.rb
```diff
@@ -51,6 +51,22 @@ def patient_list
...
+      organization_id: @organization_district.organization.id
+    }, with_medication_history: true)
+
+    redirect_to(
+      analytics_organization_district_path(id: @organization_district.district_name),
+      notice: I18n.t("patient_list_email.notice",
+        model_type: "district",
+        model_name: @organization_district.district_name)
+    )
+  end
```
harimohanraj89, line 68: Could we re-use the existing route with a param, instead of a new route?

app/controllers/analytics/facilities_controller.rb
```diff
@@ -50,6 +50,20 @@ def patient_list
...
+    PatientListDownloadJob.perform_later(recipient_email,
+      "facility",
+      {facility_id: @facility.id},
+      with_medication_history: true)
+
+    redirect_to(
+      analytics_facility_path(@facility),
+      notice: I18n.t("patient_list_email.notice", model_type: "facility", model_name: @facility.name)
+    )
+  end
```
harimohanraj89, line 65: Could we re-use the existing route with a param, instead of a new route?

app/models/blood_sugar.rb
```diff
@@ -34,4 +41,8 @@ class BloodSugar < ApplicationRecord
   def diabetic?
     blood_sugar_value >= THRESHOLDS[:high][blood_sugar_type]
   end
+
+  def to_s
+    "#{blood_sugar_value} #{BLOOD_SUGAR_UNITS[blood_sugar_type]}"
+  end
```
harimohanraj89, line 47: ❤️

app/exporters/patients_with_history_exporter.rb
```diff
@@ -0,0 +1,204 @@
...
+
+    def appointment_created_on(patient, date)
+      patient.appointments
+        .where(device_created_at: date&.all_day)
+        .order(device_created_at: :asc)
+        .first
+    end
+
+    def cache_medication_history(patient, dates)
+      @medications = {patient => {}}
```
harimohanraj89, line 168: Why store a hash with a patient key if this method only deals with one patient?

Is the intention to build up a hash of several patients? In that case, should we rewrite this as:

```
@medications ||= {}
@medications[patient] ||= {}
```

Right now, we appear to be clobbering the hash each time invoke the method.

spec/exporters/patients_with_history_exporter_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+        device_created_at: 3.months.ago,
+        facility: facility,
+        patient: patient).sort_by(&:name),
+      *create_list(:prescription_drug,
+        3,
+        device_created_at: 3.months.ago,
+        facility: facility,
+        patient: patient).sort_by(&:name)
+    ]
+  end
```
harimohanraj89, line 37: We should introduce a lot more nuance to the test data we set up here. Here are some thoughts:

* Create drugs on specific appointment days
* Delete some drugs on appointment days

app/exporters/patients_with_history_exporter.rb
```diff
@@ -0,0 +1,204 @@
...
+          "BP #{i} Other Medications"]
+      end,
+      "Latest Blood Sugar Date",
+      "Latest Blood Sugar Value",
+      "Latest Blood Sugar Type"].flatten.compact
+  end
+
+  def self.csv_fields(patient)
+    registration_facility = patient.registration_facility
+    latest_bps = patient.latest_blood_pressures.first(DISPLAY_BLOOD_PRESSURES)
```
harimohanraj89, line 98: Don't we need to grab one more BP? How do we populate the "medication updated?" correctly on the oldest BP?

app/exporters/patients_with_history_exporter.rb
prabhanshuguptagit: This fetches and keeps medications for each patient only, since we populate the CSV from latest -> oldest but the `medication updated?` calculation needs to refer to older medications while populating later medications. This is just to avoid double querying. Maybe a better name would be `fetch_medication_history`?

spec/exporters/patients_with_history_exporter_spec.rb
prabhanshuguptagit: The drugs test are already covered in `prescription_drug_spec`. A couple of cases I'd like to add here though are:
- Have a `bp_4`
- Have medications on `bp_4` and make sure they show up correctly.

app/exporters/patients_with_history_exporter.rb
prabhanshuguptagit: Hrm that's correct. Right now it gets the medication before the oldest BP as `none` and compares to that. Thanks for spotting :eyes:

app/controllers/analytics/districts_controller.rb
prabhanshuguptagit: I slightly prefer the separate actions, because we're going to keep both line lists around and we could expect these to diverge over time. Couldn't think of a strong reason to merge them in one, wdyt?

harimohanraj89: My thought was, this is the beginning of various "flavors" of a line list. Imagine a more sophisticated mechanism to say "I want blood sugars, I don't want medication history, I want regn info", etc. However, if we do that, it'll probably live somewhere brand new anyway.

Agree, let's avoid early optimization and keep them separate 👍

app/exporters/patients_with_history_exporter.rb
harimohanraj89: That may be a better name indeed. If there's only a single key we're dealing, perhaps we don't need the `patient` key at all then.

```
{
  date_1: [stuff],
  date_2: [stuff]
}
```

spec/exporters/patients_with_history_exporter_spec.rb
harimohanraj89: Ah great points, and agree with the holes you've suggested filling 👍

prabhanshuguptagit:
@harimohanraj89 On a431f4b, I verified that the ordering is fixed after this change. I haven't been able to reproduce this in a spec so far, my guess is this only happens when there is more than one batch of patients and/or a combination of the `.order` in `PatientListDownloadJob`.

prabhanshuguptagit:
Addressed the design [feedback](https://simpledotorg.slack.com/archives/CK95QFJA2/p1593607556154900)

PR #1102, opened 2020-06-25
Fix WhatsApp Twilio delivery status

timcheadle:
**Story card:** none

## Because

We need to capture Twilio session IDs to handle delivery status notifications, and WhatsApp uses `MessageSid` instead of `SmsSid`.

## This addresses

This hotfix allows us to accept both values.

### DIFF
--- a/app/controllers/api/v3/twilio_sms_delivery_controller.rb
+++ b/app/controllers/api/v3/twilio_sms_delivery_controller.rb
@@ -21,7 +21,7 @@ class Api::V3::TwilioSmsDeliveryController < ApplicationController
   end
 
   def delivery_status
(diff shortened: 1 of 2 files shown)


app/controllers/api/v3/twilio_sms_delivery_controller.rb
```diff
@@ -16,6 +16,10 @@ def update_params
     details
   end
 
+  def message_session_id
+    params["MessageSid"] || params["SmsSid"]
+  end
+
   def delivery_status
     params["SmsStatus"] || params["MessageStatus"] || TwilioSmsDeliveryDetail.results[:unknown]
```
kitallis, line 24: Should we invert this clause to

```suggestion
    params["MessageStatus"] || params["SmsStatus"] || TwilioSmsDeliveryDetail.results[:unknown]
```

as well?

PR #1085, opened 2020-06-17
DM info in linelist

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/173298755

## Because

We want to include blood sugars in the line list

## This addresses

This PR does the following things to the patient line list download:

* Makes the "Latest BP Date" the first column in the "Latest BP" section
* Add three new columns to the patient line list: 
  * "Latest Blood Sugar Date"
  * "Latest Blood Sugar Value" (eg. `150 mg/dL`)
  * "Latest Blood Sugar Type" (eg. `Random`)
* Populates ~blood sugar data into the new columns if and only if the patient had a blood sugar reading on the same day as their latest BP~ patient's latest blood sugar with date

## Additional notes

* This PR also tackles a few N+1 queries that were remaining in the line list. The line list query should now be free of N+1's.
* In order to combat the memory usage spike from avoiding N+1 queries, we've reduced the patient batch size `1000 -> 20`

PR #1109, opened 2020-06-26
Linelist ordering bug

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/173508399

This PR addresses a bug introduced in #1085 , where some necessary explicit ordering was removed, attempting to rely on ordered associations. However, due to the batching that occurs in the patient exporter, when eager loading ordered associations, the ordering is not respected. Hence we abandon eager-loading ordered associations, and instead make explicit ordered queries iteratively instead.


app/exporters/patients_exporter.rb
```diff
@@ -31,10 +31,6 @@ def self.csv(patients)
           :address,
           :medical_history,
           :current_prescription_drugs,
-          :latest_bp_passports,
```
kitallis, line 34: I think we might still be able to `includes` the unordered versions of `bp_passports`, `appointments`, `blood_pressures` and `blood_sugars`. wdyt?

harimohanraj89: I think we could. However, to minimize the number of changes we make while fixing this bug, I figured we'd get things back to where they were (no working query optimizations) before we take another look. Wdyt?

kitallis: I think it's ok, I'm only suggesting it because the line list is notoriously slow and getting rid of these might exacerbate it – we could consider waiting and watching, but it is an actively used feature.

I'll leave it up to you to make a call.

kitallis: I think it's ok, I'm only suggesting it because the line list is notoriously slow and getting rid of these might exacerbate it – we could consider waiting and watching, but it is an actively used feature.

I'll leave it up to you to make a call.

harimohanraj89: The tradeoff for speed is memory usage, so I don't wanna make those changes just yet. 👍

kitallis: Makes sense, I guess we need a deeper reconstruction of this mechanism, simple trade-offs don't apply.

PR #1110, opened 2020-06-26
Delete emails of deleted admins

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/173559305

## Because

Currently, when an admin is destroyed, their email authentication record stays around, preventing us from being able to re-invite them, due to uniqueness constraints.

## This addresses

Cascadingly deleting associated `user_authentications` and `email_authentications` when an admin is destroyed. 

* We cannot use `dependent: :destroy` because email auths are associated to users through a join table, and `dependent: :destroy` does not destroy records joined this way, it only destroys the join records themselves.
* Ideally, admins should be soft-deleted. We are consciously punting that for now. It seems like a good task to take up when we work on dashboard audit logs.


app/models/user.rb
```diff
@@ -73,6 +73,8 @@ class User < ApplicationRecord
     :authenticatable_salt,
     :invited_to_sign_up?, to: :email_authentication, allow_nil: true
 
+  after_destroy :destroy_email_authentications
```
vkrmis, line 76: How about :phone_number_authentications? Should we figure out a deletion strategy for non-admin users?

harimohanraj89: We could/should, but I've chosen to keep this fix tactical, addressing just the use-case we understand thoroughly today.

On the surface, I'm less concerned with app users since we have a robust "Allow/Deny Sync" mechanism for em.

PR #1113, opened 2020-06-26
Get all registrations with one query for new dashboard, up to 24 months of data

rsanheim:
Pivotal card: https://www.pivotaltracker.com/story/show/171110822

This retrieves all registrations for selected facilities with one query, up to 24 months of data. We then use the registration data to determine how far back to go to determine control data. This also means we make one query for registration counts where previously we were doing 1 for every month.

I tried to use the date_to_period_sql for this but ended skipping it, as it seemed to make things more complicated without any real benefit.  

I may need to dig into https://phili.pe/posts/timestamps-and-time-zones-in-postgresql/ to
understand how timezones are working with PostgreSQL here, when I have some serious dedicated time to wrap my head around it.


app/services/district_report_service.rb
```diff
@@ -20,25 +23,52 @@ def call
...
+
+  def registration_counts
+    where_clause = ActiveRecord::Base.sanitize_sql_array([
+      "registration_facility_id in (?) and medical_histories.hypertension = ?",
+      @facilities, "yes"
+    ])
+
+    @registration_counts ||= Patient.connection.select_all(<<-SQL)
+      WITH cte AS (
+        SELECT date_trunc('month', "recorded_at") AS month, count(*) AS month_ct
```
kitallis, line 57: I think you'd potentially need to make `date_trunc` timezone sensitive here since we've stepped out of the rails world.

It's easy to show why `date_trunc` doesn't behave the way one would think it does by a few simple queries:

1. Regular `date_trunc`:

```
> select id, (DATE_TRUNC('month', recorded_at)) from blood_pressures limit 1;

                  id                  |     date_trunc
--------------------------------------+---------------------
 9cb6b949-f134-4abe-b0a4-47334c585b55 | 2019-11-01 00:00:00
```

2. `date_trunc` with time-zone resolution

```
> select id, (DATE_TRUNC('month', recorded_at)) AT TIME ZONE 'Asia/Kolkata' from blood_pressures limit 1;

                  id                  |         timezone
--------------------------------------+---------------------------
 9cb6b949-f134-4abe-b0a4-47334c585b55 | 2019-11-01 00:00:00+05:30

```

3. `date_trunc` with accurate timezone conversion

```
select id, (DATE_TRUNC('month', (recorded_at::timestamptz) AT TIME ZONE 'Asia/Kolkata')) AT TIME ZONE 'Asia/Kolkata' from blood_pressures limit 1;
                  id                  |         timezone
--------------------------------------+---------------------------
 9cb6b949-f134-4abe-b0a4-47334c585b55 | 2019-11-01 00:00:00+05:30
```

You can clearly see the difference in output between 1 & 2/3. `date_trunc` effectively kicks out the timezone by cutting up the date against UTC (default session tz) and returns a `timestamp without timezone`

2 & 3. are invisibly confusing but they mean different things.

2 means that it truncates the stored database value to a month (which is still in UTC) and then **assumes** the value in IST and converts it to timestamptz.

3 means that it **interprets** the value in the session timezone (UTC), convert it to IST, truncate it to a month and then convert it back to a timestamp with tz.


The subtle difference between 2. and 3. is that 2. assumes a UTC date to be in IST, 3. **corrects** the date to be in IST (just like Rails would). 1. doesn't do anything, it treats it in UTC. 

This is the idea behind `date_to_period_sql`.

Errors from this are usually seen when you hit month (or any period) boundaries.

app/services/district_report_service.rb
```diff
@@ -20,25 +23,52 @@ def call
...
   end
 
+  def lookup_registration_count(date)
+    lookup_date = date.beginning_of_month.to_date.to_s
+    row = registration_counts.find { |r| r["date"] == lookup_date }
+    return 0 unless row
+    row["running_ct"].to_i
+  end
+
+  def registration_counts
```
kitallis, line 49: Could this whole thing just be written as?

```
Facility.first.registered_patients.with_hypertension.group_by_period(:month, :recorded_at, last: 24).count
```

app/models/concerns/sql_helpers.rb
```diff
@@ -1,32 +1,28 @@
 module SQLHelpers
-  extend ActiveSupport::Concern
```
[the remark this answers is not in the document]
kitallis: Makes sense.

app/services/district_report_service.rb
kitallis: One caveat to be aware whilst using `group_by_period` and the likes is that it doesn't generate the in-betweener dates iff the min date in the database is much later than the point in time you're asking it to go back to, for eg:

if you want last 30 days of data, and you only have last 15 days in the database, it won't populate the first 15 days with zeroes.

The way I've typically handled this in the past is that I leave the wholesale date knowledge up to the view; the view knows what it needs to populate; the query knows what it currently has (it doesn't try to backfill) and then you smash the 2 together.

app/services/district_report_service.rb
rsanheim: I _think_ I follow all of that. 😁 

I think for this initial ship for beta 1, it will be fine to let it all be done in UTC.  Converting everything in this query to correctly be in IST was proving to be really tricky, and I don't think its worth the complexity to just have a series of high level running totals be displayed in IST vs UTC.  These aren't 'per minute' values, so I don't think having the month cutoff be UTC will even be noticeable while we are in a beta early-access period.

app/services/district_report_service.rb
rsanheim: > Could this whole thing just be written as?
> 
> ```
> 
> ```

It could, but we would need to add the appropriate association(s) on FacilityGroup to retrieve the counts for the set of facilities inside a facility group. We need these counts at the District (i.e. FacilityGroup) level, not Facility.

I'll add that to a list of followup improvements I'm keeping here: https://www.pivotaltracker.com/story/show/173575455

kitallis: Cool, that sounds good.

Fwiw, if you just need a wholesale count for a `FacilityGroup`, one could do:

```
FacilityGroup.first.patients.with_hypertension.group_by_period(:month, :recorded_at, last: 24).coun
```

Without any code-changes.

It'll be slightly more work if we need to preserve Facility IDs.

app/services/district_report_service.rb
kitallis: That seems fair.

app/services/district_report_service.rb
rsanheim: I realized now why none of these simpler solutions would work. We need a running total from the beginning of time of registered patients in a FacilityGroup. Groupdate makes it easy to get the totals for every month for a particular period, but it doesn't handle the cumulative total bit. 

I'm going to look into getting the counts via groupdate and then totaling via ruby to see how that looks.

PR #1114, opened 2020-06-26
Put dashboard v2 behind a Flipper feature flag

rsanheim:
**Story card:** Pivotal card: https://www.pivotaltracker.com/story/show/171110822

## Because

We only want the new dashboards to be turned on for a certain set of users. We also want to be turn on the new dashboard for **all** users when the real production ship date comes.

## This addresses

Putting the new v2 dashboards behind a flipper feature flag, and adding some convenience features to make it easier to turn on.  This includes showing all possible permission levels in the user admin edit screen, because other turning on the "flipper UI" has to be done via rails console, which seems less than desirable.

Here is the top nav bar with the "Dashboard v2" in the header, which will only show for users who have `dashboard_v2` turned on.  Non feature flagged users will see the normal nav bar.

image

The feature should look like this for a set of user(s):

image

### DIFF
--- a/app/controllers/dashboard/districts_controller.rb
+++ b/app/controllers/dashboard/districts_controller.rb
@@ -1,5 +1,6 @@
 class Dashboard::DistrictsController < AdminController
   layout "application"
+  before_action :feature_flag_required
   skip_after_action :verify_policy_scoped
   around_action :set_time_zone
 
@@ -26,6 +27,10 @@ class Dashboard::DistrictsController < AdminController
   end
 
   private
+
+  def feature_flag_required
+    redirect_to root_url unless current_admin.feature_enabled?(:dashboard_v2)
+  end
 
   def district_params
     params.permit(:selected_date, :id)
--- a/app/policies/manage/admin/user_policy.rb
+++ b/app/policies/manage/admin/user_policy.rb
@@ -40,7 +40,7 @@ class Manage::Admin::UserPolicy < ApplicationPolicy
   end
 
   def allowed_permissions
-    Permissions::ALL_PERMISSIONS.slice(*user_permissions).values
+    Permissions::ALL_PERMISSIONS.values
   end
 
   class Scope < Scope
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -66,7 +66,13 @@
                 <li class="nav-item">
                   <%= link_to "Dashboard", root_path, class: "nav-link #{active_controller?("organizations", "facility_groups")}" %>
                 </li>
+                <% if current_admin.feature_enabled?(:dashboard_v2) %>
+                <li class="nav-item">
+                  <%= link_to "Dashboard v2", dashboard_districts_path, class: "nav-link #{active_controller?("organizations", "facility_groups")}" %>
+                </li>
+                <% end %>
               <% end %>
+
 
               <% if policy(:dashboard).overdue_list? %>
                 <li class="nav-item">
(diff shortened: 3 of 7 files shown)


rsanheim:
Question for @simpledotorg/server-developers : is it possible to write a query that would grab all the User records who are CVHO's? If so, we could create a user group which would make turning this on or off much easier for all of them at once.

Without that, it may not be worth using flipper for this, as the hidden link approach sounds like it is working well enough.

kitallis:
> Question for @simpledotorg/server-developers : is it possible to write a query that would grab all the User records who are CVHO's? If so, we could create a user group which would make turning this on or off much easier for all of them at once.
> 
> Without that, it may not be worth using flipper for this, as the hidden link approach sounds like it is working well enough.

Barring any excursions from the default permissions of a CVHO (like custom permissions), we should be able to determine all Users with supervisor permissions like this,

```
cvho_access_level = Permissions::ACCESS_LEVELS.find { |access| access[:name].eql?(:supervisor) }[:default_permissions)
User.includes(:user_permissions).where(user_permissions: { permission_slug: cvho_permissions })
```

Correct me if I'm missing a simpler way to do this, @simpledotorg/server-developers.

app/policies/manage/admin/user_policy.rb
```diff
@@ -40,7 +40,7 @@ def allowed_access_levels
   end
 
   def allowed_permissions
-    Permissions::ALL_PERMISSIONS.slice(*user_permissions).values
+    Permissions::ALL_PERMISSIONS.values
```
harimohanraj89, line 43: Full review pending, but does this let users assign permissions to themselves/their peers that they shouldn't be allowed to? Eg. can a CVHO give themselves access to Flipper now? It depends on: Who all are allowed to manage users?

rsanheim: Good question - I don't know who all has the permissions to 'manage admins'. I guess I assume if someone has that ability, they are a super-admin or something along those lines?  Could someone w/ production access run a query to see how many people have access to manage admins?

rsanheim:
Note: this is **non-blocking** for the beta 1 release we want to get out by end of June. There are too many questions here on the production permissions data to ship this soon, and the hidden URL approach is good enough for now.

app/policies/manage/admin/user_policy.rb
harimohanraj89: According to our [permissions file](https://github.com/simpledotorg/simple-server/blob/master/app/policies/permissions.rb#L1), org admins, CVHOs, _and_ superadmins can do this, so it extends well beyond just us.

app/views/layouts/application.html.erb
```diff
@@ -66,8 +66,14 @@
                 <li class="nav-item">
                   <%= link_to "Dashboard", root_path, class: "nav-link #{active_controller?("organizations", "facility_groups")}" %>
                 </li>
+                <% if current_admin.feature_enabled?(:dashboard_v2) %>
+                <li class="nav-item">
+                  <%= link_to "Dashboard v2", dashboard_districts_path, class: "nav-link #{active_controller?("dashboard/districts")}" %>
```
harimohanraj89, line 71: Let's call this "Reports"? Let's check the text with design.

app/controllers/dashboard/districts_controller.rb
```diff
@@ -27,6 +28,10 @@ def show
 
   private
 
+  def feature_flag_required
+    redirect_to root_url unless current_admin.feature_enabled?(:dashboard_v2)
```
harimohanraj89, line 32: 404 if we wanna be super watertight, not give away that a feature is hidden here? :smile:

app/views/layouts/application.html.erb
harimohanraj89: In fact, I'd consider naming the feature flag `reports` too 🤔

rsanheim:
Going to close this until we have proper roles in the system, which will make working with Flipper much easier.

PR #1115, opened 2020-06-26
standardrb fix _most_ files in db and lib

rsanheim:
ignores schema.rb and one import file `to_patients.rb` that I think can go away


lib/tasks/create_admin_user.rake
```diff
@@ -1,17 +1,17 @@
...
+  abort "Requires <email>, <password>" unless args[:email].present? && args[:password].present?
 
   name = args[:name]
   email = args[:email]
   password = args[:password]
 
   begin
     CreateAdminUser.create_owner(name, email, password)
-  rescue StandardError => e
+  rescue => e
```
kitallis, line 14: TIL

PR #1118, opened 2020-06-29
Add Benchmarks to district reports

rsanheim:
**Story card:** https://www.pivotaltracker.com/story/show/171110822

This adds the 'benchmark' data for the district reports. We determine the top district by looking at all districts control rate, and selecting the district with the highest rate. The currently selected district is then compared to that one as the benchmark.

Branching off of https://github.com/simpledotorg/simple-server/pull/1113


app/services/district_report_service.rb
```diff
@@ -69,41 +98,40 @@ def registration_counts
...
-        no_bp: query.cohort_missed_visits_count,
-        uncontrolled: query.cohort_uncontrolled_bps.count
-      }.with_indifferent_access
-    end
+  def percentage(numerator, denominator)
+    return 0 if denominator == 0
+    (numerator.to_f / denominator) * 100
+  end
+
+  def top_district_quarter_stats(district)
```
rsanheim, line 110: This code is not pretty, but I want to get this first version live and work on the next levels of reports before really refactoring. This will need to handle facility level as well as possibly a future state level, and that will change how to best organize this code.

app/services/district_report_service.rb
```diff
@@ -30,11 +31,39 @@ def compile_control_and_registration_data
       time = selected_date.advance(months: n).end_of_month
       formatted_period = time.to_s(:month_year)
 
-      @data[:controlled_patients][formatted_period] = controlled_patients(time).count
+      @data[:controlled_patients][formatted_period] = controlled_patients_count(time)
       @data[:registrations][formatted_period] = lookup_registration_count(time)
     end
   end
 
+  # We want to return cohort data for the current quarter for the selected date, and then
```
rsanheim, line 39: This part is just moving existing code around.

app/services/district_report_service.rb
```diff
@@ -69,41 +98,40 @@ def registration_counts
...
-      .under_control
-      .order("patient_id, bp_recorded_at DESC, bp_id")
-      .where(registration_facility_id: facilities)
-      .where("(year = ? AND month = ?) OR (year = ? AND month = ?) OR (year = ? AND month = ?)",
-        beg_range.year.to_s, beg_range.month.to_s,
-        mid_range.year.to_s, mid_range.month.to_s,
-        end_range.year.to_s, end_range.month.to_s)
-    LatestBloodPressuresPerPatientPerMonth.from(sub_query, "latest_blood_pressures_per_patient_per_months")
+  def top_district_benchmarks
+    districts_by_rate = FacilityGroup.all.each_with_object({}) { |district, hsh|
```
rsanheim, line 126: TODO for future PR: figure out what orgs the current user has access to, and pass that into the service.

app/services/district_report_service.rb
```diff
@@ -69,41 +96,25 @@ def registration_counts
...
+      controlled = ControlledPatientsQuery.call(facilities: district.facilities, time: selected_date).count
```
timcheadle, line 110: Do we need to make sure we only look at hypertension patients here? Something like the `Patient.with_hypertension` scope that's being applied below.

I'm worried this latest BP view is picking up non-HTN patients.

rsanheim:
Follow up PR to make the registration counts piece simpler here: https://github.com/simpledotorg/simple-server/pull/1124

app/views/dashboard/districts/show.html.erb
```diff
@@ -243,96 +243,87 @@
...
           </div>
         </div>
-        <div class="d-lg-flex mb-lg-0 pb-lg-0">
-          <div class="mb-1 mb-lg-0 pt-lg-4 pb-lg-4 flex-lg-2 align-lg-center justify-lg-center br-lg br-lg-grey-light">
-            <p class="mb-0 ls-145">
-              Q2 cohort missed visit
+        <div class="mb-3 pb-3 d-lg-flex mb-lg-2 pb-lg-0">
+          <div class="mb-1 mb-lg-0 pt-lg-3 pb-lg-4 flex-lg-2 align-lg-center justify-lg-center">
+            <p class="mb-0 ls-145 c-grey-dark">
+              More benchmarks TBD
```
timcheadle, line 314:
```suggestion
              More benchmarks coming soon
```

app/services/district_report_service.rb
rsanheim: Fixed that up in 8a83bc04d943eb043ced2e010417159e8885e9a5

app/views/dashboard/districts/show.html.erb
rsanheim: Oh ya, this will get resolved in the next PR to land

rsanheim: er maybe not, i'll just merge it here

PR #1120, opened 2020-06-30
Bugfix: User deletion

harimohanraj89:
Fixes a bug in #1110 by eagerly loading the associated email authentication records before destroying the joining user authentication records.

### DIFF
--- a/app/jobs/seed_users_data_job.rb
+++ b/app/jobs/seed_users_data_job.rb
@@ -45,7 +45,7 @@ class SeedUsersDataJob
     @patient_traits ||=
       {
         registered_patient: {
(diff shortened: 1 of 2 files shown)


kitallis:
Just to expand on my original thought, could something like this work too?

```
  has_many :user_authentications, dependent: :destroy

  has_many :email_authentications,
    through: :user_authentications,
    source: :authenticatable,
    source_type: "EmailAuthentication"
    dependent: :destroy
```

harimohanraj89:
I explored that, but we can't use `dependent` for habtm relationships. It will only destroy the join records, not the `EmailAuthentication` records 😭

PR #1121, opened 2020-06-30
Make cron respect timezone

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/173219451

## Because

Timezone for cron jobs needs to be configurable, and cannot be read from `Rails.application.config.countries`, since the app environment is not loaded here.

## This addresses

Loads the Rails env and sets the cron timezone based on the country config.


harimohanraj89:
The env var is already populated in all environments, and the cron jobs will default to "Asia/Kolkata" if missing.

For Ethiopia, yes we are adding `CRON_TIME_ZONE` to the server configs in https://github.com/simpledotorg/deployment/pull/225

harimohanraj89:
Closing in favor of #1125

PR #1122, opened 2020-06-30
Create more historical data

harimohanraj89:
**Story card:** https://www.pivotaltracker.com/story/show/173558782

Adds 24 months of historical seed data in our test environments instead of 9.


app/jobs/seed_users_data_job.rb
```diff
@@ -45,7 +45,7 @@ def patient_traits
     @patient_traits ||=
       {
         registered_patient: {
-          time_fn: -> { Faker::Time.between(from: 9.month.ago, to: Time.now) },
+          time_fn: -> { Faker::Time.between(from: 24.months.ago, to: Time.now) },
```
rsanheim, line 48: It may be worth extracting this to a constant - `SEED_DATA_AGE` or something like that. Then we could easily expand or contract this for everything that needs to go back n months.

app/jobs/seed_users_data_job.rb
```diff
@@ -89,7 +89,7 @@ def per_patient_traits
     @per_patient_traits ||=
       {
         ongoing_bp: {
```
rsanheim, line 91: What is the difference between ongoing and retroactive here in terms of how seed data is created?

harimohanraj89: At first glance, I thought that `ongoing` creates BPs for patients that are "regularly returning to care", and `retroactive` creates BPs for patients that have "stopped coming back to care". I just noticed that we're omitting `recorded_at` from the latter, so there may be more going on.

kitallis: On second thought, I think removing the `recorded_at` is maybe not the best thing? I'm not sure.

Basically if you provide a `recorded_at`, it will set the patient registration date to be that if the BP is backdated (thereby modifying the patient trait that you are intentionally creating at certain dates)

If you don't provide a `recorded_at`, it will take the `device_created_at` anyway and modify it regardless.

I think we can just remove that exclusion, just adds confusion.

Since the app now always sends a recorded app for BPs anyway, it's also more realistic to not remove it and let the shape of the generated data take whatever form it needs to.

app/jobs/seed_users_data_job.rb
```diff
@@ -27,6 +27,9 @@ class SeedUsersDataJob
 
   SYNC_PAYLOAD_SIZE = 20
 
+  SEED_DATA_AGE = 24.months
```
kitallis, line 30: Non-blocking: I think these constants are ok, but the intention of the traits is to really have whatever curve you want to build for the various traits, so pulling them out to a couple of numbers isn't really that valuable, even if they are mostly the same for now.

app/jobs/seed_users_data_job.rb
```diff
@@ -45,8 +48,8 @@ def patient_traits
     @patient_traits ||=
       {
         registered_patient: {
-          time_fn: -> { Faker::Time.between(from: 9.month.ago, to: Time.now) },
-          size_fn: -> { rand(1..10) },
+          time_fn: -> { Faker::Time.between(from: SEED_DATA_AGE.ago, to: Time.now) },
+          size_fn: -> { rand(20..30) },
```
kitallis, line 52: Haha, these are a lot of patients? I hope the review apps can handle them.

PR #1127, opened 2020-07-01
Quick report fixes

claudiovallejo:
## Because

Making minor polishes and additions for initial Reports release

## This addresses

* Added alert disclaimer banner, "Report early access: Thanks for your patience as we improve this page"
* Updated 'Cumulative registrations' copy, "1,314 patients in Mar 2020" to match labels used within card
* Removed 'Benchmarks' card bottom spacing for small devices

**Screenshots**

image

image

### DIFF
--- a/app/views/dashboard/districts/show.html.erb
+++ b/app/views/dashboard/districts/show.html.erb
@@ -1,3 +1,6 @@
+<div class="alert alert-primary">
+  <span class="fw-bold">Report early access:</span> Thanks for your patience as we improve this page
+</div>
 <div class="dashboard">
   <nav class="breadcrumb">
     <%= link_to 'All reports', dashboard_districts_path %>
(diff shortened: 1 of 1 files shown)


app/views/dashboard/districts/show.html.erb
```diff
@@ -1,3 +1,6 @@
+<div class="alert alert-primary">
+  <span class="fw-bold">Report early access:</span> Thanks for your patience as we improve this page
```
rsanheim, line 2: I think we can make this alert a bit more clear:

```suggestion
  <span class="fw-bold">Early access:</span> This page is under active development, thanks for your patience as we work on improving it
```

claudiovallejo: Thanks for the feedback @rsanheim !

PR #1130, opened 2020-07-02
Test push from github to transifex

prabhanshuguptagit:
**Story card:** NA

## Because

Checking out if we can push translations from github to transifex. This is to figure out how to merge the upcoming help translations and deprecate the existing ones without breakage.


kitallis:
Test commit to figure out how exactly bidi sync in transifex works.

- Pushing a fake italian file

PR #1131, opened 2020-07-02
Refactoring report queries & services

rsanheim:
**Story card:** https://app.clubhouse.io/simpledotorg/story/229/allow-new-district-reports-to-drill-into-facility-level-details

## Because

We want our control rate data to be consistent and have clean back end code, so that we can add facility level reporting easily.

## This addresses

Refactoring to a single `ControlRateService` object, which we can re-use for gettng the rates for a district, a facility (forthcoming), and for the top district benchmark.

This also has quite a few other refactorings to prepare the way for facility level reports.


app/services/control_rate_service.rb
```diff
@@ -0,0 +1,93 @@
...
+    else
+      [region]
+    end
+    @range = range
+    @date = date
+    @end_of_date_range = date || range.end
+    logger.info "#{self.class} created for range: #{range} region: #{region.id} #{region.name}"
+  end
+
+  delegate :logger, to: Rails
```
harimohanraj89, line 22: Shwifty 😎

app/services/control_rate_service.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  def controlled_patients(time)
+    end_range = time.end_of_month
+    mid_range = time.advance(months: -1).end_of_month
+    beg_range = time.advance(months: -2).end_of_month
+    # We need to avoid the default scope to avoid ambiguous column errors, hence the `with_discarded`
+    # Note that the deleted_at scoping piece is applied when the SQL view is created, so we don't need to worry about it here
+    sub_query = LatestBloodPressuresPerPatientPerMonth
+      .with_discarded
```
harimohanraj89, line 74: Is `.unscoped` a more semantic way to achieve the same thing?

app/services/control_rate_service.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  def registrations(time)
+    registration_counts[time.beginning_of_month.to_date]
+  end
+
+  # Calculate all registration counts for entire range, or for the single date provided
+  def registration_counts
+    @registration_counts ||= if range
+      region.patients.with_hypertension
```
harimohanraj89, line 52: A region's expected to respond to `patients`, I assume? What kinds of regions are we going to support? For example, if we're using a facility or facility group (district), could we use a more explicit association? Eg. `region.registered_patients`? 

Hmm, I see that the `patients` association on facility groups does pull registered patients. I know it sounds silly, but could we set up another duplicate association called `registered_patients` that does the same thing? It gives us two benefits:
* Consistent and specific interface for this service to compute registrations
* Sets us up nicely to leave this service alone when we switch other touchpoints on the dashboard to start using `assigned_facility` for things.

harimohanraj89: Notably, the `Facility` model has both `patients` and `registered_patients` associations, and I think we want to use the latter here to compute control rate. The former returns all patients that have ever visited the facility. In fact, could you point me to the precise definition for this control rate? I want to make sure we're on the same page.

spec/services/top_region_service_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  it "gets top district benchmarks" do
+    darrang = FactoryBot.create(:facility_group, name: "Darrang", organization: organization)
+    darrang_facilities = FactoryBot.create_list(:facility, 2, facility_group: darrang)
+    kadapa = FactoryBot.create(:facility_group, name: "Kadapa", organization: organization)
+    _kadapa_facilities = FactoryBot.create_list(:facility, 2, facility_group: kadapa)
+    koriya = FactoryBot.create(:facility_group, name: "Koriya", organization: organization)
+    koriya_facilities = FactoryBot.create_list(:facility, 2, facility_group: koriya)
+
+    Timecop.freeze("April 1sth 2020") do
```
harimohanraj89, line 27: `1st`?

spec/services/district_report_service_spec.rb
```diff
@@ -94,19 +97,19 @@ def refresh_views
...
     expected_controlled_patients.default = 0
     expected_registrations = {
-      "Jan 2020" => 4, "Feb 2020" => 4, "Mar 2020" => 6, "Apr 2020" => 6, "May 2020" => 6, "Jun 2020" => 6
+      "Dec 2018" => 0, "Jan 2020" => 4, "Feb 2020" => 4, "Mar 2020" => 6, "Apr 2020" => 6, "May 2020" => 6, "Jun 2020" => 6
     }
     expected_registrations.default = 2
-    expect(result[:controlled_patients].size).to eq(18)
-    expect(result[:registrations].size).to eq(18)
+    # expect(result[:controlled_patients].size).to eq(19)
+    # expect(result[:registrations].size).to eq(19)
```
harimohanraj89, line 112: We can trash these assertions if we're not using them.

app/services/top_region_service.rb
```diff
@@ -0,0 +1,22 @@
+# Currently this object returns the "top district (aka facility group)" by control rate from a set of organizations.
+class TopRegionService
```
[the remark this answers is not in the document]
harimohanraj89: That seems fine to me. This thing helps us find the best performer for our benchmarks section?

app/services/control_rate_service.rb
```diff
@@ -0,0 +1,93 @@
...
+    end_range = time.end_of_month
+    mid_range = time.advance(months: -1).end_of_month
+    beg_range = time.advance(months: -2).end_of_month
+    # We need to avoid the default scope to avoid ambiguous column errors, hence the `with_discarded`
+    # Note that the deleted_at scoping piece is applied when the SQL view is created, so we don't need to worry about it here
+    sub_query = LatestBloodPressuresPerPatientPerMonth
+      .with_discarded
+      .select("distinct on (latest_blood_pressures_per_patient_per_months.patient_id) *")
+      .under_control
+      .with_hypertension
```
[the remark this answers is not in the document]
harimohanraj89: Do you mean the subquery hoops? I'd say you can go ahead and add the two diagnosis fields from the medication history model to this view right away and start using them. What do you think?

spec/services/district_report_service_spec.rb
rsanheim: Fixed these and added them back.

app/services/control_rate_service.rb
rsanheim: @harimohanraj89 you can find the definitions at this doc: https://docs.google.com/document/d/1MFRYAuy7pgXFZSH_p0YMLAIyGUldXB0KFYZg2jt8KiY/edit#heading=h.m8132ooz7gej -- look under the "Reports: State Level" heading for these new report definitions.

rsanheim: One thing to note is that we will want to get the patient status into the views as well, I'll add that as a card to update the materialized view(s) i'm using.  Here is that new story: https://app.clubhouse.io/simpledotorg/story/282/add-necessary-fields-to-materialized-views-for-district-reports

app/services/control_rate_service.rb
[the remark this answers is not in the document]
harimohanraj89: Totes. In that case, which hoops _did_ you have to jump through to get that scope working?

app/services/control_rate_service.rb
harimohanraj89: My thought was: If it's specifically the discarded scope that's throwing off this query, then `with_discarded` is good. If it's _any_ default scope throwing this thing off, then I'd prefer `unscoped`, as what we're really looking to do is reset.

Not a huge deal, I'll let you take a call! 👍

rsanheim: Ahh, I see, yeah I think the only thing throwing this off was the `with_discarded` piece. 👍

PR #1132, opened 2020-07-03
Assigned facility: API changes

prabhanshuguptagit:
**Story card:** https://www.pivotaltracker.com/epic/show/4556557

## Because

We want to define an **Assigned facility** which is the facility that is responsible for the patient’s follow-up care and control. 

## This addresses

These are just the API changes required for the app to begin treating the `assigned_facility_id` parameter as a first-class property. The changes required for the reports will come subsequently in other PRs.

- [x] Add `registration_facility` to patient responses
- [x] Add `assigned_facility` to the patient object in the server data model
- [x] Add data migration to back-fill `assigned_facility` = `registration_facility`
- [x] Add `assigned_facility` as an optional parameter to patient sync APIs
- [x] Add `registration_facility` as an optional parameter to patient sync APIs
- [x] Validate that user can access the `assigned_facility` and the `registration_facility` from the patient payload
- [x] Defaults to using facility from headers if `registration_facility_id` is not sent
- [x] Set `assigned_facility` = `registration_facility` if  `assigned_facility` from API is `null` or not present
- [x] Regenerate Swagger docs

Android QA test suite [run passed](https://github.com/simpledotorg/simple-server/runs/881801142?check_suite_focus=true) .


db/data/20200703072612_set_patient_assigned_facility.rb
```diff
@@ -0,0 +1,13 @@
+class SetPatientAssignedFacility < ActiveRecord::Migration[5.2]
+  def up
+    Patient.in_batches(of: 10_000) do |batch|
```
harimohanraj89, line 3: Consider adding `.where(assigned_facility_id: nil)`

app/controllers/api/v3/patients_controller.rb
```diff
@@ -42,15 +41,16 @@ def unscope_associations
...
     if validator.invalid?
+      logger.debug "Patient had errors: #{validator.errors_hash}"
       NewRelic::Agent.increment_metric("Merge/Patient/schema_invalid")
       {errors_hash: validator.errors_hash}
     else
-      patients_params_with_metadata = single_patient_params.merge(metadata: metadata)
-      transformed_params = Api::V3::PatientTransformer.from_nested_request(patients_params_with_metadata)
-      patient = MergePatientService.new(transformed_params).merge
+      transformed_params = Api::V3::PatientTransformer.from_nested_request(single_patient_params)
+      patient = MergePatientService.new(transformed_params, request_metadata: request_metadata).merge
```
kitallis, line 53: We explicitly pass in the `request_metadata` to the service instead of merging it into the patient params. This avoids conflating the `request_user_id` and `registration_user_id` (same for facility) which behaves differently for new patients and patients being updated.

The other reason is that it also clearly segregates the omnipresent "metadata" from the actual parameters by treating them as a different entity: `request_metadata`.

app/services/merge_patient_service.rb
```diff
@@ -1,64 +1,61 @@
...
+    @request_metadata = request_metadata
+    @existing_patient = Patient.with_discarded.find_by(id: payload["id"])
   end
 
   def merge
-    existing_patient_attributes = Patient.with_discarded.find_by(id: payload["id"])&.attributes
     merged_address = Address.merge(payload[:address]) if payload[:address].present?
 
-    patient_attributes = payload.except(:address, :phone_numbers, :business_identifiers)
+    patient_attributes =
```
kitallis, line 11: These transformations are better treated as a pipeline operations so that the order is always clear.

app/services/merge_patient_service.rb
```diff
@@ -1,64 +1,61 @@
...
-    end
+    merged_business_ids = merge_business_identifiers(payload[:business_identifiers], merged_patient)
 
-    if merged_patient.deleted_at.present? && existing_patient_attributes&.dig("deleted_at").nil?
-      # Patient has been soft-deleted by the client, server should soft-delete the patient and their associated data
-      # patient_attributes[:metadata][:registration_user_id] contains the current user's id
-      merged_patient.update(deleted_by_user_id: patient_attributes[:metadata][:registration_user_id])
-      merged_patient.discard_data
-    end
+    touch_patient(merged_patient, merged_address, merged_phone_numbers, merged_business_ids)
```
kitallis, line 23: Just methods pulled out by semantic meaning.

spec/services/merge_patient_service_spec.rb
```diff
@@ -0,0 +1,95 @@
+require "rails_helper"
+
+RSpec.describe MergePatientService, type: :model do
```
kitallis, line 3: This class is doing a lot without any tests, so we've written tests for most of its behavior since it was scary to change it otherwise.

app/services/merge_patient_service.rb
```diff
@@ -1,64 +1,61 @@
...
-    when :new
-      with_request_metadata
-    when Set[:updated, :old]
-      existing_metadata = Patient.find(patient_attributes[:id]).slice(*metadata_keys)
-      patient_attributes.merge(existing_metadata).except(:metadata)
-    else
-      patient_attributes.except(:metadata)
+  attr_reader :existing_patient, :payload, :request_metadata
+
+  def set_metadata(patient_params)
```
kitallis, line 33: This is really the method that has had the most impact with this refactor.

We keep the behavior where new patients get their metadata updated through either the headers or through the params and the patients being updated don't. We can change this behavior later easily when we want to update `registration_facility` or `registration_user`.

However, here now, this doesn't do any weird metadata pulling/de-structuring, it just refers to the `new_patient_metadata` or `existing_patient_metadata` functions respectively as needed.

app/services/merge_patient_service.rb
```diff
@@ -74,7 +71,42 @@ def merge_business_identifiers(business_identifier_params, patient)
...
+      #
+      # We can fix this issue, but it requires re-working the merge function, so we'll currently just
+      # track the incidence rate, so we can plan for a fix if necessary.
+      log_update_discarded_patient if patient.discarded?
+    end
+  end
+
+  def new_patient_metadata
+    {
+      registration_facility_id: payload[:registration_facility_id].presence || request_metadata[:request_facility_id],
```
kitallis, line 100: Here, now, it's explicit what the `request_facility_id` (headers) is and what `registration_facility_id` is.

spec/services/merge_patient_service_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+          )
+
+        payload = Api::V3::PatientTransformer.from_nested_request(patient_attributes)
+        merged_patient = described_class.new(payload, request_metadata: metadata).merge
+
+        expect(merged_patient[:assigned_facility_id]).to eq(registration_facility.id)
+      end
+    end
+
+    it "should discard_data when deleted_at exists and is not already deleted" do
```
kitallis, line 40:
```suggestion
    it "should discard_data when deleted_at exists and the patient is not already deleted" do
```

app/validators/api/v3/patient_payload_validator.rb
```diff
@@ -37,7 +43,35 @@ def past_date_of_birth
     end
   end
 
+  def user_can_access_assigned_facility
+    if assigned_facility_id.present? && !can_user_access_facility?(assigned_facility_id)
+      errors.add(
+        :assigned_facility_does_not_belong_to_user,
+        "Assigned facility must belong to the Facility Group of the User"
+      )
+    end
```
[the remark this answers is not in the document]
kitallis: @prabhanshuguptagit Let's in fact add the guard clause in `can_user_access_facility?`.

app/validators/api/v3/patient_payload_validator.rb
```diff
@@ -37,7 +43,35 @@ def past_date_of_birth
...
+  end
+
   def schema
     Api::V3::Models.nested_patient
   end
+
+  private
+
+  def can_user_access_facility?(facility_id)
+    request_user.present? && request_user.facility.facility_group.facilities.where(id: facility_id).present?
```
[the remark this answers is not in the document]
prabhanshuguptagit: Gave it a shot in 28e2238. I've not touched other parts that do similar checks to limit scope for now.

spec/services/merge_patient_service_spec.rb
rsanheim: ✨ ✨ ✨

app/services/merge_patient_service.rb
```diff
@@ -74,7 +71,46 @@ def merge_business_identifiers(business_identifier_params, patient)
...
 
-  def log_update_discarded_patient(merged_patient)
-    NewRelic::Agent.increment_metric("MergePatientService/update_discarded_patient") if merged_patient.discarded?
+  def discarded?(patient)
+    patient.deleted_at.present? && existing_patient&.deleted_at.nil?
+  end
+
+  def discard_patient_data(patient)
+    # Patient has been soft-deleted by the client, server should soft-delete the patient and their associated data.
+    patient.update(deleted_by_user_id: request_metadata[:request_user_id])
```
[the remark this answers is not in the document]
kitallis: I wouldn't want to touch this code at the moment since it's not a part of the PR, it's just been moved over structurally without logical changes.

However, we typically don't throw exceptions while syncing, errors are silently returned as 200 with an error map and retried by the client eventually (this happens through Validators). 

If there are things that could fail, we should validate them upfront: the `Mergeable` concern takes care of this. The object is already validated at this point. Not convinced that we should be defensive here, esp since this is a single patient out of a batch of patients that would end up failing if we raised wholesale without catching it appropriately.

db/data/20200703072612_set_patient_assigned_facility.rb
```diff
@@ -0,0 +1,13 @@
+class SetPatientAssignedFacility < ActiveRecord::Migration[5.2]
+  def up
+    Patient.where(assigned_facility_id: nil).in_batches(of: 10_000) do |batch|
```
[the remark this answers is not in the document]
prabhanshuguptagit: I did some benchmarking on local with ~150k records before and 10_000 went through smoothly (~15 seconds for 10_000 vs 25s for 1_000). Haven't done a memory benchmark but I think it should be fine on prod boxes as well.

app/models/user.rb
```diff
@@ -89,6 +89,10 @@ def registration_facility_id
 
   alias facility registration_facility
 
+  def facility_in_facility_group?(facility_id)
```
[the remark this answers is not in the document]
kitallis: The naming suggestions make sense, fixed. 👍🏽

As for the validations themselves, this is necessary so that the user cannot assign a `registration_facility` or an `assigned_facility` for a patient to a facility group that is outside their purview of access – this would be a data breach of a kind.

If the intention was to allow the user to assign a patient to any facility in the system, then it'd make sense to not have this validation, but since the validation is scoped to their facility group, it should be considered an invalid attempt from an API perspective (even though it's also disallowed from the UI in the app).

Does that seem fair?

prabhanshuguptagit:
> Do you know how long the migration will take to run? I imagine it will take a while to churn through all the patient records.

On local with ~150k patient this takes roughly 15 seconds. Production has ~250k patients.


> I would suggest splitting apart the migration from the code changes, so that we can deal with the migration piece and any possible problems isolated from the code. Once the migration is shipped, the code PR should be pretty quick and painless.

There isn't much change happening in this PR functionally other than the migration, the code changes will be only put to use a bit later by the app. In case there is breakage and we have to rollback, we will have to rollback both commits anyways. Also the split PRs will probably end up getting released simultaneously, not sure if there are much benefits to splitting it up .

app/models/user.rb
```diff
@@ -89,6 +89,10 @@ def registration_facility_id
 
   alias facility registration_facility
 
+  def authorized_facility?(facility_id)
+    facility.presence && registration_facility.facility_group.facilities.where(id: facility_id).present?
```
harimohanraj89, line 93: Consider `registration_facility.present?`

kitallis:
@harimohanraj89 If our client APIs were ready to communicate, it would make sense to split this. But since they aren't I'm ok shipping these together. However, I'm confused about this:

> If we ship together, there's a window of time between code deployment and migrations where our API will try to write to a field that doesn't exist yet, and will fail.

Our deploys first run migrations and then _activate_ the code right? This is always true for all cap deployments. And if the migration fails, the code is rolled-back along with the migration. The steps are following:

* Rsync code
* Schema migrate
* Data migrate
* Restart passenger and sidekiq to affect changes

I'm not sure how we'll reach a situation where the API is accepting stuff before the migrations are run?

harimohanraj89:
Oh! That's good news! In that case, you're right, we can launch this as one big blob if we're confident of the migration performance 👍

app/models/user.rb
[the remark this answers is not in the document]
rsanheim: > If the intention was to allow the user to assign a patient to any facility in the system, then it'd make sense to not have this validation, but since the validation is scoped to their facility group, it should be considered an invalid attempt from an API perspective (even though it's also disallowed from the UI in the app).
> 
> Does that seem fair?

Yup, that makes sense.

app/services/merge_patient_service.rb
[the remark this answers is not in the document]
rsanheim: Got it. I think assuming that updates or saves will always work, even after validations, is a dangerous assumption, because there are always other things that can go wrong beyond the individual record's data quality (database is backed up, fk constraints, race conditions, etc).

I agree that tackling that problem piecemeal as part of this PR does not feel warranted, though. 👍

rsanheim:
> Our deploys first run migrations and then _activate_ the code right? This is always true for all cap deployments. And if the migration fails, the code is rolled-back along with the migration. The steps are following:
> 
>     * Rsync code
> 
>     * Schema migrate
> 
>     * Data migrate
> 
>     * Restart passenger and sidekiq to affect changes
> 
> 
> I'm not sure how we'll reach a situation where the API is accepting stuff before the migrations are run?

Correct me if I'm wrong, but until you (or capistrano) runs the `deploy:restart` command, Passenger is still serving requests to the app from the old release. My concern w/ bundling up any complex db / data migrations with code changes is that your live code is dealing with a schema that is changing out from underneath itself. That could mean tables will be locked as migrations run, or it means that the schema Rails knows about is actually not the schema in the database anymore.

One way we can deal with that is just take the server down before running migrations - maybe that is part of our manual deploy process? That approach can be totally fine, especially if we can schedule off-hours delpoyments.

Another way to handle it is to just run migrations and hope that no-one is using the app against a locked table during that time =).  The mobile app will retry sync operations later, so maybe that covers most of the potential timeouts that would result.

For this PR in particular, the add columns are safe, but adding indexes [is not](https://github.com/doctolib/safe-pg-migrations#concurrent-indexes). It looks like we can make it safe if we run it outside of a transaction with `CREATE INDEX CONCURRENTLY` see https://thoughtbot.com/blog/how-to-create-postgres-indexes-concurrently-in.

Finally, one big reason for dividing up migrations from code changes is that the migrations often deserve very detailed review to consider data integrity and migration safety.  These concerns are only going to grow as our deployments and data set sizes grow. Managing PostgreSQL in production is fun yay!! 😁 😁 

Anyways, just wanted to try to explain my thoughts and reservations a bit more on this. Approving so that I'm not blocking progress here 👍 .

PR #1133, opened 2020-07-07
BD demo facility cleanup

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/268/bd-demo-facility-cleanup

## Because

The NFH would like us to deactivate their demo facility. We used this facility for training purposes.

## This addresses

* One task to:
  * Move real patients registered in the demo facility to Beanibazar (they are all supposed to be in Beanibazar)
  * Discard all test patient data in the demo facility
* Another task to:
  * Soft-Delete the facility
  * We will do this after the NHF has moved all users to real facilities


lib/tasks/scripts/delete_bangladesh_demo_facility.rb
```diff
@@ -0,0 +1,85 @@
...
+  private
+
+  def move_real_patients_to_beanibazar
+    log "Moving real patients to Beanibazar..."
+    real_patients.update(registration_facility: beanibazar)
+  end
+
+  def discard_test_patients
+    log "Discarding test patients..."
+    demo_facility.reload.registered_patients.where.not(id: REAL_PATIENT_IDS).each(&:discard_data)
```
kitallis, line 62: Not a big deal, but better to do `real_patients.pluck(:id)` over `REAL_PATIENT_IDS`?

spec/tasks/scripts/delete_bangladesh_demo_facility_spec.rb
```diff
@@ -0,0 +1,66 @@
+require "rails_helper"
+require "tasks/scripts/delete_bangladesh_demo_facility"
+
+RSpec.describe DeleteBangladeshDemoFacility do
+  describe ".handle_patients" do
+    it "moves the designated patients to Beanibazar and discards the rest" do
+      moveable_patient_ids = %w[
```
kitallis, line 7: Could just do `DeleteBangladeshDemoFacility::REAL_PATIENT_IDS` here?

lib/tasks/scripts/delete_bangladesh_demo_facility.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  private
+
+  def move_real_patients_to_beanibazar
+    log "Moving real patients to Beanibazar..."
+    real_patients.update(registration_facility: beanibazar)
+  end
+
+  def discard_test_patients
```
kitallis, line 60: What about the constituent dependencies for these test patients? Like BPs etc.?

harimohanraj89: The task invokes the `discard_data` method on patients, which takes care of this dependent data.

spec/tasks/scripts/delete_bangladesh_demo_facility_spec.rb
harimohanraj89: I didn't want the test to be self-fulfilling 😅  If the list of IDs changes for any reason, I want to make sure that we're aware of it through a failing test.

kitallis: Ah, gotcha, this is a regression list. Makes sense 👍

lib/tasks/scripts/delete_bangladesh_demo_facility.rb
kitallis: Keep forgetting this...

lib/tasks/scripts/delete_bangladesh_demo_facility.rb
harimohanraj89: Hehe I can't really reason for one over the other, so I'll leave as is 😅  Since this is a controlled task in a single env with specific IDs, I think it doesnt' really matter one way or the other.

PR #1136, opened 2020-07-08
Clean BD phone numbers

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/295/clean-up-bangladesh-phone-numbers

## Because

Over 90% of patient phone numbers in Bangladesh incorrectly include a leading zero.

## This addresses

Adding a rake task to remove the leading zero from ~9000 patient phone numbers.


spec/lib/tasks/scripts/clean_bangladesh_phone_numbers_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      expect(no_leading_zero_1.reload.number).to eq("1212121212")
+      expect(no_leading_zero_2.reload.number).to eq("3434343434")
+    end
+
+    it "does nothing in a dryrun" do
+      leading_zero_1 = create(:patient_phone_number, number: "01234567890")
+      leading_zero_2 = create(:patient_phone_number, number: "09876543210")
+
+      no_leading_zero_1 = create(:patient_phone_number, number: "1212121212")
+      no_leading_zero_2 = create(:patient_phone_number, number: "3434343434")
```
prabhanshuguptagit, line 27: These could go in a `let`.

harimohanraj89: They could 😄 I'm trying an Rspec approach that Rob showed me that minimizes `let` indirection by setting up test data inline. What do you think?

prabhanshuguptagit: The way I usually do these is to move up any obviously common data to the `lets` and setting up specific data inside the `its`. This _sort of_ seems to fit that bill.

PR #1137, opened 2020-07-08
Rehaul permissions (take 1)

vkrmis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/40/create-roles-and-permissions

## Because

## This addresses

- [x] Add `Access` model that encapsulates the Role and its scope
- [x] Convenience scope-resolution methods on `Access` (for facilities, facility_groups, orgs etc.)
- [x] Very simple data migration to show how we can migrate existing users
- [x] The [facilities](https://api-sandbox.simple.org/admin/facilities) page, the [My Facilities](https://api-sandbox.simple.org/my_facilities) page and the [facility groups](https://api-sandbox.simple.org/admin/facility_groups/new) page (including sub-pages from within) have been ported over to the new system for reference
- [x] Vet all the policy uses across the code-base and ensure that this system will work

This the model is structured in a way that the pundit usage is an extremely thin layer over our `Access` abstractions, it basically supports scoping by the relevant role (admin or non-admin).

The overall philosophy is the following:

* When you're authorizing certain things, be explicit about it in the controllers (we do this through pundit atm).
* When you're scoping data or authorizing pages, we **always** walk-up-the-tree, if you want `Facility`, check if you have Facility permission or higher... or higher and so on.
* We never walk-down the tree, for example, fetch the `Organization` if you have `Facility` access only. This choice would mean that we have to refactor a lot of our view code to keep our existing pages in parity.

##### write (or manage)
* For authorization, check the appropriate record for access.
* For scoping, return back whatever you can access on that scope.

##### read

* For authorizing, for aggregates, you just need *any* analyst permission. For PII, you just need *any* admin permission.
* For scoping, it keeps it really simple and just checks if you have any access.

#### Thoughts on [Take 2](https://github.com/simpledotorg/simple-server/pull/1160):

* That will basically be a carbon-copy of the `Access` abstraction, but with more code to handle authorization checks, which are currently in the policies.
* The scope resolvers in our implementation are dead simple, so they will more or less be ported over as-is.

#### Thoughts on pundit

* It's mostly a code-structuring device that's a nice-to-have for its well-understood APIs.
* IMO, it does not add additional complexity (in a [complecting](https://www.infoq.com/presentations/Simple-Made-Easy/) sort-of-way) to the code, just structure.
* It should be ok either way, even if we decide to not use it.

The API throughout the code-base is limited to just the following potentials:

```
authorize(:manage, record, :allowed?)
policy_scope(:manage, scope)
authorize(:read, :aggregates?)
authorize(:read, :identifiable_info?)
policy_scope(:read, scope)
```


app/controllers/admin/facilities_controller.rb
```diff
@@ -10,6 +10,7 @@ class Admin::FacilitiesController < AdminController
 
   def index
     authorize([:manage, :facility, Facility])
+    @admin = current_admin
```
kitallis, line 13: Find a better home for this.

app/models/facility_group.rb
```diff
@@ -18,6 +18,8 @@ class FacilityGroup < ApplicationRecord
   has_many :medical_histories, through: :patients
   has_many :communications, through: :appointments
 
+  has_many :user_resources, as: :resource
```
kitallis, line 21: Get rid.

app/models/user.rb
```diff
@@ -152,6 +153,14 @@ def has_permission?(permission_slug)
     user_permissions.where(permission_slug: permission_slug).present?
   end
 
+  def has_role?(*roles)
```
kitallis, line 156: Unroll this move.

app/models/organization.rb
```diff
@@ -6,6 +6,8 @@ class Organization < ApplicationRecord
   has_many :users
   has_many :protocols, through: :facility_groups
 
+  has_many :user_resources, as: :resource
```
kitallis, line 9: Get rid.

app/policies/upcoming/facility_policy.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  class Scope
+    attr_reader :user, :scope
+
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
```
kitallis, line 17: Simplify this scope.

app/policies/application_policy.rb
```diff
@@ -16,6 +16,10 @@ def user_has_any_permissions?(*permissions)
     end
   end
 
+  def user_has_any_roles?(*roles)
```
kitallis, line 19: Get rid.

app/models/facility.rb
```diff
@@ -25,6 +25,8 @@ class Facility < ApplicationRecord
   has_many :prescription_drugs
   has_many :appointments
 
+  has_many :user_resources, as: :resource
```
kitallis, line 28: Get rid.

app/models/user.rb
```diff
@@ -52,7 +53,7 @@ class User < ApplicationRecord
   scope :search_by_name_or_phone, ->(term) { search_by_name(term).union(search_by_phone(term)) }
 
   validates :full_name, presence: true
-  validates :role, presence: true, if: -> { email_authentication.present? }
+  validates :designation, presence: true, if: -> { email_authentication.present? }
```
kitallis, line 56: Add a migration.

app/controllers/admin/facilities_controller.rb
```diff
@@ -5,44 +5,35 @@ class Admin::FacilitiesController < AdminController
 
   before_action :set_facility, only: [:show, :edit, :update, :destroy]
   before_action :set_facility_group, only: [:show, :new, :create, :edit, :update, :destroy]
+  before_action :authorize_facility, only: [:show, :new, :create, :edit, :update, :destroy]
   before_action :initialize_upload, :validate_file_type, :validate_file_size, :parse_file,
     :validate_facility_rows, if: :file_exists?, only: [:upload]
 
   def index
-    authorize([:manage, :facility, Facility])
+    authorize([:upcoming, :manage, Facility], :allowed?)
```
kitallis, line 13: These policies are namespaced by Upcoming to not disturb the existing one on any way, it'll eventually look like,

```
authorize(:manage, Facility, :allowed?)
```

app/controllers/admin/facilities_controller.rb
```diff
@@ -5,44 +5,35 @@ class Admin::FacilitiesController < AdminController
...
 
-    if searching?
-      facilities = policy_scope([:manage, :facility, Facility]).search_by_name(search_query)
-      facility_groups = FacilityGroup.where(facilities: facilities)
+    facilities =
```
kitallis, line 15: We were previously scoping by `Organization` here, since our original permissions system could walk the scope tree bi-directionally, this is not a thing that we would do anymore for simplicity, that means however, we need to fetch the organizations back up top from the facilities in this case, rather than the other way around.

app/controllers/admin/facilities_controller.rb
```diff
@@ -85,7 +79,10 @@ def new_facility(attributes = nil)
 
   def set_facility
     @facility = Facility.friendly.find(params[:id])
-    authorize([:manage, :facility, @facility])
+  end
+
+  def authorize_facility
```
kitallis, line 84: Just pulled this out so it's clearly visible in the `before_action` callbacks.

app/controllers/my_facilities_controller.rb
```diff
@@ -96,7 +90,11 @@ def set_time_zone
     Time.use_zone(time_zone) { yield }
   end
 
-  def authorize_my_facilities
-    authorize(:dashboard, :view_my_facilities?)
+  def authorize_view
```
kitallis, line 93: This is this radical simplicity of the read permissions, the following two functions are all we need for all reporting / analytics related permissions everywhere.

app/models/access.rb
```diff
@@ -0,0 +1,38 @@
...
+  enum role: {
+    super_admin: "super_admin",
+    admin: "admin",
+    analyst: "analyst"
+  }
+
+  validates :role, presence: true
+  validates :resource_type, inclusion: {in: ALLOWED_RESOURCE_TYPES}, allow_nil: true
+
+  class << self
```
kitallis, line 16: These methods are the workhorse for all scoping/authorization related work. It's basically just a walk-up-tree type query on the individual levels.

app/models/access.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def facilities
+      resources_for(Facility)
+        .or(Facility.where(facility_group: FacilityGroup.where(organization: resources_for(Organization))))
+        .or(Facility.where(facility_group: resources_for(FacilityGroup)))
+    end
+
+    private
+
+    def resources_for(type)
```
kitallis, line 34: This is a really useful function, both now, and in the future, if we need to make exceptions to our current scoping system. It returns the actual resource object for an `Access` object.

app/models/user.rb
```diff
@@ -179,6 +180,10 @@ def resources
     user_permissions.map(&:resource)
   end
 
+  def super_admin?
```
kitallis, line 183: Just a helper on User to check if they are `super_admin?` This allows us to guard-clause/by-pass our way through any access check anywhere.

app/policies/upcoming/manage/facility_group_policy.rb
```diff
@@ -0,0 +1,27 @@
...
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
+  def allowed?
+    return true if user.super_admin?
+    user.accesses.admin.where(resource: resolve_record(record, FacilityGroup)).exists?
```
kitallis, line 11: This pattern is followed in all the policies for all managerial auth checks, notice that this is scoped by `admin`.

app/policies/upcoming/manage/facility_group_policy.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  class Scope
+    attr_reader :user, :scope
+
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
```
kitallis, line 22: This again, is a pattern repeated throughout all policies which just calls into the `Access` methods

app/policies/upcoming/read/facility_group_policy.rb
```diff
@@ -0,0 +1,15 @@
+class Upcoming::Read::FacilityGroupPolicy < Upcoming::ApplicationPolicy
+  class Scope
+    attr_reader :user, :scope
+
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
```
kitallis, line 10: This again, same thing, just not scoped by `admin`.

app/policies/upcoming/read_policy.rb
```diff
@@ -0,0 +1,9 @@
+class Upcoming::ReadPolicy < Struct.new(:user, :read)
```
kitallis, line 1: These are all the authorization checks necessary for read access, pretty straightforward.

app/policies/upcoming/read/facility_group_policy.rb
```diff
@@ -0,0 +1,15 @@
+class Upcoming::Read::FacilityGroupPolicy < Upcoming::ApplicationPolicy
```
kitallis, line 1: These are currently namespaced under `Manage` or `Read`, but they can be totally collapsed to be 1 policy if we just pass in read or write from the call-site as necessary.

app/views/admin/facility_groups/_form.html.erb
```diff
@@ -16,7 +16,10 @@
         </div>
       <% end %>
 
-      <%= form.select(:organization_id, organizations.map { |organization| [organization.name, organization.id] }) %>
+      <% unless organizations.empty? %>
```
kitallis, line 19: We're assuming that if you have FG admin permissions, you can edit your FG, but you can reassign the FG to a different Org. Based on this assumption, we'd removed the drop-down for Org select from FG edit if there are no Orgs that you can actually manage.

app/policies/upcoming/application_policy.rb
```diff
@@ -0,0 +1,29 @@
+class Upcoming::ApplicationPolicy
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
+  def resolve_record(record, record_type)
```
vkrmis, line 9: This method allows us to pass in any of the following to do authorization checks:
```
authorize(:upcoming, :manage, @facility)
authorize(:upcoming, :mange, Facility)
authorize(:upcoming, :manage, Facility.where(...))
```
When we pass a collection or class to an authorization check, the intention is to check if the user has access to _any_ records. This is useful when we want to authorize actions for a user that aren't necessarily tied to a single resource.

app/policies/upcoming/application_policy.rb
```diff
@@ -0,0 +1,29 @@
...
+  class Scope
+    attr_reader :user, :scope
+
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
+      scope.all
```
kitallis, line 26:
```suggestion
      return scope.all if user.super_admin?
```

app/policies/upcoming/application_policy.rb
```diff
@@ -0,0 +1,29 @@
...
+  class Scope
+    attr_reader :user, :scope
+
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
+      return scope.all if user.super_admin?
```
kitallis, line 26:
```suggestion
      if user.super_admin?
        scope.all
      else
        scope.none
      end
```

app/controllers/admin/facility_groups_controller.rb
```diff
@@ -51,7 +42,7 @@ def destroy
   private
 
   def set_organizations
-    @organizations = policy_scope([:manage, :facility, Organization])
+    @organizations = policy_scope([:upcoming, :manage, Organization])
```
rsanheim, line 45: Here is an alternate way we could write this filter which would apply to all controller actions:

```ruby
def require_organizations
  @organizations = current_admin.managed_organizations
  if @organizations.empty?
    render :not_authorized
  end
end
```

app/controllers/admin/facility_groups_controller.rb
```diff
@@ -60,7 +51,10 @@ def set_protocols
 
   def set_facility_group
     @facility_group = FacilityGroup.friendly.find(params[:id])
```
rsanheim, line 53: Here is a good example of the danger of doing first class model look ups - someone could craft a URL to directly hit the `destroy` action and blow away any FacilityGroup here. Maybe Pundit would catch this for us via its automatic filtering?  In any case, this could be written like this:

```
current_admin.facility_groups.friendly.find(params[:id])
```

app/models/access.rb
```diff
@@ -0,0 +1,38 @@
+class Access < ApplicationRecord
```
rsanheim, line 1: Suggestion: rename this to `Role` or maybe `UserResource`, as it is a join model between User and Resource.

app/models/access.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def facility_groups
+      resources_for(FacilityGroup)
+        .or(FacilityGroup.where(organization: resources_for(Organization)))
+    end
+
+    def facilities
+      resources_for(Facility)
+        .or(Facility.where(facility_group: FacilityGroup.where(organization: resources_for(Organization))))
+        .or(Facility.where(facility_group: resources_for(FacilityGroup)))
```
rsanheim, line 29: As much as I love ActiveRecord, for a query as crucial as this I think its often preferable to drop into SQL to make sure its doing what you want. Finding the right facilities is critical to our whole authorization system, and bugs here could have really negative impacts. Are you sure this kind of `or` chaining does the right thing if there are multiple hits along the chain?  For example, if a user has read access to a FacilityGroup and then write access to a Facility inside that group.

Compare this to a [SQL approach](https://github.com/simpledotorg/simple-server/pull/1173/files#diff-a9804da7022a39203c40f4586796c96fR31), where there are more lines of code but it is super clear what is going on:

```
SELECT facilities.* 
FROM facilities 
  INNER JOIN facility_groups ON facility_groups.id = facilities.facility_group_id 
  INNER JOIN organizations ON facility_groups.organization_id = organizations.id 
  INNER JOIN roles ON roles.resource_type = 'Organization' 
  AND roles.resource_id = organizations.id 
WHERE 
  roles.user_id = : user_id 
  AND roles.name in (: roles) 

UNION 

SELECT facilities.* 
FROM facilities 
  INNER JOIN facility_groups ON facility_groups.id = facilities.facility_group_id 
  INNER JOIN roles ON roles.resource_type = 'FacilityGroup' 
  AND facility_groups.id = roles.resource_id 
WHERE 
  roles.user_id = : user_id 
  AND roles.name in (: roles) 

UNION 

SELECT facilities.* 
FROM facilities 
  INNER JOIN roles ON roles.resource_type = 'Facility' 
  AND roles.resource_id = facilities.id 
WHERE 
  roles.user_id = : user_id 
  AND roles.name in (: roles)
```

app/policies/upcoming/read_policy.rb
rsanheim: Maybe primarily a Pundit gripe here: I don't understand how this fits in to the rest of the code here. Yes the two methods are very simple, but its magical and somehow lies outside the region scope you have created above.

app/views/admin/facilities/index.html.erb
```diff
@@ -2,10 +2,11 @@
   <h1 class="page-title">Facilities</h1>
 
   <div class="page-nav">
-    <% if policy([:manage, :facility, Facility]).upload? %>
+    <% if policy([:upcoming, :manage, FacilityGroup]).allowed? %>
```
rsanheim, line 5:
```
if current_admin.can_manage_facilities?
```

app/views/admin/facilities/index.html.erb
```diff
@@ -2,10 +2,11 @@
...
       <%= link_to 'Bulk add facilities', upload_admin_facilities_path, class: "btn btn-sm btn-default" %>
     <% end %>
-    <% if policy([:manage, FacilityGroup]).new? %>
+
+    <% if policy([:upcoming, :manage, current_admin.organization]).allowed? %>
```
rsanheim, line 9:
```
if current_admin.can_manage?(organization)
```

app/views/admin/facilities/index.html.erb
```diff
@@ -37,7 +40,7 @@
             <h3><span class="heading-label">Facility group</span><%= facility_group.name %></h3>
           </div>
 
-          <% if policy([:manage, facility_group]).edit? %>
+          <% if policy([:upcoming, :manage, facility_group]).allowed? %>
```
rsanheim, line 43:
```
if current_admin.can_manage?(facility_group)
```

app/policies/upcoming/manage/nil_class_policy.rb
```diff
@@ -0,0 +1,18 @@
+class Upcoming::Manage::NilClassPolicy < Upcoming::ApplicationPolicy
```
rsanheim, line 1: I like using null objects like this - would this play in if someone creates a policy with a nil object?

db/migrate/20200720085855_rename_column_on_users_from_role_to_designation.rb
```diff
@@ -0,0 +1,5 @@
+class RenameColumnOnUsersFromRoleToDesignation < ActiveRecord::Migration[5.2]
+  def change
+    rename_column :users, :role, :designation
```
rsanheim, line 3: Maybe a better rename here would be to `title` or `description` ?

db/migrate/20200708112930_create_user_accesses.rb
```diff
@@ -0,0 +1,16 @@
...
+    create_table :accesses, id: :uuid do |t|
+      t.references :user, null: false, index: true, type: :uuid, foreign_key: true
+      t.string :role, null: false, index: true
+
+      t.references :resource,
+        type: :uuid,
+        polymorphic: true,
+        index: {unique: true, name: "idx_accesses_on_resource_type_and_id"}
+
+      t.timestamps
```
rsanheim, line 12: Should be required.

db/migrate/20200708112930_create_user_accesses.rb
```diff
@@ -0,0 +1,16 @@
+class CreateUserAccesses < ActiveRecord::Migration[5.2]
+  def change
+    create_table :accesses, id: :uuid do |t|
+      t.references :user, null: false, index: true, type: :uuid, foreign_key: true
+      t.string :role, null: false, index: true
+
+      t.references :resource,
```
rsanheim, line 7: Would be really nice to make this required.  I think we can, it would mean introducing the [Region model](https://github.com/simpledotorg/simple-server/pull/1163) were the root node is an `application` which owns `organizations`.  Then a `super_admin` has `admin` rights on the `application`, and we don't need any special case logic anywhere! Everything would flow through the tree nicely.

app/controllers/admin/facilities_controller.rb
```diff
@@ -5,44 +5,35 @@ class Admin::FacilitiesController < AdminController
...
   def index
-    authorize([:manage, :facility, Facility])
+    authorize([:upcoming, :manage, Facility], :allowed?)
 
-    if searching?
-      facilities = policy_scope([:manage, :facility, Facility]).search_by_name(search_query)
-      facility_groups = FacilityGroup.where(facilities: facilities)
+    facilities =
+      if searching?
+        policy_scope([:upcoming, :manage, Facility]).search_by_name(search_query)
```
rsanheim, line 17: Suggestion:

```
current_admin.facilities.search_by_name(query)
```

app/controllers/admin/facilities_controller.rb
```diff
@@ -5,44 +5,35 @@ class Admin::FacilitiesController < AdminController
...
-      @organizations = policy_scope([:manage, :facility, Organization])
-      @facility_groups = @organizations.map { |organization|
-        [organization, policy_scope([:manage, :facility, organization.facility_groups])]
-      }.to_h
-      @facilities = @facility_groups.values.flatten.map { |facility_group|
-        [facility_group, policy_scope([:manage, :facility, facility_group.facilities])]
-      }.to_h
-    end
+    facility_groups = FacilityGroup.where(facilities: facilities)
+    @organizations = Organization.where(facility_groups: facility_groups)
```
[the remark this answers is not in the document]
kitallis: It's unclear to me what `orgs = current_admin.managed_organizations` is supposed to mean.

Is it supposed to mean,

1. all the organizations you can read even if you have Facility-level write permissions?
2. all the organizations you can read if you only have Org-level write permissions?

If it's 1.,

Then I don't think the [Take 3](https://github.com/simpledotorg/simple-server/pull/1173) queries will work here because you don't walk down the tree (Org -> FG -> F).

Even if it does that, it would be very confusing because you can't really **manage** those organizations if you only have Facility level permissions.

If it's 2.,

Then I don't think this page will work because this page needs to show all the facilities the user has access to and then group them by facility groups and then organizations; regardless of what level of write permissions they are at.

app/controllers/admin/facility_groups_controller.rb
kitallis: I don't think we can apply `require_organizations` blindly to all controllers since it'll depend on what page you are at and what `managed_organizations` actually means. See this [comment](https://github.com/simpledotorg/simple-server/pull/1137#discussion_r459891772).

app/controllers/admin/facility_groups_controller.rb
kitallis: I feel like this comment actually makes the point **for** Pundit in some ways, rather than against it. If it makes us squirm to look top-level model calls because who-knows-what-might-happen type scenarios, then we should probably invest in something that handles this for us automatically.

Pundit doesn't allow you to render a controller action if you don't authorize it, it'll fail in development, it'll fail the specs and it'll fail if you launch it in production. It'll behave the same way for the `destroy` action as well.

app/models/access.rb
kitallis: It used to be called `UserResource`, but we found it very unwieldy to do `user.user_resources` and felt like `Role` doesn't really capture what this model does and neither does `UserResource` – but I'd agree that it is a little more explicit in the parlance of how-to-name-join-models.

app/policies/upcoming/manage/nil_class_policy.rb
kitallis: Yup.

db/migrate/20200708112930_create_user_accesses.rb
kitallis: Yup, let's see how that pans out, we can always do that later depending on how we release things. The change shouldn't be very complicated.

app/controllers/admin/facilities_controller.rb
kitallis: Again, I'm not certain what `current_admin.facilities` means. It's not clear looking at [Take 3](https://github.com/simpledotorg/simple-server/pull/1173) how this would work.

Refer to [Take 2](https://github.com/simpledotorg/simple-server/pull/1160/files#diff-dbee6fa62dba5eb76fea1e205e2a5f54R52) to see a nicer implementation for this.

app/models/access.rb
kitallis: I disagree pretty strongly.

> Are you sure this kind of or chaining does the right thing

Yes, because we tested it and you can simply look at the SQL query to confirm it's true. Just because you write SQL does not mean your query is correct. 

For example, the query you pasted could very well be incorrect because you will end up querying potentially discarded facilities, organizations or facility groups.

If we don't trust our own abstractions, we shouldn't use them. Adding nuances like this actually makes *me* nervous.

Separately, I think this query can be **even** further simplified, take a look at [Take 2](https://github.com/simpledotorg/simple-server/pull/1160/files#diff-dbee6fa62dba5eb76fea1e205e2a5f54R52).

kitallis:
> Thanks for the detailed writeup and notes here - made this much nicer to review. 😄
> 
> I still feel like this is just too complicated and indirect and more code than we need. I really want to stress that we want to get to the simplest solution using the **least lines of code** (without resorting to golfing of course). Our authorization system is just not complicated enough to justify much of what is going on here. We basically have three roles and `read` and `write` access -- that is very bare bones and we should keep the auth code the same.
> 
> Here is one big example: maybe part of the problem here is Pundit, and if so we should rip it out to see how much less code this can be.
> 
> We have six policy objects that are boilerplate and are essentially doing the same thing. We could consolidate them down to one object, which reveals how simple our authoraization model really is, and we don't need Pundit at all. Something like this:
> 
> ```ruby
> class Upcoming::RegionPolicy < Upcoming::ApplicationPolicy
>   class Scope
>     attr_reader :user, :scope, :region_type, :access_level
> 
>     def initialize(user, scope, region_type, access_level: :read)
>       @user = user
>       @scope = scope
>       @region_type = region_type.underscore.to_s
>       @access_level = access_level
>     end
> 
>     def resolve
>       return scope.all if user.super_admin?
>       role = access_level == :read ? "analyst" : "admin"
>       accessible_regions = user.accesses.public_send(region_type), role: role)
>       scope.where(id: accessible_regions)
>     end
>   end
> end
> ```
> 
> Once you've done that, here is the actual part of that object doing work:
> 
> ```
>  accessible_regions = user.accesses.public_send(region_type), role: role)
>  scope.where(id: accessible_regions)
> ```


Agreed 👍 , we did consider collapsing it into 1 policy and it possibly makes sense to do it that way too ([ref](https://github.com/simpledotorg/simple-server/pull/1137#discussion_r458623197)). This attempt was simply to show what a 1:1 migration without a super-invasive change would look like, doing it this way would mostly allow us to do a surgical change to the existing code-base (we have ~25 controllers and ~50 views to touch).

[Here's](https://github.com/simpledotorg/simple-server/pull/1160) an attempt at the non-pundit version.

app/controllers/admin/facilities_controller.rb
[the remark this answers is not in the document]
harimohanraj89: Looks like we fetch a list of orgs just to bucket the way we render a list of authorized facilities on this page. We could do the same thing in a less twitchy way with `facilities.includes(facility_group: :organization).flat_map(&:organization)`, since `facilities` is an authorized list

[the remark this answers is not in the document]
harimohanraj89: In other words, we're not looking to fetch an authorized list of organizations here, we're just enriching an authorized list of facilities with org data so we can render it all fancy.

app/models/access.rb
harimohanraj89: I would also lean towards leveraging ActiveRecord to perform these queries ☝️ Whether we use AR or SQL, we're gonna have to unit test this thing to the moon for maximum confidence. I don't think _how_ we implement makes it easier/harder to test, and AR has some advantages as Kitty mentions.

app/models/access.rb
harimohanraj89: I'm not a fan of `UserResource`, but I think that's because I'm not a big fan of `Resource` in the first place. Also, there's enough behavior in here that I'd be ok deviating from the usual join-table naming in favor of something more semantic. Does this read better if it's an `Access` that has a `scope` instead of a `resource`? Resource seems very generic, whereas as scope is commonly used in the context of authorization.

app/models/access.rb
```diff
@@ -0,0 +1,38 @@
+class Access < ApplicationRecord
+  ALLOWED_RESOURCE_TYPES = %w[Organization FacilityGroup Facility].freeze
+
+  belongs_to :user, required: true
+  belongs_to :resource, polymorphic: true, optional: true
```
harimohanraj89, line 5: With this model, if I need to be an admin for Bathinda and Mansa, would my `User` record have two `Access` records? Both with `role: "admin"`, and one with each FG as the `resource`?

app/policies/upcoming/manage/facility_group_policy.rb
```diff
@@ -0,0 +1,27 @@
...
+      return scope.all if user.super_admin?
+      scope.where(id: user.accesses.admin.facility_groups)
```
harimohanraj89, line 24: So, all the `manage` policies will use `user.access.admin`, and all the `read` policies will use `user.accesses.analyst`?

harimohanraj89: Also, in the case of a policy scope for, say, patient PII, would this look something like:

```ruby
scope.where(registration_facility_id: user.access.manage.facilities)
# or, after assigned facility is done...
scope.where(assigned_facility_id: user.access.manage.facilities)
```

app/views/admin/facility_groups/_form.html.erb
harimohanraj89: Hehe neat optimization

db/data/20200716074943_seed_accesses.rb
```diff
@@ -0,0 +1,11 @@
+class SeedAccesses < ActiveRecord::Migration[5.2]
+  def up
+    admin = EmailAuthentication.find_by_email("admin@simple.org").user
+    admin.accesses.admin.create!(resource: FacilityGroup.first)
```
harimohanraj89, line 4: Is this just a starting point for testing? If so, could we just do this programmatically and avoid a data migration? I'd like to use data migrations for when we port everyone's permissions over, for example.

db/migrate/20200720085855_rename_column_on_users_from_role_to_designation.rb
harimohanraj89: Can we ask Tanushree what users use this field for? Is it useful?

harimohanraj89:
Phew, I think I get what's going on now! Here are my top takeaways:

* The toolbelt (those resource grabbing methods) on the `Access` model looks great
* I like using AR for those methods as well. Either way we'll have to unit test em to pieces.
* I'm not a big fan of the Pundit interface and the amount of boilerplate code we'll need to use, so I'm in favor of introducing a method set directly on `User` that we can use instead.

```
class User
  def manageable_facilities
    accesses.admin.facilities
  end
  
  def viewable_facilities
    accesses.facilities
  end

  def can_manage_facility?(facility=nil)
    # if facility is nil, then can you manage _any_ facilities?
    # else, can you manage _this_ facility?
  end
```
etc. 

While the initial method set will be kinda large (one set for org, fg, and facility each, probably), I feel like I'd have a much easier time setting up a new Dashboard page and stuff using this tookit. What do you think?

app/models/access.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  enum role: {
+    super_admin: "super_admin",
+    admin: "admin",
+    analyst: "analyst"
```
harimohanraj89, line 10: Hmmmmm are these really roles still? Should we stick to `admin` and `analyst`, or go with, say, `view` and `manage`?

harimohanraj89: Actually, on second thought, I think we should _keep_ this as roles! Here's something I really liked from [take 3](https://github.com/simpledotorg/simple-server/pull/1173/files#diff-a9804da7022a39203c40f4586796c96fR3-R7).

There's an explicit central spot where we translate which roles are allowed to manage which resources, and which roles are allowed to view which resources.

I think in this PR, we currently distribute that knowledge across several policies by selecting the appropriate `Access` scope to use (`accesses.admin` or just `accesses`). I think a more explicit capturing of this would be awesome.

app/controllers/admin/facilities_controller.rb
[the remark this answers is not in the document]
rsanheim: > It's unclear to me what `orgs = current_admin.managed_organizations` is supposed to mean.
> 
> Is it supposed to mean,
> 
>     1. all the organizations you can read even if you have Facility-level write permissions?
> 
>     2. all the organizations you can read if you only have Org-level write permissions?

Hm, maybe this is part of the disconnect here. If you are asking a current admin for the organizations they manage, its literally just that: what orgs does this user have the ability to manage (i.e update and write to).  If they can manage an org, the reading is implicit, because of our auth model says that manage rights include read rights on that particular region.

The idea behind a `managed_organizations` scope is to _only_ return orgs that a user manages for the purposes of allowing them to add facility groups under that org, or to allow them to edit the org metadata, or whatever.

I may have misunderstood what this specific page is doing, and this kind of scope would probably be better if this page lets them maintain facilities they have access to:

```
current_user.managed_facilities
```

On the other hand, if the page is more for display / reading, or if we want to return _all_ facilities a user can read and then determine edit level access in the view, we could do this:

```
current_user.readable_facilities
```

or just
```
current_admin.facilities
```

Then in the view, psuedocode like this:

```
facilities.each do |facility|
  if current_admin.can_manage?(facility)
    <%= form_for facility ... %>
  else
    # just do the read-only logic
    <%= facility.name %> ...
  end
end
```

Does that help clear things up?  The point is that the outer domain layer itself can handle returning the appropriate scoped objects, and that is all that developers should have to worry about when writing new controllers or when updating actions. Nothing about `Facility.where` or `Organization.where` tells me how records should be scoped to an admin, where as an api like `current_admin.get_me_the_things_I_can_manage` or `current_admin.get_me_things_I_have_read_access_to` does.

app/controllers/admin/facility_groups_controller.rb
rsanheim: I meant just this controller's actions.  Not _all_ controllers. I agree it depends on the context 100% !

app/controllers/admin/facility_groups_controller.rb
rsanheim: > If it makes us squirm to look top-level model calls because who-knows-what-might-happen type scenarios, then we should probably invest in something that handles this for us automatically.

Yes, exactly! Thats why we are adding a concept of 'role' or 'access' to our domain model, and then we can rely on the database and ActiveRecord to "handle this for us automatically" 😀😀😀 .  

We have the tools in our framework to handle many of these cases, its just SQL and good data modeling and then using the ORM we have.  No need for another auth-specific framework if our use case is right in the sweet spot of literally what Rails and ActiveRecord were built for.

app/models/access.rb
rsanheim: I'm thinking `region` is a name that makes more sense ...an Admin has access to a Region via a particular Role. 

I agree that Resource feels too generic right now...

app/models/access.rb
rsanheim: I agree that however we write this, it should be tested like crazy. 👍 

I can go either way on how this is written, my larger concern is that the external API for permissions is as simple as possible and doesn't require a developer to come looking at the underlying implementation to figure out what is going.

kitallis:
@simpledotorg/server-developers Closing this PR in favor of https://github.com/simpledotorg/simple-server/pull/1160, I'll however keep referencing the conversation we had here for further changes.

PR #1150, opened 2020-07-10
Add caching to region reports:

rsanheim:
Caches the resulting hash structures in Redis for the main services that return data for the new RegionReports. This drops page load times from 4 to 10+ seconds to around ~100 ms per page - see some comparison below.  The caches will expire based on the `updated_at` of the top level region being retrieved, or after a 7 day expiration.  We can change that 7 day expiration of course, I just picked it because it felt like a safe number.

Because I'm caching the results from individual objects such as `ControlRateService`, we will benefit from the perf gains across other reports that reuse these same objects. For example, loading a single Facility Report ends up priming caches for all other Facilitys' control rates for the selected date period, as it needs to grab them all for the top region benchmark. This means that the first user in may see one or two slow page loads in a single org, but after that all subsequent users' requests should be hitting the cache. We could add our own priming, but don't feel its worth it yet without trying this out first.

Also adding [rack mini profiler](https://github.com/MiniProfiler/rack-mini-profiler) to get some development-only insights into where our time is going in expensive pages. This displays the page load time in the upper left corner in dev, and you can click to get more detailed information.  

I'm also adding [request_store](https://github.com/steveklabnik/request_store) to allow us to easily bust the cache from a web request by passing a query string of `?force_cache=1`. The RequestStore provides a thread safe store that is reset after every web request or sidekiq job.  Without this we would have to thread the state through many services for these complicated reports, and given caching is a cross-cutting concern it feels fine to use a 'global-like' object for the `force_cache` state.

**Story card:** https://app.clubhouse.io/simpledotorg/story/330/speed-up-the-new-district-facility-reports

## Without Caching (FacilityGroup)

image

## With Caching (FacilityGroup)

image

## Without Caching (Facility)

image

## With Caching (Facility)

image


app/services/control_rate_service.rb
```diff
@@ -82,6 +85,26 @@ def controlled_patients(time)
...
+  def date_or_range_cache_key
+    if range
+      "#{range.min.to_s(:iso8601)}/#{range.max.to_s(:iso8601)}"
+    else
+      date.to_s(:iso8601)
+    end
+  end
+
+  def cache_version
+    "#{region.updated_at.utc.to_s(:usec)}/#{CACHE_VERSION}"
```
kitallis, line 101: What's the reason for a microsecond-level precision here?

app/controllers/reports/regions_controller.rb
```diff
@@ -10,8 +10,9 @@ def index
   end
 
   def show
-    @region = Facility.find_by!(slug: facility_params[:id])
+    @region = scope.find_by!(slug: facility_params[:id])
     authorize(:dashboard, :show?)
+    RequestStore.store[:force_cache] = true if force_cache?
```
kitallis, line 15: Actually, now that I look at this, it seems like there's literally only one layer where you'd need to pass `force_cache: true`, which is from the controller to the service. Not sure if `RequestStore` is really that valuable here, since you also need to manage setting and reading it anyway.

I imagine, defaulting to `cache:true` and setting it to false when the param is set is nearly as much code and probably simpler?

rsanheim: The problem with passing the param down is that you then need to pass it into all the dependant services. That results in one more param for every initialize call for something that really is a cross-cutting concern, and not directly related to the data we are retrieving. 

If you look at an earlier commit - https://github.com/simpledotorg/simple-server/pull/1150/commits/2e260c819610854b01eb136857aaaa083dd0d808 - you'll see the ugliness of passing around storing all that additional instance state, as opposed to using RequestStore.

app/controllers/reports/regions_controller.rb
```diff
@@ -30,8 +31,23 @@ def show
 
   private
 
+  def scope
```
timcheadle, line 34: I dig this abstraction work.

app/services/region_report_service.rb
```diff
@@ -42,24 +43,40 @@ def compile_control_and_registration_data
...
+          uncontrolled: query.cohort_uncontrolled_bps.count
+        }.with_indifferent_access
+      end
     end
   end
 
+  private
+
+  def cohort_cache_key
+    "#{self.class}/cohort_trend_data/#{region.model_name}/#{region.id}/#{organizations.map(&:id)}/#{selected_date.to_s(:iso8601)}"
```
timcheadle, line 69: Nitpick: isn't organization higher in the hierarchy than region? This seems backwards.

harimohanraj89:
I like this use of `request_store` 👍 Looks slick!

app/controllers/reports/regions_controller.rb
kitallis: Ah, that makes sense, the diff didn't capture the ugly enough.

I guess the only thing to be wary of is using these service classes is to remember to globally and appropriately set the `RequestStore.store[:force_cache]` wherever it is called (non-controller usage, tests etc.) when we want a cached response?

app/services/region_report_service.rb
```diff
@@ -42,24 +43,40 @@ def compile_control_and_registration_data
   # the previous three quarters. Each quarter cohort is made up of patients registered
   # in the previous quarter who has had a follow up visit in the current quarter.
   def compile_cohort_trend_data
-    Quarter.new(date: selected_date).downto(3).each do |results_quarter|
-      cohort_quarter = results_quarter.previous_quarter
+    Rails.cache.fetch(cohort_cache_key, version: cohort_cache_version, expires_in: 7.days, force: force_cache?) do
```
kitallis, line 46: If my [understanding](https://github.com/simpledotorg/simple-server/pull/1149#issuecomment-658212422) of `cache_version` is correct,

Recyclable cache keys are valuable if the model is constantly _touched_, which is a typical CRUD case for Rails; Facilities almost never get updated; which means this will almost always cache for 7 days.

7 days feels like a long time to cache (even monthly) reports since it's not _that_ much data to show, we can probably afford to be more aggressive since these are just aggregates, so it's not in the order of of 100s of MBs (correct me if I'm wrong). Also since the matviews backing this refresh daily anyway, caching these for 7 days isn't that helpful, we could simply just cache for 2 days to match up with the worst-case scenario of view refresh to report cache refresh time.

Let's check with @dburka or @timcheadle on whether it's ok to do this more frequently?

rsanheim: > Let's check with @dburka or @timcheadle on whether it's ok to do this more frequently?

Sure, we can change that expire time to whatever we want to after shipping this. We can also force whole districts from the web via the `force_cache` mechanism.

app/controllers/reports/regions_controller.rb
rsanheim: > I guess the only thing to be wary of using these service classes is to remember to globally and appropriately set the `RequestStore.store[:force_cache]` wherever it is called (non-controller usage, tests etc.) when we want a cached response?

Yup, I imagine this will primarily be used for busting the cache from the controller layer.  I can't imagine where we will often want to use it for non-web driven purposes, but it is there if we need it.

rsanheim: Oh, and to be clear, the default behavior should always be to fetch from the cache if `force_cache` isn't set...`force_cache` is only ever intended to mean "I know better than the caching mechanism, grab fresh data and populate the cache with it".

app/services/control_rate_service.rb
rsanheim: I was just using a builtin format that is all numeric and easy to read if we ever need to inspect or debug these keys.

app/services/region_report_service.rb
rsanheim: The only problem with shortening the time is that that means some user will hit the expiration case every 2 days and experience a slow page load first time in.  We could prefill the cache to avoid that if we wanted to.

app/services/region_report_service.rb
rsanheim: Yeah, was just following the order set in the other keys. Its not meant to be strictly hierarchical, just easy to read and adjust if need be.

PR #1154, opened 2020-07-14
Add a health check endpoint

harimohanraj89:
**Story card:** TBD

Adds a health check endpoint that we can wire up to simple availability alerts for our environments.


harimohanraj89:
We already have a health check endpoint for now.

PR #1158, opened 2020-07-16
Assigned facility view updates

prabhanshuguptagit:
**Story card:**
https://app.clubhouse.io/simpledotorg/story/404
https://app.clubhouse.io/simpledotorg/story/405

## Because

We want to port our reports to use the assigned facility field, which is currently not available in the materialized SQL views.

## This addresses

This adds patient's `assigned_facility_id` column to the reporting views. This also drops the `responsible_facility_id` column and related calculations from them. The changes are combined together to avoid updating the views multiple times.

Views touched:
- `LatestBloodPressuresPerPatientPerDay`
- `LatestBloodPressuresPerPatientPerMonth`
- `LatestBloodPressuresPerPatientPerQuarter`
- `LatestBloodPressuresPerPatient`

This doesn't add the field to `patient_summaries` yet, that can come in a separate PR with the line list query changes.


db/migrate/20200716070836_update_latest_blood_pressures_per_patient_per_months_to_version_4.rb
```diff
@@ -0,0 +1,11 @@
+class UpdateLatestBloodPressuresPerPatientPerMonthsToVersion4 < ActiveRecord::Migration[5.2]
+  def change
+    drop_view :latest_blood_pressures_per_patient_per_quarters, materialized: true, revert_to_version: 2
+    drop_view :latest_blood_pressures_per_patients, materialized: true, revert_to_version: 1
+
+    update_view :latest_blood_pressures_per_patient_per_months, version: 4, materialized: true, revert_to_version: 3
+
+    create_view :latest_blood_pressures_per_patient_per_quarters, materialized: true, version: 3
```
prabhanshuguptagit, line 8: Per quarter view: v2 is dropped and v3 is created, effectively updating it and avoiding an additional update migration.

db/views/latest_blood_pressures_per_patient_per_days_v02.sql
```diff
@@ -0,0 +1,21 @@
...
+    patients.registration_facility_id AS registration_facility_id,
+    patients.assigned_facility_id AS assigned_facility_id,
+    patients.status as patient_status,
+    blood_pressures.facility_id AS bp_facility_id,
+    blood_pressures.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE')) AS bp_recorded_at,
+    patients.recorded_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE')) AS patient_recorded_at,
+    blood_pressures.systolic AS systolic,
+    blood_pressures.diastolic AS diastolic,
+    blood_pressures.deleted_at at time zone 'utc' at time zone (SELECT current_setting('TIMEZONE')) AS deleted_at,
+    medical_histories.hypertension as medical_history_hypertension,
```
prabhanshuguptagit, line 13: This also copies over the fields added to the per month view in #1139.

prabhanshuguptagit:
> Why are we punting on the PatientSummary view?

Just to keep this PR lean and reporting-only. Also I think it will be easier to parse with the line list code changes.

PR #1159, opened 2020-07-17
Add github action to run QA suite

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/421

## Because

We want to run the android test suite against our review apps for integration/regression testing.

## This addresses

This adds an action to run the QA suite on our PRs.

Catches/things to fix:
- The suite takes tooo long to run (~15m on average, from historical data on the `simple-android` CI).
- The suite might start running before the app is deployed on heroku, it might be a rare occurrence though. One way to fix this is to make the heroku deploy a github action and sequence the order of operations.

**Update:**

Github recently added the ability to [manually trigger actions](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch). This will allow us to run the action as and when needed instead of running this on _every_ push.

To test out the manual trigger, it requires merging this PR. We should see a button like this once this action is on master:
image

Sample run of the action triggered via github's API: https://github.com/simpledotorg/simple-server/actions/runs/172493681


.github/workflows/android-qa-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+          repository: simpledotorg/simple-android
+          path: simple-android
+
+      - name: Setup JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Turn on capturing of flaky tests
+        run: cat simple-android/app/src/androidTest/resources/quarantine_ci.properties > simple-android/app/src/androidTest/resources/quarantine.properties
```
harimohanraj89, line 21: We don't need to capture flaky android tests, methinks.

.github/workflows/android-qa-tests.yml
```diff
@@ -0,0 +1,27 @@
+name: Android QA tests
+
+on: workflow_dispatch
```
harimohanraj89, line 3: This allows manual triggering?

.github/workflows/android-qa-tests.yml
prabhanshuguptagit: I copied it over from the [android CI checks](https://github.com/simpledotorg/simple-android/blob/master/.github/workflows/ci_checks.yml#L66), the assumption being these are some failing cases we shouldn't care about.

.github/workflows/android-qa-tests.yml
prabhanshuguptagit: Yes.

prabhanshuguptagit: Yes.

.github/workflows/android-qa-tests.yml
harimohanraj89: Ah, these _rescue_ flaky tests. I thought they capture them for reporting purposes when they fail. Let's confirm with the android team. If it rescues flakers, then yes, leave it in 👍

prabhanshuguptagit:
Thanks for the review! Going to merge and test this out, we'll take care of the flaky tests situation once @vinaysshenoy gets a chance to look at it.

.github/workflows/android-qa-tests.yml
vinaysshenoy: This rescues flaky tests in addition to reporting flaky tests at https://flakiness.dx.obvious.in/simple-android-instrumented-tests/reports

Note: seems like the SSL certificate has expired, which is weird since I used Certbot which _should_ automatically renew them.

.github/workflows/android-qa-tests.yml
vinaysshenoy: TIL that manual workflow runs exist.

vinaysshenoy:
Regarding the slow performance, I doubt there's a lot we can do to fix this immediately without putting some effort into test sharding and things like that.

It might make more sense to run this on a nightly schedule instead of on every PR.

PR #1160, opened 2020-07-17
Rehaul permissions (take 2)

kitallis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/40/create-roles-and-permissions

## Because

## This addresses

This is roughly the same as [Take 1](https://github.com/simpledotorg/simple-server/pull/1137), but attempts to remove pundit usages for all the cases touched upon in there.


app/models/user.rb
```diff
@@ -167,12 +176,92 @@ def self.requested_sync_approval
...
+                           .map(&:facilities))
+    read_agg_facility_groups =
+      Facility.where(id: accesses
+                           .where(role: [Role.admin.first, Role.analyst.first], resourceable_type: "FacilityGroup")
+                           .map(&:resourceable)
+                           .map(&:facilities))
+    read_agg_facilities =
+      Facility.where(id: accesses
+                           .where(role: [Role.admin.first, Role.analyst.first], resourceable_type: "Facility")
+                           .map(&:resourceable))
```
harimohanraj89, line 260: This can now collapse to

```
Facility.where(id: access
  .where(role: [Role.admin.first, Role.analyst.first])
  .map(&:resourceable)
  .map(&:facilities) 
)
```
because all three models return sensible results for `object.facilities`

app/models/user.rb
```diff
@@ -52,7 +53,7 @@ class User < ApplicationRecord
   scope :search_by_name_or_phone, ->(term) { search_by_name(term).union(search_by_phone(term)) }
 
   validates :full_name, presence: true
-  validates :role, presence: true, if: -> { email_authentication.present? }
+  validates :designation, presence: true, if: -> { email_authentication.present? }
```
harimohanraj89, line 56: Ah, we rename this to `designation` to avoid clashing with the new "roles"? Does this need a migration?

app/models/access.rb
```diff
@@ -0,0 +1,15 @@
+class Access < ApplicationRecord
+  ALLOWED_TYPES = %w[Organization FacilityGroup Facility].freeze
+
+  belongs_to :user
+  belongs_to :resourceable, polymorphic: true, optional: true
```
harimohanraj89, line 5: Can we just call this `resource`? Or `scope`?

app/models/role.rb
```diff
@@ -0,0 +1,7 @@
+class Role < ApplicationRecord
+  enum name: {
+    super_admin: "super_admin",
+    admin: "admin",
+    analyst: "analyst"
```
harimohanraj89, line 5: Let's think with Claudio about the wordages we use. Admin/analyst? Does that still make sense on a per-scope basis? What about view/manage? read/write? (ok maybe not read/write...)

app/models/user.rb
```diff
@@ -167,12 +176,92 @@ def self.requested_sync_approval
     where(sync_approval_status: :requested)
   end
 
-  def has_role?(*roles)
-    roles.map(&:to_sym).include?(role.to_sym)
+  def can_manage?(scope)
+    admin = Role.admin.first
+    
+    value =
+      case scope.class.model_name.name.to_sym
```
harimohanraj89, line 183: We could consider using those convenience methods to make this something like:

```
case scope.class.model_name.name.to_sym
when :Organization
  manage_organizations.include?(scope)
when :FacilityGroup
  manage_facility_groups.include?(scope)
when :Facility
  manage_facilities.include?(scope)
end
```

Of course, this assumes that the scope is always a single record. The method could have two parts, the second for when the scope is a relation or class.

app/models/access.rb
```diff
@@ -0,0 +1,102 @@
...
+    super_admin: "super_admin",
+    admin: "admin",
+    analyst: "analyst"
+  }
+
+  validates :role, presence: true
+  validates :resource_type, inclusion: {in: ALLOWED_RESOURCE_TYPES}, allow_nil: true
+
+  class << self
+    def can_manage_facilities?(record)
```
kitallis, line 17: These can be collapsed to a single method eventually.

app/controllers/admin/facilities_controller.rb
```diff
@@ -85,13 +94,20 @@ def new_facility(attributes = nil)
...
     @facility = Facility.friendly.find(params[:id])
-    authorize([:manage, :facility, @facility])
   end
 
   def set_facility_group
     @facility_group = FacilityGroup.friendly.find(params[:facility_group_id])
   end
 
+  def authorize_facility
+    raise Pundit::NotAuthorizedError unless current_admin.can_manage_facilities?(@facility)
```
kitallis, line 104: Just piggy-backing on the pundit exception so we can throw up an already wired-up redirect + flash message. We can use our own exceptions later.

app/controllers/admin/facilities_controller.rb
```diff
@@ -8,41 +8,42 @@ class Admin::FacilitiesController < AdminController
...
+    facilities =
+      if searching?
+        current_admin
+          .accessible_facilities(:manage)
+          .includes(facility_group: :organization)
+          .search_by_name(search_query)
+      else
+        current_admin
+          .accessible_facilities(:manage)
+          .includes(facility_group: :organization)
```
rsanheim, line 26: ✨ This reads much more direct and clear.

app/controllers/admin/facilities_controller.rb
rsanheim: 👍

app/controllers/admin/facility_groups_controller.rb
```diff
@@ -60,7 +58,14 @@ def set_protocols
...
     @facility_group = FacilityGroup.friendly.find(params[:id])
-    authorize([:manage, @facility_group])
+  end
+
+  def authorize_facility_group
+    raise Pundit::NotAuthorizedError unless current_admin.can_manage_facility_groups?(@facility_group)
+  end
+
+  def authorize_organization
+    raise Pundit::NotAuthorizedError unless current_admin.can_manage_organizations?(@facility_group.organization)
```
rsanheim, line 68: Stylistic thing - using the suffix version of unless is not ideal here, because it kinda buries the critical piece of this logic.  Some alternatives:

```ruby
def authorize_facility_group
  unless current_admin.can_manage_organizations?(@facility_group.organization)
    raise NotAuthorized
  end
end
```

Shorter, keeps the predicate check front and center

```ruby
def authorize_facility_group
  current_admin.can_manage_organizations?(@facility_group.organization) || raise NotAuthorized
end
```

Or if we want the model to tell us about it, we could add an API like this. If we are doing the raise style everywhere, I think this would be nice to add.  Note that we don't need to tell the user what kind of object we have, because the relationship is polymorphic.  The Access model should be able verify access for any type of allowable region.

```ruby
def authorize_facility_group
  current_admin.verify_manage_access!(organization)
end

class User
  def verify_manage_access!(region)
    accesses.admin.find_by(resource: region) || raise NotAuthorized
  end
end
```

app/controllers/my_facilities_controller.rb
```diff
@@ -96,7 +93,11 @@ def set_time_zone
     Time.use_zone(time_zone) { yield }
   end
 
-  def authorize_my_facilities
-    authorize(:dashboard, :view_my_facilities?)
+  def authorize_view
+    raise Pundit::NotAuthorizedError unless current_admin.can_read_aggregates?
```
rsanheim, line 97: There is some scoping missing here, right? Asking if an admin can read something without a particular target doesn't make sense.  Maybe something like:

```
current_admin.accessible_facilities(:view).any?
```

app/models/user.rb
```diff
@@ -73,6 +74,17 @@ class User < ApplicationRecord
...
     :invited_to_sign_up?, to: :email_authentication, allow_nil: true
 
+  delegate :organizations,
+    :facilities,
+    :facility_groups, to: :accesses, prefix: :accessible
+
+  delegate :can_manage_facilities?,
+    :can_manage_facility_groups?,
+    :can_manage_organizations?,
+    :can_view_identifiable_info?,
```
rsanheim, line 84: Is this for PII?  If so lets be explicit about it, its a critical piece of the auth levels:

```
admin.can_view_PII?
```

Or

```
admin.can_view_patient_information?
```

app/models/role.rb
```diff
@@ -0,0 +1,7 @@
+class Role < ApplicationRecord
```
rsanheim, line 1: This model can go away now I think?

app/views/admin/facilities/index.html.erb
```diff
@@ -2,10 +2,8 @@
   <h1 class="page-title">Facilities</h1>
 
   <div class="page-nav">
-    <% if policy([:manage, :facility, Facility]).upload? %>
+    <% if current_admin.can_manage_organizations?(Organization) %>
```
rsanheim, line 5: Would this check only ever return true for super admins?

app/models/access.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    def can_manage_organizations?(record)
+      return true if super_admin.exists?
+
+      manageable_organizations = organizations(:manage)
+
+      if record.instance_of?(Organization)
+        manageable_organizations.where(id: record).exists?
+      else
+        manageable_organizations.exists?
```
rsanheim, line 49: What else would we pass in except an instance of an Org?  

The only case not covered by these `can_` methods here would be a brand new Simple.org setup, but I think that would be a super admin type thing to create new orgs to get things up and running.  

```
if admin.can_add_new_orgs?
  <%= form_for Organization.new %>
```

and then I think:

```
def can_add_new_orgs?
  super_admin?
end
```

app/models/access.rb
```diff
@@ -0,0 +1,102 @@
...
+  class << self
+    def can_manage_facilities?(record)
+      return true if super_admin.exists?
+
+      manageable_facilities = facilities(:manage)
+
+      if record.instance_of?(Facility)
+        manageable_facilities.where(id: record).exists?
+      else
+        manageable_facilities.exists?
```
rsanheim, line 25: Same as below, what else is a valid argument other than an instance of a Facility?  We don't need class level checks anymore, we have concrete objects we can work with.  Please correct me if I'm wrong, but I can't think of where you would do an auth check like this _without_ facilities to pass into the method.

If you want to ask if an admin can add a new Facility, you ask the parent object: `admin.can_manage?(facility_group)` -> if that returns true, the user can add a facility underneath the parent facility group.

app/models/access.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    def can_manage_facility_groups?(record)
+      return true if super_admin.exists?
+
+      manageable_facility_groups = facility_groups(:manage)
+
+      if record.instance_of?(FacilityGroup)
+        manageable_facility_groups.where(id: record).exists?
+      else
+        manageable_facility_groups.exists?
```
rsanheim, line 37: Same here, no need for the else clause.  You would only ask about facility group management with instances of a facility group.

app/models/user.rb
kitallis: `admin.can_view_patient_information?` is a lot better 👍

app/views/admin/facilities/index.html.erb
kitallis: You could have org-level permissions only on IHCI, but not on an org for a different domain (like PATH).

app/models/access.rb
kitallis: The only case is that you may want to check if you can manage _any_ facility. Similarly for _any_ Org or _any_ FacilityGroup – an [example](https://github.com/simpledotorg/simple-server/blob/master/app/controllers/admin/facilities_controller.rb#L12). 

Don't know of a good way to handle that case, other than passing in the whole object-level scope.

We could consider doing, no arguments for that case maybe, `admin.can_manage_facilities?`. wdyt?

app/controllers/my_facilities_controller.rb
kitallis: We could do that, but `can_read_aggregates?` is a really one-sided relationship at the moment. It really just means [this](https://github.com/simpledotorg/simple-server/pull/1160/files#diff-dbee6fa62dba5eb76fea1e205e2a5f54R57) – basically if you're a super_admin? or you have _any_ access at all. Even if you have the lowest most restrictive Access in the system, as long as you have something, that means you can read aggregates (as it stands).

This definition could change later and we can upgrade `can_read_aggregates?` then?

app/models/access.rb
harimohanraj89: I think Rob's question here is - when do we need to check if you can manage _any_ facility? Do we have a use case for that today?

kitallis:
@simpledotorg/server-developers Closing this PR in favor of https://github.com/simpledotorg/simple-server/pull/1184, I'll however keep referencing the conversation we had here for further changes.

app/models/access.rb
kitallis: Yeah, I posted an example in the comment above, a lot of `index` pages will be like that I think.

PR #1161, opened 2020-07-17
Allow messaging "non-contactable" patients if needed

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/297/bangladesh-sms-blast

## Because

BD Phone numbers are not all marked as "mobile" numbers.

## This addresses

Adding a `contactable` flag to the `MessagePatients` service that does two things:

* Decides whether to scope the passed-in `patients` with the `contactable` scope, and
* Decides whether to use the patient's latest _mobile_ number, or phone number.

### DIFF
--- a/lib/tasks/scripts/message_patients.rb
+++ b/lib/tasks/scripts/message_patients.rb
@@ -3,14 +3,14 @@ class MessagePatients
     new(*args).call
   end
 
-  attr_reader :patients, :message, :channel, :contactable, :dryrun, :verbose, :report
+  attr_reader :patients, :message, :channel, :only_contactable, :dryrun, :verbose, :report
   VALID_CHANNELS = [:whatsapp, :sms]
 
-  def initialize(patients, message, channel: :whatsapp, contactable: true, dryrun: false, verbose: true)
+  def initialize(patients, message, channel: :whatsapp, only_contactable: true, dryrun: false, verbose: true)
     @patients = patients
     @message = message
     @channel = channel
-    @contactable = contactable
+    @only_contactable = only_contactable
     @dryrun = dryrun
     @verbose = verbose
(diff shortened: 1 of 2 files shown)


lib/tasks/scripts/message_patients.rb
```diff
@@ -59,7 +60,19 @@ def send_messages
...
+    else
+      patients
+    end
+  end
+
+  def phone_number_for(patient)
+    if contactable
+      patient.latest_mobile_number
+    else
+      patient.latest_phone_number
```
prabhanshuguptagit, line 74: What happens if the patient doesn't have a phone number? Is that a possible case?

lib/tasks/scripts/message_patients.rb
```diff
@@ -3,13 +3,14 @@ def self.call(*args)
     new(*args).call
   end
 
-  attr_reader :patients, :message, :channel, :dryrun, :verbose, :report
+  attr_reader :patients, :message, :channel, :contactable, :dryrun, :verbose, :report
   VALID_CHANNELS = [:whatsapp, :sms]
 
-  def initialize(patients, message, channel: :whatsapp, dryrun: false, verbose: true)
+  def initialize(patients, message, channel: :whatsapp, contactable: true, dryrun: false, verbose: true)
```
prabhanshuguptagit, line 9: `only_contactable: true`?

lib/tasks/scripts/message_patients.rb
harimohanraj89: Great call. In that case we should most likely no-op the patient. I'll put that in 👍

harimohanraj89:
Thanks for the review. Comments addressed in 468f8aa

PR #1163, opened 2020-07-17
Regions

rsanheim:
### Why do we need Regions?

Our current reporting / organizational structures is as follows, from highest to lowest, with Patients belonging to Facilities.

Organization -> FacilityGroup -> Facility

We know we are missing some levels in that hierarchy, namely States and Blocks:

Organization -> State -> FacilityGroup -> Block -> Facility

All of these levels can share some things: they have a name, a URL friendly slug, and a description. Most importantly, they allow us to group patients and their data into meaningful slices for reports. 

We also know that these "levels" may be different from country to country, not only in their name but also in terms of their actual structure. Bangladesh or Ethiopia may not have Blocks, or may have a different name for that sort of group.

Its clear we shouldn't keep adding levels in the structure as one-offs, because they all share common behavior and the one-off approach won't scale to other countries. Given that, we should start working towards a better domain that lets us model Regions in a consistent, unified way. 

### What are Regions?

A Region is a hierarchical grouping object with a recursive relationship. A region has many child regions, and a region belongs to one parent region, except for the root region. In other systems they are often called something like organization, department, or division. Since our groupings are _mostly_ geographic, and the term "organization" is already taken, Region seems like a good term for Simple.org

### How to get there?

The short term plan is to add Regions, and insert them into the hierarchy at the root (for an entire app instance) and for States to begin the migration process. We would modify existing models to work as if they are regions, so that we can start using a unified interface via duck typing. The long term goal is for a Region to take the place of Orgs, FacilityGroup, and Facility for grouping.  

This PR tackles that short term plan. Here is a sketch of the relationships for how things would look after this PR is merged:

regions-v1

Long term, Regions would handle the entirety of our organizational structure, and would also make assigning roles to users easier. Regions would be used for root, organization, state, district, block, facility group, and facility, and whatever other level we need. We could extract domain objects from the current Facility model, with the first candidate being the Address and Geo location info. We would probably keep around the facility and facility group models, as they would hold specific data and relationships that are only relevant to that specific level.

Here is a sketch of how things could look long term, after what would be many PRs. Obviously this is all very hand wavey and further down the line; we would want to iterate towards something like this and refine the design as we went.

regions-future-state

**Feedback welcome!** If this seems like a good plan, this first PR should be pretty easy to ship as a first step. The main thing we'd want to be careful about is the data migration in production.

### References

[Organization Structures](https://martinfowler.com/apsupp/accountability.pdf) from Fowlers' _Analysis Patterns_ distills many of the common patterns we are dealing with here in terms of Org hierarchies and knowledge levels.

* [user roles (take 3)](#1173) - this becomes easier as regions come into play
* https://app.clubhouse.io/simpledotorg/story/132/block-level-reports
* https://app.clubhouse.io/simpledotorg/story/454/state-level-reports


app/models/region.rb
```diff
@@ -0,0 +1,36 @@
...
+    root: 0,
+    organization: 10,
+    state: 20,
+    facility_group: 30,
+    facility: 40
+  }
+
+  belongs_to :parent_region, polymorphic: true, optional: true
+  has_many :_child_regions, as: :parent_region, class_name: "Region"
+  has_many :_facility_groups, inverse_of: :parent_region, foreign_key: :parent_region_id
```
harimohanraj89, line 14: Are these "private" associations because they're only here for compatibility until everything becomes a region?

app/models/region.rb
```diff
@@ -0,0 +1,36 @@
+class Region < ApplicationRecord
+  extend FriendlyId
+  friendly_id :name, use: :slugged
+  enum level: {
+    root: 0,
+    organization: 10,
+    state: 20,
+    facility_group: 30,
+    facility: 40
+  }
```
harimohanraj89, line 10: Is the idea that this enum structure allows us enough room to insert other region levels into the Simple codebase to support, say, a new country's requirements? (eg. subdistricts in bangladesh)

app/models/region.rb
```diff
@@ -0,0 +1,36 @@
...
+  has_many :_child_regions, as: :parent_region, class_name: "Region"
+  has_many :_facility_groups, inverse_of: :parent_region, foreign_key: :parent_region_id
+
+  validates :level, presence: true, uniqueness: {conditions: -> { where(level: 0) }, message: "can only have one root Region"}
+  validates :name, presence: true
+  validates :slug, presence: true, uniqueness: true
+  validates :parent_region, presence: true, unless: :root?
+
+  def self.root
+    @root ||= Region.readonly.find_or_create_by!(name: "root", level: :root)
```
harimohanraj89, line 22: It makes me nervous to create this 😅 Maybe we find_or_create in our application setup somehow?

app/models/region_type.rb
```diff
@@ -0,0 +1,3 @@
+class RegionType < ApplicationRecord
+  validates :name, presence: true
+end
```
harimohanraj89, line 3: Is this still needed, or does `Region#level` take care of this now?

rsanheim: Oh right, this can go away.  YAGNI

app/models/region.rb
rsanheim: Yup, exactly.

rsanheim: And I realized that these should be strings in the database, not integers, so that each country's database can use whatever names for these they want.

app/models/region.rb
rsanheim: Yup you got it.

app/models/region.rb
timcheadle: Making these integers feels... wrong? The names are arbitrary; it's the hierarchy links that matter.

Even in India, different areas have different structures.

app/models/region.rb
```diff
@@ -0,0 +1,36 @@
...
+  enum level: {
+    root: 0,
+    organization: 10,
+    state: 20,
+    facility_group: 30,
+    facility: 40
+  }
+
+  belongs_to :parent_region, polymorphic: true, optional: true
+  has_many :_child_regions, as: :parent_region, class_name: "Region"
```
timcheadle, line 13: Why is this association private?

harimohanraj89:
@rsanheim checking in, did you want to see this PR through to completion and merge, or was this a spike to gather thoughts? Let us know what you wanna do with it.

rsanheim:
I want to carry this forward, just been too swamped with reports to pick it back up. It will make the next round of reports works a lot easier, so hope to do it in next few weeks.

PR #1168, opened 2020-07-20
Get top registrations benchmark

rsanheim:
**Story card:** https://app.clubhouse.io/simpledotorg/story/433/top-region-benchmarks

*NOTE:* this also includes some fixes for an unrelated bug I found while looking at report data for older date periods.

image

## Because

We want more benchmark data

## This addresses

Wiring up top registration data


app/services/control_rate_service.rb
```diff
@@ -64,23 +68,31 @@ def registration_counts
   end
 
   def controlled_patients(time)
+    LatestBloodPressuresPerPatientPerMonth.with_discarded.from(bp_query(time).under_control,
```
timcheadle, line 71: Why are we including discarded BPs in these queries?

PR #1169, opened 2020-07-21
Assigned facility: Query updates

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/358

This is a long running branch for collecting the various assigned facility report changes. All per page PRs will be raised against this branch so the changes can be released at once.

[Status doc](https://docs.google.com/document/d/1YZqm0t5NhTDUkn3Ww8yRqAszKCtAiOqChpFED6U-QXE)


app/views/analytics/districts/_analytics_table.html.erb
```diff
@@ -22,8 +22,8 @@
       <thead>
       <tr>
         <th></th>
-        <th class="row-label"><strong>Total</strong></th>
-        <th class="row-label" colspan="6"><strong>Patients with BP taken</strong></th>
+        <th class="row-label"><strong>Cumulative patients</strong></th>
+        <th class="row-label" colspan="6"><strong>Follow-ups</strong></th>
```
harimohanraj89, line 26: Nice. Since this is the districts table, I think the simpler language is better 👍

app/views/reports/regions/details.html.erb
```diff
@@ -96,39 +96,21 @@
...
-  <% end %>
-<% else %>
+<% unless @region.is_a?(Facility) %>
   <% if current_admin.permissions_v2_enabled? %>
     <%= render "analytics/districts/analytics_table",
                dashboard_analytics: @dashboard_analytics,
                period: @period.type,
                row_resource: current_admin.accessible_facilities(:view_reports).order(:name),
-               row_resource_description: "Facilities",
+               row_resource_description: "Follow-ups and registrations",
```
harimohanraj89, line 105: ❤️

app/views/reports/regions/show.html.erb
```diff
@@ -232,15 +232,15 @@
                dashboard_analytics: @dashboard_analytics,
                period: :month,
                row_resource: current_admin.accessible_users(:view_reports).order(:full_name),
-               row_resource_description: "Users",
+               row_resource_description: "New registrations",
```
harimohanraj89, line 235: ❤️

app/exporters/patients_exporter.rb
```diff
@@ -51,6 +52,10 @@ def self.csv_headers
       "Patient District",
       (zone_column if Rails.application.config.country[:patient_line_list_show_zone]),
       "Patient State",
+      "Assigned Facility Name",
```
kitallis, line 55: @prabhanshuguptagit @harimohanraj89 Do we want to double-check if this should be called Preferred Facility? (not a blocker)

app/views/reports/regions/details.html.erb
```diff
@@ -90,22 +96,24 @@
...
-  <%= render 'shared/analytics_table',
-             dashboard_analytics: @dashboard_analytics,
-             period: @period.type,
-             row_resource: Facility.order(:name),
-             row_resource_description: "Facilities",
-             row_resource_display_field: :name,
-             row_resource_link: :reports_region_facility_path %>
+<% unless @region.is_a?(Facility) %>
+  <% if current_admin.permissions_v2_enabled? %>
+    <%= render "analytics/districts/analytics_table",
```
kitallis, line 101: @prabhanshuguptagit @harimohanraj89 I feel like I remember getting rid of this generic table in favor of two separate ones. Did this come back somehow?

kitallis (commented):
It's looking solid to me, I've left a few comments, let's discuss and then we can close this off.

app/views/reports/regions/details.html.erb
kitallis: And if so, do we now have 3 of these? Generic, District-level, Facility-level?

app/views/analytics/districts/_analytics_table.html.erb
prabhanshuguptagit: The "Follow ups" column was renamed to "Patients with BP taken", since theese are not follow ups at that facility anymore (these could be recorded at any facility). Just wanted to double check that this change is intended.

(For later) Will `Cumulative patients` be the ubiquitous term for "Total" patients? This number is being shown on Missed Visits and BP control as `Total patients`.

harimohanraj89: This change is only for the district table, where both will be the same.

app/exporters/patients_exporter.rb
harimohanraj89: Good question. We shall check.

app/views/reports/regions/details.html.erb
harimohanraj89: That was nice, but the reports re-use this markup now, so I re-partialized it.

There are indeed 3 versions now. I will trash the generic one (`shared/analytics_table`), it's not used.

app/views/analytics/districts/_analytics_table.html.erb
harimohanraj89: Never mind, the facility-wise breakdown below will differ.

harimohanraj89 (approved):
All remaining feedback addressed. Time for this ship to sail.

PR #1173, opened 2020-07-22
Spike: Simple User Roles (permissions take 3)

rsanheim:
*NOTE: You should review this PR [with white space hidden.](https://github.com/simpledotorg/simple-server/pull/1173/files?diff=split&w=1)*

I've been thinking about our user auth overhaul attempts, and wanted to try an even simpler approach to cut down on the amount of boilerplate and complexity. This is a spike exploring that -- this approach adds one model and would allow us to drop Pundit.

It adds a single model - `Role` - which belongs to a User and a Resource. The
Role defines whether a User is a super admin, admin, or analyst of a
Resource. Right now a resource can be an Org, FacilityGroup, or Facility. So a User has many Resources through Roles.

I _think_ this is the simplest possible data model that could work. This does not store the role permissions in the database at all, because I think they are pretty static, but we could add a `RolePermissions` and `Permissions` table if we really needed that flexibility. It wouldn't change the underlying implementation here very much.

I also added authorization checks via a Flipper feature flag
for a couple controllers to show how nice using this is.
With this approach we are using the database to handle as much of our
authorization as possible or us. This is done via a couple UNION queries to capture the
hiearchial nature of the roles, so we can get all facility groups
or all facilities a User may have access to as an ActiveRecord relation.

That means we can then chain off of that relation so that our auth code
can stay dumb and simple. Controllers should scope off `current_admin` to ensure we
are only finding resources a user should have access to.
Using this approach means resources that a user doesn't have access to
should never even be loaded from the database for a properly written
controller.

If a logged in admin tries to load some facility they shouldn't have access to, our controllers
will raise RecordNotFound errors, because we can now do things like
this:

```
current_admin.facilities.find_by!(slug: params[:slug])
```

Or read only access:

```
current_admin.facilities(role: :read).find_by!(slug: params[:slug])
```

Or finding a facility group:

```
facility_group = current_admin.facility_groups.find_by!(slug: params[:slug])
```

Or finding all facilities a user has access to:

```
current_admin.facilities
```

I realize this may not handle some more complicated edge cases, but wanted
to get it out there for feedback and thoughts before considering all the
trickier scenarios. Feedback welcome!

Clubhouse card: https://app.clubhouse.io/simpledotorg/story/154/explore-users-roles-permissions
Related PRs: #1137 and #1160


app/controllers/admin/facilities_controller.rb
```diff
@@ -9,23 +9,35 @@ class Admin::FacilitiesController < AdminController
     :validate_facility_rows, if: :file_exists?, only: [:upload]
 
   def index
-    authorize([:manage, :facility, Facility])
+    unless Flipper.enabled?(:user_roles)
```
rsanheim, line 12: We have to turn off all the Pundit checks if we want to user this new user role auth.

app/controllers/admin/facilities_controller.rb
```diff
@@ -9,23 +9,35 @@ class Admin::FacilitiesController < AdminController
...
+      authorize([:manage, :facility, Facility])
+    end
 
     if searching?
-      facilities = policy_scope([:manage, :facility, Facility]).search_by_name(search_query)
+      facilities = if Flipper.enabled?(:user_roles)
+        current_admin.facilities.search_by_name(search_query)
```
rsanheim, line 18: No more policies to go searching for! ✨✨✨

app/controllers/admin/facilities_controller.rb
```diff
@@ -9,23 +9,35 @@ class Admin::FacilitiesController < AdminController
...
     else
-      @organizations = policy_scope([:manage, :facility, Organization])
-      @facility_groups = @organizations.map { |organization|
-        [organization, policy_scope([:manage, :facility, organization.facility_groups])]
-      }.to_h
-      @facilities = @facility_groups.values.flatten.map { |facility_group|
-        [facility_group, policy_scope([:manage, :facility, facility_group.facilities])]
-      }.to_h
+      if Flipper.enabled?(:user_roles)
+        # @organizations = policy_scope([:manage, :facility, Organization])
```
rsanheim, line 29: I don't think we would need to directly look up orgs here anymore, as I think it was just for the grouping part? Not sure.

app/controllers/admin_controller.rb
```diff
@@ -1,8 +1,8 @@
 class AdminController < ApplicationController
   before_action :authenticate_email_authentication!
 
-  after_action :verify_authorized, except: [:root]
-  after_action :verify_policy_scoped, only: :index
+  after_action :verify_authorized, except: [:root], unless: -> { Flipper.enabled?(:user_roles) }
+  after_action :verify_policy_scoped, only: :index, unless: -> { Flipper.enabled?(:user_roles) }
```
rsanheim, line 5: More disabling of Pundit.

app/controllers/my_facilities_controller.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 class MyFacilitiesController < AdminController
-  include DistrictFiltering
```
rsanheim, line 4: Wanted to inline this module to see what it was actually doing.

app/models/role.rb
```diff
@@ -0,0 +1,12 @@
...
+  enum name: {
+    super_admin: "super_admin",
+    admin: "admin",
+    analyst: "analyst"
+  }
+  belongs_to :user
+  belongs_to :resource, polymorphic: true
+  validates :name, presence: true
+
+  scope :super_admin_or_admin, -> { where(name: [:super_admin, :admin]) }
```
rsanheim, line 11: Because "write" access is essentially the same for an admin and a super_admin, this scope could come in handy.

app/models/role_scoped_regions.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      SELECT facility_groups.* from facility_groups
+      INNER JOIN roles
+        ON roles.resource_type = 'FacilityGroup'
+        AND roles.resource_id = facility_groups.id
+      WHERE roles.user_id = :user_id
+        AND roles.name in (:roles)
+    SQL
+    ids = FacilityGroup.find_by_sql([sql, {roles: roles, user_id: id}]).pluck(:id)
+    FacilityGroup.where(id: ids)
```
rsanheim, line 28: Here we re-find things by id so that we return an ActiveRecord Relation instead of an Array. 

There are gems and hacks to do a Union in plain ARel, but I'm not sure its really worth it...these queries should be very fast once we get the indexes in place.

app/models/role_scoped_regions.rb
```diff
@@ -0,0 +1,73 @@
...
+      UNION
+
+      SELECT facilities.* FROM facilities
+      INNER JOIN roles
+        ON roles.resource_type = 'Facility'
+        AND roles.resource_id = facilities.id
+      WHERE roles.user_id = :user_id
+        AND roles.name in (:roles)
+
+    SQL
```
rsanheim, line 69: Unioning the three ways that a user can have access to a Facility - via the owning Org, the owning FacilityGroup, or directly on the Facility.

app/models/user.rb
```diff
@@ -14,6 +14,12 @@ class User < ApplicationRecord
 
   belongs_to :organization, optional: true
 
+  has_many :roles
+  include RoleScopedRegions
+  has_many :managed_organizations, -> { Role.super_admin_or_admin }, through: :roles, source: :resource, source_type: "Organization"
+  has_many :managed_facility_groups, -> { Role.super_admin_or_admin }, through: :roles, source: :resource, source_type: "FacilityGroup"
+  has_many :managed_facilities, through: :managed_facility_groups, source: :facilities
```
rsanheim, line 21: I initially tried to get things working w/ these has_manys and AR merges, but it wasn't working out.  Not sure we would even want to keep these assocations honestly, probably better to have the one interface in RoleScopedRegions.

spec/factories/users.rb
```diff
@@ -55,6 +55,21 @@
     [Faker::Name.first_name, Faker::Name.last_name, Faker::Internet.domain_word, n].join("-")
   end
 
+  factory :admin_v2, class: User do
```
rsanheim, line 58: Had to create a new admin factory because the existing one bakes in the notion of an org that the user has user_permissions of.

kitallis:
Closing this in favor of https://github.com/simpledotorg/simple-server/pull/1184.

PR #1176, opened 2020-07-23
Bugfix: CSV formatting

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/205

## Because

There appears to be problems escaping characters in the overdue list CSV which causes column data to bleed into other columns. Cohort downloads also has a quote escaping issue in the column headers:

image

## This addresses

This fixes the CSV generation in these downloads. Since these CSVs are rendered via ERB, templating a row sanitises special characters. For example `"` becomes a `&quot;` which some CSV clients handle correctly, some can't. Strings with a comma in between are read as separate cells because they are not wrapped in `"`.

This sanitisation is desirable in HTML views since they are executed in the browser. File downloads should contain the raw strings. Running the string through an `html_safe` spits the correct unsanitised strings into the CSV.

### DIFF
--- a/app/views/analytics/districts/show.csv.erb
+++ b/app/views/analytics/districts/show.csv.erb
@@ -1,10 +1,8 @@
-<%= [
-  'Report generated at:',
-  Time.current
-].to_csv -%>
+<% csv = CSV.generate(headers: true) do |csv| %>
+<% csv << ["Report generated at", Time.current] %>
 <% period = @period == :quarter ? 'Quarterly' : 'Monthly' %>
-<%= ["#{@organization_district.district_name} #{period} Cohort Report"].to_csv -%>
-<%= ['Facility', 'Facility type'].concat(@cohort_analytics.each_with_index.flat_map do |(report_dates, analytics), index|
+<% csv << ["#{@organization_district.district_name} #{period} Cohort Report"] %>
+<% csv << ['Facility', 'Facility type'].concat(@cohort_analytics.each_with_index.flat_map do |(report_dates, analytics), index|
   [
     'Cohort period',
     'Follow up period',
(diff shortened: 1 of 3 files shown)


app/views/analytics/districts/show.csv.erb
```diff
@@ -1,10 +1,8 @@
-<%= [
-  'Report generated at:',
-  Time.current
-].to_csv -%>
+<% csv = CSV.generate(headers: true) do |csv| %>
```
prabhanshuguptagit, line 1: This generates a CSV and calls `.html_safe` on the final output instead of doing `.to_csv.html_safe` on all the individual strings.

app/views/analytics/districts/show.csv.erb
```diff
@@ -75,26 +73,25 @@ end).to_csv -%>
...
         @dashboard_analytics.dig(facility.id, :total_registered_patients) || 0,
         *dates.map { |period| @dashboard_analytics.dig(facility.id, :follow_up_patients_by_period, period) || 0 },
         *dates.map { |period| @dashboard_analytics.dig(facility.id, :registered_patients_by_period, period) || 0 }
-      ].to_csv
-    -%>
+      ] %>
   <% end %>
 <% end %>
+<% end.html_safe %>
+<%= csv %>
```
prabhanshuguptagit, line 97: This is the crux of the change.

app/views/appointments/index.csv.erb
```diff
@@ -1,8 +1,9 @@
-<%= [
+<% csv = CSV.generate do |csv| %>
+<% csv << [
```
harimohanraj89, line 2: The ERB tags in this block don't output anything, so can we indent them for readability?

```
<% csv = CSV.generate do |csv| %>
  <% csv << [... %>
  <% end %>
<% end.html_safe %>
```

app/views/appointments/index.csv.erb
```diff
@@ -29,7 +30,8 @@
     ('High' if patient_summary.risk_level > 0),
     patient_summary.street_address,
     patient_summary.village_or_colony,
-    patient_summary.latest_phone_number,
-    "Patient Summary"
```
harimohanraj89, line 33: Ha, good catch 👍

prabhanshuguptagit:
> Were you able to recreate the problem with a local or SBX download?

😄  Yes, any addresses with a comma break the CSV. This is basically what the CSV's looked like to a client before:
image

Now:
image

harimohanraj89:
💥 :shipit: 🚀

PR #1177, opened 2020-07-24
Fix patient sync bug

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/549

## Because

We ran into [this bug](https://sentry.io/organizations/resolve-to-save-lives/issues/1803874040/?project=1217715&query=is%3Aunresolved) with the recent patient merge refactor. This happens due to a race condition when two syncs simultaneously try to create the same patient.

## This addresses

In `MergePatientService`, caching the `existing_patient` results in a case where one sync creates a patient and the other sync fails:

- Both requests fetch and memoize existing patient data (which is nil)
- Request 1 (faster) computes merge status, determines it's a `new` resource and creates it
- Request 2 (slower) computes merge status and determines it's an `update`. It tries to use the memoized `existing_patient` for fetching the metadata but fails.

This removes the memoization on `existing_patient` to avoid this race condition.

### DIFF
--- a/app/services/merge_patient_service.rb
+++ b/app/services/merge_patient_service.rb
@@ -2,10 +2,10 @@ class MergePatientService
   def initialize(payload, request_metadata:)
     @payload = payload
     @request_metadata = request_metadata
-    @existing_patient = Patient.with_discarded.find_by(id: payload["id"])
   end
 
   def merge
+    patient_before_merge = existing_patient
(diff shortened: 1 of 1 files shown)


timcheadle:
Next time, can you describe the bug in the PR description? That way we can see what happened regardless of whether we keep Sentry around or not.

timcheadle:
Wish we could add a unit test for this but I realize that's difficult.

PR #1183, opened 2020-07-27
Refactoring: preparing the way for easier quarterly cohorts and monthly cohorts

rsanheim:
This PR got bigger and more extensive than I wanted, but I'm glad with how it turned out. I was frustrated with having to deal with monthly vs quarterly periods, so I created an object `Period` to handle that. That ended up simplifying a lot of the service code necessary to grab all the data, and will also make it to work with from the view.

I have more plans for this refactoring, but want to get this piece merged so I can build off of it with subsequent pieces of the reports.

**Story card:** https://app.clubhouse.io/simpledotorg/story/465/monthly-cohort-reports

## Because

* We want to display cohorts for months and quarters
* We want to deal with monthly reports and quarterly reports w/o a lot of pain and type checking

## This addresses

* Refactoring and cleaning up so we work with Periods, and not just Dates and Quarter.
* Returning the 'grouped by period' registrations as well as the cumulative counts, because returning just one or the other is too limiting (and if we grab the grouped numbers, its easy to return the cumulative as well).


app/models/period.rb
```diff
@@ -0,0 +1,103 @@
...
+      value
+    else
+      raise ArgumentError, "unknown quarter value #{value} #{value.class}"
+    end
+    new(type: :quarter, value: quarter)
+  end
+
+  def initialize(attributes = {})
+    super
+    @type = type.intern if type
```
timcheadle, line 38: Can we make this `to_sym` instead? Always looks more readable to me

app/models/quarter.rb
```diff
@@ -1,24 +1,50 @@
...
     @number = QuarterHelper.quarter(date).freeze
+    @to_s = "Q#{number}-#{year}".freeze
   end
 
   def next_quarter
-    self.class.new(date: date.advance(months: 3))
+    advance(months: 3)
   end
 
+  alias succ next_quarter
```
timcheadle, line 40: Instead of aliasing this, should we just find/replace with `succ` and `previous`?

app/models/quarter.rb
```diff
@@ -1,24 +1,50 @@
...
   def self.current
     new(date: Date.current)
   end
 
+  def self.parse(string)
+    match = string.match(PARSE_REGEX)
+    raise ArgumentError, "String to parse as Quarter must match QX-YYYY format" unless match
+    number = Integer(match[1])
+    year = Integer(match[2])
+    quarter_month = ((number - 1) * 3) + 1
```
timcheadle, line 18: Can you make this its own utility method, like `month_to_quarter`?

app/services/control_rate_service.rb
```diff
@@ -1,25 +1,34 @@
...
+      @single_period = periods
+      Range.new(periods, periods)
+    else
+      periods
+    end
+    @quarterly_report = @periods.begin.quarter?
+    logger.info "#{self.class} created for periods: #{periods} facilities: #{facilities.map(&:id)} #{facilities.map(&:name)}"
+  end
+
+  def single_period?
```
timcheadle, line 24: Shouldn't this return a boolean?

app/services/control_rate_service.rb
```diff
@@ -29,55 +38,80 @@ def call
...
-      region.registered_patients.with_hypertension
-        .group_by_period(:month, :recorded_at, range: range)
-        .count
-        .each_with_object(Hash.new(0)) { |(date, count), hsh|
-          hsh[:running_total] += count
-          hsh[date] = hsh[:running_total]
-        }.delete_if { |date, count| count == 0 }.except(:running_total)
-    else
-      count = region.registered_patients.with_hypertension.where("recorded_at <= ?", date).count
+    @registration_counts ||= if single_period
```
timcheadle, line 70: Use `single_period?` here for consistency

app/services/control_rate_service.rb
```diff
@@ -97,15 +131,19 @@ def bp_query(time)
 
   private
 
+  def quarterly_report?
+    @quarterly_report
+  end
+
   def cache_key
-    "#{self.class}/#{region.model_name}/#{region.id}/#{date_or_range_cache_key}"
+    "#{self.class}/#{region.model_name}/#{region.id}/#{periods.begin.type}_periods/#{periods_cache_key}"
```
timcheadle, line 139: I'd move the `type_periods` prefix into the `periods_cache_key` method, keeps it contained

app/services/region_report_service.rb
```diff
@@ -20,10 +20,10 @@ def initialize(region:, selected_date:, current_user:)
 
   attr_reader :current_user
   attr_reader :data
-  attr_reader :region
   attr_reader :facilities
+  attr_reader :period
```
timcheadle, line 24: Move one line down for alpha order ;)

app/models/quarter.rb
rsanheim: We need the `succ` method so that Quarters work in ranges...and we provide `next_quarter` to have a better named version of it. =)

app/services/control_rate_service.rb
rsanheim: It returns something that is truthy, which is good for what its used for.

PR #1185, opened 2020-07-28
Access scoping + auth

kitallis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/570/new-permissions-data-model

## Because

## This addresses

Add the scoping and authorization methods with specs.


spec/models/access_spec.rb
```diff
@@ -0,0 +1,636 @@
+require "rails_helper"
+
+RSpec.describe Access, type: :model do
+  describe "Associations" do
+    it { is_expected.to belong_to(:user) }
+    pending { is_expected.to belong_to(:resource).optional }
```
kitallis, line 6: Need to figure out this test, it's a little tricky because our resources are currently _optionally optional_ 🧠

app/models/access.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    def super_admin?
+      super_admin.exists?
+    end
+  end
+
+  private
+
+  def user_has_only_one_role
```
kitallis, line 79: Since we have decided to only have 1 _kind of_ role per user across many accesses, we can't write a simple database constraint anymore. This would have to be an application constraint and/or some unwieldy check-constraint on pg that we should avoid for now.

It might seem a bit nonsensical to offer this abstraction in the first place, why have the schema encode roles in the Access if there's only ever going to be 1?

The only good reason to do this is future flexibility by trading-off correctness... pretty much no other reason why. It makes it unnecessarily difficult to comprehend why this validation exists in this class at all.

OTOH, the flexibility that it provides is that we can change our minds about this without needing a migration and/or write more nuanced/different cases.

spec/models/access_spec.rb
```diff
@@ -0,0 +1,636 @@
+require "rails_helper"
+
+RSpec.describe Access, type: :model do
```
kitallis, line 3: There's a ton of tests in here (6 times more LOC than the code itself) and they may seem a bit redundant, but they aren't, they are mostly every small case documented explicitly.

app/models/access.rb
```diff
@@ -0,0 +1,86 @@
...
+        .or(FacilityGroup.where(organization: organizations(action)))
+    end
+
+    def facilities(action)
+      resources_for(Facility, action)
+        .or(Facility.where(facility_group: facility_groups(action)))
+    end
+
+    def can?(action, model, record = nil)
+      if record&.is_a? ActiveRecord::Relation
```
kitallis, line 40: This explicit constraint is currently dubious, I'm not sure if we'll keep it or not.

app/models/access.rb
```diff
@@ -0,0 +1,86 @@
...
+      return resources.exists? if super_admin?
+      return resources.find_by_id(record).present? if record
+
+      resources.exists?
+    end
+
+    def resources_for(resource_model, action)
+      return resource_model.all if super_admin?
+
+      resource_ids = where(resource_type: resource_model.to_s, role: ACTION_TO_ROLE[action]).pluck(:resource_id)
```
harimohanraj89, line 68: Both here and in the other methods here - we need to scope by user, right? Are we doing that? Did I miss it?

app/models/access.rb
harimohanraj89: I also think we shouldn't put `role` on `Access`. A data model that _allows_ for more flexibility than our app can support means either complicated code to account for inconsistent data, or constraints we need to manage, and this isn't easily solvable with a simple unique index. Are we comfortable enough to rely on the model vaildation? Isn't this "too-flexible data model" what introduced complexity in our current permissioning system too?

For example, if I want to decide whether to show a user a "Manage" link in our nav, would I iterate through all accesses and check for manage ability? Or just the first one, assuming they're all the same?

To address the concerns we had with the "keep role on user" approach, here are my instincts, let me know if they jive with you.

**The abstraction is spread across two classes**

I think that's OK because, the rest of our app will _only_ talk to `User` anyway! With the right methods and delegation on `User`, the rest of our app shouldn't need to know about accesses at all. I think that sets us up for evolving our permissions nicely as well. The trick will be making sure that the `Access` deals only with "which resources", and the `User` deals with "which actions". 

If modeling that is a problem...
(get ready for a cheeky suggestion)
```
class Access
  delegate :role, to: :user
end
```
Let the `Access` take care of all of it, and just pull the role it needs from the user. This way our data model is correct, and our abstraction is centralized.
It means some of the methods in here will have to resort to in-memory searches instead of ActiveRecord queries, but that should be ok, since we'll be dealing with a single user's accesses at most. For example:

```ruby
resource_ids = where(resource_type: resource_model.to_s).select { |access|
  access.role.in?(ACTION_TO_ROLE[action])
}.map(&:resource_id)
```

**Naming**

I'm not a big fan of `AccessibleResource` either 😅 it's super verbose. Let's also forget about that confusing role/designation field for a second. What if users had a "role" field and many "scopes"?

app/models/access.rb
vkrmis: Usage for these methods looks like:
`@user.can?(:view)`
`@user.accessible_facilities(:view)`

So they're always scoped by the `user --has_many--> accesses` relation, don't need to pass in the user and scope explicitly.

spec/models/access_spec.rb
harimohanraj89: You can set up separate subjects for separate tests.

```ruby
context "for super admins" do
  subject { Access.new(role: :super_admin) } 
  it { is_expected.not_to belong_to(:resource) } # I think you can do negative matchers?
end

context "for everyone else" do
  subject { Access.new(role: :viewer) } 
  it { is_expected.to belong_to(:resource) } # I think you can do negative matchers?
end
```

Would that work?

spec/models/access_spec.rb
harimohanraj89: Rigorous testing of this important piece of code is pretty expected, I think :)

spec/models/access_spec.rb
kitallis: Ah nice, let me try this, I had no idea it was picking up `subject` like that.

harimohanraj89: Hehe yeah it's pretty neat. If unspecified, it spins up a `model.new` as the subject by default.

kitallis: I got this working, but in a slightly different, more explicit way. The problem with that spec helper is that it doesn't understand our nuanced constraint and throws up this error:

```
Expected Access to have a belongs_to association called resource (and for the record not to fail validation if :resource is unset; i.e., either the association should have been defined with `optional: true`, or there should not be a presence validation on :resource)
```

Basically, it doesn't allow you to use `belongs_to` having both an `optional` and a `presence` validation together. Which we do, but conditionally so. Hence, it's not technically incorrect, just that the spec helper isn't smart enough to determine that conditional.

So I re-wrote it in a more straightforward way [here](https://github.com/simpledotorg/simple-server/pull/1185/commits/e9aa57aa00d4fe38c776d98256b3e694cea33687).

app/models/access.rb
kitallis: @harimohanraj89 I generally agree with this, but I think it's futile to try to achieve what we're trying to achieve with 100% success. `User` and `Access` are tightly coupled and you're going to leak information in either direction regardless of how hard you try. Unless you try and create yet [another abstraction](http://timelessrepo.com/abstraction-creep) that wraps over both.

The real cost-benefit analysis of _where access levels / roles lie_ are the following (and I still think that's the core idea we should focus on), everything else is exercise of subjective balance:

* You get clean and clear database constraints of only 1-role-per-`User` if the role is in `User`. I'd also argue that it's better, because when designing user accesses, you want strong, reliable database constraints.
* Future flexibility of changing the idea of 1 role per `User` if it's in `Access` without a migration. We've already gone back-and-forth on the 1-role-per-`User` constraint twice and introduced two more roles in the process of simplification, it isn't terribly hard to imagine this being an actual use-case and also doesn't necessarily differ much in terms of code-complexity.

To that extent, here's an attempt at moving the `access_level` (`role` was taken) to `User` but keeping some of fetching responsibilities in `Access` (as per your suggestions) – [Role in Access](https://github.com/simpledotorg/simple-server/pull/1185/files/e9aa57aa00d4fe38c776d98256b3e694cea33687) versus [Role in User](https://github.com/simpledotorg/simple-server/pull/1185/commits/fdaccce8955b444f1fe0998584a7d81204443f7b). You can compare these side-by-side and let me know what you think. Let's decide and freeze on one approach soon.

harimohanraj89: Thanks for laying out these comparisons, it's really helpful.

My gut says "keep role in User" because
* The clean DB constraint
* Migrating isn't hard

spec/models/user_spec.rb
```diff
@@ -3,11 +3,92 @@
...
   describe "Associations" do
     it { should have_many(:user_authentications) }
+    it { should have_many(:accesses) }
   end
 
   describe "Validations" do
     it { should validate_presence_of(:full_name) }
     it_behaves_like "a record that validates device timestamps"
+
+    pending { is_expected.to validate_presence_of(:access_level) }
```
kitallis, line 13: This is intentionally left pending for when the migration is complete and we add this validation.

app/models/user.rb
```diff
@@ -11,33 +11,33 @@ class User < ApplicationRecord
     allowed: "allowed",
     denied: "denied"
   }, _prefix: true
+  enum access_level: {
+    viewer: "viewer",
```
harimohanraj89, line 15: Neat. This is where we'll also include `report_viewer` and `call_center`, right?

app/models/user.rb
```diff
@@ -53,7 +53,7 @@ class User < ApplicationRecord
 
   validates :full_name, presence: true
   validates :role, presence: true, if: -> { email_authentication.present? }
-
+  # validates :access_level, presence: true, if: -> { email_authentication.present? }
```
harimohanraj89, line 56: Keeping this off during the migration I guess? If we haven't already, let's add a story to turn it back on so we don't forget.

app/models/user.rb
```diff
@@ -179,12 +178,41 @@ def resources
     user_permissions.map(&:resource)
   end
 
+  # #########################
+  # User Access (permissions)
+  #
+  def accessible_organizations(action)
```
harimohanraj89, line 184: Could we use named arguments? Even though there's just one, I feel like these methods are definitely gonna be revisited and tweaked.

app/models/user.rb
```diff
@@ -179,12 +178,41 @@ def resources
...
+    accesses.facility_groups(action)
+  end
+
+  def accessible_facilities(action)
+    return Facility.all if power_user?
+    accesses.facilities(action)
+  end
+
+  def can?(action, model, record = nil)
+    return true if power_user?
```
harimohanraj89, line 200: This is beautiful.

app/models/user.rb
```diff
@@ -179,12 +178,41 @@ def resources
...
+    accesses.can?(action, model, record)
+  end
+  #
+  # #########################
+
   def destroy_email_authentications
-    destroyable_email_authentications = email_authentications.load
+    destroyable_email_auths = email_authentications.load
```
harimohanraj89, line 207: Haha was it really that bad? 😄

app/models/user.rb
kitallis: Yup, I added a comment in the [tests](https://github.com/simpledotorg/simple-server/pull/1185/files#r470619885) for this, forgot to add the comment here. I'll add a story as well.

app/models/user.rb
kitallis: RubyMine was constantly spitting warnings for E_TOO_LONG_VARIABLE_NAMES.

app/models/user.rb
kitallis: I'll do this towards the end, since my later PRs and Vikram's PRs depend on this API, don't want that to break right now. Will stick a card for it.

harimohanraj89: Haha that kinda validates why an interface that _doesn't_ require change across the codebase is nice 😂  Sounds good 👍

PR #1189, opened 2020-07-31
Facility details beneath each District

claudiovallejo:
**Story card:** https://app.clubhouse.io/simpledotorg/story/435/facility-details-beneath-each-district

## Because

## This addresses
* "HTN controlled" facilities table
* "No BP measure" facilities table
* "HTN not under control" facilities table
* "Cumulative registrations" table

image


app/assets/javascripts/application.js
```diff
@@ -27,6 +27,22 @@
 
 $(function () {
   $('[data-toggle="tooltip"]').tooltip()
+  
+  if($('#htn-controlled-table').length) {
```
kitallis, line 31: Let's pull these table names out into constants and just use them in both the `if` check and the actual element fetch.

kitallis: I think we can pull these selectors out into the specific page they refer to like `reports.js` or `user_analytics.js` etc.

kitallis:
Other than some minor JS restructuring, looks good to me 👍

app/assets/javascripts/application.js
claudiovallejo: Thanks @kitallis, I went ahead and moved all reports-specific selectors to `reports.js`. I'll create a separate story and PR to move analytics table selector to `user_analytics.js`.

app/assets/javascripts/reports.js
```diff
@@ -1,14 +1,42 @@
...
+
+function initializeTables() {
+  const tableSortConfig = { descending: true };
+
+  const cumulativeRegistrationsTable = document.getElementById("cumulative-registrations-table");
+  const htnNotUnderControlTable = document.getElementById("htn-not-under-control-table");
+  const noBPMeasureTable = document.getElementById("no-bp-measure-table");
+  const htnControlledTable = document.getElementById("htn-controlled-table");
+
+  if (htnControlledTable) {
```
kitallis, line 24: We seem to be doing this kind of thing a lot, in other places as well, and [not just tablesort](https://github.com/simpledotorg/simple-server/blob/master/app/assets/javascripts/user_analytics.js#L5) I could suggest an initialize function for things like this,

```javascript
function initElementById(element, func) { 
    if (document.getElementById(element)) { 
        func(); 
    } 
}

const htnControlTables = "htn-control-table"
const noBPMeasureTable = "no-bp-measure-table"

initElementById(htnControlTables,  () => new Tablesort(htnControlTables, tableSortConfig))
initElementById(noBPMeasureTable,  () => new Tablesort(noBPMeasureTable, tableSortConfig))
```

Probably not worth it for now though, maybe we can do this later.

rsanheim:
Now that we have the bulk caching in place, listing all the facility details should be doable.

I also added caching for the NoBPMeasure queries...this time doing it at the per-period level instead of for an entire range of values.  The trade off here is that other callers who need smaller or larger result sets will hit the cache, with the downside being we are going to be way chattier w/ Redis. We could use a multi fetch type thing for Redis to pretty easily clean up that piece up though.

app/services/no_bp_measure_service.rb
```diff
@@ -19,9 +22,15 @@ def call
 
   def visited_without_bp_taken_count(period)
     return 0 if facilities.empty?
+    Rails.cache.fetch(cache_key(period), version: cache_version, expires_in: CACHE_TTL, force: force_cache?) do
+      execute_sql(period)
+    end
+  end
+
+  def execute_sql(period)
```
rsanheim, line 30: alternative approach to https://github.com/simpledotorg/simple-server/pull/1260

PR #1194, opened 2020-08-03
Recent BPs table

claudiovallejo:
### Note
This is branched from the "Facility details beneath each District" PR. Let's merge the ["Facility details beneath each District" PR](https://github.com/simpledotorg/simple-server/pull/1189) first.

---

**Story card:** https://app.clubhouse.io/simpledotorg/story/448/recent-bps-card-for-facility-reports

## Because

Created the "Recent BPs" table for Facility reports in the details section

## This addresses

image

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -28,5 +28,23 @@
 $(function () {
   $('[data-toggle="tooltip"]').tooltip()
 
-  new Tablesort(document.getElementById('analytics-table'), { descending: true })
+  if($('analytics-table').length) {
+    new Tablesort(document.getElementById('analytics-table'), { descending: true })
+  }
+
+  if($('#htn-controlled-table').length) {
+    new Tablesort(document.getElementById('htn-controlled-table'), { descending: true })
+  }
+
+  if($('#no-bp-measure-table').length) {
+    new Tablesort(document.getElementById('no-bp-measure-table'), { descending: true })
+  }
+
+  if($('#htn-not-under-control-table').length) {
+    new Tablesort(document.getElementById('htn-not-under-control-table'), { descending: true })
+  }
+
+  if($('#cumulative-registrations-table').length) {
+    new Tablesort(document.getElementById('cumulative-registrations-table'), { descending: true })
+  }
 });
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -28,7 +28,7 @@
 $(function () {
   $('[data-toggle="tooltip"]').tooltip()
 
-  if($('analytics-table').length) {
+  if($('#analytics-table').length) {
     new Tablesort(document.getElementById('analytics-table'), { descending: true })
   }
 
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -47,4 +47,8 @@ $(function () {
   if($('#cumulative-registrations-table').length) {
     new Tablesort(document.getElementById('cumulative-registrations-table'), { descending: true })
   }
+
+  if($('#recent-bps').length) {
+    new Tablesort(document.getElementById('recent-bps'), { descending: true })
+  }
 });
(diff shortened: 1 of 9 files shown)


app/views/shared/_recent_bp_log.html.erb
```diff
@@ -7,8 +7,8 @@
       by this user.
     <% end %></p>
   <% if blood_pressures.any? %>
-    <div class="table-responsive table-compact pb-3">
-      <table>
+    <div>
+      <table id="recent-bps" class="table-responsive table-compact">
```
claudiovallejo, line 11: Fixing an error log. `Tablesort` expects the target element to be a table.

app/views/reports/regions/show.html.erb
```diff
@@ -272,24 +272,214 @@
...
+            <tbody>
+              <% @region.facilities.order(:name).each do |facility| %>
+                <tr>
+                  <td class="ta-left">
+                    <%= link_to(reports_region_path(facility.region_slug)) do %>
+                      <%= facility.name %>
+                    <% end %>
+                  </td>
+                  <td class="ta-left">
+                    PENDING
```
kitallis, line 370: These `PENDING` here are intentional?

app/controllers/reports/regions_controller.rb
```diff
@@ -36,6 +41,15 @@ def details
     @quarterly_registrations = @data[:quarterly_registrations]
     @top_region_benchmarks = @data[:top_region_benchmarks]
     @last_registration_value = @data[:registrations].values&.last || 0
+
+    unless @region.is_a?(FacilityGroup)
+      @recent_blood_pressures = @region
```
kitallis, line 46: Wondering if this will read neater if we move this `recent_blood_pressures` method to the `Facility` model itself (and other regions subsequently as necessary) since it's being used in multiple places now.

So then, we could just do `@region.recent_blood_pressures`.

app/controllers/reports/regions_controller.rb
```diff
@@ -36,6 +41,15 @@ def details
     @quarterly_registrations = @data[:quarterly_registrations]
     @top_region_benchmarks = @data[:top_region_benchmarks]
     @last_registration_value = @data[:registrations].values&.last || 0
+
+    unless @region.is_a?(FacilityGroup)
```
kitallis, line 45: Is it going to be shown in pages other than `Facility`-level reports? If not, then should we just make this 

```ruby
if @region.is_a?(Facility)
```

app/assets/javascripts/application.js
```diff
@@ -28,5 +28,27 @@
 $(function () {
   $('[data-toggle="tooltip"]').tooltip()
 
-  new Tablesort(document.getElementById('analytics-table'), { descending: true })
+  if($('#analytics-table').length) {
```
kitallis, line 31: I left a comment about this JS in the other [PR](https://github.com/simpledotorg/simple-server/pull/1189)

app/views/reports/regions/show.html.erb
claudiovallejo: They're intentional and addressed in (this PR)[https://github.com/simpledotorg/simple-server/pull/1189]

app/controllers/reports/regions_controller.rb
claudiovallejo: Good call @kitallis. Fixed!

app/assets/javascripts/application.js
claudiovallejo: Thanks! I'll move non-reports related conditionals/selectors to JS view files in a separate PR

app/controllers/analytics/facilities_controller.rb
```diff
@@ -11,12 +11,7 @@ def show
     set_dashboard_analytics(@period, 6)
     set_cohort_analytics(@period, @prev_periods)
 
-    @recent_blood_pressures = @facility
-      .blood_pressures
-      .includes(:patient, :user)
-      .order(Arel.sql("DATE(recorded_at) DESC, recorded_at ASC"))
-
-    @recent_blood_pressures = paginate(@recent_blood_pressures)
+    @recent_blood_pressures = paginate(@facility.recent_blood_pressures)
```
kitallis, line 14: Yay!

PR #1196, opened 2020-08-03
Reports: No BP Measure card

rsanheim:
**Story card:** https://app.clubhouse.io/simpledotorg/story/431/no-bp-measure-card

image

## Because

## This addresses

### DIFF
--- a/app/services/region_report_service.rb
+++ b/app/services/region_report_service.rb
@@ -13,9 +13,11 @@ class RegionReportService
     @range = start_period..@period
     @data = {
       controlled_patients: {},
-      registrations: {},
       cumulative_registrations: 0,
+      missed_visits: {},
+      missed_visits_rate: {},
       quarterly_registrations: [],
+      registrations: {},
       top_region_benchmarks: {}
     }.with_indifferent_access
   end
@@ -33,14 +35,29 @@ class RegionReportService
     data.merge! compile_cohort_trend_data
     data[:visited_without_bp_taken] = count_visited_without_bp_taken
     data[:visited_without_bp_taken_rate] = percentage_visited_without_bp_taken
+    data[:missed_visits] = calculate_missed_visits
+    data[:missed_visits_rate] = calculate_missed_visits_rate
     data[:top_region_benchmarks].merge!(top_region_benchmarks)
 
     data
   end
 
+  def calculate_missed_visits
+    data[:cumulative_registrations].each_with_object({}) do |(period, count), hsh|
+      hsh[period] = count - data[:controlled_patients][period] - data[:uncontrolled_patients][period]
+    end
+  end
+
+  def calculate_missed_visits_rate
+    data[:missed_visits].each_with_object({}) do |(period, count), hsh|
+      hsh[period] = percentage(count, data[:cumulative_registrations][period].to_f)
+    end
+  end
+
   # visited in last 3 months but had no BP taken
   def count_visited_without_bp_taken
-    range.each_with_object({}) do |period, hsh|
+    periods = data[:registrations].keys
+    periods.each_with_object({}) do |period, hsh|
       visits_without_bp = Appointment.from(patients_visited_without_bp_taken(period), "appointments").count("appointments.patient_id")
       hsh[period] = visits_without_bp
     end
@@ -65,7 +82,7 @@ class RegionReportService
     end_date = period.to_date
     begin_date = period.advance(months: -3).to_date
     Appointment.select("DISTINCT ON (appointments.patient_id) appointments.*")
-      .joins(:patient)
+      .joins(:patient).merge(Patient.with_hypertension)
       .joins("LEFT OUTER JOIN blood_pressures ON blood_pressures.patient_id = appointments.patient_id")
       .order("appointments.patient_id")
       .where("blood_pressures.id IS NULL")
--- a/app/services/region_report_service.rb
+++ b/app/services/region_report_service.rb
@@ -57,49 +57,18 @@ class RegionReportService
   # visited in last 3 months but had no BP taken
   def count_visited_without_bp_taken
     periods = data[:registrations].keys
-    periods.each_with_object({}) do |period, hsh|
-      visits_without_bp = patients_visited_without_bp_taken(period).count
-      hsh[period] = visits_without_bp
-    end
+    VisitedButNoBPService.new(region, periods: periods).call
   end
 
   def percentage_visited_without_bp_taken
     data[:visited_without_bp_taken].each_with_object({}) do |(period, count), hsh|
-      hsh[period] = if count&.zero?
-        0
-      else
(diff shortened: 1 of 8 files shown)


claudiovallejo:
@rsanheim viewing this branch locally and the control + uncontrol + no bp taken rates add up to 104%. Potential bug?

image

app/services/region_report_service.rb
```diff
@@ -23,20 +27,71 @@ def initialize(region:, period:, current_user:)
...
+  def patients_visited_without_bp_taken(period)
+    end_date = period.to_date
+    begin_date = period.advance(months: -3).to_date
+    Appointment.select("DISTINCT ON (appointments.patient_id) appointments.*")
+      .joins(:patient).merge(Patient.with_hypertension)
+      .joins("LEFT OUTER JOIN blood_pressures ON blood_pressures.patient_id = appointments.patient_id")
+      .order("appointments.patient_id")
+      .where("blood_pressures.id IS NULL")
+      .where("appointments.device_created_at :: DATE BETWEEN :begin AND :end", begin: begin_date, end: end_date)
+      .where("patients.registration_facility_id in (?)", facilities)
```
rsanheim, line 90: @simpledotorg/server-developers how does this query look for patients who have a visit but no BP taken?

kitallis: This looks correct mostly, but I feel like this:

```
where("patients.registration_facility_id in (?)", facilities)
```

Should instead be on the **creation** facility of the appointment instead?

It'd make sense to say that a patient is counted as "visited" if they have an appointment created at any of the facilities in the Region.

wdyt?

Also, we're not explicitly checking for prescription drugs, are we using appointments as proxy for meds updated? If so, I think that seems sensible.

harimohanraj89: My interpretation was that the facility scoping here is the "assigned facility". I don't care where you scheduled your appointment, I care that the patient I was responsible for had an appointment scheduled, aka, they visited any clinic. I suspect this will become `patients.assigned_facility_id in (?)` soon.

kitallis: 👍

app/services/visited_but_no_bp_service.rb
```diff
@@ -0,0 +1,54 @@
...
+  def no_blood_pressures_recorded(period)
+    <<-SQL
+      NOT EXISTS (
+        SELECT 1
+        FROM blood_pressures bps
+        WHERE patients.id = bps.patient_id
+          AND #{between_clause("bps", "recorded_at", period)}
+      )
+    SQL
+  end
```
rsanheim, line 46: ☝️

kitallis: This looks correct, but I don't understand where `patients` is appearing from? Raw SQL is very hard to understand when it's partially applied. Is this basically,

```
BloodPressure.where.not(patient: patients, recorded_at: begin_date...end_date).exists?
```

If so, 

Could we re-write `patients_visited_with_no_bp_taken` with a normal `left_outer_joins` instead?

```
def between(period)
  begin_date = period.advance(months: -3).start_date
  end_date = period.end_date
  Range.new(begin_date, end_date)
end

def patients_visited_with_no_bp_taken(period)
  Patient
    .distinct
    .with_hypertension
    .where(registration_facility: facilities)
    .left_outer_joins(:appointments).where(appointments: {recorded_at: between(period)})
    .left_outer_joins(:prescription_drugs).where(prescription_drugs: {recorded_at: between(period)})
    .left_outer_joins(:blood_sugars).where(blood_sugars: {recorded_at: between(period)})
    .left_outer_joins(:blood_pressures).where(blood_pressures: {id: nil})
    .where("appointments.id IS NOT NULL OR prescription_drugs.id IS NOT NULL OR blood_sugars.id IS NOT NULL")
end
```

In the future, this may compose a lot better.

rsanheim:
Going to punt on lost to follow up, need to try to get it into one query.

app/services/visited_but_no_bp_service.rb
rsanheim: I dont' think any of these ActiveRecord based approaches are really clear, to be honest 😄 . I think these kind of queries are really straining the abilities of our ORM, and would rather we just drop into SQL....especially because this should really be **one** query, instead of 24 queries repeated for each period.

I'm going to look into getting some of these queries rolled up into one via the Postgres [generate_series](https://www.postgresql.org/docs/10/functions-srf.html) functions, which will make these queries longer but should be able to get all our data in one trip.  When I do that I'll probably try it out with [GitHub::SQL](https://github.com/github/github-ds) to see how that works out.

kitallis: I don't mind dropping into SQL at all. I have used query builders (like the one you're suggesting) in the past and even [written](https://github.com/nilenso/honeysql-postgres) one and my general feeling is that they lie under the pretense of an abstraction but aren't really actually one (you can't compose relational algebra using strings, you lose the mapping to `ActiveRecord`). However, they are far better than raw SQL which generally ends up leaking subtle bugs. I also agree if it's performant and not a terribly complicated query to write raw, it's better to write a single one. If you're looking to migrate to that, I'm 👍.

However, our queries are operating on well-defined relations (and not flat / arbitrary table structures) and we can probably write ~95% of them easily with AR. Writing with AR has many advantages (and some disadvantages as well):

* They compose better
* They are easier to drive through the REPL
* Atomic pieces can be tested in isolation
* Useful default scopes are pre-applied
* The SQL building noise is hidden away from application code
* Lesser need to munge / interpolate strings awkwardly

Disadvantages:

* Can't express everything (postgres is vast and not everything is supported)
* The queries may not always be performant
* Some scopes in AR can bite you, so you have to be a diligent 

That said, writing SQL can often "feel" a lot better because, you can have one giant SQL to look at that does all the things that you think it should do and you operate within that space. For complex reporting queries that need to be performant this may very well be a good choice (I just haven't encountered a scenario to justify it yet other than the materialized views).

Specifically about the refactor,

I'm sure you can tweak the name of `scope_visits_by` to suggest it's `left_outer_join` intention clearly, my basic point was to avoid writing things like this,

```
   <<-SQL
      NOT EXISTS (
        SELECT 1
        FROM blood_pressures bps
        WHERE patients.id = bps.patient_id
          AND #{between_clause("bps", "recorded_at", period)}
      )
    SQL
```

Which is an invalid query by itself and isn't very clear that it only works when applied to specific parts of another larger SQL, which in comparison this,

```
  Patient.left_outer_joins(model).where(model => {time_field => period})
```

...is not. It's both testable and runnable in isolation.

And secondly, to not lose the original point 😛 , this was my main suggestion to arrive at the `no_bp` query,

```
  Patient.left_outer_joins(:blood_pressures).where(blood_pressures: {recorded_at: period, id: nil})
```

Just my 2c.

PR #1197, opened 2020-08-03
Added videos to the HELP in Simple

dburka:
**Story card:** No story

## Because

Added all of the training videos the HELP section

## This addresses

Making the training videos more accessible to healthcare workers.

## To do

* [ ] Ship https://github.com/simpledotorg/deployment/pull/232


harimohanraj89:
@dburka Confirming, we can add the same video to both India and Bangladesh users, right?

dburka:
@harimohanraj89 Yes... most of the videos are in English and are not IHCI-specific.

PR #1200, opened 2020-08-04
Assigned facility: Cohort charts

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/480

## Because
We are making changes to our reports to incorporate assigned facility.

## This addresses

This changes the Cohort report charts to be Assigned facility compliant. [Definition changes in detail](https://docs.google.com/document/d/1YZqm0t5NhTDUkn3Ww8yRqAszKCtAiOqChpFED6U-QXE/edit#heading=h.20ix6jjqi36g).


app/queries/cohort_analytics_query.rb
```diff
@@ -33,37 +33,37 @@ def patient_counts_by_period(period, prev_periods, from_time: Time.current)
   end
 
   def patient_counts(cohort_start, cohort_end, report_start, report_end)
-    registered_patients = registered(cohort_start, cohort_end)
-    followed_up_patients = followed_up(registered_patients, report_start, report_end)
+    cohort_patients = cohort_patients(cohort_start, cohort_end)
```
harimohanraj89, line 36: To be honest, I think it's ok to keep the `registered` terminology, since we scope the patients outside this area. Within here, I think it makes sense to say "given a set of patients, these are the total registered patients, these are the controlled, uncontrolled, and no-BPs".

As we start propagating assigned facility everywhere, we'll no longer have to get confused by the word "registration". Not a big deal either way.

app/models/facility.rb
```diff
@@ -98,7 +98,7 @@ def facilities
   end
 
   def cohort_analytics(period, prev_periods)
-    query = CohortAnalyticsQuery.new(registered_hypertension_patients)
+    query = CohortAnalyticsQuery.new(assigned_patients)
```
harimohanraj89, line 101: Are we including non-HTN patients by changing this to `assigned_patients`? We should preserve the HTN-only scoping we used to do, right?

kitallis: The scoping is preserved, will go ahead and change the name to `assigned_hypertension_patients`.

app/queries/cohort_analytics_query.rb
prabhanshuguptagit: Even within this context, the idea of pure `registrations` breaks down when we group by `assigned_facility`. I guess the `cohort_patients` method could be called `registrations`👍 . since after the grouping it becomes more like `assigned_and_registered`. Changed in 9740b1f.

app/queries/my_facilities/missed_visits_query.rb
```diff
@@ -57,8 +57,9 @@ def missed_visit_totals
 
   def total_patients_per_facility
     @total_patients_per_facility ||=
-      facilities
-        .assigned_patients
```
prabhanshuguptagit, line 61: This scope results in a `WHERE patients IN (..list of all patients in those facilities)` query which is extremely expensive to generate. Going back to `.where(assigned_facility: facilities)` for performance.

spec/exporters/patients_exporter_spec.rb
```diff
@@ -4,7 +4,12 @@
   include QuarterHelper
 
   let!(:facility) { create(:facility) }
-  let!(:patient) { create(:patient, registration_facility: facility, status: "dead") }
+  let!(:patient) {
+    create(:patient,
+      registration_facility: facility,
+      status: "dead",
+      address: create(:address, village_or_colony: Faker::Address.city)) # need a different village and zone
```
prabhanshuguptagit, line 11: This is an unrelated flaky spec fix.

spec/exporters/patients_with_history_exporter_spec.rb
```diff
@@ -5,7 +5,11 @@
 
   let!(:now) { Time.current }
 
-  let!(:patient) { create(:patient, status: "dead") }
+  let!(:patient) {
+    create(:patient,
+      status: "dead",
+      address: create(:address, village_or_colony: Faker::Address.city)) # need a different village and zone
```
prabhanshuguptagit, line 11: Another flaky spec.

app/queries/cohort_analytics_query.rb
```diff
@@ -33,26 +33,26 @@ def patient_counts_by_period(period, prev_periods, from_time: Time.current)
...
 
-    controlled_counts = controlled_patients.group(:registration_facility_id).size.symbolize_keys
+    controlled_counts = controlled_patients.group(:assigned_facility_id).size.symbolize_keys
     uncontrolled_counts = followed_up_counts.merge(controlled_counts) { |_, followed_up, controlled|
       followed_up - controlled
     }
 
     {
-      registered: {total: registered_patients.size, **registered_counts},
+      cohort_patients: {total: cohort_patients.size, **cohort_patient_counts},
```
harimohanraj89, line 53: Is this payload cached somewhere? If so, one pattern I've grown fond of is introducing a version number in the cache key. Schema changes like this can be accompanied by bump in the version number, so we don't have to go clear caches as soon as we deploy changes like this.

[Example](https://github.com/simpledotorg/simple-server/blob/progress-control-01/app/services/control_rate_service.rb#L2).

spec/models/facility_spec.rb
```diff
@@ -30,23 +30,6 @@
...
-    end
-
-    it "returns assigned patients for facilities" do
-      expect(Facility.where(id: assigned_facilities).assigned_patients).to match_array patients
-    end
-
-    it "ignores registration patients" do
-      expect(Facility.where(id: registration_facility).assigned_patients).to be_empty
-    end
-  end
```
harimohanraj89, line 48: Should we reinstate specs for `assigned_hypertension_patients`, or remove that scope from `Facility`?

app/models/facility.rb
```diff
@@ -108,7 +104,7 @@ def facilities
   end
 
   def cohort_analytics(period, prev_periods)
-    query = CohortAnalyticsQuery.new(registered_hypertension_patients)
+    query = CohortAnalyticsQuery.new(assigned_hypertension_patients)
```
harimohanraj89, line 107: Apart from naming, I believe this is the meaty change of this PR.

prabhanshuguptagit: This is one meaty change, the other one is grouping by `:assigned_facility_id` in `patient_counts`.

spec/models/facility_spec.rb
prabhanshuguptagit: Will reinstate the `assigned_hypertension_patients` specs with a shoulda matcher. 👍

app/models/facility.rb
harimohanraj89: Ah, yes. 👍

app/queries/cohort_analytics_query.rb
prabhanshuguptagit: This is [cached here](https://github.com/simpledotorg/simple-server/blob/23c26d04ef39c41db53d0aa44132911436865e99/app/controllers/analytics_controller.rb#L50). Added the version number in e4f26d6.

PR #1204, opened 2020-08-06
README details around dependencies and puma issues

timcheadle:
I ran into issues setting up a new dev environment and wanted to document the dependencies and puma issues I fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,9 +63,9 @@ rails db:setup
 
 ##### Issues with MacOS Catalina and Puma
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,6 +61,31 @@ rake yarn:install
 rails db:setup
 ```
 
+##### Issues with MacOS Catalina and Puma
+
+If you encounter issues installing the `puma` gem on MacOS Catalina, it may be due to compiler upgrades Apple issued.
+
+You may see errors like this:
```
harimohanraj89, line 68: We could just note which step in the setup process raises these errors? I'm assuming `bin/setup`?

PR #1208, opened 2020-08-07
Permissions: flesh out roles and actions

vkrmis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/128/support-for-adding-and-removing-permissions-to-an-access-level

## Because
We need more sophisticated roles than just `view` and `manage` to support current and future use cases for the dashboard

### Pages/Controllers migrated with a feature toggle:

- [x] `OrganizationsController`
- [x] `Admin::FacilitiesController`
- [x] `Admin::FacilityGroupsController`
- [x] `MyFacilitiesController`

## This addresses

Adds the roles and permissions we need

[PRD](https://www.notion.so/Admin-Permissions-fc9f306ed0a948189f342d15166ac252)
[Designs](https://www.figma.com/file/9inSvRD4dXcMfXchlyDZVV/Dashboard-(Production)?node-id=909%3A1801)


app/controllers/organizations_controller.rb
```diff
@@ -1,15 +1,22 @@
 class OrganizationsController < AdminController
   include Pagination
+  skip_after_action :verify_authorized
 
   def index
     authorize(:dashboard, :show?)
```
vkrmis, line 6: Left this in for now, will get rid of it once we add a basic data migration

app/controllers/organizations_controller.rb
```diff
@@ -1,15 +1,22 @@
...
+    current_admin.can?(:view_reports, :facility_group)
```
vkrmis, line 7: We need an `authorized?` method similar to `can?` that will throw a `Pundit::NotAuthorizedError` equivalent instead of returning a boolean.

app/controllers/organizations_controller.rb
```diff
@@ -1,15 +1,22 @@
...
 
     @users_requesting_approval = policy_scope([:manage, :user, User])
       .requested_sync_approval
       .order(updated_at: :desc)
 
     @users_requesting_approval = paginate(@users_requesting_approval)
 
-    @organizations = policy_scope([:cohort_report, Organization]).order(:name)
+    @organizations = current_admin
```
vkrmis, line 15: Our new scoping strategy is unidirectional, so we need to go from FacilityGroup to Organization manually.

app/models/access.rb
```diff
@@ -53,6 +57,10 @@ def can?(action, model, record = nil)
       end
     end
 
+    def super_admin?
```
vkrmis, line 60: Moving this up to be a public method so we can delegate from User.
Its useful for doing `user.super_admin?` checks to toggle links etc.

app/views/layouts/application.html.erb
```diff
@@ -56,19 +56,19 @@
           <% if email_authentication_signed_in? %>
             <ul class="navbar-nav mr-auto">
 
-              <% if policy(:dashboard).view_my_facilities? %>
+              <% if current_admin.can?(:view_reports, :facility) %>
```
vkrmis, line 59: Migrating existing policy checks

app/views/layouts/application.html.erb
```diff
@@ -78,31 +78,31 @@
...
-                  <% if policy([:dashboard]).manage_admins? %>
+                  <% if current_admin.can?(:manage, :facility) %>
                     <%= link_to "Admins", admins_path, class: "dropdown-item #{active_controller?("email_authentications")}" %>
                   <% end %>
-                  <% if policy([:dashboard]).manage_users? %>
+                  <% if current_admin.can?(:manage, :facility) %>
                     <%= link_to "Users", admin_users_path, class: "dropdown-item #{active_controller?("users")}" %>
                   <% end %>
-                  <% if policy([:dashboard]).view_sidekiq_ui? %>
+                  <% if current_admin.super_admin? %>
```
vkrmis, line 102: `user.super_admin?` comes in handy here

app/views/organizations/index.html.erb
```diff
@@ -1,14 +1,11 @@
...
-
       <div class="col-sm-5 col-lg-7">
         <% @organizations.each do |organization| %>
             <section id="facility-groups">
               <div class="card" title="<%=organization.name%>">
                 <h2><%= organization.name %></h2>
-
-
-                    <% policy_scope([:cohort_report, organization.facility_groups])
+                    <% current_admin.accessible_facility_groups(:view_reports).where(organization: organization)
```
vkrmis, line 8: This is potentially a bad first example to migrate because it does some facility.district weirdness.
I've made the most literal translation to the new permissions so its comparable. Might end up refactoring this quite a bit.

app/models/access.rb
```diff
@@ -1,17 +1,21 @@
...
   belongs_to :resource, polymorphic: true, optional: true
 
   enum role: {
-    viewer: "viewer",
+    super_admin: "super_admin",
     manager: "manager",
-    super_admin: "super_admin"
+    health_care_worker: "health_care_worker",
+    health_administrator: "health_administrator",
+    call_center_worker: "call_center_worker"
```
harimohanraj89, line 18: Following up from [https://simpledotorg.slack.com/archives/CK95QFJA2/p1596812158352700](here), I think this is a good place to start:
* manager
* viewer
* viewer_reports_only
* call_center

app/models/access.rb
```diff
@@ -1,17 +1,21 @@
 class Access < ApplicationRecord
   ALLOWED_RESOURCES = %w[Organization FacilityGroup Facility].freeze
   ACTION_TO_ROLE = {
-    view: [:manager, :viewer],
+    manage_overdue_list: [:manager, :health_care_worker, :call_center_worker],
+    view_reports: [:manager, :health_care_worker, :health_administrator],
```
harimohanraj89, line 5: Considering that this is the predominant view method in the dashboard, and that PII is checked only in a couple of places (overdue list, patient line list downloads), could we name this `view`?

vkrmis: We'd be relying on people to know/remember that `view` = `view_reports` and `view` != `view_pii`. 
Since `view_reports` and `view_pii` are orthogonal, I'd prefer to keep them explicit.

Separate clarification: the overdue list page will actually use the `manage_overdue_list` check, because it allows state changes as well + we wouldn't want to give users with the `call_center` role access to any PII outside of the overdue list.

app/models/access.rb
vkrmis: I'd like to be explicit about the `viewer` role having access to reports + pii. How about `all_viewer` or `viewer_all`?

app/controllers/admin/facilities_controller.rb
```diff
@@ -8,16 +8,42 @@ class Admin::FacilitiesController < AdminController
   before_action :initialize_upload, :validate_file_type, :validate_file_size, :parse_file,
     :validate_facility_rows, if: :file_exists?, only: [:upload]
 
+  skip_after_action :verify_authorized, if: -> { Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) }
+  skip_after_action :verify_policy_scoped, if: -> { Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) }
+  after_action :verify_access_authorized, if: -> { Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) }
```
vkrmis, line 13: Gating these actions based on the Flipper flag.

app/controllers/admin/facilities_controller.rb
```diff
@@ -8,16 +8,42 @@ class Admin::FacilitiesController < AdminController
...
+
   def index
-    authorize([:manage, :facility, Facility])
+    if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin)
+      current_admin.authorize(:manage, :facility)
+    else
+      authorize([:manage, :facility, Facility])
+    end
```
vkrmis, line 20:
```
if Flipper.enabled?
     check new permissions
else
     check old permissions 
end
```

This pattern gets repeated throughout. We can get rid of all these conditionals when we make the full transition

app/controllers/admin/facilities_controller.rb
```diff
@@ -8,16 +8,42 @@ class Admin::FacilitiesController < AdminController
...
+      else
+        policy_scope([:manage, :facility, Facility]).search_by_name(search_query)
+      end
+
       facility_groups = FacilityGroup.where(facilities: facilities)
 
       @organizations = Organization.where(facility_groups: facility_groups)
       @facility_groups = facility_groups.group_by(&:organization)
       @facilities = facilities.group_by(&:facility_group)
+    elsif Flipper.enabled?(:new_permissions_system_aug_2020, current_admin)
```
vkrmis, line 34: standardrb complains about nested conditionals, turned it into an `elsif` even though its a little harder to read

spec/flipper_helper.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def disable_flag(*args)
+    allow_any_instance_of(Flipper).to receive(:enabled?).with(*args).and_return(false)
+  end
+
+  def with_flag_enabled(*args)
+    enable_flag(*args)
+    yield
+    disable_flag(*args)
+  end
```
vkrmis, line 14: Added these helpers for testing behaviour when flags are enabled vs disabled.

app/models/user_access.rb
```diff
@@ -67,6 +95,48 @@ def can_access_record?(resources, record)
     resources.exists?
   end
 
+  def can_create?(model, record)
+    case model
+    when :facility
```
kitallis, line 100: Can we write this in a way where we don't repeat this case-tree? It'll could be quite confusing to add new `model` names in 3 different places.

app/models/user_access.rb
```diff
@@ -67,6 +95,48 @@ def can_access_record?(resources, record)
...
+    when :facility_group
+      resources&.find_by_id(record.organization_id).present?
+    when :organization
+      user.power_user?
+    else
+      raise ArgumentError, "Can't create #{model}."
+    end
+  end
+
+  def can_perform_action?(action, model, record = nil)
```
kitallis, line 124: This is only called when record is `nil` right? Why does it need the third argument?

app/models/user_access.rb
```diff
@@ -67,6 +95,48 @@ def can_access_record?(resources, record)
     resources.exists?
   end
 
+  def can_create?(model, record)
+    case model
+    when :facility
+      can_create_record?(:facility, record, accessible_facility_groups(:manage))
```
kitallis, line 101: Why redirect to the `can_create_record?` methods? Why not just inline that behavior here?

app/controllers/admin/facility_groups_controller.rb
```diff
@@ -51,7 +58,11 @@ def destroy
   private
 
   def set_organizations
-    @organizations = policy_scope([:manage, :facility, Organization])
+    @organizations = if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin)
+      current_admin.accessible_organizations(:manage).union(Organization.where(id: @facility_group&.organization_id))
```
vkrmis, line 62: `union` the current facility_group's organization so an admin with facility_group level access can use the edit page

app/models/user_access.rb
```diff
@@ -1,18 +1,33 @@
 class UserAccess < Struct.new(:user)
   class NotAuthorizedError < StandardError; end
+  class AuthorizationNotPerformedError < StandardError; end
 
   LEVELS = {
```
vkrmis, line 5: Defines the access levels and some metadata for the views

app/models/user_access.rb
```diff
@@ -24,10 +39,14 @@ class NotAuthorizedError < StandardError; end
   }.freeze
 
   ACTION_TO_LEVEL = {
```
vkrmis, line 41: Maps actions to levels

app/models/user_access.rb
```diff
@@ -24,10 +39,14 @@ class NotAuthorizedError < StandardError; end
...
-    view: [:manager, :viewer],
+    manage_overdue_list: [:manager, :viewer_all, :call_center],
+    view_reports: [:manager, :viewer_all, :viewer_reports_only],
+    view_pii: [:manager, :viewer_all],
     manage: [:manager]
   }.freeze
 
+  VALID_OPERATIONS = [:access_record, :access_any, :create]
```
vkrmis, line 48: We've added a new argument called `operation` to the `can?` and `authorize` methods.

This allows us to perform slightly different checks for 
- creating new records (`:create`)
- cases where no explicit record exists (`:access_any`)
- specific records (`:access_record`) 

We tried a version of this where we can infer these cases using `new_record?` and `record == nil`, but this is much cleaner.

See [this](https://github.com/simpledotorg/simple-server/blob/347206af26e89d3da909383ee45a19f676c74b19/app/models/user_access.rb#L62) for the other approach

app/models/user_access.rb
```diff
@@ -42,23 +61,27 @@ def accessible_facilities(action)
...
 
-  def can?(action, model, record = nil)
-    if record&.is_a? ActiveRecord::Relation
-      raise ArgumentError, "record should not be an ActiveRecord::Relation."
-    end
+  def can?(action, model, operation, record = nil)
+    raise ArgumentError, "record should not be an ActiveRecord::Relation." if record&.is_a? ActiveRecord::Relation
+    raise ArgumentError, "#{operation} is not a supported operation." unless operation.in? VALID_OPERATIONS
 
     case model
```
vkrmis, line 68: We dispatch on model, with each model getting its own `can_act_on_<model>` method. Keeps it clean

app/models/user_access.rb
```diff
@@ -133,6 +156,48 @@ def can_access_record?(resources, record)
     resources.exists?
   end
 
+  def can_act_on_facility?(action, operation, record)
```
vkrmis, line 159: Here we can do the appropriate checks for various operations based on the model and its nuances

spec/features/home_with_new_permissions_spec.rb
```diff
@@ -0,0 +1,146 @@
+require "rails_helper"
+
+RSpec.feature "Home page with new permissions", type: :feature do
```
vkrmis, line 3: Duplicated `spec/features/admin/facilities/index_with_new_permissions_spec.rb`, but perform the new permission checks instead. This feature spec covers both `admin/facilities_controller` and `admin/facility_groups_controller`

spec/flipper_helper.rb
```diff
@@ -0,0 +1,9 @@
+module FlipperHelpers
```
vkrmis, line 1: Helpers to toggle Flipper flags for specs

kitallis (approved):

PR #1210, opened 2020-08-07
Make registration counts consistent; various other fixes

rsanheim:
**Story card:** n/a fix for https://sentry.io/organizations/resolve-to-save-lives/issues/1829002291/?project=1217715

### DIFF
--- a/app/services/control_rate_service.rb
+++ b/app/services/control_rate_service.rb
@@ -46,17 +46,25 @@ class ControlRateService
 
       data[:registrations] = registration_counts
       data[:cumulative_registrations] = sum_cumulative_registrations
+      data[:registrations].delete_if { |period, value| !periods.cover?(period) }
+      data[:cumulative_registrations].delete_if { |period, value| !periods.cover?(period) }
+
       periods.each do |(period, count)|
-        data[:controlled_patients][period] = controlled_patients(period).count
+        controlled = controlled_patients(period).count
+        uncontrolled = uncontrolled_patients(period).count
+
+        data[:controlled_patients][period] = controlled
+        data[:uncontrolled_patients][period] = uncontrolled
+
+        # For quarterly reports the registration count is based on the cohort, so its from the previous period.
         registration_count = if quarterly_report?
-          # get the cohort counts for quarterly reports
           data[:registrations][period.previous] || 0
         else
           data[:cumulative_registrations][period]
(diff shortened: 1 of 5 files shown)


app/services/control_rate_service.rb
```diff
@@ -67,19 +75,26 @@ def call
...
     }
   end
 
   def registration_counts
-    range = (periods.begin.start_date..periods.end.end_date)
+    return @registration_counts if @registration_counts
     formatter = lambda { |v| quarterly_report? ? Period.quarter(v) : Period.month(v) }
-    result = region.registered_patients.with_hypertension.group_by_period(periods.begin.type, :recorded_at, {range: range, format: formatter}).count
-    result.drop_while { |period, count| count == 0 }.to_h
+    result = region.registered_patients.with_hypertension.group_by_period(periods.begin.type, :recorded_at, {format: formatter}).count
```
rsanheim, line 90: We now grab _all_ registrations w/ the same query, so we don't have to worry about adding in the registrations that happen before the `periods` range.

spec/services/top_region_service_spec.rb
```diff
@@ -89,7 +89,7 @@ def refresh_views
 
     refresh_views
 
-    service = TopRegionService.new([organization], Period.month(june_1.end_of_month), scope: :facility)
```
rsanheim, line 92: This was an error w/ the test setup, the actual code passes in the Period and doesn't go to end of month.

app/services/control_rate_service.rb
```diff
@@ -46,17 +46,25 @@ def call
 
       data[:registrations] = registration_counts
       data[:cumulative_registrations] = sum_cumulative_registrations
+      data[:registrations].delete_if { |period, value| !periods.cover?(period) }
+      data[:cumulative_registrations].delete_if { |period, value| !periods.cover?(period) }
```
rsanheim, line 50: This removes registration data that _is not_ part of the range requested. If we didn't do this calling code might end up displaying too much data in reports or charts.

app/services/control_rate_service.rb
harimohanraj89: I like this a lot. Grab everything and chop off what wasn't requested. Let groupdate take care of all the annoying boundary conditions and timezones and stuff.

PR #1216, opened 2020-08-10
Help bp passport 01

dburka:
**Story card:** No card

## Because

We want healthcare workers to know the BP Passport app exists

## This addresses

Adds BP Passport App info to HELP in the app.


app/assets/stylesheets/common_webview.scss
```diff
@@ -1,3 +1,36 @@
...
+$blue-mid: #9db8d2;
+$blue-light: #E9F3FF;
+$blue-dark: #0c3966;
+$green-light: #E0FFED;
+$green-mid: #5CFF9D;
+$green: #00B849;
+$green-dark: #007A31;
+
+$font-condensed: 'Roboto Condensed', sans-serif;
+$font-small: 11px;
```
harimohanraj89, line 30: I see these vars declared but don't see them used below. Are these overriding the parent theme?

config/locales/api/help/en.html
```diff
@@ -34,11 +34,24 @@
...
+                
+            </a>
+            <h3><span style="color: #FF3355; font-size: 90%; float: right;">NEW!</span> App for your patients</h3>
+            <p>Patients can now chart their own BPs and blood sugars with a simple app called the "BP Passport" app.</p>
+            <p>Ask your patients to go to the Android or iOS app store and install the "BP Passport" app.</p>
+            <p><a href="#" onclick="open_window('help-bp-passport-app'); return false">Learn more...</a></p>
```
harimohanraj89, line 53: Clicking this doesn't scroll to the top of the overlay. I noticed we did the scroll magic in the progress control tab. Can we port the same magic here too?

app/assets/stylesheets/common_webview.scss
dburka: I should have done this ages ago and set the universal variables as a partial that can be included in any/all of the SASS files. Did this now. Please check to make sure this is done correctly.

config/locales/api/help/en.html
dburka: Yes, can you help me to do this?

harimohanraj89:
Sweet, I've rejiggered the SASS variables a bit, and added the auto-scroll-to-top. I'm too deep in this PR now, I've requested another review from a server dev :)

app/assets/javascripts/help.js
```diff
@@ -1,13 +1,28 @@
 // Hides and shows each page of Help
 
-function open_window(id) {
+function openWindow(id, parentId) {
```
kitallis, line 3: This case is also fixed in this open [PR](https://github.com/simpledotorg/simple-server/pull/1221), but it should be a clean merge regardless.

app/assets/stylesheets/help.scss
```diff
@@ -1,3 +1,5 @@
+@import "partials/webview";
+
 html, body {
```
kitallis, line 3: Slightly unrelated, but related to `help.scss` refactor: I noticed that there's a help screen CSS [class](https://github.com/simpledotorg/simple-server/blob/master/app/assets/stylesheets/application.scss#L49) in `application.scss` – Is that supposed to be there and/or can we move it here?

PR #1218, opened 2020-08-10
Bug: Reports header active tab

claudiovallejo:
**Story card:** https://app.clubhouse.io/simpledotorg/story/805/fix-header-styling-bug

## Because

Selecting a new date from the dropdown in reports removes the red font color from the active section.

## This addresses

* Removes `utf8=✓` to the url by adding `enforce_utf8: false` to the `form_for`
* **Note:** Seems like this is part of Rails to fix an issue with IE not rendering characters [read more here](https://stackoverflow.com/questions/4104474/rails-3-utf-8-query-string-showing-up-in-url)


rsanheim:
This needs to have conflicts resolved.

PR #1219, opened 2020-08-11
No BP Measure queries / charts

rsanheim:
**Story card:** https://app.clubhouse.io/simpledotorg/story/431/no-bp-measure-card

## Because

## This addresses


app/services/no_bp_measure_service.rb
```diff
@@ -0,0 +1,88 @@
...
+      end
+    end
+  end
+
+  def missed_visits_for(period)
+    if type == :missed_visits
+      visit_start_range = period.advance(years: -1).start_date
+      visit_end_range = period.advance(months: -3).start_date
+    else
+      visit_start_range = "-infinity"
```
rsanheim, line 29: This is [a valid Postgres special data/time](https://www.postgresql.org/docs/10/datatype-datetime.html#DATATYPE-DATETIME-SPECIAL-TABLE).

app/services/no_bp_measure_service.rb
```diff
@@ -0,0 +1,88 @@
...
+      visit_end_range = period.advance(years: -1).start_date
+    end
+    bind_attributes = {
+      hypertension: "yes",
+      facilities: facilities.map(&:id),
+      visit_start_range: visit_start_range,
+      visit_end_range: visit_end_range,
+      bp_start_range: period.advance(months: -3).start_date,
+      bp_end_range: period.end_date
```
rsanheim, line 38: Once we remove the ability to see the current month, this can be changed to just be start_date to start_date.  As it stands, this could be _almost_ four months of control data.

PR #1222, opened 2020-08-12
Invite Admin Flow (UI for new Access system)

kitallis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/847/as-a-power-user-i-can-invite-other-access-levels-for-the-selected-organizations

## Because

We want a newer, better, faster, stronger permissions system.

## This addresses

* Add the Invite admin screen flow behind a flipper toggle
* Allow granting access based on your access level and current accesses

## This does not address

* Editing an admin
* Viewing accesses for the admin against the new system

## Dark launch

This currently doesn't do a great job at dark launching because it's overly complicated to make that work. 

* Making a new devise controller is complicated because sessions state is not shared, so if you migrate to a different controller you will log out.
* I can't easily dispatch to a different view from the devise-inherited controller methods because the `new` function doesn't yield internally.

Instead, what we have for now, is that if you have the flipper toggle enabled, you'll see the new invite flow (but unable to see the old flow) or else you would see the old flow as usual.

## Screens

**Common**

Note that the access level dropdown will change depending on the current user's level.

Screenshot 2020-08-26 at 2 40 29 PM

**For a manager**

Screenshot 2020-08-26 at 2 39 27 PM

**For a power user**

Note that if you have full access, the fraction is not shown.

Screenshot 2020-08-26 at 2 39 55 PM

**Note:** Viewers will not be able to invite admins.


app/presenters/admin_access_presenter.rb
```diff
@@ -0,0 +1,18 @@
+class AdminAccessPresenter < SimpleDelegator
```
kitallis, line 1: This `Presenter` isn't currently pulling its weight yet, but hopefully with view + edit it will. If it doesn't, then well, I'll just stick these few methods back in the controller.

app/views/email_authentications/invitations/_access_tree.html.erb
```diff
@@ -0,0 +1,53 @@
+<div class="card access-tree">
```
kitallis, line 1: Pulling this out in the partial will come in handy when we use the same general thing for edit and view both.

app/views/email_authentications/invitations/new.html.erb
```diff
@@ -1,12 +1,33 @@
...
-                    submit_text: 'Invite Admin',
-                    submit_method: 'POST' %>
\ No newline at end of file
+        <%= form.label "Facility Access *" %>
+        <div id="facility-access">
+          <%= render "email_authentications/invitations/access_tree",
+              access_tree: current_admin.access_tree,
+              form: form %>
+        </div>
+        <%= form.hidden_field :selected_facilities, id: "facility-access-selections", value: [], multiple: true %>
```
kitallis, line 17: This thing is magically populated by JS to mimic the conventional Rails Way™ of dealing with collections. Since this is not a nested form, these facilities are populated by multiple hidden tags under `selected_facilities` through [this](https://github.com/simpledotorg/simple-server/pull/1222/files#diff-be260484baa204aaf798736297ef46acR19) function, which then gets read as a well-formed array in the controller when the form is submitted.

app/models/user_access.rb
```diff
@@ -59,6 +59,63 @@ def can?(action, model, record = nil)
     end
   end
 
+  def access_tree(action)
```
kitallis, line 62: This method looks like it's only for views (and it probably is), but it's better documented and tested and placed here because it's a critical piece that makes important DB queries and must be well tested (like any other `UserAccess` method).

app/models/user_access.rb
```diff
@@ -59,6 +59,63 @@ def can?(action, model, record = nil)
...
+        .map(&:facility_group)
+        .map { |fg|
+          facilities_in_facility_group =
+            facility_tree.select { |facility, _| facility.facility_group == fg }
+
+          [fg,
+            {
+              can_access: can?(action, :facility_group, fg),
+              facilities: facilities_in_facility_group,
+              total_facilities: fg.facilities.size
```
kitallis, line 81: We only expose totals, not the actual objects. Being overly defensive here since we're dealing with authorization.

app/controllers/email_authentications/invitations_controller.rb
```diff
@@ -8,22 +8,36 @@ def new
...
-        next if permission_params.blank?
-
-        permission_params.each do |attributes|
-          user.user_permissions.create!(attributes.permit(
-            :permission_slug,
-            :resource_id,
-            :resource_type
-          ))
+    if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin)
+      raise UserAccess::NotAuthorizedError unless current_admin.can?(:manage, :facility)
```
kitallis, line 12: This will eventually be replaced by the new authorize method from @vkrmis's [PR](https://github.com/simpledotorg/simple-server/pull/1208).

app/models/user_access.rb
```diff
@@ -79,4 +145,29 @@ def resources_for(resource_model, action)
...
+        selected_facilities -= selected_facilities_in_fg
+      end
+    end
+
+    selected_facilities.each do |f|
+      resources << {resource: f} if can?(:manage, :facility, f)
+    end
+
+    resources.flatten
+  end
```
harimohanraj89, line 172: This is where the magic happens. Love it.

app/controllers/email_authentications/invitations_controller.rb
harimohanraj89: Why are we checking if the user can manage _facilities_?

app/controllers/email_authentications/invitations_controller.rb
```diff
@@ -1,29 +1,51 @@
...
+      authorize([:manage, :admin, current_admin])
+    end
+
     super
   end
 
   def create
-    user = User.new(user_params)
-    authorize([:manage, :admin, user])
+    if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin)
```
harimohanraj89, line 17: Why do we key off the flag here? I thought that the `create` method will continue to do its thing, and we'll use Flipper to redirect darklaunched users to the `new_create` path instead?

harimohanraj89 (changes requested):
Initial round of questions from me!

app/controllers/email_authentications/invitations_controller.rb
kitallis: I removed the `new_create` method, it didn't buy us much. Instead it spills awkward devise code all over the routes files and paths in views. It'll be easier to remove this code later if it's all in the same method, so I just kept it here.

app/controllers/email_authentications/invitations_controller.rb
kitallis: Hm, I think that's the best approximation we have no? We only have the idea of `manage` and `view`, roughly. So, even if you can manage a single facility, that's enough for you to invite folks over.

app/models/user_access.rb
kitallis: I would like to re-write this a little more _functionally_, but it works pretty well for now, so I've kept it around.

app/controllers/email_authentications/invitations_controller.rb
harimohanraj89: Ohh that's right! We don't really have a concept of managing _users_ do we. Would a better approximation simply be - is the user a manager?

kitallis: Hm, then we'll have to ensure if they are `manager` or `power_user` or any other `manager-esque` permission. One could make a `manager?` function on user or something. But it felt like using a standard pattern for all pages is better: like `authorize`. I did consider a `manage_user` action, which is simply a collection of `access_levels` that are allowed to manage a user. But I thought to wait on that until I get on migrating the user/admin pages. That may fit best here.

app/views/email_authentications/invitations/new.html.erb
kitallis: Thanks. It really isn't, I did that because it was annoying to figure out how collection checkboxes worked. I've implemented your suggestion [here](https://github.com/simpledotorg/simple-server/pull/1222/commits/b12ec7964ea60a3dc8c33970b6d736c52521bdf0). Works well and shaves off some JS ❇️ .

app/controllers/email_authentications/invitations_controller.rb
harimohanraj89: Oh great point. We want to ask if you can perform the _action_ "manage", not if you are the role "manager". Yup, this looks fine until we introduce some concept of managing users 👍

app/assets/javascripts/invite_admin.js
```diff
@@ -0,0 +1,103 @@
...
+  // get checked state of siblings
+  // are every or some siblings checked (using Boolean as test function)
+  const checkStatus = siblings.map(check => check.checked)
+  const every = checkStatus.every(Boolean)
+  const some = checkStatus.some(Boolean)
+
+  // check parent if all siblings are checked
+  // set indeterminate if not all and not none are checked
+  parent.checked = every
+  parent.indeterminate = !every && every !== some
```
harimohanraj89, line 80: Can this be written as `some && !every`?

app/assets/javascripts/invite_admin.js
```diff
@@ -0,0 +1,103 @@
...
+
+  // recurse until check is the top most parent
+  if (element != parent) updateParentCheckedState(parent, selector)
```
harimohanraj89, line 83: `!==`?

harimohanraj89 (approved):
A couple last bits of feedback, but this is darklaunched and holds up other work. Let's push it out and tweak it in future PRs.

app/assets/javascripts/invite_admin.js
danySam: No, it does not rely on loose equality. I'll change this.

PR #1237, opened 2020-08-19
Reports: Tooltip improvements

claudiovallejo:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1018/language-and-tooltip-improvements

## Because

* Tooltips are hard to read
* Language is inconsistent

## This addresses
**Tooltips are easier to read**
* Updated tooltip styling
* Showing a single tooltip per chart data point

**Consistent labels**
* Using same labels for all chart legends
* Using same labels for all chart tooltips

**Screenshots**
image
image
image
image
image

### DIFF
--- a/.semaphore/bangladesh_demo_deployment.yml
+++ b/.semaphore/bangladesh_demo_deployment.yml
@@ -1,0 +1,25 @@
+version: v1.0
+name: Pipeline 3
+blocks:
+  - name: Deploy to Bangladesh Demo
+    task:
+      jobs:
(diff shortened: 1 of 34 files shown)


dburka:
I left some comments in Slack. Biggest one is that the tooltip values are in reverse order from the graph — so if green/controlled is at the bottom of a bar chart, it's at the top of the tooltip. This makes them hard to read especially in the grey "No recent BP" card.

dburka (approved):
Looks good. Good fixes.

PR #1239, opened 2020-08-21
Add teleconsultation phone number to User

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/990

## Because

We need an additional field on User to support different primary and telemedicine/Whatsapp phone numbers.

## This addresses

- Adds `teleconsultation_phone_number` to the User model
- Adds `teleconsultation_phone_number` to `/v4/users/activate` and `/v4/users/me` responses (Defaults teleconsultation_phone_number to primary phone number if not present)
- Adds field to Edit user form

Once this is merged to production, we can kick off https://app.clubhouse.io/simpledotorg/story/987/prepare-to-migrate-telemed-numbers-to-user


app/models/user.rb
```diff
@@ -134,7 +144,7 @@ def update_with_phone_number_authentication(params)
     )
 
     transaction do
-      update!(user_params) && phone_number_authentication.update!(phone_number_authentication_params)
+      update(user_params) && phone_number_authentication.update!(phone_number_authentication_params)
```
kitallis, line 147: Why this change?

prabhanshuguptagit: This page returned a 400 earlier if an update failed, we want to show the errors on the form now. 
image

app/controllers/admin/users_controller.rb
```diff
@@ -89,6 +89,7 @@ def user_params
     params.require(:user).permit(
       :full_name,
       :phone_number,
+      :teleconsultation_phone_number,
```
harimohanraj89, line 92: Should we model this as an isd code and number, like we do for the existing telemed numbers? I think we found that works, and gives us flexibility as we don't hard-couple the phone number to the country.

harimohanraj89 (commented):
One question about ISD codes.

app/controllers/admin/users_controller.rb
prabhanshuguptagit: Added in 9e00e02. Small nuance:
- Currently `teleconsultation_phone_numbers` are sent as `+91222222222` since whatsapp expects `912222222222`.
- User `phone_numbers` are sent as `22222222222` because phone number auths don't have a country code.

We will prepend the isd code before sending it in the `user` payload so that `teleconsultation_phone_number` will always be in the `+91` format. This does mean that the user payload will have `phone_number` and `teleconsultation_phone_number` in different formats. I'll run this by mobile devs, not expecting this to cause any issues. Just something to be aware of.

image

Ran the UI change by claudio ✅

harimohanraj89: Coolbeans. Yeah it's a little weird because the app doesn't care about the teleconsult number exposed in this payload, it only cares about the "MOs in a facility" endpoint, where we can consistently deliver phone numbers with ISD codes, like we do now.

Don't worry _too_ much about getting it perfect, I think there's gonna be some inconsistency no matter what we decide - either between these numbers and existing telemed numbers, or between these telemed numbers and user phone auths.

app/models/user.rb
```diff
@@ -101,6 +103,11 @@ def registration_facility_id
 
   alias facility registration_facility
 
+  def teleconsultation_phone_number_with_isd
```
harimohanraj89, line 106: Consider `full_teleconsultation_phone_number`

harimohanraj89 (approved):
✨ 🚀 🌔

app/transformers/api/v4/user_transformer.rb
```diff
@@ -4,7 +4,8 @@ def to_response(user)
       Api::V4::Transformer.to_response(user)
         .merge("registration_facility_id" => user.registration_facility.id,
                "phone_number" => user.phone_number,
-               "password_digest" => user.phone_number_authentication.password_digest)
+               "password_digest" => user.phone_number_authentication.password_digest,
+               "teleconsultation_phone_number" => user.teleconsultation_phone_number_with_isd)
```
harimohanraj89, line 8: Are the plain database fields also exposed in the API automatically? Do we need to explicitly exclude them?

prabhanshuguptagit: The `teleconsultation_isd_code` needs to be removed, oops. Added in 88220b5c.

app/models/user.rb
```diff
@@ -105,6 +105,7 @@ def registration_facility_id
 
   def full_teleconsultation_phone_number
     defaulted_teleconsult_number = teleconsultation_phone_number || phone_number
+    teleconsultation_isd_code ||= Rails.application.config.country["sms_country_code"]
```
harimohanraj89, line 108: I'd love to make the ISD code required, and prepopulate the form input with the country config. We can aggressively force our data to be high quality here, I think.

prabhanshuguptagit: This is only for users who don't have a teleconsultation number added. They will have the number and isd code both set to `null`. We'll make the code required in the form.

harimohanraj89: Ah great point 👍 Nice.

PR #1241, opened 2020-08-24
Warm dashboard reports caches every day

rsanheim:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1068/warm-up-all-dashboard-caches-on-a-regular-schedule

## Because

we want pages to be fast on first page load

## This addresses

preload the cache every day during off hours


app/services/no_bp_measure_service.rb
```diff
@@ -18,6 +18,9 @@ def call
   end
 
   def visited_without_bp_taken_for(period)
+    if facilities.empty?
+      return 0
+    end
```
prabhanshuguptagit, line 23: Inline this?

app/services/no_bp_measure_service.rb
```diff
@@ -18,6 +18,9 @@ def call
   end
 
   def visited_without_bp_taken_for(period)
```
prabhanshuguptagit, line 20: Could we call this method something like `visited_without_bp_taken_count` since it's only a count and not the actual patients?

spec/services/top_region_service_spec.rb
```diff
@@ -16,6 +16,34 @@ def refresh_views
...
 
+  it "has a call convienence method" do
+    darrang = FactoryBot.create(:facility_group, name: "Darrang", organization: organization)
+    darrang_facilities = FactoryBot.create_list(:facility, 2, facility_group: darrang)
+    kadapa = FactoryBot.create(:facility_group, name: "Kadapa", organization: organization)
+    _kadapa_facilities = FactoryBot.create_list(:facility, 2, facility_group: kadapa)
+    koriya = FactoryBot.create(:facility_group, name: "Koriya", organization: organization)
+    koriya_facilities = FactoryBot.create_list(:facility, 2, facility_group: koriya)
+
+    Timecop.freeze("April 1sth 2020") do
```
prabhanshuguptagit, line 27: 👾

app/models/period.rb
```diff
@@ -27,6 +27,15 @@ def self.quarter(value)
...
     @type = type.intern if type
+    @value = if @value.is_a?(String)
+      if quarter?
+        Quarter.parse(@value)
+      else
+        @value.to_date
+      end
+    else
+      @value
+    end
```
prabhanshuguptagit, line 38: Is there a way where we can avoid shadowing `@value`? Also the `if-else` could go in a separate method.

spec/services/top_region_service_spec.rb
```diff
@@ -16,6 +16,34 @@ def refresh_views
     end
   end
 
+  it "has a call convienence method" do
```
prabhanshuguptagit, line 19: [nit] This spec also seems to be validating some data, could you update the `it` description.

prabhanshuguptagit (commented):
The `Warmer` service looks good👌 . We should stress test this out on sandbox (or even the new `perf`, let me know if you want to pair on running it).

app/models/period.rb
rsanheim: Yup, I didn't really like this. I cleaned it up a bit to try to make it more clear.

app/services/no_bp_measure_service.rb
rsanheim: 👍

app/services/no_bp_measure_service.rb
rsanheim: 👍

PR #1242, opened 2020-08-25
Introduce a UserAccess model

kitallis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/128/support-for-adding-and-removing-permissions-to-an-access-level

## Because

The User class is lifting a lot of access-related weight, move it out to a more dedicated place.

I first pulled it out into a concern, but it didn't deserve one.

Then I pulled it out into a module/mixin, but mixins are a very poor abstraction that don't mean anything, esp not a great place to put concrete domain logic.

So I eventually pulled up everything from `User` and `Access` up into a concrete class that defines the `Access` behavior between `User` and `Access` and just use the Delegator pattern to forward methods from `User` to `UserAccess` for convenience.

This also makes the behavior easier to test and doesn't spill over the specs in 2 places.

## This addresses

* Add `UserAccess`
* Delegate methods from `User` to `UserAccess`
* Migrate `Access` and `User` permissions-related specs to `UserAccess`


app/models/user_access.rb
```diff
@@ -0,0 +1,82 @@
+class UserAccess < Struct.new(:user)
+  class NotAuthorizedError < StandardError; end
```
kitallis, line 2: This will come in handy later.

app/models/user_access.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def can_access_record?(resources, record)
+    return true if user.power_user?
+    return resources.find_by_id(record).present? if record
+    resources.exists?
+  end
+
+  def resources_for(resource_model, action)
+    return resource_model.all if user.power_user?
+    return resource_model.none unless ACTION_TO_LEVEL.fetch(action).include?(user.access_level.to_sym)
```
kitallis, line 72: This clause has been moved up, it used to be like [this](https://github.com/simpledotorg/simple-server/pull/1242/files#diff-dbee6fa62dba5eb76fea1e205e2a5f54L58) before. It makes more sense this way since we're just pulling accesses off a concrete `User`:

harimohanraj89 (approved):
Makes sense to me. Shipit :shipit:

PR #1243, opened 2020-08-25
HOTFIX: Cache the cohort_stats in progress tabs

kitallis:
## Because

The cohort report queries on the progress tabs are causing a cascading strain on the database.

## This addresses

This is a quick fix to cache them for 7 days until we figure out a better solution.


harimohanraj89 (approved):
:shipit: 🚀  🩹

PR #1244, opened 2020-08-25
Link telemed MO users to facilities

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/994

## This addresses

- Link facilities to users as MOs, as a has_and_belongs_to_many relationship
- Add `capabilities` to the `/v4/users/activate` and `/v4/users/me` responses


app/transformers/api/v4/user_transformer.rb
```diff
@@ -4,12 +4,19 @@ def to_response(user)
       Api::V4::Transformer.to_response(user)
         .merge("registration_facility_id" => user.registration_facility.id,
                "phone_number" => user.phone_number,
-               "password_digest" => user.phone_number_authentication.password_digest)
+               "password_digest" => user.phone_number_authentication.password_digest,
+               "capabilities" => app_capabilities(user))
```
harimohanraj89, line 8: How about `user.app_capabilities` instead? There's already tons of capability logic in the user model.

swagger/v4/swagger.json
```diff
@@ -1176,6 +1188,18 @@
...
+      "type": "object",
+      "properties": {
+        "can_teleconsult": {
+          "type": "string",
+          "enum": [
+            "yes",
+            "no"
+          ]
+        }
+      }
```
harimohanraj89, line 1202: Will we update the `FacilityTeleconsultationsController` after the data migration is complete? I guess that's the last step?

app/models/teleconsultation_medical_officer.rb
```diff
@@ -0,0 +1,2 @@
+class TeleconsultationMedicalOfficer < User
+end
```
harimohanraj89, line 2: Can we set up the associations without this subclass? It seems unnecessary 🤔

db/migrate/20200824112700_facilities_teleconsultation_medical_officers.rb
```diff
@@ -0,0 +1,10 @@
+class FacilitiesTeleconsultationMedicalOfficers < ActiveRecord::Migration[5.2]
+  def change
+    create_join_table :facilities,
+      :users,
+      column_options: {type: :uuid},
+      table_name: "facilities_teleconsultation_medical_officers" do |t|
+      t.index [:facility_id, :user_id], name: :index_facilities_teleconsult_mos_on_facility_id_and_user_id
+    end
+  end
+end
```
harimohanraj89, line 10: Is there any merit in setting up a first-class join table like "TeleconsultationMedicalOfficer" that joins facilities and users? Do we foresee the need for more teleconsultation logic in the near future that would live there?

harimohanraj89: PS Either way, we should set up a model to access this join table. Otherwise, it becomes awkward to ask questions like "Show me the list of teleconsult MOs".

harimohanraj89 (changes requested):

app/transformers/api/v4/user_transformer.rb
prabhanshuguptagit: I was slightly conflicted about this, since this is mostly transformation logic. (`true -> "yes", false -> "no"`). This is another variant I tried:

```
CAPABILITY_VALUES = {
 true => "yes",
 false => "no"
}.freeze

def app_capabilities(user)
  {can_teleconsult: User::CAPABILITY_VALUES[user.can_teleconsult?]}
end
```

prabhanshuguptagit: Actually this ^ + `user.app_capabilities` doesn't seem too bad.

swagger/v4/swagger.json
prabhanshuguptagit: The plan is to leave this controller untouched. Any nurses who wish to create new style teleconsults will have to update their apps anyways, and we expect this endpoint to stop being used after a while.

app/transformers/api/v4/user_transformer.rb
harimohanraj89: Can we simplify the data model of this capabilities blob? What about:

```
"user": {
  "capabilities": {
    "can_teleconsult": true
  }
}
```
or _even_ simpler
```
"user": {
  "capabilities": ["can_teleconsult"]
}
```

app/models/teleconsultation_medical_officer.rb
prabhanshuguptagit: I added this class for encapsulating some of the serialization-deserialization logic for the MO sync endpoint. This is somewhat premature but will come handy very soon.

swagger/v4/swagger.json
harimohanraj89: Oh, we do? I thought we'd replace the contents with MOs through the new data model? That way nurses don't _have_ to update right?

PS, how is the new version of the app planning to fetch telemed numbers from the server?

app/transformers/api/v4/user_transformer.rb
prabhanshuguptagit: 1. The mobile devs prefer `"yes/no"` over `true/false` because it sits well with the other APIs (medical history has yes/no answers)
2. We considered this before and then moved to this hash map version since it's a bit more flexible (if we want to support more than yes/no types of `capabilities`).

swagger/v4/swagger.json
prabhanshuguptagit: The new app will fetch the list from `teleconsultation_medical_officers/sync` and populate the bottom sheet with that data. We _could_ serialize the MOs in this old API but not sure if it's worth the effort. It's likely that all teleconsult nurses will be asked to update to be able to send that new deeplink message.

app/transformers/api/v4/user_transformer.rb
harimohanraj89: Gotcha. Then your suggestion above seems nice to me as well!

swagger/v4/swagger.json
harimohanraj89: Understood. Nice.

db/migrate/20200824112700_facilities_teleconsultation_medical_officers.rb
prabhanshuguptagit: I'd like to not call the join model `TeleconsultationMedicalOfficer` since it doesn't represent a real MO, it's just a relation underneath. wdyt about 21ec1bb? This has the advantage that `TeleconsultationMedicalOfficer.all` _are_ users. One doesn't have to do:
```
t = TeleconsultationMedicalOfficers.first
u = t.user // t should already be a user
```

We could add _a_ model to back the join table if we need it. The join table doesn't do anything interesting though currently and we could always add it later, I think.

harimohanraj89: I like this 👍  Let's do it!

harimohanraj89 (approved):
Looks great 🚀

PR #1245, opened 2020-08-25
Cleaner URLs for reports; remove the hacky 'region_slug'

rsanheim:
Never really liked that approach to dealing with the polymorphic
routes...this is a bit more explicit and the URLs we end up with are
much cleaner

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -160,11 +160,9 @@ Rails.application.routes.draw do
   get "/dashboard/districts/", to: redirect("/reports/districts/")
   get "/dashboard/districts/:slug", to: redirect("/reports/districts/%{slug}")
   get "/reports/districts/", to: redirect("/reports/regions/")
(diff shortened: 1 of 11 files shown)


config/routes.rb
```diff
@@ -160,15 +160,14 @@
   get "/dashboard/districts/", to: redirect("/reports/districts/")
   get "/dashboard/districts/:slug", to: redirect("/reports/districts/%{slug}")
   get "/reports/districts/", to: redirect("/reports/regions/")
-  get "/reports/districts/:slug", to: redirect("/reports/regions/facility_group-%{slug}")
+  # get "/reports/districts/:slug", to: redirect("/reports/regions/facility_group-%{slug}")
```
claudiovallejo, line 163: @rsanheim is this intentionally commented out?

rsanheim: oops, nope, ill remove it

PR #1246, opened 2020-08-26
Update banner email

claudiovallejo:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1087/update-old-reports-banner-copy

## Because

`support@simple.org` emails go to the #support channel on Slack

## This addresses

Replaces `support@simple.org` with `help@simple.org`.


harimohanraj89 (approved):
Looks like we're replacing `team@`, not `support@`, but still, sounds good.

Do we know where `help@` goes?

PR #1248, opened 2020-08-26
Production deployment preparation

harimohanraj89:
* Add production deployment pipelines
* Add auto-deploys to SBX, QA, and Demo environments for `release-*` tags

### DIFF
--- a/.semaphore/development_deployment.yml
+++ b/.semaphore/development_deployment.yml
@@ -46,5 +46,9 @@ agent:
 promotions:
   - name: Bangladesh Demo Deployment
     pipeline_file: bangladesh_demo_deployment.yml
+    auto_promote:
+      when: tag =~ '^release-.*' AND result = 'passed'
   - name: India Demo Deployment
     pipeline_file: india_demo_deployment.yml
+    auto_promote:
+      when: tag =~ '^release-.*' AND result = 'passed'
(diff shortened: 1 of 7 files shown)


.semaphore/development_deployment.yml
```diff
@@ -46,5 +46,9 @@ agent:
 promotions:
   - name: Bangladesh Demo Deployment
     pipeline_file: bangladesh_demo_deployment.yml
+    auto_promote:
+      when: tag =~ '^release-.*' AND result = 'passed'
```
kitallis, line 50: You should be able to use this [regex](https://github.com/simpledotorg/simple-server/blob/master/bin/deploy#L311) for the date.

harimohanraj89: Is this a good time to simplify that? Maintaining that seems tedious, but we can use our deployment script to start generating tags that look like `release-2020-08-25-01`

kitallis: Oh, that's right, it should be ok for now.

harimohanraj89: I've gone ahead and modified the tag created by `bin/deploy` in this PR itself, since that should probably be an atomic change.

I plan to open one more PR that will remove the deployment steps from the `bin/deploy` script, and also port the Slack notification to Semaphore.

kitallis (approved):
⏩

Issue #1255, opened 2020-08-27
DB perf tracking issue


rsanheim:
We had a DB related outage last night on August 27th.  This is a tracking issue to discuss remediation steps, fixes, and to track progress.

### References

* https://aws.amazon.com/blogs/database/scaling-your-amazon-rds-instance-vertically-and-horizontally/
* https://stackoverflow.com/questions/4832459/aws-rds-instance-upgrade-down-time
* https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html

### Checklist

* [x] back off on materialized view refresh
* [x] measure time to refresh caches on sandbox before any changes
* [x] look into what it takes to upgrade Postgres RDS instance size
* [x] upgrade to a ~`db.r5`~ `r4.xlarge` type database on sandbox -- these are the "latest generation memory optimized" instance types, and are supported for PostgreSQL.
  *Note that we couldn't use the r5 or higher because our Postgres version isn't recent enough*
* [x] verify
* [x] repeat for prod but use the `r4.x2large` instance
* [x] turn back on region cache warming (with flipper toggle) https://github.com/simpledotorg/simple-server/pull/1257
**NOTE:** I got blocked here because I needed to deploy a new ENV var to production, and that requires SSH access to ship `deployment`PRs
* [x] Merge and deploy latest `deployment` the new slack webhook ENV var to production https://github.com/simpledotorg/deployment/pull/246 (this is required for the next step!)
* [x] Deploy latest production `simple-server` so that we get https://github.com/simpledotorg/simple-server/pull/1258 deployed - kit/hari
* [x] Run the cache warmer manually on production to verify the DB is fine and that we get notifications in Slack #alerts - kit/hari:
```
RAILS_ENV=production bundle exec rails runner "Reports::RegionCacheWarmer.call"
``` 
* [ ] Verify production report page load times -- it should generally be 500 ms or lower for all the pages if they are all in the cache.  You can find them starting at `/reports/regions` - kit/hari - left some notes [here](https://github.com/simpledotorg/simple-server/issues/1255#issuecomment-682582124).
* [x] Update Terraform configuration to be in sync with our new database instance types https://github.com/simpledotorg/deployment/commit/ef29a2231bb2a0b7882c06246374b441dea979b7 - prabhanshu
[...]

PR #1256, opened 2020-08-27
Back off on materialized view refresh:

rsanheim:
Do it twice a week at 5 am India time.

This is to help us get the DB perf back in order.  We may not want to back it off this much long term, but for now twice a week is fine so we have some breathing room.

https://github.com/simpledotorg/simple-server/issues/1255


rsanheim:
Merging w/o approval cuz I need this live for db testing.

PR #1257, opened 2020-08-27
Add slack notifications to RegionCacheWarmer and turn it back on

rsanheim:
Also adding ability to disable via Flipper so we don't have to redeploy
if we wanna turn it off again.  Want to give this a shot now that we are on a larger database box.

/cc #1255

### DIFF
--- a/app/services/reports/region_cache_warmer.rb
+++ b/app/services/reports/region_cache_warmer.rb
@@ -10,7 +10,7 @@ module Reports
       @period = period
       @original_force_cache = RequestStore.store[:force_cache]
       RequestStore.store[:force_cache] = true
-      notifier.ping "#{self.class.name} Starting ..."
+      notify "#{self.class.name} Starting ..."
     end
 
     attr_reader :original_force_cache, :period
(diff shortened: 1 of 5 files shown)


app/services/reports/region_cache_warmer.rb
```diff
@@ -4,26 +4,72 @@ def self.call
       new.call
     end
 
+    def self.create_slack_notifier
+      return if ENV["SIMPLE_SERVER_DEPLOYMENT_NOTIFICATIONS_WEBHOOK"].blank?
```
rsanheim, line 8: We wanna use a new webhook but adding env vars to prod is painful right now, so reusing what we have for now.

config/schedule.rb
```diff
@@ -23,17 +23,15 @@ def local(time)
...
+every :day, at: local("02:00 am"), roles: [:cron] do
+  runner "Reports::RegionCacheWarmer.call"
+end
 
 every [:sunday, :wednesday], at: local("12:30am"), roles: [:cron] do
   rake "refresh_materialized_db_views"
 end
 
-every :month, at: local("02:00 am"), roles: [:seed_data] do
+every :month, at: local("04:00 am"), roles: [:seed_data] do
```
rsanheim, line 34: Don't want this to overlap w/ the cache warming on sandbox / etc.

PR #1260, opened 2020-08-28
Cache NoBPMeasure result

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1102/cache-nobpmeasureservice-results

## Because

While testing reports performance, we discovered that the NoBPMeasureService was not caching its results like the other services were. Judging from the cache-related methods already in the class, this is probably just an oversight.

## This addresses

Caches the result of the NoPBMeasureService.


rsanheim:
We may want to cache at the lower level of granularity here - i.e. one cache entry per period / region. That advantage of that is that our cache hit rate will be higher when we have different callers of this service who are asking for different ranges. I'll throw up a PR for that approach today.

rsanheim:
>  I'll throw up a PR for that approach today.

Oh, right, I did it as part of the Facility details PR [here](https://github.com/simpledotorg/simple-server/pull/1189/files#r480391659).

harimohanraj89:
Oh, awesome. Ok, closing this.

PR #1268, opened 2020-08-31
Teleconsult MO sync endpoint

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/999

## Because

We need a `facility_medical_officers` endpoint for the app to pull the list of available medical officers.

## This addresses

Adds a `GET /v4/facility_medical_officers/sync` API.

### Notes

This resource is slightly different from regular sync resources:
- It is pull-only, i.e, the app will never update and push these records to the server. There is no `POST sync` action.
- Since the resource isn't a real thing in the DB (it's a relation), there isn't a natural cursor that can be used to create a process token and chunk these records up. We could manufacture one using the users and facility's `updated_at`. We have however chosen the simpler approach of not sending a process token and sending all the records in a single response. This is fine because the number of records is going to be small (one record per facility, for all facilities in a facility group).


app/transformers/api/v4/teleconsultation_medical_officer_transformer.rb
```diff
@@ -0,0 +1,9 @@
+class Api::V4::TeleconsultationMedicalOfficerTransformer
+  class << self
+    def to_response(medical_officer)
+      medical_officer
+        .slice("id", "full_name", "teleconsultation_phone_number")
```
kitallis, line 5: Why are we slicing these things from here? Maybe drop a comment here?

app/controllers/api/v4/teleconsultation_medical_officers_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  private
+
+  def facility_medical_officers(facility)
+    {id: facility.id,
+     facility_id: facility.id,
+     medical_officers: transform_medical_officers(facility.teleconsultation_medical_officers),
+     created_at: Time.current,
+     updated_at: Time.current,
+     deleted_at: Time.current}
```
harimohanraj89, line 17: It seems a bit inconsistent that this endpoint would return a nested array of objects, each containing an array of MOs. Instead, I think we should return a flat array of medical officer objects, each of which contains a `facility_id` in its attributes. What do you think?

harimohanraj89 (changes requested):
One big structural suggestion. The rest is fairly straightforward :)

app/controllers/api/v4/teleconsultation_medical_officers_controller.rb
prabhanshuguptagit: There's two reasons to make this nested:
- MOs are not implemented as a "real resource", both on the server and the app. This resource represents only a relation between the facility and users. Both a flat resource and a nested resource would need the same effort to serialize/deserialize.
- The payload size becomes much smaller when we nest.

Another additional (later discovered) advantage of this is that we can send all these facility records in one response. The number of flattened records will be much larger and we will have to chunk it, which is slightly tricky. Nesting solves this problem.

prabhanshuguptagit: > MOs are not implemented as a "real resource"

Actually, this may be incorrect. Given that we are sending the name and phone number along, the app _might_ be constructing a MO model internally. There is no need for it to be associating these to `users`. Regardless, nesting still has those advantages.

This data should ideally be in the facility's payload, yes (conditionally added after login). This is a separate endpoint only because it's more ergonomic for the app this way.

Calling it `teleconsultation_facilities` makes sense, although `teleconsultation_medical_officers` appears more illustrative of its purpose (this endpoint is primarily to get MO data on the app). wdyt?

harimohanraj89: Haha yeah tricky. I think I'd slightly lean towards calling it teleconsultation facilities. Reason being: Even though we do primarily want MO info, I'd place a little more weight on ensuring the top-level objects returned by the API match its name. For example, I wouldn't expect a `facilities` endpoint to return facility _groups_, with facility info inside.

kitallis (changes requested):

app/controllers/api/v4/teleconsultation_medical_officers_controller.rb
vkrmis: Could also call it `FacilityMedicalOfficers`

prabhanshuguptagit: Changed it to `/v4/facility_medical_officers/sync` in d7409f6.

harimohanraj89 (approved):
Looks great to me now.

kitallis (approved):
すごい!!

PR #1269, opened 2020-08-31
Performance improvments for ControlRateService / cache times

rsanheim:
Small fix so we know which prod environments these alerts are coming from.


app/services/control_rate_service.rb
```diff
@@ -75,22 +75,16 @@ def controlled_patients(period)
   end
 
   def bp_monthly_query(period)
-    date = period.to_date
-    end_range = date.end_of_month
-    mid_range = date.advance(months: -1).end_of_month
-    beg_range = date.advance(months: -2).end_of_month
+    control_range = period.blood_pressure_control_range
```
harimohanraj89, line 78: This control range starts and ends precisely at midnight of the local time zone, right?

harimohanraj89 (approved):
One time-zone comment. Pre-emptive approval to unblock you 🚀

app/services/control_rate_service.rb
rsanheim: Well, its using dates, so it beings on the day of the previous control range, and ends on the last day of the range. So, actually, the beginning piece should *not* be `>=`, it should be `>`.  I'll fix that.

```ruby
  it "can return its blood pressure control range" do
    range = jan_1_2020_month_period.blood_pressure_control_range
    expect(range.begin).to eq(Date.parse("October 31st 2019"))
    expect(range.end).to eq(Date.parse("January 31st 2020"))

    range = Period.month("July 1st 2020").blood_pressure_control_range
    expect(range.begin).to eq(Date.parse("April 30th 2020"))
    expect(range.end).to eq(Date.parse("July 31st 2020"))
  end
```

PR #1271, opened 2020-09-01
Deployment Refinement

harimohanraj89:
In light of our Semaphore deployment pipeline, the only thing we need to do now to make deployments is to cut a release tag.

In order to start modifying our deployment tooling for this flow, this PR introduces a `bin/release` script. It's a simpler version of the deployment script that does the following:

* Print a changelog based on the last commit
* Cut a release tag (ask for help if there's a conflict)
* Push the release tag and point to our semaphore page

It does not
* Automatically publish the changelog to Slack. Let's do that manually for a little bit until we can get Semaphore to do it for us.

Todo:
* [x] Add an extra "Are you sure" if the current branch is not master.

### DIFF
--- a/app/assets/javascripts/common/reports.js
+++ b/app/assets/javascripts/common/reports.js
@@ -1,15 +1,42 @@
-window.addEventListener("DOMContentLoaded", initializeCharts);
+window.addEventListener("DOMContentLoaded", function() {
+  initializeTables();
+  initializeCharts();
+});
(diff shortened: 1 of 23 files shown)


rsanheim:
> [x] Add an extra "Are you sure" if the current branch is not master.

I added this because I want to use this script to changelog the release I did earlier today. 😄

PR #1277, opened 2020-09-02
Index "twilio_sms_delivery_details.session_id"

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1138/add-an-index-to-twilio-sms-delivery-details-table

## Because

We have DB and request strain when we fire off reminders and get callbacks from Twilio. We spend most of our time finding the right `TwilioSmsDeliveryDetail` record by session ID.

## This addresses

This PR indexes that `session_id` column for quick finds.

Note: Do we need to run this thing concurrently? Can someone help me verify against a prod-like dataset? ie, how long will this migration take to run in IHCI production?


db/schema.rb
```diff
@@ -512,6 +513,7 @@
     t.index ["access_level"], name: "index_users_on_access_level"
     t.index ["deleted_at"], name: "index_users_on_deleted_at"
     t.index ["organization_id"], name: "index_users_on_organization_id"
+    t.index ["teleconsultation_phone_number"], name: "index_users_on_teleconsultation_phone_number"
```
harimohanraj89, line 516: @prabhanshuguptagit looks like this index was missed from the schema when the migration was added. Perhaps the migration was run on a local db and then modified to add an index? In any case, I want to confirm that this index _should_ be here.

prabhanshuguptagit: I think I added the index later and missed committing the schema, yes. This is correct.

db/schema.rb
```diff
@@ -757,6 +881,7 @@
           END AS current_age,
       p.gender,
       p.status,
+      p.assigned_facility_id,
```
harimohanraj89, line 884: @prabhanshuguptagit I wanted to check this one too. Did this get missed, is it supposed to be here? I dropped and recreated my db locally before running these migrations.

kitallis (approved):

db/schema.rb
prabhanshuguptagit: That... shouldn't be. This will only be added in #1169, not sure why it shows up here. Also I did a drop and migrate on master, I don't see this in my schema.

db/schema.rb
```diff
@@ -658,6 +660,101 @@
      FROM (appointments ap
        JOIN facilities f ON ((ap.facility_id = f.id)));
   SQL
+  create_view "latest_blood_pressures_per_patient_per_days", materialized: true, sql_definition: <<-SQL
```
prabhanshuguptagit, line 663: Can we skip committing these schema reordering chunks?

harimohanraj89: Yup I'll clean up the schema 👍

db/schema.rb
harimohanraj89: Actually, looks like it was added in https://github.com/simpledotorg/simple-server/pull/1158, so it _should_ be here 👍

harimohanraj89: Yup, looks like I had a dirty schema somewhere during this process. I've reset it now, thanks 👍

prabhanshuguptagit (approved):

PR #1278, opened 2020-09-02
Update deployment docs

harimohanraj89:
Update deployment documentation to use the new `bin/release` script.


kitallis (approved):
LGTM.

I suppose we can add the slack notifier from semaphore a bit later, right? If I have the time I can take a shot at it. It's literally just a `curl` request I imagine.

harimohanraj89:
Yup. The only tricky bit will be grabbing the right changelog at that time. We can stick the Slack webhook into our project secrets easily, no biggie.

PR #1282, opened 2020-09-02
Migrate permissions for controllers

vkrmis:
**Story card(s):** 
- [x] Migrate Admin::OrganizationsController
- [x] Migrate Admin::ProtocolsController
- [x] Migrate Admin::ProtocolDrugsController

## Because

We conditionally use the new permissions for some controllers (using a Flipper flag). 
We need to migrate the rest of our app to start using the new permissions

## This addresses

This PR migrates the controllers listed above to use the new migrations. ~These migrations are not gated by the Flipper flag, and **should only be merged after a data migration strategy is in place to transition existing users to the new permissions**~

These migrations have been gated, safe to merge now.


app/controllers/admin/protocol_drugs_controller.rb
```diff
@@ -2,25 +2,44 @@ class Admin::ProtocolDrugsController < AdminController
...
 
+  skip_after_action :verify_authorized, if: -> { Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) }
+  skip_after_action :verify_policy_scoped, if: -> { Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) }
+  after_action :verify_authorization_attempted, if: -> { Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) }
+
   def index
-    authorize([:manage, ProtocolDrug])
-    @protocol_drugs = policy_scope([:manage, ProtocolDrug])
+    if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin)
+      authorize1 { current_admin.power_user? }
```
harimohanraj89, line 11: This is a good first guess, let's figure out who should be allowed to manage protocol drugs.

Praveen will not be a power user, right? Nilenso will have to manage protocol drugs if we ship this?

app/models/user_access.rb
```diff
@@ -83,6 +83,18 @@ def accessible_users(action)
       .where(phone_number_authentications: {registration_facility_id: accessible_facilities(action)})
   end
 
+  def accessible_protocols(_action)
```
harimohanraj89, line 86: Hmm, now that we have other models that utilize `accessible_*` methods, should we rename the `resources_for` method to `regions_for`?

app/views/admin/organizations/index.html.erb
```diff
@@ -13,10 +19,14 @@
     <div class="row">
       <div class="col-10">
         <h5 class="mb-1">
-          <% if policy([:manage, organization]).edit? %>
+          <% if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) %>
```
harimohanraj89, line 22: There might be a bug here. Doesn't this bit mean once we enable the flipper flag, _everyone_ will see a link?

app/views/admin/organizations/index.html.erb
```diff
@@ -32,10 +42,16 @@
         </div>
       </div>
       <div class="col-2 text-right">
-        <% if policy([:manage, organization]).destroy? %>
+        <% if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) %>
```
harimohanraj89, line 45: There might be a bug here. Doesn't this bit mean once we enable the flipper flag, everyone will see a link?

app/views/admin/protocol_drugs/index.html.erb
```diff
@@ -18,11 +18,16 @@
         <td><%= protocol_drug.dosage %></td>
         <td><%= protocol_drug.rxnorm_code %></td>
         <td><%= protocol_drug.protocol %></td>
-        <% if policy([:manage, protocol_drug]).edit? %>
+        <% if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) %>
```
harimohanraj89, line 21: There might be a bug here. Doesn't this bit mean once we enable the flipper flag, everyone will see links?

app/views/admin/protocol_drugs/show.html.erb
```diff
@@ -18,6 +18,10 @@
   <%= @protocol_drug.protocol %>
 </p>
 
-<% if policy([:manage, @protocol_drug]).edit? %>
+<% if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) %>
```
harimohanraj89, line 21: There might be a bug here. Doesn't this bit mean once we enable the flipper flag, everyone will see a link?

spec/controllers/reports/regions_controller_spec.rb
```diff
@@ -71,7 +71,7 @@ def refresh_views
     end
   end
 
-  context "show" do
+  pending "show" do
```
harimohanraj89, line 74: Are we going to unpending these specs before shipping?

harimohanraj89 (changes requested):
Feedback below. Some overall themes:

* I think several of our Flipper view checks aren't correct. They seem to show links to everyone if the flag's enabled, instead of just power users.
* I see several pend-ed specs. We're going to unpend them before merging, right?

app/views/admin/organizations/index.html.erb
vkrmis: These checks are actually unnecessary, the organizations are already scoped by the controller, and the controller actions have authorization checks. Not sure why we put them in in the first place tbh.

app/views/admin/organizations/index.html.erb
vkrmis: _ditto_

app/views/admin/protocol_drugs/index.html.erb
vkrmis: _ditto_

app/views/admin/protocol_drugs/show.html.erb
vkrmis: _ditto_

app/views/admin/organizations/index.html.erb
harimohanraj89: Oh I see. If you can get to this page in the first place, then you can see everything. Cool, that works for me 👍

spec/controllers/reports/regions_controller_spec.rb
vkrmis: Yup, forgot to do this.

app/controllers/admin/protocol_drugs_controller.rb
vkrmis: Yup, would limiting this to organization managers work? Protocols are basically independent.

harimohanraj89: Yeah it's kinda weird, eh? I really like the idea of `authorize1 { current_admin.accessible_organizations(:manage).any? }`. If you think that makes sense, let's do that right away 👌 

If we're not sure, let's leave as is and revisit later. In the interim, we'll manage protocols 😅

spec/features/home_spec.rb
```diff
@@ -14,16 +14,21 @@
   let!(:path_clinic) { create(:facility, facility_group: path_group, name: "Dr. Amir Singh") }
 
   before do
+    enable_flag(:new_permissions_system_aug_2020, admin)
     sign_in(admin.email_authentication)
-    visit root_path
+    # Root path has moved to MyFacilities#Overview
+    # visit root_path
+    visit organizations_path
```
harimohanraj89, line 21: This isn't a home_spec anymore 🤔  Is it valuable? Should we just axe it?

harimohanraj89:
Last couple of comments from me. This is lookin' solid.

app/controllers/admin/protocol_drugs_controller.rb
vkrmis: I've done this, and reused the idea in the `accessible_protocols` and `accessible_protocol_drugs` implementations.

app/models/user_access.rb
vkrmis: Not fully convinced we should do this, would mean we should also change `access.resource` to `access.region`. Lets leave it for later? Since `resources_for` is a private method, we can change this easily.

spec/features/home_spec.rb
vkrmis: Lets leave it in for now. It tests some access things, general navigation, scoping for the user approvals partial etc. We should eventually migrate it to reflect the new home page (my facilities).

app/models/user_access.rb
harimohanraj89: Ah I see. Yup sounds good to me 👍

app/models/user_access.rb
```diff
@@ -97,6 +97,24 @@ def accessible_users(action)
       .where(phone_number_authentications: {registration_facility_id: accessible_facilities(action)})
   end
 
+  def accessible_protocols(action)
+    return Protocol.none unless [:manage].include?(action)
```
harimohanraj89, line 101: Why return none if you ask for protocols you can view? Can we simplify this to:

```
return Protocol.all if power_user?
return Protocol.all if accessible_organizations(:manage).any?

Protocol.none
```

harimohanraj89 (approved):
One last question about the `accessible_protocol_*` methods, but this looks great 👌

app/models/user_access.rb
vkrmis: Hmm. Can do this, however, there isn't a simple `view` action. The other supported actions (`view_reports`, `view_pii`, and `manage_overdue_list`) don't really make sense in the context of protocols or protocol drugs.

We could add a `view` action, but we don't currently have a case for `view`ing protocols or drugs you can't manage.

harimohanraj89: I'm sorry, I shouldn't have use the word "view". I understand that protocol-related models are "all or nothing". Either you can manage all of them, or you can't do anything. My question was more scoped to the logic of this method. Can we simplify? ☝️

harimohanraj89: Speaking of viewing protocols, one touchpoint for viewing protocols is selecting one for your facility group, right? Are we planning to expose `ProtocolDrug.all` in the selection dropdown? I think that's ok, as long as we're making that conscious decision.

vkrmis: Ah, I misread the question. wdyt of:
```
return Protocol.all if power_user?
return Protocol.all if 𝗮𝗰𝘁𝗶𝗼𝗻 == :𝗺𝗮𝗻𝗮𝗴𝗲 && accessible_organizations(:manage).any? 
Protocol.none
```
 Mostly so we prevent usages like `accessible_protocols(:view_pii)` which don't make sense.

vkrmis: > Speaking of viewing protocols, one touchpoint for viewing protocols is selecting one for your facility group, right? Are we planning to expose ProtocolDrug.all in the selection dropdown? I think that's ok, as long as we're making that conscious decision.

Yup, this is the existing behaviour

harimohanraj89: Yup that looks perfect 👍

PR #1286, opened 2020-09-03
Miscellaneous permissions fixes

vkrmis:
**Story card:** none

## This addresses

- [x] Fix `OrganizationsController#index` regressions
- [x] Fix `Admin::UsersController#index` regressions

### DIFF
--- a/app/models/user_access.rb
+++ b/app/models/user_access.rb
@@ -72,7 +72,10 @@ class UserAccess < Struct.new(:user)
     User.admins.where(organization: user.organization)
   end
 
-  memoize def accessible_users
+  memoize def accessible_users(action)
+    return User.non_admins if bypass?
+    return User.none if action_to_level(action).include?(:manage)
+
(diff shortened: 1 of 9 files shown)


kitallis (approved):

app/models/user_access.rb
```diff
@@ -72,7 +72,10 @@ class AuthorizationNotPerformedError < StandardError; end
     User.admins.where(organization: user.organization)
   end
 
-  memoize def accessible_users
+  memoize def accessible_users(action)
+    return User.non_admins if bypass?
+    return User.none if action_to_level(action).include?(:manage)
```
harimohanraj89, line 77: Do we need to update test coverage for this change? Are we missing test coverage here?

vkrmis: We are definitely missing coverage for multiple methods in UserAccess, we're adding that separately today.

PR #1287, opened 2020-09-03
Gate Admin::FacilitiesController correctly

vkrmis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1154/couldn-t-create-a-facility-in-a-facilitygroup-because-we-accidentally-un-gated-the-the-controller-and-made-them-use-the-new

## Because

A permissions bug is causing new permissions methods to be invoked for users who haven't been migrated.

## This addresses

Gating the Admin::FacilityController actions correctly so the new permissions aren't invoked accidentally.

### DIFF
--- a/app/controllers/admin/facilities_controller.rb
+++ b/app/controllers/admin/facilities_controller.rb
@@ -3,8 +3,12 @@ class Admin::FacilitiesController < AdminController
   include Pagination
   include SearchHelper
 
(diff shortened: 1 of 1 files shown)


harimohanraj89 (approved):
Ah I see, the facility group fetching wasn't gated correctly.

Let's take this for a spin real quick for gated and non-gated users and :shipit:

kitallis (approved):
❇️

PR #1288, opened 2020-09-03
Telemed: Add MOs to facility UI 

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1003

## Because

A facility can have multiple telemed MOs. The facility edit page should allow searching users by name and adding them to the facility.

## This addresses

This adds a typeahead search to add MOs to a facility. Adds a `admin/users.json` endpoint for querying search.

This also removes the old `teleconsultation_phone_number` and `teleconsultation_isd_code` fields in the edit form and from the bulk upload CSV.

Figma: https://www.figma.com/file/9inSvRD4dXcMfXchlyDZVV/Dashboard-(Production)
In action: https://imgur.com/a/KLqAjdz


app/models/facility.rb
```diff
@@ -18,7 +18,8 @@ class Facility < ApplicationRecord
   has_many :phone_number_authentications, foreign_key: "registration_facility_id"
   has_many :users, through: :phone_number_authentications
   has_and_belongs_to_many :teleconsultation_medical_officers,
-    class_name: "TeleconsultationMedicalOfficer",
+    -> { distinct },
+    class_name: "User",
```
prabhanshuguptagit, line 22: Turns out doing `has_many` with `class_name: TeleconsultationMedicalOfficer` does not work out for us as expected. The `default_scope` on `TeleconsultationMedicalOfficer` means there is no way to check if a user is already added as an MO to a facility, since the results are already scoped. This was also somewhat incorrect, semantically. 

Still keeping the `TeleconsultationMedicalOfficer` class around for other semantic goodness.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -52,6 +52,27 @@ def login_user
...
+        search_query = "Doctor"
+
+        get :index, format: :json, params: {search_query: search_query}
+
+        expect(response).to be_successful
+        expect(JSON(response.body).first["full_name"]).to eq "Doctor Jack"
+      end
+
+      specify do
+        expect(User).to receive(:search_by_name_or_teleconsultation_phone_number).with(search_query)
```
harimohanraj89, line 71: Should we `allow` this to receive this and return a specific user? We can then assert that the user that's found is sent in the response correctly.

app/models/facility.rb
```diff
@@ -140,9 +140,7 @@ def dashboard_analytics(period: :month, prev_periods: 3, include_current_period:
      longitude: "longitude (optional)",
      facility_size: "size (optional)",
      enable_diabetes_management: "enable_diabetes_management (true/false)",
-     enable_teleconsultation: "enable_teleconsultation (true/false)",
-     teleconsultation_phone_number: "teleconsultation_phone_number",
```
harimohanraj89, line 144: Is it ok to revoke support for this? Do we need to let anybody know?

Should we continue to accept phone numbers and gracefully associate the right users if they exist?

harimohanraj89 (changes requested):
My thoughts below!

app/views/shared/_typeahead_search.html.erb
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  let debouncedSearch = debounce(search);
+
+  function showSpinner() {
+    let spinner = $(".typeahead-spinner").first().clone();
+    spinner.css({display: "block"});
+
+    populateDropdown(spinner);
+  }
```
harimohanraj89, line 81: We should put this JS in the JS pipeline as well.

prabhanshuguptagit: This JS needs to be in an erb file since there is some fancy rails templating going on. 
```
  function resultsJSONToHTML(results) {
    return results.map(function (result) {
      return <%= search_result_to_row %>(result);
    })
  }
```

This can be moved into separate `typeahead.js.erb`?

app/models/facility.rb
prabhanshuguptagit: These fields in the CSV are for the old old teleconsultation numbers, when we had one number per facility. We could let CVHOs submit one user's numbers in the template 🤷 , although it might be easier for them to upload the CSV and then add them all from the UI.

We will continue sending these fields in the app, so the API contract doesn't change.

app/models/user.rb
```diff
@@ -57,12 +57,19 @@ def app_capabilities
   has_many :accesses, dependent: :destroy
 
   pg_search_scope :search_by_name, against: [:full_name], using: {tsearch: {prefix: true, any_word: true}}
+  pg_search_scope :search_by_teleconsultation_phone_number,
+    against: [:teleconsultation_phone_number],
+    using: {tsearch: {prefix: true, any_word: true}}
```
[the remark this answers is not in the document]
prabhanshuguptagit: This lets users do partial word searches. I think we shouldn't support this. 👍

app/models/facility.rb
harimohanraj89: Hmm yeah tricky. I'd run this one by Praveen or Mahima to learn how uploads + telemed is being used, and at least let folks know that this is coming down the pike.

app/views/shared/_typeahead_search.html.erb
harimohanraj89: Oh I see. Yeah that seems quite convoluted indeed 🤔  Could we leverage another template and handle the JSON results client-side, like we do for the already-added-user cards? This stuff needing to be in here is a good red flag that we may be able to clean up the implementation a bit 😅

app/controllers/admin/users_controller.rb
```diff
@@ -39,12 +39,20 @@ def index
...
+        @users =
+          if searching?
+            paginate(users.search_by_name_or_phone(search_query))
+          else
+            paginate(users)
+          end
+      end
+
+      format.json do
+        @users = users.search_by_name_or_teleconsultation_phone_number(search_query)
```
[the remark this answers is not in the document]
prabhanshuguptagit: Agree, I was avoiding a new one since there was a lot of churn on this page going on with the permissions stuff in parallel. I _think_ that has settled down now, added the controller in e1373e0. This uses the old system, I'll flag it to @vkrmis to take a look.

app/assets/stylesheets/partials/_spacing.scss
```diff
@@ -0,0 +1,41 @@
...
+.p-0px { padding: 0px; }
+.pt-0px { padding-top: 0px; }
+.pt-4px { padding-top: 4px; }
+.pt-20px { padding-top: 20px; }
+.pt-24px { padding-top: 24px; }
+.pb-4px { padding-bottom: 4px; }
+.pb-8px { padding-bottom: 8px; }
+.pl-16px { padding-left: 16px; }
+.px-1px { padding-right: 1px; padding-left: 1px; }
+.px-20px { padding-right: 20px; padding-left: 20px; }
```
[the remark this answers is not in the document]
prabhanshuguptagit: I added some spacing classes that were missing in `_reports.scss` and then moved them all into a `_spacing.scss`. I just didn't want to put mine in `_reports.scss`, hence.

app/views/shared/_typeahead_search.html.erb
prabhanshuguptagit: We are handling the JSON all on client side, this bit is just to support the typeahead as a reusable component. This allows us to pass in JS functions to the partial that will do the JSON to HTML business:
```
 <%= render 'shared/typeahead_search',
                       placeholder: "Find medical officers by name or phone number",
                       url: "/admin/users.json",
                       results_transformer: "filterExistingUsers",
                       search_result_to_row: "userJSONToRow"
...
```
This gives the user a lot of flexibility to structure/style the typeahead as they wish.

harimohanraj89: Oh I see, hmm. I'd be tempted to hold off on the abstraction until we have a second, maybe even third, use case for typeahead. Keep it simple now, avoid guessing what the abstraction would look like without a concrete use case. What do you think?

app/assets/javascripts/common/facility_teleconsultation_fields.js
```diff
@@ -0,0 +1,96 @@
+function toggleTeleconsultationFields() {
```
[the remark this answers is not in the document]
prabhanshuguptagit: ➕

app/views/shared/_typeahead_search.html.erb
prabhanshuguptagit: Yeah this seems like overkill, I'll revert back to the plebian version.

app/assets/javascripts/common/facility_teleconsultation_fields.js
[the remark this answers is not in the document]
prabhanshuguptagit: Refactored in 4d4f714.

app/controllers/admin/users_controller.rb
```diff
@@ -39,19 +39,30 @@ def index
...
+        paginate(users.search_by_name_or_phone(search_query))
+      else
+        paginate(users)
       end
+  end
 
-      format.json do
-        @users = users.search_by_name_or_teleconsultation_phone_number(search_query)
-      end
+  def search
```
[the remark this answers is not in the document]
prabhanshuguptagit: This endpoint is only being called from the facility edit page, we are still searching for users. I think this namespace is correct, does `teleconsult_search` sound better?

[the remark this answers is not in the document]
harimohanraj89: `teleconsult_search` sounds 👌  💯

app/assets/javascripts/standalone/facility_teleconsult_fields.js
```diff
@@ -0,0 +1,74 @@
...
+    $("body").on("click", ".mo-search .typeahead-dropdown-row", function () {
+      let user = $(this).data();
+      teleconsultationFields.addMedicalOfficer(user);
+      teleconsultationFields.userSearch.emptyTypeahead();
+    });
+
+    $("body").on("click", ".remove-medical-officer", function () {
+      let userID = $(this).attr('data-user-id');
+      teleconsultationFields.removeMedicalOfficer(userID);
+    });
```
harimohanraj89: The classes look great. Could we place these JS ditties back in `common` so they're part of a unified JS bundle for the app? If we can introduce a convenient interface to this class, we can then move the _initialization_ (which we don't want to do on every page) to a tiny snippet of inline JS on the page itself. For example:

```html
blah blah erb markup blah blah

<script>
  facilityTeleconsultFields = new FacilityTeleconsultFields();
  facilityTeleconsultFields.listen();
</script>
```

```js
  // blah blah teleconsultfieldsstuff
  this.listen() {
        let teleconsultationFields = this;

    // Populate MOs from search results
    $("body").on("click", ".mo-search .typeahead-dropdown-row", function () {
      let user = $(this).data();
      teleconsultationFields.addMedicalOfficer(user);
      teleconsultationFields.userSearch.emptyTypeahead();
    });

    $("body").on("click", ".remove-medical-officer", function () {
      let userID = $(this).attr('data-user-id');
      teleconsultationFields.removeMedicalOfficer(userID);
    });
  }
```

See #1297 for inspiration 😄

app/assets/javascripts/standalone/user_teleconsult_search.js
```diff
@@ -0,0 +1,104 @@
...
+
+      clearTimeout(timeout);
+      timeout = setTimeout(later, wait);
+    };
+  }
+
+  this.debouncedSearch = this.debounce(this.search);
+}
+
+let debouncedUserSearch = new UserTeleconsultSearch().debouncedSearch;
```
harimohanraj89: Same feedback here. I'd love to move this "class" back to the common JS bundle by tactically moving the initialization to an inline snippet of JS.

harimohanraj89 (approved):
Looks awesome. Last few bits of feedback below. 🚀

app/assets/javascripts/standalone/facility_teleconsult_fields.js
prabhanshuguptagit: This was done in 6dfa50e.

harimohanraj89: Lol, did I review an outdated diff 🤦 

Neato. I still like the idea of wrapping the event listener registration in a method, rather than letting them implicitly run on initialization. We could even run them explicitly on registration inside a constructor method. Shall we do that?

prabhanshuguptagit: Sounds good 💯

PR #1289, opened 2020-09-03
Bugfix: Edit User autocomplete

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1156

## Because

Browsers are trying to incorrectly autocomplete the PIN field in the Edit User form. [Bug reported on Slack](https://simpledotorg.slack.com/archives/CFHKJ5WJY/p1599124237058200)

## This addresses

Disables autocomplete in the Edit User form.

### DIFF
--- a/app/views/admin/users/_form.html.erb
+++ b/app/views/admin/users/_form.html.erb
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="col-md-6">
-    <%= bootstrap_form_with(model: [:admin, user], local: true, label_errors: true) do |form| %>
+    <%= bootstrap_form_with(model: [:admin, user], local: true, label_errors: true, autocomplete: "off") do |form| %>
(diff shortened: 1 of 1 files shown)


harimohanraj89:
Does this change disable some custom autocompletion, or does it disable browsers from trying to pre-populate passwords? 🤔 PS this disables autocomplete for the whole form, right? Not just PIN? Looks like the teleconsult number _and_ PIN were being autocompleted with Praveen's username and password?

prabhanshuguptagit:
> Does this change disable some custom autocompletion, or does it disable browsers from trying to pre-populate passwords?
The latter. It's this attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

this disables autocomplete for the whole form, right?
For the whole form, yes, sorry missed this in the description.

kitallis (approved):

PR #1293, opened 2020-09-04
Bugfix redux: Disable autocomplete on Edit User

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1156

## Because

The Edit user form is being autocompleted. This was tried fixing in #1289 but that didn't work as expected.

## This addresses

Turns out `autocomplete="off"` on the form does not work as intended as of 2020: https://stackoverflow.com/a/57810447. 

This sets `autocomplete="new-password"` on the password field and `autocomplete="off"` on the field before as per [MDN's latest guidelines](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion). This disables autofill by default. The autocomplete suggestions still show up. It seems like there is no way to hide those in 2020, at least not if your field is called `email`, `phone_number` or `password`.

image

### DIFF
--- a/app/views/admin/users/_form.html.erb
+++ b/app/views/admin/users/_form.html.erb
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="col-md-6">
-    <%= bootstrap_form_with(model: [:admin, user], local: true, label_errors: true, autocomplete: "off") do |form| %>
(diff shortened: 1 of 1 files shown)


harimohanraj89 (approved):
This is messed up, good luck 😂

PR #1294, opened 2020-09-04
Assigned facility - Matview indices

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1171/index-assigned-facility-in-all-materialized-views

* Adds an index to `assigned_facility_id` on the heavily used `latest_blood_pressures_per_patient_per_months` materialized view
* Removes the `latest_blood_pressures_per_patient_per_days` while we're at it. It is no longer used.


db/schema.rb
```diff
@@ -760,6 +858,7 @@
           END AS current_age,
       p.gender,
       p.status,
+      p.assigned_facility_id,
```
harimohanraj89, line 861: Looks like this got missed until now?

PR #1296, opened 2020-09-04
Assigned facility reports

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1159/update-all-dashboard-reports-to-use-assigned-facility-instead-of-registration-facility

## Because

Assigned facility changes need to propagate to the new reports too.

## This addresses

* Modifies the four report cards to use assigned facility instead of registration facility for its calculations ([see doc](https://docs.google.com/document/d/1YZqm0t5NhTDUkn3Ww8yRqAszKCtAiOqChpFED6U-QXE/edit#heading=h.4q8qtwxbrqxq))
* Adds an index for assigned facility ID to the matview heavily used by reports
* Removes an unused matview

## To do

* [x] Implement design and copy changes
* [ ] [Not doing this until we get CVHO feedback] Rename "registrations" variables in the service classes to make more sense to developers 

### Overview

image

Screen Shot 2020-09-16 at 4 01 19 PM

---

### Details

Screen Shot 2020-09-16 at 4 01 24 PM

---

### Cohort Report

Screen Shot 2020-09-16 at 4 01 28 PM


app/services/control_rate_service.rb
```diff
@@ -58,7 +58,7 @@ def sum_cumulative_registrations
   def registration_counts
     return @registration_counts if defined? @registration_counts
     formatter = lambda { |v| quarterly_report? ? Period.quarter(v) : Period.month(v) }
-    result = region.registered_patients.with_hypertension.group_by_period(periods.begin.type, :recorded_at, {format: formatter}).count
+    result = region.assigned_patients.with_hypertension.group_by_period(periods.begin.type, :recorded_at, {format: formatter}).count
```
kitallis, line 61: Oh, we're scoping `with_hypertension` here, want to use `assigned_hypertension_patients` here directly?

db/migrate/20200904103339_drop_daily_bp_matview.rb
```diff
@@ -0,0 +1,5 @@
+class DropDailyBpMatview < ActiveRecord::Migration[5.2]
```
kitallis, line 1: 🙏

kitallis (changes requested):
Just a quick in-between review.

app/services/control_rate_service.rb
harimohanraj89: That was my first thought as well. Facility groups don't have that interface available, so I decided to keep the interface simple and just compose what we need here. Facility groups do feature an `assigned_patients` assocation through facilities quite nicely.

kitallis:
@harimohanraj89 Is the only main thing pending on this the copy changes?

harimohanraj89:
Yup. Once Claudio confirms the necessary copy changes, I'll add them in here and then ship. I wanted to get eyes on the query changes early.

kitallis (approved):
@harimohanraj89 This looks good to me. Let's wait until Monday before shipping this? For now, we can simply merge it in the base PR.

PR #1297, opened 2020-09-07
Improve user acess tree loading performance

kitallis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1108/make-rendering-the-access-tree-and-granting-access-more-efficient

## Because

When inviting or editing admins, large access trees can take up to 10s to load for users with a lot of access (managers or Orgs, power users).

## This addresses

- [x] Load the access tree asynchronously using `render_async` to improve perf while loading huge access trees
- [x] Remove `UserAccessTree` over a faster, easier way of rendering access trees from within `AdminAccessPresenter`
- [x] For edits, give space to the UI thread to render the access tree to avoid UI-locking when pre-checking checkboxes
- [x] Make the JS prototypical to harness better namespacing and page-level segregation

_async tree loading in action_
async-tree-load


app/assets/javascripts/common/admins.js
```diff
@@ -1,97 +1,117 @@
...
   accessLevelListener()
-  selectAccessLevels()
-  selectAllListener()
+  accessLevelSelector()
+}
+
+//
+// wait for the render_async hook before loading the access tree related JS
+//
+document.addEventListener('render_async_load', function (_event) {
```
kitallis, line 27: 1/2:

This is one of the main changes in this file, we're loading the access tree related JS after a callback hook from `render_async_load`, once it has finished loading.

app/assets/javascripts/common/admins.js
```diff
@@ -1,97 +1,117 @@
...
+
+  // go through all the checkboxes that are pre-checked and update their parents accordingly
+  for (const checkbox of checkboxes) {
+    if (!checkbox.checked) continue
 
-  for (const checkbox of checkedCheckboxes) {
-    updateParentCheckedState(checkbox, ACCESS_LIST_INPUT_SELECTOR)
+    // a large tree can take a lot of time to load on the DOM,
+    // so we queue up our updates by requesting frames so as to not cause overwhelming repaints
+    requestAnimationFrame(function () {
```
kitallis, line 46: 2/2:

This is the second useful change, when we're in the edit screen, and we pre-check the appropriate check-boxes, we still need to update the parents for indeterminate state ([can't provide indeterminate state from HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Indeterminate_state_checkboxes)).

This needs to loop through all the checked checkboxes and apply `updateParentCheckedState` on each, which can be quite expensive and I've seen slow rendering times because of it.

Here we queue up these updates so that it doesn't block the UI thread and make the page slow to render. It works quite well.

app/assets/javascripts/common/admins.js
```diff
@@ -1,97 +1,117 @@
+//
```
kitallis, line 1: This file has a bunch of refactoring, but the main two optimizations are inline commented down below.

app/presenters/admin_access_presenter.rb
```diff
@@ -18,31 +18,47 @@ def permitted_access_levels
...
       {
-        data: access_tree.facilities,
+        data: visible_facilities,
         render_partial: "email_authentications/invitations/facility_access_tree",
         root: :facility
       }
     end
   end
+
+  memoize def organization_tree
```
kitallis, line 43: Removed the `UserAccessTree` class in favor of these tiny 3-4 methods that now power the access trees.

app/assets/javascripts/common/admins.js
```diff
@@ -0,0 +1,213 @@
...
+    children.forEach(child => {
+      // reset indeterminate state for children
+      child.indeterminate = false
+      child.checked = parent.checked
+    })
+  },
+
+  onAsyncLoaded: function () {
+    const _self = this
+    document.addEventListener('render_async_load', function (_event) {
```
kitallis, line 117: 1/2:

This is one of the main changes in this file, we're loading the access tree related JS after a callback hook from render_async_load, once it has finished loading.

app/assets/javascripts/common/admins.js
```diff
@@ -0,0 +1,213 @@
...
+    const checkboxes = nodeListToArray(ACCESS_LIST_INPUT_SELECTOR, this.facilityAccess())
+
+    // go through all the checkboxes that are pre-checked and update their parents accordingly
+    for (const checkbox of checkboxes) {
+      if (!checkbox.checked) continue
+
+      // a large tree can take a lot of time to load on the DOM,
+      // so we queue up our updates by requesting frames so as to not cause overwhelming repaints
+      const _self = this
+      requestAnimationFrame(function () {
```
kitallis, line 140: 2/2:

This is the second useful change in this file, when we're in the edit screen, and we pre-check the appropriate check-boxes, we still need to update the parents for indeterminate state ([can't provide indeterminate state from HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Indeterminate_state_checkboxes)).

This needs to loop through all the checked checkboxes and apply `updateParentCheckedState` on each, which can be quite expensive and I've seen slow rendering times because of it.

Here we queue up these updates so that it doesn't block the UI thread and make the page slow to render. `requestAnimationFrame` works well here.

app/helpers/admin_access_helper.rb
```diff
@@ -3,26 +3,33 @@ def access_facility_count(available)
     "#{available} #{"facility".pluralize(available)}"
   end
 
-  def access_checkbox(form, name, resource, page: :new, checked_fn: -> { false })
-    return access_resource_label(resource) if page.eql?(:show)
+  def access_checkbox(name, resource, page: :new, checked_fn: -> { false })
+    case page
+      when :show
+        content_tag(:div, class: "form-check__show") { access_resource_label(resource) }
```
harimohanraj89, line 9: It took me a while to grok what's going on in here. Why overload this method if we're not returning a checkbox in this case? Should we use two different methods?

app/presenters/admin_access_presenter.rb
```diff
@@ -18,31 +18,47 @@ def permitted_access_levels
...
+  memoize def visible_organizations
+    admin.accessible_organizations(:any)
+  end
+
+  memoize def visible_facility_groups
+    admin.accessible_facility_groups(:any)
+  end
+
+  memoize def visible_facilities
+    admin.accessible_facilities(:any)
```
harimohanraj89, line 62: This isn't related to this PR, but could you shed some light on what the `any` action does, and how it should be used?

app/views/admins/show.html.erb
```diff
@@ -24,21 +24,24 @@
...
+        <div class="spinner-wrapper">
+          <div class="spinner spinner-border text-primary" role="status">
+            <span class="sr-only">Loading...</span>
+          </div>
+        </div>
+      <% end %>
     </div>
   </div>
+  <script>
+    new AdminAccess().onAsyncLoaded()
```
harimohanraj89, line 43: I'd love to pass the `facility-access` id to the initialization here. `new AdminAccess('#facility-access')`. It decouples the markup specifics from the JS class, and eventually lets us instantiate multiple AdminAccess panels too. Since it's low lift, can we make that change?

Even if there are three or four different handles that need to passed in, I think it's still worth decoupling. What do you think?

app/views/email_authentications/invitations/_facility_access_tree.html.erb
```diff
@@ -1,18 +1,18 @@
...
           <span class="spacer"></span>
         <% end %>
 
         <%= access_checkbox(
-                form,
                 :facilities,
                 facility,
                 page: page,
-                checked_fn: -> { user_being_edited.visible?(:facility, facility) }
+                checked_fn: -> { true }
```
harimohanraj89, line 13: Why are the boxes always checked? Is this intentional?

app/views/email_authentications/invitations/new.html.erb
```diff
@@ -15,27 +15,32 @@
...
+            <% end %>
           </div>
         </div>
 
         <%= form.submit "Send Invite", class: "btn btn-primary", id: "form-submit", hidden: :hidden %>
       <% end %>
     </div>
   </div>
+  <script>
+    const adminAccessInvite = new AdminAccessInvite()
```
harimohanraj89, line 40: Similar to above, I'd love to pass handles to the necessary DOM elements as constructor params. What do you think?

app/views/email_authentications/invitations/new.html.erb
```diff
@@ -15,27 +15,32 @@
...
+    adminAccessInvite.onAsyncLoaded()
+    adminAccessInvite.onDOMLoaded()
```
harimohanraj89, line 42: `adminAccessInvite.initialize()`?

harimohanraj89 (changes requested):

app/helpers/admin_access_helper.rb
kitallis: We can do that, I stuck it in here to keep the views simple, if I do separate methods, the check will go in the view, since the partials are shared.

app/controllers/admins_controller.rb
```diff
@@ -139,4 +152,8 @@ def user_params
       device_updated_at: Time.current
     }.compact
   end
+
+  def access_tree_for_page
+    params[:page].to_sym
```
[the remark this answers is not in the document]
kitallis: Yeah, it's added as a [query param](https://github.com/simpledotorg/simple-server/pull/1297/files/3ac2296bcc00cbb0e1c1740580ee5ec6d69581fb#diff-009fada69f121bd468eeb071eb7d9521R25) to the route.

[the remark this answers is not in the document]
kitallis: ...I'm going to rename this method to `page_for_access_tree`. The name is inverted.

app/helpers/admin_access_helper.rb
kitallis: I have an idea to make this clearer, let me give it a shot.

app/views/email_authentications/invitations/_facility_access_tree.html.erb
kitallis: It is, but I understand how this isn't obvious. The `checked_fn` implicitly checks if the page is `edit?` and hence when this page is called for edits, the admin in question is the editee which is pre-filtered by whichever facilities they can see anyway, so they should all be checked.

Let me find a way to make this more obvious.

kitallis: What a great catch, this was in fact a regression I introduced, it should now be,

```
user_being_edited.visible_facilities.include?(facility)
```

...which makes sense.

vkrmis (approved):

kitallis:
@harimohanraj89 I'm merging this in. I've addressed all your comments, if you wanna talk more, we can do it next week. Highlights of the changes:

* [Made the collapsing logic very simple](https://github.com/simpledotorg/simple-server/pull/1297/commits/a6c6dc6a72c9e66437f34d10557abd040b4ea353)
* Checkbox v label is now [inlined](https://github.com/simpledotorg/simple-server/pull/1297/commits/11b41e52a12de70a443985762be766691bf58e34) and helpers are [logic-free](https://github.com/simpledotorg/simple-server/pull/1297/commits/76c72eb262264833d2dea6bddd6b605bf69188dd) (just render plain HTML)
* Some other minor here-and-theres (renaming things etc.)

PR #1299, opened 2020-09-08
A few quick fixes around permissions

kitallis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1190/few-quick-bug-fixes-around-the-new-permissions-system

## Because

During tests with Praveen we caught a few real and potential issues. Addressed them in here.

## This addresses

- [x] Only power users can edit other power users (this was kind of already functional)
- [x] Non power users cannot edit power user (this was kind of already functional)
- [x] Only list admins based on their shared access to organizations
- [x] Edit page was showing multiple save buttons
- [x] Only allow power users to edit access levels
- [x] Automatically add new users as they are granted access to the new access system


app/controllers/email_authentications/invitations_controller.rb
```diff
@@ -25,6 +25,10 @@ def create
           new_user.save!
 
           current_admin.grant_access(new_user, selected_facilities)
+
+          #
+          # Automatically add new users as they are granted access to the new access system
+          Flipper.enable_actor(:new_permissions_system_aug_2020, Flipper::Actor.new(new_user.flipper_id))
```
vkrmis, line 31: Nice!

app/views/layouts/application.html.erb
```diff
@@ -110,7 +110,7 @@
                       <%= link_to "Protocols", admin_protocols_path, class: "dropdown-item #{active_controller?("admin/protocols")}" %>
                     <% end %>
 
-                    <% if current_admin.accessible_admins(:manage).any? %>
+                    <% if current_admin.accessible_facilities(:manage).any? %>
```
vkrmis, line 113: We're using `current_admin.accessible_facilities(:manage).any?` as a proxy for a lot of checks. Wondering if we can do any better.

vkrmis (approved):

vkrmis:
lgtm ✨

app/views/layouts/application.html.erb
kitallis: Yeah, we should not call these so many times, for one, we should wrap all of them in a single block.

Separately, this one actually should be `accessible_admins` – it's more correct. But I ended up using `accessible_facilities` for speed, since the other one is slower and will have an impact on all pages. But it's a good enough proxy for the admin check for now.

kitallis: I'll drop a comment about this.

PR #1300, opened 2020-09-08
Users table on new reports

rsanheim:
**Story card:** https://app.clubhouse.io/simpledotorg/story/445/detailed-users-table-in-facility-reports

Adding detailed user breakdown on new reports for facilities.

This reuses the same shared partial that the other reports are using as well as the same query under the hood.

image


claudiovallejo (approved):
Looks good Rob! Made a few small CSS updates to fix spacing between cards.

PR #1302, opened 2020-09-08
Monthly cohorts

rsanheim:
Get monthly cohorts working for feature parity w/ old reports.

**Story card:** https://app.clubhouse.io/simpledotorg/story/465/monthly-cohort-reports

**Screenshot**
image

## TODO

* [x] update labels for monthly cohorts to be correct


app/models/period.rb
```diff
@@ -87,6 +97,12 @@ def previous
     end
   end
 
+  def downto(number)
```
kitallis, line 100: Ha, took me a bit to understand this, but I get it. I also can't think of a simpler way that isn't imperative,

```ruby
# worse
def downto(number)
  periods = [self]
    
  number.times do
    periods << periods.last.previous
  end

  periods
end
```

app/services/cohort_service.rb
```diff
@@ -1,36 +1,37 @@
...
 
   # Each quarter cohort is made up of patients registered in the previous quarter
   # who has had a follow up visit in the current quarter.
   def call
     result = {quarterly_registrations: []}
-    quarters.each do |results_quarter|
-      quarter_data = compute_quarter(results_quarter)
-      result[:quarterly_registrations] << quarter_data
+    range.each do |period|
+      result[:quarterly_registrations] << compute(period)
```
kitallis, line 17: This is still outputting `quarterly_registrations`? How does that work? My understanding is that the range of these reports are now period-type agnostic, so we'll have to collect the results generically now?

kitallis (commented):
The strategy makes a lot of sense. I'm a little unclear how the different result-sets will be collected eventually, but I assume that is to come.

app/controllers/reports/regions_controller.rb
```diff
@@ -50,24 +50,19 @@ def details
 
   def cohort
     authorize(:dashboard, :show?)
+    period_params = report_params[:period].presence || {type: :month, value: Date.current.beginning_of_month}
```
harimohanraj89, line 53: `report_params[:period].presence || CohortService.default_period`?

app/models/period.rb
harimohanraj89: `downto` is standard enough that I think this is fine 😅

app/services/cohort_service.rb
```diff
@@ -1,36 +1,37 @@
 class CohortService
-  attr_reader :region, :quarters
   CACHE_VERSION = 1
   CACHE_TTL = 7.days
+  attr_reader :range
+  attr_reader :region
 
-  def initialize(region:, quarters: nil)
+  def initialize(region:, range: nil)
```
harimohanraj89, line 7: How about `periods:` instead of `range:`? Consistent with `ControlRateService`

app/services/cohort_service.rb
```diff
@@ -1,36 +1,37 @@
...
-  def compute_quarter(results_quarter)
-    Rails.cache.fetch(cache_key(results_quarter), version: cache_version, expires_in: CACHE_TTL, force: force_cache?) do
-      cohort_quarter = results_quarter.previous_quarter
-      period = {cohort_period: :quarter,
-                registration_quarter: cohort_quarter.number,
-                registration_year: cohort_quarter.year}
+  def compute(period)
+    Rails.cache.fetch(cache_key(period), version: cache_version, expires_in: CACHE_TTL, force: force_cache?) do
+      cohort_period = period.previous
+      period = {cohort_period: cohort_period.type,
```
harimohanraj89, line 27: We're clobbering the method argument `period` with this local variable 😅

app/views/reports/regions/_header.html.erb
```diff
@@ -21,16 +21,21 @@
   <p class="mb-12px fs-xxlarge fw-bold">
     <%= @region.name %>
   </p>
+  <%
+    # We want to maintain the request params for Overview and Detail nav links, but _only_ in the
+    # case where its _not_ a quarter (as those pages don't support quarters).
+    url_params = (params.dig(:period, :type) == "quarter") ? {} : request.params
```
harimohanraj89, line 27: Any harm in _not_ excluding quarters? Do those actions gracefully ignore unsupported params, or do they loudly fail?

app/services/cohort_service.rb
```diff
@@ -1,36 +1,37 @@
...
 
-  def initialize(region:, quarters: nil)
+  def initialize(region:, range: nil)
     @region = region
-    @quarters = quarters || default_quarters
+    @range = range || default_range
   end
 
   # Each quarter cohort is made up of patients registered in the previous quarter
   # who has had a follow up visit in the current quarter.
```
harimohanraj89, line 13: We should update this comment's wording to be period-type-agnostic.

app/services/cohort_service.rb
harimohanraj89: Since each element of the array we're building here is rich with cohort info, we could even drop this top-level key, and return an array. Should we do that?

app/services/cohort_service.rb
```diff
@@ -1,36 +1,37 @@
...
-      period = {cohort_period: :quarter,
-                registration_quarter: cohort_quarter.number,
-                registration_year: cohort_quarter.year}
+  def compute(period)
+    Rails.cache.fetch(cache_key(period), version: cache_version, expires_in: CACHE_TTL, force: force_cache?) do
+      cohort_period = period.previous
+      period = {cohort_period: cohort_period.type,
+                registration_quarter: cohort_period.value.try(:number),
+                registration_year: cohort_period.value.try(:year),
+                registration_month: cohort_period.value.try(:month)}
```
harimohanraj89, line 30: I guess we don't _need_ to, but should we version-bump the cache for good measure?

app/services/cohort_service.rb
```diff
@@ -1,36 +1,37 @@
...
   private
 
-  def compute_quarter(results_quarter)
-    Rails.cache.fetch(cache_key(results_quarter), version: cache_version, expires_in: CACHE_TTL, force: force_cache?) do
-      cohort_quarter = results_quarter.previous_quarter
-      period = {cohort_period: :quarter,
-                registration_quarter: cohort_quarter.number,
-                registration_year: cohort_quarter.year}
+  def compute(period)
+    Rails.cache.fetch(cache_key(period), version: cache_version, expires_in: CACHE_TTL, force: force_cache?) do
```
harimohanraj89, line 25: `cache_key(period)` will gracefully generate separate keys for monthly cohorts. Nice.

Can we update the name of the argument in the method definition below to `period`?

app/views/reports/regions/_header.html.erb
```diff
@@ -43,24 +48,45 @@
...
-
-          <% (1..6).each do |num|
-              selected = false
-              date = Date.current.beginning_of_month.advance(months: -num)
-              period = Period.new(type: :month, value: date)
+      <% if action_name == "cohort" %>
+        <div class="dropdown-menu" aria-labelledby="dropdownMenuLink">
+          <% [:quarter, :month].each do |period_type|
+            period = Period.new(type: period_type, value: @period.to_date)
+            selected = true if period.type == @period.type && period == @period
```
harimohanraj89, line 61: Hmmm, should we relax our `Period` comparator to return false if types are different?

harimohanraj89 (dismissed):
Feedback below! A few definite blockers. Feel free to get Tim to override our reviews once addressed.

app/services/cohort_service.rb
rsanheim: Yup, just pushed up a change to just return the array. 👍

app/views/reports/regions/_header.html.erb
rsanheim: They will fail, because we try to build the `Period` from params for all the actions, and a Period can built w/ a type of `month` or `quarter`. But only the `cohort` stuff deals w/ quarters (for now).

app/views/reports/regions/_header.html.erb
rsanheim: Yup did that.

app/controllers/reports/regions_controller.rb
timcheadle: Agreed, let's delegate this default to the service as we're doing in `set_period`

app/helpers/reports/regions_helper.rb
```diff
@@ -3,4 +3,8 @@ def percentage_or_na(value, options)
     return "N/A" if value.blank?
     number_to_percentage(value, options)
   end
+
+  def cohort_report_type(period)
+    "#{period.type.to_s.humanize}ly report"
```
timcheadle, line 8: Not sure we need the "report" verbage in here, should we leave that up to the view?

timcheadle (commented):
I'm also getting different months in the new vs old reports:

**New reports:**
image

**Old reports:**
image

claudiovallejo:
@rsanheim this looks good! Made a small visual update — I reversed the order of the bar graph's title/subtitle.

claudiovallejo (approved):

app/helpers/reports/regions_helper.rb
rsanheim: This is just a view helper, so I think this is fine.

app/models/period.rb
rsanheim: Yeah, its a ruby stdlib method on a few core classes: https://apidock.com/ruby/Integer/downto

PR #1303, opened 2020-09-10
Gate RegionsController for new permissions

vkrmis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1134/migrate-the-regionscontroller-and-its-actions

## Because

## This addresses

Migrates the RegionsController and it's views to use the new permissions (gated by a Flipper flag)

### DIFF
--- a/app/controllers/reports/regions_controller.rb
+++ b/app/controllers/reports/regions_controller.rb
@@ -15,7 +15,6 @@ class Reports::RegionsController < AdminController
     if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin)
       authorize1 { current_admin.accessible_facilities(:view_reports).any? }
       @organizations = current_admin.accessible_facilities(:view_reports)
-        .includes(facility_group: :organization)
         .flat_map(&:organization)
         .uniq
         .compact
--- a/app/controllers/reports/regions_controller.rb
+++ b/app/controllers/reports/regions_controller.rb
@@ -8,14 +8,30 @@ class Reports::RegionsController < AdminController
   before_action :find_region, except: :index
   around_action :set_time_zone
 
-  def index
-    authorize(:dashboard, :show?)
+  skip_after_action :verify_authorized, if: -> { Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) }
+  after_action :verify_authorization_attempted, if: -> { Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) }
 
-    @organizations = policy_scope([:cohort_report, Organization]).order(:name)
+  def index
+    if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin)
+      authorize1 { current_admin.accessible_facilities(:view_reports).any? }
+      @organizations = current_admin.accessible_facilities(:view_reports)
+        .includes(facility_group: :organization)
+        .flat_map(&:organization)
+        .uniq
+        .compact
+        .sort_by(&:name)
+    else
+      authorize(:dashboard, :show?)
+      @organizations = policy_scope([:cohort_report, Organization]).order(:name)
+    end
   end
 
   def show
-    authorize(:dashboard, :show?)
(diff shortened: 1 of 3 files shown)


app/controllers/reports/regions_controller.rb
```diff
@@ -8,14 +8,30 @@ class Reports::RegionsController < AdminController
...
 
-    @organizations = policy_scope([:cohort_report, Organization]).order(:name)
+  def index
+    if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin)
+      authorize1 { current_admin.accessible_facilities(:view_reports).any? }
+      @organizations = current_admin.accessible_facilities(:view_reports)
+        .includes(facility_group: :organization)
+        .flat_map(&:organization)
+        .uniq
+        .compact
```
prabhanshuguptagit, line 21: Which scope here introduces the dups and nils? Should we move these to inside `accessible_facilities` if that's the source? Should we move this to `accessible_organizations` and use that instead?

app/views/reports/regions/index.html.erb
```diff
@@ -15,10 +15,26 @@
...
-            <%= link_to(reports_region_path(district, report_scope: "district"), class: "link-row") do %>
-              <i class='fas fa-angle-right light'></i>
-              <%= district.name %>
+          <% if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) %>
+            <% current_admin.accessible_facilities(:view_reports)
+               .includes(facility_group: :organization)
+               .where(facility_groups: {organization: organization})
+               .flat_map(&:facility_group)
+               .uniq
+               .compact
```
prabhanshuguptagit, line 24: Same here, if we could call `accessible_facility_groups` that would be nice.

app/controllers/reports/regions_controller.rb
vkrmis: `@organizations` is used to show links for organizations where the current_admin has access to any facilities under them. We need to `.uniq.compact` this so we only show one link for each organization.

We can't use `accessible_organizations` for this because in the new permissions system, having access to a single facility under an organization isn't the same as having access to the whole organization. This means `accessible_organizations` only returns organizations for power users and organization managers.

This worked in the old permissions system because the [:cohort_report, Organization] policy would traverse the tree in both directions and include organizations in the scope even if the user only had access to a single facility under them.

app/views/reports/regions/index.html.erb
vkrmis: _ditto_ ^

app/controllers/reports/regions_controller.rb
prabhanshuguptagit: 👍  Does it make sense to add convenience methods for getting such orgs?

prabhanshuguptagit (approved):

kitallis (approved):

app/views/reports/regions/index.html.erb
```diff
@@ -15,10 +15,26 @@
       <section>
         <div class="card">
           <h2><%= organization.name %></h2>
-          <% policy_scope([:cohort_report, organization.facility_groups]).order(:name).each do |district| %>
-            <%= link_to(reports_region_path(district, report_scope: "district"), class: "link-row") do %>
-              <i class='fas fa-angle-right light'></i>
-              <%= district.name %>
+          <% if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) %>
+            <% current_admin.accessible_facilities(:view_reports)
+               .includes(facility_group: :organization)
```
kitallis, line 20: I think this can be removed, we've already [inlined](https://github.com/simpledotorg/simple-server/blob/master/app/models/user_access.rb#L65) this `includes`.

vkrmis:
@prabhanshuguptagit we went back and forth on this a bit. We had a version of `visible_*` methods at some point, but removed them to keep things simple and unambiguous. If this becomes a common enough need, we can revisit and add some convenience methods.

PR #1304, opened 2020-09-10
Gate Analytics Controllers and views

vkrmis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1209/migrate-the-analytics-controller-and-its-actions

## Because

## This addresses

Migrates the Analytics Controller views to use the new permissions (gated by a Flipper flag)


app/views/analytics/facilities/show.html.erb
```diff
@@ -49,28 +62,55 @@
...
-      <% end %>
-    </p>
-    <p>
-      <%= link_to analytics_facility_patient_list_with_history_path(@facility) do %>
-        <i class="far fa-file-excel mr-2"></i>
-        Patient line list (with medication history)
-      <% end %>
-    </p>
+  <% if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) %>
+    <% if current_admin.accessible_facilities(:view_reports).include?(@facility) %>
```
harimohanraj89, line 66: This should be `:view_pii`, right? For the line list?

app/views/shared/_analytics_table.html.erb
```diff
@@ -73,37 +73,73 @@
...
-              <% else %>
-                <%= link_to resource.send(row_resource_display_field), send(row_resource_link, resource) %>
-              <% end %>
-            </td>
-
-            <!--Registered patients-->
-            <td style="text-align: center;">
-              <%= dash_if_zero(dashboard_analytics.dig(resource.id, :total_registered_patients)) %>
-            </td>
+      <% if Flipper.enabled?(:new_permissions_system_aug_2020, current_admin) %>
```
harimohanraj89, line 76: Can we simplify these big chunks of ERB branching by pulling out just the bits we need to switch on? In both cases we're getting some collection in two different ways, but iterating through it the same way.

Can we use the Flipper if-else just to set a local `resources` var, and then have a single `resource.each` block?

harimohanraj89 (changes requested):
Some questions and ideas!

app/controllers/analytics/districts_controller.rb
```diff
@@ -4,6 +4,7 @@ class Analytics::DistrictsController < AnalyticsController
 
   before_action :set_organization_district
   skip_after_action :verify_authorized
+  skip_after_action :verify_authorization_attempted
```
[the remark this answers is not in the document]
vkrmis: The `DistrictsController` has never had any (the previous line skips the old authorization check). Unsure why this is the case, so I haven't changed it.

app/views/analytics/facilities/show.html.erb
```diff
@@ -18,19 +18,32 @@
...
 
-<%= render 'shared/analytics_table',
-  dashboard_analytics: @dashboard_analytics,
-  period: @period,
-  row_resource: User.order(:full_name),
-  row_resource_description: "Users",
-  row_resource_display_field: :full_name,
-  row_resource_link: :admin_user_path %>
-
-<% if policy([:cohort_report, @facility]).view_health_worker_activity? %>
```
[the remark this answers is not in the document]
vkrmis: Missed this, switching it to `view_pii`

app/views/analytics/facilities/show.html.erb
vkrmis: Missed this, switching it to `view_pii`

app/views/shared/_analytics_table.html.erb
vkrmis: Will do this.

I've optimized for easily deleting the `else` branches, but it does lead to big mostly-duplicated chunks.

vkrmis: This is done

kitallis (approved):
❇️

PR #1305, opened 2020-09-10
Fix broken API docs

prabhanshuguptagit:
**Story card:** -

## Because
Our swagger docs broke. This is because we missed committing the latest `swagger.json` with the teleconsult API changes.
image

## This addresses

This runs a `rake docs` and updates `swagger.json`.


harimohanraj89 (approved):
Nice. If this works in the review app, :shipit:

PR #1306, opened 2020-09-10
Migrate Appointments Controller

vkrmis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1133/migrate-the-appointmentscontroller-and-its-actions
## Because

## This addresses

Migrates the AppointmentsController and it's views to use the new permissions (gated by a Flipper flag)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -85,7 +85,7 @@
 
               <% if current_admin.accessible_facilities(:manage_overdue_list).any? %>
                 <li class="nav-item">
-                  <%= link_to "Overdue patients", appointments_path, class: "nav-link #{active_controller?("appointments")}" %>
+                  <%= link_to "Overdue", appointments_path, class: "nav-link #{active_controller?("appointments")}" %>
                 </li>
               <% end %>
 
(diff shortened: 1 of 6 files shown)


app/views/layouts/application.html.erb
```diff
@@ -85,7 +85,7 @@
 
               <% if current_admin.accessible_facilities(:manage_overdue_list).any? %>
                 <li class="nav-item">
-                  <%= link_to "Overdue patients", appointments_path, class: "nav-link #{active_controller?("appointments")}" %>
+                  <%= link_to "Overdue", appointments_path, class: "nav-link #{active_controller?("appointments")}" %>
```
harimohanraj89, line 88: Why this change?

harimohanraj89 (approved):
This looks super straightforward, nice. :shipit: :shipit: :shipit:

app/views/layouts/application.html.erb
vkrmis: accidentally included " patients" in an earlier PR, the old-permissions branch of the layout just says "Overdue" now, so I've removed the extra " patients" here.

harimohanraj89: Noice 😎

PR #1307, opened 2020-09-10
Install Skylight

timcheadle:
## Because

We're exploring https://www.skylight.io as an alternative to New Relic

## This addresses

Initial setup of Skylight. The necessary `SKYLIGHT_AUTHENTICATION` key is in 1password.


harimohanraj89 (approved):
Took this for a spin on SBX with the auth key. Looks good.

Confirmed that other environments (like QA) still function correctly if no key is configured. This should be safe to ship to production.

PR #1309, opened 2020-09-10
Refactor feature check responsibility to user for actor flipper checks

rsanheim:
For all of our feature flag checks instead of 

```ruby
 if Flipper.enabled?(:name, current_admin)
```

allow this:

```ruby
if current_admin.feature_enabled?(:name)
```

and for features used everywhere (like permissions), go one step further and add a first class method:

```ruby
if current_admin.permissions_v2_enabled?
```

This makes for more readable code throughout all our feature checks, and
also guides callers to use the one way of checking actor gates on our
current_admin / current_user.  Also, having one method to grep for makes it easier to rip out feature flag checks once things have shipped.

For the permissions checks, I'm using `permissions_v2` as the feature name, because `new_permissions_system_aug_2020` is awkward and also its now September 😀 .  Also, going with a plain version number is clear and explicit for systems like this where we would _hopefully_ have low turnover of numbers over time.

I also renamed any sort of auth check methods to match the `_v2` naming convention - this makes it much easier to rename them to be the primary method when the old permissions code can be ripped out.


kitallis:
@rsanheim Makes sense to do this 👍 . The name of the feature flag has the date of creation rather than the current time of the year. But maybe v2 is enough? This is probably v3 of permissions though, or maybe v4, version numbers with feature names didn't really make any sense, because it completely depends on when one joined the project 😝.

harimohanraj89:
Can we check if this breaks if there's no current admin? 😅 I want to make sure this won't break if you're not signed in.

As for the name, we're going to rip this feature flag out after launch, so I don't mind either way. 🤷

rsanheim:
>  But maybe v2 is enough? This is probably v3 of permissions though, or maybe v4, version numbers with feature names didn't really make any sense, because it completely depends on when one joined the project

Yeah I guess I was thinking "v2" just means the "next version".  We could also do that sort of naming - `permissions_next_enabled?`, though that also could be confusing.  I mostly just wanted a more concise name for scanning and reading code.

app/controllers/admins_controller.rb
```diff
@@ -113,7 +113,7 @@ def verify_params
   #
   # This is a temporary `verify_params` method that will exist until we migrate fully to the new permissions system
   #
-  def 🆕verify_params
```
rsanheim, line 116: please, no emoji in code 😁 .  I know editors and ruby can handle it, but it makes grepping or finding things harder.

app/presenters/admin_access_presenter.rb
```diff
@@ -3,15 +3,22 @@
...
   end
 
   def display_access_level
-    OpenStruct.new(UserAccess::LEVELS.fetch(admin.access_level.to_sym))
+    access_level = if admin_access_level
+      UserAccess::LEVELS.fetch(admin_access_level.to_sym)
+    else
+      {name: "(Not Set)", description: "N/A"}
+    end
+    OpenStruct.new(access_level)
```
rsanheim, line 21: This was needed so that the admin listing page wouldn't blow up for users who don't have an access level. Found this while trying to kick the tires of perms v2 in dev.

rsanheim (commented):
Some notes on further refactorings I did.

rsanheim:
Going to go ahead and merge this because its touching a lot of files and we are going to have merge conflict hell if we don't get it landed in master.  This should make cleaning up the permissions code easier once we are ready to make v2 the actual production permissions system.

PR #1310, opened 2020-09-11
Migrate teleconsult MOs to MVP

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1027

## Because

The current telemed implementation has phone numbers linked to facilities. We want to move to linking users with facilities instead.

## This addresses

Adds a rake task to link MOs to facilities if their phone number matches the current `teleconsultation_phone_numbers`. This looks at user's `phone_number` and `teleconsult_phone_number` to find a match.


lib/tasks/scripts/link_teleconsultation_medical_officers.rb
```diff
@@ -0,0 +1,37 @@
+class LinkTeleconsultationMedicalOfficers
+  class << self
+    def call
+      linked_users_count, unlinked_numbers_count = 0, 0
+
+      Facility.all.each do |facility|
```
harimohanraj89, line 6: Use `Facility.find_each` to avoid instantiating a bajillion Ruby objects to save memory!

lib/tasks/scripts/link_teleconsultation_medical_officers.rb
```diff
@@ -0,0 +1,37 @@
...
+      Rails.logger.info "Could not find a match for #{unlinked_numbers_count} numbers."
+    end
+
+    private
+
+    def link_medical_officer(medical_officer, facility)
+      unless facility.teleconsultation_medical_officers.include?(medical_officer)
+        facility.teleconsultation_medical_officers << medical_officer
+
+        Rails.logger.info "Linked #{medical_officer.id} to #{facility.name}"
```
harimohanraj89, line 27: We should log that we did this _after_ we `facility.save!` 😅

spec/tasks/scripts/link_teleconsultation_medical_officers_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      facility_1_mos = facility_1_phone_numbers.map { |phone_number|
+        create(:user, teleconsultation_phone_number: phone_number)
+      }
+
+      described_class.call
+
+      expect(facility_1.reload.teleconsultation_medical_officers).to match_array [facility_1_mos, existing_mo].flatten
+    end
+  end
```
harimohanraj89, line 69: Let's add another spec where we don't create any users, and expect the linker to run correctly.

lib/tasks/scripts/link_teleconsultation_medical_officers.rb
```diff
@@ -0,0 +1,37 @@
+class LinkTeleconsultationMedicalOfficers
+  class << self
+    def call
+      linked_users_count, unlinked_numbers_count = 0, 0
+
+      Facility.all.each do |facility|
+        facility.teleconsultation_phone_numbers.each do |number|
+          medical_officer = find_medical_officer_by_number(number["phone_number"])
```
harimohanraj89, line 8: Finding both linked and unlinked users in the same method is a bit complex 😅 Could we split this up a bit, to something like:

```ruby
phone_number = number["phone_number"]

if already_linked?(phone_number)
  already_linked_users_count += 1
else if linkable?(phone_number)
  link(phone_number)
  linked_users_count += 1
else
  unlinked_users_count += 1
end
```

harimohanraj89 (changes requested):
Some feedback. Lookin' good.

harimohanraj89 (approved):
The diff got borked 😅  but your latest commits look solid. Godspeed.

PR #1313, opened 2020-09-14
Minor permissions fixes/improvements

vkrmis:
**Story card:** 

## Because

Found some minor issues with the new permissions work

## This addresses

Includes the following fixes:
- [x] Layout should say "Home" instead of "My Facilities"
- [x] Delete Facility on the Edit Facility page doesn't check if facility is deletable
- [x] Delete FacilityGroup on the Edit FacilityGroup page doesn't check if facility group is deletable
- [x] Admins#index is not sorted
- [x] Admins edit screen has all facilities pre-selected


app/helpers/admin_access_helper.rb
```diff
@@ -3,7 +3,7 @@ def access_facility_count(available)
     "#{available} #{"facility".pluralize(available)}"
   end
 
-  def access_checkbox(name, resource, checked)
+  def access_checkbox(name, resource, checked: nil)
```
kitallis, line 6: @vkrmis Default it to `false` instead. It can only be `true` or `false`.

kitallis (approved):
Looks solid. Just left a small comment.

app/models/facility.rb
```diff
@@ -255,6 +255,10 @@ def _destroy
     end
   }
 
+  def discardable?
```
harimohanraj89, line 258: Let's slap some test coverage on these `discardable` methods.

rsanheim: Would there be cases were we want to cascade the discard to the associated records here?  I can imagine that for 90%+ of cases, there will be some associated records on facility or facility groups, so this would almost always block the discard.

vkrmis: We don't want to cascade discards, we only want to allow a discard from the UI is brand new organizations/facility groups/facilities that were created accidentally etc.

The other case is when there are duplicate facilities, but those need the data migrated manually before we allow the discard.

harimohanraj89: I suspect this will allow for deletion if you _just_ set something up and you wanna trash it.

harimohanraj89: Deleting a facility/group that's in use is... much bigger :/ I don't think it'll be a usable feature since we don't have a way to, say, move patients from one facility to another...

PR #1315, opened 2020-09-15
Add patient list downloads to new reports

rsanheim:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1217/add-patient-line-list-to-main-tab

For feature parity w/ old reports.

image

### DIFF
--- a/app/controllers/reports/patient_lists_controller.rb
+++ b/app/controllers/reports/patient_lists_controller.rb
@@ -19,7 +19,7 @@ class Reports::PatientListsController < AdminController
 
     PatientListDownloadJob.perform_later(recipient_email, region_class, download_params)
 
-    redirect_back(fallback_location: reports_region_path(@region, report_scope: "facility"))
+    redirect_back(fallback_location: reports_region_path(@region, report_scope: params[:report_scope]))
   end
 
   private
--- a/app/controllers/reports/patient_lists_controller.rb
+++ b/app/controllers/reports/patient_lists_controller.rb
@@ -25,9 +25,9 @@ class Reports::PatientListsController < AdminController
   private
 
   def find_region
-    id = filtered_params[:id]
+    slug = filtered_params[:id]
     klass = region_class.classify.constantize
-    @region = klass.find(id)
+    @region = klass.find_by!(slug: slug)
   end
 
   def filtered_params
--- a/app/controllers/reports/patient_lists_controller.rb
+++ b/app/controllers/reports/patient_lists_controller.rb
@@ -10,9 +10,12 @@ class Reports::PatientListsController < AdminController
         authorize_v2 { current_admin.accessible_facilities(:view_pii) }
       end
     else
-      authorize([:cohort_report, region_class.classify.constantize], :patient_list?)
+      policy_scope([:cohort_report, region_class.classify.constantize])
     end
     @region = scope.find_by!(slug: params[:id])
+    unless current_admin.permissions_v2_enabled?
+      authorize([:cohort_report, @region], :patient_list?)
+    end
 
     recipient_email = current_admin.email
     download_params = if region_class == "facility_group"
--- a/app/controllers/reports/patient_lists_controller.rb
+++ b/app/controllers/reports/patient_lists_controller.rb
@@ -1,0 +1,47 @@
+class Reports::PatientListsController < AdminController
+  skip_after_action :verify_policy_scoped
+  before_action :find_region
+
+  def show
+    if current_admin.permissions_v2_enabled?
+      authorize_v2 { current_admin.accessible_facilities(:view_reports).any? }
+    else
+      authorize(:dashboard, :show?)
+    end
+
+    recipient_email = current_admin.email
+
+    download_params = if region_class = "facility_group"
+      {id: @region.id}
+    else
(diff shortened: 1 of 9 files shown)


app/controllers/reports/patient_lists_controller.rb
```diff
@@ -0,0 +1,44 @@
+class Reports::PatientListsController < AdminController
+  skip_after_action :verify_authorized, if: -> { current_admin.permissions_v2_enabled? }
+  after_action :verify_authorization_attempted, if: -> { current_admin.permissions_v2_enabled? }
+
+  def show
+    scope = if current_admin.permissions_v2_enabled?
+      if region_class == "facility_group"
+        authorize_v2 { current_admin.accessible_facility_groups(:view_pii) }
+      else
+        authorize_v2 { current_admin.accessible_facilities(:view_pii) }
```
rsanheim, line 10: This authorize_v2 block feels unnecessary w/ the new permissions system...since the scope already handles the authorization for us.

app/controllers/reports/patient_lists_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+    else
+      authorize([:cohort_report, region_class.classify.constantize], :patient_list?)
```
rsanheim, line 13: I can't figure out how to use the pundit system to return the scope of records that the admin has patient_list access to...which is what I want so that we can be consistent w/ the v2 way of doing things here.

app/controllers/reports/patient_lists_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+    @region = scope.find_by!(slug: params[:id])
```
rsanheim, line 15: Finally, here is the check that does the work.  If we know that the "scope" is already limited to things the admin has access too, we can rely on `find_by!` to either return something the admin has access to or raise an ActiveRecord::RecordNotFound.

app/policies/cohort_report/facility_group_policy.rb
```diff
@@ -1,4 +1,16 @@
 class CohortReport::FacilityGroupPolicy < ApplicationPolicy
+  def patient_list?
+    user_has_any_permissions?(
+      [:download_patient_line_list, nil],
+      [:download_patient_line_list, record.organization],
+      [:download_patient_line_list, record]
+    )
+  end
```
rsanheim, line 8: This was an attempt to add the pundit check for patient_list rights for the Facility Group.  This doesn't seem right tho.

image

rsanheim (commented):
some notes on where I'm blocked here.

app/controllers/reports/patient_lists_controller.rb
rsanheim: To clarify here: Not trying to block any progress 😁 . We can circle back around later after permissions v2 is shipped to improve things - just trying to clarify now that I'm getting some real experience w/ the new perms system.

kitallis: I think that's fine, you can totally go ahead and not use the `authorize_v2` if you think you don't need it. And since you don't have an `index` method, you can _probably_ avoid it. I'm not at all married to having it around, it's the least interesting thing in the entirety of the permissions work :)

But here are some concrete reasons why it could be helpful to have *something* like it:

1. When you are dealing with collections, you will never call `find_by`. This is true for all index actions in all controllers. This means you can potentially scope things that you don't have authority to scope.
2. Wrapping it in a block allows us a generic API to provide a guard-rail against actions that have not been authorized, it'll immediately throw an exception if you don't wrap your actions in a block. This is obviously not as powerful as Pundit, because the Pundit guardrails would explicitly work for the specific things you are scoping not an arbitrary call to a block.
3. It's easier to see that authorization is taken care of when reading code. It's explicit and consistent instead of a hidden catch-all on `RecordNotFound` or in every controller.

It adds *some* value, but maybe not a whole lot. It was definitely easier to migrate from old to new using this because it keeps the old functionality in parity.

This idea, and pundit, and the million other auth libraries in Rails are all generally based off of the [PERM metamodel](https://casbin.org/docs/en/how-it-works) from https://casbin.org.

app/controllers/reports/patient_lists_controller.rb
kitallis: Since pundit is based off of PERM, it identifies policy checking and scoping as two distinct operations on a resource.

What you are doing here is a policy check / authorization in pundit-parlance, I believe what you need is a scope:

```
policy_scope([:cohort_report, region_class.classify.constantize])
```

And potentially call,

```
policy([:manage, region_class.classify.constantize]).patient_list?
```

...in the views where you want to hide/show the buttons, like: [this](https://github.com/simpledotorg/simple-server/blob/master/app/views/analytics/facilities/show.html.erb#L82) and [this](https://github.com/simpledotorg/simple-server/blob/master/app/views/analytics/districts/show.html.erb#L73).

That should be sufficient.

app/policies/cohort_report/facility_group_policy.rb
kitallis: This looks accurate to me.

claudiovallejo (approved):

PR #1316, opened 2020-09-15
Teleconsult sync API

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1005

## Because

We are adding Teleconsultations as a first class entity as part of the MVP to capture telemed records.

## This addresses

- Create a `Teleconsultation` model
  - Add a `requested_medical_officer_id` to keep track of the MO that was requested to perform a teleconsult.
  - Add a `medical_officer_id` which tracks the MO who actually performed the teleconsult.

- Add the `/v4/teleconsultations/sync` API endpoint

Notes: 
- This is a push only endpoint, since the app never fetches these records.
- The `request` and `record` are synced independently to the server. A user (nurse or MO) specifies only one of these and sets the other key to `null` on the app.
- Prescription drugs are linked to teleconsultations by a `teleconsultation_id` on `prescription_drugs`(#1320)


app/controllers/api/v4/teleconsultations_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      teleconsultation_params.permit(
+        :id,
+        :patient_id,
+        :medical_officer_id,
+        :request,
+        :updated_at,
+        :deleted_at,
+        :created_at,
+        request: Teleconsultation::REQUEST_ATTRIBUTES,
+        record: Teleconsultation::RECORD_ATTRIBUTES << {prescription_drugs: []}
```
prabhanshuguptagit, line 33: Strong Parameters needs nested arrays need to be specified.

app/models/teleconsultation.rb
```diff
@@ -0,0 +1,42 @@
+class Teleconsultation < ApplicationRecord
+  include Mergeable
+
+  belongs_to :patient
+  belongs_to :medical_officer, class_name: "User", foreign_key: :medical_officer_id
+  belongs_to :facility, optional: true
+  belongs_to :requester, class_name: "User", foreign_key: :requester_id, optional: true
+
+  REQUEST_ATTRIBUTES = %w[requester_id facility_id requested_at request_completed]
```
kitallis, line 9: `freeze` the `REQUEST_ATTRIBUTES` and `RECORD_ATTRIBUTES` as well? 

Or add `# frozen_string_literal: true` on top of the file.

spec/models/teleconsultation_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      FactoryBot.build(:teleconsultation,
+        medical_officer: medical_officer,
+        teleconsultation_type: "audio",
+        patient_took_medicines: "yes",
+        patient_consented: "yes",
+        medical_officer_number: "",
+        prescription_drugs: prescription_drug_ids,
+        facility: facility)
+    end
+    it "returns the teleconsult record data" do
```
kitallis, line 49: [NIT] Space between `end` and `it`.

app/transformers/api/v4/teleconsultation_transformer.rb
```diff
@@ -0,0 +1,20 @@
...
+        .except(*Teleconsultation::RECORD_ATTRIBUTES)
+        .merge({"request" => teleconsultation.request,
+                "record" => teleconsultation.record})
+    end
+
+    def from_request(teleconsultation)
+      request, record = teleconsultation["request"], teleconsultation["record"]
+      payload = Api::V4::Transformer.from_request(teleconsultation).except("request", "record")
+
+      payload.merge!(request) if request
```
kitallis, line 15: What do we think about making the data model change right away and update the PD's `teleconsultation_id` in a `MergeTeleconsultationService` over storing the PD IDs in an array on `Teleconsultation`?

It'd make the migration easier later, since you'll only have to change code-paths and not migrate any data. We can `except` that `teleconsultation_id` from the `prescription_drug` payload while syncing until the app can understand it.

kitallis (commented):
Looks pretty solid, just left a couple of minor comments and a question.

app/transformers/api/v4/teleconsultation_transformer.rb
harimohanraj89: The annoying part is, there's no guarantee the PD will be synced when this request is received. That's part of the reason why we decided to bite the bullet and strip out these arrays everywhere - [ref](https://simpledotorg.slack.com/archives/CV9V1DWLV/p1600251647036000)

spec/factories/teleconsultations.rb
```diff
@@ -0,0 +1,36 @@
...
+end
+
+def build_invalid_teleconsultation_payload
+  build_teleconsultation_payload.merge("created_at" => nil)
+end
+
+def updated_teleconsultation_payload(existing_teleconsultation)
+  update_time = 10.days.from_now
+  build_teleconsultation_payload(existing_teleconsultation).merge("updated_at" => update_time)
+end
```
harimohanraj89, line 40: Is the factory the right place to define these methods? Can they be defined within the spec scopes where they're needed?

app/models/teleconsultation.rb
```diff
@@ -0,0 +1,42 @@
+class Teleconsultation < ApplicationRecord
+  include Mergeable
+
+  belongs_to :patient
+  belongs_to :medical_officer, class_name: "User", foreign_key: :medical_officer_id
+  belongs_to :facility, optional: true
```
harimohanraj89, line 6: Is this optional because the `record` may arrive first?

app/models/teleconsultation.rb
```diff
@@ -0,0 +1,42 @@
+class Teleconsultation < ApplicationRecord
+  include Mergeable
+
+  belongs_to :patient
+  belongs_to :medical_officer, class_name: "User", foreign_key: :medical_officer_id
```
harimohanraj89, line 5: Does this need to be optional? We won't have it when the request comes in first right?

app/models/teleconsultation.rb
```diff
@@ -0,0 +1,42 @@
+class Teleconsultation < ApplicationRecord
```
harimohanraj89, line 1: Generic question: what's a good way for us to query how many teleconsultations are incomplete (requested, but never recorded).

harimohanraj89 (changes requested):
My thoughts below!

app/controllers/api/v4/teleconsultations_controller.rb
```diff
@@ -0,0 +1,37 @@
+class Api::V4::TeleconsultationsController < Api::V4::SyncController
+  def sync_from_user
+    __sync_from_user__(teleconsultation_params)
+  end
+
+  private
+
+  def merge_if_valid(teleconsultation_params)
+    validator = Api::V4::TeleconsultationPayloadValidator.new(teleconsultation_params)
```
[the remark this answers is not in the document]
prabhanshuguptagit: How does cb7e6cc look?

app/controllers/api/v4/teleconsultations_controller.rb
prabhanshuguptagit: Disappeared in 7c7e037.

app/models/facility.rb
```diff
@@ -29,6 +29,7 @@ class Facility < ApplicationRecord
   has_many :patients, -> { distinct }, through: :encounters
   has_many :prescription_drugs
   has_many :appointments
+  has_many :teleconsultations
```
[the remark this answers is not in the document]
prabhanshuguptagit: Yes. Does it warrant being called `teleconsultation_requests`? I felt `teleconsultations` makes sense by itself since there is no concept of `teleconsultation_records` for a facility.

app/models/teleconsultation.rb
prabhanshuguptagit: We have this information in the request also since the nurse knows which MO the teleconsult is sent to.

app/models/teleconsultation.rb
prabhanshuguptagit: Yes.

app/controllers/api/v4/teleconsultations_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      {record: teleconsultation}
+    end
+  end
+
+  def teleconsultation_params
+    params.require(:teleconsultations).map do |teleconsultation_params|
+      teleconsultation_params.permit(
+        :id,
+        :patient_id,
+        :medical_officer_id,
```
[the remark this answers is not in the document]
prabhanshuguptagit: The app sends both these IDs as part of the payload. This is not for authorising the record, only for storage. Is there a reason to infer these from the headers instead?

app/models/teleconsultation.rb
prabhanshuguptagit: Currently we could use the `recorded_at` timestamp, Something like:
```
def self.incomplete
  where(`recorded_at IS NOT NULL`)
end
```

prabhanshuguptagit: When we add the `request_completed` flag:
```
def self.incomplete
  where(`recorded_at IS NOT NULL`).or(`request_completed = NULL`)
end 
```

app/controllers/api/v4/teleconsultations_controller.rb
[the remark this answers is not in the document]
harimohanraj89: Doesn't that mean as a nurse, I can send arbitrary MO ids, and spoof being a doctor? It's ok for the app to send them, but we shouldn't write them unless we're sure it's coming from the MO who performed the teleconsult. How can we do that? Maybe a `if request_user != mo_id { remove_the_param_from_payload }`?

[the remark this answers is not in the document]
harimohanraj89: Or even, `if request_user is an MO { write the mo_id }`?

app/models/teleconsultation.rb
harimohanraj89: I seeeeeee, so the MO ID sent in the request payload is "who has been asked to do the teleconsult".

Let me address [my previous comment](https://github.com/simpledotorg/simple-server/pull/1316/files#r490089933) here as well in one big bundle then.

Conceptually, there's a difference between the MO that's _requested_ to do the teleconsult, and the MO that _does_ the teleconsult.

If we want them to be the same, we should enforce that when we receive a `record` from the MO. This is a stricter model, not allowing, for example, MOs to forward these links to each other ("I'm busy, can you take this one").

In the more lenient model, the MO requested doesn't have to be the same as the MO who does the teleconsult. However, in that case, these are two separate concepts that we should capture distinctly. We can rename this to `requested_medical_officer` or something to make it very clear, and introduce a `medical_officer` association that stores which MO actually did the teleconsultation. Our API can do this in a fairly straightforward manner:

```ruby
unless teleconsult.completed?
  payload.merge(medical_officer_id: request_user.id)
end
```

This wont' require any API changes I believe. This will also allow us to make smart inferences about the usage of telemedicine - eg. how much link sharing is going on, _can_ we enforce the stricter model in the future, etc.

What do you think? Let me know if you wanna chat.

prabhanshuguptagit: Ah, that makes total sense. Yes, imo we should introduce a check here, instead of allowing `requested_medical_officer` and `medical_officer` to be separate things. Something like:

```
authorise record update if payload["medical_officer_id"] = request_user.id
```

Nurses can send requests for the same patient to many doctors in the first place, so I think we shouldn't support/instrument for the lenient model.

app/models/facility.rb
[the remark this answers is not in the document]
harimohanraj89: Yup, the name is fine. I was just clarifying the semantics :)

app/models/teleconsultation.rb
```diff
@@ -0,0 +1,42 @@
...
+  REQUEST_ATTRIBUTES = %w[requester_id facility_id requested_at request_completed]
+  RECORD_ATTRIBUTES = %w[
+    recorded_at
+    teleconsultation_type
+    patient_took_medicines
+    patient_consented
+    medical_officer_number
+    prescription_drugs
+  ]
+  TELECONSULTATION_ANSWERS = {
```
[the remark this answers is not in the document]
prabhanshuguptagit: Split'em up in 50b6b92. Couldn't figure out a good name for "the yes/no answer fields" 😅

app/controllers/api/v4/teleconsultations_controller.rb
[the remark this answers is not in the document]
prabhanshuguptagit: Update: 156cb2d

spec/factories/teleconsultations.rb
prabhanshuguptagit: These generic "invalid" and "updated" teleconsultation are helpful to have, yeah. It's nice to be able to quickly get an invalid payload in any spec. Other sync resources already have these little methods in factories, just made similar ones for teleconsultations.

app/models/teleconsultation.rb
prabhanshuguptagit: Added the `requested_medical_officer` in 0beb9d5

app/controllers/api/v4/teleconsultations_controller.rb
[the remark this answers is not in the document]
harimohanraj89: This is addressed below.

app/models/teleconsultation.rb
harimohanraj89: To close the loop, we decided to go with the lenient `requested_medical_officer` approach so we can first learn how big or small this problem is, before introducing constraints into the system. Additionally, constraining MOs on the serverside cannot be mirrored by a constraint in the app today, so it would lead to a misleading user experience where a seemingly successful teleconsultation would be rejected by the server.

app/validators/api/v4/teleconsultation_payload_validator.rb
```diff
@@ -9,10 +9,25 @@ class Api::V4::TeleconsultationPayloadValidator < Api::V3::PayloadValidator
     :deleted_at,
     :created_at
   )
+  attr_writer :request_user_id
 
   validate :validate_schema
+  validate :authorized_teleconsult_record, if: -> { record.present? }
+
+  def authorized_teleconsult_record
+    unless request_user.can_teleconsult?
```
harimohanraj89, line 18: Since we're using the forgiving `find_by` downstairs, we should assume that `request_user` can be nil. Can we do `request_user&.can_teleconsult?`

app/models/user.rb
```diff
@@ -46,6 +46,12 @@ def app_capabilities
   has_many :appointments
   has_many :medical_histories
   has_many :prescription_drugs
+  has_many :teleconsultation_requests,
```
harimohanraj89, line 49: Can we rename these associations to `requested_teleconsultations` and `recorded_teleconsultations`? That's more consistent with the idea that both associations return a collection of teleconsultations.

harimohanraj89 (approved):
This is lookin slick ✨ 

I've left some last suggestions for ya. In addition could we also:
* Update the PR desc
* Add teleconsultations to the [data dictionary](https://docs.google.com/spreadsheets/d/1LOn9IzfuQXog2agGbNphweMNTV27K84yctjPs6b7USI/edit#gid=0)?

app/services/merge_teleconsultation_service.rb
```diff
@@ -0,0 +1,36 @@
...
+    payload
+  end
+
+  def set_medical_officer(payload)
+    payload.except!(:medical_officer_id) if is_request? && Teleconsultation.find_by(id: payload[:id])
+    payload[:medical_officer_id] = @request_user.id if is_record?
+    payload
+  end
+
+  def is_request?
```
kitallis, line 29: No `is_*` necessary?

app/services/merge_teleconsultation_service.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def initialize(payload, request_user)
+    @payload = payload
+    @request_user = request_user
+  end
+
+  def merge
+    teleconsultation = @payload
+      .yield_self { |payload| set_requested_medical_officer(payload) }
```
kitallis, line 13: Really wish we were using Ruby 2.6, then we could use `then` over `yield_self`.

PR #1320, opened 2020-09-16
Telemed: Add fields to PrescriptionDrug

prabhanshuguptagit:
**Story card:**  https://app.clubhouse.io/simpledotorg/story/1121

## Because

We want to log a prescription drug's frequency and the duration for teleconsultations.
image

## This addresses

- Adds optional `frequency` and `duration_in_days` fields to PrescriptionDrug
- Adds an optional `teleconsultation_id` field to PrescriptionDrug


app/schema/api/v3/models.rb
```diff
@@ -252,7 +252,10 @@ def prescription_drug
          is_protocol_drug: {type: :boolean},
          is_deleted: {type: :boolean},
          patient_id: {"$ref" => "#/definitions/uuid"},
-         facility_id: {"$ref" => "#/definitions/uuid"}
+         facility_id: {"$ref" => "#/definitions/uuid"},
+         frequency: {type: :string, enum: PrescriptionDrug.frequencies.keys},
+         duration_in_days: {type: :integer},
```
harimohanraj89, line 257: Can frequency and duration be nullable? Or are we relying on the client to omit these params from the request for regular PDs? Can we confirm with the android team?

spec/factories/prescription_drugs.rb
```diff
@@ -8,6 +8,8 @@
     rxnorm_code { Faker::Code.npi }
     device_created_at { Time.current }
     device_updated_at { Time.current }
+    frequency { "OD" }
+    duration_in_days { 10 }
```
harimohanraj89, line 12: I would keep the regular factory _without_ these, and introduce a `for_teleconsultation` trait that adds them (as well as a teleconsultation ID, depending on which PRs get merged first.)

harimohanraj89 (changes requested):
Looks great. A couple note.

app/validators/api/v3/prescription_drug_payload_validator.rb
```diff
@@ -8,6 +8,9 @@ class Api::V3::PrescriptionDrugPayloadValidator < Api::V3::PayloadValidator
     :is_protocol_drug,
     :patient_id,
     :facility_id,
+    :frequency,
+    :duration_in_days,
+    :teleconsultation_id,
```
kitallis, line 13: I suppose we don't need to explicitly omit `teleconsultation_id` from the response payload for the older apps? Shall we confirm?

harimohanraj89: That did get me thinking, though. Should we authorize setting `teleconsultation_id` to just telemedicine MOs?

prabhanshuguptagit: Makes sense. A nurse cannot see the teleconsult prescription screen, so it should not be possible for them to add `teleconsultation_ids` but we could add a check nonetheless. There's another case worth considering:
- A user who had `can_teleconsult` true logs a teleconsult with some drugs
- User's `can_teleconsult` permission is revoked from server side
- User tries to sync the drugs from the teleconsult after and fails

harimohanraj89: I think that's ok, and the intended outcome, right?

prabhanshuguptagit: In this case we should allow the user to log those drugs, I think. The drugs _were_ prescribed to the patient and maybe it's better to collect that information than rejecting it.

harimohanraj89: Hmm, fair point. If it's not crazy to capture that auth in the code, go for it. If it is crazy, then we can ignore it.

prabhanshuguptagit: It isn't too hard to do this check (yet another payload validation), but I think we can rely on the app here for simplicity.

app/models/patient.rb
```diff
@@ -49,7 +49,7 @@ class Patient < ApplicationRecord
     -> { where(identifier_type: "simple_bp_passport").order(device_created_at: :desc) },
     class_name: "PatientBusinessIdentifier"
 
-  has_many :current_prescription_drugs, -> { where(is_deleted: false) }, class_name: "PrescriptionDrug"
+  has_many :current_prescription_drugs, -> { where(is_deleted: false).order(created_at: :desc) }, class_name: "PrescriptionDrug"
```
prabhanshuguptagit, line 52: Something about the field additions caused the ordering to become indeterministic.

harimohanraj89: Why does this association need to be ordered? I recall that ordered associations have caused us problems when we compose queries together in ActiveRecord.

app/validators/api/v3/prescription_drug_payload_validator.rb
harimohanraj89: Whoops, I may have been confusing. I meant if it's hard to tackle the edge case you mentioned, we can ignore the edge case. I think we should authorize the teleconsultation_id here regardless. What do you think

prabhanshuguptagit: Yeah, the authorisation can't be done basically without losing on the edge case information. I'd say we let this field be liberal.

harimohanraj89 (approved):
The ordering's not a big deal. :shipit:

app/models/patient.rb
prabhanshuguptagit: The current_prescription_drugs were being ordered weirdly (not by date or name) by Postgres, hence this to have some order (a spec broke).

> I recall that ordered associations have caused us problems when we compose queries together in ActiveRecord.
The `in_batches` in the exporter removes `order` scopes, however we call this method separately on a `patient` so it works for us.

PR #1321, opened 2020-09-17
More polishing and more regression clean-ups for permissions

kitallis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1249/user-permissions-regressions

## Because

Found a few power-user related edge-cases on various pages.

## This addresses

**Note:** Hide whitespace changes while going through the diff.

- [x] Ensure in `verify_params_v2` that a `power_user` can have blank `selected_facilities`
- [x] Edit screen for a `power_user` should hide facility accesses
- [x] `OrganizationsConstroller#index` should be accessible by organization managers and power users
- [x] Simplify the permissions for the nav-bar links
- [x] Add specs for `InvitationsController#create`
- [x] Improve `AdminsController` specs
- [x] Migrate Users section on Admin::FacilitiesController#show
- [x] Migrate Admin::FacilitiesController#upload


app/assets/javascripts/common/admins.js
```diff
@@ -2,13 +2,14 @@
 // elements
 //
 const ACCESS_LIST_INPUT_SELECTOR = "input.access-input"
+const ACCESS_LEVEL_ID = "access-level"
```
vkrmis, line 5: Why is this indirection necessary?

kitallis: Because it's being used in multiple places, once via JQuery (because of the selectpicker library) and once via JS. Since JS is very forgiving about errors, it's best to use a constant for these things to avoid regressions.

app/controllers/admin/organizations_controller.rb
```diff
@@ -7,7 +7,7 @@ class Admin::OrganizationsController < AdminController
 
   def index
     if current_admin.permissions_v2_enabled?
-      authorize_v2 { current_admin.power_user? }
+      authorize_v2 { current_admin.accessible_organizations(:manage).any? }
```
vkrmis, line 10: Updated the PR description to match this ^

app/controllers/email_authentications/invitations_controller.rb
```diff
@@ -123,4 +125,13 @@ def invite_params
   def temporary_password
     SecureRandom.base64(16)
   end
+
+  def validate_selected_facilities?
+    selected_facilities.blank? && user_params[:access_level] != User.access_levels[:power_user]
```
vkrmis, line 130: Why use `User.access_levels[:power_user]` instead of just `"power_user"`?

vkrmis (approved):
Looks good, just left some minor questions inline

app/controllers/email_authentications/invitations_controller.rb
kitallis: Just feel like we shouldn't rely on naked enum references around the code-base, it becomes hard to track down what the source of truth is and what it's actually supposed to be called.

vkrmis: Hmm, not sure how much of a difference there is in this usage though, we're just replacing `"power_user"` with `:power_user`. The burden of knowledge is exactly the same. Feel free to ignore this, I realize its pretty nitpicky.

app/views/admin/facilities/show.html.erb
```diff
@@ -64,61 +64,60 @@
...
+          </div>
+          <div>
+            <a href="tel:<%= user.phone_number %>" class="btn btn-sm btn-phone btn-outline-primary"><i class="fas fa-phone"></i> <%= user.phone_number %>
+            </a>
           </div>
-          <div class="col text-right">
+        </div>
+        <div class="col text-right">
+          <% if current_admin.permissions_v2_enabled? %>
+            <%= link_to 'Edit', edit_admin_user_path(user), class: "btn btn-sm btn-outline-primary" %>
```
vkrmis, line 96: Left out the `Delete` link. The old permissions implementation always returns `false` for `policy([:manage, :user, user]).destroy?`

app/views/admin/facilities/show.html.erb
```diff
@@ -64,61 +64,60 @@
...
-              %>
-              <div class="text-info mt-2 mb-2">Access: <%= user.sync_approval_status_reason %></div>
-            </div>
-            <div>
-              <a href="tel:<%= user.phone_number %>" class="btn btn-sm btn-phone btn-outline-primary"><i class="fas fa-phone"></i> <%= user.phone_number %>
-              </a>
-            </div>
+<div class="card">
+  <h2>Users</h2>
+  <% if @facility_users.present? %>
```
vkrmis, line 69: Assigning @facility_users in the controller instead of fetching directly here.

app/controllers/admin/facilities_controller.rb
```diff
@@ -77,6 +77,12 @@ def index
   end
 
   def show
+    @facility_users = if current_admin.permissions_v2_enabled?
+      current_admin.accessible_users(:manage)
```
kitallis, line 81:
```suggestion
      current_admin
        .accessible_users(:manage)      
```

PR #1322, opened 2020-09-17
Migrate approval notifier emails to also use new permissions

vkrmis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1245/migrate-user-approval-emails-to-incorporate-admins-with-new-permissions

## Because

The user approval emails depend on old permissions to decide who needs to get approval notification emails.

## This addresses

Adds admins with new permissions to the list of admins who need to see these emails.

### DIFF
--- a/app/mailers/approval_notifier_mailer.rb
+++ b/app/mailers/approval_notifier_mailer.rb
@@ -29,9 +29,9 @@ class ApprovalNotifierMailer < ApplicationMailer
 
   def supervisor_emails
     permissions_users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.facility_group).map(&:user)
-    accesses_users = User.admins.where(access_level: :manager)
+    accesses_users = User.admins.manager_access
       .select { |admin| admin.accessible_facilities(:manage).include?(user.facility) }
-      .reject { |admin| admin.accesses.map(&:resource_type).include?("Organization") }
+      .reject { |admin| admin.accesses.map(&:resource).include?(user.organization) }
 
     users = (permissions_users + accesses_users).uniq.compact
     users.map(&:email).join(",")
@@ -39,10 +39,10 @@ class ApprovalNotifierMailer < ApplicationMailer
 
   def organization_owner_emails
     permissions_users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.organization).map(&:user)
-    accesses_users = User.admins.where(access_level: :manager)
+    accesses_users = User.admins.manager_access
       .select { |admin|
       admin.accessible_facilities(:manage).include?(user.facility) &&
-        admin.accesses.map(&:resource_type).include?("Organization")
+        admin.accesses.map(&:resource).include?(user.organization)
     }
 
     users = (permissions_users + accesses_users).uniq.compact
@@ -51,7 +51,7 @@ class ApprovalNotifierMailer < ApplicationMailer
 
   def owner_emails
     permissions_users = UserPermission.where(permission_slug: :approve_health_workers, resource: nil).map(&:user)
-    accesses_users = User.admins.where(access_level: :power_user)
+    accesses_users = User.admins.power_user_access
 
     users = (permissions_users + accesses_users).uniq.compact
     users.map(&:email).join(",")
--- a/app/mailers/approval_notifier_mailer.rb
+++ b/app/mailers/approval_notifier_mailer.rb
@@ -23,18 +23,37 @@ class ApprovalNotifierMailer < ApplicationMailer
 
   private
 
+  # permissions_users are admins as per the old permissions system
+  # accesses_users are admins as per the new permissions system
+  # we need to send emails to the superset of the two till we flip everyone over to new permissions
+
   def supervisor_emails
-    users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.facility_group).map(&:user)
+    permissions_users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.facility_group).map(&:user)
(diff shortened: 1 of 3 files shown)


app/mailers/approval_notifier_mailer.rb
```diff
@@ -23,18 +23,37 @@ def reset_password_approval_email(user_id:)
...
   private
 
+  # permissions_users are admins as per the old permissions system
+  # accesses_users are admins as per the new permissions system
+  # we need to send emails to the superset of the two till we flip everyone over to new permissions
+
   def supervisor_emails
-    users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.facility_group).map(&:user)
+    permissions_users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.facility_group).map(&:user)
+    accesses_users = User.admins.where(access_level: :manager)
```
rsanheim, line 32: I think this can just be `User.admin.manager` the enum gives us a scope.

app/mailers/approval_notifier_mailer.rb
```diff
@@ -23,18 +23,37 @@ def reset_password_approval_email(user_id:)
...
+      .reject { |admin| admin.accesses.map(&:resource_type).include?("Organization") }
+
+    users = (permissions_users + accesses_users).uniq.compact
     users.map(&:email).join(",")
   end
 
   def organization_owner_emails
-    users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.organization).map(&:user)
+    permissions_users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.organization).map(&:user)
+    accesses_users = User.admins.where(access_level: :manager)
```
rsanheim, line 42: same here

app/mailers/approval_notifier_mailer.rb
```diff
@@ -23,18 +23,37 @@ def reset_password_approval_email(user_id:)
...
+  # permissions_users are admins as per the old permissions system
+  # accesses_users are admins as per the new permissions system
+  # we need to send emails to the superset of the two till we flip everyone over to new permissions
+
   def supervisor_emails
-    users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.facility_group).map(&:user)
+    permissions_users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.facility_group).map(&:user)
+    accesses_users = User.admins.where(access_level: :manager)
+      .select { |admin| admin.accessible_facilities(:manage).include?(user.facility) }
+      .reject { |admin| admin.accesses.map(&:resource_type).include?("Organization") }
```
rsanheim, line 34: This could incorrectly reject people who manage some other Org, right? This is an edge case that doesn't exist today, but wanted to point it out.

app/mailers/approval_notifier_mailer.rb
```diff
@@ -23,18 +23,37 @@ def reset_password_approval_email(user_id:)
...
     users.map(&:email).join(",")
   end
 
   def organization_owner_emails
-    users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.organization).map(&:user)
+    permissions_users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.organization).map(&:user)
+    accesses_users = User.admins.where(access_level: :manager)
+      .select { |admin|
+      admin.accessible_facilities(:manage).include?(user.facility) &&
+        admin.accesses.map(&:resource_type).include?("Organization")
```
rsanheim, line 45: I feel like for this check you really want to ensure the admin manages the Organization that the facility is in, and not just any Org.

app/mailers/approval_notifier_mailer.rb
vkrmis: Good catch, didn't consider that. Fixing.

app/mailers/approval_notifier_mailer.rb
```diff
@@ -23,18 +23,37 @@ def reset_password_approval_email(user_id:)
...
+  # permissions_users are admins as per the old permissions system
+  # accesses_users are admins as per the new permissions system
+  # we need to send emails to the superset of the two till we flip everyone over to new permissions
+
   def supervisor_emails
-    users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.facility_group).map(&:user)
+    permissions_users = UserPermission.where(permission_slug: :approve_health_workers, resource: user.facility_group).map(&:user)
+    accesses_users = User.admins.manager_access
+      .select { |admin| admin.accessible_facilities(:manage).include?(user.facility) }
+      .reject { |admin| admin.accesses.map(&:resource).include?(user.organization) }
```
kitallis, line 34: @vkrmis Is this different from,

```ruby
admin.accessible_organizations(:manage).where(id: user.organization).any?
```

...this?

app/mailers/approval_notifier_mailer.rb
kitallis: Shall we encode this as a test case?

kitallis (approved):
Looks good mostly, left 1 main comment.

app/mailers/approval_notifier_mailer.rb
vkrmis: Yes, `supervisor_emails` are `facility` and `facility_group` managers, so `accessible_organizations(:manage)` doesn't work. It would work for `organization_owner_emails` though, I'll change it there.

vkrmis:
> Shall we encode this as a test case?

I've added more data for admins that shouldn't get the emails. This should be covered by the existing cases since we're asserting `contains_exactly` for the `email.to` and `email.cc` fields.

PR #1325, opened 2020-09-18
Allow forced cache refresh on old reports

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1250/allow-old-reports-caches-to-be-force-refreshed

Allow old reports caches to be force-refreshed with a force_cache true query param, similar to reports.


kitallis (approved):
Locking gud.

Do you want to add a dumb test that checks that the request store is set if the query param is sent as a quick controller spec?

PR #1328, opened 2020-09-18
Reports v2 feature parity: Add graphics downloads 

rsanheim:
**Story card:** https://app.clubhouse.io/simpledotorg/story/456/create-download-dropdown

There is a bunch of code we could refactor here from the old reports, but I don't want to tackle that until the old controllers are gone.  Right now they are too tightly coupled to really clean this up nicely.


app/controllers/reports/regions_controller.rb
```diff
@@ -110,6 +111,27 @@ def download
     end
   end
 
+  def whatsapp_graphics
```
kitallis, line 114: Do you want to stick a quick controller spec for this action?

kitallis (approved):
lgtm, just added a note to add some controller specs for this new action. I've been noticing that we are lacking controller specs quite a bit across the board.

app/controllers/reports/regions_controller.rb
rsanheim: Yup done. 👍

PR #1331, opened 2020-09-21
Regions

rsanheim:
**Story card:** ch1455 https://app.clubhouse.io/simpledotorg/story/1455/create-region-schema-run-dry-run-of-region-backfill-in-prod

## Because

We want a clean(er) way to represent hierarchical geographic areas in our domain model.  Regions are a way to model Organization -> District (ie FacilityGroup) -> Block -> Facility, and they allow us to use the same object at each level for situations where that is what we want - for example with reporting.

## This addresses

Adds a Region model and RegionType model.  Adds a backfill with a dry_run mode, and this backfill **will not** run in production automatically via any migrations for now. We want to run it in dry run mode and then examine the log output to see what happens and what sort of data cleanup is necessary.

Future PRs will fix up the backfill to be more robust as needed, and then wire up the real backfill to run via a data migration.


app/models/region.rb
```diff
@@ -0,0 +1,51 @@
...
+    region_name = name || source.name
+    region = Region.new name: region_name, kind: kind
+    region.send :set_slug
+    region.source = source if source
+    region.path = "#{parent.path}.#{region.slug.tr("-", "_")}"
+    region.save!
+    region
+  end
+
+  def self.backfill!
```
rsanheim, line 28: Putting the backfill method here for easier testing.

app/models/region.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    instance = Region.create! name: "India", path: "India", kind: instance_kind
+    Organization.all.each do |org|
+      org_region = create_region_from(source: org, kind: org_kind, parent: instance)
+
+      org.facility_groups.each do |facility_group|
+        facility_group_region = create_region_from(source: facility_group, kind: facility_group_kind, parent: org_region)
+
+        facility_group.facilities.group_by(&:zone).each do |block_name, facilities|
+          block_region = create_region_from(name: block_name, kind: block_kind, parent: facility_group_region)
```
rsanheim, line 43: We probably don't want to create blocks until the data is cleaned up on prod, but I wanted to include it here to see how it would work.

app/models/region_kind.rb
```diff
@@ -0,0 +1,14 @@
+class RegionKind < ApplicationRecord
```
rsanheim, line 1: This could've been named `RegionType`, but using the name `type` tends to conflict w/ Rails STI.  This is basically where we define the region hiearchy for each Simple instance.

rsanheim (commented):
Notes for revieweres.

app/models/region.rb
```diff
@@ -0,0 +1,51 @@
...
+    region = Region.new name: region_name, kind: kind
+    region.send :set_slug
+    region.source = source if source
+    region.path = "#{parent.path}.#{region.slug.tr("-", "_")}"
+    region.save!
+    region
+  end
+
+  def self.backfill!
+    instance_kind = RegionKind.find_by!(name: "Instance")
```
vkrmis, line 29: Can we use something like `Root` or `RootRegion` instead of `Instance` for these top level `RegionKinds`? `Instance` doesn't really convey anything to me semantically.

app/models/region_kind.rb
vkrmis: Do `RegionKind`s have to be in the database? I'm wondering if we can just initialize them using some config hash-map instead?

Feels like creating a table and db rows for this might be overkill.

app/models/region.rb
rsanheim: Yeah we went back and forth on the name for the top level thing.  I'll change it to `Root`.

app/models/region_kind.rb
rsanheim: They do have to be in the DB, for a couple reasons:

* we need to know the canonical structure of Regions for building out the admin UI -- we could infer it from Region paths, but that would be error-prone and dangerous
* the names of Regions will vary between countries. For example, in India we have blocks, in Bangladesh we will have [upazila](https://en.wikipedia.org/wiki/Upazila).

rsanheim:
ch1455

app/models/region_type.rb
```diff
@@ -0,0 +1,14 @@
+class RegionType < ApplicationRecord
```
vkrmis, line 1: Definitely prefer `RegionType` to `RegionKind`, but wasn't there some concern about using `Type` in a model name?

app/services/region_backfill.rb
```diff
@@ -0,0 +1,127 @@
...
+  end
+
+  def create_regions
+    root_type = find_region_type("Root")
+    org_type = find_region_type("Organization")
+    facility_group_type = find_region_type("FacilityGroup")
+    block_type = find_region_type("Block")
+    facility_type = find_region_type("Facility")
+
+    current_country_name = CountryConfig.current[:name]
```
vkrmis, line 39: We may want to ensure this only can only be run in India (especially if `dry_run = false`), since the Region hierarchy is likely going to be different elsewhere.
Don't want to trigger this accidentally via a migration or manually in other countries.

vkrmis:
Mostly looks good, may want to be a little extra defensive and ensure this can only be triggered in India envs.

vkrmis (approved):

app/models/region_type.rb
harimohanraj89: `type` as a field name is problematic, but model should be fine.

app/models/region_type.rb
```diff
@@ -0,0 +1,14 @@
...
+  validates :name, presence: true, uniqueness: true
+  validates :path, presence: true
+
+  attr_accessor :parent
+
+  before_validation :set_path, if: :parent
+
+  def set_path
+    self.path = "#{parent.path}.#{name}"
+  end
```
harimohanraj89, line 13: Why do we need to do manual parent and path stuff in this class?

config/initializers/country_config.rb
```diff
@@ -55,6 +55,10 @@ class CountryConfig
   def self.for(abbreviation)
     CONFIGS[abbreviation]
   end
+
+  def self.current
+    Rails.application.config.country
```
harimohanraj89, line 60: Oh... this is... nice...

app/models/region.rb
```diff
@@ -0,0 +1,33 @@
+class Region < ApplicationRecord
+  ltree :path
+  extend FriendlyId
+  friendly_id :name, use: :slugged
+
+  validates :name, presence: true
+  validates :slug, presence: true, uniqueness: true
```
harimohanraj89, line 7: Don't think we can enforce uniqueness of the slug, can we? What happens if two blocks have the same name?

app/models/region.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  MAX_LABEL_LENGTH = 255
+
+  # A label is a sequence of alphanumeric characters and underscores.
+  # (In C locale the characters A-Za-z0-9_ are allowed).
+  # Labels must be less than 256 bytes long.
+  def name_to_path_label
+    name.gsub(/\W/, "_").slice(0, MAX_LABEL_LENGTH)
+  end
```
harimohanraj89, line 24: Where is this label used?

app/models/region.rb
rsanheim: Ugh.  True.

rsanheim: Er, no, they need to be unique actually.  FriendlyID should be able to handle name dupes and append a number to get a unique one.  We need them unique so that we can use them for URLs and such.

app/models/region.rb
rsanheim: Just for our backfill.

app/services/region_backfill.rb
rsanheim: 👍 added a fail safe check

PR #1332, opened 2020-09-22
Fix JS to be Uglifier compatible

kitallis:
**Story card:** N/A

## Because

Our deploys are breaking because Uglifier is not feature-complete and breaks for random things around certain ES6+ syntax.

## This addresses

In our case, the array destructuring syntax was breaking Uglifier (with harmony). Apparently a known [issue](https://github.com/lautis/uglifier/issues/175).


danySam (approved):
Looks good ;)

PR #1333, opened 2020-09-22
Block-level syncing

kitallis:
**Story card:** [ch1344](https://app.clubhouse.io/simpledotorg/story/1344/change-patient-sync-boundary-to-block-instead-of-facility-group)

## Because

Refer to this [PRD](https://docs.google.com/document/d/1Cflct0Y-44IRUVw_5-NptcnNSX1UgAPBiXqoXHq22io/edit#).

## This addresses

- Powered via the Region≈Block system
- Adds a `sync_region_id` to the process token
- Accepts a `HTTP_X_SYNC_REGION_ID` header to keep a track


rsanheim:
This looks promising for a quick fix - are you able to test it with production like data set sizes?

app/models/zone.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def initialize(facility)
+    @facility = facility
+  end
+
+  def blood_pressures
+    BloodPressure
+      .includes(encounter: {facility: :facility_group})
+      .where(facilities: {facility_group: facility_group})
+      .where(facilities: {zone: zone})
```
rsanheim, line 12: Would be good to run some EXPLAIN on this w/ facilities with lots of data, and also just get some comparison numbers between this PR and the production query to retrieve blood_pressures.

kitallis: Yeah, I'll definitely do that. 

But after some thought, I think we need to change the definition of how these BPs are calculated. I'm calculating them using encounters (the same way we do this for FGs currently). But I think this approach will not fly, because it can result in dangling BPs on the device.

If BPs are scoped by encounters, that means they are scoped by the facility of the BP. This is fine if the user is downloading all the patient data for the FG. But if we're syncing only block-level data, we should only sync those patients that the User can see in their block and then scope the BPs by patients on the device, rather than by facility.

If we don't do that, then this can happen:

```
Patient 1 registers in Facility 1
Patient 1 goes to Facility 1 in Block 1 to take BP1
Patient 1 goes to Facility 2 in Block 2 to take BP2
Patient 1 goes to Facility 2 in Block 2 to take BP3
Patient 1 goes to Facility 2 in Block 2 to take BP4

Nurse 1 from Facility 1 would be able to see the patient, and only BP1
Nurse 2 from Facility 2 can't see this patient because they are not registered here, but will sync BPs from BP2 to BP4.
```

This is quite non-ideal (even if the example is a little contrived).

It should instead be,

```
Patient 1 registers in Facility 1
Patient 1 goes to Facility 1 in Block 1 to take BP1
Patient 1 goes to Facility 2 in Block 2 to take BP2
Patient 1 goes to Facility 2 in Block 2 to take BP3
Patient 1 goes to Facility 2 in Block 2 to take BP4

Nurse 1 from Facility 1 should be able to see Patient 1 and all their BPs (regardless of where they were recorded)
Nurse 2 from Facility 2 can't see this patient
```

app/controllers/api/v3/sync_controller.rb
```diff
@@ -35,10 +35,6 @@ def disable_audit_logs?
     false
   end
 
-  def sync_api_toggled_on?
```
kitallis, line 38: Dead code.

spec/controllers/shared_examples/shared_spec_for_sync_controller.rb
```diff
@@ -183,83 +183,6 @@ def discard_patient(record)
   end
 end
 
-RSpec.shared_examples "a working V1 sync controller sending records" do
```
kitallis, line 186: Dead code.

app/models/region.rb
```diff
@@ -54,6 +54,23 @@ def log_payload
     attrs.symbolize_keys
   end
 
+  def self.sources
+    includes(:source).map(&:source)
+  end
+
+  def syncable_patients
```
kitallis, line 61: I'll bench this in isolation, but I think this will be a lot slower than before.

app/transformers/api/v3/facility_transformer.rb
```diff
@@ -10,8 +10,16 @@ def to_response(facility)
...
-               sync_region_id: facility.facility_group_id,
+               sync_region_id: sync_region_id(facility),
                protocol_id: facility.protocol.try(:id))
     end
+
+    def sync_region_id(facility)
+      if Flipper.enabled?(:region_level_sync)
+        facility.region.block.id
+      else
+        facility.facility_group.id
```
harimohanraj89, line 21: I guess this is inconsistent now 😅  Should the FG be `facility_group.region.id`? So we consistently send back region IDs for everything? I guess we can do that now that our data's all clean.

harimohanraj89: Actually, that might trigger resyncs on _all_ devices. Best to leave as is, and revisit after block-level sync is shipped to all (most) places.

harimohanraj89 (commented):
Some initial thoughts. Lookin' great!

app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -61,11 +64,27 @@ def current_facility_processed_since
...
+    def sync_region
+      if block_level_sync?
+        current_block
+      else
+        current_facility_group
+      end
+    end
+
+    def block_level_sync?
+      current_user.feature_enabled?(:region_level_sync) && process_token[:sync_region_id].present?
```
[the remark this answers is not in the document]
kitallis: Should we do both? Both levels of granularity will be super flexible. That way we can do Praveen **and** release incrementally.

app/controllers/api/v3/blood_sugars_controller.rb
```diff
@@ -13,7 +13,7 @@ def sync_to_user
   private
 
   def region_records
-    BloodSugar.syncable_to_region(current_sync_region).for_v3
+    BloodSugar.syncable_to_region(sync_region).for_v3
```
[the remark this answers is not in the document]
kitallis: Ah, yes, that'd work.

app/controllers/api_controller.rb
```diff
@@ -31,25 +31,33 @@ def current_facility_group
     current_user.facility.facility_group
   end
 
+  def current_block
+    current_user.facility.region.block
```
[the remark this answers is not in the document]
kitallis: Great catch.

app/models/region.rb
```diff
@@ -68,8 +85,8 @@ def log_payload
     # Generates has_many type of methods to fetch a region's descendants
     # e.g. organization.facilities
     define_method(region_type.pluralize) do
-      if ancestor_types(region_type).include?(self.region_type)
-        descendants.where(region_type: region_type)
+      if self_and_ancestor_types(region_type).include?(self.region_type)
+        self_and_descendants.where(region_type: region_type)
```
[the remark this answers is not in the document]
kitallis: Yup, precisely.

app/models/facility.rb
```diff
@@ -242,4 +242,8 @@ def teleconsultation_phone_numbers_with_isd
   def discardable?
     registered_patients.none? && blood_pressures.none? && blood_sugars.none? && appointments.none?
   end
+
+  def syncable_patients
+    registered_patients.with_discarded
```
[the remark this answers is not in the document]
kitallis: We can do the other magic things in `Facility`, `FacilityGroup` and `Organization` as well for consistency, but I've left them as they were before, since I figured that Region is effectively acting as the special-case sync class (for blocks).

Don't have a good answer to this.

app/transformers/api/v3/facility_transformer.rb
kitallis: I considered this. It shouldn't be `facility_group.region.id` both because it'll trigger a resync and more importantly, the reason why this check is here is for rollback-ability. We want to be able to get back to previous known good state trivially.

app/models/facility_group.rb
```diff
@@ -75,6 +76,10 @@ def cohort_analytics(period:, prev_periods:)
     query.call
   end
 
+  def syncable_patients
+    registered_patients.with_discarded
+  end
```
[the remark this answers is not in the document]
kitallis: I'm not sold on which one is better, I think getting syncable patients from the region makes sense and so does asking the patient to fetch the syncable patients for a region. They are both conceptually sound. Although, I seem to have a slight preference of how we had it before, purely because following the codepaths overall, there's less indirection.

Don't think there will be multiple queries no?

I think, `fg.syncable_patients` should present the same time-complexity of a query like `Patient.where(reg_facility: fg.facilities)`.

I have another idea, that I didn't explore to keep it simple, that I'll take a shot at and see if y'all like it.

app/controllers/concerns/api/v3/sync_to_user.rb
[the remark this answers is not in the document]
rsanheim: I think for the first release of this targeting _just_ the user seems fine, and limits the scope of things to worry about for a pretty critical change. 

We can easily expand to flagging 'region-like' actors in the future if we want to.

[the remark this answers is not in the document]
rsanheim: Why check on the `sync_region_id` being present as well? When this feature is enabled, we should be able to expect that id being in the payload, and if not we will raise and error and deal with it somehow (I hope) ?  If that is the case, we should make the feature flagged code work the same, and assume the token will exist.

app/models/region.rb
```diff
@@ -54,6 +54,23 @@ def log_payload
...
+    Patient
+      .with_discarded
+      .where(registration_facility: facilities.sources)
+      .or(Patient
+            .with_discarded
+            .where(assigned_facility: facilities.sources))
+      .union(Patient
+               .with_discarded
+               .joins(:appointments)
+               .where(appointments: {facility: facilities.sources}))
```
rsanheim, line 71: I'm confused about this query -- isn't this doing a lot more work than the previous query on Patient?  

```
  scope :syncable_to_region, ->(region) {
    with_discarded.where(registration_facility: region.facilities)
  }
```

We should keep the query the same if at all possible, there are enough other changes going on here at the same time. 😀

app/models/region.rb
```diff
@@ -80,11 +97,14 @@ def log_payload
...
   def initialize_path
     logger.info(class: self.class, msg: "got reparent_to: #{reparent_to.name}, going to initialize new path")
-    self.path = if reparent_to.path.present?
-      "#{reparent_to.path}.#{path_label}"
-    else
-      path_label
-    end
+
+    self.path =
+      if reparent_to.path.present?
```
rsanheim, line 102: this formatting change makes me 😢 - is this your local editor preferences? I don't think standard does this

kitallis: Haha. I did this intentionally. I cannot bear the sight of an incorrectly aligned `if-else` ladder, it just hasn't ever been that way in the history of programming and somehow standard forces you to align the 3/4 of the ladder to the left and the `if` next to the RHS making for a completely arbitrary looking conditional.

I'd much rather see the `if-else` ladder, read like how it's supposed to:

```ruby
if <c>
**things
else
**other_things
end
```
or

```ruby
a = if <c>
    **things
    else
    **other_things
    end
```

Not,

```ruby
thing = if <c>
**thing
else
**other_things
end
```

I agree with [this](https://stackoverflow.com/a/13863289).

app/models/region.rb
kitallis: Yeah, I think I might change the others to be this way also or conditionally enforce this query only for blocks.

We cannot launch block-level syncing without these additional changes (so I think it must be in this PR ideally), because the syncability is just not a wide enough net to be scalable. [ref](https://docs.google.com/document/d/1Cflct0Y-44IRUVw_5-NptcnNSX1UgAPBiXqoXHq22io/edit#heading=h.gegdriq7f9oe).

app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -32,10 +32,13 @@ def processed_until(records)
...
-      {current_facility_id: current_facility.id,
-       current_facility_processed_since: processed_until(current_facility_records) || current_facility_processed_since,
-       other_facilities_processed_since: processed_until(other_facility_records) || other_facilities_processed_since,
-       resync_token: resync_token}
+      {
+        current_facility_id: current_facility.id,
+        current_facility_processed_since: processed_until(current_facility_records) || current_facility_processed_since,
+        other_facilities_processed_since: processed_until(other_facility_records) || other_facilities_processed_since,
+        resync_token: current_resync_token,
+        sync_region_id: current_sync_region_id
```
prabhanshuguptagit, line 40:
```suggestion
        sync_region_id: current_sync_region_id || current_facility_group.id
```

prabhanshuguptagit, line 40:
```suggestion
        sync_region_id: sync_region.id
```

app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -61,11 +64,27 @@ def current_facility_processed_since
...
-      request.headers["HTTP_X_RESYNC_TOKEN"]
+    def resync_token_modified?
+      process_token[:resync_token] != current_resync_token
+    end
+
+    def sync_region_modified?
+      process_token[:sync_region_id] != current_sync_region_id
+    end
+
+    def sync_region
```
prabhanshuguptagit, line 78:
```
sync_region -> current_sync_region_id
```

app/controllers/concerns/api/v3/sync_to_user.rb
prabhanshuguptagit, line 87: Check if `requested_sync_region_id` is present instead of `process_token[:sync_region_id].present?`.

app/models/region.rb
```diff
@@ -54,6 +54,23 @@ def log_payload
     attrs.symbolize_keys
   end
 
+  def self.sources
+    includes(:source).map(&:source)
+  end
+
+  def syncable_patients
+    Patient
```
prabhanshuguptagit, line 62: This should support different patients for different region types:
FG -> registered patients
Facility -> registered patients
Block -> registered + assigned + appointed

app/controllers/concerns/api/v3/sync_to_user.rb
prabhanshuguptagit: Case where this change might break: 
- We enable block level sync for a user, they haven't resynced facilities yet. The `SYNC_REGION_ID` header is set to the facility group.
- App syncs and starts receiving blocks in the response. Since the `process_token.sync_region_id` is blank on first call, resync is triggered. Also `process_token.sync_region_id` is set to Block region's ID.
- They will send the same old `SYNC_REGION_ID` in the next request, we trigger a resync again.

If we mirror the header in `process_token.sync_region_id` (currently this way) there's a difference between what we are sending (block level data) and what the `process_token.sync_region_id` represents (facility group data).

app/controllers/api/v3/facilities_controller.rb
```diff
@@ -30,8 +30,10 @@ def transform_to_response(facility)
   end
 
   def response_process_token
-    {other_facilities_processed_since: processed_until(other_facility_records) || other_facilities_processed_since,
-     resync_token: resync_token}
+    {
+      other_facilities_processed_since: processed_until(other_facility_records) || other_facilities_processed_since,
+      resync_token: current_resync_token
```
prabhanshuguptagit, line 35: `resync_token` is a meaningful name already, I think we don't need the `current` suffix unlike the user and facility header.

app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -61,11 +64,27 @@ def current_facility_processed_since
...
     end
 
-    def resync_token
-      request.headers["HTTP_X_RESYNC_TOKEN"]
+    def resync_token_modified?
+      process_token[:resync_token] != current_resync_token
+    end
+
+    def sync_region_modified?
+      process_token[:sync_region_id] != current_sync_region_id
```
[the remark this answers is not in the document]
prabhanshuguptagit: Fixed:
```
def sync_region_modified?
  return if requested_sync_region_id.blank?
  process_token[:sync_region_id] != requested_sync_region_id
end
```

app/controllers/concerns/api/v3/sync_to_user.rb
prabhanshuguptagit: This has been addressed by making `current_sync_region` sensitive to the various possible `sync_region_id`'s.

app/models/region.rb
```diff
@@ -54,6 +54,36 @@ def log_payload
     attrs.symbolize_keys
   end
 
+  def syncable_patients
+    case region_type
```
vkrmis, line 58: Is this conditional meant to be a temporary fix?

app/controllers/api_controller.rb
```diff
@@ -31,12 +31,40 @@ def current_facility_group
...
   end
 
+  def current_sync_region
+    # This method selectively permits only FacilityGroup sync (via facility group ID)
+    # and block-level sync (via regions) and offers facility group as a safe fallback.
+    # Over time, the facility group ID support can be dropped and this method can
+    # allow other region types as well
+    return current_facility_group if requested_sync_region_id.blank?
+    return current_facility_group if requested_sync_region_id == current_facility_group.id
+    return current_block if block_level_sync?
```
harimohanraj89, line 41: This looks nice. If I understand correctly,
* It's the FG if the client has no idea what a sync group is
* It's the FG if block-level sync is on, but the client still asks for an FG's data (because it saved the FG ID when block-level sync was off, and hasn't updated its facility info yet)

app/controllers/api_controller.rb
```diff
@@ -31,12 +31,40 @@ def current_facility_group
...
+  def block_level_sync?
+    current_user.block_level_sync? && requested_sync_region_id == current_block.id
+  end
+
+  def current_block
+    # Fetching current block from current_facility is safer
+    # than fetching it by Region.find(requested_sync_region_id)
+    # since the requested_sync_region_id can be an FG id.
+    # This can be replaced in the future when facility group ID support is dropped.
+    current_facility.region.block_region
```
harimohanraj89, line 55: I'd posit we stick to getting the block from the current facility permanently. Regardless of what the user is asking to sync, the current facility (which comes from the X-Facility-ID header, right?) is where the user is.

app/models/facility.rb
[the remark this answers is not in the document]
harimohanraj89: Can you help me understand if we still need to do some updating here? The code here is inconsistent with our intended "patients in a block" definition, right?

app/models/region.rb
harimohanraj89: Why not apply the block definition to all regions? Is performance a concern?

app/models/facility.rb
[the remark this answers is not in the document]
harimohanraj89: Update: I see the `region` does the heavy lifting now. We can... delete these methods from facility and facility group?

app/models/user.rb
```diff
@@ -236,6 +236,10 @@ def power_user?
     power_user_access? && email_authentication.present?
   end
 
+  def block_level_sync?
+    Flipper.enabled?(:regions_prep) && feature_enabled?(:region_level_sync)
```
harimohanraj89, line 240: Nice, dependent on the earlier feature flag.

public/manifest/production.json
```diff
@@ -2,21 +2,15 @@
...
-    },
-    {
-      "country_code": "ET",
-      "endpoint": "https://api.et.simple.org/api/",
-      "display_name": "Ethiopia",
-      "isd_code": "251"
+      "country_code": "IN",
+      "endpoint": "https://api-production-perf-2.simple.org/api/",
+      "display_name": "II Performance",
+      "isd_code": "91"
```
harimohanraj89, line 13: 🛑  We have to revert these manifest changes 🛑

harimohanraj89:
Looking pretty darn good overall. I've left my last questions/thoughts and [one blocker](https://github.com/simpledotorg/simple-server/pull/1333#discussion_r537493860).

After that, looks good for approval.

app/models/facility.rb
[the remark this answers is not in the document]
harimohanraj89: Hmm, in that case, it seems to me like we should remove them for a couple of reasons:
* It seems overly defensive to have these methods in here
* Regions are carrying their own weight quite well, minimizing the chances that a facility or facility group will need to be used

What do you think?

spec/controllers/api/v3/blood_pressures_controller_spec.rb
```diff
@@ -1,36 +1,33 @@
 require "rails_helper"
 
 RSpec.describe Api::V3::BloodPressuresController, type: :controller do
+  before { Flipper.enable(:regions_prep) }
```
rsanheim, line 4: For this sort of wide-reaching feature flag, we will want to be testing **both** code paths for awhile to ensure we are able to turn the flag off for an easy revert after shipping.  It wouldn't surprise me to see the feature flagged code live on into January given holidays and also the fact that mobile rollout and resync takes time.

Something like this will work for the entire controller spec:

```ruby
[:enable, :disable].each do |flag_toggle|
  RSpec.describe "Api::V3::BloodPressuresController with regions_prep #{flag_toggle}", type: :controller do
    before { Flipper.public_send(flag_toggle, :regions_prep) }
...
```

spec/controllers/api/v3/blood_sugars_controller_spec.rb
```diff
@@ -1,36 +1,33 @@
 require "rails_helper"
 
 RSpec.describe Api::V3::BloodSugarsController, type: :controller do
+  before { Flipper.enable(:regions_prep) }
```
rsanheim, line 4: Need to test for both states of the flag here like above.

spec/controllers/api/v3/encounters_controller_spec.rb
```diff
@@ -1,16 +1,15 @@
 require "rails_helper"
 
 RSpec.describe Api::V3::EncountersController, type: :controller do
+  before { Flipper.enable(:regions_prep) }
```
rsanheim, line 4: Need to test for both states of the flag here as well.

spec/controllers/api/v3/facilities_controller_spec.rb
```diff
@@ -1,8 +1,8 @@
 require "rails_helper"
 
 RSpec.describe Api::V3::FacilitiesController, type: :controller do
+  before { Flipper.enable(:regions_prep) }
```
rsanheim, line 4: same as above, or alternatively maybe a good approach is to have sub contexts in side the controller, one that enables the flag and one that disables?  Maybe it depends on the specifics of the context and setup required here.

spec/controllers/shared_examples/shared_spec_for_sync_controller.rb
```diff
@@ -357,6 +280,208 @@ def discard_patient(record)
...
+  let!(:patient_in_request_facility) { create(:patient, :without_medical_history, registration_facility: request_facility) }
+  let!(:patient_in_same_block) { create(:patient, :without_medical_history, registration_facility: facility_in_same_block) }
+  let!(:patient_assigned_to_block) { create(:patient, :without_medical_history, assigned_facility: facility_in_same_block) }
+  let!(:patient_with_appointment_in_block) {
+    create(:patient, :without_medical_history)
+      .yield_self { |patient| create(:appointment, patient: patient, facility: facility_in_same_block) }
+      .yield_self { |appointment| appointment.patient }
+  }
+  let!(:patient_in_other_block) { create(:patient, :without_medical_history, registration_facility: facility_in_other_block) }
+  let!(:patient_in_other_facility_group) { create(:patient, :without_medical_history, registration_facility: facility_in_other_group) }
```
rsanheim, line 308: When we have this much dependent state happening via `let!`, it's really hard to track what happens for each individual test and what the actual defendant objects at play are.  Also, adding any new specs for this context will end up w/ all this overhead, even if the new specs don't really need all the state.

I realize there is a lot of setup necessary here to have good tests, and these are critical code paths to be tested. My suggestion is that for future PRs we work towards inlining setup or using plain ruby methods to consolidate setup and then call it normally. We don't need tests to be DRY, in fact they are going to be much more readable and useful for us in the future if we have more things inline.

The Thoughtbot crew have written a bunch on the dangers of let: 
* https://thoughtbot.com/blog/lets-not
* https://thoughtbot.com/blog/my-issues-with-let

I also noticed the [RSpec docs themselves](https://www.rubydoc.info/github/rspec/rspec-core/RSpec%2FCore%2FMemoizedHelpers%2FClassMethods%3Alet) recommend against heavy usage of `let` like this:

> Note: let _can_ enhance readability when used sparingly (1,2, or maybe 3 declarations) in any given example group, but that can quickly degrade with overuse. YMMV.

Please consider this feedback non-blocking, but we should start to refactor away from this sort of `let` overuse in the future.

app/controllers/api_controller.rb
rsanheim: Are we still getting a new header for the `region_type` as well as the `region_id` ? I thought I saw that in earlier discussions.

If so that could allow us to be safer when region sync is enabled as we could constraint the lookup by both:

```
current_user.accessible_regions.where(id: requested_sync_region_id, region_type: requested_sync_region_type)
```

app/models/facility.rb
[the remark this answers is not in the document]
rsanheim: Agree w/ @harimohanraj89, removing this kind of duplicated code and letting Regions handle it was one of the big reasons for them. 

Lets ax any code that isn't being called. (blocking change)

app/models/facility_group.rb
```diff
@@ -138,6 +138,10 @@ def cohort_analytics(period:, prev_periods:)
     query.call
   end
 
+  def syncable_patients
+    registered_patients.with_discarded
```
rsanheim, line 142: 🪓 Don't think this is called, right?  Everything will go through Region or Patient I believe.

app/models/organization.rb
```diff
@@ -50,4 +51,8 @@ def districts
   def discardable?
     facility_groups.none? && users.none? && appointments.none?
   end
+
+  def syncable_patients
```
rsanheim, line 55: 🪓 

If we are syncing all patients in an Org we are gonna have a bad time. 😁

app/models/patient.rb
```diff
@@ -56,12 +56,10 @@ class Patient < ApplicationRecord
 
   attribute :call_result, :string
 
-  scope :search_by_address,
-    ->(term) { joins(:address).merge(Address.search_by_street_or_village(term)) }
-
+  scope :syncable_to_region, ->(region) { region.syncable_patients }
```
rsanheim, line 59: image

app/models/region.rb
```diff
@@ -54,6 +54,36 @@ def log_payload
...
+          .union(assigned_patients)
+          .union(appointed_patients)
+      else
+        registered_patients
+    end
+  end
+
+  def registered_patients
+    Patient
+      .with_discarded
```
rsanheim, line 70: Callers should be responsible for adding `with_discarded` if that is really what they want...its  dangerous to make this the default implementation as it bypasses our soft-delete.

(blocking)

app/models/region.rb
```diff
@@ -54,6 +54,36 @@ def log_payload
...
+      .where(registration_facility: facility_regions.pluck(:source_id))
+  end
+
+  def assigned_patients
+    Patient
+      .with_discarded
```
rsanheim, line 76: Same here, adding `with_discarded` would break RegionReports, as we definitely don't want soft deleted patients there.

See https://github.com/simpledotorg/simple-server/pull/1477/files#diff-7a3a5da2640933d23aba82fb308d4cb9827e42c27ed05ba979cdb9aa6b849637R51

app/models/region.rb
rsanheim: I just use whatever standard auto formats to honestly. That is the point of standard - to not have to manually adjust formatting at all, ever, so we don't even have these discussions. 😁 😁

rsanheim (changes requested):
Really glad to see this so close. 🤩 

I left some blocking changes, primarily around removing any code that is not necessary and also minimizing changes that aren't directly related to getting block sync live.  I tried to be explicit about what things are included under the scope of 'request changes' vs not.

The testing issue I raise could be done in a follow up PR, but if we go that route we really need to do it *right after* this lands to avoid any regression.

app/controllers/api_controller.rb
prabhanshuguptagit: We aren't getting a separate header for `region_type`. In fact this was one of the reasons we went with `Region`, since it allows us to drop this header 😄 .

app/models/region.rb
kitallis: I mentioned this [earlier](https://github.com/simpledotorg/simple-server/pull/1333#discussion_r522784347) and it might have gotten lost in the thread. I think this is the whole purpose of the PR, the rest is just structural changes. 

Block-level sync is not useful if we're just syncing registered patients. We **cannot** launch block-level sync without the neighboring queries.

app/models/facility.rb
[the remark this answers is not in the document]
prabhanshuguptagit: Agree. Will ax these👍

app/models/facility_group.rb
prabhanshuguptagit: This is called [by `Patient` like so](https://github.com/simpledotorg/simple-server/pull/1333/files/59aab26c1b0ac4ad3791b388c0c2af76b21da996#diff-94dcc014f62f0d9b6084ad98f82f74cd40c86b029ef1691914bdbc960c2f42dcR59). This method is how we keep sync backwards compatible.

app/models/organization.rb
prabhanshuguptagit: Addressed in 8fdfb24.

spec/controllers/shared_examples/shared_spec_for_sync_controller.rb
kitallis: I think @harimohanraj89 and I wrote some of these initially and I'm aware of the dangers of `let`. But for these specs, I think it actually does the opposite when you inline, putting this much setup data in each test makes the plethora of cases unreadable and it's very hard to figure out what the test case is actually doing and/or if there's any difference in the setup data for each of them or not – all this is noise for understanding these cases.

These test cases are really really important, we mess this up and our syncs go for a toss. So we wanted to make them as short and easy to read as possible. And more importantly consistent across the board so that one sync controller doesn't behave differently from the rest.

None of this applies in our case, I think:

Screenshot 2020-12-08 at 11 42 58 AM

* none of it is alien, the tests don't work if you remove any one
* all this setup is necessary for *all* tests in the block-level sync block
* they are as slow as necessary

We could move this to a function, but that doesn't really solve the problem, it just hides away code from one place to another. The setup complexity remains. I generally prefer in-lining seed data, but 90% of the tests we write are fairly non-critical, unlike this.

In fact this silo-ing of the `let`s was so helpful that it literally paved the way for our manual QA because we translated the seed data nearly 1-to-1, [ref](https://docs.google.com/document/d/1Cflct0Y-44IRUVw_5-NptcnNSX1UgAPBiXqoXHq22io/edit#heading=h.5zi8c8m688p4).

app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -61,11 +64,18 @@ def current_facility_processed_since
...
+
+    def resync_token_modified?
       process_token[:resync_token] != resync_token
     end
 
-    def resync_token
-      request.headers["HTTP_X_RESYNC_TOKEN"]
+    def sync_region_modified?
+      return if requested_sync_region_id.blank?
+      process_token[:sync_region_id] != requested_sync_region_id
```
harimohanraj89, line 78: Would a datadog metric here be useful to gauge how often nurses switch blocks?

app/models/region.rb
kitallis: Summary:

* We need this, cannot launch without it
* If we start doing this for all other regions, we shouldn't blindly do it every everything else, the sync boundaries for blocks happen to be quite atypical, this may not be the case for others. We should discuss and then implement. Cordoning this off is a safer bet.
* We should also think twice before doing it for bigger regions because of perf reasons, this will fly for blocks because you're starting very narrow, but it won't work for States, that's too much data.

harimohanraj89: > We cannot launch block-level sync without the neighboring queries.

Agree. To clarify, my comment here was to ask if we can apply the "unioned" definition across the board, but the conditional here makes the change _more_ surgical. It preserves current behavior for facility groups, and only changes the sync definition for blocks. Let's leave as is.

app/models/region.rb
kitallis: Good catch!

app/controllers/concerns/api/v3/sync_to_user.rb
prabhanshuguptagit: Yes! A follow up PR that adds metrics is in the works 😄

app/models/region.rb
prabhanshuguptagit: Addressed in 8911b319.

public/manifest/production.json
```diff
@@ -2,21 +2,15 @@
   "v1": [
     {
       "country_code": "IN",
-      "endpoint": "https://api.simple.org/api/",
```
prabhanshuguptagit, line 5: Revert manifest changes

prabhanshuguptagit (changes requested):

spec/controllers/api/v3/blood_pressures_controller_spec.rb
kitallis: Thanks @rsanheim! These specs are a bit tricky since they only partially depends on the flag. We've been thinking about the `regions_prep` flag as something that is extremely short-term. In fact, we're looking to raise a PR to remove it this week itself. The `region_level_sync` flag is what will stick around long-term (across months), this `regions_prep` was a super transitionary flag, the quicker we get rid of the easier it is for everyone.

This is largely why we've avoided massive structural spec changes just to support this flag (more or less everywhere), the mental overhead and the overall cost just doesn't justify the upkeep.

However, using your above snippet (and without too much headache) we managed to get this working with a slight modification to avoid running the region specs when the `regions_prep` isn't enabled. It's covered [here](https://github.com/simpledotorg/simple-server/pull/1333/commits/f343427699cf80a265bfc15ae0902876e78cd5d1?diff=unified&w=1). However, now our specs take 10m to run on CI :)

Let us know if that's cool.

spec/controllers/api/v3/blood_sugars_controller_spec.rb
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/1333/commits/f343427699cf80a265bfc15ae0902876e78cd5d1?diff=unified&w=1.

spec/controllers/api/v3/facilities_controller_spec.rb
kitallis: Addressed in https://github.com/simpledotorg/simple-server/pull/1333/commits/f343427699cf80a265bfc15ae0902876e78cd5d1?diff=unified&w=1

prabhanshuguptagit (approved):

spec/controllers/shared_examples/shared_spec_for_sync_controller.rb
rsanheim: > We could move this to a function, but that doesn't really solve the problem, it just hides away code from one place to another. The setup complexity remains.

My point about moving it to a function and calling it from specs was that it would be easier to follow in the future for whoever has to maintain this in two years.  I agree totally that all the setup is necessary and these tests are critical, I just want to be able to understand them quickly later. :)

rsanheim (approved):
Approving to unblock, excited to see this get out. thanks for the thoughtful responses to my feedback. 😀

PR #1339, opened 2020-09-23
Telemed: Fix facility medical officers API 

prabhanshuguptagit:
**Story card:** -

## Because

The `v4/facility_medical_officers` endpoint is sending resources with an incorrect `deleted_at`. The `deleted_at` should always be `null` since this is an artificial resource that will always be synced.
image

[Slack ref](https://simpledotorg.slack.com/archives/CV9V1DWLV/p1600850974004900)

## This addresses

This fixes the `deleted_at` to be `null`.


vkrmis (approved):
Looks good

harimohanraj89 (approved):

PR #1344, opened 2020-09-24
Bug: Total count not updating when user has only facility-level access tree

danySam:
**Story card:** <story link>

## Because

Total count not updating when the user has only facility-level access tree

## This addresses

Fixed by adding a guard clause to check if checked element has a parent


kitallis (approved):
Makes sense to me.

PR #1345, opened 2020-09-24
Fix UI regression in ProtocolDrugs creation

vkrmis:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1315/ui-regression-in-protocoldrug-creation

## Because

A UI regression with the new permissions hides the button to create new protocol drugs from Organization Managers

## This addresses

- Fixes the UI regression
- Bypasses `authorize_2` checks for power_users

### DIFF
--- a/app/views/admin/protocol_drugs/index.html.erb
+++ b/app/views/admin/protocol_drugs/index.html.erb
@@ -37,7 +37,7 @@
 <br>
 
 <% if current_admin.permissions_v2_enabled? %>
-  <% if current_admin.accessible_organizations(:manage).any? %>
+  <% if current_admin.accessible_organizations(:manage).any? || current_admin.power_user? %>
     <%= link_to 'New protocol drug', new_admin_protocol_drug_path %>
   <% end %>
 <% else %>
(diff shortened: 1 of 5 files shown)


app/views/admin/protocol_drugs/index.html.erb
```diff
@@ -37,7 +37,7 @@
 <br>
 
 <% if current_admin.permissions_v2_enabled? %>
-  <% if current_admin.accessible_organizations(:manage).any? %>
+  <% if current_admin.accessible_organizations(:manage).any? || current_admin.power_user? %>
```
harimohanraj89, line 40: Shouldn't `accessible_organizations` return a non-empty set of orgs for power users?

vkrmis: Yes, with one edge case: when no organizations exist in the system yet. Adding this to UI checks so the application is usable at initial setup.

harimohanraj89: Ah the blank slate problem. Neat. 👍

harimohanraj89 (approved):

PR #1352, opened 2020-09-25
Rename "Assigned facility" to "Preferred facility" in linelists

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/1329/rename-assigned-facility-to-preferred-facility-in-the-line-lists

## Because

We use "Preferred" in all user-facing touchpoints, so our line lists should be consistent.


prabhanshuguptagit (approved):
🚀

PR #1354, opened 2020-09-25
Add security env to cap and manifest

prabhanshuguptagit:
**Story card:** <story link>

## Because

We need to spin up a security env for the CyRAACS audit.

## This addresses

This adds the security env created in https://github.com/simpledotorg/deployment/pull/260 to cap and app manifest.


harimohanraj89 (approved):
Looks like a clean revert of the important bits from https://github.com/simpledotorg/simple-server/pull/1292

PR #1355, opened 2020-09-25
Ignore breaking data migration

prabhanshuguptagit:
**Story card:** NA

## Because

We found an old one-time data migration that breaks on a fresh `db:migrate`. This happens due to a conflict between the `users` schema in the data migration and newly added validations in the User model.

## This addresses

This ignores the one time migration to fix `db:migrate`. This is safe to ignore since it has been already run on all envs.

### DIFF
--- a/db/data/20200528063206_set_patient_soft_deletion_info.rb
+++ b/db/data/20200528063206_set_patient_soft_deletion_info.rb
@@ -1,15 +1,19 @@
 class SetPatientSoftDeletionInfo < ActiveRecord::Migration[5.2]
   def up
-    bot_user = User.find_or_create_by!(full_name: "Data Migration Bot",
-                                       sync_approval_status: "denied",
-                                       sync_approval_status_reason: "Bot user doesn't require sync") { |user|
-      user.device_created_at = Time.current
-      user.device_updated_at = Time.current
-    }
-
-    Patient.with_discarded.discarded.in_batches(of: 1_000) do |batch|
-      batch.update_all(deleted_by_user_id: bot_user.id, deleted_reason: "unknown")
-    end
+    # This migration conflicts with the validations added on User later
+    # (teleconsultation_phone_number). Commenting this out since it's a
+    # one time migration that's already been run on all envs.
+    #
+    # bot_user = User.find_or_create_by!(full_name: "Data Migration Bot",
+    #                                    sync_approval_status: "denied",
+    #                                    sync_approval_status_reason: "Bot user doesn't require sync") { |user|
+    #   user.device_created_at = Time.current
+    #   user.device_updated_at = Time.current
+    # }
+    #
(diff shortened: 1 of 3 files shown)


vkrmis (approved):

db/schema.rb
```diff
@@ -142,9 +142,7 @@
     t.index ["user_id"], name: "index_communications_on_user_id"
   end
 
-  create_table "data_migrations", id: false, force: :cascade do |t|
-    t.string "version", null: false
-    t.index ["version"], name: "unique_data_migrations", unique: true
```
harimohanraj89, line 147: Why did this disappear?

prabhanshuguptagit: This seems to be something introduced by the rails upgrade in 778fbf6f

db/data/20200528063206_set_patient_soft_deletion_info.rb
```diff
@@ -1,15 +1,19 @@
...
+    # bot_user = User.find_or_create_by!(full_name: "Data Migration Bot",
+    #                                    sync_approval_status: "denied",
+    #                                    sync_approval_status_reason: "Bot user doesn't require sync") { |user|
+    #   user.device_created_at = Time.current
+    #   user.device_updated_at = Time.current
+    # }
+    #
+    # Patient.with_discarded.discarded.in_batches(of: 1_000) do |batch|
+    #   batch.update_all(deleted_by_user_id: bot_user.id, deleted_reason: "unknown")
+    # end
```
harimohanraj89, line 16: Could we instead rescue the exception, skip it only when it's problematic?

```ruby
def up
  # try to do the thing
rescue ActiveRecord::RecordInvalid 
  Rails.logger.info "This data migration could not be run. Skipping."
end
```

harimohanraj89 (changes requested):
Some thoughts!

db/schema.rb
harimohanraj89: Weird. So is this column and index in our db or not? Disappearing from the schema without a corresponding migration means our schema and DB are out of sync, right?

db/data/20200528063206_set_patient_soft_deletion_info.rb
prabhanshuguptagit: At the point in time this migration is run, this will always fail because the subsequent migration that adds the columns hasn't been added yet.

harimohanraj89: Hmm yeah fair. Yeah this is clearer 👍

db/schema.rb
prabhanshuguptagit: Just the syntax for this migration seems to have changed. The column and index is still present (line below).

harimohanraj89 (approved):

PR #1372, opened 2020-09-29
Ranked facilities

timcheadle:
**Story card:** [ch1419](https://app.clubhouse.io/simpledotorg/story/1419/add-ranked-facilities-controller-and-views)

## Because

We want to provide CVHOs with a summary of which facilities are performing well and which ones need help relative to each other.

## This addresses

Build a controller action and views to display `Ranked facilities` under the `Home` section.

### Known issues:

- This adds OPD load estimation based on `facility_size`. We should look at production averages to determine better estimates than what I wrote


app/models/reports/performance_score.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def registrations_score
+      REGISTRATIONS_SCORE_WEIGHT * registrations_rate
+    end
+
+    def registrations_rate
+      registrations = @result.registrations.values.last
+
+      # If the target is zero, return 100% if any registrations occurred
+      if target_registrations <= 0
```
rsanheim, line 38: I'm hitting an error locally related to this part of the scoring:

```
[2020-10-02T14:56:28.514-05:00] FATAL: NoMethodError (undefined method `/' for nil:NilClass):

[2020-10-02T14:56:28.515-05:00] FATAL: app/models/reports/performance_score.rb:42:in `registrations_rate'
app/models/reports/performance_score.rb:31:in `registrations_score'
app/models/reports/performance_score.rb:15:in `overall_score'
app/controllers/my_facilities/ranked_facilities_controller.rb:41:in `block in show'
app/controllers/my_facilities/ranked_facilities_controller.rb:41:in `sort_by'
app/controllers/my_facilities/ranked_facilities_controller.rb:41:in `show'
app/controllers/my_facilities/ranked_facilities_controller.rb:65:in `block in set_time_zone'
app/controllers/my_facilities/ranked_facilities_controller.rb:65:in `set_time_zone'
app/controllers/admin_controller.rb:24:in `switch_locale'
```

I'm guessing you need to add a check for nil `registrations`.

rsanheim (approved):
Approving to unblock, but note that there is one edge case that blew up locally for me.  Could be related to empty facilities I've created in dev, but I imagine that could also happen in sandbox / prod.

If for some reason the cache isnt' warmed, this will basically do that for us, right? 😏 It needs to basically load every single facilities data.

app/controllers/my_facilities/ranked_facilities_controller.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+
+class MyFacilities::RankedFacilitiesController < AdminController
```
kitallis, line 3: What prompted a new controller for this? There's only one action and it seems like it could fit well within `MyFacilitiesController` no? It's duplicating a bunch of setup code, which would mean changing it in two places whenever we do so.

app/models/reports/performance_score.rb
```diff
@@ -0,0 +1,49 @@
+module Reports
+  class PerformanceScore
+    CONTROL_SCORE_WEIGHT = 0.5
+    VISITS_SCORE_WEIGHT = 0.3
+    REGISTRATIONS_SCORE_WEIGHT = 0.2
+    TARGET_REGISTRATION_RATE = 0.1
+
+    def initialize(region:, result:)
```
kitallis, line 8: Should we call this `reports_result`? Took me a while to figure out what `result` this is.

app/models/reports/performance_score.rb
```diff
@@ -0,0 +1,49 @@
...
+      @result = result
+    end
+
+    def overall_score
+      # max score: 100
+      (control_score + visits_score + registrations_score)
+    end
+
+    def control_score
+      CONTROL_SCORE_WEIGHT * @result.controlled_patients_rate.values.last
```
kitallis, line 19: Could you explain the significance of `controlled_patients_rate.values.last`? What does it mean to pull the last `controlled_patients_rate` value out?

Also, it seems like the `Result` object has a method called [last_value](https://github.com/simpledotorg/simple-server/blob/master/app/models/reports/result.rb#L51) which seems like it does the same thing  that you could use (although I don't understand the meaning behind that either).

app/controllers/my_facilities/ranked_facilities_controller.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
```
kitallis, line 1: Should we add some controller specs for this as well?

kitallis (approved):
Approved to avoid blocking, left some comments inline.

app/models/reports/performance_score.rb
timcheadle: Added code to handle `nil` for all `Result` used here

app/controllers/my_facilities/ranked_facilities_controller.rb
timcheadle: It's primarily because I'm using `Reports::RegionService` which requires different setup than the My Facilities queries. I wanted to use exactly the same data that we're showing in reports.

app/models/reports/performance_score.rb
timcheadle: I definitely don't want to confuse; renamed it to `reports_result`

app/models/reports/performance_score.rb
timcheadle: the `Result` object contains a hash of values keyed by dates, so you can get the control rate for May 2019, for example. `values.last` returns the most recent value. I suppose switching to a more consistent method may make sense for this (to cover for weirdness in hash sorting, for example). What say you, @rsanheim? Regardless, that's a separate PR.

app/controllers/my_facilities/ranked_facilities_controller.rb
timcheadle: We're iterating on this so I skipped controller/integration specs for now.

PR #1374, opened 2020-09-30
Add a sync_region_id to Facility API

prabhanshuguptagit:
**Story card:** [ch1414](https://app.clubhouse.io/simpledotorg/story/1414/introduce-a-sync-group-concept-in-the-api)

## Because

We want to introduce a “sync group” concept in the API that the app can use to determine whether to trigger a resync or not when a user switches facilities. For now, populate the facility's block name in there.

## This addresses

This adds a `sync_group_id` to `v3/facility` sync resource.

[Updated API docs](https://simple-review-pr-1374.herokuapp.com/api-docs#tag/facility)


app/transformers/api/v3/facility_transformer.rb
```diff
@@ -10,6 +10,7 @@ def to_response(facility)
           "teleconsultation_phone_numbers")
         .merge(config: {enable_diabetes_management: facility.enable_diabetes_management,
                         enable_teleconsultation: facility.enable_teleconsultation},
+               sync_group_id: facility.zone,
```
harimohanraj89, line 13: Let's release this PR with `sync_group_id: facility.facility_group_id` so we can transparently add this field to the API right away.

kitallis (approved):
lgtm

PR #1375, opened 2020-09-30
Patient centric sync

harimohanraj89:
**Story card:** [ch1341](https://app.clubhouse.io/simpledotorg/story/1341/change-all-sync-resources-to-patient-based-boundaries)

## Because

All sync resources currently sync based on encounters. This means that a facility's BPs are defined as "BPs taken at the facility". This was ok since our sync boundary was a whole facility group.

However, in preparation for block-level sync, we should change these boundaries to be patient-based. ie, BPs for a facility should be "BPs taken anywhere belonging to patients synced to this facility".

## This addresses

* Adds a `syncable_to_region` scope to all sync resources
* Utilizes this new scope in all sync controllers


spec/controllers/api/v3/appointments_controller_spec.rb
```diff
@@ -135,14 +137,16 @@ def create_record_list(n, options = {})
...
-        FactoryBot.create_list(:appointment, 2, facility: facility_in_same_group, updated_at: 5.minutes.ago)
-        FactoryBot.create_list(:appointment, 2, facility: request_facility, updated_at: 7.minutes.ago)
+
+        other_patient = create(:patient)
+        create_record_list(2, patient: other_patient, updated_at: 3.minutes.ago)
+        create_record_list(2, facility: facility_in_same_group, updated_at: 5.minutes.ago)
+        create_record_list(2, facility: request_facility, updated_at: 7.minutes.ago)
       end
 
       it "only sends data for facilities belonging in the sync group of user's registration facility" do
```
prabhanshuguptagit, line 152: `belonging to patients in the...`?

spec/controllers/api/v3/prescription_drugs_controller_spec.rb
```diff
@@ -107,14 +109,15 @@ def create_record_list(n, options = {})
     end
 
     describe "syncing within a facility group" do
-      let(:facility_in_same_group) { FactoryBot.create(:facility, facility_group: request_user.facility.facility_group) }
-      let(:facility_in_another_group) { FactoryBot.create(:facility) }
+      let(:facility_in_same_group) { create(:facility, facility_group: request_user.facility.facility_group) }
+      let(:facility_in_another_group) { create(:facility) }
+      let(:other_patient) { create(:patient) }
```
prabhanshuguptagit, line 114: `other_patient` should belong to `facility_in_another_group` here.

prabhanshuguptagit (changes requested):
Looks solid ✨ . We should give this a spin on `perf-1`and `perf-2` and see if the numbers tally up with the list here https://simpledotorg.slack.com/archives/C018TPYH3J9/p1600942515010400. Not sure how we would check the number of resources synced on the app, ~maybe we write a script~? a debug build! :)

app/models/patient.rb
```diff
@@ -82,6 +82,10 @@ class Patient < ApplicationRecord
       .merge(PatientPhoneNumber.phone_type_mobile)
   }
 
+  scope :syncable_to_region, ->(region) {
+    with_discarded.where(registration_facility: region.facilities)
```
rsanheim, line 86: ✨ I really dig the incremental improvements we are getting with the whole region abstraction. Eventually this `region` can just be a Block region, and things will work nicely.

harimohanraj89: Exactly! That's what I had in mind with this change as well.

image

prabhanshuguptagit (approved):

PR #1376, opened 2020-09-30
Bugfix: encounter generation

prabhanshuguptagit:
**Story card:** [ch1442](https://app.clubhouse.io/simpledotorg/story/1442/fix-bp-sync-bug)

## Because

BP syncs are failing for some users with a `500` due to an edge case in encounter generation. When a user tries to sync BP/BS in a soft-deleted facility, saving the encounter fails due to a `belongs_to` rails validation on `facility_id`. This causes the observation creation to fail with:

```
insert or update on table "observations" violates foreign key constraint "fk_rails_47dc1837b8"
DETAIL:  Key (encounter_id)=(9859dd29-a0fd-5420-9c1b-00c425b87c6f) is not present in table "encounters".
```

For example, this is happening currently at the facility `CHC BHAM` which was soft deleted since a duplicate facility with the same name existed. Users who have BPs and BSs recorded at that facility on their device are currently receiving a 500 everytime those BPs/BSs are tried to be synced.

[Sentry ref](https://sentry.io/organizations/resolve-to-save-lives/issues/1914505499/?environment=production&project=1217715&query=is%3Aunresolved)

## This addresses

To mitigate failing syncs for those users, this adds a validation to BP and BS resources that checks for the facility's existence. By returning an error for the culprit BPs/BSs, the apps mark them as erroneous and eventually clear them from the device. 

There is no clean way to accept those BPs since the facility doesn't exist anymore. Having encounters without a facility in the system might have far-reaching effects on reporting etc.

### DIFF
--- a/spec/controllers/api/v3/blood_pressures_controller_spec.rb
+++ b/spec/controllers/api/v3/blood_pressures_controller_spec.rb
@@ -265,9 +265,7 @@ RSpec.describe Api::V3::BloodPressuresController, type: :controller do
 
     context "for a soft deleted facility" do
       before :each do
-        request.env["HTTP_X_USER_ID"] = request_user.id
-        request.env["HTTP_X_FACILITY_ID"] = request_facility.id
-        request.env["HTTP_AUTHORIZATION"] = "Bearer #{request_user.access_token}"
+        set_authentication_headers
       end
 
       it "returns an error and does not create the blood pressure" do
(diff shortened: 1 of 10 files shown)


spec/controllers/api/v3/blood_pressures_controller_spec.rb
```diff
@@ -262,6 +262,25 @@ def create_record_list(n, options = {})
         end
       end
     end
+
+    context "for a soft deleted facility" do
```
harimohanraj89, line 266: I think `for a deleted facility` should suffice.

harimohanraj89 (approved):
Looks great. Let's get these users' sync back online 👍

PR #1378, opened 2020-09-30
Remove old reports

rsanheim:
**Story card:** [ch1438](https://app.clubhouse.io/simpledotorg/story/1438/remove-old-reports)

**Do not merge** until October 5th

Removes all old reports (i.e. Analytics controllers) including views, routes, and controller.


app/helpers/reports/regions_helper.rb
```diff
@@ -4,6 +4,11 @@ def reports_region_facility_path(facility, options = {})
     reports_region_path(facility, options)
   end
 
+  def reports_region_district_path(district, options = {})
```
harimohanraj89, line 7: Why a helper for this instead of setting the `report_scope`, like we do for the facility paths?

app/views/analytics/districts/show.html.erb
```diff
@@ -1,130 +0,0 @@
...
-  </nav>
-
-  <%= render "period_dropdown" %>
-
-  <h1><%= @organization_district.district_name %></h1>
-</div>
-
-<%= render 'shared/cohort_charts', cohort_analytics: @cohort_analytics, quarter: current_quarter, year: current_year %>
-<% if current_admin.permissions_v2_enabled? %>
-  <%= render 'follow_ups_table',
```
harimohanraj89, line 18: There are some partials like this sitting in this directory that can now be moved to the `reports/region` area, since those are the only reports that use em now.

harimohanraj89 (commented):
Some early feedback since most of us are out tomorrow.

Also - Can we clean up the routes as well?

app/views/analytics/districts/show.html.erb
rsanheim: Yup, removed those.

app/helpers/reports/regions_helper.rb
rsanheim: I've got helpers for both facility and facility group now, just haven't switched over to use them everywhere (yet).

app/views/reports/regions/index.html.erb
```diff
@@ -13,7 +13,7 @@
     <div class="col-sm-5 col-lg-7">
       <% @organizations.each do |organization| %>
       <section>
-        <div class="card">
+        <div class="card org-card">
```
claudiovallejo, line 16: @rsanheim what does the `org-card` class do?

claudiovallejo (approved):

app/views/reports/regions/show.html.erb
```diff
@@ -244,7 +244,7 @@
                row_resource_display_field: :full_name,
                row_resource_link: :admin_user_path %>
   <% else %>
-    <%= render "analytics/facilities/analytics_table",
+    <%= render "new_registrations_table",
```
harimohanraj89, line 247: I think we can call this `registrations_table` for consistency, what do you think?

harimohanraj89 (approved):
One last suggestion from me. Rest looks 🔥 🚀

app/views/reports/regions/index.html.erb
rsanheim: It lets us better target these classes in some of our integration tests - see https://github.com/simpledotorg/simple-server/pull/1378/files#diff-6f0d4ecd603cef429123ac37759e41fe.

app/views/reports/regions/show.html.erb
rsanheim: There already is a `registrations_table` in mix here -- we are running out of names. 😁

PR #1382, opened 2020-10-01
Reports: "Visit details..." updates

claudiovallejo:
**Story card:** [ch1390]

## Because

We want to make Reports great :)

## This addresses

* Move "Visit details of all hypertensive patients" card to "Overview" section
* Clean up `details` in `regions_controller.rb`
* Show 6 months of data in "Visit details of all hypertensive patients"

**Screenshots**
image
image


app/views/reports/regions/show.html.erb
```diff
@@ -224,6 +224,87 @@
     </div>
   </div>
 </div>
+<div class="mt-8px mx-0px mb-16px p-20px bg-white br-4px bs-small d-lg-flex fd-lg-column justify-lg-between h-lg-full w-lg-full mt-lg-0">
```
rsanheim, line 227: I would recommend extracting this to a partial to keep `show` from getting any longer.

rsanheim (approved):
One small review item.

PR #1388, opened 2020-10-05
Bugfix: Patient line list

prabhanshuguptagit:
**Story card:** [ch1463](https://app.clubhouse.io/simpledotorg/story/1463/fix-medical-history-linelist-bug)

## Because

We caught a bug [on sentry](https://sentry.io/organizations/resolve-to-save-lives/issues/1800173727/?project=1217715) in the medication history line list: 
image

## This addresses

This happens when concurrent line list downloads are queued. In `PatientWithHistoryExporter`, `@medications` is a class instance variable and gets mutated by multiple sidekiq workers. Sidekiq explicitly warns against [using class variables](https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting#writing-thread-safe-code).  This changes `@medications` to be an instance variable.


app/exporters/patients_with_history_exporter.rb
```diff
@@ -1,7 +1,7 @@
 require "csv"
 
-module PatientsWithHistoryExporter
-  extend QuarterHelper
+class PatientsWithHistoryExporter
```
kitallis, line 3: Let's make the other Exporter a class also for consistency?

harimohanraj89 (approved):
Nice, they're both classes now. Fun fact, I never realized that these were modules until today!

PR #1389, opened 2020-10-05
Run integration tests for the review apps automatically (not-required)

kitallis:
## Because

If automatic tests are run manually, it's unclear which endpoint to hit to run the tests against. Ideally it should be SBX, and not the review app, as it's currently configured to do.

## This addresses

Simply run the integration tests on each PR instead of a manual trigger. We can tweak this later to trigger only sync tests (since those are the most relevant) to increase build speed (mobile-devs are working on it).


prabhanshuguptagit (approved):
Do we need to explicitly wait for the heroku env to spin up before running this? It _might_ fail on the first run, not a big deal though.

rsanheim:
> Simply run the integration tests on each PR instead of a manual trigger. We can tweak this later to trigger only sync tests (since those are the most relevant) to increase build speed (mobile-devs are working on it).

How much time will this add to the CI cycle? I'm concerned that this will add noise to every PR cycle, because we will need to investigate if the failing / pending check is this long running integration test?

kitallis:
@rsanheim They take around 10-15m to run. However, I think if we say that `master` should always be deployable then it's imperative that we know syncing works correctly before we merge branches to `master` and running integration tests (sync tests) is another facet of that confirmation.

However, given the low churn on sync-related changes on an ongoing basis, we can take a middle-ground, where we simply keep these QA tests completely optional and still allow people to merge based on their own judgement even if the QA tests don't finish running.

In the future, @vinaysshenoy is planning to split out the sync tests from the integration suite which will make these tests run a lot faster and we can look to making these mandatory.

vinaysshenoy:
> @rsanheim They take around 10-15m to run. However, I think if we say that `master` should always be deployable then it's imperative that we know syncing works correctly before we merge branches to `master` and running integration tests (sync tests) is another facet of that confirmation.
> 
> However, given the low churn on sync-related changes on an ongoing basis, we can take a middle-ground, where we simply keep these QA tests completely optional and still allow PRs to be merged based on people's own judgement even if the QA tests don't finish running. I've changed the setting to reflect that.
> 
> In the future, @vinaysshenoy is planning to split out the sync tests from the integration suite which will make these tests run a lot faster and we can look to making these mandatory.

We could also run them on a nightly schedule against QA after deploys are done so they don't block PRs.

PR #1391, opened 2020-10-05
Don't show current period of registrations

rsanheim:
This keeps it consistent with what the currently selected period is

**Story card:** [ch1327](https://app.clubhouse.io/simpledotorg/story/1327/show-last-6-months-not-including-current-month-in-new-registrations-card)

### DIFF
--- a/app/controllers/reports/regions_controller.rb
+++ b/app/controllers/reports/regions_controller.rb
@@ -46,10 +46,10 @@ class Reports::RegionsController < AdminController
                                                         period: @period).call
       }
     else
-      @show_current_period = true
+      @show_current_period = false
       @dashboard_analytics = @region.dashboard_analytics(period: :month,
                                                          prev_periods: 6,
-                                                         include_current_period: true)
+                                                         include_current_period: false)
     end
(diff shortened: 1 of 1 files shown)


app/controllers/reports/regions_controller.rb
```diff
@@ -46,10 +46,10 @@ def show
                                                         period: @period).call
       }
     else
-      @show_current_period = true
+      @show_current_period = false
```
harimohanraj89, line 49: Is this used in the view? Can we rip it out and hard-code falses where necessary, since it can't be anything else?

harimohanraj89 (approved):
Oh, well that was easy 😅 

One non-blocking comment from me. 🚀

app/controllers/reports/regions_controller.rb
rsanheim: I don't wanna change too much while all of this is in flight (and we have views shared by old reports and new).

I think long term we need to pass in the `Period` object to this tables to really do things right, but that will be easier once the old stuff is all gone.

PR #1392, opened 2020-10-05
Turn on Datadog APM everywhere except dev

rsanheim:
After pairing with @prabhanshuguptagit this morning a bit we realized it would be good to have APM turned on everywhere to start - it makes it easier to track down errors or problems even in non-prod environents.

We can always dial this back later if we want to.


harimohanraj89 (approved):
Hehe I was on the fence about the limited activation as well. Let's do it.

PR #1393, opened 2020-10-05
Add statsd instrumentation to Sidekiq

rsanheim:
For:

```
    global metrics
    queue metrics
    worker metrics (processing, runtime)
    job metrics (processing_time and success / failure)
```

via https://github.com/phstc/sidekiq-statsd

**Story card:** [ch1356](https://app.clubhouse.io/simpledotorg/story/1356/setup-datadog-in-production)


harimohanraj89 (approved):
Neat. Data Dog will automatically slurp up these new stats and expose them? 

PS why the rename to `config/initializers/10_statsd.rb`? Do we need it to run before the sidekiq initializer? 😅  Since it's just a singleton class declaration inside, could we move it to `lib` and require it explicitly where necessary?

rsanheim:
> PS why the rename to `config/initializers/10_statsd.rb`

Yeah, its for the load time order.  We can move it to lib though.

PR #1403, opened 2020-10-07
Reports: Add region total rows and copy/styling updates

claudiovallejo:
**Story card:** [ch1399](https://app.clubhouse.io/simpledotorg/story/1399/add-all-rows-to-facility-comparison-tables)

## Because

1. It makes it easier to see (and verify) accuracy of facility data
2. It shows the region's name as the "All" header (per CVHOs request)

## This addresses

**Add total rows to facility comparison tables**
image

**Replace "All" with region's name in District Report tables**
image

**Replace "All" with region's name in Facility Report tables**
image

**Update cohort chart key label and tooltip copy**
image
image

**Format cohort chart tooltips to properly display "patient" in singular/plural**
image

### DIFF
--- a/app/views/reports/regions/cohort.html.erb
+++ b/app/views/reports/regions/cohort.html.erb
@@ -52,7 +52,7 @@
                     data-toggle="tooltip"
                     data-placement="top"
                     data-trigger="hover focus click"
-                    title="<%= number_with_delimiter(cohort["uncontrolled"]) %> patients visited with HTN not under control"
+                    title="<%= number_with_delimiter(cohort["uncontrolled"]) %> patients visited with BP not controlled"
                     style="width: <%= uncontrolled_percent %>;"
                   >
                     <%= (cohort["uncontrolled"] > 1 && uncontrolled_percent == 0) ? "< 1" : uncontrolled_percent %>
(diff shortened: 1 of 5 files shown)


app/views/reports/regions/cohort.html.erb
```diff
@@ -42,7 +42,7 @@
                     data-toggle="tooltip"
                     data-placement="top"
                     data-trigger="hover focus click"
-                    title="<%= number_with_delimiter(cohort["no_bp"]) %> patients didn't have a BP taken"
+                    title="<%= number_with_delimiter(cohort["no_bp"]) %> <% if cohort["no_bp"] == 1 %>patient<% else %>patients<% end %> didn't have a BP taken"
```
rsanheim, line 45: [pluralize](https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize) is your friend here (and below) .  No need to do this by hand. 😁

rsanheim (approved):
One place where you can use a Rails view helper, otherwise LGTM.  approving to unblock.

app/views/reports/regions/cohort.html.erb
harimohanraj89: Thanks Rails :heart:

claudiovallejo: Pluralize is so nice ❤️

PR #1405, opened 2020-10-08
Use the localized name for districts in the banner

harimohanraj89:
**Story card:** [ch1536](https://app.clubhouse.io/simpledotorg/story/1536/update-facility-group-header-to-be-region-specific)

## Because

The word "District" doesn't always make sense in all countries

Screen Shot 2020-10-08 at 5 26 43 PM

## This addresses

Use the localized word for "District" in the helpful facility management banner.

Screen Shot 2020-10-08 at 5 27 24 PM


app/views/admin/facilities/index.html.erb
```diff
@@ -22,7 +22,9 @@
 <% else %>
   <div class="alert alert-info alert-dismissible fade show" role="alert">
     <i class="far fa-question-circle mr-2"></i> A <span class="term">Facility Group</span> is usually a
-    <span class="term">District</span>. First, add a new facility group and then add facilities inside it.
+    <span class="term"><%= t('helpers.label.facility.district').gsub(/\s\*$/, '').strip %></span>.
```
kitallis, line 25: Ugh, don't know why we have the `*` in the locale files. It should be in the markup :|

harimohanraj89: Hehe yeah. I chose this surgical approach instead of detangling that right now 😅

kitallis (approved):

PR #1410, opened 2020-10-09
Only show supported genders in the progress tab

kitallis:
**Story card:** [ch1537](https://app.clubhouse.io/simpledotorg/story/1537)

## Because

Ethiopia, for example, should not have the `transgender` option enabled.

## This addresses

Populate the gender dropdowns via the country configuration.


harimohanraj89 (approved):
:shipit:

PR #1411, opened 2020-10-09
Manually mark mobile numbers

harimohanraj89:
**Story card:** [ch1534](https://app.clubhouse.io/simpledotorg/story/1534/investigate-and-fix-non-ihci-reminders)

## Because

Non-IHCI environments don't have a way to identify mobile numbers. IHCI uses Exotel.

## This addresses

Marking all patient phone numbers as mobile every night, if `FORCE_MARK_PATIENT_MOBILE_NUMBERS=true`

### DIFF
--- a/app/services/mark_patient_mobile_numbers.rb
+++ b/app/services/mark_patient_mobile_numbers.rb
@@ -6,10 +6,21 @@ class MarkPatientMobileNumbers
   def call
     return unless ENV["FORCE_MARK_PATIENT_MOBILE_NUMBERS"] == "true"
 
+    count = eligible_numbers.count
+    notify("Force-marking #{count} as mobile numbers")
+
     eligible_numbers.update_all(phone_type: "mobile")
+
+    notify("Finished force-marking #{count} as mobile numbers")
   end
+
+  private
 
   def eligible_numbers
     PatientPhoneNumber.where.not(phone_type: "mobile")
+  end
+
+  def notify(msg)
+    Rails.logger.info msg: msg, class: self.class.name
   end
 end
--- a/app/services/mark_patient_mobile_numbers.rb
+++ b/app/services/mark_patient_mobile_numbers.rb
@@ -4,7 +4,7 @@ class MarkPatientMobileNumbers
   end
 
   def call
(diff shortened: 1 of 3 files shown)


app/services/mark_patient_mobile_numbers.rb
```diff
@@ -0,0 +1,26 @@
...
+    new.call
+  end
+
+  def call
+    return unless ENV["FORCE_MARK_PATIENT_MOBILE_NUMBERS"] == "true"
+
+    count = eligible_numbers.count
+    notify("Force-marking #{count} as mobile numbers")
+
+    eligible_numbers.update_all(phone_type: "mobile")
```
kitallis, line 12: There's 12k phone numbers that this will update in a single query everyday. Should we consider batching, or do you think this suffices for now?

kitallis (approved):
lgtm

app/services/mark_patient_mobile_numbers.rb
```diff
@@ -0,0 +1,26 @@
+class MarkPatientMobileNumbers
+  def self.call
+    new.call
+  end
+
+  def call
+    return unless ENV["FORCE_MARK_PATIENT_MOBILE_NUMBERS"] == "true"
```
kitallis, line 7: Do you want to use Flipper for this?

harimohanraj89: I thought about it. I think I like the idea of long-running switches like this in Flipper. You know what, I'll do it 👍

app/services/mark_patient_mobile_numbers.rb
harimohanraj89: It's ~7k in BD today. It's a single SQL query that won't stress app memory or performance. I think a single sweep is ok.

app/services/mark_patient_mobile_numbers.rb
kitallis: I was hoping, that we eventually migrate all our feature toggles or flags of all kind to flipper, keep bare-minimum config in the config.

harimohanraj89: I was on the fence, but I think you've sold me. Done in 5156590

PR #1412, opened 2020-10-09
Address nil type numbers too

harimohanraj89:
**Story card:** [ch1534](https://app.clubhouse.io/simpledotorg/story/1534/investigate-and-fix-non-ihci-reminders)

Follow-up bugfix for #1411


kitallis (approved):
nailed it

PR #1394, opened 2020-10-06
Wrap up permissions changes (take 2)

vkrmis:
**Story card:** [ch1513](https://app.clubhouse.io/simpledotorg/story/1513/cleanup-pundit-and-old-permissions-code)

## This addresses

- [x] ApplicationController
- [x] AdminController
- [x] ApplicationLayout
- [x] Users Factory
- [x] Shared Partials
- [x] Routes
- [x] Remove Pundit
- [x] Remove pundit usage from ApprovalNotifierMailer 
- [x] Delete policies
- [x] Migrate/delete rake scripts
- [x] Migrate feature specs
- [x] Rename authorize_v2
- [x] Dev-env data migration to upgrade existing admins permissions

A final cleanup PR is still necessary and will come up next week: [ch1546](https://app.clubhouse.io/simpledotorg/story/1546/final-cleanup-of-old-permissions)

PR #1413, opened 2020-10-09
District filtering bugfix

harimohanraj89:
## Because

Hot fix for #1394 - The new RankedFacilitiesController was using some artifacts (`DistrictFiltering` module) that were removed in #1394.

## This addresses

Removing the DistrictFiltering and adding a simple controller test to ensure it's not broken.


vkrmis (approved):
Looks good :+1:

PR #1414, opened 2020-10-09
Re-add delete admin button

vkrmis:
**Story card:** [ch1547](https://app.clubhouse.io/simpledotorg/story/1547)

## Because

We disabled the "Delete admin" button while changing the permissions model

## This addresses

Adds the "Delete admin" button using the new permission model


kitallis (approved):
lgtm

PR #1416, opened 2020-10-09
Reports: Value formatting

claudiovallejo:
**Story card:** [ch1549](https://app.clubhouse.io/simpledotorg/story/1549/chart-borders-and-value-formatting)

## Because
All numeric values should be comma-separated

## This addresses
Creates a new function to format all numeric values displayed in chart tooltips

### DIFF
--- a/app/assets/javascripts/common/reports.js
+++ b/app/assets/javascripts/common/reports.js
@@ -496,10 +496,10 @@ function onePlotTooltip(config) {
     registrationsPeriodEndNode.innerHTML = periodStartNode.getAttribute("data-period-start");
   } else {
     rateNode.innerHTML = dataPoints[0].value + "%";
-    totalPatientsNode.innerHTML = Object.values(totalPatients)[dataPoints[0].index];
+    totalPatientsNode.innerHTML = formatNumberWithCommas(Object.values(totalPatients)[dataPoints[0].index]);
     periodStartNode.innerHTML = Object.values(periodInfo)[dataPoints[0].index].bp_control_start_date;
     periodEndNode.innerHTML = Object.values(periodInfo)[dataPoints[0].index].bp_control_end_date;
-    registrationsNode.innerHTML = Object.values(adjustedRegistrations)[dataPoints[0].index];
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/common/reports.js
```diff
@@ -570,12 +570,16 @@ function stackedBarChartTooltip(config) {
...
+    controlledPatientsNode.innerHTML = formatNumberWithCommas(controlledPatients[dataPoints[0].index]);
     periodStartNodes.forEach(node => node.innerHTML = periodInfo[dataPoints[0].index].bp_control_start_date);
     periodEndNodes.forEach(node => node.innerHTML = periodInfo[dataPoints[0].index].bp_control_end_date);
-    cumulativeRegistrationsNodes.forEach(node => node.innerHTML = adjustedRegistrations[dataPoints[0].index]);
+    cumulativeRegistrationsNodes.forEach(node => node.innerHTML = formatNumberWithCommas(adjustedRegistrations[dataPoints[0].index]));
   }
+}
+
+function formatNumberWithCommas(value) {
+  return value.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
```
kitallis, line 584: @claudiovallejo Could you post a screenshot of how the numbers must be formatted to see if we can find a simpler solution to this? It's hard to parse this regex to figure out what it's trying to do.

kitallis (changes requested):

app/assets/javascripts/common/reports.js
kitallis: Also, I'm assuming this will eventually have to be localized, hence, if possible, we should try to do this in Ruby.

claudiovallejo: @kitallis I would add a screenshot, but fake data in SBX are all less than 1,000 so no commas are displayed. But here's an example: `12345 -> 12,345`.

Is it ok if we fix in JS now? I'm doing some cleanup work today and tomorrow and can ask Rob's help to move this to Ruby.

kitallis (approved):

PR #1417, opened 2020-10-12
Enable Rails 5.2 Content Security Policy

kitallis:
**Story card:** [ch1563](https://app.clubhouse.io/simpledotorg/story/1563/add-a-content-security-policy)

## Because

Our app doesn't expose a [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) (CSP).

## This addresses

Create one for all pages (for the time being) and configure any third-party hosts as necessary.

Sentry has a neat feature that can be used to [report CSP violations](https://sentry.io/settings/resolve-to-save-lives/projects/simple-server/security-headers/), we have set our CSP to `report-only` mode and which should show up as alerts on Slack if anything is violated. This will allow us to find any violations that we may not catch immediately and can fix them incrementally, rather than outright breaking the loading of the page.

Companion config changes: https://github.com/simpledotorg/deployment/pull/273


prabhanshuguptagit (approved):
Should merge https://github.com/simpledotorg/deployment/pull/273 before merging.

PR #1420, opened 2020-10-12
Remove real info from app.json

harimohanraj89:
**Story card:** TBD - [chXXXX](URL)

## Because

There's some real people info in the `app.json`, which powers review apps.

## This addresses

Replacing that real info with test emails.


kitallis (approved):
lg

PR #1423, opened 2020-10-13
Disable the TaggedLogging patch on review apps (ie Heroku)

rsanheim:
**Story card:** [ch1579](https://app.clubhouse.io/simpledotorg/story/1579/review-apps-are-breaking)

It appears the tagged logging patch for datadog + tagged logging is breaking Heroku review apps.

Trying to disable that patch to get things back up and running.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -28,7 +28,6 @@ module SimpleServer
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
 
-    # Don't generate system test files.
     config.generators do |g|
       g.system_tests false
       g.helper false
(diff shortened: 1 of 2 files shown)


kitallis (approved):
Thanks for looking into this @rsanheim. This appears to be working 👍, not sure what exactly is wrong though. Merging in to unblock review apps.

rsanheim:
> Thanks for looking into this @rsanheim. This appears to be working 👍, not sure what exactly is wrong though. Merging in to unblock review apps.

Yup, my hunch is that Heroku overrides logger settings somehow, because I know they want STDOUT / simple formatting.  That probably conflicts with the JSON / tagged log patch.

PR #1424, opened 2020-10-14
Skip sending registration_approval_email in non production environments

vkrmis:
**Story card:** [ch1584](https://app.clubhouse.io/simpledotorg/story/1584/non-production-environments-are-sending-too-many-emails-for-new-registrations)

## Because

Non production environments are sending too many registration_approval_emails

## This addresses

Disable these emails for non-production environments


app/mailers/approval_notifier_mailer.rb
```diff
@@ -2,14 +2,18 @@ class ApprovalNotifierMailer < ApplicationMailer
...
   def registration_approval_email(user_id:)
-    @user = User.find(user_id)
-    subject = I18n.t("registration_approval_email.subject",
-      full_name: @user.full_name,
-      org_name: @user.facility_group.organization.name)
-    mail(subject: subject,
-         to: supervisor_emails,
-         cc: organization_owner_emails,
-         bcc: owner_emails)
+    if ENV["SIMPLE_SERVER_ENV"] == "production"
```
kitallis, line 5: I think you can simply use `SIMPLE_SERVER_ENV` directly since that constant is set [here](https://github.com/simpledotorg/simple-server/blob/master/config/application.rb#L22).

kitallis (approved):
Looks good, left a minor comment.

kitallis:
Wup, I think some important tests are failing... @vkrmis .

vkrmis:
> Wup, I think some important tests are failing... @vkrmis.

Fixed. There are specs in `api/v3/users_controller_spec.rb` that check some of the same behaviour. Needed to stub `SIMPLE_SERVER_ENV` there as well.

PR #1426, opened 2020-10-14
Reports: Activity Tables + Current Month

claudiovallejo:
**Story card:** [ch1593](https://app.clubhouse.io/simpledotorg/story/1593/update-facility-follow-ups-query)

## Because

* We need to show data for current month
* CVHOs want to see activity tables for District and Facility Reports
* Activity tables show all follow-up, new registration, and cumulative registration data
* Follow-up data doesn't duplicate patient follow-ups where the patient visited twice in the same month

## This addresses

**Resurrected follow-up query and tests (copied it from #1169)**

**Removed "New registrations" card from Facility Reports "Overview"**

**Added "Facility activity" table to District Reports "Details"**

image

**Added "User activity" table to Facility Reports "Details"**

image

**Reports show data for current month**

image

### Implementation Notes

* Introduce an `ActivityService` class that can fetch registrations and followups that occurred _at_ a facility, regardless of where patients are assigned, using the model scopes that power the Progress tab today
* The service class groups by period (month, day, etc), and can also group by additional dimensions like gender (for progress tab), facility (for district analytics), or user (for facility analytics)
* The 3 touchpoints that use this class now for consistent regn and followup numbers are:
  * Progress tab `UserAnalyticsPresenter`
  * District activity `DistrictAnalyticsQuery`
  * Facility activity `FacilityAnalyticsQuery`
* The use of a single service class ensures that the numbers always "add up". Grouped-by-facility numbers always add up to district numbers, etc, across all touchpoints
* Uniqueness of followups will continue to follow the Progress tab's behavior for now: For monthly numbers, 2 visits in a month count as 1 followup. For daily numbers, 2 visits in day count as 1 followup.


app/queries/facility_analytics_query.rb
```diff
@@ -54,6 +55,46 @@ def registered_patients_by_period
...
+                .select("blood_pressures.recorded_at AS bp_recorded_at")
+                .order("blood_pressures.patient_id",
+                  Arel.sql(BloodPressure.date_to_period_sql("blood_pressures.recorded_at", @period)),
+                  "blood_pressures.recorded_at"),
+          "patients")
+        .group("bp_user_id")
+        .group_by_period(:month, "bp_recorded_at")
+        .count
+
+    group_by_user_and_date(@follow_up_patients_by_period, :follow_up_patients_by_period)
```
rsanheim, line 95: Is this any different from the query in the districts query object? At first glance they appear the same, which makes me think it should be its own object that can just work with a Region.

rsanheim (changes requested):
Have a question on what looks like a duplicated query?

app/queries/facility_analytics_query.rb
harimohanraj89: The facility query does a lot of grouping by _user_, whereas the district query groups by _facility_. It's a bit awkward to detangle them cleanly :/

harimohanraj89: Hmm, perhaps we _could_ pull the base query into a separate object and leave the grouping and data structure massaging in here. I'll take a stab at it and see if it turns out ok.

app/controllers/reports/regions_controller.rb
```diff
@@ -33,18 +33,16 @@ def show
         hsh[facility.name] = Reports::RegionService.new(region: facility,
                                                         period: @period).call
       }
-    else
-      @show_current_period = false
-      @dashboard_analytics = @region.dashboard_analytics(period: :month,
-                                                         prev_periods: 6,
-                                                         include_current_period: false)
```
harimohanraj89, line 40: @claudiovallejo We don't wanna show this info on the overview anymore, I guess?

claudiovallejo: @harimohanraj89 that's right — I removed this query because we removed the new registrations table from Facility Reports "Overview"

app/queries/facility_analytics_query.rb
claudiovallejo: @rsanheim the queries are practically the same. The difference is that the districts query groups results by facility and the facility query groups results by user. If you still think it should work with a Region, I'd appreciate your guidance on how to make that change

app/queries/district_analytics_query.rb
```diff
@@ -73,6 +91,46 @@ def patients_with_bp_by_period
     group_by_facility_and_date(@patients_with_bp_by_period, :patients_with_bp_by_period)
   end
 
+  def follow_up_patients_by_period
```
kitallis, line 94: This query is doing some crazy things, I think we should re-think this.

It tries hard to assign a single visit to any user to match numbers in the facility reports. If a patient visits 2 separate users in the same month and the same facility, it only picks one of those users (the first one) and ends up not assigning the other user a visit. 

This metric is dubious at best because it under-counts the user's work.

I think we should simplify this and make it represent the truth better; multiple visits across users should be counted for all users.

harimohanraj89: Ooh that's a good call. Counting both visits is preferable, and I didn't realize that it would be easier to do so. I think we can avoid a lot of this `from`, `distinct` convolution.

Let me take a stab at this as well.

kitallis: I might be able to dig the original original query.

harimohanraj89: This is helpful, thanks. Yup, I see a `hypertension_followups_*` scope in our code today that seems to make this a lot easier. Stay tuned 🤞

claudiovallejo: @kitallis @harimohanraj89 thanks for the feedback. I updated the district and facility `follow_up_patients_by_period` queries to use the first version of the query referenced above by Kitty. Rob and I also updated the respective query specs.

app/queries/district_analytics_query.rb
```diff
@@ -73,6 +91,34 @@ def patients_with_bp_by_period
...
+        .left_outer_joins(patient: [:medical_history])
+        .joins(:facility)
+        .where(facilities: { id: facilities })
+        .where(deleted_at: nil)
+        .where("medical_histories.hypertension = ?", "yes")
+        .group('facilities.id')
+        .group_by_period(@period, 'blood_pressures.recorded_at')
+        .where("patients.recorded_at < #{date_truncate_string}")
+        .order('facilities.id')
+        .distinct("recorded_at")
```
claudiovallejo, line 116: @harimohanraj89 I added `distinct("recorded_at")` so that multiple BPs recorded for a patient in a single day are counted once.

app/queries/district_analytics_query.rb
claudiovallejo: END OF DAY UPDATE: @harimohanraj89 @kitallis there was a discussion on Slack with Daniel (thread starts here)[https://simpledotorg.slack.com/archives/CK95QFJA2/p1603225062040200] and we'd like to update the follow-ups query to count unique patient visits with a BP recorded per day within a month. I updated the footnotes to reflect this new query, here's what's pending:
* Update district and facility follow-up queries so that BPs are unique by day
* Create respective tests (I took a stab at creating the facility test)


Would love to pair with either of you tomorrow if you have a chance so we can ship tomorrow

kitallis: @harimohanraj89 @claudiovallejo I'm a little confused about this new definition, we can discuss this in person more clearly, but:

In either district or facility follow-ups, the idea of follow-ups has had nothing to do with visits up until now, plainly stated follow-ups are like this:

```
Have registered before the beginning of month M at any facility
AND
Have diagnoed HTN = yes
AND
Have assigned facility at facility F 
AND
Have at least 1 Blood Pressure reading in month M at facility F
```

If we do it by visits, then if a Patient visits 30 times in a month each day, the follow-ups for that month will be 30. That historically has never been how we've computed follow-ups. See the [scenarios](https://docs.google.com/document/d/16ga4Ug-t-aBvy_p0OfxbOacBDgH7LdIC7VCP4N1y_Cg/edit#heading=h.5whpx7awcvca) here for reference.

Secondly, how are we planning to group these numbers in the facility analytics, by user?

app/controllers/reports/regions_controller.rb
```diff
@@ -33,20 +33,19 @@ def show
...
     authorize { current_admin.accessible_facilities(:view_reports).any? }
 
-    @dashboard_analytics = @region.dashboard_analytics(period: @period.type, prev_periods: 6)
+    @show_current_period = true
+    @dashboard_analytics = @region.dashboard_analytics(period: @period.type,
+                                                       prev_periods: 6,
+                                                       include_current_period: true)
 
     if @region.is_a?(Facility)
+      @facility_progress = UserAnalyticsPresenter.new(@region)
```
harimohanraj89, line 48: @claudiovallejo I think we can remove this now in favor of grabbing the data straight from `@dashboard_analytics`. AKA, replace the top "Facility Progress" table with data from `@dashboard_analytics`.

app/queries/district_analytics_query.rb
harimohanraj89: Closing this loop after oodles of discussion. Rather than nail this definition the first time around, let's instead aim for _consistency_ across facility details, district details, and progress tab. This PR is introducing some centralized service classes to fetch this info, so we can iterate on the definition easily.

app/queries/district_analytics_query.rb
```diff
@@ -25,52 +26,40 @@ def call
...
-      Patient
-        .group("assigned_facility_id")
-        .where(assigned_facility: facilities)
-        .hypertension_follow_ups_by_period(@period, last: @prev_periods)
-        .count
+  def follow_up_patients_by_period
+    result = ActivityService.new(region, last: nil, group: BloodPressure.arel_table[:facility_id]).follow_ups
 
-    group_by_facility_and_date(@patients_with_bp_by_period, :patients_with_bp_by_period)
+    group_by_date_and_facility(result, :follow_up_patients_by_period)
```
harimohanraj89, line 62: Instead of all this in-memory aggregation, do you think we could just run another `ActivityService` query? Perhaps...

```ruby
ActivityService.new(region, last: nil).follow_ups
```

...perhaps no `group` will return results that are grouped only by date, which is what we want here?

app/presenters/user_analytics_presenter.rb
```diff
@@ -292,22 +292,10 @@ def monthly_htn_or_dm_stats
...
-    control_rate_start = control_rate_end.advance(months: -HTN_CONTROL_MONTHS_AGO)
-    controlled_visits = ControlRateService.new(current_facility, periods: control_rate_start..control_rate_end).call
-
-    registrations =
-      current_facility
-        .registered_hypertension_patients
-        .group_by_period(:month, :recorded_at, last: MONTHS_AGO)
-        .group(:gender)
-        .count
+    activity_service = ActivityService.new(current_facility, include_current_period: false)
```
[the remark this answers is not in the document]
harimohanraj89: Todo: Allow the time period to be specified, and have this service power the daily stats too.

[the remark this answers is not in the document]
harimohanraj89: Todo: This service should probably power _all_ the stats (HTN and DM) for maximum consistency 🤔

app/services/activity_service.rb
```diff
@@ -0,0 +1,34 @@
+class ActivityService
```
harimohanraj89, line 1: Todo: Test coverage for this (which can emulate the `UserAnalyticsPresenter` to start)

spec/queries/district_analytics_query_spec.rb
```diff
@@ -221,36 +245,19 @@
...
           {
             facility_2.id =>
               {
-                patients_with_bp_by_period: {
-                  four_months_back => 0,
-                  three_months_back => 1,
-                  two_months_back => 0,
-                  one_month_back => 0
+                follow_up_patients_by_period: {
+                  three_months_back => 1
```
harimohanraj89, line 255: I've updated the assertion here because:
* I'm confident that the view handles zeros well through some click tests
* The other payloads in this data structure do _not_ include zero-keys, which leads me to believe that this one was a fluke

spec/queries/district_analytics_query_spec.rb
```diff
@@ -198,8 +222,8 @@
 
     before do
       Timecop.travel(three_months_back) do
-        create(:blood_pressure, patient: patients.first, facility: facility_2)
-        create(:blood_pressure, patient: patients.second, facility: facility_2)
+        create(:blood_pressure, :with_encounter, patient: patients.first, facility: facility_2)
```
harimohanraj89, line 225: The new followup scopes we're using rely on encounters being present for the BPs ([ref](https://github.com/simpledotorg/simple-server/blob/reports-activity-tables/app/models/facility.rb#L103)). I think this requirement is redundant, but I've chosen to update the test setup for now.

I don't know if it's safe to assume that all BPs have an encounter because #famouslastwords. I'd like to update the `Facility.hypertension_follow_ups_by_period` scope to operation on `Patient.stuff` instead of `patients.stuff` instead.

harimohanraj89 (commented):
Yo @rsanheim, I've left some dev notes for you in my review below 👇 Several todos, some areas for debate.

All three touchpoints (district details, facility details, progress tab) are using the new service. Claudio and I took it for a spin and the numbers are consistent. 80%, just 20% to go 😉

app/services/activity_service.rb
kitallis, line 1: 👌

app/models/patient.rb
```diff
@@ -61,17 +61,17 @@ class Patient < ApplicationRecord
...
-    follow_ups_with(BloodSugar, period, last: last)
+  scope :diabetes_follow_ups_by_period, ->(period, at_region: nil, current: true, last: nil) {
+    follow_ups_with(BloodSugar, period, at_region: at_region, current: current, last: last)
       .with_diabetes
   }
 
-  scope :hypertension_follow_ups_by_period, ->(period, last: nil) {
-    follow_ups_with(BloodPressure, period, last: last)
+  scope :hypertension_follow_ups_by_period, ->(period, at_region: nil, current: true, last: nil) {
+    follow_ups_with(BloodPressure, period, at_region: at_region, current: current, last: last)
```
harimohanraj89, line 74: Added a bit more versatility to these scopes to use them for more touchpoints now.

* `at_region`: data for BPs/BSs recorded _at_ facilities in the region (regardless of where the patient is registered/assigned)
* `current`: include/exclude the current period in the grouped results. This is just a groupdate option that we need to accept and pass down.

app/queries/facility_analytics_query.rb
```diff
@@ -67,15 +68,18 @@ def cache_key
     ].join("/")
   end
 
-  def group_by_user_and_date(query_results, key)
+  def group_by_date_and_user(result, key)
```
harimohanraj89, line 71: Heckin bamboozled.

spec/queries/facility_analytics_query_spec.rb
```diff
@@ -134,8 +180,8 @@
 
     before do
       Timecop.travel(three_months_back) do
-        create(:blood_pressure, :with_encounter, patient: patients.first, facility: facility, user: users.first)
-        create(:blood_pressure, :with_encounter, patient: patients.second, facility: facility, user: users.first)
+        create(:blood_pressure, patient: patients.first, facility: facility, user: users.first)
+        create(:blood_pressure, patient: patients.second, facility: facility, user: users.first)
```
harimohanraj89, line 184: It shouldn't matter if you're missing an encounter.

rsanheim (commented):
Some thoughts - also it looks like you should run `standardrb --fix` to get some of the linting fixed up here.

spec/services/activity_service_spec.rb
```diff
@@ -0,0 +1,744 @@
...
+      end
+
+      activity_service = ActivityService.new(facility_1, diagnosis: :all)
+
+      Timecop.freeze(aug_3) do
+        expect(activity_service.follow_ups).to eq(june_1 => 4, july_1 => 3, aug_1 => 4)
+      end
+    end
+
+    it "counts multiple visits in a month as one monthly followup" do
```
rsanheim, line 452: This definition makes sense to me, and seems a lot simpler to correctly implement than the alternatives. I know this is different from some of the other definitions that have been thrown around, but getting this consistent first is great.

app/models/patient.rb
```diff
@@ -86,14 +86,18 @@ class Patient < ApplicationRecord
     with_discarded.where(registration_facility: region.facilities)
   }
 
-  def self.follow_ups_with(model_name, period, time_column: "recorded_at", last: nil)
+  def self.follow_ups_with(model_name, period, time_column: "recorded_at", region: nil, last: nil)
```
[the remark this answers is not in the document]
harimohanraj89: Yup, I'm valuing consistency right now 😄 The ActivityService specs hammer this thing's edge cases right now. I'm yet to look at the Patient specs themselves.

app/services/activity_service.rb
```diff
@@ -0,0 +1,59 @@
...
+    relation = relation.group_by_period(period, :recorded_at, last: last)
+    relation = relation.group(group) if group.present?
+    relation.count
+  end
+
+  def follow_ups
+    relation = follow_ups_relation
+    relation = relation.group(group) if group.present?
+    relation.count
+  end
```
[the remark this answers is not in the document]
harimohanraj89: Yeah, same!

harimohanraj89:
Test failures point to likely real problems. I'll dig in.

rsanheim (approved):
Gonna throw an approval on here because this is clearly a big step forward.  Nice refactor !

rsanheim:
Some note on the failures I see (and wow, is there a way to copy paste from semaphore w/o the color codes?

```
1) Reports::RegionsController show retrieves district data04:50
     Failure/Error: expect(data[:controlled_patients].size).to eq(9) # sanity check​04:50
     ​04:50
       expected: 9​
            got: 10

       (compared using ==)
     # ./spec/controllers/reports/regions_controller_spec.rb:139:in `block (3 levels) in <top (required)>'​04:50
```
(a few of these happened)
      
I was mostly using this to get a handle on how much data was getting returned, and less as a 'the result size must always be n' type test.  The difference of one is weird, but maybe expected w/ other changes here?

In any case, if this change makes sense I wouldn't worry about changing the spec to the new size.

harimohanraj89:
Thanks for the context on the specs @rsanheim. I'll remove the sanity checks if needed (I think it's some "current period" stuff, probably).

There's also one bad query in there, I'll def take a look at that. Some fun arel usage probably awaits.

rsanheim:
```
ActionView::Template::Error:​04:50
       Reports::Result does not have #dig method
```

This has to do w/ the Result interface changes that happened in master, probably just need `to_hash`.  Make sure latest from master is merged first though, as I made changes around this in other PRs.

rsanheim:
> Thanks for the context on the specs @rsanheim. I'll remove the sanity checks if needed (I think it's some "current period" stuff, probably).

Ohhh right.  That may be a thing we change back for the reports. I'm not sure we want to show the 'current period' by default on those pages (when there are no params)...as I think the default view should be the last full completed month.

PR #1428, opened 2020-10-14
Change Region reports to cache all data and return only what is requested

rsanheim:
**Story card:** [ch1662](https://app.clubhouse.io/simpledotorg/story/1662/improve-dashboard-caching)

We currently cache the full set of BP control data for every region for the past 24 months. The problem with this is that admins may request an older set of data, for example, they can change the month dropdown to two months to request a set of data that doesn't match what we have in the cache.

This results in us retrieving _all_ the control data for the associated regions, and then caching it all as well. So we do a bunch of unnecessary SQL and also cache a lot of duplicated data (in this case, it would be duping 22 months of data).

Here we are changing our caching strategy to cache **all** the data for regions, and then controllers or APIs can retrieve just the data they need from the cache.

## Checklist

* [x] How to set a good cache key?
* [x] More testing w/ the other analytics / API code paths that use ControlRateService


app/models/reports/result.rb
```diff
@@ -71,44 +100,57 @@ def last_key(key)
...
+        current_registrations = registrations[period]
+        total = current_registrations + previous_registrations
+        running_totals[period] = total
+      }
+    end
+
     # "Missed visits" is the remaining registerd patients when we subtract out the other three groups.
-    def count_missed_visits
-      self.missed_visits = range.each_with_object({}) { |(period, visit_count), hsh|
+    def calculate_missed_visits(range)
```
rsanheim, line 127: The fact that some methods like this one take a range to do calculations and others inside Result operate on `full_data_range` not is a sign that this Result object is doing too much and isn't the right abstraction. 

This result object handles the historical results for registrations, controlled/uncontrolled counts, and also missed visits. Some of these are fetched by the ControlRateService but other parts of the data are fetched by other services.  I think there is a follow up PR here to refactor towards a simpler, more focused result object that only cares about a set of data in form of `{period1 => datum1, period2 => datum2, period3 => datum3}`, etc. Then the outer layers will just assemble those results, and the caching parts and ranger filtering parts will become simpler.

app/services/control_rate_service.rb
```diff
@@ -10,60 +10,60 @@ def initialize(region, periods:)
     @region = region
     @facilities = region.facilities
     # Normalize between a single period and range of periods
-    @periods = if !periods.is_a?(Range)
+    @report_range = if !periods.is_a?(Range)
```
rsanheim, line 13: Rename here for clarity, as there are two ranges in play from this layer down - the "report range" clients ask for, and the "full_data_range" that exists for a Region. The full data region is the entire set of data from the first registration for a Region to the current date.

app/services/control_rate_service.rb
```diff
@@ -10,60 +10,60 @@ def initialize(region, periods:)
...
-      periods.each do |(period, count)|
-        results.controlled_patients[period] = controlled_patients(period).count
-        results.uncontrolled_patients[period] = uncontrolled_patients(period).count
-      end
-
-      results.calculate_percentages(:controlled_patients)
-      results.calculate_percentages(:uncontrolled_patients)
-      results
-    end
+    all_cached_data.report_data_for(report_range)
```
rsanheim, line 35: This is the major change here: we grab all the cached data and then ask the result to filter it to just the report range.

app/services/control_rate_service.rb
```diff
@@ -109,23 +109,12 @@ def bp_quarterly_query(period)
...
-  end
-
-  def periods_cache_key
-    value = if periods.is_a?(Range)
-      "#{periods.begin.value}/#{periods.end.value}"
-    else
-      period.value
-    end
-    "#{periods.begin.type}_periods/#{value}"
+    "#{self.class}/#{region.model_name}/#{region.id}/#{report_range.begin.type}/#{Date.current}"
```
rsanheim, line 117: Our cache key now only cares about the period type (:month or :quarter), and the date that the queries are run.

app/services/reports/region_service.rb
```diff
@@ -12,8 +12,7 @@ def initialize(region:, period:)
       @region = region
       @period = period
       start_period = period.advance(months: -(MAX_MONTHS_OF_DATA - 1))
-      @range = start_period..@period
-      @result = Result.new(@range)
```
rsanheim, line 16: remove some unnecessary early initialization here.

app/services/reports/region_service.rb
```diff
@@ -23,12 +22,12 @@ def initialize(region:, period:)
...
-      result.merge! ControlRateService.new(region, periods: range).call
+      result = ControlRateService.new(region, periods: range).call
       result.visited_without_bp_taken = NoBPMeasureService.new(region, periods: range).call
       result.calculate_percentages(:visited_without_bp_taken)
-      result.count_missed_visits
-      result.calculate_missed_visits_percentages
-      result.calculate_period_info
+      result.calculate_missed_visits(range)
+      result.calculate_missed_visits_percentages(range)
+      result.calculate_period_info(range)
```
rsanheim, line 30: We have to tell the Result what range we want the calculations for. 😟

lib/json_logger.rb
```diff
@@ -1,3 +1,6 @@
+# Don't merge the datadog trace info in some environments as it clutters up logs
+DATADOG_TRACE_INFO_ENABLED = !(Rails.env.development? || Rails.env.test?)
```
rsanheim, line 2: the trace info in dev can be really noisy, so setting it here as a constant to turn it off but making it easy to toggle back on. Sometimes the trace info is useful if you are debugging how the datadog tracing pieces are working.

spec/services/control_rate_service_spec.rb
```diff
@@ -246,10 +246,67 @@ def refresh_views
...
   end
 
-  it "has a reasonable cache key" do
-    periods = Period.month(july_2018)..Period.month(july_2020)
-    service = ControlRateService.new(facility_group_1, periods: periods)
-    expected_key = "ControlRateService/FacilityGroup/#{facility_group_1.id}/month_periods/#{july_2018.to_date}/#{july_2020.to_date}"
-    expect(service.send(:cache_key)).to eq(expected_key)
+  context "caching" do
+    let(:memory_store) { ActiveSupport::Cache.lookup_store(:redis_store) }
+    let(:cache) { Rails.cache }
```
rsanheim, line 251: These pieces were the driver for https://github.com/simpledotorg/simple-server/pull/1451

rsanheim (commented):
This PR is a response to seeing some requests for reports still hitting 2+ second response times, and also preparing the way for more flexible report time windows in the future.

I knew this would be a difficult change given some of the earlier assumptions made around caching report data. I'm not real happy with how complicated some of this turned out, but I think we can get things to a better place w/ followup refactorings. Check out my inline comments for some thoughts on that.

For context: we were tightly coupling the set of cached data with what the set of data _most_ users requested. We cached the past 24 months of control data by default, but if you requested a different window you wouldn't hit the cache and would have a bad time.

We really need to support variable time windows of data, but for performance and simplicity sake it makes sense to request and cache _all_ the data for a Region. But to do that means we grab the cached data in one step, and then filter that data down to what the client asked for in a second step. Thats what this PR does.

These changes get us to a much better place performance wise, but there is a lot to be done here to improve the clarity of the code.

rsanheim:
>  More testing w/ the other analytics / API code paths that use ControlRateService

The tests are passing for this part, but wanted to ping @harimohanraj89 and others to ask for help to get more confidence that the API analytics end points would be all good w/ these changes.

app/services/control_rate_service.rb
```diff
@@ -10,60 +10,60 @@ def initialize(region, periods:)
...
+  private
+
+  def all_cached_data
+    Rails.cache.fetch(cache_key, version: cache_version, expires_in: 7.days, force: force_cache?) {
+      fetch_all_data
     }
   end
 
+  def fetch_all_data
+    results.registrations = registration_counts
```
kitallis, line 47: It does feel like the `Result` object is doing too much, at least purely from an API perspective (if not inherent behavior, like you [mentioned](https://github.com/simpledotorg/simple-server/pull/1428/files#r508985228))?

It can currently,

* Mutate data that it stores from the outside, by assigning stuff in the data hashmap (eg. `results.earliest_registration_period=`)
* Read data off the data hashmap by period (eg. `adjusted_registrations_for`)
* Read a different kind of data, that it computes (eg. `count_cumulative_registrations`)
* Read aggregates from the data it stores (eg. `results.calculate_percentages`)
* Rinse the data it currently holds (eg. `fill_in_nil_registrations`)

Guardfile
```diff
@@ -57,8 +57,4 @@ guard :rspec, cmd: "bin/rspec" do
   watch(rails.spec_helper) { rspec.spec_dir }
   watch(rails.routes) { "#{rspec.spec_dir}/routing" }
   watch(rails.app_controller) { "#{rspec.spec_dir}/controllers" }
-
-  # Capybara features specs
-  watch(rails.view_dirs) { |m| rspec.spec.call("features/#{m[1]}") }
-  watch(rails.layouts) { |m| rspec.spec.call("features/#{m[1]}") }
```
harimohanraj89, line 63: Lol.

config/schedule.rb
```diff
@@ -31,6 +31,10 @@ def local(time)
   runner "MarkPatientMobileNumbers.call"
 end
 
+every :day, at: local("12:30am"), roles: [:cron] do
```
kitallis, line 34: What's this change? The `refresh_materialized_views` task is still under the `db` namespace right? This looks duplicated from line#26 above, but without the ns.

kitallis (commented):
Just a first pass to try and understand the code flow, will try and leave final thoughts once I'm done with the whole thing.

app/models/reports/result.rb
```diff
@@ -31,21 +40,40 @@ def [](key)
...
+
+    # Return all periods for the entire set of data for a Region - from the first registrations until
+    # the most recent period
+    def full_data_range
+      (earliest_registration_period..current_period)
     end
 
-    def merge!(other)
-      @data.merge! other
+    def fill_in_nil_registrations
```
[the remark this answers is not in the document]
rsanheim: We do - the reason being the differences between these too types of behavior, which parts of the view rely on right now (for better or for worse)

```
Hash.new(0).values.last => nil
Hash.new(0).values.size => 0

# hsh = grab registrations etc
fill_in_nil_registrations(hsh)
hsh.values.last => 0
hsh.values.size => 24
```

config/schedule.rb
rsanheim: This was probably accidental from a merge, I'll double check and clean up.

app/assets/javascripts/common/reports.js
```diff
@@ -275,20 +275,31 @@ function initializeCharts() {
   );
   cumulativeRegistrationsGraphConfig.options.tooltips = {
     enabled: false,
-    custom: function (tooltipModel) {
-      let period = null;
-      if(tooltipModel.dataPoints) {
-        let label = tooltipModel.dataPoints[0].label;
-        period = data.periodInfo[label];
+    custom: function (tooltip) {
+      console.log(this);
```
claudiovallejo, line 279: You missed removing these `console.log`s

app/services/control_rate_service.rb
rsanheim: I agree - my plan is to refactor towards a better, more focused Result object in subsequent PRs. This all got messy with the whole "cached result" vs "report result" differences that have come about.

claudiovallejo (approved):

PR #1430, opened 2020-10-15
Use the new cohort_patients key instead of registered

harimohanraj89:
**Story card:** [ch1600](https://app.clubhouse.io/simpledotorg/story/1600/fix-patients-column-in-cohort-downloads)

## Because

Some downloads were trying to use an older version of the cohort query results

## This addresses

Updating download touchpoints to use the correct cohort query results:
* Monthly cohort reports
* Quarterly cohort reports
* Whatsapp graphics (web and downloadable)

These touchpoints are using the `CohortAnalyticsQuery`. I think we should decommission that thing in favor of the `CohortService`. What do you think?

### DIFF
--- a/app/views/reports/regions/cohort.csv.erb
+++ b/app/views/reports/regions/cohort.csv.erb
@@ -16,7 +16,7 @@
 <% @cohort_analytics.each_with_index do |(report_dates, analytics), index| %>
   <%
     cohort_start, report_start = report_dates
-    registered = analytics.dig(:registered, :total) || 0
+    registered = analytics.dig(:cohort_patients, :total) || 0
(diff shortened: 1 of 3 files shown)


kitallis (approved):
ugh

no tests broke because of this change?

harimohanraj89:
Yuuuuup. I'm torn whether to introduce tests for this that will break, or wait until we replace this with the `CohortService`. What do you think?

harimohanraj89:
I'm gonna ship as-is, to fix the downloads asap.

kitallis:
That sounds good, was just going to respond 👍

PR #1431, opened 2020-10-15
Add cap task to run rails runner tasks

vkrmis:
**Story card:** none

## Because

We don't have a way to run `rails runner` tasks via capistrano

## This addresses

Adds a `cap deploy:runner` task


kitallis (approved):
lgtm

PR #1434, opened 2020-10-15
Clean up mat view refresh; change schedule to every night

rsanheim:
**Story card:** [ch1603](https://app.clubhouse.io/simpledotorg/story/1603/change-materialized-view-schedule-to-be-everynight)

Schedule it for everynight and clean up the code a bit for setting the timestamp.

### DIFF
--- a/app/controllers/my_facilities/ranked_facilities_controller.rb
+++ b/app/controllers/my_facilities/ranked_facilities_controller.rb
@@ -42,13 +42,7 @@ class MyFacilities::RankedFacilitiesController < AdminController
   private
 
   def set_last_updated_at
-    last_updated_at =
-      begin
-        Time.parse(Rails.cache.fetch(Constants::MATVIEW_REFRESH_TIME_KEY))
-      rescue TypeError, ArgumentError
-        nil
-      end
-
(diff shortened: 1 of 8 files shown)


app/controllers/my_facilities/ranked_facilities_controller.rb
```diff
@@ -42,13 +42,7 @@ def show
   private
 
   def set_last_updated_at
-    last_updated_at =
-      begin
-        Time.parse(Rails.cache.fetch(Constants::MATVIEW_REFRESH_TIME_KEY))
```
rsanheim, line 47: We don't need to parse things out of the cache - Rails does a Marshall.dump when setting things in the cache, so when we read it will come out as a `Time` correctly w/ the timezone and everything.

config/initializers/constants.rb
```diff
@@ -1,3 +0,0 @@
-module Constants
```
rsanheim, line 1: No need for this anymore.

harimohanraj89 (approved):
Love it. Yeah, class-ifying this was the right move. That's why it was so awkward to set up those constants and stuff 😅

PR #1435, opened 2020-10-15
Bug fix for password update: Use safe navigator in case any of these objects are nil

rsanheim:
[ch1602](https://app.clubhouse.io/simpledotorg/story/1602/warden-callback-breaks-on-bad-password-update)

Bug fix for https://sentry.io/organizations/resolve-to-save-lives/issues/1958880545/?project=1217715&referrer=slack


harimohanraj89 (approved):
Oh, this was our code? I mistakenly assumed it was Warden internals.

Yup, this makes sense to me. What's the condition where this `user` was nil, causing exceptions?

rsanheim:
> Yup, this makes sense to me. What's the condition where this `user` was nil, causing exceptions?

I'm guessing it would be an orphaned email_auth record, or maybe edge case where a password update fails due to mismatched passwords? 

We are only setting this user id into the RequestStore so that DataDog can then submit it with logs, so we definitely don't want this to ever fail w/ bad data.

PR #1436, opened 2020-10-16
Bathinda and Mansa support

harimohanraj89:
**Story card:** [ch1657](https://app.clubhouse.io/simpledotorg/story/1657/separate-reports-for-bathinda-and-mansa)

## Because

Sometimes a facility group doesn't correspond to a district.
* Bathinda
* Mansa
* Chennai

## This addresses

Adding a hidden report for a "Facility District" - a district based on facility addresses - that duck-types the Region interface.

This powers:
* https://api.simple.org/reports/regions/facility_district/Bathinda
* https://api.simple.org/reports/regions/facility_district/Mansa
* https://api.simple.org/reports/regions/facility_district/Chennai

which assembles facilities based on their address's district, instead of their facility group.

### Todo

* [x] Fix downloads
* [x] Add test coverage


app/models/facility_district.rb
```diff
@@ -0,0 +1,37 @@
...
+  def facilities
+    scope.where(district: name)
+  end
+
+  def dashboard_analytics(period:, prev_periods:)
+    query = DistrictAnalyticsQuery.new(name, facilities, period, prev_periods, include_current_period: true)
+    query.call
+  end
+
+  def cohort_analytics(period:, prev_periods:)
```
vkrmis, line 21: So these query calls will piggyback on the cache mechanism in the queries?
Nice!

app/models/facility_district.rb
```diff
@@ -0,0 +1,37 @@
+class FacilityDistrict
+  attr_reader :name, :scope
+
+  alias_method :id, :name
+  alias_method :to_param, :name
+
+  def initialize(name:, scope: Facility.all)
```
vkrmis, line 7: Should we scope by Org as well here? Only realistically an issue in Maharashtra, if there are overlapping districts between PATH and IHCI.

vkrmis: I checked this, there aren't any overlapping districts. Should be okay for a quick fix.

vkrmis (approved):
lgtm

app/models/facility_district.rb
harimohanraj89: Yup that was my exact guess as well. Thanks for confirmation.

rsanheim (approved):

app/models/facility_district.rb
harimohanraj89: Yup, just like they do on `FacilityGroup` and `Facility` already! 😄

PR #1438, opened 2020-10-16
Spring cleaning

harimohanraj89:
**Story card:** TBD

* Remove a bunch of anonymized mailer/download stuff
* Remove the `OrganizationDistrict`

### Reviewers

Can you please sanity check that the things removed are truly not needed? Are they still being used by some code paths, or used manually in a console or rake task somewhere?


kitallis (approved):
this lgtm, don't think any of this stuff is used anymore.

PR #1439, opened 2020-10-16
Make Block a required field

kitallis:
**Story card:** [ch1343](https://app.clubhouse.io/simpledotorg/story/1343/clean-up-existing-facility-block-zone-data)

## Because

We are soon going to be cleaning up all our block data to be first-class models. But in the meanwhile, let's at least disallow people from leaving empty blocks because they are harder to clean-up later.

## This addresses

Makes the block (or zone) field mandatory in upload / UI.

### DIFF
--- a/app/models/facility.rb
+++ b/app/models/facility.rb
@@ -64,6 +64,7 @@ class Facility < ApplicationRecord
 
   auto_strip_attributes :name, squish: true, upcase_first: true
   auto_strip_attributes :district, squish: true, upcase_first: true
+  auto_strip_attributes :zone, squish: true, titleize: true
 
   with_options if: :import do |facility|
     facility.validates :organization_name, presence: true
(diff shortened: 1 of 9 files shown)


prabhanshuguptagit (approved):

app/models/facility.rb
```diff
@@ -81,6 +82,7 @@ class Facility < ApplicationRecord
   validates :district, presence: true
   validates :state, presence: true
   validates :country, presence: true
+  validates :zone, presence: true
```
rsanheim, line 85: This should only be validated on `create`, correct? Otherwise we are going to have invalid facilities in prod that have empty zones.

rsanheim: discussed on slack, this is desired so that if admins do edit any facilities, they will be forced to enter a zone.  Facility updates are very rare in production.

PR #1440, opened 2020-10-16
Reports: Facility Tables copy fix

claudiovallejo:
**Story card:** [ch1585](https://app.clubhouse.io/simpledotorg/story/1585/facility-table-copy-fix)

## Because

The facility tables denominator copy is not right.

## This addresses

**Updated facility tables denominator copy in District Reports**

image


kitallis (approved):
looks good to me 👍

PR #1442, opened 2020-10-16
Try new branch for new Heroku / env var fix

rsanheim:
Maybe we need a brand new PR to pick up the ENV var that I've added to the 'parent' app in Heroku? I'm confused as to why its just breaking on this one Sentry env var.


rsanheim:
nope this isn't the problem.

PR #1445, opened 2020-10-19
Telemed: Add request status to API

prabhanshuguptagit:
**Story card:** [ch1644](https://app.clubhouse.io/simpledotorg/story/1644/add-request-status-to-teleconsultation)

## Because

We want nurses to be able to set the request's completion status as part of the teleconsultation request.

## This addresses

Add the `requester_completion_status` field to the API.

image


app/models/teleconsultation_medical_officer.rb
```diff
@@ -1,3 +1,3 @@
 class TeleconsultationMedicalOfficer < User
-  default_scope { joins(:teleconsultation_facilities) }
+  default_scope { joins(:teleconsultation_facilities).distinct }
```
prabhanshuguptagit, line 2: This is an unrelated tiny fix.

app/schema/api/v4/models.rb
```diff
@@ -254,7 +254,8 @@ def teleconsultation
             properties: {
               requester_id: {"$ref" => "#/definitions/uuid"},
               facility_id: {"$ref" => "#/definitions/uuid"},
-              requested_at: {"$ref" => "#/definitions/timestamp"}
+              requested_at: {"$ref" => "#/definitions/timestamp"},
+              request_completed: {type: [:string, "null"], enum: Teleconsultation.request_completeds.keys << nil}
```
vkrmis, line 258: Lets add 
```
ActiveSupport::Inflector.inflections do |inflect|
  inflect.irregular "request_completed", "requests_completed"
end
```
to config/initializers/inflections.rb

That way we can use `Teleconsultation.requests_completed` instead of `Teleconsultation.request_completeds`

vkrmis (commented):
Do we want to separately infer completion if we know the request has been completed?
Eg: The nurse has not marked completion, but the server received the MO's log. That'll let us use this field as the canonical indicator for reporting the number of completed requests etc.

harimohanraj89:
I'd go the other way. Capture events as they happen, and draw inferences in the reporting layer. That keeps our data flexible enough to answer questions like: How many times is there inconsistency between the nurse's report and the doctors, etc etc

prabhanshuguptagit:
> Capture events as they happen, and draw inferences in the reporting layer.

I also would like to add something like this when we start working on telemed reporting:

```
def self.completed
  where(request_completed: :yes).or(where('recorded_at IS NOT NULL'))
end
```

vkrmis:
Sorry for nit-picking this, but imo, in that case we should rename this field to disambiguate it from completion. Maybe `marked_complete_by_requester` or something. Otherwise we end up with two very similar/overlapping definitions of `completed`, and that confusion gets transmitted to conversations, reporting copy and everything downstream.

harimohanraj89:
I like that suggestion a lot 👍

harimohanraj89:
PS: Wherever we land up, @prabhanshuguptagit can you document these fields in our [data dictionary](https://docs.google.com/spreadsheets/u/1/d/1LOn9IzfuQXog2agGbNphweMNTV27K84yctjPs6b7USI/edit?usp=drive_web&ouid=108027733417617261123) please? 🙂

prabhanshuguptagit:
Makes sense. `marked_complete_by_requester` would be incorrect because it does not correspond to the `yes`/`no`/`waiting` status filled in by the nurse, can't think of a better name here. `request_completion_status`? 🤔

app/schema/api/v4/models.rb
prabhanshuguptagit: `requests_completed` sounds like a scope (of completed requests), shall we let this be since it's only going to be used in one place?

prabhanshuguptagit:
Called it `requester_completion_status` in 49604bf.

harimohanraj89 (approved):
Yup, this looks great 👍

PR #1448, opened 2020-10-19
Ranked facility fixes

timcheadle:
**Story card:** [ch1648](https://app.clubhouse.io/simpledotorg/story/1648/add-score-and-fix-registration-rate-bugs)

## Because

Registration rate should never exceed 100%. Also, we should show the score and fix OPD estimates.

## This addresses

- Show a `Score` column
- Max out registration rate at 100%
- Improve OPD estimates:
   - Community = 450/month
   - Small = 1800
   - Medium = 3000
   - Large = 7500

### DIFF
--- a/spec/models/reports/performance_score_spec.rb
+++ b/spec/models/reports/performance_score_spec.rb
@@ -102,6 +102,12 @@ describe Reports::PerformanceScore, type: :model do
       @opd_load = 0
       expect(perf_score.registrations_rate).to eq(0)
     end
+
+    it "maxes at 100 if registrations exceeds target" do
+      @registrations = 500
+      @opd_load = 100
+      expect(perf_score.registrations_rate).to eq(100)
+    end
   end
 
   describe "#target_registrations" do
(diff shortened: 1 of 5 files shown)


spec/models/reports/performance_score_spec.rb
```diff
@@ -102,6 +102,12 @@
       @opd_load = 0
       expect(perf_score.registrations_rate).to eq(0)
     end
+
+    it "maxes at 100 if registrations exceeds target" do
+      @registrations = 500
+      @opd_load = 100
```
harimohanraj89, line 108: Ivars for test setup? #derty 😅 

TBH it's quite neat.

harimohanraj89 (approved):
Makes sense to me ✨

rsanheim (approved):
I'm guessing the odd test setup is because Reports::Result is a bit tricky to fill w/ fake data?

spec/models/reports/performance_score_spec.rb
timcheadle: It's because otherwise I'd have to construct a new `double` for every test. If you know a better way please show! :)

PR #1449, opened 2020-10-19
Fix facility upload templates to require zone_or_block

timcheadle:
## Because

https://github.com/simpledotorg/simple-server/pull/1439 made zone a required field, but we never updated the CSV templates

## This addresses

Fixes both facility upload CSV templates to require `zone_or_block`


harimohanraj89 (approved):
Ah. Yes. :shipit:

PR #1452, opened 2020-10-20
Simplify the syncing codepaths

kitallis:
**Story card:** [ch1655](https://app.clubhouse.io/simpledotorg/story/1655/make-some-sync-logic-improvements-to-make-way-for-region-syncs)

## Because

There's a bunch of logical redundancy and cruft around how the sync mechanism works; makes it hard to understand the codepaths.

## This addresses

Remove the `PrioritisableByFacility` mixin in favor of simply delegating it to `SyncToUser` and make way for easily modifying the `sync_region` for all the sync resources in a generic way.

Additionally, we've modified how `current_facility_records` and `other_facility_records` are synced. They are synced scoped by the patients in the `current_facility`, rather than the facility itself. This is conceptually more accurate (albeit equivalent in terms of data to the previous change we made in https://github.com/simpledotorg/simple-server/pull/1375).


app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -1,20 +1,22 @@
 module Api::V3::SyncToUser
   extend ActiveSupport::Concern
+
   included do
```
kitallis, line 4: Instead of an additional mixin called `PrioritisableByFacility` (or `PrioritisableByPatient` it should've been called now), we stick the default and most common behavior in `SyncToUser` for `current_facility_records`, `other_facility_records`.

Certain controllers like Facilities and Protocols can override it to avoid syncing scoped by `region_records` since we sync all of them all the time.

Notice that we've also changed the scope to be `patient` here instead of `facility`; this is more in-line with a proper patient-centric sync.

app/controllers/api/v3/appointments_controller.rb
```diff
@@ -15,10 +13,6 @@ def metadata
 
   private
 
-  def facility_group_records
```
kitallis, line 18: We've removed these individual calls to `syncable_to_region` in favour of a `region_records` that behaves the exact same way [everywhere](https://github.com/simpledotorg/simple-server/pull/1452/files#diff-40d0adce51092a7a607ada431136b5c8dc5447e754449277afd39d5a42c5a826R5). Exceptions are added as necessary to a couple of controllers.

This keeps the proximity of `region_records` with the methods that actually make use of it. So now,

* `region_records`
* `current_facility_records`
* `other_facility_records`

All, fly together, rather than cut awkwardly in different places among different controllers.

app/controllers/api/v3/protocols_controller.rb
```diff
@@ -9,11 +9,17 @@ def sync_to_user
     __sync_to_user__("protocols")
   end
 
-  def find_records_to_sync(since, limit)
```
kitallis, line 12: Dead code.

app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -1,20 +1,22 @@
...
-      current_facility_group
-        .send(model_name.name.underscore.pluralize.to_sym)
-        .with_discarded
+    def region_records
+      model_name.syncable_to_region(current_sync_region)
     end
 
     def current_facility_records
-      []
+      region_records
```
kitallis, line 10: Instead of an additional mixin called `PrioritisableByFacility` (or `PrioritisableByPatient` it should've been called now), we stick the default and most common behavior in `SyncToUser` for `current_facility_records`, `other_facility_records`.

Certain controllers like Facilities and Protocols can override it to avoid syncing scoped by `region_records` since we sync all of them all the time.

Notice that we've also changed the scope to be `patient` here instead of `facility`; this is more in-line with a proper patient-centric sync.

It used to be,

```ruby
current_facility_records.where(facility: current_facility).updated_since...
```

Which was: _give me all the blood_pressures for patients in the current_facility_group AND filter them by the current_facility_.

Now it's,

```ruby
region_records.where(patient: Patients.in_current_facility).updated_since...
```

Which is: _give me all the blood_pressures for patients in the current sync region AND filter them by patients in the current_facility_.

kitallis: This is potentially slower (we'll do some perf tests), but conceptually more correct.

app/controllers/api/v3/patients_controller.rb
```diff
@@ -13,22 +13,23 @@ def request_metadata
...
   end
 
   def other_facility_records
     other_facilities_limit = limit - current_facility_records.count
-    facility_group_records
-      .includes(:address, :phone_numbers, :business_identifiers)
-      .where.not(registration_facility: current_facility.id)
+
+    region_records
+      .where.not(registration_facility: current_facility)
```
harimohanraj89, line 32: Makes sense. Patients are special because they're central to sync boundaries now.

harimohanraj89 (approved):
This look slickkkkkk ✨ 

Very intuitive refactor, nice spec updates.

I reckon the next area to rejigger soon will be the use of the time cursor. I _think_ we'll soon need to decouple the cursor from the `current_facility_records/other_facility_records` methods so that we can compose results nicely. We will soon need to start ignoring the cursor sometimes.

prabhanshuguptagit:
Tested the sync times on datadog. All seems well with the patient scoping change.

Screenshot 2020-10-28 at 3 43 09 PM
[red - primary]
[orange - feature]

PR #1454, opened 2020-10-21
Import canonical zones into Region

prabhanshuguptagit:
**Story card:** [ch1667](https://app.clubhouse.io/simpledotorg/story/1667/import-canonical-blocks-into-region)

## Because

We have an exhaustive list of blocks for IHCI which we want to use to populate the Zone field in the Edit Facility form.

## This addresses

This adds a task to import the canonical blocks list into the Region model. It associates the blocks to the correct Org -> _District > Zone_ hierarchy.

Since the import task depends on the _Country > Org_ hierarchy to exist, it should be run after the [region backfill](https://github.com/simpledotorg/simple-server/blob/master/app/services/region_backfill.rb) has been run. More organizations across countries can be added and the same task can be used later to import zones.


app/services/region_backfill.rb
```diff
@@ -44,8 +44,8 @@ def create_regions
     end
     root_type = find_region_type("Root")
     org_type = find_region_type("Organization")
-    facility_group_type = find_region_type("FacilityGroup")
-    block_type = find_region_type("Block")
```
rsanheim, line 48: I was thinking about this some more and had some good discussions w/ @claudiovallejo about upcoming block level reports.  

I think that if we are using the term "Block" in the UI for IHCI reports, we should stick with Block as the terminology in our code. Its the word that we will be using most often when discussing with designers or users of the system, so it makes sense to keep the name in our Region hierarchy consistent with that.

We can also watch for [an answer](https://simpledotorg.slack.com/archives/CK95QFJA2/p1603305906080800) from Praveen to help get some more confidence in this name.

app/services/region_backfill.rb
```diff
@@ -44,8 +44,8 @@ def create_regions
...
+    facility_group_type = find_region_type("District")
```
rsanheim, line 47: 👍

app/services/region_backfill.rb
```diff
@@ -85,9 +85,9 @@ def create_region_types
     unless dry_run?
       root = RegionType.create! name: "Root", path: "Root"
       org = RegionType.create! name: "Organization", parent: root
-      facility_group = RegionType.create! name: "FacilityGroup", parent: org
-      block = RegionType.create! name: "Block", parent: facility_group
-      _facility = RegionType.create! name: "Facility", parent: block
+      facility_group = RegionType.create! name: "District", parent: org
+      zone = RegionType.create! name: "Zone", parent: facility_group
```
rsanheim, line 89: I think we should stick with Block here to keep a [ubiquitous language](https://www.martinfowler.com/bliki/UbiquitousLanguage.html) with our primary deployment.

lib/tasks/scripts/import_zones_to_regions.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def create_region_from(name:, type:, parent:)
+    region = Region.new(name: name, type: type)
+    region.path = [parent.path, region.name_to_path_label].compact.join(".")
+    region.save!
+    region
+  end
+
+  def log(message)
+    puts message if verbose
```
rsanheim, line 65: Can you also log these to the Rails logger, even if verbose is false? You can setup a child logger similar to how the RegionBackfill does, which will let us see all the log messages easily in datadog.

config/data/india/ihci/canonical_zones.yml
```diff
@@ -0,0 +1,379 @@
+---
```
rsanheim, line 1: I think after thinking on this a bit I agree with y'all that we should just import all the canonical Regions...it will definitely help us standardize things going forward.

rsanheim (changes requested):
My main bit of feedback is to keep the `Block` name and add more logging, otherwise I think this is a good approach.

lib/tasks/scripts/import_zones_to_regions.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  private
+
+  memoize def organization_region
+    Region.find_by!(name: organization_name, type: RegionType.find_by!(name: "Organization"))
+  end
+
+  memoize def zone_region_type
+    RegionType.find_by!(name: "Zone")
```
rsanheim, line 42: `name: "Block"`

lib/tasks/scripts/import_zones_to_regions.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+  end
+
+  private
+
+  memoize def organization_region
+    Region.find_by!(name: organization_name, type: RegionType.find_by!(name: "Organization"))
+  end
+
+  memoize def zone_region_type
```
rsanheim, line 41: `block_region_type`

app/services/region_backfill.rb
```diff
@@ -44,28 +44,37 @@ def create_regions
...
+        result[facility.state][facility.district] ||= {}
+        result[facility.state][facility.district][facility.zone] ||= []
+        result[facility.state][facility.district][facility.zone] << facility
+      }
+
+      org_region = create_region_from(source: org, region_type: org_type, parent: root_region)
+      hierarchical_facilities.map do |state, districts|
+        state_region = create_region_from(name: state, region_type: state_type, parent: org_region)
+        districts.map do |district, zones|
+          district_region = create_region_from(name: district, region_type: district_type, parent: state_region)
```
rsanheim, line 67: I noticed Facility#district is required at the model level, but not schema level.  Probably something we should make not null.

app/services/region_backfill.rb
```diff
@@ -44,28 +44,37 @@ def create_regions
...
+          zones.map do |zone, facilities|
```
rsanheim, line 68: Can you do a `gsub(/zone/, 'block')` in this file? All the names here were `block` in here previously, which will make it easier to keep things consistent and also to find in the logs as we go forward.

app/services/region_backfill.rb
```diff
@@ -83,13 +92,14 @@ def find_region_type(name)
   def create_region_types
     logger.info msg: "create_region_types"
     unless dry_run?
-      root = RegionType.create! name: "Root", path: "Root"
-      org = RegionType.create! name: "Organization", parent: root
-      facility_group = RegionType.create! name: "FacilityGroup", parent: org
-      block = RegionType.create! name: "Block", parent: facility_group
-      _facility = RegionType.create! name: "Facility", parent: block
+      RegionType::HIERARCHY.reduce(nil) do |parent_region_type, region_type|
```
rsanheim, line 95: Do you need to reduce with a `nil` here, or could you just use `each` ? I think things would be more clear if you special cased the Root region type explictly here.

lib/tasks/import_zones_to_regions.rake
```diff
@@ -0,0 +1,8 @@
+require "tasks/scripts/import_zones_to_regions"
+
+desc "Import zones from canonical list to Region"
+task :import_zones_to_regions, [:organization_name, :dry_run] => :environment do |_t, args|
+  abort "Requires <organization_name>" unless args[:organization_name].present?
+
+  ImportZonesToRegions.import(args[:organization_name], dry_run: args[:dry_run])
```
rsanheim, line 7: `gsub(/Zones/, "Blocks")`

app/services/region_backfill.rb
prabhanshuguptagit: I left these ones to be zones intentionally, since they represent the Facility field that are being copied over to the region `block`. Feel free to change.

app/models/facility.rb
```diff
@@ -79,6 +79,8 @@ class Facility < ApplicationRecord
     facility.validates :name, presence: true
   end
 
+  alias_attribute :block, :zone
```
rsanheim, line 82: @prabhanshuguptagit I added this to make life easier as we get the language more consistent.  We should be able to migrate to the new field over time via this alias ( and we can do the actual schema rename after the region backfill).

rsanheim (approved):
Got names a bit more consistent and added more tests for the State creation.

@prabhanshuguptagit I went w/ a slightly simpler approach for the backfill, I create all the States first and then just iterate through every facility and create Districts as needed.  It was a bit simpler to get tests passing with that approach.

PR #1455, opened 2020-10-21
Populate facility zone via Region

prabhanshuguptagit:
**Story card:** [ch1328](https://app.clubhouse.io/simpledotorg/story/1328/change-facility-zone-field-to-be-a-drop-down-instead-of-text-field)

## Because

We want to regulate the Zone data from the Facility edit form to be clean and correct.

## This addresses
image

This makes the `Zone` field a drop-down that is populated via the Region model. 

It is feature-flagged off (using flipper) so it can be merged right away. If flipper flag is off, there's no dropdown and it'd be just a normal text field like status quo.

## Next steps

This is just a small step in the direction of making Regions first-class to support Region-level syncing. We need to do a bunch of other things to execute this project successfully:

* We should populate the missing Block data
* Run the `RegionBackfill` task
* Enable this flag
* Add some callbacks in all our existing Region-aware models to create first-class Regions for them on create/update
* Implement Region-level sync


app/models/region.rb
```diff
@@ -31,4 +31,28 @@ def log_payload
     attrs["errors"] = errors.full_messages.join(",") if errors.any?
     attrs.symbolize_keys
   end
+
+  # These methods are generated on class load and updates to RegionTypes
+  # will not be available until the class is reloaded.
+  RegionType.all.map do |region_type|
```
harimohanraj89, line 37: Will a surgical `method_missing` be more robust against changing RegionTypes? I know there's a stigma attached to it, like rescuing errors, but I think it's a useful tool when applied properly.

**Update:** On second thought, Region Types should be _super_ stable in any given deployment. As you were, gentlemen.

app/views/admin/facilities/_form.html.erb
```diff
@@ -12,7 +12,11 @@
       <h3 class="mt-5">Address</h3>
       <%= form.text_field :street_address, id: :facility_street_address %>
       <%= form.text_field :village_or_colony, id: :facility_village_or_colony %>
-      <%= form.text_field :zone, id: :facility_zone, required: true %>
+      <% if Flipper.enabled?(:facility_zone_dropdown) %>
+        <%= form.select :zone, @available_zones.zip(@available_zones), {id: :facility_zone, include_blank: true}, required: true %>
```
harimohanraj89, line 16: Zip 👏

app/controllers/admin/facilities_controller.rb
```diff
@@ -5,6 +5,7 @@ class Admin::FacilitiesController < AdminController
 
   before_action :set_facility, only: [:show, :edit, :update, :destroy]
   before_action :set_facility_group, only: [:show, :new, :create, :edit, :update, :destroy]
+  before_action :set_available_zones, only: [:new, :create, :edit, :update]
```
harimohanraj89, line 8: It might be cleaner to put this behind the feature flag too. I know we _should_ be able to set these zones and not use them (famous last words), but I'm unclear how this'll play out in places we haven't run the backfill yet (eg. Bangladesh)

app/controllers/admin/facilities_controller.rb
```diff
@@ -118,6 +119,20 @@ def set_facility_group
...
+        facility_region.district.blocks
+      else
+        Region.where(type: RegionType.find_by_name("Block"))
+      end
+
+    @available_zones = zones.order(:name).pluck(:name)
+  end
+
+  def facility_region
+    @facility_region ||= Region.find_by(source_id: @facility.id) if @facility
```
harimohanraj89, line 134: `if @facility.id`? The `new` action could have a blank `@facility` object with a `nil` ID.

I know it's not the case right now (since the creation happens in the controller action), but it's fairly common practice to extract that kind of stuff to a `before_action`, and I wouldn't want this to break if we do that.

app/controllers/admin/facilities_controller.rb
```diff
@@ -118,6 +119,20 @@ def set_facility_group
     @facility_group = current_admin.accessible_facility_groups(:manage).friendly.find(params[:facility_group_id])
   end
 
+  def set_available_zones
+    zones =
+      if facility_region
+        facility_region.district.blocks
+      else
+        Region.where(type: RegionType.find_by_name("Block"))
```
harimohanraj89, line 127: When we don't have a specific facility, we still have the context of the facility group available (see method above). We should limit the selectable blocks to one of the blocks in the facility group (district).

harimohanraj89 (changes requested):

app/models/region.rb
kitallis: Yeah, considered it. Came to the same conclusion as you.

app/controllers/admin/facilities_controller.rb
prabhanshuguptagit: Oops. Yeah, we _should_ feature flag this, since this will ship before the backfill is run anyways (even on IHCI).

kitallis (approved):
lgtm

app/services/region_backfill.rb
```diff
@@ -68,8 +68,8 @@ def create_regions
           count_invalid(block_type)
           logger.info msg: "Skipping creation of Facility #{facility.name} because the block field (ie zone) is blank",
                       error: "block is blank",
-                      block_name: block_name,
-                      facilities: facilities.map(&:name)
+                      block_name: facility.block,
```
kitallis, line 71: Thanks, all this was breaking for me too.

PR #1462, opened 2020-10-23
My Facilities: Mobile-friendly filters

claudiovallejo:
**Story card:** [ch1680](https://app.clubhouse.io/simpledotorg/story/1680/my-facilities-filter-improvements)

## Because

Filtering on desktop is not intuitive or obvious. This new design is responsive and mobile-friendly.

## This addresses

**Selecting an option from any filter (block, facility, or period) submits a new form**

image

**Add block and facility size filters to "Ranked Facilities"**

image

image

**Add block and facility size filters to "BP control"**

image

image

**Add block and facility size filters to "Missed visits"**

image

image

**Add block and facility size filters to "Registrations"**

image

image


dburka:
Not blocker @claudiovallejo , but "Block" isn't sticky when you navigate between sections of My Facilities. "Facility Size" is nice and sticky tho!

Otherwise this looks good.

dburka:
Also, really minor thing @claudiovallejo but I used H5 headers in the new cards on the overview screen of My Facilities. Can you make them match the other cards with H3?

app/controllers/concerns/my_facilities_filtering.rb
```diff
@@ -38,11 +38,7 @@ def set_only_new_facilities
     end
 
     def facilities_by_size(facilities)
-      if (@facility_sizes - @selected_sizes).empty?
-        facilities
-      else
-        facilities.where(facility_size: @selected_sizes)
-      end
+      facilities.where(facility_size: @selected_sizes).or(facilities.where(facility_size: nil))
```
rsanheim, line 41: Why grab facilities with nil sizes?

rsanheim (approved):
Approving to unblock, tho I did have one question on one of the scopes that seems a little weird.

I would definitely kick the tires on sandbox with this quite a bit, as it is changing quite a few things. Holler in slack and I can help QA it as well.

app/controllers/concerns/my_facilities_filtering.rb
claudiovallejo: Thanks @rsanheim! Just realizing facility size is a mandatory field in the UI. Removing this!

claudiovallejo:
> Not blocker @claudiovallejo , but "Block" isn't sticky when you navigate between sections of My Facilities. "Facility Size" is nice and sticky tho!
> 
> Otherwise this looks good.

Ah yes, thanks Daniel! Fixed!

app/views/my_facilities/missed_visits.html.erb
```diff
@@ -1,40 +1,64 @@
...
+            aria-haspopup="true"
+            aria-expanded="false"
+          >
+            <%= PeriodSelection::PERIODS[:missed_visits][@selected_period] %>
+          </a>
+          <div class="dropdown-menu" aria-labelledby="dropdownMenuLink">
+            <% PeriodSelection::PERIODS[:missed_visits].each do |period, period_name| %>
+              <button
+                class="dropdown-item filter-by-period <%= "active" if period.to_s == @selected_period.to_s %>"
+                data-value="<%= period %>"
```
prabhanshuguptagit, line 46: The period selection JS at the bottom can be removed by tweaking the button to submit the form:
```
<button 
  class="dropdown-item <%= "active" if period.to_s == @selected_period.to_s %>"
  value="<%= period %>"
  type="submit" 
  name="period" 
  form="query-filters">
  <%= period_name %>
</button>
```

app/views/shared/_size_and_zone_filters.html.erb
```diff
@@ -0,0 +1,86 @@
...
+        class="dropdown-item <%= "active" unless @selected_zones.size == 1 %> filter-by-zone"
+        data-value=""
+      >
+        All blocks
+      </button>
+      <div class="dropdown-divider"></div>
+      <% @zones.each do |zone| %>
+        <button
+          class="dropdown-item query-filter-zone <%= "active" if @selected_zones.size == 1 && @selected_zones.first == zone %> filter-by-zone"
+          data-value="<%= zone %>"
```
prabhanshuguptagit, line 30: `my_facilities_filters.js` can also be removed in a similar fashion as above by making this:

```
<button
  class="dropdown-item query-filter-zone <%= "active" if @selected_zones.size == 1 && @selected_zones.first == zone %>"
  value="<%= zone %>"
  type="submit"
  name="zone"
  form="query-filters">
  <%= zone.capitalize %>
</button>
```

app/views/shared/_size_and_zone_filters.html.erb
```diff
@@ -0,0 +1,86 @@
...
+    aria-haspopup="true"
+    aria-expanded="false"
+  >
+    <% if @selected_zones.size == 1 %>
+      <%= @selected_zones.first.capitalize %>
+    <% else %>
+      All blocks
+    <% end %>
+  </a>
+  <div class="dropdown-menu" aria-labelledby="dropdownMenuLink">
```
prabhanshuguptagit, line 18: The zones dropdown seems to `overflow: scroll` on mobile but it's getting huge on desktop. Shall we set a  `max-height` on desktop as well?
image

prabhanshuguptagit (changes requested):

app/controllers/concerns/my_facilities_filtering.rb
```diff
@@ -38,11 +38,7 @@ def set_only_new_facilities
     end
 
     def facilities_by_size(facilities)
-      if (@facility_sizes - @selected_sizes).empty?
-        facilities
-      else
-        facilities.where(facility_size: @selected_sizes)
-      end
+      facilities.where(facility_size: @selected_sizes)
```
prabhanshuguptagit, line 41: @claudiovallejo Facility size was recently made a mandatory field, so there's still old data without any `size` set. I think we'll have to keep the `if (@facility_sizes - @selected_sizes).empty?` clause for when all sizes are selected.

app/views/my_facilities/missed_visits.html.erb
claudiovallejo: @prabhanshuguptagit thanks for the feedback! If we remove the JS at the bottom, the selected period is cleared if the user selects an option from a different dropdown. We need the JS to set the right values in the hidden inputs.

Let me know if you know how to address this without JS, as it would also allow us to remove the JS from the block and size filters

app/views/shared/_size_and_zone_filters.html.erb
claudiovallejo: Great point! Setting `max-height` on desktop 👍

rsanheim (approved):

PR #1463, opened 2020-10-26
Automate weekly telemed report

vkrmis:
**Story card:** [ch1711](https://app.clubhouse.io/simpledotorg/story/1711/automate-weekly-report)

## Because

We currently run the weekly telemed report manually, this is time consuming and potentially errror-prone

## This addresses

Automate this report by fetching data from mixpanel via their [JQL](https://developer.mixpanel.com/reference/jql#jql-api-reference) API


config/schedule.rb
```diff
@@ -43,3 +43,9 @@ def local(time)
   rake "db:purge_users_data"
   rake "db:seed_users_data"
 end
+
+every :monday, at: local("11:00 am"), roles:[:cron] do
+  if FeatureToggle.enabled?("WEEKLY_TELEMED_REPORT")
+    rake "reports:telemedicine"
```
rsanheim, line 49: Can you put this feature flag into Flipper instead?

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -1,27 +1,56 @@
...
+    QUERY
+  end
+
+  def fetch_mixpanel_data
+    url = URI.parse("https://mixpanel.com/api/2.0/jql")
+
+    response = begin
+      HTTP.basic_auth(user: ENV["MIXPANEL_API_SECRET"], pass: nil)
+        .post(url, json: {format: "csv", script: @query})
+    rescue HTTP::Error => e
```
kitallis, line 38: Don't need to `rescue` and `raise` it back here, just let it bubble up.

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -1,27 +1,56 @@
...
-    mixpanel_csv = File.read(@mixpanel_csv_path)
+    mixpanel_csv = fetch_mixpanel_data
     mixpanel_data = CSV.parse(mixpanel_csv, headers: false)
 
     @mixpanel_data[:hydrated] = mixpanel_data.drop(1).map { |row|
-      facility = User.find(row[1]).registration_facility
-      {user_id: row[1],
-       date: Date.parse(row[2]),
-       clicks: row[3].to_i,
+      facility = User.find(row[0]).registration_facility
```
kitallis, line 50: I would suggest, make dumb functions out of this:

```ruby
def user_id(row)
  row[0]
end
```

or at least inline them as variables

```ruby
user_id = row[0]
date = row[1]
```

etc.

kitallis (commented):
Left some early comments inline (I realize it's in draft). Excited for this to be automated!

lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -3,32 +3,80 @@
...
+          mixpanel.reducer.count()
+        );
+      }
+    QUERY
+  end
+
+  def fetch_mixpanel_data
+    url = URI.parse("https://mixpanel.com/api/2.0/jql")
+
+    response = HTTP.basic_auth(user: ENV["MIXPANEL_API_SECRET"], pass: nil)
```
harimohanraj89, line 35: Shall we `ENV.fetch` here so it fails loud and clear if not configured? Eg. If we turn it on in BD accidentally.

harimohanraj89 (approved):
:shipit: Let's make our lives easier 👌 Also, let's check in manually the first few times to make sure it works 😅

vkrmis:
> :shipit: Let's make our lives easier 👌 Also, let's check in manually the first few times to make sure it works 😅

👍 This only emails the report to nilenso folks btw, there's still a manual step that we're NOT automating: copying this report to the google doc.

PR #1464, opened 2020-10-26
Remove zone validation

prabhanshuguptagit:
**Story card:** -

## Because

Facilities without a zone are being marked invalid since zone is now a required field. This caused the telemed MO link task to [fail](https://simpledotorg.slack.com/archives/CFHMC60P5/p1603699647096500).

## This addresses

This loosens the model validation until all facilities in prod have a block set. The input field in the form is still set to `required`.


harimohanraj89 (approved):
Looks fine. If possible, can we add the validation on `create` only, allowing facilities to be _updated_ seamlessly?

PR #1465, opened 2020-10-26
Datadog: Log request headers

prabhanshuguptagit:
**Story card:** -
## Because

We need to peek at the request headers in Datadog which aren't logged (tagged) anywhere by default.

## This addresses

This configures datadog's rack middleware to log the custom headers we use. [ref](https://docs.datadoghq.com/tracing/setup/ruby/#rack)


config/initializers/00_datadog.rb
```diff
@@ -10,6 +10,7 @@
   c.tracer.enabled = SEND_DATA_TO_DD_AGENT
   c.version = SimpleServer.git_ref(short: true)
   c.use :rails, analytics_enabled: true
+  c.use :rack, headers: {request: ["X-USER-ID", "X-FACILITY-ID", "X-SYNC-GROUP-ID"], response: ["Content-Type", "X-Request-ID"]}
```
harimohanraj89, line 13: Is the `X-Request-ID` set by DataDog?

PS does the capitalization have to be inconsistent between request and response headers?

harimohanraj89 (approved):
Couple non-blocking questions.

Let's test and :shipit:

config/initializers/00_datadog.rb
prabhanshuguptagit: The `Request ID` is set either by rails or nginx, not sure which one. This was already being logged in datadog as a [default](https://github.com/DataDog/dd-trace-rb/pull/389)

The headers are case insensitive, I just stuck to what they are conventionally being called around everywhere.

harimohanraj89: I guess that's fair. You could also camel(?)case em all if you want. Whichever you think is prettier.

PR #1467, opened 2020-10-27
Workaround for the Host-Header Injection vulnerability

kitallis:
**Story card:** [ch1730](https://app.clubhouse.io/simpledotorg/story/1730/host-header-injection-vulnerability)

## Because

From our security audit, this was one of the pending vulnerabilities that **wasn't** related to MITM attacks in particular. A [host-header](https://www.acunetix.com/blog/articles/automated-detection-of-host-header-attacks/) injection is an attack where the adversary can supply a malicious host header with an arbitrary domain or link, possibly to direct a user to that particular link.

This can be modified either by modifying the `Host` header or the `X-Forwarded-For` header. Since our app is behind SSL, and nginx manages the certficates, modifying `Host` should immediately be disallowed. However, `X-Forwarded-For` can still be messed with (we were also able to reproduce this issue).

The severity of this issue can vary from:

1. Simply a redirect to a malicious site,
2. Affecting other users if in case our pages are cached, effectively caching malicious pages (cache-poisoning)

pt.2 is unlikely because rails fragment / page caching does not cache by the `root_url` and we do not have a caching layer between nginx and rails, so we should be safe.

pt.1 is still possible.

Read more about the audit in this [doc](https://drive.google.com/drive/u/0/folders/1iTrURtw0QZhzZMGdS3w1xkUNOkIN27N4).

## This addresses

We have taken this approach documented in [secure_rails](https://github.com/ankane/secure_rails#host-header-injection), which will work for the most part.

This approach however is not fool-proof, as documented in this [thread](https://github.com/rails/rails/issues/29893#issuecomment-319768403). 

tl;dr: 

* if we directly use `request.host` in our application code (which we currently don't) – we'd potentially read the `X-Forwarded-For` header in there.
* we can't allow multiple domains as an allowlist (we currently don't need to) since this option can only allow setting one host.

For all other use-cases, this fix will work (made sure that the tests run by the audit also work 👍)

This issue is solved in a much better way in [Rails 6](https://github.com/rails/rails/issues/29893#issuecomment-467591014).

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -18,6 +18,7 @@ Rails.application.configure do
   # Full error reports are disabled and caching is turned on.
   config.consider_all_requests_local = false
   config.action_controller.perform_caching = true
+  config.action_controller.default_url_options = {host: ENV.fetch("SIMPLE_SERVER_HOST")}
 
   # Attempt to read encrypted secrets from `config/secrets.yml.enc`.
   # Requires an encryption key in `ENV["RAILS_MASTER_KEY"]` or


vkrmis:
> if we have multiple domains to allowlist, we can't do that (we currently don't) since this option can only allow setting one host.

Does this apply to the FQDN? If it does, would it disallow `dashboard.*` urls? Is there a way to do this in the nginx config instead of the Rails app?

kitallis:
> > if we have multiple domains to allowlist, we can't do that (we currently don't) since this option can only allow setting one host.
> 
> Does this apply to the FQDN? If it does, would it disallow `dashboard.*` urls? Is there a way to do this in the nginx config instead of the Rails app?

This is a good question. It won't disallow, but it will force URLs to be redirected to `SIMPLE_SERVER_HOST`. I think it can be done using `nginx` – but it seemed quite tricky to do so. Currently, nginx has no knowledge of this header, it just passes it through.

kitallis:
Closing this PR in favor of https://github.com/simpledotorg/deployment/pull/280 – since that's a better solution than hacking Rails' root URLs.

PR #1472, opened 2020-10-28
Fix Region Backfill

prabhanshuguptagit:
**Story card:** [ch1455](https://app.clubhouse.io/simpledotorg/epic/1339/block-level-syncing)

## Because

The Region backfill is spinning up a separate district and zone for each facility.

## This addresses

Fixes the backfill script to check if the region exists before creating a new one. This also makes the script idempotent so we can run this script more than once to backfill as and when the CVHOs send us newer block data.

### DIFF
--- a/app/services/region_backfill.rb
+++ b/app/services/region_backfill.rb
@@ -50,20 +50,20 @@ class RegionBackfill
     facility_type = find_region_type("Facility")
 
     root_parent = NullRegion.new(name: "__root__")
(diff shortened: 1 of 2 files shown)


rsanheim (approved):

rsanheim:
Pushed up an additional commit to keep slugs consistent with the source objects, otherwise Facility Regions were getting created with different slugs for some reason. Discovered that while working on https://github.com/simpledotorg/simple-server/pull/1477.

PR #1473, opened 2020-10-28
Make strong password a bit more lax

rsanheim:
Bump down min entropy to 17 - the default is 18.

**Story card:** [chXXXX](URL)

## Because

## This addresses


harimohanraj89 (approved):
Is 17 low enough? Do we know if some attempted passwords would have worked? I know it goes against our instincts 😅  but if we're gonna lower the requirement, may as well ensure it makes things tangible easier for our users.

PR #1475, opened 2020-10-28
Overdue: Filter copy fix

claudiovallejo:
**Story card:** [ch1749](https://app.clubhouse.io/simpledotorg/story/1749/overview-copy-fix)

## Because

There should be no space between `<` and `365` in the filter label

## This addresses

**No space between symbol and number**

image


claudiovallejo:
Closing because this change should be made in #1478

PR #1477, opened 2020-10-28
Block level reports via Regions

rsanheim:
**Story card:** [ch132](https://app.clubhouse.io/simpledotorg/story/132/block-level-reports)

## Because

We want to start moving towards reporting based on Regions instead of Facility or FacilityGroups.

## This addresses

Adding block level reports behind a feature flag


app/controllers/reports/regions_controller.rb
```diff
@@ -141,14 +155,25 @@ def find_region
     end
   end
 
+  def find_region_v2
+    region_type = report_params[:report_scope]
+    @region = if region_type == "facility_district"
+      FacilityDistrict.new(name: report_params[:id], scope: current_admin.accessible_facilities(:view_reports))
```
rsanheim, line 161: @harimohanraj89 can you remind me what the long term plan for "FacilityDistrict" is? Can this go away once we have regions fully in placed and production data cleaned up?

harimohanraj89: Totes. Our goal should still be to axe this thing asap.

I chatted with Vinay. Since the app can intelligently purge data outside a user's sync group regularly now, we should be able to rearrange facilities and groups more freely now. The rough order of operations should be:
* Launch block-level sync
* Test in SBX what happens if we split facilities in one facility group into two groups
* Test in SBX what happens if we combine facilities from multiply facility groups into one group
* If it looks good, make the data changes in prod
  * Split Bathinda and Mansa into two FGs
  * Combine Chennai into a single FG
* Axe the `FacilityDistrict`

app/controllers/reports/regions_controller.rb
```diff
@@ -141,14 +155,25 @@ def find_region
...
+    else
+      Region.where(region_type: region_type).find_by!(slug: report_params[:id])
```
harimohanraj89, line 163: Consistently fueled by Regions. Love it. It's happening!

harimohanraj89 (changes requested):

app/models/region.rb
```diff
@@ -22,6 +22,24 @@ class Region < ApplicationRecord
...
+    end
+    # Don't leave around the old, auto generated methods to avoid confusion
+    undef_method "#{type}_regions?"
+  end
+
+  def assigned_patients
+    if source
+      source.assigned_patients
+    else
+      Block.new(name: name, facilities: facilities).assigned_patients
```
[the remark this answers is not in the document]
rsanheim: yeah I think that makes sense. I'll give it a shot. 👍

app/models/block.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  def initialize(name:, facilities:)
+    @name = name
+    @facilities = facilities
+  end
+
+  def assigned_patients
+    Patient.where(assigned_facility: facilities)
+  end
+end
```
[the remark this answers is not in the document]
rsanheim: Yup, removed this model. 👍

app/models/region.rb
```diff
@@ -38,6 +56,15 @@ def short_uuid
     SecureRandom.uuid[0..7]
   end
 
+  def facilities
+    if facility_region?
+      [source]
```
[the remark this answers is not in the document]
rsanheim: yup this is cleaner now

spec/controllers/reports/regions_controller_spec.rb
```diff
@@ -175,6 +175,123 @@ def refresh_views
     end
   end
 
+  context "show v2" do
+    render_views_on_ci
```
rsanheim, line 179: rendering views in a local TDD loop was too painful, and I'm not really testing anything here deep in the views anyways other than the fact that they render and don't blow up

spec/support/logger_helper.rb
```diff
@@ -0,0 +1,9 @@
+module LoggerHelper
+  def logger
```
rsanheim, line 2: make the rails logger easily available in all specs

app/views/reports/regions/_block_tables.html.erb
```diff
@@ -0,0 +1,277 @@
+<h3 class="mt-48px mb-8px fs-28px c-black">
```
rsanheim, line 1: This whole view could probably be consolidated w/ the facility_tables view, but not yet...

rsanheim (commented):
Some comments for reviewers.

app/models/facility.rb
```diff
@@ -200,6 +200,24 @@ def discardable?
...
+  def facility_region?
+    true
+  end
+
+  # For regions compatibility
+  def district_region?
+    false
+  end
+
+  def self.state_region_name(facility_attributes)
```
harimohanraj89, line 213: What's this for? Is it used?

app/models/region.rb
```diff
@@ -38,6 +48,19 @@ def short_uuid
...
 
+  def assigned_patients
+    Patient.where(assigned_facility: facilities)
+  end
+
+  def facilities
+    if facility_region?
+      Facility.where(id: source_id)
+    else
+      ids = facility_regions.pluck(:source_id)
```
harimohanraj89, line 59: nitpick: Maybe call this `source_ids` for consistency and no confusion?

app/controllers/reports/regions_controller.rb
```diff
@@ -26,12 +26,18 @@ def show
     @new_registrations = @last_registration_value - (@data[:cumulative_registrations].values[-2] || 0)
     @adjusted_registration_date = @data[:adjusted_registrations].keys[-4]
 
-    if @region.is_a?(FacilityGroup) || @region.is_a?(FacilityDistrict)
+    if @region.district_region?
```
harimohanraj89, line 29: There's [one more touchpoint](https://github.com/simpledotorg/simple-server/pull/1477/files#diff-15a670f34e2a19569d76af44251cfd75268f11cfd403b931241305f2c5becfd9R75) now that needs this `district_region?` applied.

rsanheim: Hm, this link isn't working - can you point me to the file / line number?

app/controllers/reports/regions_controller.rb
```diff
@@ -132,6 +138,14 @@ def set_force_cache
   end
 
   def find_region
+    if current_admin.feature_enabled?("region_reports")
+      find_region_v2
```
harimohanraj89, line 142: When enabled, _all_ reports will be fueled by regions instead of "sources" (like facilities or facility groups), right? That's a pretty significant change, I think. Any extra testing we need to do? Verify that the caching doesn't get in the way, etc?

app/models/facility.rb
rsanheim: nope, dead code - removed.

app/models/region.rb
rsanheim: 👍

app/controllers/reports/regions_controller.rb
rsanheim: Yup, once the flag is on that is true.  The plan is to ship this soon so I can verify on sandbox and also get some help from others on QA'ing it.  I'll definitely be doing some updates for caching and testing in follow up PRs before we turn the flag on for everyone.

rsanheim (commented):
@harimohanraj89 responding to some feedback.

app/controllers/reports/regions_controller.rb
harimohanraj89: Funtimes 🤦  Here ya go

https://github.com/simpledotorg/simple-server/blob/master/app/controllers/reports/regions_controller.rb#L69

It's the `download` action in the same controller.

app/controllers/reports/regions_controller.rb
harimohanraj89: Neatneatneat. Sounds like a good plan.

app/models/facility.rb
harimohanraj89: Phew, awesome. It was just gnarly enough for me to be thankful that it's dead code 😂

spec/support/logger_helper.rb
harimohanraj89: Is this worth the extra effort/indirection? 😅  I don't see the logger helper used anywhere yet, I think? 🤔

harimohanraj89 (approved):
A bunch of final miscellaneous thoughts, but an unblocking approval to you, good sir.

It's feature-flagged, so we can get this merged ASAP and continue our QA process as you mentioned, to test feature flag toggling, caching behavior, weird districts like Bathinda and Mansa, etc.

PR #1481, opened 2020-10-29
Reports number formatting with numeral.js

claudiovallejo:
**Story card:** [ch1576](https://app.clubhouse.io/simpledotorg/story/1576/localized-number-format-with-numeral-js)

## Because

It's hard to read regex

## This addresses

Uses `numeral.js` to format numbers with commas. Sets us up to render localized number formats in the future.


claudiovallejo:
@rsanheim @kitallis I've setup numeral.js to replace regex to format numbers in Reports. Is locale-specific number formatting necessary? If so, would we format numbers for India, Bangladesh, and Ethiopia locales?

app/views/reports/regions/show.html.erb
```diff
@@ -300,4 +300,5 @@
 <div id="data-json" style="display: none;">
   <%= raw @data.to_json %>
 </div>
+<script src="//cdnjs.cloudflare.com/ajax/libs/numeral.js/2.0.6/numeral.min.js"></script>
```
harimohanraj89, line 303: What happens if this asset fails to load, eg if the CDN is down/slow? Does the page no longer display numbers? Perhaps we could define a placeholder `numeral` that returns the plain number?

```js
var numeral = function(value) {
  return {
    format: function() { return value }
  }
}
```

that can get overridden once the library loads.

harimohanraj89 (changes requested):
I'm starting to feel that these JS assets that are critical to our reports working should be included in our asset pipeline, rather than loading from CDNs. I think we should add them to our `package.json` as dependencies, and load them with some `require`s in `application.js`. What do you think?

app/views/reports/regions/show.html.erb
claudiovallejo: @harimohanraj89 thanks, that's a great point I didn't think of. Do I just do a `yarn install numeral`?

harimohanraj89: You can run `bin/yarn add numeral` (and also `bin/yarn add chart.js` if you'd like), and then "require" them in `application.js` just like we [require tablesort](https://github.com/simpledotorg/simple-server/blob/master/app/assets/javascripts/application.js#L19).

kitallis:
@claudiovallejo 

> Is locale-specific number formatting necessary?

I don't think so, since `numeral.js` supports localization, we're a step ahead already. I think this is good enough for now.

app/views/reports/regions/show.html.erb
harimohanraj89: PS hit me up if you wanna pair on this later today.

claudiovallejo: > You can run `bin/yarn add numeral` (and also `bin/yarn add chart.js` if you'd like), and then "require" them in `application.js` just like we [require tablesort](https://github.com/simpledotorg/simple-server/blob/master/app/assets/javascripts/application.js#L19).

@harimohanraj89 I tried requiring the `chart.js` module as well but Rails expects a file named `chart.js` to exist in our js directory when I add `//= require chart.js` (that's the name of the module...). I did a quick search online, but couldn't find a fix. Do you know of an easy fix for this? Or should we punt and change chart.js in the future?

rsanheim: I think you can skip adding chart.js to the asset pipeline for now, that can be a separate PR.

rsanheim (approved):

PR #1482, opened 2020-10-29
Reports JS refactor

claudiovallejo:
**Story card:** [ch1449](https://app.clubhouse.io/simpledotorg/story/1449/reports-js-refactor)

## Because

`reports.js` is a mess. It's hard to read and hard to edit.

## This addresses

* Listing function declarations sequentially
* Removing as many abstractions as possible
* More verbose, but easier to read graph configuration objects


rsanheim (approved):
I think this is definitely going in the right direction here. Also interesting that its only some ~40 more lines of code when everything is made explicit & inline like this! 👍

PR #1485, opened 2020-10-29
Persist params for missed visits section

claudiovallejo:
**Story card:** [ch1779](https://app.clubhouse.io/simpledotorg/story/1779/persist-params-for-missed-visits-section)

## Because

The "period" param should persist when user navigates between "Missed visits" and "Registrations" sections in "Home".

## This addresses

Adds "period" as a param to persist for "Missed visits"


harimohanraj89 (approved):
The "day" period doesn't apply to missed visits, but I think that's ok.

One thing to consider: Perhaps we can pass `period` to the other pages as well (ranked facilities, overview, bp control) so that you can click to one of those pages and back, and your period selection will _still_ remain.

kitallis (approved):

PR #1486, opened 2020-10-30
Upsert Regions through source models – I

kitallis:
**Story card:** [ch1783](https://app.clubhouse.io/simpledotorg/story/1783/triggers-to-automatically-create-regions-in-the-background-when-new-source-models-are-created-or-updated)

## Because

Keep the `Region` model in sync as `RegionSource` models are created or updated.

## This addresses

Add model-level callbacks.

**Note:** This PR only adds the callbacks to the `Organization` model. For `FacilityGroup` and `Facility`, separate PRs will be raised.


prabhanshuguptagit (commented):
lgtm

app/models/region.rb
```diff
@@ -1,20 +1,25 @@
...
   validates :region_type, presence: true
 
   belongs_to :source, polymorphic: true, optional: true
 
-  before_discard do
-    self.path = nil
-  end
-
-  MAX_LABEL_LENGTH = 255
+  before_validation :set_path
```
rsanheim, line 21: I don't think we want to set the path every time a region is validated -- the path should be stable and unchanging most of the time, since the slug itself and the hiearchy of regions is unchanging most of the time.

rsanheim (changes requested):

app/services/region_backfill.rb
```diff
@@ -90,12 +90,8 @@ def find_or_create_region_from(parent:, region_type:, name: nil, source: nil)
     return existing_region if existing_region
 
     region = DryRunRegion.new(Region.new(name: region_name, region_type: region_type), dry_run: dry_run?, logger: logger)
-    if source
-      region.slug = source.slug
```
rsanheim, line 94: Why wouldn't we want to copy over the source slug? This will keep some things simpler from a dashboard reports perspective.

app/models/region.rb
rsanheim: Pushing up some changes to handle this concern and limit the scope of when we are resetting a path, I'll comment on the code once I'm sure all tests are passing.

app/models/organization.rb
```diff
@@ -14,6 +14,27 @@ class Organization < ApplicationRecord
 
   auto_strip_attributes :name, squish: true, upcase_first: true
 
+  # ----------------
+  # Region callbacks
+  after_create :create_region
+  before_update :update_region
+
+  def create_region
+    return if region&.persisted?
```
rsanheim, line 23: This will handle the case where Regions are driving things in the future, so we don't accidentally create unnecessary regions where one is already created.

app/models/organization.rb
```diff
@@ -14,6 +14,27 @@ class Organization < ApplicationRecord
...
+    parent = Region.find_by!(region_type: Region.region_types[:root])
+    region = build_region(name: name, description: description, reparent_to: parent)
+    region.region_type = Region.region_types[:organization]
+    region.save!
+  end
+
+  def update_region
+    return unless name_changed? || description_changed?
```
rsanheim, line 31: Avoid unnecessary updates here if the name or description has not changed.

app/services/region_backfill.rb
```diff
@@ -89,13 +89,8 @@ def find_or_create_region_from(parent:, region_type:, name: nil, source: nil)
...
     return existing_region if existing_region
 
-    region = DryRunRegion.new(Region.new(name: region_name, region_type: region_type), dry_run: dry_run?, logger: logger)
-    if source
-      region.slug = source.slug
-      region.source = source
-    end
-    region.path = [parent.path, region.path_label].compact.join(".")
-
+    region = DryRunRegion.new(Region.new(name: region_name, region_type: region_type, reparent_to: parent), dry_run: dry_run?, logger: logger)
```
rsanheim, line 92: `reparent_to` now handles setting the path for us

rsanheim (approved):
some notes for rationale behind these changes.

rsanheim:
Going to merge this to unblock and keep progress moving - we can chat or open new PRs to handle any follow up 😁 .

app/models/organization.rb
kitallis: Make sense!

app/models/organization.rb
kitallis: I had this, but removed it because of paranoia. But I think it should be ok to do this...

app/services/region_backfill.rb
kitallis: We also had something similar here, but removed it again, because it seemed like introducing a new API for re-parenting was confusing and not obvious. I'm open to this, but reluctantly so.

app/services/region_backfill.rb
prabhanshuguptagit: This is causing problems due to bad data in sandbox/staging. Some FGs have facilities in multiple states which breaks the slug uniqueness and those regions don't get created. We already have the `source` to map, is it necessary to match slugs? The other way is to make sure the data is extremely clean (hierarchically speaking) everywhere.

app/models/region.rb
```diff
@@ -10,11 +12,11 @@ class Region < ApplicationRecord
...
   belongs_to :source, polymorphic: true, optional: true
 
-  before_discard do
-    self.path = nil
-  end
-
-  MAX_LABEL_LENGTH = 255
+  # To set a new path for a Region, assign the parent region via `reparent_to`, and the before_validation
+  # callback will assign the new path.
+  attr_accessor :reparent_to
```
prabhanshuguptagit, line 17: Can we continue using the `@parent` writer method instead of adding a `reparent_to`? imo `parent:` makes for a sweet API, given that we're adding this method just to make setting paths ergonomic.

app/models/organization.rb
prabhanshuguptagit: I feel like we can get away with not having to check this, it's one more conditional to keep in mind (there's already too many) while writing all the callbacks.

app/models/region.rb
```diff
@@ -60,6 +61,21 @@ def log_payload
 
   private
 
+  def initialize_path?
+    @initialize_path
+  end
+
+  def initialize_path
+    errors.add(:reparent_to, "must assign new parent to initialize path") unless reparent_to
```
kitallis, line 69: This is unreachable code right? It will never be called if we have the `if: :reparent_to` upstairs.

app/models/region.rb
```diff
@@ -60,6 +61,21 @@ def log_payload
...
+    logger.info(class: self.class, msg: "got reparent_to: #{reparent_to}, going to initialize new path")
+    self.path = "#{reparent_to.path}.#{path_label}"
+    self.reparent_to = nil
```
prabhanshuguptagit, line 72: Why do we need to set this back to `nil`?

app/models/region.rb
rsanheim: We discussed in a zoom, but just for posterity - the goal here was to keep the destructive api of 'reparent' - i.e. setting a parent and changing the path as very separate from the read only API of retrieving the parent.

app/models/region.rb
rsanheim: Yup, barring any sort of very very unlikely race condition. 👍

app/models/region.rb
rsanheim: Being defensive and making sure we mark the record as "done reparenting".  There are cases where validation callbacks on a model could be triggered multiple times.

PR #1489, opened 2020-10-30
Use the slug for the Region path instead of the name

rsanheim:
The slug will not change over time, whereas the name can

**Story card:** [ch1786](https://app.clubhouse.io/simpledotorg/story/1786/use-the-region-slug-for-its-ltree-path)

## Because

## This addresses


app/models/region.rb
```diff
@@ -19,8 +19,8 @@ class Region < ApplicationRecord
   # A label is a sequence of alphanumeric characters and underscores.
   # (In C locale the characters A-Za-z0-9_ are allowed).
   # Labels must be less than 256 bytes long.
-  def name_to_path_label
-    name.gsub(/\W/, "_").slice(0, MAX_LABEL_LENGTH)
+  def path_label
+    slug.gsub(/\W/, "_").slice(0, MAX_LABEL_LENGTH)
```
kitallis, line 23: I suppose we still need to do this after Friendly ID, even though it prefers underscores because of [uniqueness](https://github.com/norman/friendly_id/blob/master/lib/friendly_id/slugged.rb#L100) and probably just general care-taking of the health of this path.

lib/tasks/scripts/import_blocks_to_regions.rb
```diff
@@ -56,7 +56,8 @@ def find_or_create_region(name:, type:, parent:)
 
   def create_region_from(name:, type:, parent:)
     region = Region.new(name: name, type: type)
-    region.path = [parent.path, region.name_to_path_label].compact.join(".")
+    region.send(:set_slug) unless region.slug
```
kitallis, line 59: Don't think I understood why this is necessary. If FID [runs](https://github.com/norman/friendly_id/blob/master/lib/friendly_id/slugged.rb#L250) `set_slug` already as a `before_validation` why does it matter if it's set previously or not?

app/models/region.rb
kitallis, line 23: As we discussed over our catchup, I think this approach is pretty acceptable, and at least a lot better than the alternative of using `name` and then recursively having to manage paths or whatever to avoid name-clashing.

Since Friendly ID maintains uniqueness and hence provides [this backoff](https://github.com/norman/friendly_id/blob/master/lib/friendly_id/slugged.rb#L226) behavior in race-conditions, we should add a uniqueness clause to the path label field as well; because no two paths should ever clash.

kitallis (approved):
Just left a few comments, otherwise lgtm 👾

lib/tasks/scripts/import_blocks_to_regions.rb
kitallis: Would we not have to do this in `path_label` as well?

It seems like you can't have a `path` as blank, but `path` depends on `slug` (which also can't be blank), so you'd have to set slug before you set the path, since the `set_path` method will blow during `before_validation` otherwise.

I don't know how to correctly solve this, but I find this dependency chain of path versus slug quite confusing.

PR #1490, opened 2020-11-02
Remove RegionType

prabhanshuguptagit:
**Story card:** [ch1455](https://app.clubhouse.io/simpledotorg/story/1455/create-region-schema-run-dry-run-of-region-backfill-in-prod)

## Because

Following the conversation from last week, this removes the `RegionType` class. 

### Rationale:

`Region types` are mostly static information given an environment/country. We don't expect it to change at runtime once they've been set. Since the code also depends on the region types being present at boot, it's a bit inconvenient to load `RegionTypes` in the app and stub them in tests. This information sits better inside code (or config) given the amount of flexibility we desire in the near-long term.

## This addresses

Removes `RegionType`, the related associations and adds an enum `region_type` on `Region`. The hierarchy is intentionally the same across all envs and conforms to `IHCI`. When we decide to make this different per country, this can be moved to the `CountryConfig`.


app/models/region.rb
```diff
@@ -26,33 +29,49 @@ def name_to_path_label
...
+      if ancestor_types(region_type).include?(self.region_type)
+        descendants.where(region_type: region_type)
       else
-        raise NoMethodError, "undefined method #{region_type.name.pluralize.underscore} for #{self} of type #{type.name}"
+        raise NoMethodError, "undefined method #{region_type.pluralize.underscore} for #{self} of type #{self.region_type}"
       end
     end
   end
+
+  private
```
kitallis, line 60: These methods are ❇️

spec/models/region_spec.rb
```diff
@@ -50,20 +50,17 @@
   end
 
   context "association helper methods" do
-    it "generates the appropriate has_one or has_many type methods based on the available RegionTypes" do
+    it "generates the appropriate has_one or has_many type methods based on the available region types" do
       facility_group_1 = create(:facility_group, organization: create(:organization))
       create(:facility, facility_group: facility_group_1)
 
       RegionBackfill.call(dry_run: false)
```
kitallis, line 57: Can we leave a comment here; that reminds us to move away from `RegionBackfill` in this test to actually creating factory data?

app/models/region.rb
kitallis, line 60: These methods are ❇️

kitallis (approved):
this looks good to me; should help simplify migration code a fair bit.

PR #1491, opened 2020-11-02
Progress tab "Updated date" fix

claudiovallejo:
**Story card:** [ch1798](https://app.clubhouse.io/simpledotorg/story/1798/progress-tab-visual-fixes)

## Because

* "Updated date" is not showing
* No space between "Today" card and top header

Screenshot_20201102-114627

## This addresses

* Displays "Updated date" text
* Updates "Updated date" element spacing

image


harimohanraj89:
Do we know why this broke in the first place?

claudiovallejo:
> Do we know why this broke in the first place?

I don't @harimohanraj89. Replacing `float: right;` with `text-align: right;` fixed it in my local, but don't know if the "Updated at" text hasn't been showing since the last update to this file in Aug 2020.

harimohanraj89:
Hmm, it may be [this change here](https://github.com/simpledotorg/simple-server/pull/1458/files#diff-66384dd8f7be999bdaabb1ac3ff9bcf7974b1717b0ba613acc7f09dc1b915443L298-L302). 

You could confirm by checking out the previous commit locally and seeing if the progress tab is fine there - `git checkout fb5c39381ea7259ffa15abd5f6906f9ec6a7def2`. 

Neato, if that's the change that caused it, then I'm fairly confident that we're good now :smile: I was worried it might be some android web view update thingy or something.

claudiovallejo:
> Hmm, it may be [this change here](https://github.com/simpledotorg/simple-server/pull/1458/files#diff-66384dd8f7be999bdaabb1ac3ff9bcf7974b1717b0ba613acc7f09dc1b915443L298-L302).
> 
> You could confirm by checking out the previous commit locally and seeing if the progress tab is fine there - `git checkout fb5c39381ea7259ffa15abd5f6906f9ec6a7def2`.
> 
> Neato, if that's the change that caused it, then I'm fairly confident that we're good now 😄 I was worried it might be some android web view update thingy or something.

I checked out the previous commit and the problem is still there, so it's likely a CSS bug and not an Android web view.

harimohanraj89 (approved):
Gotcha, godspeed!

PR #1492, opened 2020-11-03
Remove teleconsult fields

harimohanraj89:
**Story card:** [ch1251](https://app.clubhouse.io/simpledotorg/story/1251/remove-old-teleconsultation-fields)

## Because

The old teleconsultation code paths are now obsolete.

## This addresses

Removal of the following items:
* Old teleconsultation fields on the facility model
* Feature flag
* Backfill data scripts (this assumes that all MOs are accurately captured in the new data model)


app/models/user.rb
```diff
@@ -134,9 +134,9 @@ def registration_facility_id
   alias_method :facility, :registration_facility
 
   def full_teleconsultation_phone_number
-    defaulted_teleconsult_number = teleconsultation_phone_number.presence || phone_number
-    teleconsultation_isd_code ||= Rails.application.config.country["sms_country_code"]
```
harimohanraj89, line 138: We had a bug here. Existing ISD codes were not respected because this statement initialized a `teleconsultation_isd_code` local variable, which would always be `nil` on the right side.

app/models/facility.rb
```diff
@@ -243,15 +218,12 @@ def teleconsultation_enabled?
   end
 
   def teleconsultation_phone_number_with_isd
-    teleconsultation_phone_number = teleconsultation_phone_numbers.first
-    return if teleconsultation_phone_number.blank?
-
-    Phonelib.parse(teleconsultation_phone_number.isd_code + teleconsultation_phone_number.phone_number).full_e164
+    teleconsultation_phone_numbers_with_isd.first&.dig(:phone_number)
```
harimohanraj89, line 221: Grab the first MO's number instead of the facility attribute. This method powers the `api/v4/facility_teleconsultations` endpoint that may still be in use.

app/models/facility.rb
```diff
@@ -243,15 +218,12 @@ def teleconsultation_enabled?
...
   end
 
   def teleconsultation_phone_numbers_with_isd
-    teleconsultation_phone_numbers.map do |phone_number|
-      {phone_number: Phonelib.parse(phone_number.isd_code + phone_number.phone_number).full_e164}
+    teleconsultation_medical_officers.map do |medical_officer|
+      {phone_number: medical_officer.full_teleconsultation_phone_number}
```
harimohanraj89, line 226: Grab the MO's numbers instead of the facility attribute. This method powers the `api/v4/facility_teleconsultations` endpoint that may still be in use.

kitallis: Hrm, didn't realize this method was sticking this data in a hash. Can't this just return the list and the hash stuff can happen at the call site? This hash seems a little arbitrary in this model.

Ignore if too much change.

app/models/user.rb
kitallis: What would have been the consequence of this? Numbers without ISD codes would've broken?

kitallis (approved):
looks pretty solid to me 👍 – left a few minor comments

app/models/user.rb
harimohanraj89: Numbers _have_ to have ISD code. If all MOs added to IHCI had India phone numbers, then there would be no impact. If an MO had a non-india phone number, we'd clobber their ISD code with +91, and their number would be borked.

I think all MOs have India numbers, we should be good.

spec/models/facility_spec.rb
```diff
@@ -266,38 +266,6 @@
                                            import: true)
     end
 
-    context "with old teleconsultation fields" do
-      let(:upload_file) { fixture_file_upload("files/upload_facilities_test_old.csv", "text/csv") }
```
prabhanshuguptagit, line 270: We can delete this fixture file as well.

app/models/facility.rb
prabhanshuguptagit: Should we nudge praveen once again to send out a notification to update apps? This change might surprise users on the oldest teleconsult version.

prabhanshuguptagit (changes requested):
Some more things that can be dropped ✨

app/models/facility.rb
harimohanraj89: As discussed, I'll clarify the impacts with Praveen before shipping this cleanup.

app/models/facility.rb
harimohanraj89: If the touchpoints are few, I'll do it right away. If they are numerous, I'll leave it alone.

app/models/facility.rb
harimohanraj89: Clarified. No comms necessary, we can ship this.

prabhanshuguptagit (approved):
Like how much cleaner `Facility` is now 🧼✨

PR #1494, opened 2020-11-03
Reports: Comma-formatted numbers

claudiovallejo:
**Story card:** [ch1824](https://app.clubhouse.io/simpledotorg/story/1824/comma-separated-numbers)

## Because

We should display numbers in a consistent format across the Dashboard. This PR ensures numbers are:
1. Displayed as `-` if `0`
2. Comma-formats numbers (i.e. `1234` → `1,234`)

## This addresses

Displays comma-formatted numbers in the following Dashboard sections

1. Reports "Details"
2. Reports "Cohort reports"


app/views/reports/regions/details.html.erb
```diff
@@ -59,16 +59,16 @@
                 Total
               </td>
               <td class="row-total ta-center">
-                <%= dash_if_zero(@dashboard_analytics.sum { |_, row| row[:total_registered_patients] || 0 }) %>
```
harimohanraj89, line 62: We're no longer using dashes if the number is zero? Is that intentional?

harimohanraj89 (changes requested):
One question from me.

app/views/reports/regions/details.html.erb
claudiovallejo: I did this intentionally, as it's the only table where we do this. I will get Daniel's approval first though.

harimohanraj89: Looks like we want to stick with dashes according to Slack.

I did a quick check and it's safe to pass dashes to the `number_with_delimiter` methods, so we can update these lines to:
```erb
<%= number_with_delimiter(dash_if_zero(@dashboard_analytics.blah_blah)) %>
```

claudiovallejo: Great, thank you @harimohanraj89 !

harimohanraj89 (approved):

PR #1497, opened 2020-11-04
Populate states through dropdown

prabhanshuguptagit:
**Story card:** [1343](https://app.clubhouse.io/simpledotorg/story/1343/clean-up-existing-facility-block-zone-data)

## Because
We want to set states at the FG level instead of the facility level to avoid clashes with the Region hierarchy (all facilities should belong to the same state).

## This addresses

This adds an exhaustive list of states for each country which populates a dropdown in the FG form. The state regions are created on the fly if they don't exist already. This is a quicker solution than creating a state management UI in the org page, which can come in _after_ block level sync is wrapped up.
image

This also adds Region upsert callbacks for Facility groups a la https://github.com/simpledotorg/simple-server/pull/1486.


app/models/facility_group.rb
```diff
@@ -39,6 +48,8 @@ def toggle_diabetes_management
       set_diabetes_management(true)
     elsif diabetes_enabled?
       set_diabetes_management(false)
+    else
+      true
```
prabhanshuguptagit, line 52: This is an unrelated fix to `toggle_diabetes_management`. It returns a `nil` if there is no change, which is broken behaviour.

app/models/facility_group.rb
```diff
@@ -69,4 +80,25 @@ def cohort_analytics(period:, prev_periods:)
...
+      source: self,
+      reparent_to: state_region,
+      region_type: Region.region_types[:district]
+    )
+  end
+
+  def update_region
+    region.reparent_to = state_region
+    region.name = name
+    region.save!
```
rsanheim, line 96: Can you check to see if any of these attributes changed to decide if we need to do the update at all here?

prabhanshuguptagit: I think we can avoid that check, just for the sake of reducing complexity. There's already quite a few conditionals to keep in mind. wdyt? https://github.com/simpledotorg/simple-server/pull/1486#discussion_r517017412

config/data/canonical_states.yml
```diff
@@ -0,0 +1,43 @@
...
+  - Comilla
+  - Dhaka North
+  - Dhaka South
+  - Gazipur
+  - Khulna
+  - Mymensingh
+  - Narayanganj
+  - Rajshahi
+  - Rangpur
+  - Sylhet
```
prabhanshuguptagit, line 43: This does not have Ethiopia states currently. We can run Region Backfill + enable flipper on ET once we have that information.

rsanheim: These are the Ethiopia "states" - there are 12 of them.  In Ethiopia they are known as "regions": ..."There are currently ten regional states and two chartered cities".

https://en.wikipedia.org/wiki/Regions_of_Ethiopia#List_of_regions_and_city_administrations

I don't think there is a reason to not add them right away?

prabhanshuguptagit: The CSV that we received has names in the amharic script, not sure if we're planning on using english or amharic. I'll ask for this on slack. Meanwhile the state names are not a blocker, I meant.

kitallis:
@prabhanshuguptagit I think we should disable **editing** the `State` from the dropdown? Only select whilst creating it?

config/data/canonical_states.yml
```diff
@@ -0,0 +1,47 @@
...
+Ethiopia:
+  - አዲስ አበባ
+  - ትግራይ
+  - Oromiyaa
```
prabhanshuguptagit, line 47: This is not the complete list of Ethiopian states, we can update this in a later PR.

app/controllers/admin/facility_groups_controller.rb
```diff
@@ -1,7 +1,11 @@
 class Admin::FacilityGroupsController < AdminController
+  COUNTRYWISE_STATES = YAML.load_file("config/data/canonical_states.yml")
```
harimohanraj89, line 2: Could we load and store this in the `CountryConfig`?

harimohanraj89 (commented):
Looks cool, minor suggestions from me. Are we gonna try to get away from the callback pattern in this PR?

app/controllers/admin/facility_groups_controller.rb
prabhanshuguptagit: Addressed in 9eb755f

prabhanshuguptagit:
> Are we gonna try to get away from the callback pattern in this PR?

https://github.com/simpledotorg/simple-server/pull/1500 is the PR in question, where the callback pattern was being overused a bit. That has been fixed now.

harimohanraj89 (approved):
My feedback's addressed 🚀

PR #1499, opened 2020-11-05
Add Teleconsultation to purge controller

prabhanshuguptagit:
**Story card:** -

## Because

We are not purging teleconsultations via the purge controller for QA

## This addresses

Adds `Teleconsultation` to `PurgeUsersData`.


kitallis (approved):
look gud

PR #1501, opened 2020-11-05
Dont override slug generation for factory created facilities

rsanheim:
This overrides what FriendlyId would do by default, which is not
something we want to mess with in our test data. Doing the slug generation via the factory also obscures the
default behavior of FriendlyId where it appends a UUID for unique slugs.

### DIFF
--- a/spec/factories/facilities.rb
+++ b/spec/factories/facilities.rb
@@ -16,10 +16,6 @@ FactoryBot.define do
     enable_teleconsultation { true }
     monthly_estimated_opd_load { 300 }
 
-    sequence :slug do |n|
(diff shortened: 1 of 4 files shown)


timcheadle (commented):
Where does this PR obscure the UUID suffix?

rsanheim:
> Where does this PR obscure the UUID suffix?

I clarified the description here -- the previous code was overriding the FriendlyId slug uniqueness logic, which will use the UUID.  By removing the slug generation from our factories, our tests will operate more like the slug creation in production.

dburka (approved):
Nice slugs.

harimohanraj89:
Agree.

kitallis:
Looks good to me.

PR #1502, opened 2020-11-05
Region backfill - better handle slug dupes

rsanheim:

We want to make sure we handle duplicate slugs when backfilling regions. This adds a test for that case and ensure we smartly copy over the slug from the source, or fallback to a uniq suffix when needed in the case of duplicate Region slugs.

### DIFF
--- a/app/models/region.rb
+++ b/app/models/region.rb
@@ -21,6 +21,11 @@ class Region < ApplicationRecord
   REGION_TYPES = %w[root organization state district block facility].freeze
   enum region_type: REGION_TYPES.zip(REGION_TYPES).to_h
 
+  # Override the auto-generated root method (via enum) to return the one, single root Region
+  def self.root
+    Region.find_by(region_type: :root)
+  end
+
   # A label is a sequence of alphanumeric characters and underscores.
   # (In C locale the characters A-Za-z0-9_ are allowed).
   # Labels must be less than 256 bytes long.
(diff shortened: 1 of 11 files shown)


dburka (approved):
Slugs

app/services/region_backfill.rb
```diff
@@ -91,7 +91,9 @@ def find_or_create_region_from(parent:, region_type:, name: nil, source: nil)
     region = DryRunRegion.new(Region.new(name: region_name, region_type: region_type, reparent_to: parent), dry_run: dry_run?, logger: logger)
     if source
       region.source = source
-      region.slug = source.slug
+      # Attempt to set the slug on the region the same as the source -- we are using FriendlyId's set_slug
+      # so that the built-in uniqueness logic will work for duplicates.
+      region.set_slug(source.slug)
```
rsanheim, line 96: This will correctly handle dupe slugs when necessary.

app/models/region.rb
```diff
@@ -21,6 +21,11 @@ class Region < ApplicationRecord
   REGION_TYPES = %w[root organization state district block facility].freeze
   enum region_type: REGION_TYPES.zip(REGION_TYPES).to_h
 
+  # Override the auto-generated root method (via enum) to return the one, single root Region
+  def self.root
+    Region.find_by(region_type: :root)
```
rsanheim, line 26: We should only ever have one root Region.

harimohanraj89: Makes sense, we don't wanna support multiple region trees since we have orgs to split up different deployments in the same instance.

PR #1503, opened 2020-11-05
Overhaul seed data generation

rsanheim:
**Story card:** [ch1904](https://app.clubhouse.io/simpledotorg/story/1904/improve-our-seed-data-generation-in-dev)

### Overview

This PR implements a new seed data generation approach that is much faster and direct. It allows us to easily configure different parameters as we iterate so we can get to a much more production-like dataset. You can take a look at the `.env.seed.*` config files at the top of this PR to see how things are configured for different dataset sizes.

#### Create a fast, small dataset - perfect for getting a new machine or new developer up and running:

```
time rails db:seed
```

#### Create a large, production sized dataset - this takes about ~35 minutes on my iMac Pro:

```
SEED_TYPE=large rails db:seed
```

This created 470k patients with 1.6M BPs and 235k appointments across 1000 facilities.

#### Create a dataset that is like production in terms of facilities and facility groups, but has 30% of the patient data:

```
SEED_TYPE=large SCALE_FACTOR=0.30 rails db:seed
```

Full database reset to with a medium sized dataset:

```
SEED_TYPE=medium rails db:reset
```

On my imac Pro the seed part of this takes less than a minute to create ~9k patients with 22k BPs. The overall time is a bit over a minute, as it includes the database reset.

### Implementation notes

* Uses the [activerecord-import][1] gem to directly insert rows to the database in multi-value statements
* Uses the [parallel][2] gem to process multiple facilities worth of data concurrently
* Introduces [weighted random sampling][3] for many of our data choices to better match a production-like distribution

[1]: https://github.com/zdennis/activerecord-import
[2]: https://github.com/grosser/parallel
[3]: https://gist.github.com/O-I/3e0654509dd8057b539a


lib/seed/runner.rb
```diff
@@ -0,0 +1,144 @@
...
+          facility_birth_date = Faker::Time.between(from: 3.years.ago, to: 1.day.ago)
+          benchmark("[#{slug} Seeding patients for a #{facility_size} facility") do
+            patients = patients_to_create(facility_size).times.map { |num|
+              build_patient(user, oldest_registration: facility_birth_date)
+            }
+            patient_result = Patient.import(patients, recursive: true)
+            result[:patient] = patient_result.ids.size
+          end
+          patient_info = facility.assigned_patients.pluck(:id, :recorded_at)
+          bp_result = BloodPressureSeeder.call(config: config, facility: facility, user: user)
```
vkrmis, line 87: Does this mean that _all_ the BPs for a patient will be at the same facility, and recorded by the same user?

lib/seed/runner.rb
```diff
@@ -0,0 +1,144 @@
...
+    def build_patient(user, oldest_registration:)
+      recorded_at = Faker::Time.between(from: oldest_registration, to: 1.day.ago)
+      FactoryBot.build(:patient,
+        address: nil,
+        recorded_at: recorded_at,
+        registration_user: user,
+        registration_facility: user.facility)
+    end
+
+    def create_appts(patient_info, user)
```
vkrmis, line 126: Can we tie appointments to the patient's BPs/encounters rather than generate them separately? We can keep the `some patients dont get appointments` behaviour

That way we'd end up with a more organic pattern of visits -> follow up appointments. Scheduled date should tie into a previous encounter's creation date so we have some patients going overdue, LTFU etc when we seed this data.

vkrmis:
Looks mostly good to scale data generation, left some comments inline.

Taking a step back, wdyt about making Encounters the primary model here, rather than BPs? That will give us a good base to incrementally add in various pieces going forward.

So we'd generate `n` encounters per patient and go from there for what data to create at each encounter - BPs, Drugs, Sugars, Appointments etc.

rsanheim:
> Taking a step back, wdyt about making Encounters the primary model here, rather than BPs? That will give us a good base to incrementally add in various pieces going forward.
> 
> So we'd generate `n` encounters per patient and go from there for what data to create at each encounter - BPs, Drugs, Sugars, Appointments etc.

I think thats a good idea - would just require moving around some of the logic here to start w/ encounters and go from there.

rsanheim:
Got this working on review apps via #1545 - so sandbox should be pretty straightforward as well.

claudiovallejo (approved):

PR #1504, opened 2020-11-06
Fix ethiopia address model

vkrmis:
**Story card:** [ch1884](https://app.clubhouse.io/simpledotorg/story/1884/fix-ethiopia-address-fields)

## Because

We accidentally inverted the hierarchy of Zones and Woredas in our initial model

## This addresses

Flips these two fields to fix the hierarchy, and includes a data migration to fix existing data


kitallis (approved):
This looks good @vkrmis, I'm not sure if we need the migration since I think we're going to be re-imported the Facilities in the local languages, but it doesn't hurt to port them over regardless – so it's 👍

vkrmis:
> This looks good @vkrmis, I'm not sure if we need the migration since I think we're going to be re-importing the Facilities in the local languages, but it doesn't hurt to port them over regardless – so it's 👍
> 
> It'd be great if you could update the description etc. and cut a card for this if it's not already there.

Done, I was hoping to get a look at Ethiopia prod to see if the data migration was necessary or not. Demo doesn't have much going on.

PR #1505, opened 2020-11-06
Restrict teleconsult mo addition

harimohanraj89:
**Story card:** [ch1641](https://app.clubhouse.io/simpledotorg/story/1641/mos-outside-of-facility-group-can-be-added-to-facility)

## Because

If an MO is added to facility's teleconsult list, but the MO is from a different facility group, they won't be able to see patient records anyway.

## This addresses

Updates the MO search box to return results only in the current facility group.

### But shouldn't we add a validation to the facility or MO model too?

I considered adding a validation to, say, the join table between facilities and MOs, but introducing a restriction like that in the data model becomes complex very fast. For example, what if you add an MO to a facility, then reassign the MO user to a different facility group? To be thorough, we'd have to, say, destroy all teleconsult join records when the MO is switched to another facility group. I chose not to add this complexity.

Instead, I will:
* [x] Add a controller-level check on the facility controller's `update` action to strip out user IDs outside the FG.


kitallis (approved):
This mostly lgtm 👍 I assume, this missing scenario hasn't happened yet and is just an edge-case we knew about?

harimohanraj89:
Yup, exactly 👍

PR #1507, opened 2020-11-06
Better URL friendly slugs via slug candidates

rsanheim:
append the region type and then a short UUID to try to find a unique
slug for Region.  These will be better than things like:

 `region-name-54cf800b-1a85-4562-936e-95a6482483e4`


timcheadle (approved):

app/models/region.rb
```diff
@@ -26,6 +26,18 @@ def self.root
     Region.find_by(region_type: :root)
   end
 
+  def slug_candidates
+    [
```
kitallis, line 30: Is this data-structure something that FriendlyID understands?

harimohanraj89: According to [this doc](https://norman.github.io/friendly_id/file.Guide.html#Candidates) it would appear yes.

app/models/region.rb
```diff
@@ -26,6 +26,18 @@ def self.root
...
+      :name,
+      [:name, :region_type],
+      [:name, :region_type, :short_uuid]
+    ]
+  end
+
+  def short_uuid
+    SecureRandom.uuid[0..7]
+  end
```
harimohanraj89, line 39: Ultimate fallback, throw some randomness at the end? Nice.

harimohanraj89 (approved):
Looks great to me. I thought FriendlyID was good at appending a simple sequence ID to the end, like `_1`. Was it in fact appending the ID to the end, which in our case is an unwieldy UUID?

rsanheim:
> ooks great to me. I thought FriendlyID was good at appending a simple sequence ID to the end, like `_1`. Was it in fact appending the record's ID to the end, which in our case is an unwieldy UUID?

It actually appends just a new random UUID - https://github.com/norman/friendly_id/blob/e9a591b09c70e788fb35a7f45c5e146e94355f2c/lib/friendly_id/slugged.rb#L320-L326.  It used to append plain numerics but they moved away from that to avoid race conditions with concurrent scenarios.

PR #1509, opened 2020-11-09
Fix ENV var access

vkrmis:
**Story card:** none

## Because

A [previous PR](https://github.com/simpledotorg/simple-server/pull/1504) had breaking ENV access in a data migration

## This addresses

Fixes the breaking access ^


kitallis (approved):
@vkrmis Can you make sure you've test this locally?

PR #1510, opened 2020-11-09
Validate presence of MOs if teleconsultation is enabled

harimohanraj89:
**Story card:** [ch1888](https://app.clubhouse.io/simpledotorg/story/1888/disallow-enabling-teleconsultation-if-no-mos-are-added-to-facility)

## Because

You shouldn't be able to enable teleconsultation if there are no MOs added to the facility.

## This addresses

Adds a presence validation for telemedicine MOs if teleconsultation is enabled without any associated MOs. It's not very pretty, but I think it'll do for this edge case.

## Screenshots

---

Screen Shot 2020-11-09 at 11 15 03 PM

---

Screen Shot 2020-11-09 at 11 15 05 PM


spec/models/facility_spec.rb
```diff
@@ -2,18 +2,18 @@
...
-    it { should have_many(:prescription_drugs) }
-    it { should have_many(:patients).through(:encounters) }
-    it { should have_many(:appointments) }
-    it { should have_many(:teleconsultations) }
-    it { should have_and_belong_to_many(:teleconsultation_medical_officers) }
-
-    it { should have_many(:registered_patients).class_name("Patient").with_foreign_key("registration_facility_id") }
-    it { should have_many(:assigned_patients).class_name("Patient").with_foreign_key("assigned_facility_id") }
-    it { should have_many(:assigned_hypertension_patients).class_name("Patient").with_foreign_key("assigned_facility_id") }
+    it { is_expected.to have_many(:users) }
```
prabhanshuguptagit, line 5: Should we be adopting one over the other going forward?

prabhanshuguptagit (approved):
Looks solid👌

spec/models/facility_spec.rb
```diff
@@ -185,11 +185,28 @@
...
+    it { is_expected.to validate_presence_of(:district) }
+    it { is_expected.to validate_presence_of(:state) }
+    it { is_expected.to validate_presence_of(:country) }
+    it { is_expected.to validate_numericality_of(:pin) }
+
+    describe "teleconsultation medical officers" do
+      context "when teleconsultation is enabled" do
+        subject { Facility.new(enable_teleconsultation: true) }
+
+        specify do
```
kitallis, line 198: Make this `it` as well?

kitallis (approved):
lg

spec/models/facility_spec.rb
harimohanraj89: Oh yeah, good point. No reason for this one to be `specify`.

spec/models/facility_spec.rb
harimohanraj89: `expect` is the preferred approach, as `should` does a lot of object patching to work. `is_expected` is shorthand for `expect(subject)` 👍

PR #1511, opened 2020-11-10
Import missing IHCI blocks

prabhanshuguptagit:
**Story card:** [1342](https://app.clubhouse.io/simpledotorg/story/1342/get-a-canonical-block-zone-list)

## Because

For blocks and zones to become first-class citizens for sync boundaries, we need to ensure that block/zone data is clean. 

## This addresses

This adds a task to import block names for facilities which don't have block names set. This data is a consolidation of the CSVs that we sent out to Praveen to fill.


lib/tasks/import.rake
```diff
@@ -28,4 +30,10 @@ namespace :import do
     end
     puts "Created #{created_facilities} facilities"
   end
+
+  desc "Import IHCI facility block names"
+  task ihci_facility_block_names: :environment do
+    return "Cannot run task in this env" if CountryConfig.current[:name] != "India" && ENV["SIMPLE_SERVER_ENV"] != "production"
```
harimohanraj89, line 36: Consider removing the `production` check so we can verify in, say, Demo.

lib/tasks/scripts/import_facility_block_names.rb
```diff
@@ -0,0 +1,22 @@
...
+  def self.import(file)
+    facilities_updated = 0
+    facilities_not_found = 0
+
+    CSV.foreach(file, headers: true) do |row|
+      district = row["district"]
+      facility_name = row["facility"]
+      block = row["block"]
+
+      facility = Facility.where(name: facility_name, district: district)
```
harimohanraj89, line 11: `find_by`? Do we expect more than one?

lib/ihci/data/facility_blocks_list.csv
```diff
@@ -0,0 +1,601 @@
+state,facility,district,block,,,
```
harimohanraj89, line 1: OK for this info to be checked into the repo?

harimohanraj89 (approved):
Some questions, but looks great.

lib/tasks/scripts/import_facility_block_names.rb
prabhanshuguptagit: There's a bunch of dupe facilities in prod and it's hard to tell which is the "correct" one. Just another bad data edge case 😅 .

lib/ihci/data/facility_blocks_list.csv
prabhanshuguptagit: Should be ok, this information is publicly accessible. This does make the name of facilities public. Will check.

PR #1512, opened 2020-11-11
Freeze and prefill address fields

prabhanshuguptagit:
**Story card:** [ch1908](https://app.clubhouse.io/simpledotorg/story/1908/freeze-updating-fg-and-facility-address-fields)

## Because

We should freeze updates to the Org -> FG -> facility hierarchy so it's easier to maintain parity with `Region`. We should also set the `State` and `Country` address fields in `Facility` via the region hierarchy and not ask the user to supply these fields anymore.

## This addresses

- [X] New Facility form: District is pre-filled to FG's name, State is pre-filled to Facility group's state, Country is pre-filled to `Root.region`. State and country fields are frozen.
- [X] Edit Facility form: State and country fields are frozen.
- [x] New Facility via bulk upload: State and country fields are dropped from template and are set via `Region`
- [X] Edit FG form: Organization field is frozen on edit


app/controllers/admin/facilities_controller.rb
```diff
@@ -108,7 +108,9 @@ def upload
 
   def new_facility(attributes = nil)
     @facility_group.facilities.new(attributes).tap do |facility|
-      facility.country ||= Rails.application.config.country[:name]
+      facility.district ||= @facility_group.name
+      facility.state ||= @facility_group.region.state.name if Flipper.enabled?(:regions_prep)
```
rsanheim, line 112: Instead of gating this on the feature flag, could you instead just gate it on the existence of the the state region?  That way the state will populate correctly as soon as regions are filled.

prabhanshuguptagit: We're using the `regions_prep` flag to mean: "Region backfill has been run and all regions are in place". My worry with doing say `@facility_group.region&.state&.name` is that we might forget to clean it up and that extra code will sit around forever. The flag is easier to grep and clean.

kitallis (approved):
I have some thoughts on cleaning up the CSV code bunching up in the Facility model. But let's do that separately to avoid blocking this PR.

PR #1524, opened 2020-11-12
Remove unassociated facilities

kitallis:
**Story card:** [ch1905](https://app.clubhouse.io/simpledotorg/story/1905/remove-non-essential-data-from-ihci-production)

## Because

We have a bunch of unassociated facilities that are unused. They were originally associated but disassociated later because of bad name or something else. This feature is not useful anymore since previously facilities used to be created before FGs, this is not the case anymore and now this data will only adds more entropy to our move towards Regions.

## This addresses

Remove unassociated facilities from India:production and remove the feature on the view layer as well, just show the list of associated facilities.

Screenshot 2020-11-12 at 4 49 19 PM

### DIFF
--- a/app/views/admin/facility_groups/_form.html.erb
+++ b/app/views/admin/facility_groups/_form.html.erb
@@ -26,28 +26,20 @@
                         disabled: action_name == "edit" ? true : false %>
       <% end %>
 
-      <label>Associated Facilities</label>
-      <% if facility_group.facilities.present? %>
-        <ul class="pl-3">
-          <% facility_group.facilities.each do |facility| %>
-            <li>
-              <%= link_to facility.name, admin_facility_group_facility_path(facility_group, facility), target: :_blank %>
-            </li>
-          <% end %>
-        </ul>
-      <% end %>
+      <%= form.select(:protocol_id, protocols.order(:name).map { |protocol| [protocol.name, protocol.id] }) %>
 
-      <div class="mt-5 mb-5">
-        <h3>Diabetes enabled?</h3>
-        <div class="alert alert-primary mt-3" role="alert">
+      <div class="form-group">
+        <label class="mb-0">Diabetes enabled?</label>
+        <small class="form-text text-muted mt-0 mb-3">
           <% is_diabetes_enabled = facility_group.diabetes_enabled? %>
 
           <% if is_diabetes_enabled && facility_group.facilities.present? %>
             Diabetes management is already enabled for all facilities in this group.
           <% else %>
-            This will enable Diabetes management for all associated facilities in this group.
+            This will enable diabetes management for all associated facilities in this group.
           <% end %>
-        </div>
+        </small>
+
         <%= form.check_box :enable_diabetes_management,
             {id: :facility_group_enable_diabetes_management,
                 checked: is_diabetes_enabled,
@@ -55,8 +47,6 @@
                 onclick: "confirmDisableDM(this)"},
             true, false %>
       </div>
-
-      <%= form.select(:protocol_id, protocols.order(:name).map { |protocol| [protocol.name, protocol.id] }) %>
 
       <%= form.primary("Save facility group") %>
     <% end %>
--- a/app/views/admin/facility_groups/_form.html.erb
+++ b/app/views/admin/facility_groups/_form.html.erb
@@ -26,18 +26,19 @@
                         disabled: action_name == "edit" ? true : false %>
       <% end %>
 
-      <% if facility_group.facilities.present? %>
-        <%= form.collection_check_boxes :facility_ids, facility_group.facilities, :id, :name, label: "Associated facilities", checked_value: 1 %>
-      <% end %>
-
-      <% if Facility.where(facility_group: nil).present? %>
-        <%= form.collection_check_boxes :facility_ids, Facility.where(facility_group: nil), :id, :name, label: "Unassociated facilities", checked: true %>
-      <% else %>
-        <% if facility_group.new_record? %>
-          <div class="row">
-            <div class="col-md-9 mb-3 text-warning">All existing facilities belong to a facility group.</div>
-          </div>
+      <label>Associated Facilities</label>
+      <%= content_tag(:ul, class: "") do %>
+        <% facility_group.facilities.each do |facility| %>
+          <li>
+            <%= link_to facility.name, admin_facility_group_facility_path(facility_group, facility), target: :_blank %>
+          </li>
         <% end %>
+      <% end if facility_group.facilities.present? %>
+
+      <% if facility_group.new_record? %>
+        <div class="row">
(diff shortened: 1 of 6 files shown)


app/views/admin/facility_groups/_form.html.erb
```diff
@@ -26,18 +26,19 @@
...
+        <% facility_group.facilities.each do |facility| %>
+          <li>
+            <%= link_to facility.name, admin_facility_group_facility_path(facility_group, facility), target: :_blank %>
+          </li>
         <% end %>
+      <% end if facility_group.facilities.present? %>
+
+      <% if facility_group.new_record? %>
+        <div class="row">
+          <div class="col-md-9 mb-3 text-warning">All existing facilities belong to a facility group.</div>
```
prabhanshuguptagit, line 40: We can remove this text now.🎊

kitallis: dun

app/views/admin/facility_groups/_form.html.erb
```diff
@@ -26,19 +26,14 @@
...
-      <% end %>
-
-      <% if Facility.where(facility_group: nil).present? %>
-        <%= form.collection_check_boxes :facility_ids, Facility.where(facility_group: nil), :id, :name, label: "Unassociated facilities", checked: true %>
-      <% else %>
-        <% if facility_group.new_record? %>
-          <div class="row">
-            <div class="col-md-9 mb-3 text-warning">All existing facilities belong to a facility group.</div>
-          </div>
+      <label>Associated Facilities</label>
```
harimohanraj89, line 29: Is it indeed useful to have the associated facilities here? I wanna make sure we're being mindful of [what we learned recently](https://simpledotorg.slack.com/archives/CK95QFJA2/p1605017237290200)

app/views/admin/facility_groups/_form.html.erb
```diff
@@ -26,19 +26,14 @@
...
+      <%= content_tag(:ul, class: "pl-3") do %>
+        <% facility_group.facilities.each do |facility| %>
+          <li>
+            <%= link_to facility.name, admin_facility_group_facility_path(facility_group, facility), target: :_blank %>
+          </li>
         <% end %>
-      <% end %>
+      <% end if facility_group.facilities.present? %>
```
harimohanraj89, line 36: Could we just 

```
<% if facility_group.facilities.present? %>
  <ul>
...
```

Not sure this nifty clause saves us much codes, what do you think?

db/data/20201112075313_remove_unassociated_facilities_from_production.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveUnassociatedFacilitiesFromProduction < ActiveRecord::Migration[5.2]
+  def up
+    return if SIMPLE_SERVER_ENV != "production"
+    return if ENV["DEFAULT_COUNTRY"] != "IN"
+
+    Facility.where(facility_group_id: nil).delete_all
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
harimohanraj89, line 10: How about a no-op log instead, so it doesn't get in the way if we need to reverse another migration?

db/data/20201112075313_remove_unassociated_facilities_from_production.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveUnassociatedFacilitiesFromProduction < ActiveRecord::Migration[5.2]
+  def up
+    return if SIMPLE_SERVER_ENV != "production"
+    return if ENV["DEFAULT_COUNTRY"] != "IN"
+
+    Facility.where(facility_group_id: nil).delete_all
```
harimohanraj89, line 6: Do we need to `discard_data` instead? Are we sure there's no data attached to these facilities?

app/views/admin/facility_groups/_form.html.erb
```diff
@@ -26,19 +26,14 @@
                         disabled: action_name == "edit" ? true : false %>
       <% end %>
 
-      <% if facility_group.facilities.present? %>
-        <%= form.collection_check_boxes :facility_ids, facility_group.facilities, :id, :name, label: "Associated facilities", checked_value: 1 %>
```
harimohanraj89, line 30: This deletion means we should be able to delete something in the corresponding controller too, right?

harimohanraj89 (changes requested):
Lookin good. Some suggestions/questions below.

db/data/20201112075313_remove_unassociated_facilities_from_production.rb
kitallis: Yes, [checked](https://simpledotorg.slack.com/archives/GJ2AUUASK/p1605075304000700). Praveen also checked with the CVHOs to whom these facilities may belong to.

These are truly bad facilities, unlike, say PATH, who served some purpose in the past.

app/views/admin/facility_groups/_form.html.erb
kitallis: Maybe? I also want to make minimal changes here since the broader goal is making sure only good data remains in the system. Keeping this around makes sure there's not much radical change to this UI.

kitallis:
> Lookin good. Some suggestions/questions below.

Responded and/or addressed!

app/views/admin/facility_groups/_form.html.erb
harimohanraj89: Spatially yes, but conceptually it's a bigger change I think. In any case, I'll let you close the loop with Claudio on it. Not a blocker for this PR.

harimohanraj89: BTW this label can go inside the conditional below.

harimohanraj89 (approved):

app/views/admin/facility_groups/_form.html.erb
kitallis: Ha, I dislike trivial conditionals blocking UI elements – makes the markup hard to read. But... changed (don't feel strongly).

app/views/admin/facility_groups/_form.html.erb
kitallis: > BTW this label can go inside the conditional below.

> Update: Or we can put a "None" in gray, like the "No data to show" pattern in the reports.

Oh, good point.

> Spatially yes, but conceptually it's a bigger change I think. In any case, I'll let you close the loop with Claudio on it. Not a blocker for this PR.

Yeah, I think the idea is that we need to re-think our Region UI (including this page). This is probably the wrong time to implement (not so much to start thinking about the design) because Regions are in a very weird intermediate state.

For now, I'm mostly going by Daniel's approval of removing [this](https://simpledotorg.slack.com/archives/GJ2AUUASK/p1605075341001500).

kitallis: Removing, approved by Daniel. HAHA.

PR #1526, opened 2020-11-12
User does not need to belong to an Organization

kitallis:
## Because

Dashboard users have access to an Organization through `Access` now.
App users don't need to be associated to an Organization anyway, since they are linked via a Facility.

## This addresses

Remove `organization_id` from `users` table and related changes.


rsanheim:
@kitallis does this need anymore work besides fixing up the test suite?

PR #1527, opened 2020-11-12
Region callbacks for Facility

kitallis:
**Story card:** [ch1783](https://app.clubhouse.io/simpledotorg/story/1783/triggers-to-automatically-create-regions-in-the-background-when-new-source-models-are-created-or-updated)

## Because

This is the last of the concrete Region sources that must create a Region in the background for any updates or creates.

## This addresses

* Add a create + update callback on Facility.
* Added support to ensure that once the `regions_prep` feature flag is removed and callbacks become real, a huge number of tests won't start suddenly failing.


app/models/facility.rb
```diff
@@ -101,6 +101,31 @@ class Facility < ApplicationRecord
...
+  after_update :update_region, if: -> { Flipper.enabled?(:regions_prep) }
+
+  def create_region
+    Region.create!(
+      name: name,
+      source: self,
+      reparent_to: block_region,
+      region_type: Region.region_types[:facility]
+    )
+  end
```
rsanheim, line 116: We talked about this in our zoom sync, but I don't think we should try to create regions for a Facility in a callback...they will be created either:

* via the backfill
* or via the controllers that are directly responsible for mutating Regions

This introduces too many ways that facility/region state can change out from under us in surprising ways.

rsanheim (changes requested):

app/models/facility.rb
kitallis: I think it's OK to not do parent regions like Block through the callbacks (as status quo), but I feel really uncomfortable doing concrete data integrity procedures via the API code-paths alone.

We need a robust, single place, that ensures that Regions are always created for source Regions when they are created or edited. The backfill script does not serve this purpose.

I would like to be able to say: _I have a created a facility (regardless of the medium) and I'm guaranteed to have a Region for it, no matter what_.

app/models/facility_group.rb
```diff
@@ -36,9 +36,31 @@ def state
...
+  end
+
+  def update_region
+    region.reparent_to = state_region
+    region.name = name
+    region.save!
+  end
+
+  def state_region
+    Region.state.find_by_name(state)
```
kitallis, line 60: Remove `state` creation, just find.

spec/factories/facilities.rb
```diff
@@ -32,5 +32,20 @@
       zone { Faker::Address.block }
       facility_type { "PHC" }
     end
+
+    transient do
```
kitallis, line 36: Tests don't know how to create regions for `states` and `blocks`, since they are only created through the API endpoints. Create these regions by default.

app/models/organization.rb
```diff
@@ -16,23 +16,33 @@ class Organization < ApplicationRecord
...
+    create_region!(
+      name: name,
+      description: description,
+      reparent_to: Region.root,
+      region_type: Region.region_types[:organization]
+    )
   end
 
   def update_region
-    return unless name_changed? || description_changed?
```
kitallis, line 31: This is an unnecessary optimization that makes this callback more complicated than it needs to.

rsanheim: I'm curious - was it causing any problems?

kitallis: Not particularly, it just seemed like a _feel-good_ check rather than adding any serious value. Besides, since it's now an `after_update` callback, I don't think dirty is expected to work. And it seems like a better fit for an `after_*` callback semantically.

Also, I think if we were to defend against unnecessary updates, it should ideally be written as:

```ruby
if name_changed?
  region.name = name
elsif description_changed?
  region.description = description
end

region.save!
```

Which just seems too much for a really simple idempotent operation. 

It also tends to make it not clear whether it's happening because we want to avoid unnecessary updates or if there's a hard-constraint because of which this shouldn't be run. 

In my mind, it's simpler for me to think about this as: _no matter what, region gets updated_.

spec/controllers/admin/facility_groups_controller_spec.rb
```diff
@@ -1,6 +1,8 @@
 require "rails_helper"
 
 RSpec.describe Admin::FacilityGroupsController, type: :controller do
+  let(:organization) { FactoryBot.create(:organization) }
+  let(:protocol) { FactoryBot.create(:protocol) }
```
prabhanshuguptagit, line 5:
```suggestion
```

prabhanshuguptagit (approved):

PR #1529, opened 2020-11-13
Validate block in Facility model

prabhanshuguptagit:
**Story card:** [ch1925](https://app.clubhouse.io/simpledotorg/story/1925/restrict-state-and-block-names-in-facility)

## Because

We added a dropdown to select blocks from the dashboard, we should also be checking this at the model layer (mainly for CSVs).

## This addresses

This adds a model validation that checks if the facility's zone is in the list of zones in FG.


app/validators/admin/csv/facility_validator.rb
```diff
@@ -29,22 +29,14 @@ def duplicate_rows
...
-      import_facility = Facility.new(facility_params(facility))
+      import_facility = Facility.new(facility)
       row_errors << [row_num, import_facility.errors.full_messages.to_sentence] if import_facility.invalid?
     end
     group_row_errors(row_errors).each { |error| @errors << error } if row_errors.present?
   end
 
   private
 
-  def facility_params(facility)
```
prabhanshuguptagit, line 40: This code is unused now.

kitallis (approved):

app/models/facility.rb
```diff
@@ -239,6 +240,18 @@ def facility_is_unique
     errors.add(:facility, "already exists") if organization.present? && facility_group.present? && facility.present?
   end
 
+  def block_allowed
```
kitallis, line 243: Replace this code snippet with:

```ruby
  def block_allowed
    unless facility_group.region.blocks.pluck(:name).include?(block)
      errors.add(:zone, "not present in the facility group")
    end
```

...once https://github.com/simpledotorg/simple-server/pull/1554 is merged in.

app/models/facility.rb
```diff
@@ -88,6 +88,7 @@ class Facility < ApplicationRecord
       message: "must be added to enable teleconsultation"
     }
   validates :enable_diabetes_management, inclusion: {in: [true, false]}
+  validate :block_allowed, if: -> { Flipper.enabled?(:regions_prep) }
```
harimohanraj89, line 91: `valid_block`?

harimohanraj89 (approved):
:shipit:

PR #1531, opened 2020-11-16
Rename region family methods to be more explicit

rsanheim:
Because we sitll have source models like Facility and FacilityGroup, we
should name the methods that can move up or down the region tree very
explicitly ...other wise its very confusing if you see calls like the following:

```
  region.facilities
  region.districts
  region.facility_groups
```

Does that return actual Facilities, or Regions of type facility?

This changes the ancestor methods to have a _region suffix and the
descendant methods to a have a _regions suffix. It also gives the scopes
provided by the enum a _regions suffix for consistency.

This came up in the context or region reports /cc #1477 

[ch132](https://app.clubhouse.io/simpledotorg/story/132/block-level-reports)

### DIFF
--- a/app/models/region.rb
+++ b/app/models/region.rb
@@ -22,7 +22,6 @@ class Region < ApplicationRecord
   REGION_TYPES = %w[root organization state district block facility].freeze
   enum region_type: REGION_TYPES.zip(REGION_TYPES).to_h, _suffix: "regions"
 
-  # Override the auto-generated root method (via enum) to return the one, single root Region
   def self.root
     Region.find_by(region_type: :root)
   end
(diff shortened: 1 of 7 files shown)


kitallis (approved):
Yup, this was bothering me as well and wanted to do the same. Looks good.

app/models/region.rb
```diff
@@ -20,7 +20,7 @@ class Region < ApplicationRecord
   before_discard :remove_path
 
   REGION_TYPES = %w[root organization state district block facility].freeze
-  enum region_type: REGION_TYPES.zip(REGION_TYPES).to_h
+  enum region_type: REGION_TYPES.zip(REGION_TYPES).to_h, _suffix: "regions"
 
   # Override the auto-generated root method (via enum) to return the one, single root Region
```
prabhanshuguptagit, line 25: Removing this comment.

prabhanshuguptagit (approved):

rsanheim:
@prabhanshuguptagit thanks for the merge!

PR #1532, opened 2020-11-17
Support asdf-vm for bin/setup's Ruby installation

ogirginc:
## Because

[asdf] is a version manager which allows a unified API for all
supported languages.

[asdf]: https://github.com/asdf-vm/asdf

## This addresses

adding support for asdf-vm if the installation is done through
the provided `bin/setup` script.


kitallis:
Hi @ogirginc, thanks a ton for the submission!

This looks clean and seems like it mustn't affect existing users of `rbenv` or `rvm` in any way. However, we should probably ensure that is in fact true since none of us are `asdf` users.

Heads-up, the team is on a break in the last week of November, so if we don't get a chance to try this, we might only get to it by the first week of December or later.

Cheers!

ogirginc:
Hi @kitallis. It should be fine because, under the hood, it just [uses](https://github.com/asdf-vm/asdf-ruby#use) the `rbenv`'s `ruby-build`.

Have a nice & safe break!

PR #1544, opened 2020-11-17
Refactor State region creation

kitallis:
This roughly pertains to [ch1783](https://app.clubhouse.io/simpledotorg/story/1783/triggers-to-automatically-create-regions-in-the-background-when-new-source-models-are-created-or-updated).

## This addresses

* It deprecates the `ManageDistrictRegionService` in favor of methods on `FacilityGroup`.
* Increase test coverage of the State region / Block region creation flow.
* The last [PR](https://github.com/simpledotorg/simple-server/pull/1531) didn't migrate all calls to the region family names fully. This changes all of them.


app/models/facility_group.rb
```diff
@@ -53,14 +55,43 @@ def toggle_diabetes_management
...
+    remove_block_regions!
+  end
+
+  def create_block_regions!
+    return unless Flipper.enabled?(:regions_prep)
+    return if new_blocks.blank?
+
+    new_blocks.map { |name|
+      Region.block_regions.create!(name: name, reparent_to: region)
+    }
```
prabhanshuguptagit, line 80: Idea: we return a `.all?` on the `.map` so we can chain the methods in the controller. That will get rid of the `if .. true else false`.

app/models/facility_group.rb
```diff
@@ -53,14 +55,43 @@ def toggle_diabetes_management
     end
   end
 
-  def report_on_patients
```
kitallis, line 56: Dead code.

prabhanshuguptagit (approved):

PR #1548, opened 2020-11-19
[Seed Data] Ensure facilities regions make sense

rsanheim:
This fixes some issues I turned up when trying to use region reports in dev on a large seed data set:

* Ensure all facilities in the same facility group share the same state,
otherwise things are invalid when we start to work with them as regions
* Ensure facilities in the same facility group are distributed across
some shared blocks - previously they were each getting assigned their
own block

**Story card:** [ch1904](https://app.clubhouse.io/simpledotorg/story/1904/improve-our-seed-data-generation-in-dev)


lib/seed/facility_seeder.rb
```diff
@@ -61,15 +69,22 @@ def call
...
         facility_group_id, facility_group_name = *row
-        number_of_facilities_per_facility_group.times {
+        number_facilities = number_of_facilities_per_facility_group
+        state = Seed::FakeNames.instance.state
+        blocks = Seed::FakeNames.instance.blocks.sample(facilities_per_block)
+        number_facilities.times {
           size = weighted_facility_size_sample
           type = SIZES_TO_TYPE.fetch(size).sample
 
+          # TODO set the facility state here to match the parent district!
```
kitallis, line 79: Are you confused about something around the modeling here or is this just a comment to come back to?

lib/seed/facility_seeder.rb
```diff
@@ -61,15 +69,22 @@ def call
...
+          # also, what about ze blocks?
           attrs = {
             district: facility_group_name,
             facility_group_id: facility_group_id,
             facility_size: size,
-            facility_type: type
+            facility_type: type,
+            state: state,
```
kitallis, line 86: Makes sense.

kitallis (approved):
Approving to unblock, although I'll have to dig deeper into these script to understand more.

PR #1549, opened 2020-11-19
Only load Capybara things when running feature specs

rsanheim:
There is no reason to require chrome and load capybara when you just
want to run a model test. This shaves off some startup time for rspec
for the common unit test case.

(no story)


kitallis (approved):
makes sense & lg

harimohanraj89 (approved):
I approve this message.

PR #1552, opened 2020-11-19
Move facility group block syncing to service object

rsanheim:
**Story card:** No story, this is a refactoring to keep our region sync code clean

This came out of various reviews and discussions with @kitallis in slack.


kitallis:
@rsanheim Can we talk about this? I really really don't want to do this. I would like to avoid these in-between behavior driven service objects where simple functions on real entities will suffice for concrete non-transformational ideas.

Objects that narrowly carve out behavior (without modelling an entity) like this don't seem particularly useful (other than moving a piece code from one place to another),
Screenshot 2020-11-20 at 7 50 06 AM

What does it mean to say?
```
CreateFacilityGroup(fg).create
CreateFacilityGroup(fg).update

# or even
SomeOtherSideEffectyFacilityGroupBehaviour.create
SomeOtherSideEffectyFacilityGroupBehaviour.update
```

When there's already yet another object whose exact purpose is that,
Screenshot 2020-11-20 at 7 50 46 AM

FG/District is at the center of ensuring integral Region creation across the codebase. Here's some considerations before a FG is created in this state of mid-proliferation of Regions work:

1. A Facility Group must not exist without a Region
2. A Facility Group must be responsible for making sure that it belongs in a State region before it gets created, because 1. will not work without it.
3. A Facility Group must be responsible for seeding Blocks that it can hold, because if it doesn't, then Facilities can't be created within it.

These are almost rules for FGs, and I really would like to keep all these ideas in one-place, instead of pushing them in various other places. In fact, if we really want to go further down this path, I would rather add functions like [this](https://github.com/simpledotorg/simple-server/blob/master/app/models/user.rb#L172) to the base model than creating more objects in-between.

rsanheim:
> * A Facility Group must not exist without a Region
> * A Facility Group must be responsible for making sure that it belongs in a State region before it gets created, because 1. will not work without it.
> * A Facility Group must be responsible for seeding Blocks that it can hold, because if it doesn't, then Facilities can't be created within it.

The current callback approach already is confusing with regard to these these items. What order do they run in? How can you tell the state of the object from one callback to the next?

If we want to handle this in one place, that's fine, but littering the behavior through side-effect driven callbacks isn't really a good solution for that.

rsanheim:
> What does it mean to say?

also, just for context, this PR is still in a draft state, so the names and what not aren't finalized at all 😁

kitallis:
> The current callback approach already is confusing with regard to these these items. What order do they run in? How can you tell the state of the object from one callback to the next?
> 
> If we want to handle this in one place, that's fine, but littering the behavior through side-effect driven callbacks isn't really a good solution for that.

As it stands, only source regions are created through callbacks. States and Blocks are not. Blocks have their own UI, so it's reasonable to expect them to be created via the API paths. State will get its own UI soon, so it's reasonable to expect that to be created through the API paths as well in the future (through Organizations, ideally).

Source models don't have any such hooks. Hence, they are callbacks. It feels like putting something that is a classic temporary side-effect for integrity purposes would be a decent fit for this mechanism (and since it feels like an ok compromise, given that we aren't going the DB-triggers route).

There are no nested callbacks or a chain of multiple callbacks firing at the same time. When you create a Facility, a region is guaranteed for creation. The object that you get back is a Facility object. It simply ends there.

> also, just for context, this PR is still in a draft state, so the names and what not aren't finalized at all 😁

Ah, yeah. That's fine. I just wanted to drop some early considerations about these decisions since I was curious about the need for this PR as it came out just a day after we merged the callback / UI PRs. And also before we even actually used it in production 🎤

app/controllers/admin/facility_groups_controller.rb
```diff
@@ -89,24 +96,4 @@ def facility_group_params
...
-
-  def update_params
-    facility_group_params.except(:state)
-  end
-
-  def create_successful?
-    @facility_group.create_state_region!
-
-    if @facility_group.save && @facility_group.toggle_diabetes_management
-      @facility_group.update_block_regions!
```
rsanheim, line 101: The above three operations were all operating in their own separate transaction, creating a lot of places where things can fail and we are left with a FacilityGroup and/or Region in an invalid state that is out of sync. This PR is refactoring towards a better pattern where we can better see all the things going on w/ create and update, so that we can keep everything in a single transaction for data integrity.

app/controllers/admin/facility_groups_controller.rb
```diff
@@ -89,24 +96,4 @@ def facility_group_params
...
-      true
-    end
-  end
-
-  def update_successful?
-    if @facility_group.update(update_params) && @facility_group.toggle_diabetes_management
-      @facility_group.update_block_regions!
```
rsanheim, line 108: Same problem here as the create above.

app/controllers/admin/facility_groups_controller.rb
kitallis: Isn't the whole action wrapped in a single transaction and all 3-4 things happen together? We can move that inside the function if it reads better, I think. But logically, it seems like it's doing the right thing.

app/models/facility_group_region_sync.rb
```diff
@@ -0,0 +1,51 @@
+# This object keeps Facility Group in sync with related regions
+#
+# These callbacks are medium-term temporary.
+# This class and the Region callbacks should ideally be totally superseded by the Region class.
+# Keep the callbacks simple (avoid branching and optimization), idempotent (if possible) and loud when things break.
+class FacilityGroupRegionSync < SimpleDelegator
```
kitallis, line 6: Thinks this makes sense, overall 👍

app/controllers/admin/facility_groups_controller.rb
rsanheim: Yup, but wrapping the entire action is too broad. We really want the transaction only around the database level concerns, and the controller action will be doing other things beyond just saving the models (it could kick off jobs, or send an email, or render the view...etc).

I don't know the behavior if a transaction were to fail when the view has already started rendering, but I don't want to try and find out :). We just want to minimize the scope of transactions in general to the actual model level changes that need it.

kitallis: That's fair, I think it could just be within `create_successful?` and `update_successful?` – which I guess is what you've effectively done.

rsanheim: Yeah, I'd like to have it in the model layer entirely somehow, but that gets awkward with how much the controller is coupled to the model now via virtual attrs.  This is a good enough first step.

timcheadle (approved):
Much cleaner approach 👍

rsanheim (commented):
@simpledotorg/server-developers Leaving some comments to explain what is all going on here. I'm pretty happy with how much more readable the FG controller spec is now, and want to continue to move away from implicit state in our specs to more explicit wherever possible, even if that means more lines of code. The payoff in terms of readability is massive.

PR #1554, opened 2020-11-20
Refactor Facility CSV upload 

kitallis:
## Because

Became very unwieldy to add simple new validations on Facility like [this](https://github.com/simpledotorg/simple-server/pull/1529) easily.

## This addresses

It pulls out the CSV parser and CSV-specific validations into their own classes.

The validations need to validate both on a collection and on a per-row basis, they are both handled within the same class. The model validations are also triggered from within for good measure.


app/validators/csv/facilities_validator.rb
```diff
@@ -0,0 +1,107 @@
...
+    include Memery
+
+    validates :name, presence: true
+    validates :organization_name, presence: true
+    validates :facility_group_name, presence: true
+    validate :facility_is_unique
+    validate :organization_exists
+    validate :facility_group_exists
+
+    attr_accessor(*Facility.new.attributes.keys.map(&:to_sym))
```
prabhanshuguptagit, line 73: Do we need accessors for all the attributes, looks like just `:name` will do?

app/parsers/csv/facilities_parser.rb
```diff
@@ -0,0 +1,110 @@
...
+      next if facility.values.all?(&:blank?)
+      facilities << Facility.new(attach_meta(facility))
+    end
+  end
+
+  def extract_facility(row)
+    COLUMNS.map { |attr, col_name| [attr, row[col_name]] }.to_h
+  end
+
+  def attach_meta(facility_attrs)
```
prabhanshuguptagit, line 78: Could break this up into a `default_blanks_to_false` and `set_region_data`.

app/validators/csv/facilities_validator.rb
kitallis: True, I just became lazy, will fix.

prabhanshuguptagit (approved):

PR #1556, opened 2020-11-30
Enable block level sync per user

prabhanshuguptagit:
**Story card:** -

## Because

When we enable block level sync for a user, we need to make sure the facilities are touched in order to trigger a resync. [Slack ref](https://simpledotorg.slack.com/archives/C018TPYH3J9/p1605886730074600)

## This addresses

This adds a quick script to make enabling block level sync for users easy.

### DIFF
--- a/lib/tasks/enable_block_level_sync.rake
+++ b/lib/tasks/enable_block_level_sync.rake
@@ -8,7 +8,7 @@ task enable_block_level_sync: :environment do |_t, args|
     if user
       enable_block_level_sync(user)
     else
-      Rails.logger.info "User #{user_id} not found" unless user
+      Rails.logger.info "User #{user_id} not found"
     end
   end
 end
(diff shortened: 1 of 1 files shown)


kitallis (approved):
looks good.

lib/tasks/enable_block_level_sync.rake
```diff
@@ -0,0 +1,17 @@
+desc "Enable block level sync for users"
+task enable_block_level_sync: :environment do |_t, args|
+  # bundle exec enable_block_level_sync[user_id_1,user_id_2,...]
+
+  user_ids = args.extras
+  Rails.logger.info user_ids.map { |user_id|
```
kitallis, line 6: @prabhanshuguptagit can you inline this logger, so we get separate logs for each user?

PR #1558, opened 2020-12-01
Explicitly require seed lib

rsanheim:
we need this for factories that depend on it now

Fix https://sentry.io/organizations/resolve-to-save-lives/issues/2066819385

no story


kitallis (approved):
lgtm

PR #1559, opened 2020-12-02
Include facility breakdown in facilitydistrict cohort report downlodas

harimohanraj89:
**Story card:** [ch1947](https://app.clubhouse.io/simpledotorg/story/1947/facility-district-reports-cohort-download-bug)

## Because

Per-facility breakdown was missing from the Bathinda and Mansa cohort report downloads.

## This addresses

Treats `FacilityDistrict` like a `District` for downloads, which adds back the per-facility breakdowns for Bathinda and Mansa cohort reports.


kitallis (approved):
looks good.

PR #1560, opened 2020-12-02
Fix some issues around regions preparatory work

kitallis:
## Because

We enabled the flag on SBX and tested all the work around these [PRs](https://github.com/simpledotorg/simple-server/pulls?q=is%3Apr+is%3Aclosed+label%3A%22block+level+sync%22) and found a few issues.

## This addresses

- [x] Bulk upload facility was not creating regions for facilities
- [x] State presence validation (on Facility) runs even if FG doesn't exist during bulk upload (this causes confusing error messages)
- [x] Valid block validation (on Facility) runs even if FG doesn't exist
- [x] Dedupe block creation in the UI by trimming the names to reduce erroneous names

### DIFF
--- a/app/jobs/import_facilities_job.rb
+++ b/app/jobs/import_facilities_job.rb
@@ -12,6 +12,10 @@ class ImportFacilitiesJob < ApplicationJob
       import_facilities << import_facility
     end
 
-    Facility.import!(import_facilities, validate: true)
+    ActiveRecord::Base.transaction do
+      Facility.import!(import_facilities, validate: true)
+      # import! can't run callbacks, so we manually ensure that we create regions
+      import_facilities.each(&:make_region) if Flipper.enabled?(:regions_prep)
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


app/jobs/import_facilities_job.rb
```diff
@@ -12,6 +12,10 @@ def perform(facilities)
       import_facilities << import_facility
     end
 
-    Facility.import!(import_facilities, validate: true)
+    ActiveRecord::Base.transaction do
+      Facility.import!(import_facilities, validate: true)
+      # import! can't run callbacks, so we manually ensure that we create regions
+      import_facilities.each(&:make_region) if Flipper.enabled?(:regions_prep)
+    end
```
harimohanraj89, line 19: Drats. Even though we used model callbacks, looks like there was another entryway huh 😅  I'm not suggesting that one way was better than the other, just sympathizing 😄

harimohanraj89 (approved):
:shipit:

app/jobs/import_facilities_job.rb
kitallis: It sucks. But regardless of the region creation approach, we'd have to do this here, so I'm ok pawning this off as an "edge-case" since we're using this separate gem to make these uploads faster.

PR #1561, opened 2020-12-02
Hard delete organization data

prabhanshuguptagit:
**Story card:** [ch1905](https://app.clubhouse.io/simpledotorg/story/1905/remove-non-essential-data-from-ihci-production)

## Because

We want to clear some unused orgs in IHCI prod (PATH and Screening data Wardha) to make the transition to `Region`s easier.

## This addresses

This adds a rake task to delete an org and its associated data. 

An earlier attempt at cleanup for PATH didn't go correctly and we just deleted:
- the PATH org,
- all facility groups belonging to PATH.

This went fine because we don't have a FK constraint between facilities and FGs. Currently there are ~130 facilities (and all the associated records, patients etc) with `facility_group_id` as `nil` which all belong to PATH. Hence the script has a small method to specifically cleanup PATH data.


db/migrate/20201202074942_add_address_index_to_patient.rb
```diff
@@ -0,0 +1,5 @@
+class AddAddressIndexToPatient < ActiveRecord::Migration[5.2]
```
kitallis, line 1: Need to make this concurrent.

vkrmis (approved):
lgtm.
Is it worth discarding these records instead of deleting them?

db/migrate/20201202074942_add_address_index_to_patient.rb
```diff
@@ -0,0 +1,7 @@
...
+  disable_ddl_transaction!
+
+  def change
+    add_index :patients, :address_id, algorithm: :concurrently
```
kitallis, line 5: We need this to speed up deletes. Deletes check for foreign-keys on other tables and the [EXPLAIN](https://www.postgresql.org/message-id/49f698a3-e4f3-d4e4-f769-d29b39bbb735%40gmail.com) plans don't necessarily show that they are slow because of this.

prabhanshuguptagit:
@vkrmis We considered soft deleting it initially, but it would leave a bunch of unassociated data sitting around. It's cleaner to hard delete PATH at least, since we have got a green light to get rid of it.

PR #1513, opened 2020-11-11
Localize progress tab

vkrmis:
**Story card:** [ch1909](https://app.clubhouse.io/simpledotorg/story/1909/localize-progress-tab)

## Because
The progress tab is currently not localized

## This addresses

This extracts strings from progress tab views to prepare for localization

PR #1562, opened 2020-12-03
Move progress tab strings to the correct file

vkrmis:
**Story card:** none

## Because

An earlier [PR](https://github.com/simpledotorg/simple-server/pull/1513) for progress tab localization put the strings to be localized in `config/locales/en.yml`
This file isn't currently being translated by Transifex.

## This addresses

Move progress tab strings to `config/locales/api/en.yml`


harimohanraj89 (approved):
Makes sense.

PR #1565, opened 2020-12-04
Periodically check the state of regions

kitallis:
## Because

We don't want to get this wrong, it's crucial to proper functioning of the app.

## This addresses

Adds a layer of periodic polling to ensure everything is in good shape.


config/schedule.rb
```diff
@@ -45,3 +45,7 @@ def local(time)
     rake "reports:telemedicine"
   end
 end
+
+every 5.minutes, roles: [:cron] do
```
kitallis, line 49: All queries on a production-sized dataset run in under 500ms. This should be safe to run on production several times.

app/services/regions_integrity_check.rb
```diff
@@ -0,0 +1,138 @@
...
+    def sources_without_regions
+      (source.to_set - region.to_set).to_a
+    end
+
+    def regions_without_sources_count
+      if region.uniq.count > source.uniq.count
+        (region.to_set - source.to_set).count
+      else
+        0
+      end
```
harimohanraj89, line 128:
```
[
  (region.to_set - source.to_set).count,
  0
].max
```
?

app/services/regions_integrity_check.rb
```diff
@@ -0,0 +1,138 @@
...
+      facilities: facilities,
+      facility_groups: facility_groups,
+      blocks: blocks,
+      states: states
+    }
+  end
+
+  def organizations
+    {
+      source: Organization.pluck(:id),
```
harimohanraj89, line 39: Should we `to_set` here at the source to simplify the inspection methods below?

app/services/regions_integrity_check.rb
```diff
@@ -0,0 +1,138 @@
...
+      else
+        0
+      end
+    end
+
+    def init_inconsistencies
+      {
+        missing_regions: [],
+        missing_sources_count: 0
+      }
```
harimohanraj89, line 135: I'd just shove this data structure into `initialize`. Don't think we need the indirection.

`ok?` can ensure the first key is empty and the second key is zero.

What do you think?

app/services/regions_integrity_check.rb
```diff
@@ -0,0 +1,138 @@
...
+      @checked && (@inconsistencies == init_inconsistencies)
+    end
+
+    private
+
+    attr_writer :inconsistencies
+
+    def set_inconsistencies
+      @inconsistencies[:missing_regions] += sources_without_regions
+      @inconsistencies[:missing_sources_count] = regions_without_sources_count
```
harimohanraj89, line 116: Would this be more helpful if it collected a list of region IDs we can inspect, rather than a count?

app/services/regions_integrity_check.rb
```diff
@@ -0,0 +1,138 @@
...
+  def organizations
+    {
+      source: Organization.pluck(:id),
+      region: Region.organization_regions.pluck(:source_id)
+    }
+  end
+
+  def facilities
+    {
+      source: Facility.pluck(:id),
```
harimohanraj89, line 46: How do we need to use `with_discarded` for these integrity checks?

harimohanraj89 (changes requested):
Drats, left standalone comments by mistake. 

Review above! ☝️

app/services/regions_integrity_check.rb
kitallis: Nah, I think neither `RegionBackfill` nor callbacks care about discarded records, so it wouldn't matter if they went out of sync anyway. I **think** it's ok to not have it.

harimohanraj89: Does deleting facilities delete the region too? I'm thinking if we delete a facility, will we get a "region without source" show up in this check?

kitallis: [RegionSource](https://github.com/simpledotorg/simple-server/blob/master/app/models/region_source.rb#L4) handles this.

kitallis:
> Drats, left standalone comments by mistake.
> 
> Review above! ☝️

@harimohanraj89 So I changed the script drastically because it wasn't letting me mold it freely. I rewrote parts of it and removed the test that checks for "missing sources"; it instead checks for "duplicate regions" which is more valuable. Regions without sources is harmless, so not really something getting scared about.

I also benched the queries overall on prod-sized data and this was my samples:

```
# benchmark { RegionsIntegrityCheck.sweep }

I, [2020-12-08T14:48:08.348171 #45463]  INFO -- : Benchmarking (1012.2ms)
I, [2020-12-08T14:49:38.327484 #45463]  INFO -- : Benchmarking (1106.5ms)
I, [2020-12-08T14:49:45.556005 #45463]  INFO -- : Benchmarking (967.6ms)
I, [2020-12-08T14:49:51.610416 #45463]  INFO -- : Benchmarking (1015.1ms)
I, [2020-12-08T14:50:00.146900 #45463]  INFO -- : Benchmarking (989.6ms)
I, [2020-12-08T14:50:06.171613 #45463]  INFO -- : Benchmarking (1071.7ms)
I, [2020-12-08T14:50:20.987763 #45463]  INFO -- : Benchmarking (1056.3ms)
I, [2020-12-08T14:50:28.620415 #45463]  INFO -- : Benchmarking (1105.1ms)
```

Since it has increased in run time by 500ms from before, I have changed the schedule from 5 minutes to every hour.

app/services/regions_integrity_check.rb
```diff
@@ -0,0 +1,198 @@
...
+    self
+  end
+
+  private
+
+  def organizations
+    missing_regions =
+      (Organization.pluck(:id).to_set -
+        Region.organization_regions.pluck(:source_id).to_set
+      ).to_a
```
harimohanraj89, line 46: Do you need to `to_set`? Does that handle duplicates?

harimohanraj89 (approved):
Walked through this sanity check with you on zoom. Looks cooooool.

PR #1566, opened 2020-12-04
Pause hard-deletes and remove PATH delete tasks

kitallis:
## Because

Just a clean-up task. This script is a little scary to have around, so let's make sure we disable for now and enable it back when we need it manually (i.e. for cleaning up Screening Data Wardha on production)

## This addresses

* Disallow running the `hard_delete` script, force people to make code-changes to use.
* Remove PATH delete tasks.


harimohanraj89 (approved):
Axe path destruction, and disable generic org destruction. Neat.

PR #1568, opened 2020-12-07
Region backfill: create region for duplicate sources too

prabhanshuguptagit:
**Story card:** -

## Because

The region backfill script skipped creating regions for some duplicate facilities when we ran it in prod. We fixed regions for those facilities manually, this is a fix for the bug.


kitallis (approved):
although I don't think we'll run backfill ever again, but this change makes sense and is more robust

PR #1569, opened 2020-12-07
Create regions when importing source models

kitallis:
**Story card:** TBD
## Because

Importing using the `activerecord-import` gem bypasses most of the callbacks that Rails supports, since it runs in it's own codepath and uses a separate bulk query. This means it cannot reliably guarantee creating regions on its own.

This is problematic because the new seed script never created any regions for source models.

## This addresses

Add a `Facility#import_with_regions` and `FacilityGroup#import_with_regions`.

### DIFF
--- a/app/jobs/import_facilities_job.rb
+++ b/app/jobs/import_facilities_job.rb
@@ -12,10 +12,6 @@ class ImportFacilitiesJob < ApplicationJob
       import_facilities << import_facility
     end
 
-    ActiveRecord::Base.transaction do
-      Facility.import!(import_facilities, validate: true)
-      # import! can't run callbacks, so we manually ensure that we create regions
-      import_facilities.each(&:make_region) if Flipper.enabled?(:regions_prep)
-    end
+    Facility.import_with_regions(import_facilities, validate: true)
   end
 end
--- a/lib/seed/facility_seeder.rb
+++ b/lib/seed/facility_seeder.rb
@@ -64,7 +64,7 @@ module Seed
       facility_groups = number_of_facility_groups.times.map {
         FactoryBot.build(:facility_group, organization_id: organization.id, state: nil)
       }
-      fg_result = FacilityGroup.import(facility_groups, returning: [:id, :name], on_duplicate_key_ignore: true)
+      fg_result = FacilityGroup.import_with_regions(facility_groups, returning: [:id, :name], on_duplicate_key_ignore: true)
 
       facility_attrs = []
       fg_result.results.each do |row|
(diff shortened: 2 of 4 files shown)


lib/seed/facility_seeder.rb
```diff
@@ -64,7 +64,7 @@ def call
       facility_groups = number_of_facility_groups.times.map {
         FactoryBot.build(:facility_group, organization_id: organization.id, state: nil)
       }
-      fg_result = FacilityGroup.import(facility_groups, returning: [:id, :name], on_duplicate_key_ignore: true)
+      fg_result = FacilityGroup.import_with_regions(facility_groups, returning: [:id, :name], on_duplicate_key_ignore: true)
```
rsanheim, line 67: We won't need this when https://github.com/simpledotorg/simple-server/pull/1571 lands.

app/jobs/import_facilities_job.rb
```diff
@@ -12,10 +12,6 @@ def perform(facilities)
       import_facilities << import_facility
     end
 
-    ActiveRecord::Base.transaction do
-      Facility.import!(import_facilities, validate: true)
-      # import! can't run callbacks, so we manually ensure that we create regions
-      import_facilities.each(&:make_region) if Flipper.enabled?(:regions_prep)
-    end
+    Facility.import_with_regions(import_facilities, validate: true)
```
rsanheim, line 15: Think we need some tests that exercise code paths with and without regions prep turned on.

What happens if someone imports facilities with new blocks?  We probably should handle that in the loop above on line 8, assuming `facilities` is the attributes from a CSV?

kitallis: So, the tests for this job are dumb because the job is really dumb. It doesn't bother validating here because the validation for CSVs has already happened since the errors need to be presented to the user before the job is even kickstarted. So even if this job throws errors, you know something bad has happened in the pipeline.

This new [class](https://github.com/simpledotorg/simple-server/blob/master/app/validators/csv/facilities_validator.rb) takes care of this from this [PR](https://github.com/simpledotorg/simple-server/pull/1554) (both with and without `regions_prep`) and it correctly reports missing zones if there are any since that particular [validation](https://github.com/simpledotorg/simple-server/blob/master/app/models/facility.rb#L92) already exists on the model.

lib/seed/facility_seeder.rb
kitallis: We can ditch this since you've inlined stuff, which looks good, but I was thinking I'd keep this anyway since it'd be nice to replace this

```
import_facilities.each(&:make_region) if Flipper.enabled?(:regions_prep) 
```

with

```
Facility.import_with_regions(import_facilities, validate: true)
```

in application code under `app/jobs/import_facilities_job.rb`.

kitallis:
Closing this because proposed changes are complete and any other refactoring is non-critical.

PR #1570, opened 2020-12-07
Create parent regions in seed script

kitallis:
**Story card:** TBD

## Because

Parent regions always need to be explicitly created. The seed script wasn't doing this.

## This addresses

Create states and blocks through the region factories. This builds upon https://github.com/simpledotorg/simple-server/pull/1569.

### DIFF
--- a/lib/seed/facility_seeder.rb
+++ b/lib/seed/facility_seeder.rb
@@ -62,19 +62,22 @@ module Seed
       puts "Creating #{number_of_facility_groups} FacilityGroups..."
 
       facility_groups = number_of_facility_groups.times.map {
-        FactoryBot.build(:facility_group, organization_id: organization.id, state: nil)
+        state_region = FactoryBot.create(:region, :state, name: Seed::FakeNames.instance.state, reparent_to: organization.region)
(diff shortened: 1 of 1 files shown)


lib/seed/facility_seeder.rb
```diff
@@ -62,19 +62,22 @@ def call
       puts "Creating #{number_of_facility_groups} FacilityGroups..."
 
       facility_groups = number_of_facility_groups.times.map {
-        FactoryBot.build(:facility_group, organization_id: organization.id, state: nil)
+        state_region = FactoryBot.create(:region, :state, name: Seed::FakeNames.instance.state, reparent_to: organization.region)
```
kitallis, line 65: This will create a `State` per `FacilityGroup`, is that ok?

rsanheim: Hm, we'd rather have a bunch of districts in facility groups, which more closely matches prod.  I'll take a shot at it.

rsanheim: Check out https://github.com/simpledotorg/simple-server/pull/1571, I'm also going to kick off a large benchmark run here to make sure things stay fast.

kitallis: Closing this PR.

PR #1571, opened 2020-12-08
Create all necessary regions in seed process

rsanheim:
Creating Regions in our FacilitySeeder so that things will be valid in non-prod envs.

I'm keeping this all inside the FacilitySeeder for now, because we really need this to stay fast and drive it via ActiveRecord Import. Also, the logic will get cleaner and simpler once the legacy fields are removed (`zone` and `state` fields).

### DIFF
--- a/app/models/region.rb
+++ b/app/models/region.rb
@@ -16,7 +16,9 @@ class Region < ApplicationRecord
   # To set a new path for a Region, assign the parent region via `reparent_to`, and the before_validation
   # callback will assign the new path.
   attr_accessor :reparent_to
+  attr_accessor :parent_path
   before_validation :initialize_path, if: :reparent_to
+  before_validation :_set_path_for_seeds, if: :parent_path
   before_discard :remove_path
 
   REGION_TYPES = %w[root organization state district block facility].freeze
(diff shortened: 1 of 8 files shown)


rsanheim:
Looks good at first glance:

```
[~/src/simpledotorg/simple-server (region-fixes-for-seed)]> rails c
Region.counwarning: parser/current is loading parser/ruby25, which recognizes
warning: 2.5.8-compliant syntax, but you are running 2.5.1.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
t
Running via Spring preloader in process 19618
Loading development environment (Rails 5.2.4.4)
[1] pry(main)> Region.count
{:msg=>"  \e[1m\e[35m (56.8ms)\e[0m  \e[1m\e[34mSELECT COUNT(*) FROM \"regions\" WHERE \"regions\".\"deleted_at\" IS NULL\e[0m"}
=> 1128
[2] pry(main)> Facility.count
{:msg=>"  \e[1m\e[35m (6.4ms)\e[0m  \e[1m\e[34mSELECT COUNT(*) FROM \"facilities\" WHERE \"facilities\".\"deleted_at\" IS NULL\e[0m"}
=> 941
[3] pry(main)>
```

kitallis (approved):
this looks good 👍  – hopefully if there's any pending gaps with regions or whatever, this [script](https://github.com/simpledotorg/simple-server/pull/1565) will catch it.

rsanheim:
> this looks good 👍 – hopefully if there's any pending gaps with regions or whatever, this [script](https://github.com/simpledotorg/simple-server/pull/1565) will catch it.

👍 yup we should see pretty quick if things get out of sync

app/models/region.rb
```diff
@@ -16,7 +16,9 @@ class Region < ApplicationRecord
   # To set a new path for a Region, assign the parent region via `reparent_to`, and the before_validation
   # callback will assign the new path.
   attr_accessor :reparent_to
+  attr_accessor :parent_path
   before_validation :initialize_path, if: :reparent_to
+  before_validation :_set_path_for_seeds, if: :parent_path
```
rsanheim, line 21: This is hacky, needs to be refactored or removed soon if we keep this approach for seed data.

PR #1575, opened 2020-12-09
Follow up fixes for seed process

rsanheim:
**Story card:** [ch2070](https://app.clubhouse.io/simpledotorg/story/2070/seed-data-patients-need-addresses) and [ch2071](https://app.clubhouse.io/simpledotorg/story/2071/ensure-updated-at-timestamps-are-realistic-looking-for-seed-data)

Some fixes for seed data to ensure things work with the mobile app.


lib/seed/blood_pressure_seeder.rb
```diff
@@ -82,12 +84,14 @@ def call
...
         encounters << {
           blood_pressure_id: bp_id,
+          created_at: recorded_at,
           id: SecureRandom.uuid,
           device_created_at: recorded_at,
           device_updated_at: recorded_at,
           encountered_on: recorded_at,
           facility_id: facility.id,
           patient_id: patient_id,
+          updated_at: recorded_at,
```
kitallis, line 94: `Faker::Time.between(from: recorded_at, to: 1.day.ago)` has nanosecond randomization right? If that's the case this should be mostly solid.

kitallis (approved):
looks good overall, left a small question for confirmation

.env.seed.large
```diff
@@ -1,3 +1,4 @@
+NUMBER_OF_STATES=6
```
kitallis, line 1: I think we can probably do 8-10 states in `large`.

lib/seed/blood_pressure_seeder.rb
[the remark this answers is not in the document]
rsanheim: I would assume so ? The issue wasn't so much having one or two BPs w/ the same `updated_at`, It was that they _all_ had the same timestamps because we weren't overriding updated_at, correct? If so this will be more than enough randomness.

PR #1588, opened 2020-12-09
Add task to disable block level sync

prabhanshuguptagit:
**Story card:** -

This is a follow up to https://github.com/simpledotorg/simple-server/pull/1556. This adds a rake task to make disabling block level sync for a user convenient.


lib/tasks/block_level_sync.rake
```diff
@@ -0,0 +1,47 @@
...
+        Rails.logger.info "User #{user_id} not found"
+      end
+    end
+  end
+end
+
+def set_block_level_sync(user, enable)
+  ActiveRecord::Base.transaction do
+    if user.phone_number_authentication
+      user.facility_group.facilities.update_all(updated_at: Time.current)
```
rsanheim, line 36: Why do we need to update all these facility timestamps? This will have the side effect of busting a lot of reporting related caches which use updated_at as their `cache_version`.

prabhanshuguptagit: We need to change the `updated_at` of these facilities so that these facilities are synced with the new `sync_region_id` to the app, which triggers the block-level sync for other resources. This task is going to be around for a very short term (only until we are enabling this feature on a per user basis). Once we launch to all users we won't have to do this anymore.

prabhanshuguptagit: Which also means that once we enable a user, reports of that FG will open slowly for a bit. Will keep in mind!

kitallis: Approving, since there's no real way of resolving this, it's a circular dependency effectively,

* You need the facilities to be updated to synchronize to the device
* Caching facilities depends on their updated timestamps

We can think of a separate timestamp field to set cache versions against potentially since that might be an easier change as the sync mechanism is quite deeply integrated with `updated_at`.

But given that this situation is going to be restricted to enabling/disabling one-off users, I don't think we need to do something about this immediately.

kitallis (approved):

PR #1590, opened 2020-12-10
Adjust facility performance scores based on ideal clinic performance

timcheadle:
**Story card:** [ch2059](https://app.clubhouse.io/simpledotorg/story/2059/add-facility-grades-based-on-performance-score)

## Because

Previously it was impossible for a clinic to achieve a perfect score. This change adjusts the scores, so if a clinic has the following, they get a perfect score:

- 70% control rate or higher
- 80% return visits rate (20% missed visits or less)
- 10% of OPD load registered (or higher)

## This addresses

- Updates scoring algorithm to match the PRD
- Updates specs to account for new adjusted scores
- Use a specific `Period` to get the rates from `Reports::Result` rather than relying on hash ordering


rsanheim (approved):
LGTM.  calculations look pretty straight forward

PR #1876, opened 2020-12-11
Block-level sync fixes

kitallis:
## Because

For the new apps that understand `sync_region_id`; they won't have the process token with `sync_region_id` available on the very time they sync (after the server gets deployed). This will cause the comparison for `sync_region_modified?` to fail and effectively trigger a forced resync. This should be avoided because there's no need to trigger this resync until we selectively enable block-level sync for users and/or all users eventually. When we do trigger block-sync for all users, the forced resync will be restricted to the patient data in their block; which should be smaller and less strenuous.

**New apps:**

* Request 1: `sync_region_id` header is present, no `sync_region_id` in process token
* Server triggers an unnecessary forced-resync

**Old apps:**

No change.

## This addresses

Makes an exception for a force-resync when the process token does not have a `sync_region_id`

This is the "truth table" of what should happen:

```
| header sync_region_id | processtoken sync_region_id | force_resync?     |
|-----------------------+-----------------------------+-------------------|
| blank                 | blank                       | false             |
| blank                 | set                         | false             |
| set                   | blank                       | false             |
| set                   | set                         | true if different |
```


spec/controllers/api/v3/appointments_controller_spec.rb
```diff
@@ -169,16 +169,21 @@ def create_record_list(n, options = {})
             before { request.env["HTTP_X_SYNC_REGION_ID"] = request_facility_group.id }
 
             context "when process_token's sync_region_id is empty" do
-              it "syncs facility group records from beginning of time" do
-                process_token = make_process_token(current_facility_processed_since: Time.current,
-                                                   other_facilities_processed_since: Time.current)
+              it "syncs facility group records" do
+                facility_group_records = [
```
prabhanshuguptagit, line 173: This is the culprit case: we shouldn't treat this as a special case since users _will_ receive `process_token.sync_region_id` in the next request and do a resync in following request if they need to.

kitallis:
Tested on the perf env, does the right thing 👍

vkrmis (approved):
Lgtm

harimohanraj89 (approved):
If I understand correctly, this fix works because the server will decide when to insert the `sync_region_id` into the client's process token, and that'll only happen if the feature flag is enabled for the user.

When we launch this fix and keep the feature flag off, clients will be sending requests with:
* sync region _header_ present
* sync region _process token_ absent
and we'll treat those users as normal, not forcing resync. We can then start enabling the feature gradually, and when we do, 
* the client's first request will look like above ☝️
* the server will populate the sync region in that first request's response process token
* subsequent requests will contain a sync region id in the process token

Did I get that right?

kitallis:
> If I understand correctly, this fix works because the server will decide when to insert the `sync_region_id` into the client's process token, and that'll only happen if the feature flag is enabled for the user.
> 
> When we launch this fix and keep the feature flag off, clients will be sending requests with:
> 
>     * sync region _header_ present
> 
>     * sync region _process token_ absent
>       and we'll treat those users as normal, not forcing resync. We can then start enabling the feature gradually, and when we do,
> 
>     * the client's first request will look like above ☝️
> 
>     * the server will populate the sync region in that first request's response process token
> 
>     * subsequent requests will contain a sync region id in the process token
> 
> 
> Did I get that right?


That's correct. The mistake we made earlier was forget that the *server is always responsible for updating changes to process tokens in the first request*, anything that happens after that is when changes are propagated. We got this so wrong that we literally wrote a spec describe this incorrect behavior, which we just reverted.

PR #1877, opened 2020-12-11
Metrics for block-level sync

kitallis:
**Story card:** [ch2073](https://app.clubhouse.io/simpledotorg/story/2073/add-metrics-to-measure-performance-of-block-level-sync)

## Because

Measure and track in different ways how block-sync is performing. 

## This addresses

- [x] Number of resyncs for block-sync
- [x] Users attempting block-level sync
- [x] Users enabled for block level sync via rake task
- [x] De-dupe patients against passport id
- [x] Alert if someone is in a resync loop - [datadog ref](https://app.datadoghq.com/monitors/27808323)
- [x] Improve time tracing around constituent parts of a sync request

Added a [a block-level sync dashboard on datadog](https://app.datadoghq.com/dashboard/5gh-pzf-66k/block-level-sync---prabhanshu?from_ts=1608469183462&live=true&to_ts=1608641983462) that will get populated from this data.


app/controllers/api/v3/facilities_controller.rb
```diff
@@ -16,9 +16,11 @@ def current_facility_records
   end
 
   def other_facility_records
-    Facility
-      .with_discarded
-      .updated_on_server_since(other_facilities_processed_since, limit)
+    Statsd.instance.time("other_facility_records.Facility") do
```
kitallis, line 19: These `time` calls are now timing the old sync code (post revert), but whenever we ship the new-new perf changes, the body can be swapped and the timer can stay.

app/controllers/api/v3/sync_controller.rb
```diff
@@ -35,6 +39,27 @@ def disable_audit_logs?
     false
   end
 
+  def log_block_level_sync_metrics(response_key)
```
kitallis, line 42: Centralize all the block-sync logging in one place so that the loggers don't end up being called multiple times during the code-path and are easy to find in one place.

app/controllers/api/v3/sync_controller.rb
```diff
@@ -11,10 +11,14 @@ def __sync_from_user__(params)
   end
 
   def __sync_to_user__(response_key)
-    AuditLog.create_logs_async(current_user, records_to_sync, "fetch", Time.current) unless disable_audit_logs?
+    records = records_to_sync
```
harimohanraj89, line 14: Shall we put proper ivar caching into the methods to prevent this kind of inefficiency in future?

app/services/report_duplicate_passports.rb
```diff
@@ -0,0 +1,56 @@
...
+      .having("COUNT(DISTINCT patient_id) > 1")
+      .select(:identifier)
+  end
+
+  private
+
+  def passport_assigning_facility
+    "COALESCE((metadata->'assigning_facility_id'), (metadata->'assigningFacilityUuid'))::text"
+  end
+end
```
harimohanraj89, line 56: What does the output of this report generally look like? Can we post some logs from a test run for good measure?

harimohanraj89 (changes requested):
Couple of thoughts. Lookin' pretty solid.

app/services/report_duplicate_passports.rb
kitallis: Let me fetch some logs and post a counter...

app/controllers/api/v3/sync_controller.rb
kitallis: 👍🏼 https://github.com/simpledotorg/simple-server/pull/1921 is doing a bunch of nice memoization, I think we can do this there. I'll add a task for it in the PR.

app/services/report_duplicate_passports.rb
kitallis: Logs –

Screenshot 2020-12-28 at 5 02 36 PM


Counter –

Screenshot 2020-12-28 at 5 04 19 PM

kitallis: This happens to be the actual number as it stands on production...

harimohanraj89:
Noice.

harimohanraj89 (approved):

PR #1888, opened 2020-12-14
Reports fix: Default to getting data for the current month

rsanheim:
This was still falling back to previous month, which is not what we
want anymore. It also causes regions w/ data from _just_ the current
month to raise errors.

**Story card:** [ch2091](https://app.clubhouse.io/simpledotorg/story/2091/dashboard-throwing-500-for-some-districts-in-ihci)

## Because

## This addresses


app/controllers/reports/regions_controller.rb
```diff
@@ -124,12 +124,8 @@ def set_facility_keys
     @facility_keys = [district, *facilities]
   end
 
-  def default_period
-    Period.month(Date.current.last_month.beginning_of_month).attributes
-  end
-
   def set_period
-    period_params = report_params[:period].presence || default_period
+    period_params = report_params[:period].presence || Reports::RegionService.default_period.attributes
```
rsanheim, line 128: Going to use the one `default_period` everywhere now.

This means that region reports will now show the current months data when there is no period selected, which should be better for performance reasons because that is what we are caching in Region cache warmer.

prabhanshuguptagit (approved):
Lgtm

PR #1892, opened 2020-12-16
Reports authorization fixes

rsanheim:
**Story card:** [ch2094](https://app.clubhouse.io/simpledotorg/story/2094/fix-authorization-for-region-reports)

Fixing some authorization blocks that weren't doing what was expected before.

### DIFF
--- a/app/controllers/reports/regions_controller.rb
+++ b/app/controllers/reports/regions_controller.rb
@@ -19,7 +19,9 @@ class Reports::RegionsController < AdminController
   end
 
   def show
-    authorize { current_admin.accessible_facilities(:view_reports).any? }
+    facilities = authorize { current_admin.accessible_facilities(:view_reports) }
+    p facilities
+    @region = facilities.find_by!(name: params[:id])
 
     @data = Reports::RegionService.new(region: @region, period: @period).call
     @last_registration_value = @data[:cumulative_registrations].values&.last || 0
--- a/app/controllers/reports/regions_controller.rb
+++ b/app/controllers/reports/regions_controller.rb
@@ -6,7 +6,7 @@ class Reports::RegionsController < AdminController
   before_action :set_period, only: [:show, :details, :cohort]
   before_action :set_page, only: [:details]
   before_action :set_per_page, only: [:details]
-  before_action :find_region, except: :index
+  before_action :find_region, except: [:index]
   around_action :set_time_zone
 
   def index
@@ -19,10 +19,6 @@ class Reports::RegionsController < AdminController
   end
 
   def show
-    facilities = authorize { current_admin.accessible_facilities(:view_reports) }
-    p facilities
-    @region = facilities.find_by!(name: params[:id])
-
     @data = Reports::RegionService.new(region: @region, period: @period).call
     @last_registration_value = @data[:cumulative_registrations].values&.last || 0
     @new_registrations = @last_registration_value - (@data[:cumulative_registrations].values[-2] || 0)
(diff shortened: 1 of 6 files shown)


app/controllers/reports/regions_controller.rb
```diff
@@ -134,29 +132,30 @@ def set_force_cache
...
+        owning_facility_group = block.district_region.source
+        unless current_admin.accessible_facility_groups(:view_reports).include?(owning_facility_group)
+          raise UserAccess::NotAuthorizedError
+        end
+        block
+      when "Facility"
+        current_admin.accessible_facilities(:view_reports).find_by!(slug: params[:id])
+      else
+        raise "bad region_class #{region_class}"
+      end
```
rsanheim, line 153: This is getting pretty ugly here now, but once we add Regions to our UserAccess model as a first class thing it will get much simpler -- ideally something like:

```ruby
current_admin.accessible_regions.find_by!(slug: params[:id], region_type: params[:region_type) || 
  current_admin.accessible_ancestors_of(slug: params[:id], region_type: params[:region_type)
```

The `lpath` functionality should help us get there with something like this.

app/controllers/reports/regions_controller.rb
```diff
@@ -134,29 +132,30 @@ def set_force_cache
...
+    }
+    @region = if current_admin.feature_enabled?(:region_reports)
+      region_source.region
```
rsanheim, line 156: For region reports, we want the actual `Region` ...

app/controllers/reports/regions_controller.rb
```diff
@@ -134,29 +132,30 @@ def set_force_cache
...
     else
-      Region.where(region_type: region_type).find_by!(slug: report_params[:id])
+      region_source
```
rsanheim, line 158: and for legacy reports, we want the source, so `Facility`, `FacilityGroup`, etc...

spec/models/user_access_spec.rb
```diff
@@ -70,13 +70,14 @@
 
         it "returns the organizations an admin can perform actions on" do
           permission_matrix.each do |admin, action, current_resource, expected_resources|
-            admin.accesses.create(resource: current_resource)
+            admin.accesses.create!(resource: current_resource)
```
rsanheim, line 73: All of these tests were silently re-testing the same cases for each user, because the second `create` would fail, so we wouldn't have the new user access, so we were just re-testing the first case for every user in the matrix.

Gotta use `create!` for creating models via AR in tests like this, always.

spec/models/user_access_spec.rb
```diff
@@ -70,13 +70,14 @@
...
 
             expect(admin.accessible_organizations(action)).to match_array(expected_resources),
               error_message(admin,
                 action,
                 expected_resources,
                 admin.accessible_organizations(action))
+            admin.accesses.delete_all
```
rsanheim, line 80: We have to manually clear out the user_accesses so the next iterations in the matrix succeed.

rsanheim (commented):
notes for @simpledotorg/server-developers on this

spec/models/user_access_spec.rb
kitallis: great catch 👏🏼

app/controllers/reports/regions_controller.rb
kitallis: yeah, this would become a fairly smooth query once we make it region-ready, completely alright with this density for now

app/controllers/reports/regions_controller.rb
```diff
@@ -134,29 +132,30 @@ def set_force_cache
...
-    end
-  end
-
-  def find_region_v2
-    region_type = report_params[:report_scope]
-    @region = if region_type == "facility_district"
-      FacilityDistrict.new(name: report_params[:id], scope: current_admin.accessible_facilities(:view_reports))
+    region_source = authorize {
+      case region_class
+      when "FacilityDistrict"
```
kitallis, line 137: I thought we got rid of `FacilityDistrict`?

PR #1893, opened 2020-12-16
Region breadcrumbs and show 'children data'

rsanheim:
Ads [ch1988 - region breadcrumbs](https://app.clubhouse.io/simpledotorg/story/1988/add-region-breadcrumbs) and also cleans up how we are showing 'children data' for a region, which makes it easier to show the tables for the correct things. So a District will show its blocks in the tables, and a Block will show its facilities.

This is done by asking the model itself what its children records are, so that things will work correctly regardless of the `region_reports` feature flag.


timcheadle (approved):
Solid cleanup!

PR #1894, opened 2020-12-16
Multi-lingual "Resources" print materials

claudiovallejo:
**Story card:** [ch2103](https://app.clubhouse.io/simpledotorg/story/2103/country-specific-print-materials)

## Because

India, Bangladesh, and Ethiopia program managers need access to country-specific print materials :)

## This addresses

**India print materials**
image

**Bangladesh print materials**
image

**Ethiopia print materials**
image


harimohanraj89 (approved):
Neat.

PR #1895, opened 2020-12-16
Regions report index listing

rsanheim:
**Story card:** [ch1990](https://app.clubhouse.io/simpledotorg/story/1990/better-region-listing-page)

## Because

With first class region reports we need a better index listing that will:

* show the hierarchy of regions an admin can access
* properly show and auth blocks as well as districts and facilities
* try to re-use as much of the existing 'region hierarchy' JS we already have

I had to jump thru some hoops here because user access is based on Facility and FacilityGroup, and not Region...but to really handle 'region reports', we want to determine auth rights based on a Region. So there are new auth code paths that allow us to determine what region an admin can view reports on along the hierarchy.  We will need some additional hacks to get auth around state level reports soon as well.

Once we move user access to Regions as a first class thing, this will get all much simpler, as we can use that table directly and fall back on tree to determine cascading authorization in fairly straight forward manner.


app/assets/javascripts/common/admins.js
```diff
@@ -216,8 +202,18 @@ AdminAccess.prototype = {
...
-  initialize: function () {
-    this.onAsyncLoaded()
+  onDOMLoaded: function() {
+    document.addEventListener('DOMContentLoaded', (_event) => {
+      this.resourceRowCollapseListener()
+    });
+  },
+
+  initialize: function (async = true) {
+    if(async) {
```
rsanheim, line 212: We need to initialize the collapsable behavior for async and non-async versions of the page.

app/controllers/reports/regions_controller.rb
```diff
@@ -11,11 +11,23 @@ class Reports::RegionsController < AdminController
...
-      .sort_by(&:name)
+
+    if current_admin.feature_enabled?(:regions_reports)
+      @organizations = current_admin.accessible_facilities(:view_reports)
+        .flat_map(&:organization)
+        .uniq
+        .compact
+        .sort_by(&:name)
+    else
+      @accessible_regions = current_admin.accessible_facility_regions(:view_reports).each_with_object({}) { |facility, result|
```
rsanheim, line 22: Here we grab the lowest level accessible node for the current admin, a facility, and then build the tree back up from that to be able to show the correct hierarchy in the view.

app/controllers/reports/regions_controller.rb
```diff
@@ -93,6 +105,19 @@ def whatsapp_graphics
 
   private
 
+  def accessible_district?(district)
```
rsanheim, line 108: We need a permissions cache to determine if we should show just the name of a district or block in the index view, or if we should show a link to the report of that particular region.

app/models/user_access.rb
```diff
@@ -70,18 +70,40 @@ def accessible_facilities(action)
...
+    paths = Region.district_regions.where(source_id: facility_group_ids, source_type: "FacilityGroup").pluck(:path)
+    globs = paths.map { |path| " '#{path}.*' "}.join(",")
+    Region.block_regions.where("path ? ARRAY[#{globs}]::lquery[]")
+  end
+
+  def accessible_facility_regions(action)
+    facility_ids = accessible_facilities(action).pluck("facilities.id")
+    Region.facility_regions.where(source_id: facility_ids)
+  end
+
```
rsanheim, line 89: All the above methods go thru our user_access system, and then find the source objects to determine the matching Region level authorization.

app/models/user_access.rb
```diff
@@ -70,18 +70,40 @@ def accessible_facilities(action)
...
 
     resource_ids =
       [
         manageable_facilities.pluck("facilities.id"),
         manageable_facilities.map(&:facility_group_id),
-        manageable_facilities.map { |facility| facility.organization.id }
+        manageable_facilities.map(&:organization_id),
+        manageable_facility_groups.map(&:id),
+        manageable_facility_groups.map(&:organization_id),
+        manageable_orgs.map(&:id)
```
rsanheim, line 106: These helped me get some specs passing, and seemed to match the intent of what this method is supposed to be doing...we want to show any admins for resources that the current admin has `manage` rights to, and not just at the facility level.

app/views/reports/regions/_index_tree.html.erb
```diff
@@ -0,0 +1,75 @@
...
+
+  <% regions.each do |org, regions_by_district| %>
+
+  <div class="card access-tree show">
+    <div class="access-tree__section-header">
+      <%= org.name %>
+    </div>
+
+    <ul>
+      <% regions_by_district.each do |district, regions_by_block| %>
```
rsanheim, line 11: Descend through the tree, and show the accessible _or_ just the parent regions of things that the user may have access to.

app/views/reports/regions/_index_tree.html.erb
```diff
@@ -0,0 +1,75 @@
...
+      <li>
+        <div class="access-item">
+
+          <div class="access-item__dropdown">
+            <i class="fas fa-caret-down light"></i>
+          </div>
+
+          <div class="form-check show">
+            <%= link_to_if(accessible_district?(district), district.name, reports_region_path(district, report_scope: district.region_type)) %>
```
rsanheim, line 20: `link_to_if` to use the permissions cache to see if we should like to the region or just show it as a navigation guide.

app/views/reports/regions/_index_tree.html.erb
```diff
@@ -0,0 +1,75 @@
...
+              </div>
+
+            <ul class="access-item-wrapper block">
+            <% regions.sort_by(&:name).each do |region| %>
+              <li>
+                <div class="access-item <%= region.region_type %> collasped ">
+                  <span class="spacer"></span><span class="spacer"></span><span class="spacer"></span>
+
+                  <div class="form-check show">
+                    <%= link_to region.name, reports_region_path(region, report_scope: region.region_type) %>
```
rsanheim, line 51: If we've gotten this far, we know the facility is one the user can view reports on.

app/views/reports/regions/index.html.erb
```diff
@@ -1,25 +1,35 @@
...
+        </div>
+
+      <% else %>
+
+      <div class="col-sm-5 col-lg-7">
+        <% @organizations.each do |organization| %>
+        <section>
+          <div class="card org-card">
+            <h2><%= organization.name %></h2>
+            <% current_admin.accessible_facilities(:view_reports)
```
rsanheim, line 16: All this can go away soon once we have validated this approach is prod.

spec/models/user_access_spec.rb
```diff
@@ -6,41 +6,41 @@
...
-    let!(:admin_4) { create(:admin, :call_center, :with_access, resource: organization_3) }
-    let!(:admin_5) { create(:admin, :call_center) }
-
-    let!(:manager) { create(:admin, :manager) }
-    let!(:viewer_all) { create(:admin, :viewer_all) }
-    let!(:viewer_reports_only) { create(:admin, :viewer_reports_only) }
-    let!(:call_center) { create(:admin, :call_center) }
-
-    let!(:protocol_1) { create(:protocol) }
-    let!(:protocol_drug_1) { create(:protocol_drug) }
```
rsanheim, line 43: `let!` is almost always not what we want, and makes tests harder to use and reason about...or in this case, impossible to turn on feature flags early enough to change how test objects are created.

spec/models/user_access_spec.rb
```diff
@@ -160,6 +160,31 @@
...
+          block_2 = create(:region, :block, reparent_to: facility_group_3_2.region)
+          _block_3 = create(:region, :block, name: "block_1", reparent_to: facility_group_1.region)
+          manager.accesses.create!(resource: organization_3)
+          expect(manager.accessible_block_regions(:manage)).to match_array([block_1, block_2])
+          viewer_all.accesses.create!(resource: organization_3)
+          expect(viewer_all.accessible_block_regions(:manage)).to match_array([])
+          expect(viewer_all.accessible_block_regions(:view_reports)).to match_array([block_1, block_2])
+        end
+
+        it "returns any blocks underneath any accessible_facility groups" do
```
rsanheim, line 177: We have no first class notion of block access, so we are using the owning district (facility group) as a proxy for that.

app/assets/javascripts/common/admins.js
```diff
@@ -52,9 +40,7 @@ AdminAccess.prototype = {
 
   resourceRowCollapseListener: function () {
     const collapsibleItems = [
-      this.facilityAccessItemsPadding(),
-      this.facilityAccessItemsAccessRatio(),
-      this.facilityAccessItemsDropdown()
+      document.getElementsByClassName("access-item")
```
rsanheim, line 43: This simplified how the collage handler worked so that it would respond for all levels of the hierarchy.... we don't need to grab all these elements specifically, we can use the overall access row.

rsanheim (commented):
I have a couple tests to clean up Wednesday, but otherwise this is in a pretty good state to merge and turn on feature flags for early access users.

rsanheim:
One more thing: I feel like there is some sort of really nice cohesive ystem where we can combine the path based navigation of `ltree` on Region with our polymorphic user access model to really make this **much** easier. Something like:

```
current_admin.accesible_regions_for_type(:district, access_level: :manager).to_tree # district can be whatever level you want to start from
# returns the tree structure of any accessible regions, ideally in just a few SQL queries that join Region to UserAccess

# or maybe
current_admin.accesible_child_regions_for(some_specific_district, access_level: :any).to_tree 
# give me the accessible regions just underneath some particular region
```

...but how that all works isn't clear to met yet. Hopefully we work towards that in Jan as we build out state reports. It could also help with perf in the API side of the app.

app/models/user_access.rb
kitallis: I think this is correct, yes.

app/assets/javascripts/common/admins.js
kitallis: Does this change the behavior from before or it was superfluous to begin with?

kitallis (commented):
@rsanheim I did one pass and things are making sense to me, couple of questions before I dig deeper:

* Does this change the permissions granting UI in anyway, and are we doing that in this attempt?
* We aren't adding any state permissions yet right? Just blocks?

spec/models/user_access_spec.rb
```diff
@@ -160,6 +160,31 @@
         end
       end
 
+      context "#accessible_blocks" do
+        before { Flipper.enable(:regions_prep) }
```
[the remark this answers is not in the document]
rsanheim: Will do.

app/assets/javascripts/common/admins.js
rsanheim: I think there is a slight change in how a row responds when you click on a checkbox, in that it will also collapse the row if you deselect it. I played with it a bit and it felt totally reasonable from a UX perspective.

PR #1897, opened 2020-12-17
Improve Sync To User performance (batch 2)

kitallis:

## Because

Whilst perf-testing block-syncs, we discovered various inefficencies in our sync (GET) code-paths that get exacerbated when many users re-sync at the same time.

A full list of improvements and the rationale is [here](https://github.com/simpledotorg/simple-server/issues/1904).

## This addresses

This improves how the overall query for fetching the records that are required to syncing are queried. Making some structural changes to construction of the overall query avoids making inefficient sub-selects.


app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -2,33 +2,31 @@ module Api::V3::SyncToUser
...
+        .where(patient: other_patient_records)
         .updated_on_server_since(other_facilities_processed_since, other_facilities_limit)
     end
 
     private
 
-    def records_to_sync
-      current_facility_records + other_facility_records
+    def model
+      controller_name.classify.constantize.for_sync
```
rsanheim, line 25: I think I understand why you are returning a AR scope here instead of the actual model, but its a confusing mismatch ....maybe rename the method to something like `scope` to better represent that its the base scope for other methods to build off of?  Or maybe `region_scope`, `sync_scope`, `default_sync_scope` ?

kitallis: Yeah, I wasn't happy with continuing to calling it `model`. We'd removed `model` earlier because I disliked it's hanging presence with invisible meanings. But I think `sync_scope` makes sense 👍🏼

app/models/prescription_drug.rb
```diff
@@ -21,9 +21,7 @@ class PrescriptionDrug < ApplicationRecord
   validates :is_protocol_drug, inclusion: {in: [true, false]}
   validates :is_deleted, inclusion: {in: [true, false]}
 
-  scope :syncable_to_region, ->(region) {
-    with_discarded.where(patient: Patient.syncable_to_region(region))
-  }
+  scope :for_sync, -> { with_discarded }
```
harimohanraj89, line 24: Can we add a couple tiny tests for this scope across these models?

app/models/facility.rb
```diff
@@ -209,4 +209,8 @@ def valid_block
       errors.add(:zone, "not present in the facility group")
     end
   end
+
+  def syncable_patients
+    registered_patients.with_discarded
+  end
```
harimohanraj89, line 215: Is this a bit dissonant? "Syncable" in general means you are registered in, assigned to, or plan to visit. We're using a simplified definition here because it doesn't really matter when used for facility prioritization. However, does this set us up to do things like facility-level sync incorrectly?

harimohanraj89: To address the dissonance, I suspect we'd either:
* update the query here to include all three clauses - This doesn't seem correct because it doesn't maintain feature parity in the sync API, and is also less performant probably
* update the name of this method? - Something along the lines of `prioritized_patients` or something? Making it clear that this boundary is not a facility-level sync boundary, but a looser boundary _within_ another block-or-FG-level sync?

app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -2,33 +2,30 @@ module Api::V3::SyncToUser
...
-    def region_records
-      model = controller_name.classify.constantize
-      model.syncable_to_region(current_sync_region)
-    end
-
     def current_facility_records
-      region_records
-        .where(patient: prioritized_patients)
+      model_sync_scope
+        .where(patient: current_facility.syncable_patients.pluck(:id))
```
harimohanraj89, line 7: Would a couple semantic methods help with readability?

```ruby
def current_facility_patient_ids
  @current_facility_patient_ids ||= current_facility.syncable_patients.pluck(:id)
end

def other_facility_patient_ids
  @other_facility_patient_ids ||= current_sync_region.syncable_patients.pluck(:id) - current_facility_patient_ids
end
```

app/models/region.rb
```diff
@@ -88,10 +88,10 @@ def syncable_patients
     case region_type
       when "block"
         registered_patients.with_discarded
-          .union(assigned_patients.with_discarded)
+          .or(assigned_patients.with_discarded)
           .union(appointed_patients.with_discarded)
```
harimohanraj89, line 92: Why does this remain a `union`?

harimohanraj89 (changes requested):
Couple questions/suggestions from me.

app/models/region.rb
kitallis: Can't `or` with a scope that has `includes` or `joins`.

app/controllers/concerns/api/v3/sync_to_user.rb
kitallis: They might, I can play around with this. But our general approach was to reduce the number of methods in this mixin. Every method that we add, is an opportunity for someone to override it in some random place in some controller that inherits from it. The thinner this mixin is in terms of pure method volume, the cleaner the internal API is.

I'll take a shot at this and see if it helps anyway. Will respond back here.

app/models/facility.rb
kitallis: `prioritized_patients` works I think 👍🏼

harimohanraj89 (approved):
Lookin' slick ✨

PR #1898, opened 2020-12-17
Improve Sync To User performance (batch 3)

prabhanshuguptagit:

## Because

Whilst perf-testing block-syncs, we discovered various inefficencies in our sync (GET) code-paths that get exacerbated when many users re-sync at the same time.

A full list of improvements and the rationale is [here](https://github.com/simpledotorg/simple-server/issues/1904).

## This addresses

This cleans up some cpu-intensive inefficient code from our sync to user pipeline. Most noticeably:
- Use [OJ](https://github.com/ohler55/oj) instead of the default rails serializer (`to_json`). This is a drop-in solution and improves our `render` times by ~2.5x.
- Some cleaning up in the transformers.


app/controllers/api/v3/sync_controller.rb
```diff
@@ -11,12 +11,14 @@ def __sync_from_user__(params)
   end
 
   def __sync_to_user__(response_key)
-    AuditLog.create_logs_async(current_user, records_to_sync, "fetch", Time.current) unless disable_audit_logs?
+    records = records_to_sync
+
+    AuditLog.create_logs_async(current_user, records, "fetch", Time.current) unless disable_audit_logs?
```
rsanheim, line 16: Can we remove this ? If we still need the logging, we could send it to Datadog I guess, but its adding overhead to the sync path here:

```
    records_by_class = records.group_by { |record| record.class.to_s }

    records_by_class.each do |record_class, records_for_class|
      log_data = {
        user_id: user.id,
        record_class: record_class,
        record_ids: records_for_class.map(&:id),
        action: action,
        time: time
      }.to_json
```

Note that the above is *not* done async, its done before the handoff to the sidekiq job.

If no-one has looked at these logs in the past month or so, maybe we can just drop them.

app/transformers/api/v3/transformer.rb
```diff
@@ -1,31 +1,33 @@
...
   class << self
-    def from_request(attributes_of_payload)
-      rename_attributes(attributes_of_payload, key_mapping)
+    def from_request(payload_attributes)
+      rename_attributes(payload_attributes, from_request_key_mapping)
     end
 
     def to_response(model)
-      rename_attributes(model.attributes, inverted_key_mapping).as_json
+      rename_attributes(model.attributes, to_response_key_mapping).as_json
```
rsanheim, line 8: Do we need `as_json` here ? I think `to_json` does that for us, at so `Oj.dump` would do the same.  We may be invoking `as_json` twice for every payload, worth testing at least.

app/controllers/api/v3/sync_controller.rb
```diff
@@ -11,12 +11,14 @@ def __sync_from_user__(params)
...
     render(
-      json: {
-        response_key => records_to_sync.map { |record| transform_to_response(record) },
+      json: Oj.dump({
+        response_key => records.map { |record| transform_to_response(record) },
         "process_token" => encode_process_token(response_process_token)
-      },
+      }, mode: :compat),
```
rsanheim, line 21: Have you tried using the Oj optimized versions of encoding for some of the core objects? Trying that for dates, times, arrays and hashes would be interesting to see if it helps serialize time.  See https://github.com/ohler55/oj/blob/develop/pages/JsonGem.md ... I think we would need calls like this in an initializer somewhere:

```
Oj.add_to_json(Array)
Oj.add_to_json(Hash)
etc...
```

app/controllers/api/v3/sync_controller.rb
kitallis: We debated this a fair bit, and yes that block of code is synchronous. No one has looked at them not just in the past month or so, but ever, in the last 3 years.

However, the argument for keeping this is a fail-safe I think, it's like insurance.

I'll post some numbers shortly around all of these, which will show that the time consumed by this is relatively small to the other things, which is why we didn't bother cleaning this up.

app/transformers/api/v3/transformer.rb
prabhanshuguptagit: I tried removing this in https://github.com/simpledotorg/simple-server/pull/1889/commits/3fbcacc71ed1ff898d813e7d148ad544c5ecf5a5 which caused [these specs](https://resolvetosavelives.semaphoreci.com/workflows/2a444ff2-59b3-4f1d-ac4e-5d8dbdfb55cf) to fail. I haven't dug into this fully but `as_json` does something weird with timestamps.

```
[3] pry(main)> e = Encounter.first
=> #<Encounter:0x00007fcb8b256f80
....
 encountered_on: Thu, 23 Jan 2020,
 created_at: Thu, 23 Jan 2020 07:36:27 UTC +00:00,
 updated_at: Thu, 23 Jan 2020 07:36:27 UTC +00:00>
[4] pry(main)> e.as_json
=> {
 ...
 "encountered_on"=>Thu, 23 Jan 2020,
 "created_at"=>Thu, 23 Jan 2020 07:36:27 UTC +00:00,
 "updated_at"=>Thu, 23 Jan 2020 07:36:27 UTC +00:00}
[5] pry(main)> e.as_json.as_json
=> { "encountered_on"=>"2020-01-23",
 "created_at"=>"2020-01-23T07:36:27.163Z",
 "updated_at"=>"2020-01-23T07:36:27.163Z"}
```
I didn't want to change this just in case the app is relying on this behaviour, just to decrease the surface of change. Will dig into this some more.

app/controllers/api/v3/sync_controller.rb
vkrmis: We can't drop or move these to datadog afaik.

The purpose of this is to be able to audit which users have accessed specific data. 
The example scenario we've used for this is: _a legal authority asks us for information on which users have accessed data related to x._

Unless something has changed so that this is no longer a scenario we need to provide logs for, we'll need to keep this around in some form. Datadog doesn't work because we only retain data for 30 days there.

harimohanraj89: Correcto, this is a compliance issue that we need to be bullet-proof against. We'll have to take the perf hit for it. That being said, if there is any data massaging that we can push into an async job, all the better.

prabhanshuguptagit: image

Fwiw this takes _extremely_ little time (only ~1% of the response time, ~20ms). I think we can keep this around, even without moving out the data massaging and come back to this later in a fun friday cleanup.

app/transformers/api/v3/transformer.rb
prabhanshuguptagit: Couldn't figure out a fix here. Calling `Oj.add_to_json` doesn't have any effect here either. Leaving this be for now, want to avoid breaking timestamps for apps unknowingly.

app/transformers/api/v3/transformer.rb
```diff
@@ -1,31 +1,33 @@
...
       {
         "created_at" => "device_created_at",
         "updated_at" => "device_updated_at"
       }
     end
 
-    def inverted_key_mapping
-      key_mapping.invert
+    def to_response_key_mapping
+      from_request_key_mapping.invert
```
harimohanraj89, line 30: Love this naming change.

harimohanraj89 (approved):
Looks good to me.

PR #1899, opened 2020-12-18
Add teleconsultation completion numbers

vkrmis:
**Story card:** [ch2120](https://app.clubhouse.io/simpledotorg/story/2120/add-telemed-request-completion-numbers-to-the-weekly-report)

## Because

We currently only report counts for teleconsult requests that have been marked completed (`yes`). Want to show counts for other statuses as well:

- `no`
- `waiting`
- no status set (`nil`)

## This addresses

Adds these additional counts to the report. Also changes the report schedule so we can send the reports out earlier in the day.


harimohanraj89 (approved):
Love it.

PR #1900, opened 2020-12-18
Add more efficient sync indexes

kitallis:

## Because

Simply having an `updated_at` index is in fact slower than not having it, across different access patterns for our app.

## This addresses

A better index is a composite index on `patient_id` and `updated_at` that is overall faster. More info on it [here](https://github.com/simpledotorg/simple-server/issues/1904).


harimohanraj89 (approved):
Looks good. What about Encounters? I see they are a sync resource in our API.

PR #1901, opened 2020-12-18
Remove less efficient sync indexes

kitallis:

## Because

We don't need these indexes anymore with the advent of #1900. These must be removed in order for the postgres query planner to start using the new ones.

## Note

Please don't merge this until #1900 is deployed. Ideally this and #1900 should be separate deploys, so that if the concurrent indexes take time to build, the current indexes aren't removed from underneath, temporarily making queries slower.


harimohanraj89 (approved):
Unfortunate we have to remove these for #1900 to take effect 😅  However, looks like we added these [specifically to improve sync performance](https://github.com/simpledotorg/simple-server/pull/1084).

🚀

PR #1902, opened 2020-12-18
Clean-up some pending linter TODOs

kitallis:
Remove:

- lib/bangladesh/import/to_patients.rb
- app/controllers/admin/facilities_controller.rb
- app/controllers/admins_controller.rb
- .rubocop.yml

Since all our files are standardized, as of this point, the TODO isn't necessary. The remaining exceptions can just be baked in. Add a temporary clause for `lib/tasks/hard_delete.rake` for `Lint/UnreachableCode` until it's revived and/or deleted as a part of `.standard.yml` itself.


.standard_todo.yml
```diff
@@ -2,9 +2,8 @@
 # Remove from this list as you refactor files.
 ---
 ignore:
-- bundle/bin/*
-- db/schema.rb
-- db/data_schema.rb
-- lib/bangladesh/import/to_patients.rb
-- app/controllers/admin/facilities_controller.rb
-- app/controllers/admins_controller.rb
+  - bundle/bin/*
```
kitallis, line 5: This is proper yml formatting, [as seen here](https://github.com/testdouble/standard/blob/master/.standard.yml).

harimohanraj89 (approved):
Should we do something about the hard deletion task? I'd vote to put the `hard_delete.rake` issue back into `todo` and either make the code paths reachable again (carefully of course), or axe the task. We can always revive it from the old PR. Perhaps we could simply no-op the task as well, so that it's easy to `git blame` the old PR to revive if necessary.

```ruby
def hard_delete_stuff
  # Visit <PR LINK> to revive this task
end
```

kitallis:
@harimohanraj89 I'll stick it back into the TODO for now and merge.

PR #1903, opened 2020-12-18
Add UTs to State list in India

kitallis:
**Story card:** [ch2121](https://app.clubhouse.io/simpledotorg/story/2121)

## Because

Missed adding Union Territories to State list in India

## This addresses

Adds them.


harimohanraj89 (approved):
🚀

PR #1907, opened 2020-12-21
Fix duplicate patient scopes

harimohanraj89:
**Story card:** [ch1182](https://app.clubhouse.io/simpledotorg/story/1182/number-of-total-registered-patients-is-different-in-facility-view-and-district-view)

## Because

The `with_hypertension` and `with_diabetes` scopes return duplicate patient records if a patient has, for whatever reason, more than one medical history record.

## This addresses

Teaching the scopes to de-dupe patients. We don't make attempts to correct the data or enforce correct data because:

> Has-One relationships are always tough :thinking:
> 
> On one hand, we could enforce clean data, but how? A validation? That doesn't protect against concurrent processes writing two medical histories for the same patient.
> 
> So, a supporting unique index on medical_history.patient_id? That makes it hard to do things like soft-delete medical histories, right?
> 
> So instead of cleaning up the data, I feel like we should just be robust to these kinds of data issues, and slap the necessary distinct clause on the problematic patient scopes. It makes sense conceptually to do so.

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -58,8 +58,8 @@ class Patient < ApplicationRecord
 
   scope :syncable_to_region, ->(region) { region.syncable_patients }
   scope :search_by_address, ->(term) { joins(:address).merge(Address.search_by_street_or_village(term)) }
-  scope :with_diabetes, -> { joins(:medical_history).merge(MedicalHistory.diabetes_yes) }
-  scope :with_hypertension, -> { joins(:medical_history).merge(MedicalHistory.hypertension_yes) }
+  scope :with_diabetes, -> { joins(:medical_history).merge(MedicalHistory.diabetes_yes).distinct }
+  scope :with_hypertension, -> { joins(:medical_history).merge(MedicalHistory.hypertension_yes).distinct }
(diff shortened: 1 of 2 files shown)


kitallis:
@harimohanraj89 Could you elaborate on this?

```
So, a supporting unique index on medical_history.patient_id? That makes it hard to do things like soft-delete medical histories, right?
```

harimohanraj89:
Certainly. If we slapped a unique constraint on `medical_histories.patient_id`, we'd be sure that nobody can accidentally add multiple medical histories on a patient

eg. user double-clicks a button, two concurrent process each create a medical history. With the index, one of the processes will encounter an exception.

However, if we add that index, then we can no longer do the following cleanly:
* Register a patient with a medical history
* Soft-delete the patient's medical history (for whatever reason) and create a new one.
This'll fail because even though on MH is soft deleted, it'll still cause trouble with the unique index.
I'd imagine us wanting to do this kind of soft-deletion as some sort of bad-data-recovery process, where we always soft-delete existing medical histories before creating a new one.

In summation, I wish the medical history was fields on the patient record 😅

kitallis (approved):
@harimohanraj89 Makes sense. I think this is an acceptable thing to do as long as we have the `has_one` 🚀

PR #1910, opened 2020-12-22
Remove regions prep

kitallis:
**Story card:** [ch2138](https://app.clubhouse.io/simpledotorg/story/2138/deprecate-regions-preparation-flag)

## Because

Regions are now 100% live, we can deprecate `regions_prep`.

## This addresses

* Remove it from both app and specs
* Refactor some code blocks as we go along

## Note for reviewers

Let us know if you'd like us to cut this PR up, we didn't feel the need for it since the changes were fairly straightforward and well contained. Please use the [no-whitespace mode](https://github.com/simpledotorg/simple-server/pull/1910/files?diff=unified&w=1) for reviewing, it's a lot easier to see through all the changes.


spec/controllers/admin/facility_groups_controller_spec.rb
```diff
@@ -164,49 +144,6 @@
 
       expect(response).to have_http_status(:bad_request)
     end
-  end
-
-  describe "PUT #update with regions_prep enabled" do
```
kitallis, line 169: This was a duplicated block with and without `regions_prep`, now everything is for enabled in the original block.

spec/controllers/admin/facility_groups_controller_spec.rb
```diff
@@ -241,31 +178,8 @@
     end
   end
 
-  describe "DELETE #destroy with regions prep disabled" do
```
kitallis, line 244: This was a duplicated block with and without `regions_prep`, now everything is for enabled in the original block.

spec/factories/facility_groups.rb
```diff
@@ -13,17 +13,11 @@
     protocol
 
     transient do
-      create_parent_region { Flipper.enabled?(:regions_prep) }
+      create_parent_region { true }
     end
 
     before(:create) do |facility_group, options|
       if options.create_parent_region
-        if facility_group.organization.region.nil?
```
kitallis, line 21: This should never happen anymore afaict. Everything should have a region for itself. In tests, both FGs and Facilities should always have a parent region created as well (unless specified otherwise).

spec/services/region_backfill_spec.rb
```diff
@@ -1,6 +1,8 @@
 require "rails_helper"
 
 RSpec.describe RegionBackfill, type: :model do
+  before { skip "RegionBackfill is now deprecated and should be removed eventually." }
```
kitallis, line 4: I've currently just flagged off these specs because they shouldn't need to be run anymore since regions are always created. But I'm not sure if I'm missing anything, as far as I understand, `RegionBackfill` is a kind of `regions_prep`, so it should go away with it.

spec/controllers/admin/facility_groups_controller_spec.rb
rsanheim: 👍 ⚡

rsanheim (approved):
This is gonna massively improve CI time and also make things a lot easier to reason about now that we can rely on region always being there.

prabhanshuguptagit (changes requested):
lgtm. Leaving a blocker to check on the remaining region integrity sweep errors before merge (on SBX, QA and ET I think).

PR #1911, opened 2020-12-22
Add a sort order for teleconsultations.requester_completion_status to count teleconsultations consistently

vkrmis:
**Story card:** [ch2125](https://app.clubhouse.io/simpledotorg/story/2125/teleconsult-request-counts-don-t-add-up)

## Because

Telemed reporting uses a `distinct` filter on patients to avoid counting multiple requests for each patient. However, doing this while also breaking down the requests by `requester_completion_status` was causing an inconsistency in the total counts.

## This addresses

This adds an explicit hierarchy to select requests based on their completion type `('yes' > 'no' > 'waiting' > nil)`, and loads them into memory when first read. All subsequent counts are performed on this, and totals remain consistent.


lib/tasks/scripts/telemedicine_reports.rb
```diff
@@ -384,12 +384,14 @@ def facility_measures(facility, p_start, p_end)
...
-    teleconsult_marked_waiting = teleconsult_requests.where(requester_completion_status: "waiting")
-    teleconsult_not_marked = teleconsult_requests.where(requester_completion_status: nil)
+      .select("DISTINCT ON (patient_id) patient_id, requester_completion_status, recorded_at")
+      .order(Arel.sql("patient_id, array_position(array['yes', 'no', 'waiting']::varchar[], requester_completion_status)"))
+      .to_a
+    teleconsult_records = teleconsult_requests.reject { |request| request.recorded_at.nil? }
+    teleconsult_marked_completed = teleconsult_requests.select { |request| request.requester_completion_status == "yes" }
+    teleconsult_marked_incomplete = teleconsult_requests.select { |request| request.requester_completion_status == "no" }
+    teleconsult_marked_waiting = teleconsult_requests.select { |request| request.requester_completion_status == "waiting" }
+    teleconsult_not_marked = teleconsult_requests.select { |request| request.requester_completion_status.nil? }
```
harimohanraj89, line 394: Switching to in-memory counts because of the new distincting introduced above? This seems fine for now, but will it become a problem when our teleconsult activity grows larger?

harimohanraj89 (approved):
One concern about scalability. Code changes look fine to me. If I understand correctly, sometimes a patient has multiple requests, and we're prioritizing which one we choose to mark their request completion status.

lib/tasks/scripts/telemedicine_reports.rb
vkrmis: It could become problematic. This array is teleconsults distinct by patient, per facility, per week. Right now it tends to be < 10. If this starts to become a larger number that we continue tracking, we should probably move it over to a regular dashboard instead of this weekly csv export. We could use materialized views like we do with other reports to keep it scalable.

harimohanraj89: Sounds good. Let's make sure that the activity (mem usage, code execution time, query performance) is showing up on DataDog already so we can reliably tell when this is no longer scalable :smile:

vkrmis:
> One concern about scalability. Code changes look fine to me. If I understand correctly, sometimes a patient has multiple requests, and we're prioritizing which one we choose to mark their request completion status.

Yes, for this report, we want to deduplicate multiple teleconsult requests for the same patient. So if there are _n_ teleconsults for a patient within the same week, we only count them as one request. The problem with that was adding a `where` clause to the count query would pick a different request each time, based on the completion status. One way we could avoid doing this is to invalidate "duplicate" teleconsult requests elsewhere.

harimohanraj89:
Nah, I think this approach is better. Keep all the data, and do the work at reporting-time 👍

PR #1914, opened 2020-12-23
Fix rendering issues with progress tabs

kitallis:
**Story card:** [ch2137](https://app.clubhouse.io/simpledotorg/story/2137/typos-in-progress-tab)

## Because

Some of the progress tab translations were spilling un-escaped HTML because we weren't rendering them safely.

## This addresses

There's mainly [three](https://guides.rubyonrails.org/i18n.html#using-safe-html-translations) ways to solve the problem of rendering HTML from locale files,

1. Put `_html` in front of the keys,
```yaml
hello_html: <b>hello!</b>
```
2. Put an `html` key nested under, 
```yaml
title:
  html: <b>title!</b>
```
3. Render the translations `raw`
```ruby
<%= raw t('welcome') %>
```

I've chosen 3. because it's the least invasive in terms of the translation files being modified. Since those files are auto-generated, it's better to not manually fiddle with them too much. Also, since the input is fairly static and since `raw` is a wrapper over `html_safe` we needn't worry about XSS attacks.

I've made ~most~all of the `t()` calls wrap around `raw` in case we need to add HTML in other places in the future (in addition to the ones that were actually breaking).

### DIFF
--- a/app/views/api/v3/analytics/user_analytics/show.html.erb
+++ b/app/views/api/v3/analytics/user_analytics/show.html.erb
@@ -6,7 +6,7 @@
   <meta name="MobileOptimized" content="320">
   <meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,minimum-scale=1,shrink-to-fit=no"/>
   <meta name="apple-mobile-web-app-capable" content="yes"/>
-  <title><%= t("analytics.page_title") %></title>
+  <title><%= raw t("analytics.page_title") %></title>
 
   <%= inline_stylesheet("user_analytics.css") %>
 </head>
@@ -17,7 +17,7 @@
 
   <p class="updated-date">
     <span>
-      <%= t("analytics.updated", time: @user_analytics.last_updated_at) %>
+      <%= raw t("analytics.updated", time: @user_analytics.last_updated_at) %>
     </span>
   </p>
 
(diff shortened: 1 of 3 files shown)


app/views/api/v3/analytics/user_analytics/show.html.erb
```diff
@@ -112,9 +112,9 @@
   <!-- monthly hypertension control view -->
   <div class="card">
     <a href="#"  onclick="openWindow('progress-help-controlled', 'progress-start'); return false" class="info"><span>?</span></a>
-    <h3><%= t("analytics.hypertension_controlled") %></h3>
+    <h3><%= raw t("analytics.hypertension_controlled") %></h3>
     <p class="desc">
-      <%= t("analytics.hypertension_controlled_desc") %>
+      <%= raw t("analytics.hypertension_control_desc") %>
```
kitallis, line 117: This is a separate issue, we were calling the wrong locale key, which was rendering like,

Before: https://simpledotorg.slack.com/files/UFL4Z1QBV/F01HBBWD0MB/correction_3.jpeg
After: https://receive.nilenso.com/api/download/a43b19e2-44f7-11eb-ba6a-6adbae69d9d8

(github's image uploader is broken)

harimohanraj89 (approved):
Funtimes 😅  Yeah, as long as we control the locales, and the locales are the only things we are `raw`ing here, this should be fine. Agreed, updating the locales is cumbersome - and also is no more "secure" than this approach.

PR #1915, opened 2020-12-23
Fixed the medium breakpoint of the navigation so there isn't a broken…

dburka:
… in between breakpoint between large and small.

**Story card:** [chXXXX](URL)

## Because

## This addresses


harimohanraj89 (approved):
Just took it for a spin. Looks great 👍

PR #1916, opened 2020-12-23
Rollout block-level sync

prabhanshuguptagit:
**Story card:** [2086](https://app.clubhouse.io/simpledotorg/story/2086/gradual-roll-out-block-level-sync)

## Because

We need to be able to gradually bump the rollout percentage of block level sync enabled users.

## This addresses

This adds a rake task to assist.


lib/tasks/block_level_sync.rake
```diff
@@ -28,6 +28,30 @@ namespace :block_level_sync do
...
+  desc "Bump block level sync for users"
+  task :bump_percentage, [:percentage] => :environment do |_t, args|
+    percentage = args.percentage
+
+    existing_percentage = Flipper[:block_level_sync].percentage_of_actors_value
+    existing_enabled_users = enabled_user_ids.to_set
+
+    ActiveRecord::Base.transaction do
+      Flipper.enable_percentage_of_actors(:block_level_sync, existing_percentage + percentage)
+      Facility.update_all(updated_at: Time.current)
```
rsanheim, line 41: Do we need to do this? Its going to bust every cache we have across the entire site for any sort of reporting.

rsanheim: I changed the cache key for reports, so this should be okay now.

lib/tasks/block_level_sync.rake
```diff
@@ -1,47 +1,24 @@
...
-      if enable
-        Flipper.enable(:block_level_sync, user)
-      else
-        Flipper.disable(:block_level_sync, user)
-      end
-      Rails.logger.info "Block level sync #{enable ? "enabled" : "disabled"} for #{user.id}"
-    else
-      Rails.logger.info "User #{user.id} does not have a phone number authentication"
-    end
+  desc "Bump block level sync for users"
```
harimohanraj89, line 18: Nitpick: I'd go with `set_percentage`, as we may wanna reduce the percentage too :sweat_smile:

lib/tasks/scripts/block_level_sync.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+    end
+  end
+
+  def bump(percentage)
+    existing_percentage = Flipper[:block_level_sync].percentage_of_actors_value
+    existing_enabled_user_ids = enabled_user_ids.to_set
+
+    ActiveRecord::Base.transaction do
+      Flipper.enable_percentage_of_actors(:block_level_sync, existing_percentage + percentage)
```
harimohanraj89, line 47: If we go with a "set percentage" approach, we can update this.

lib/tasks/scripts/block_level_sync.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      newly_enabled_user_ids = enabled_user_ids.to_set - existing_enabled_user_ids.to_set
+      newly_enabled_user_ids.each { |user_id| touch_facilities(User.find(user_id)) }
+
+      newly_enabled_user_ids.each do |user_id|
+        Rails.logger.info msg: "Block level sync enabled for #{user_id}",
+                          block_level_sync_enabled_user_id: user_id
```
harimohanraj89, line 54: Let's throw the new percentage into a log here as well.

lib/tasks/scripts/block_level_sync.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+    end
+
+    Rails.logger.info msg: "Triggering the cache warmer to ensure that facilities have the latest cache..."
+    Reports::RegionCacheWarmer.call
```
harimohanraj89, line 59: According to Rob's latest comment, the reports caches are no longer affected by `Facility.touch`. Can we remove this now?

lib/tasks/scripts/block_level_sync.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def bump(percentage)
+    existing_percentage = Flipper[:block_level_sync].percentage_of_actors_value
+    existing_enabled_user_ids = enabled_user_ids.to_set
+
+    ActiveRecord::Base.transaction do
+      Flipper.enable_percentage_of_actors(:block_level_sync, existing_percentage + percentage)
+
+      newly_enabled_user_ids = enabled_user_ids.to_set - existing_enabled_user_ids.to_set
+      newly_enabled_user_ids.each { |user_id| touch_facilities(User.find(user_id)) }
```
harimohanraj89, line 50: Touching facilities one _user_ at a time seems inefficient. If two users from the same FG are enabled, we'd be touching the FG's facilities twice, right? Can we instead assemble the full flat unique list of facilities and then touch them?

lib/tasks/scripts/block_level_sync.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def touch_facilities(user)
+    user
+      .facility_group
+      .facilities
+      .update_all(updated_at: Time.current)
+  end
+
+  def enabled_user_ids
+    User.all.select { |u| u.feature_enabled?(:block_level_sync) }.pluck(:id)
```
harimohanraj89, line 72: Consider something like the following to avoid instantiating a bunch of `User` objects:
```ruby
User.find_each.each_with_object([]) do |ids, user|
  ids.push(user.id) if u.feature_enabled?(:block_level_sync)
end
```

harimohanraj89: PS does `Array#pluck` work? The `pluck` after the `select` looks shady to me 🤔

spec/lib/tasks/scripts/block_level_sync_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      expect(users.map(&:block_level_sync?).count(&:itself)).to be_between(17, 21)
+    end
+
+    it "increases the percentage" do
+      users = create_list(:user, 20)
+      allow(Reports::RegionCacheWarmer).to receive(:call).and_return(true)
+
+      BlockLevelSync.bump(95)
+
+      expect(users.map(&:block_level_sync?).count(&:itself)).to be_between(17, 21)
```
harimohanraj89, line 76: Why would there be 21 users if we're only creating 20 above? 

PS is there any way to get a more precise assertion? I'm guessing we're using a range because there's some randomness. Can we instead (or also) query the the Flipper percentage for the specific feature flag?

harimohanraj89 (changes requested):
Feedback below!

lib/tasks/block_level_sync.rake
kitallis: Done! https://github.com/simpledotorg/simple-server/pull/1916/commits/8ba088d7ae6b1189ea0194942d7aecf389817025

lib/tasks/scripts/block_level_sync.rb
kitallis: Done! https://github.com/simpledotorg/simple-server/pull/1916/commits/8ba088d7ae6b1189ea0194942d7aecf389817025

lib/tasks/scripts/block_level_sync.rb
kitallis: Done! https://github.com/simpledotorg/simple-server/pull/1916/commits/8ba088d7ae6b1189ea0194942d7aecf389817025

lib/tasks/scripts/block_level_sync.rb
kitallis: Rails monkey-patches `Array` to add `pluck.` 

I also don't think this will give us any real measurable gains since there's only 6.5k users. Definitely not with the default `find_each` batch size of 1000. I've changed this regardless since you suggested some code and modified the batch size to be a `100`in https://github.com/simpledotorg/simple-server/pull/1916/commits/8ba088d7ae6b1189ea0194942d7aecf389817025.

spec/lib/tasks/scripts/block_level_sync_spec.rb
kitallis: I added another clause to check the actual percentage value here https://github.com/simpledotorg/simple-server/pull/1916/commits/8ba088d7ae6b1189ea0194942d7aecf389817025 and modified the specs accordingly.

lib/tasks/scripts/block_level_sync.rb
kitallis: I don't particularly want to change this because I don't think it matters. It's a simple script that piggy-backs on an existing method that does it by user. I doubt we'll see any useful gains by doing this.

lib/tasks/scripts/block_level_sync.rb
kitallis: I'm not certain, looking at https://github.com/simpledotorg/simple-server/commit/ba08f5ebabb8670e5bbb6d7ae56225cd63a19e26 suggests that we still use `updated_at` for cache keys?

lib/tasks/scripts/block_level_sync.rb
harimohanraj89: That's a fair point. Keeping the code simple for a transient script like this makes sense 👍

harimohanraj89 (approved):
Looks great 👌

lib/tasks/scripts/block_level_sync.rb
harimohanraj89: Closing the loop: Yup we confirmed that the `cache_key` method that's being used in the reports caches now does still use `updated_at` implicitly, so let's leave this as is.

lib/tasks/block_level_sync.rake
kitallis: @rsanheim It seems like from https://github.com/simpledotorg/simple-server/commit/ba08f5ebabb8670e5bbb6d7ae56225cd63a19e26 we're still using the builtin `cache_key` which will implicitly use an `updated_at`? Am I looking at the wrong thing?

rsanheim: No, that's correct, we are using the `updated_at` for the `cache_version` from the facility.  So touching facilities will bust cache keys implicitly...I suppose it isn't a huge deal, as they would all get re-warmed on the next nightly cache warm run.

PR #1921, opened 2020-12-25
Sync performance fixes

prabhanshuguptagit:
**Story card:** [2086](https://app.clubhouse.io/simpledotorg/story/2086/gradual-roll-out-block-level-sync)

## Because

We discovered that the perf improvements for block level sync tend to slow down FG syncing. We reverted the improvements (#1920) to unblock deploys while a fix was figured out. 

## This addresses

This combines the reverted perf improvements (#1898 and #1897) and fixes them to be FG compatible.

## Note

Please use https://github.com/simpledotorg/simple-server/issues/1926 as reference for the changes here.


app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -2,46 +2,48 @@ module Api::V3::SyncToUser
...
-          .updated_on_server_since(current_facility_processed_since, limit)
+      time(__method__) do
+        @current_facility_records ||=
+          model_sync_scope
+            .where(patient: current_facility.prioritized_patients.select(:id))
+            .updated_on_server_since(current_facility_processed_since, limit)
       end
     end
 
+    # this is performance-critical code, be careful while refactoring it
```
harimohanraj89, line 14: Let's link to the issue in this comment too.

app/controllers/concerns/api/v3/sync_to_user.rb
```diff
@@ -90,5 +92,13 @@ def sync_region_modified?
       return if process_token[:sync_region_id].blank?
       process_token[:sync_region_id] != requested_sync_region_id
     end
+
+    def time(method_name, &block)
+      raise ArgumentError, "You must supply a block" unless block
```
harimohanraj89, line 97: Is this necessary? Can we just let the `yield` below blow up with a `Block not provided`?

harimohanraj89 (approved):
I think we've done a _ton_ of research and test driving of these changes. They look good!

I have one crazy idea - what happens if we replace `current_facility_records` and `other_facility_records` with `current_facility_records` and `all_facility_records`? 

Sure, we'll be sending the current facility records twice, but is that worth it - given the alternative is a gnarly codepath that requires cautionary comments and performance observation every time we touch it?

app/controllers/concerns/api/v3/sync_to_user.rb
kitallis: Oh, right, of course 👍🏼

PR #1927, opened 2020-12-30
Clean-up unused code

kitallis:

## Because

Remove code that is clearly not in use anymore. 

## This addresses

Each commit describes what is removed explicitly. Please don't add any refactors to this PR, only remove code that is obviously and clearly not in use.


app/services/region_backfill.rb
```diff
@@ -115,4 +115,44 @@ def count_invalid(region_type)
     invalid_counts[region_type] ||= 0
     invalid_counts[region_type] += 1
   end
+
+  # Wraps a Region object and allows us to easily avoid persistence
+  # methods when we are exercising Regions in a backfill dry_run scenario
+  class DryRunRegion < SimpleDelegator
```
kitallis, line 121: Removed the `DryRunRegion` model and inlined it into `RegionBackfill` script. I've kept the script around just in case we need it in some off chance something is broke, but reduced the amount of footprint in the codebase.

harimohanraj89 (approved):
It all looks great. Looks like our review apps just invoke `db:seed`, which invokes `Seed::Runner.call`, which is the new thing 👍

PR #1928, opened 2020-12-30
Clean-up unused image assets

kitallis:

## Because

There are images in our assets that are not used. 

## This addresses

Remove them to avoid loading them and reduce asset payload size. Removes ~600kb of image assets.


harimohanraj89 (approved):
Are the images included in the asset payload, or fetched separately? 🤔 

In any case, this looks great.

PR #1929, opened 2020-12-30
Block reports fixes

harimohanraj89:
**Story card:** [ch2175](https://app.clubhouse.io/simpledotorg/story/2175/block-reports-user-wise-data-missing)

User-wise data tables in the Details section of facility reports were missing.

The underlying cause is: The Details section wasn't handling the distinction between facilities and non-facility regions correctly. Comments are inline for reviewers.

**Recommend reviewing with whitespace ignored**


app/controllers/reports/regions_controller.rb
```diff
@@ -53,8 +53,8 @@ def details
                                                        prev_periods: 6,
                                                        include_current_period: true)
 
-    if @region.respond_to?(:recent_blood_pressures)
-      @recent_blood_pressures = paginate(@region.recent_blood_pressures)
+    if @region_source.respond_to?(:recent_blood_pressures)
+      @recent_blood_pressures = paginate(@region_source.recent_blood_pressures)
```
harimohanraj89, line 57: We weren't feature flagging correctly here. Regions never have a `recent_blood_pressures` method. 

Rather than introducing the `recent_blood_pressures` method onto regions, I chose to simply always ask the region source.

app/controllers/reports/regions_controller.rb
```diff
@@ -154,7 +154,7 @@ def set_force_cache
   end
 
   def find_region
-    region_source = authorize {
+    @region_source = authorize {
```
harimohanraj89, line 157: Hold onto this so we can use it above (for recent BPs)

app/models/region.rb
```diff
@@ -89,7 +89,11 @@ def cohort_analytics(period:, prev_periods:)
   end
 
   def dashboard_analytics(period:, prev_periods:, include_current_period: true)
-    DistrictAnalyticsQuery.new(self, period, prev_periods, include_current_period: include_current_period).call
+    if facility_region?
+      FacilityAnalyticsQuery.new(self, period, prev_periods, include_current_period: include_current_period).call
+    else
+      DistrictAnalyticsQuery.new(self, period, prev_periods, include_current_period: include_current_period).call
+    end
   end
```
harimohanraj89, line 97: This is one crux of the bugfix.

app/views/reports/regions/details.html.erb
```diff
@@ -1,208 +1,206 @@
...
-                New registrations<sup>2</sup>
-              </th>
-              <th colspan="6">
-                Follow-up patients<sup>3</sup>
-              </th>
-            </tr>
-            <tr class="sorts" data-sort-method="thead">
-              <th class="row-label sort-label sort-label-small ta-center" data-sort-default>
-                Users
+<% if @region.facility_region? %>
```
harimohanraj89, line 2: This is the other crux of the bugfix.

app/views/reports/regions/details.html.erb
```diff
@@ -1,208 +1,206 @@
 <%= render "header", period_selector: false %>
-<% if @region.is_a?(Facility) %>
-  <% if current_admin.accessible_facilities(:view_reports).include?(@region) %>
```
harimohanraj89, line 3: We are already authorizing the region in the [controller](https://github.com/simpledotorg/simple-server/blob/master/app/controllers/reports/regions_controller.rb#L156-L182), so we don't need to do it again here.

spec/models/region_spec.rb
```diff
@@ -233,7 +233,68 @@
     end
   end
 
-  context "#syncable_patients" do
+  describe "#cohort_analytics" do
```
kitallis, line 236: 👍🏼

app/models/region.rb
```diff
@@ -89,7 +89,11 @@ def cohort_analytics(period:, prev_periods:)
...
 
   def dashboard_analytics(period:, prev_periods:, include_current_period: true)
-    DistrictAnalyticsQuery.new(self, period, prev_periods, include_current_period: include_current_period).call
+    if facility_region?
```
kitallis, line 92: Oh, so, since `FacilityAnalyticsQuery` was never being responded to, we never saw the user details?

app/controllers/reports/regions_controller.rb
kitallis: Hrm, since we have `cohort_analytics` and `dashboard_analytics` in regions already, does it not make sense to have `recent_blood_pressures` in there as well?

I'm thinking that we don't really ever need to call `region_source` for any actual work, it's just for internal dispatch.

kitallis (commented):
@harimohanraj89 Looks mostly good to me, but I just dropped in a question for you.

app/models/region.rb
harimohanraj89: Ha, exactly. Every region thought its dashboard analytics was supposed to be district analytics - so when we started _using_ regions in reports, the facility reports borked.

app/controllers/reports/regions_controller.rb
harimohanraj89: Let me give it a spin and see what it looks like. The things I'm worried about are 
- the recent BP knowledge getting spread across several places (even if it's fetched by the query, it'll have to be paginated in the controller).
- caching concerns. I _think_ we'll wanna keep the recent BP log realtime - I'll have to see if that plays nicely with the caching setup of the analytics queries.

harimohanraj89: @kitallis I took a look, and the caching does make it a bit awkward to handle. We currently assume that all data from our "queries" (eg. FacilityAnalyticsQuery) are cached, repeatable without fear, and represent data up to the same point in time.

This is probably a non-starter to put the recent BP log in there for now, what do you think?

kitallis: @harimohanraj89 Let's tackle this separately then, since this is a bugfix that prevents usage.

kitallis (approved):

app/controllers/reports/regions_controller.rb
rsanheim: I did some digging here to get back up to speed on all this, and the cache keys for the analytic queries are handled differently than the newer code paths that RegionService calls into.

The older analytic queries use keys that are coarser grained (notice the `mon_year` in the key below).  They rely on a TTL based expiration in production of one day. So I guess they _will_ update on their own as users retrieve pages that use those queries, with the performance hit for the first user in.

```ruby
  def cache_key
    [
      self.class.name,
      @facility.id,
      @period,
      @prev_periods,
      @from_time.to_s(:mon_year),
      CACHE_VERSION
    ].join("/")
  end
```

The newer queries use cache keys that key off `Date.current`, which work in tandem w/ the cache warmer that runs through all Regions every night to keep the data fresh for every morning. So users should never take the perf hit with these queries assuming all the caches stay fully loaded correctly throughout the day.

```ruby
  def cache_key
    "#{self.class}/#{region.cache_key}/#{report_range.begin.type}/#{Date.current}"
  end
```

> I'm thinking that we don't really ever need to call `region_source` for any actual work, it's just for internal dispatch.

This is definitely true, as we want a Region to be the only external API for these reports...that will allow us to stop caring about the source in the controllers at all once we enable `region_reports` for everyone.

So I think the path forward is to change the keys in the analytic queries to key off current date and add them to the cache warmer nightly run, and then we should be retrieve all the data from the Region and drop any knowledge of source from the controller altogether.

rsanheim (commented):
Catching back up on this issue from over the break...

PR #1931, opened 2020-12-31
Region source slug mismatch

harimohanraj89:
**Story card:** [ch2182](https://app.clubhouse.io/simpledotorg/story/2182/region-source-slug-mismatch)

## Because

The links on the index page were broken when a region's slug didn't match its source's slug. The cause of breakage was: Link were crafted using the region's slug - but when clicked, the region lookup on the report page was using the URL param to search for _source_ records. This didn't matter in most cases, but is conceptually incorrect and can become a problem.

## This addresses

* Include the region lookup under the feature flag, so we directly look up regions by slug when feature flag is active
* Reuse the convenience method `accessible_block_regions` to perform block lookup and authorization


kitallis (approved):
lg

app/controllers/reports/regions_controller.rb
```diff
@@ -153,34 +154,37 @@ def set_force_cache
...
-
   def find_region
-    @region = if current_admin.feature_enabled?(:region_reports)
-      region_source.region
+    @region ||= if current_admin.feature_enabled?(:region_reports)
+      authorize {
+        case region_class
+        when "FacilityDistrict"
+          scope = current_admin.accessible_facilities(:view_reports)
+          FacilityDistrict.new(name: report_params[:id], scope: scope)
```
kitallis, line 163: Should this not be `FacilityDistrict.new().region`?

harimohanraj89: The `FacilityDistrict` is a weird one. It doesn't have any backing regions - I think it needs to stay as is, but let me doublecheck its behavior with the flag both off and on.

harimohanraj89: Took it for a detailed spin. The special "Facility DIstrict" links still work great in both cases - and links to these pages are not exposed anywhere on the dashboard 👍

PR #1935, opened 2020-12-31
Query namespace updatesRemove MyFacilities namespace from dashboard querie

harimohanraj89:

The concept of "My Facilities" is slowly disappearing from the dashboard, and the associated pages are now just dashboard pages under the "Home" section.

To avoid coupling our queries with a specific product structure, let's remove the `MyFacilities` namespace from these queries.


rsanheim (approved):
Makes sense to me.

PR #1937, opened 2021-01-04
Purge API Usage logger

kitallis:

## Because

This logger was not used and this kind of analysis can already be acquired from datadog since we send app version headers to all the sync logs.

## This addresses

Remove the `api_usage_logger` and related files.


harimohanraj89 (approved):
Byebye

PR #1939, opened 2021-01-04
My Facilities Improvements: BP Controlled

claudiovallejo:
**Story card:** [ch2189](https://app.clubhouse.io/simpledotorg/story/2189/bp-controlled-tab)

## Because

These updates are hidden behind the `my_facilities_improvements` feature flag. This PR updates the data displayed in the "BP controlled" tab in "My Facilities". The "BP controlled" tab now displays:
* Total assigned patients
* Total registered patients
* 6-month change
* 6 months of data

## This addresses

image


app/views/my_facilities/blood_pressure_control.html.erb
```diff
@@ -1,209 +1,318 @@
...
+                <% (@start_period..@period).each do |period| %>
+                  <th class="row-label sort-label sort-label-small" data-sort-method="number" colspan="2">
+                    <%= period %>
+                  </th>
+                <% end %>
+              </tr>
+            </thead>
+            <tbody>
+              <% @facilities_by_size[size].each do |facility| %>
+                <% six_month_rate_change = @data_for_facility[facility.name]["controlled_patients_rate"][@period] - @data_for_facility[facility.name]["controlled_patients_rate"][@start_period] || 0 %>
```
rsanheim, line 67: This should be moved to a helper method.

app/views/my_facilities/blood_pressure_control.html.erb
```diff
@@ -1,209 +1,318 @@
...
+                <tr data-row="<%= facility.slug %>" data-trend-color="<%= six_month_rate_change >= 0 ? "green" : "red" %>">
+                  <td class="type-title">
+                    <%= link_to(reports_region_path(facility, report_scope: "facility")) do %>
+                      <%= facility.name %>
+                    <% end %>
+                  </td>
+                  <td>
+                    <%= number_with_delimiter(@data_for_facility[facility.name]["adjusted_registrations"].values.last || 0) %>
```
rsanheim, line 75: We should add a new helper function for this - it can follow the pattern set here: 

https://github.com/simpledotorg/simple-server/blob/85b004c202e0b79353896f2462eea75fa8c52391/app/helpers/dashboard_helper.rb#L2-L5

app/views/my_facilities/blood_pressure_control.html.erb
```diff
@@ -1,209 +1,318 @@
...
+                  <td>
+                    <%= number_with_delimiter(@data_for_facility[facility.name]["cumulative_registrations"].values.last || 0) %>
+                  </td>
+                  <td>
+                    <div class="w-80px h-20px">
+                      <canvas id=<%= facility.slug %>></canvas>
+                    </div>
+                  </td>
+                  <td class="fw-bold ta-center <%= six_month_rate_change >= 0 ? "c-green-dark" : "c-red-dark" %>">
+                    <%= six_month_rate_change >= 0 ? "+" : "" %><%= number_to_percentage(six_month_rate_change, precision: 0) %>
```
rsanheim, line 86: helper method here also pls

app/views/my_facilities/blood_pressure_control.html.erb
```diff
@@ -1,209 +1,318 @@
...
-              <input class="selected-month" name="registration_month" value="<%= registration_month %>" form="query-filters" hidden>
-              <input class="selected-year" name="registration_year" value="<%= registration_year %>" form="query-filters" hidden>
-              <input class="selected-cohort-period" name="cohort_period" value="<%= cohort_period %>" form="query-filters" hidden>
+            </tbody>
+          </table>
+        <% end %>
+      </div>
+    </div>
+  </div>
+  <script src="https://cdn.jsdelivr.net/npm/chart.js@2.8.0"></script>
```
rsanheim, line 107: everything above can go into a new partial, and everything in the else can go into another partial as well.

Basically, separate out the old view code from the new and name the partials after that...maybe `bp_control_details_v1` and `bp_control_details_v2`

app/controllers/my_facilities_controller.rb
```diff
@@ -10,6 +10,7 @@ class MyFacilitiesController < AdminController
   PERIODS_TO_DISPLAY = {quarter: 3, month: 3, day: 14}.freeze
 
   around_action :set_time_zone
+  before_action :set_period, except: [:index]
```
rsanheim, line 13: Is this filter really just for the blood_pressure_control action here? If so, it should have an `only` declaration for that instead of the `except`

rsanheim (commented):
Some comments on cleaning up the views and such.

app/views/my_facilities/blood_pressure_control.html.erb
rsanheim: And any other `number || 0` can use the same helper so we don't have to have this logic in the view

app/views/my_facilities/blood_pressure_control.html.erb
claudiovallejo: @rsanheim what's the value of moving the `a - b || 0` into it's own helper?

rsanheim: Good question. One thing it does is keeps our views simple and more readable, which makes them easier to maintain in the future. We can also test helper functions, which probably isn't necessary for how simple this is but often these things get more complicated over time.

I haven't seen how you've changed things yet, but something like this would be a lot more readable:

```
  six_month_rate_change(facility)
```

or maybe

```
  six_month_rate_change(facility, "controlled_patients_rate")
```
if its needed for other rates.

app/views/my_facilities/_bp_controlled_details_v2.html.erb
```diff
@@ -0,0 +1,106 @@
...
+              <% (@start_period..@period).each do |period| %>
+                <th class="row-label sort-label sort-label-small" data-sort-method="number" colspan="2">
+                  <%= period %>
+                </th>
+              <% end %>
+            </tr>
+          </thead>
+          <tbody>
+            <% @facilities_by_size[size].each do |facility| %>
+              <% six_month_rate_change = @data_for_facility[facility.name]["controlled_patients_rate"][@period] - @data_for_facility[facility.name]["controlled_patients_rate"][@start_period] || 0 %>
```
rsanheim, line 66: Now that the `period` s here are way clearer, you could keep it inline. 

Long term we should really get this into a view component, will make this all much clearer.

rsanheim (approved):

app/views/my_facilities/blood_pressure_control.html.erb
claudiovallejo: Thanks @rsanheim !

PR #1945, opened 2021-01-05
Discard stale appointments

vkrmis:
**Story card:** [ch1953](https://app.clubhouse.io/simpledotorg/story/1953/cannot-remove-patients-from-overdue-list)

## Because

Some patients have multiple appointments in the scheduled state. This is due to older app versions accidentally creating multiple appointments for patients. The overdue list becomes difficult to use in this state, because patients who have been contacted, and had their status updated will still show up as overdue thanks to the extra appointments.

## This addresses

Adds a rake task that:
- Discards all but the latest scheduled appointment, for patients who have multiple scheduled appointments.
- Discards far future (> 1 month) appointments in favour of closer ones
- Mutates a far future appointment and schedules it in a month if no other scheduled appointments exist


lib/tasks/data_fixes.rake
```diff
@@ -17,4 +18,17 @@ namespace :data_fixes do
...
+
+  desc "Clean up invalid scheduled appointments (multiple scheduled appointments for a patient)"
+  task discard_invalid_scheduled_appointments: :environment do
+    patients_ids = Appointment
+      .where(status: "scheduled")
+      .group(:patient_id).count
+      .select { |_k, v| v > 1 }
+      .keys
+
+    Patient.with_discarded.where("id in (?)", patients_ids).each do |patient|
```
prabhanshuguptagit, line 30:
```suggestion
    Patient.with_discarded.where(id: patients_ids).each do |patient|
```

lib/tasks/data_fixes.rake
```diff
@@ -17,4 +18,17 @@ namespace :data_fixes do
          "patients: #{patient_count}, BPs: #{bp_count}, blood sugars: #{bs_count}, "\
          "appointments: #{appointment_count}, prescriptions: #{prescription_drug_count}"
   end
+
+  desc "Clean up invalid scheduled appointments (multiple scheduled appointments for a patient)"
+  task discard_invalid_scheduled_appointments: :environment do
+    patients_ids = Appointment
```
prabhanshuguptagit, line 24: Should we pass all patients to `DiscardInvalidAppointments` and do this there? This logic and the looping could go in there then.

prabhanshuguptagit (approved):
Minor suggestions, approving to unblock.

lib/tasks/data_fixes.rake
vkrmis: Thought about it, but decided to go this way to be a little more consistent. 
The other `data_fixes` script here also operates on one (user) record at a time.

PR #1947, opened 2021-01-06
ActiveJob / ActionMailer should use sidekiq by default

kitallis:
## Because

We don't run mailers as proper sidekiq jobs because our default active job adapter is `:inline`

## This addresses

Change the adapter to be `:sidekiq`

### DIFF
--- a/config/sidekiq-heroku.yml
+++ b/config/sidekiq-heroku.yml
@@ -1,6 +1,5 @@
 :concurrency: 2
 :queues:
-  - default
-  - mailers
-  - [low, 1]
   - [high, 3]
+  - [default, 2]
+  - [low, 1]
--- a/config/sidekiq-heroku.yml
+++ b/config/sidekiq-heroku.yml
@@ -1,5 +1,6 @@
 :concurrency: 2
 :queues:
   - default
+  - mailers
   - [low, 1]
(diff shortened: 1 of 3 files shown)


config/sidekiq-heroku.yml
```diff
@@ -1,5 +1,6 @@
 :concurrency: 2
 :queues:
   - default
+  - mailers
```
harimohanraj89, line 4: We could instead just set the mailer queue to `default`. I don't think we gain much by maintaining a separate queue.

```ruby
# config/environments/production.rb
config.action_mailer.deliver_later_queue_name = 'default_mailer_queue'
```

kitallis: @harimohanraj89 I think that's fine. Done https://github.com/simpledotorg/simple-server/pull/1947/commits/ef7b4bbbe8259559d83ee4a394f7957839017991.

This also puts proper priority for `default` because otherwise it's [order](https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues) dependent.

harimohanraj89: Lol good one.

harimohanraj89 (approved):
:shipit:

PR #1948, opened 2021-01-06
Inherit the parent info while creating Facility factories

kitallis:
**Story card:** -

## Because

Creating a standalone Facility from factories can cause a potential mismatch in the Region hierarchy since all it's de-normalized parent info is hard-coded. We discovered this problem in this [demo data gen run](https://simpledotorg.slack.com/archives/CFHMC60P5/p1609916415061000).

## This addresses

Inherit the parent info from the parents.


harimohanraj89 (approved):
What happens today if someone makes a new facility in prod and enters a "district" that's different from the facility's FG name?

prabhanshuguptagit:
The district field can be set to anything right now, we don't particularly care about it. The region's district hierarchy is driven through FGs and so we pick the name from there.

PR #1949, opened 2021-01-06
Compress line list CSVs

prabhanshuguptagit:

## Because

While investigating the line list performance issues, we found that our line list attachments are exceeding the size limit on sendgrid's SMTP API. The size limit is 20 MB, which most of our medical history line lists are larger than.

## This addresses

This compresses the line lists as zip archives before attaching them. The zipped files are well under the limit (~5MB is the current worst case), so we shouldn't hit the limit soon again. `.zip` files can be opened natively on Windows 10 (doesn't need any specialised software like WinZip) and MacOS (using the native archive utility), so this shouldn't have much impact on the user experience.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -61,6 +61,7 @@ gem "roo", "~> 2.8.0"
 gem "rspec-rails", "~> 3.7"
 gem "rswag", "~> 1.6.0"
 gem "ruby-progressbar", require: false
+gem "rubyzip"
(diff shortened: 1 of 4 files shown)


harimohanraj89 (approved):
This is very cool. Rubyzip seems solid. Let's test several permutations (big FGs, small FGs, with/without med history) and make sure zipping works across the board 👍

PR #1950, opened 2021-01-06
Speed-up patient line lists

kitallis:

## Because

They are computed in real-time and aren't cached in any way, which makes them really slow.

## This addresses

Introduce a materialized version of the `PatientSummary` view called `MaterializedPatientSummary` and use that to populate the line lists.

We can start using this view for other things slowly as well,

* Overdue download
* Appointments index
* Query the view to compute `risk_level` and `days_overdue` in the patient model 

And deprecate `PatientSummary` in due time.

* [x] Make the `id` index a unique key
* [x]  Can this view be refreshed concurrently?
* [x] Eager-load the last n appointments and last n bps
* [x] Refactor `formatted_medications`
* [x] Fix timezone issues with all date operations in the matview
* [x] Double-check attachment filenames in Sandbox or review.


db/views/materialized_patient_summaries_v01.sql
```diff
@@ -0,0 +1,118 @@
+SELECT p.recorded_at,
```
kitallis, line 1: This is intentionally formatted this way so that we can see all the exposed columns properly.

db/views/materialized_patient_summaries_v01.sql
```diff
@@ -0,0 +1,118 @@
...
+       CONCAT(date_part('year', p.recorded_at), ' Q', EXTRACT(QUARTER
```
kitallis, line 2: This file is intentionally formatted this way so that we can see all the exposed columns properly (this is also valid formatting for SQL, via sqlformat.org).

app/exporters/patients_with_history_exporter.rb
```diff
@@ -93,53 +54,52 @@ def csv_headers
...
-    registration_facility = patient.registration_facility
-    assigned_facility = patient.assigned_facility
+  def csv_fields(patient_summary)
+    patient = patient_summary.patient
     latest_bps = patient.latest_blood_pressures.first(DISPLAY_BLOOD_PRESSURES + 1)
-    latest_blood_sugar = patient.latest_blood_sugar
-    latest_appointment = patient.latest_scheduled_appointment
-    latest_bp_passport = patient.latest_bp_passport
-    fetch_medication_history(patient, latest_bps.map(&:recorded_at))
+    all_medications = fetch_medication_history(patient, latest_bps.map(&:recorded_at))
```
kitallis, line 69: This pertains to the refactor (down below) to try to remove the stateful `@medications` out of this class and just pass it around instead.

app/exporters/patients_exporter.rb
```diff
@@ -1,9 +1,8 @@
-require "csv"
-
 class PatientsExporter
+  require "csv"
   include QuarterHelper
 
-  BATCH_SIZE = 20
+  BATCH_SIZE = 500
```
kitallis, line 5: This performs a bit better because we're not creating that many throwaway objects anymore, so we can strain memory a bit to decrease overall time taken.

app/exporters/patients_with_history_exporter.rb
```diff
@@ -1,50 +1,10 @@
...
-  include QuarterHelper
-
-  BATCH_SIZE = 20
-  BLOOD_SUGAR_TYPES = {
-    random: "Random",
-    post_prandial: "Postprandial",
-    fasting: "Fasting",
-    hba1c: "HbA1c"
-  }.with_indifferent_access.freeze
+class PatientsWithHistoryExporter < PatientsExporter
```
kitallis, line 1: Inherit to avoid duplication (and hence errors) for the exact same setup code.

db/migrate/20210107092746_create_materialized_patient_summaries.rb
```diff
@@ -0,0 +1,5 @@
+class CreateMaterializedPatientSummaries < ActiveRecord::Migration[5.2]
+  def change
+    create_view :materialized_patient_summaries, materialized: true
```
kitallis, line 3: Building this for the first time takes ~30-40s on the perf box. The migration will refresh it once synchronously.

db/migrate/20210107112850_add_patient_id_index_to_materialized_patient_summary.rb
```diff
@@ -0,0 +1,7 @@
+class AddPatientIdIndexToMaterializedPatientSummary < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def change
+    add_index :materialized_patient_summaries, :id, algorithm: :concurrently
```
kitallis, line 5: This is very important for making the view useful, otherwise it'll continue to be as slow. The `id` here is the patient id.

db/views/materialized_patient_summaries_v01.sql
```diff
@@ -0,0 +1,118 @@
...
+       reg_facility.facility_type                                                                                     AS registration_facility_type,
+       reg_facility.district                                                                                          AS registration_district,
+       reg_facility.state                                                                                             AS registration_state,
+       p.assigned_facility_id                                                                                         AS assigned_facility_id,
+       assigned_facility.name                                                                                         AS assigned_facility_name,
+       assigned_facility.facility_type                                                                                AS assigned_facility_type,
+       assigned_facility.district                                                                                     AS assigned_facility_district,
+       assigned_facility.state                                                                                        AS assigned_facility_state,
```
kitallis, line 25: These are new fields I've added.

db/views/materialized_patient_summaries_v01.sql
```diff
@@ -0,0 +1,118 @@
...
+       CONCAT(date_part('year', latest_blood_pressure.recorded_at), ' Q', EXTRACT(QUARTER
+                                                                                  FROM
+                                                                                  latest_blood_pressure.recorded_at)) AS latest_blood_pressure_quarter,
+       latest_blood_pressure_facility.name                                                                            AS latest_blood_pressure_facility_name,
+       latest_blood_pressure_facility.facility_type                                                                   AS latest_blood_pressure_facility_type,
+       latest_blood_pressure_facility.district                                                                        AS latest_blood_pressure_district,
+       latest_blood_pressure_facility.state                                                                           AS latest_blood_pressure_state,
+       latest_blood_sugar.id                                                                                          AS latest_blood_sugar_id,
```
kitallis, line 36: This is a new field that I've added, so we can support `to_s`.

db/views/materialized_patient_summaries_v01.sql
```diff
@@ -0,0 +1,118 @@
...
+                AND latest_blood_sugar.blood_sugar_value >= 300)
+                OR (latest_blood_sugar.blood_sugar_type = 'post_prandial'
+                    AND latest_blood_sugar.blood_sugar_value >= 300)
+                OR (latest_blood_sugar.blood_sugar_type = 'fasting'
+                    AND latest_blood_sugar.blood_sugar_value >= 200)
+                OR (latest_blood_sugar.blood_sugar_type = 'hba1c'
+                    AND latest_blood_sugar.blood_sugar_value >= 9.0)) THEN 1
+            ELSE 0
+           END)                                                                                                       AS risk_level,
+       latest_bp_passport.id                                                                                          AS latest_bp_passport_id,
```
kitallis, line 77: This is a new column added (`latest_bp_passport.id`) explicitly so we can add a `belongs_to` and do `shortcode` from ruby.

app/exporters/patients_with_history_exporter.rb
```diff
@@ -93,53 +54,52 @@ def csv_headers
...
+      "Latest Blood Sugar Type"
+    ].flatten.compact
   end
 
-  def csv_fields(patient)
-    registration_facility = patient.registration_facility
-    assigned_facility = patient.assigned_facility
+  def csv_fields(patient_summary)
+    patient = patient_summary.patient
     latest_bps = patient.latest_blood_pressures.first(DISPLAY_BLOOD_PRESSURES + 1)
```
harimohanraj89, line 68: Do we need to `includes :blood_pressures, :appointments`?

harimohanraj89 (changes requested):
One concern from me. The rest looks great. I believe our patient line list specs don't need to be updated at all, right? Sweet.

app/exporters/patients_with_history_exporter.rb
```diff
@@ -64,11 +75,16 @@ def csv_headers
...
+    latest_bps = patient_summary
+                   .latest_blood_pressures
+                   .first(DISPLAY_BLOOD_PRESSURES + 1)
+
+    all_medications = fetch_medication_history(patient_summary, latest_bps.map(&:recorded_at))
     zone_column_index = csv_headers.index(zone_column)
-    patient_appointments = patient.appointments.order(device_created_at: :desc).load
+
+    patient_appointments = patient_summary
+                             .appointments
```
prabhanshuguptagit, line 86: V nice

prabhanshuguptagit (approved):

PR #1951, opened 2021-01-06
Fix state name in FacilitySeeder

prabhanshuguptagit:
**Story card:** -

## Because

The `FacilitySeeder` creates facilities with state names like ` #<Region:0x00007fb2c8ea8af8>`.

## This addresses

Fix state names.

### DIFF
--- a/lib/seed/facility_seeder.rb
+++ b/lib/seed/facility_seeder.rb
@@ -131,7 +131,7 @@ module Seed
             facility_group_id: facility_group_id,
             facility_size: size,
             facility_type: type,
-            state: state,
+            state: state.name,
             zone: blocks.sample
           }
           facility_attrs << FactoryBot.build(:facility, :seed, attrs)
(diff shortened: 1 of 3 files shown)


harimohanraj89 (approved):

rsanheim:
These facilities already have proper state Regions attached to them via what the seed process does.

Does this mean the text field on Facility needs to be in sync w/ the Region model? If so, we should just remove it as soon as possible....its legacy and another way for things to fall out of sync.

prabhanshuguptagit:
The regions are properly associated yeah, this field doesn't _have_ to be in sync for correctness. Just something I noticed in data I generated locally.

Thinking of it, I think we can start removing it right away now that `regions_prep` is done everywhere.

rsanheim:
I pushed up a fix to make sure we are using the state names Daniel sent over, as well as some more assertions here.

rsanheim:
> Thinking of it, I think we can start removing it right away now that `regions_prep` is done everywhere.

Yeah, we should definitely do that, as its pretty confusing right now to have a text field that doesn't really match the actual region.

PR #1952, opened 2021-01-06
Upgrade ddtrace

kitallis:
## Because

There's reason to believe that the current ddtrace is not doing so well with long running jobs. [ref](https://github.com/DataDog/dd-trace-rb/issues/818), [ref](https://github.com/DataDog/dd-trace-dotnet/issues/1002).

They do not show up in the APM traces,
Screenshot 2021-01-06 at 8 47 38 PM

But we can see them in logs,
Screenshot 2021-01-06 at 8 49 49 PM

## This addresses

Upgrade ddtrace from `0.40.0` to `0.43.0`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -126,7 +126,7 @@ GEM
       thor (~> 0.20.3)
     data_migrate (6.3.0)
       rails (>= 5.0)
-    ddtrace (0.40.0)
+    ddtrace (0.43.0)
       msgpack
     devise (4.7.1)
       bcrypt (~> 3.0)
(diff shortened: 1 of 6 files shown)


prabhanshuguptagit (approved):

rsanheim:
Worth a shot - mind posting the CHANGELOG from ddtrace? Was there any reference to better handling background jobs?

kitallis:
@rsanheim On second look, it seems like the [CHANGELOG](https://github.com/DataDog/dd-trace-rb/blob/master/CHANGELOG.md#0430---2020-11-18) suggest that an attempt to fix the issue was in `0.40.0` – can't find anything particularly interesting after that.

Merging this in anyway, because it doesn't hurt.

PR #1953, opened 2021-01-06
State reports

rsanheim:
Wires up basic state level region reports. 

For the index view: We are using recursion here so that when we decide to limit what regions are returned for other countries, things will just work.  This will allow things to work with both 'full level' reports (for
India) and countries that only want to display certain levels in the
hiearchy

Using https://github.com/github/view_component - docs are located at [https://viewcomponent.org](https://viewcomponent.org)

**Story card:** [ch2213](https://app.clubhouse.io/simpledotorg/story/2213/show-available-state-reports-in-index-view)


app/components/reports/region_tree_component.html.erb
```diff
@@ -0,0 +1,25 @@
+<ul class="access-item-wrapper">
+  <% children.each do |parent, children| %>
+  <li>
+    <div class="access-item collapsed <%= parent.region_type %>">
+      <% depth(parent).times do %>
```
harimohanraj89, line 5: Can we simply pass the depth into the recursive function call? That'll keep the view code _extremely_ agnostic of our region structure, right?

app/components/reports/region_tree_component.rb
```diff
@@ -0,0 +1,20 @@
...
+  def initialize(parent:, children:)
+    @children = children
+    @parent = parent
+  end
+
+  OFFSET = 2
+
+  def depth(region)
+    depth = Region::REGION_TYPES.index(region.region_type) - OFFSET
+    depth += 1 if region.child_region_type.nil?
```
harimohanraj89, line 13: Can you tell me what this +1 is for?

spec/components/reports/region_tree_component_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+  pending "add some examples to (or delete) #{__FILE__}"
+
+  # it "renders something useful" do
+  #   expect(
+  #     render_inline(described_class.new(attr: "value")) { "Hello, components!" }.css("p").to_html
+  #   ).to include(
+  #     "Hello, components!"
+  #   )
+  # end
+end
```
harimohanraj89, line 13: We can axe this if we don't plan to write view-y specs for this thing.

app/views/reports/regions/_children.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        <%= link_to_if(true, parent.name, reports_region_path(parent, report_scope: parent.region_type)) %>
+      </div>
+
+      <div class="access-item__padding"></div>
+    </div>
+
+    <%= render "children", children: children if children && children.any? %>
+  </li>
+  <% end %>
+</ul>
```
harimohanraj89, line 29: Is this no longer required, since we're using the view component?

harimohanraj89 (changes requested):
Some questions/suggestions from me!

app/components/reports/region_tree_component.rb
rsanheim: This is one reason we want this as a method and not as an arg passed in I think.

We want the depth to be based off the highest level region type, minus the offset, and then we need to add one for the case where we _don't_ display the dropdown icon ...which is the case where there aren't any child regions to show.

app/views/reports/regions/_children.html.erb
rsanheim: nope, deleting!

claudiovallejo (approved):

app/components/reports/region_tree_component.html.erb
rsanheim: We can, but I think I'd like to wait until I tackle the non India region hierarchy before changing this much more...as I think this will change depending on how I implement that.

app/components/reports/region_tree_component.rb
harimohanraj89: Ahhh I see, is that because the dropdown icon bumps the text over to the right? That sounds like a view concern that we can solve with appropriate styling, since conceptually, the depth of one item shouldn't change based on whether it has children or not.

In any case, we're not going to do depth-in-method-call as you mentioned above ☝️ , so I'll let you decide if you've got the motivation to figure out the CSS for that 😅  No big deal either ways 👍

app/controllers/reports/regions_controller.rb
```diff
@@ -14,15 +14,16 @@ def index
     if current_admin.feature_enabled?(:region_reports)
       accessible_facility_regions = authorize { current_admin.accessible_facility_regions(:view_reports) }
       cache_key = "#{current_admin.cache_key}/regions/index"
-      cache_version = accessible_facility_regions.cache_key
+      cache_version = "#{accessible_facility_regions.cache_key} / v2"
```
harimohanraj89, line 17: This change ensure we bust the index view cache when we deploy, right?

Shall we strip the spaces out of the cache version for idiomatic-ness?

app/controllers/reports/regions_controller.rb
```diff
@@ -14,15 +14,16 @@ def index
...
-          org, district, block = ancestors.values_at("organization", "district", "block")
+          org, state, district, block = ancestors.values_at("organization", "state", "district", "block")
           result[org] ||= {}
-          result[org][district] ||= {}
-          result[org][district][block] ||= []
-          result[org][district][block] << facility
+          result[org][state] ||= {}
+          result[org][state][district] ||= {}
+          result[org][state][district][block] ||= []
+          result[org][state][district][block] << facility
```
harimohanraj89, line 26: I love how natural this feels because of our regions.

app/models/user_access.rb
```diff
@@ -57,6 +57,22 @@ def accessible_organizations(action)
       .includes(facility_groups: :facilities)
   end
 
+  class UnsupportedAccessRequest < RuntimeError; end
+
+  # We determine a state's view_reports access via the district manager user access. Any other access requests for
+  # a state receive an error and are not supported.
```
harimohanraj89, line 63: Would it be a bit clearer to word the first half as: 

Users can view reports for a state if they can manage any district within the state.

app/models/user_access.rb
```diff
@@ -75,6 +91,10 @@ def accessible_district_regions(action)
     Region.district_regions.where(source_id: facility_group_ids)
   end
 
+  # We determine accessible blocks via the parent district authorization - so we need to jump through some hoops
+  # here given our current user access model.
```
harimohanraj89, line 95: Shall we put the actual auth rule in this comment while we're here? Eg.

A block is accessible to a user only if the parent district is also accessible.

app/controllers/reports/regions_controller.rb
```diff
@@ -195,6 +211,8 @@ def find_region
 
   def region_class
     @region_class ||= case report_params[:report_scope]
+    when "state"
```
harimohanraj89, line 214: I _think_ we don't use this magical `region_class` as a class anywhere 🤔  That means this method is doing some trivial mapping, plus an inclusion check.

Can we: 
1. trash this method
1. `switch` on the `report_scope` in `find_region`
1. Rely on the inclusion check there?

spec/models/user_access_spec.rb
```diff
@@ -160,6 +160,37 @@
         end
       end
 
+      context "#accessible_state_regions" do
+        it "managers of a district have view_reports access to the districts' states" do
+          facility_group_1 = create(:facility_group, name: "district 1")
+          facility_group_2 = create(:facility_group, name: "district 2")
+          pp facility_group_1.state_region
+          pp facility_group_2.state_region
```
harimohanraj89, line 168: Oh hai. We can remove these prints I think.

app/components/reports/region_tree_component.rb
rsanheim: > Ahhh I see, is that because the dropdown icon bumps the text over to the right? That sounds like a view concern that we can solve with appropriate styling, since conceptually, the depth of one item shouldn't change based on whether it has children or not.
> 
> In any case, we're not going to do depth-in-method-call as you mentioned above ☝️ , so I'll let you decide if you've got the motivation to figure out the CSS for that 😅 No big deal either ways 👍

Yup that is right.

I'm reusing all the CSS and DOM structure from the admin side, so changing the CSS just for this code path would be more trouble that its currently worth. Certainly worth revisiting though.

app/controllers/reports/regions_controller.rb
rsanheim: Yup, and yup. :grin

app/controllers/reports/regions_controller.rb
rsanheim: yup, I'm going to do this in a followup PR because I have a slightly larger refactoring in mind here.

PR #1958, opened 2021-01-08
Add drug stock fields to protocol drugs

vkrmis:
**Story card:** [ch2220](https://app.clubhouse.io/simpledotorg/story/2220/add-drug-category-and-stock-tracked-to-protocoldrug-data-model-and-ui)
[PRD](https://docs.google.com/document/d/13bPC3CjWWbaNV2D_hVD3Flu7A4NUWB5DdNVo-MX8i6Y/edit#heading=h.2tfhwc7ycgb)

## Because

We need to enrich the protocol drugs data model with some new fields, in preparation for Drug Stock work 

## This addresses

Adds `drug_category` and `stock_tracked` to `ProtocolDrug`s 

Screenshot 2021-01-11 at 7 29 58 PM


app/models/protocol_drug.rb
```diff
@@ -5,6 +5,15 @@ class ProtocolDrug < ApplicationRecord
   validates :name, presence: true
   validates :dosage, presence: true
 
+  enum drug_categories: {
```
harimohanraj89, line 8: `drug_category` singular? Does plural work too?

app/views/admin/protocol_drugs/_form.html.erb
```diff
@@ -1,9 +1,15 @@
...
-      <%= form.text_field :name, id: :protocol_drug_name %>
-      <%= form.text_field :dosage, id: :protocol_drug_dosage %>
-      <%= form.text_field :rxnorm_code, id: :protocol_drug_rxnorm_code, label: "RxNorm code" %>
-
-      <%= form.primary %>
-    <% end %>
-</div>
\ No newline at end of file
+  <%= bootstrap_form_with(model: [:admin, @protocol, @protocol_drug], local: true, label_errors: true) do |form| %>
+    <%= form.text_field :name, id: :protocol_drug_name %>
```
harimohanraj89, line 3: Shall we make the other design changes to this form while we're here? Namely:
* Name, dosage, and Rxnorm on a single line
* "Type" under that
* "Track" under that

harimohanraj89 (changes requested):
Couple thoughts.

app/models/protocol_drug.rb
vkrmis: Shouldn't, fixing.

vkrmis: Plural accidentally worked for what I was using it for 🤦 
`ProtocolDrug.drug_categories`

app/controllers/admin/protocol_drugs_controller.rb
```diff
@@ -56,7 +56,9 @@ def protocol_drug_params
     params.require(:protocol_drug).permit(
       :name,
       :dosage,
-      :rxnorm_code
+      :rxnorm_code,
+      :drug_category,
+      :stock_tracked
```
ssrihari, line 61: From the PRD, it seems like we want this to restrict what is seen on the dashboard. Does it also restrict the drugs that can be tracked? 

If we don't want to restrict the data-entry, maybe we should rename the field to reflect viewing, or reporting, something on those lines. `stock_tracked_on_dashboard`?

ssrihari: Vikram explained over a conversation. Seems like we want to keep the two concerns of viewing-on-dashboard, and collecting-data as the same thing for now.

We can revisit splitting these two concerns later, but that would involve a data migration. 

cc @harimohanraj89 @kitallis

harimohanraj89: Sounds good 👍

harimohanraj89 (approved):
Coooolbeans.

PR #1963, opened 2021-01-11
Update transferred patient status from call results

prabhanshuguptagit:
**Story card:** [ch2234](https://app.clubhouse.io/simpledotorg/story/2234/data-script-to-mark-patients-as-migrated-based-on-recent-appointment)

## Because

We want to exclude patients who have migrated away from Simple while reporting. This information currently resides as the call result's `cancel_reason` in `Appointment`. This is set from the:
- overdue list on dashboard
- overdue list on the app

## This addresses

This adds a data migration to copy over the migration information to `Patient`, by setting the `status` field to `migrated`. This also makes sure that any future call result updates (both via dashboard and the app) will update the Patient's status.


lib/tasks/scripts/mark_transferred_patients.rb
```diff
@@ -0,0 +1,18 @@
...
+        .select("DISTINCT ON (patient_id) *")
+        .order("patient_id, updated_at DESC")
+
+    migrated_patient_ids =
+      Appointment
+        .from(latest_appointment_per_patient, "appointments")
+        .where(cancel_reason: TRANSFER_CANCEL_REASONS)
+        .pluck(:patient_id)
+
+    Patient.where(id: migrated_patient_ids).update_all(status: "migrated")
```
rsanheim, line 16: Would we want to update a patient if they are marked as `dead`?

Also, can you output the number of updated records to the log so that we can monitor how much this updates?

prabhanshuguptagit: Dead patients are already being updated https://github.com/simpledotorg/simple-server/blob/d87efa16e2803e1b63d34de8d7fa8806a59fb214/app/controllers/appointments_controller.rb#L76

rsanheim: > Would we want to update a patient if they are marked as `dead`?

Sorry, I meant if a patient's existing status is marked as something non-active (say, dead), would we want to overwrite that w/ `migrated` ?

prabhanshuguptagit: Hmm, that should be unlikely since patients get removed from the overdue list once the appointment status is updated. Both on the app and the dashboard.

prabhanshuguptagit: We can still override the status I think, in case someone does manage to update them again. Shouldn't do any harm.

lib/tasks/scripts/mark_transferred_patients.rb
```diff
@@ -0,0 +1,21 @@
...
+    migrated_patient_ids =
+      Appointment
+        .from(latest_appointment_per_patient, "appointments")
+        .where(cancel_reason: TRANSFER_CANCEL_REASONS)
+        .pluck(:patient_id)
+
+    patients = Patient.where(id: migrated_patient_ids)
+    Rails.logger.info "Status set to 'migrated' for #{patients.count}"
+
+    patients.update_all(status: "migrated")
```
kitallis, line 19: This won't resync the patients, intentional?

db/data/20210111092259_set_transferred_patient_status.rb
```diff
@@ -0,0 +1,9 @@
+class SetTransferredPatientStatus < ActiveRecord::Migration[5.2]
+  def up
+    MarkTransferredPatient.call
```
kitallis, line 3: Note: measure timings on perf environment before running on prod.

kitallis (approved):
This lgtm logically.

Non-blocking question, to take a stock of any incidental complexities in our API:

* We are updating Patient status using the cancel reason.
* App also sends the Patient status using the Patient API, but only for `dead` patients.
* App does not mark other Patient statuses'.

To keep things consistent, should we also have the app update the patient status in addition to doing this through the API, because otherwise there's just too many cross-service nuances to keep a track of.

prabhanshuguptagit:
> Should we also have the app update the patient status in addition to doing this through the API

Yep, we decided on doing this in both places. Android folks will pick this up in the next iteration.

lib/tasks/scripts/mark_transferred_patients.rb
prabhanshuguptagit: Oops, unintentional. Thanks.

PR #1964, opened 2021-01-11
Add the DrugStock model

vkrmis:
**Story card:** [ch2221](https://app.clubhouse.io/simpledotorg/story/2221/add-the-drugstock-data-model)

## Because

We want to start tracking and recording the available stock of protocol drugs at facilities 

## This addresses

Adds the DrugStock: 
- model
- model specs
- factory


app/models/drug_stock.rb
```diff
@@ -0,0 +1,9 @@
+class DrugStock < ApplicationRecord
+  belongs_to :facility
+  belongs_to :user
+  belongs_to :protocol_drug
+
+  validates :in_stock, numericality: true, presence: true
+  validates :received, numericality: true, presence: true
```
harimohanraj89, line 7: Can we sanity check whether `received` can be empty? I think it might be possible. Only `in_stock` is required for the basic reporting we want to do. `received` lets us calculate consumption later, so I reckon people might wanna leave it blank to start, at least sometimes.

db/schema.rb
```diff
@@ -151,6 +151,21 @@
     t.index ["version"], name: "unique_data_migrations", unique: true
   end
 
+  create_table "drug_stocks", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "facility_id", null: false
+    t.uuid "user_id", null: false
+    t.uuid "protocol_drug_id", null: false
+    t.integer "in_stock", null: false
+    t.integer "received", null: false
```
harimohanraj89, line 159: Ditto to above, should this be nullable?

app/models/protocol.rb
```diff
@@ -2,6 +2,7 @@ class Protocol < ApplicationRecord
   has_many :protocol_drugs, -> { order(:updated_at) }
 
   has_many :facility_groups
+  has_many :drug_stocks
```
harimohanraj89, line 5: This should be on the protocol _drug_, right?

spec/factories/drug_stocks.rb
```diff
@@ -0,0 +1,11 @@
+FactoryBot.define do
+  factory :drug_stock do
+    id { SecureRandom.uuid }
+    association :facility, strategy: :create
+    association :user, strategy: :create
+    association :protocol_drug, strategy: :create
+    recorded_at { Time.current }
+    in_stock { 500 }
+    received { 500 }
```
harimohanraj89, line 9: Let's make these numbers different for better entropy in our tests. Might as well make the `received` number significantly less than `in_stock` as well while we're at it, for conceptual accuracy (~100)

harimohanraj89 (changes requested):
A few  questions and comments.

app/models/protocol.rb
vkrmis: Oh, whoops

app/models/drug_stock.rb
vkrmis: Shouldn't they enter 0 then?

vkrmis: I've made the change for now, can add the check if we decide to later

harimohanraj89: Coolbeans. Yeah `0` and `null` should be treated differently. `0` means "nothing was received", `null` means "I don't know".

app/models/drug_stock.rb
```diff
@@ -0,0 +1,9 @@
+class DrugStock < ApplicationRecord
+  belongs_to :facility
+  belongs_to :user
+  belongs_to :protocol_drug
+
+  validates :in_stock, numericality: true, presence: true
+  validates :received, numericality: true
+  validates :recorded_at, presence: true
```
ssrihari, line 8: Should this go into `ApplicationRecord`?

ssrihari: Resolved after talking to @vkrmis. Not all application records are "recorded", like `Facility`.

harimohanraj89 (approved):
Godspeed :shipit:

PR #1968, opened 2021-01-11
Fix admin deletion again

harimohanraj89:
**Story card:** [ch2267](https://app.clubhouse.io/simpledotorg/story/2267)

## Because

In #1110, we ensured that admins can be deleted and reinvited with the same email address.

Then, in #1414, we introduced a regression because we _discarded_ admins instead of _destroying_ them.

Here, we add similar data cleanup hooks to `after_discard`. We also address the root cause of why admins need to be deleted most of the time - to re-trigger an invitation.

## This addresses

* Ensuring again that admins can be deleted and invited again with the same email address
* Adding a "Resend invite" link (screenshots below)

### The "Resend invite" link

Screen Shot 2021-01-12 at 12 21 57 AM

### The success messaging

Screen Shot 2021-01-12 at 12 22 02 AM

### The failure messaging (if user claimed their invitation by the time you re-trigger

Screen Shot 2021-01-12 at 12 22 33 AM


app/controllers/admins_controller.rb
```diff
@@ -55,6 +55,23 @@ def update
     redirect_to admins_url, notice: "Admin was successfully updated."
   end
 
+  def resend_invitation
```
kitallis, line 58: Should we write a controller spec for this?

It'll be a little awkward to assert since everything is just a redirect with a `notice` or `alert` without any real status codes but at least the branching will get tested over?

kitallis (approved):
I'm approving to unblock, just left a comment about a potential controller spec.

PR #1943, opened 2021-01-05
Drug stock design mockup

harimohanraj89:
**Story card:** [Drug stock](https://app.clubhouse.io/simpledotorg/epic/1338/drug-stock)

PR #1975, opened 2021-01-13
Add form to track drug stock for a given facility and month

vkrmis:
**Story card:** [ch2222](https://app.clubhouse.io/simpledotorg/story/2222/add-drug-stock-form-to-dashboard-behind-a-feature-flag)
[PRD](https://docs.google.com/document/d/13bPC3CjWWbaNV2D_hVD3Flu7A4NUWB5DdNVo-MX8i6Y/edit#)

## Because

We need to add a way for Dashboard users to submit drug stock information

## This addresses

- Copy over the view html from https://github.com/simpledotorg/simple-server/pull/1943
- Only show facilities in the reports page when the protocol drug is enabled for tracking stock
- Render form in a separate page (same page modal is complex, and not performant)
- Pre-fill form with latest drug stocks for the chosen month
- Refresh page on changing the month via the month picker
- Add a datepicker widget so that the month picker is usable across browsers
- Handle the form request on the controller#create, create multiple drug stock entries in a transaction
- Add necessary validations, and handle empty form submission
- Redirect back to reports page with the previously chosen filters
- Highlight drug stock in navigation when active
- Feature flag the drug stock link and controller actions
- Rename recorded_at to for_end_of_month
- Access protocol drugs through facilities
- Add auth checks for new and create actions

## Quick demo
https://user-images.githubusercontent.com/1003238/105482921-18dc2b80-5ccf-11eb-98ed-94804d589358.mp4


app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+                    .includes(facility_group: :protocol_drugs)
+                    .where(protocol_drugs: {stock_tracked: true})
+  end
+
+  def new
+    render partial: "form"
+  end
+
+  def create
+    drug_stocks = DrugStock.transaction do
```
kitallis, line 23: Do we need a transaction here? If the create fails, it'll be silent anyway I believe, so there'd be nothing to rollback, it'll continue creating the others.

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  def authorize_my_facilities
+    authorize { current_admin.accessible_facilities(:view_reports).any? }
+  end
+
+  def set_facility
+    @facility = current_admin.accessible_facilities(:view_reports).find_by_id(params[:facility_id])
+  end
+
+  def drug_stocks_params
+    params.permit(:recorded_at, drug_stocks: [:received, :in_stock, :protocol_drug_id])
```
kitallis, line 55: Want to line these up vertically, so that diffs in the future show up clearly?

app/assets/javascripts/common/drug_stocks.js
```diff
@@ -0,0 +1,21 @@
...
+        let in_stock = protocol_drug.querySelector('.in_stock');
+        if(received.value.trim() != '') {
+            in_stock.required = true;
+        } else {
+            in_stock.required = false;
+        }
+    });
+};
+
+function validateAndCloseDrugStockForm(facility_id) {
```
kitallis, line 16: Non-blocking suggestion:

Might be nice to class-ify this JS if possible. We've been trying to move to a little bit of prototypical JS lately. It doesn't entirely apply in this scenario because the function acts as a closure that doesn't arbitrarily get called, but namespacing will ensure that for sure.

It could be something like `new DrugStock().markRequired()` or something.

[ref](https://github.com/simpledotorg/simple-server/blob/51be010af90cd5de81da56563456026ebaa52ff3/app/assets/javascripts/common/admins.js#L200), [ref](https://github.com/simpledotorg/simple-server/blob/51be010af90cd5de81da56563456026ebaa52ff3/app/assets/javascripts/common/user_teleconsult_search.js).

app/views/my_facilities/drug_stocks/_form.html.erb
```diff
@@ -0,0 +1,54 @@
+<div class="modal fade" id=<%= "drug_report_modal_#{facility.id}" %> tabindex="-1" role="dialog" aria-labelledby="Drug Report" aria-hidden="true">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content" style="text-align: left">
+      <div class="modal-header">
+        <h2 class="modal-title">Drug stock report</h2>
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true"><i class="fas fa-times"></i></span>
+        </button>
+      </div>
+      <%= bootstrap_form_with(url: "/my_facilities/drug_stocks",
```
kitallis, line 10: Want to use rails' bidi paths here?

kitallis (commented):
Looking good! Left some comments inline. Let me know once this is ready for full review and I'll make another pass at it.

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -0,0 +1,26 @@
+class MyFacilities::DrugStocksController < AdminController
+  include Pagination
+  include MyFacilitiesFiltering
+
+  layout "my_facilities"
+
+  before_action :authorize_my_facilities
+  after_action :verify_authorization_attempted
+
+  DRUG_CATECORIES = []
```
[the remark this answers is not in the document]
vkrmis: This is unused, gone

app/controllers/my_facilities/drug_stocks_controller.rb
vkrmis: We're showing a notice/alert based on the whole payload succeeding or failing. Better than failing silently.

app/assets/javascripts/common/drug_stocks.js
vkrmis: This js is gone now

app/controllers/my_facilities/drug_stocks_controller.rb
ssrihari: I actually wanted the transaction semantics. I'll change this to `create!`, ya?

app/views/my_facilities/drug_stocks/_form.html.erb
ssrihari: This file is gone. The change has been made in `new.html.erb` now.

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+    end
+
+    report_url_with_filters = session[:report_url_with_filters]
+    session[:report_url_with_filters] = nil
+    if drug_stocks.empty?
+      redirect_to report_url_with_filters
+    elsif drug_stocks.all?(&:valid?)
+      redirect_to report_url_with_filters, notice: "Saved drug stocks"
+    elsif drug_stocks.any?(&:invalid?)
```
vkrmis, line 45: `create!` fails immediately on invalid inputs, so we return a `400 Bad Request` instead of the response with this message. Need to keep `create` or catch the exception.

spec/controllers/my_facilities/drug_stocks_controller_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+  describe "GET #index" do
+    context "as power_user" do
+      it "returns a success response" do
+        sign_in(power_user.email_authentication)
+
+        get :index, params: {}
+        expect(response).to be_successful
+      end
+
+      it "only include facilities with tracked protocol drugs" do
```
vkrmis, line 38: flaky

app/controllers/my_facilities/drug_stocks_controller.rb
ssrihari: Ah yes, missed handling tx failure. Handled now in 94ab4cd4fbf6ef4693db28c7858306aecdfcbe37.

spec/controllers/my_facilities/drug_stocks_controller_spec.rb
ssrihari: Strange, I don't see anything in the test that makes this flaky.

ssrihari:
> Looking good! Left some comments inline. Let me know once this is ready for full review and I'll make another pass at it.

@kitallis ready

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+      drug_stock[:in_stock].blank? && drug_stock[:received].blank?
+    end
+  end
+
+  def authorize_my_facilities
+    authorize { current_admin.accessible_facilities(:view_reports).any? }
+  end
+
+  def set_facility
+    @facility = authorize { current_admin.accessible_facilities(:manage).find_by_id(params[:facility_id]) }
```
kitallis, line 66: I think it's generally a safer pattern to do `find_by!(id: <>)` within `authorize` blocks because it'll naturally raise a `ActiveRecord::RecordNotFound`.

Returning `falsey` is also supported, but I prefer throwing an exception in these auth scenarios.

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+    @drug_stocks = drug_stock_list.each_with_object({}) { |drug_stock, acc|
+      acc[drug_stock.protocol_drug.id] = drug_stock
+    }
+  end
+
+  def create
+    report_url_with_filters = session[:report_url_with_filters]
+    session[:report_url_with_filters] = nil
+
+    drug_stocks = DrugStock.transaction do
```
kitallis, line 31: I think this is fine, but the combination of both starting a transaction and a per record validity check feels like a lot of work in the controller.

Do you want to try moving this persistence stuff into the model, add a function that bulk saves stuff?

kitallis (approved):
LGTM. Left a couple of comments, approved to unblock.

app/controllers/my_facilities/drug_stocks_controller.rb
ssrihari: Thanks @kitallis, we'll fix this in https://github.com/simpledotorg/simple-server/pull/2012.

app/controllers/my_facilities/drug_stocks_controller.rb
ssrihari: We'll also look into this with https://github.com/simpledotorg/simple-server/pull/2012.

PR #1977, opened 2021-01-13
Exclude dead patients in reports

prabhanshuguptagit:
**Story card:** [ch2227](https://app.clubhouse.io/simpledotorg/story/2227/exclude-dead-patients-from-all-necessary-queries)

## Because

We want to exclude dead patients from our reports.

## This addresses

These excludes the patients from following:
- [x] Region reports (except registrations)
- [X] My Facilities reports (except registrations)
- [X] Patient line list. This also adds a column to display the patient status (active, migrated)

The changes are flippered behind a `report_with_exclusions` flag. This will be enabled once all the exclusions work has gone in (LTFU, new chart etc.). This does not edit the chart footnotes yet, it will be updated in one of the subsequent PRs.


app/services/no_bp_measure_service.rb
```diff
@@ -58,6 +61,7 @@ def execute_sql(period)
         AND "medical_histories"."deleted_at" IS NULL
         AND "medical_histories"."hypertension" = :hypertension
         AND "patients"."assigned_facility_id" in :facilities
+        AND "patients"."status" != :exclude_status
```
prabhanshuguptagit, line 64: @rsanheim It seems like github-ds doesn't support Postgres yet. I was hoping to do a `NOT IN ('dead', 'migrated')`. Might try writing this in Activerecord, if that works out we'll be able to use the scope as well.

prabhanshuguptagit: I tried interpolating the `('dead', 'migrated')` list like
```
AND "patients"."status" NOT IN :excluded_statuses
```
That didn't work out because it inserts single quotes around the brackets.

app/services/control_rate_service.rb
```diff
@@ -60,11 +63,25 @@ def fetch_all_data
     results
   end
 
+  def registration_counts_with_exclusions
+    return @registration_counts_with_exclusions if defined? @registration_counts_with_exclusions
+    formatter = lambda { |v| quarterly_report? ? Period.quarter(v) : Period.month(v) }
+
+    @registration_counts_with_exclusions =
+      region.assigned_patients
+        .for_reports(with_exclusions: with_exclusions)
```
rsanheim, line 72: If `with_exclusions` is false, this will duplicate what we are already storing in registrations, right?

app/services/no_bp_measure_service.rb
```diff
@@ -34,50 +37,42 @@ def cache_keys_for_period
...
+                OR prescription_drugs.id IS NOT NULL
+                OR blood_sugars.id IS NOT NULL")
+      .where("NOT EXISTS
+                 (SELECT 1
+                  FROM blood_pressures bps
+                  WHERE patients.id = bps.patient_id
+                  AND bps.recorded_at > ?
+                  AND bps.recorded_at <= ?)", start_date, end_date)
+      .distinct("patients.id")
+      .count
```
rsanheim, line 67: This is going to be harder to read and harder to maintain than the SQL version. I'll take a shot at keeping it in GitHub SQL...I'm pretty sure it can handle IN clauses.

rsanheim: pushed up a couple commits -- I think the only thing we needed here was parens around the in clause because `GitHub ::SQL` doesn't add that automagically.

prabhanshuguptagit: The primary reason I tried moving this to AR was to support the `for_reports` scope. I wanted to avoid defining the exclusions and doing that `if with_exclusions` here. That makes it harder to maintain when the exclusions change, and adding LTFU will be trickier too. This query is fairly straightforward with a bunch of joins and wheres stacked together, no special PG functions to drop into SQL for.

Separately, it feels like there’s way too many ways in the codebase to query data from the database (AR, raw SQL,  Github SQL). It’s not clear which one is good for what. We should ideally try to stick to one interface for consistency and simplicity. I prefer sticking to AR as much as I can, because it's easy to compose with the rest of the query codebase which is heavily scope reliant. wdyt?

app/services/control_rate_service.rb
prabhanshuguptagit: This applies the with_hypertension clause in that case, which I think we should've been adding already to the denominator.

app/services/no_bp_measure_service.rb
prabhanshuguptagit: @rsanheim As I was spiking out LTFU exclusions, I'm finding the scope makes the query much smaller and easy to port. I'll revert back to the AR query for now.

rsanheim: > It’s not clear which one is good for what. We should ideally try to stick to one interface for consistency and simplicity. I prefer sticking to AR as much as I can, because it's easy to compose with the rest of the query codebase which is heavily scope reliant. wdyt?

I think AR is great for simple queries where you are just joining scopes together and grabbing a simple, straight forward set of records.

SQL is better when you need to really get deep into the guts of the query - so if there are subselects, CTE's, or UNIONs, and especially if the query is one that needs to be tuned for performance.  SQL is also much easier to iterate on between a local Postgres session and the app, because you can just copy / paste the query between your local database tool and the code.  Trying to do this with logged AR queries is a pain and makes it hard to really tweak the SQL when necessary.

The other thing to remember here is that these queries could really benefit from further tuning -- we are currently rerunning queries for many periods when we could instead run **one** query to get all time periods - see [this answer](https://stackoverflow.com/a/14877465) on Stack Overflow about using CTE along with `generate_series` to grab data for all periods in one query. We could be doing this for all our reports, but we would want to be in plain SQL to do it in any sort of reasonable way.

rsanheim (approved):
👍 approving to unblock progress -- we can always iterate more on the SQL / AR differences later as all of these things evolve

PR #1983, opened 2021-01-14
My Facilities Improvements: District filters

claudiovallejo:
**Story card:** [ch2228](https://app.clubhouse.io/simpledotorg/story/2228/add-district-filter)

## Because

## This addresses

**Default view**
image

**Empty state**
image

* The first district is selected by default
* The district dropdown displays all districts a user has access to (sorted alphabetically)
* The blocks dropdown shows all blocks in the selected district (sorted alphabetically)
* The facility sizes dropdown shows facility sizes for facilities in the selected district (sorted from largest to smallest)
* Selecting a new district sets the block and facility size filters to "All" to avoid selecting a block/facility size that doesn't exist


app/controllers/concerns/my_facilities_filtering.rb
```diff
@@ -5,38 +5,49 @@ module MyFacilitiesFiltering
...
+    end
+
     def populate_zones
-      @zones = current_admin.accessible_facilities(:view_reports).pluck(:zone).uniq.compact.sort
+      @zones = @facilities.pluck(:zone).uniq.compact.sort
     end
 
-    def set_selected_sizes
+    def set_selected_facility_group
+      @selected_facility_group = params[:facility_group] ? @facility_groups.find_by(slug: params[:facility_group]) : @facility_groups.first
```
harimohanraj89, line 40: With respect to ordering mindfully - is there a better default we can choose than `facility_groups.first`?

app/controllers/concerns/my_facilities_filtering.rb
```diff
@@ -5,38 +5,49 @@ module MyFacilitiesFiltering
...
+
+    def filter_facilities
+      filtered_facilities = facilities_by_size(@facilities)
       facilities_by_zone(filtered_facilities)
     end
 
     private
 
+    def populate_facility_groups
+      @facility_groups = current_admin.accessible_facility_groups(:view_reports).order(:name)
```
harimohanraj89, line 24: `accessible_facility_groups` will give you the list of facility groups that a user has _full_ access to. In other words, the user can access _all_ facilities within that facility group.

However, the filters need to expose facility groups that the user has _any_ access to. In other words, facility groups where the user has _at least one_ facility access.

For example, if I have reports access to _one_ facility in District A and _one_ facility in District B, I expect to be able to select District A or District B from this dropdown. And in both cases, I expect to see data for only _one_ facility inside. Currently, when I tried to set up this permission for a test user, the page breaks, as `@facility_groups` becomes an empty array.

app/controllers/concerns/my_facilities_filtering.rb
claudiovallejo: That's a good question @harimohanraj89 — the facility groups are sorted by alphabetical order and I think the first district in that list makes most sense since the list only include facility groups the admin has access to. Do you have an idea for a better default?

app/controllers/concerns/my_facilities_filtering.rb
claudiovallejo: Ah thanks for pointing this out @harimohanraj89! The example you mention is right — My Facilities should show all facilities an admin has access to and not just all facilities within districts the admin has full access to.

What do I need to change from the `populate_facility_groups` method to make sure it works as your example?

app/controllers/concerns/my_facilities_filtering.rb
harimohanraj89: Nope, I don't 😅  I also kinda missed the fact that these will be alphabetically sorted. Sounds good to me!

timcheadle (approved):
Solid work!

PR #1984, opened 2021-01-14
Reports cache improvements

rsanheim:
This could be resulting in cache misses if the cache warmer fails, or if
dates are mismatched, or if some else strange is afoot.

Since we are busting the cache ourselves every night, we can have a more
general cache_key here.

no story


app/services/control_rate_service.rb
```diff
@@ -115,7 +109,7 @@ def quarterly_report?
   end
 
   def cache_key
-    "#{self.class}/#{region.cache_key}/#{report_range.begin.type}/#{Date.current}"
+    "#{self.class}/#{region.cache_key}/#{@periods.end.cache_key}"
```
rsanheim, line 112: This will stop any cache misses based on date mismatches.

spec/services/top_region_service_spec.rb
```diff
@@ -1,123 +0,0 @@
-require "rails_helper"
-
-RSpec.describe TopRegionService, type: :model do
```
rsanheim, line 3: This is all dead code now.

rsanheim (commented):
Notes for @simpledotorg/server-developers

PR #1993, opened 2021-01-20
Seed generation fixes

harimohanraj89:
**Story card:** [ch2325](https://app.clubhouse.io/simpledotorg/story/2325/generate-a-fresh-large-seed-data-set-in-sandbox)

* [x] Add "demo" as a recognized seed env
* [x] Set correct patient statuses

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -13,6 +13,8 @@ class Patient < ApplicationRecord
   ANONYMIZED_DATA_FIELDS = %w[id created_at registration_date registration_facility_name user_id age gender]
   DELETED_REASONS = %w[duplicate unknown accidental_registration].freeze
 
+  enum status: STATUSES.zip(STATUSES).to_h, _prefix: true
+
   enum reminder_consent: {
     granted: "granted",
     denied: "denied"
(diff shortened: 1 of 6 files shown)


app/models/patient.rb
```diff
@@ -18,6 +13,23 @@ class Patient < ApplicationRecord
   ANONYMIZED_DATA_FIELDS = %w[id created_at registration_date registration_facility_name user_id age gender]
   DELETED_REASONS = %w[duplicate unknown accidental_registration].freeze
 
+  enum status: STATUSES.zip(STATUSES).to_h, _prefix: true
```
kitallis, line 16: Hm, should we not have a prefix? Then we can continue to do `Patient#status`.

kitallis (approved):
Looking good, I just left a note about the `status` prefix.

app/models/patient.rb
harimohanraj89: The `_prefix` actually applies to the scope and inquiry methods on the patient. ie,

```ruby
# prefix
Patient.status_active
patient.status_active?

# no prefix
Patient.active
patient.active?
```

I was unable to avoid the prefix because the `could_not_contact_reasons` enum is not prefixed, and has a conflicting enum value "dead". I think prefixing is a good practice in general to avoid these clashes, but didn't want to modify the existing `could_not_contact_reasons` here because I'm unsure of the scope of its usage.

kitallis: Ah, sorry, my bad. Makes sense.

PR #1994, opened 2021-01-20
Search refinements

rsanheim:
**Story card:** [ch1989](https://app.clubhouse.io/simpledotorg/story/1989/ability-to-search-for-region-by-name)
## Because

Follow up fixes for search:

* better handling of empty / invalid search requests
* don't show orgs in the results, they aren't useful
* better event binding to capture "enter" key
* better placeholder text

### DIFF
--- a/app/assets/javascripts/common/regions_search.js
+++ b/app/assets/javascripts/common/regions_search.js
@@ -54,7 +54,7 @@ RegionsSearch = function () {
       this.clearDropDown()
       return false
     }
-    if (searchQuery.length) {
+    if (searchQuery && searchQuery.length) {
       this.showSpinner();
       $.ajax({
         url: e.form.action,
--- a/app/assets/javascripts/common/regions_search.js
+++ b/app/assets/javascripts/common/regions_search.js
@@ -50,7 +50,7 @@ RegionsSearch = function () {
   this.searchRequest = (e) => {
     let searchQuery = e.value;
 
-    if (searchQuery.length) {
+    if (searchQuery && searchQuery.length) {
       this.showSpinner();
       $.ajax({
         url: e.form.action,
@@ -61,6 +61,9 @@ RegionsSearch = function () {
           this.populateSearchResults(searchQuery, res)
         }
       })
+    } else {
+      console.log("invalid or empty search query")
+      return false
     }
   }
 
--- a/app/assets/javascripts/common/regions_search.js
+++ b/app/assets/javascripts/common/regions_search.js
@@ -40,17 +40,21 @@ RegionsSearch = function () {
       })
       this.populateDropdown(html);
     } else {
-      let html = $("template#no-results-found").html();
-      let $html = $(html);
-      $html.find(".search-query").html(searchQuery);
-      this.populateDropdown($html)
+      this.populateDropdown(this.noResultsFound(searchQuery))
     }
+  }
+
+  this.clearDropDown = () => {
+    $(".typeahead-dropdown").empty();
   }
 
   this.searchRequest = (e) => {
     let searchQuery = e.value;
-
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/common/regions_search.js
```diff
@@ -40,17 +40,21 @@ RegionsSearch = function () {
...
 
+  this.clearDropDown = () => {
+    $(".typeahead-dropdown").empty();
+  }
+
   this.searchRequest = (e) => {
     let searchQuery = e.value;
-
-    if (searchQuery.length) {
+    if (searchQuery === "") {
```
rsanheim, line 53: Need to handle this case specifically for when a user uses backspace to clear out a query and start over.

app/assets/javascripts/common/regions_search.js
```diff
@@ -61,6 +65,9 @@ RegionsSearch = function () {
           this.populateSearchResults(searchQuery, res)
         }
       })
+    } else {
+      console.log("invalid or empty search query")
```
rsanheim, line 69: Just logging for now, as I haven' seen this situation come up any more for now.

app/controllers/regions_search_controller.rb
```diff
@@ -1,27 +1,27 @@
...
-        result[org] ||= {}
-        result[org][state] ||= {}
-        result[org][state][district] ||= {}
-        result[org][state][district][block] ||= []
-        result[org][state][district][block] << facility
+        state, district, block = ancestors.values_at("state", "district", "block")
+        result[state] ||= {}
+        result[state][district] ||= {}
+        result[state][district][block] ||= []
+        result[state][district][block] << facility
```
rsanheim, line 16: No more orgs in the results.

app/controllers/regions_search_controller.rb
```diff
@@ -1,27 +1,27 @@
...
       }
     }
-    @query = params.permit(:query)[:query]
+    @query = params.permit(:query)[:query] || ""
     regex = /.*#{Regexp.escape(@query)}.*/i
     results = search(@accessible_regions, regex)
     json = results.sort_by(&:name).map { |region|
       {
-        ancestors: region.cached_ancestors.where.not(region_type: "root").order(:path).map { |a| a.name }.join(" > "),
+        ancestors: region.cached_ancestors.where.not(region_type: ["root", "organization"]).order(:path).map { |a| a.name }.join(" > "),
```
rsanheim, line 24: No orgs in the ancestor chain.

app/views/reports/regions/index.html.erb
```diff
@@ -22,9 +23,22 @@
...
+          const form = $("#search-form")
+          const input = form.find("#search_input")
+          form.on("submit", function(event) {
+            search(input);
+            return false;
+          })
+          form.on("input", function(event) {
+            search(input);
+            return false;
+          })
```
rsanheim, line 36: Handle both the submit and the input event, which captures any change in the search field as well as just hitting "enter" on the form itself.

app/views/reports/regions/_search.html.erb
```diff
@@ -10,7 +10,7 @@
                          id: :search_query,
                          class: "typeahead-input form-control",
                          hide_label: true,
-                         placeholder: "Find regions",
+                         placeholder: "Search for a district, block, or facility",
```
claudiovallejo, line 13: @rsanheim can we use "Search for a region or facility name" as placeholder text? Users that only have access to blocks or facilities won't be able to search by district. Using "regions" instead of specifying regions allows us to use a single placeholder text for all dashboard roles.

rsanheim: Sure 👍

PR #1995, opened 2021-01-20
Search authorization / extract RegionAccess

rsanheim:
A user should only see regions in their search results that they actually have access to.

Need to do some refactoring for this to consolidate this behavior and prep for future improvements.

**Story card:** [ch1989](https://app.clubhouse.io/simpledotorg/story/1989/ability-to-search-for-region-by-name)


app/models/region_access.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def accessible_district?(region, action)
+    return true if user.power_user?
+    accessible_region_ids(region.region_type, action).include?(region.id)
+  end
+
+  def accessible_block?(region)
+    return true if user.power_user?
+    @accessible_block_ids ||= user.accessible_block_regions(:view_reports).pluck(:id)
```
prabhanshuguptagit, line 40: Why is the interfacing different here from say in accessible_state?

prabhanshuguptagit (commented):
This is probably unrelated to the scope of this PR but I think there is some simplicity and performance to be gained in `UserAccess` if we do some smart joins. lmk if you want to pair on it sometime.

rsanheim:
> This is probably unrelated to the scope of this PR but I think there is some simplicity and performance to be gained in `UserAccess` if we do some smart joins. lmk if you want to pair on it sometime.

@prabhanshuguptagit You may be interested in [this epic](https://app.clubhouse.io/simpledotorg/epic/2217/migrate-user-access-to-regions), which will make our UserAccess much more cohesive and performant.

PR #1998, opened 2021-01-21
Guide: migrate patient data for user from one Facility to another

kitallis:
**Story card:** [ch2361](https://app.clubhouse.io/simpledotorg/story/2361/user-registered-in-the-wrong-facility)

## Because

Related to this [request](https://app.clubhouse.io/simpledotorg/story/2361/user-registered-in-the-wrong-facility) from the field.

## This addresses

Add a how-to guide under `wiki/` to deal with this in the future.


harimohanraj89 (approved):
Love it.

doc/wiki/move-patient-data-for-a-user.md
```diff
@@ -0,0 +1,33 @@
...
+
+a. The user has data locally, but cannot sync because the facility is soft-deleted (point 1. from above).  
+b. The user has data locally, can sync, but is syncing to the wrong facility (point 2. from above).
+
+Fixing b. can be automated by following these steps:
+
+* Move user data from source facility to destination facility
+
+```bash
+bundle exec cap india:production deploy:rake task=data_fixes:move_user_data_from_source_to_destination_facility[<user-id>,<source-facility-id>,<destination-facility-id]
```
vkrmis, line 20:
```suggestion
bundle exec cap india:production deploy:rake task=data_fixes:move_user_data_from_source_to_destination_facility[<user-id>,<source-facility-id>,<destination-facility-id>]
```

doc/wiki/move-patient-data-for-a-user.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+* Reattach the user to the correct facility using the edit UI (eg. https://api.simple.org/admin/users/user_id/edit) and picking the correct registration facility from the drop down:
+
+![](resources/select-user-facility.png)
+
+Fixing a. is tricky since the user's syncs would be failing (because facility is soft-deleted) and hence we can't recover their data. We can currently only fix this manually if we deem it important enough:
+
+* Un-delete the soft-deleted facility temporarily
```
vkrmis, line 29: Any region related nuances to be aware of here?

kitallis: Good point, I guess I should add those in here?

kitallis: Done.

vkrmis: Somewhere else, with a link here also works.

vkrmis (approved):

PR #1999, opened 2021-01-21
Add additional BP passport analytics

kitallis:
**Story card:** [ch2249](https://app.clubhouse.io/simpledotorg/story/2249/merge-cross-block-duplicate-patients)

## Because

We need more BP Passport related data to make an informed call for [merging duplicate patients](https://docs.google.com/document/d/1TItK2-kpQj91lNmOhzmq2CP1J15AB5knceobwZOIKoA/edit#).

## This addresses

* Adds a bunch of metrics and queries for Datadog.
* Generates a trend report for the various metrics as a PDF and allows mailing it directly to a power user.


harimohanraj89 (approved):
Looks pretty good. This is standalone tooling that generates those PDFs right? Let's get it in there. We can iterate as needed.

PR #2001, opened 2021-01-21
Exclude LTFU patients in reports

prabhanshuguptagit:
**Story card:** [ch2230](https://app.clubhouse.io/simpledotorg/story/2230/exclude-lost-to-follow-up-from-region-reports-overview)

## Because

LTFU patients are HTN patients who have not had a BP recorded in the past 12 months and have been registered for at least 12 months. We want to exclude these patients from our reports.

## This addresses

- [x] Removes LTFU patients from all reporting. This is flippered by the `with_exclusions` flag.
- [x] Adds a toggle in BP control card in Region reports for power users to include/exclude on the fly:
image


app/models/concerns/patient_reportable.rb
```diff
@@ -1,18 +1,35 @@
...
     scope :excluding_dead, -> { where.not(status: :dead) }
+    scope :excluding_ltfu, ->(ltfu_as_of: Date.today) do
+      where(id: latest_bp_within_ltfu_period(ltfu_as_of).select(:patient_id))
+    end
 
-    scope :for_reports, ->(with_exclusions: false) do
+    scope :for_reports, ->(with_exclusions: false, exclude_ltfu_as_of: nil) do
       if with_exclusions
-        with_hypertension.excluding_dead
+        if exclude_ltfu_as_of
```
kpethtel, line 16: I believe we could avoid this branching logic by rewriting this as:
`
skope = with_hypertension
skope = skope.excluding_dead if with_exclusions
skope = skope.excluding_ltfu(ltfu_as_of: exclude_ltfu_as_of) if exclude_ltfu_as_of
skope
`
Admittedly, this could cause issues if `with_exclusions` is false and `exclude_ltfu_as_of` is truthy, but that shouldn't happen in principle and could be prevented.

app/models/latest_blood_pressures_per_patient_per_quarter.rb
```diff
@@ -3,11 +3,18 @@ class LatestBloodPressuresPerPatientPerQuarter < ApplicationRecord
...
+    primary_key: :patient_id,
+    foreign_key: :patient_id
 
   def self.refresh
     Scenic.database.refresh_materialized_view(table_name, concurrently: true, cascade: false)
   end
 
   scope :with_hypertension, -> { where("medical_history_hypertension = ?", "yes") }
   scope :excluding_dead, -> { where.not(patient_status: :dead) }
+  scope :excluding_ltfu, ->(ltfu_as_of: Date.today) do
```
kpethtel, line 17: Given the repetitions between this and per_month, would it make sense to extract these into a module?

app/models/concerns/patient_reportable.rb
prabhanshuguptagit: We're planning on enabling exclusions all across at once with a flipper flag, so there _will_ be an interim period when `exclude_ltfu_as_of` will be set and `with_exclusions` will be false. It doesn't look pretty but  `with_exclusions` will get cleaned up after release.

app/models/latest_blood_pressures_per_patient_per_quarter.rb
prabhanshuguptagit: We could, I'll try to add this change if the scope of this PR doesn't end up being too big (which I'm afraid it's going to be). Will refactor in a separate PR otherwise 👍🏼 .

app/models/reports/result.rb
```diff
@@ -173,9 +187,21 @@ def calculate_period_info(range)
...
+        self[:registrations_with_ltfu][period.previous] || 0
+      else
+        adjusted_registrations_with_ltfu_for(period)
+      end
+    end
+
+    def registrations_for_percentage_calculation(period, key)
+      if key == :controlled_patients_with_ltfu
+        return registrations_for_percentage_calculation_with_ltfu(period)
+      end
```
prabhanshuguptagit, line 201: Not super fond of this, but basically we want to be able to switch the denominator when calculating control rate for `with_ltfu` patients.

app/services/control_rate_service.rb
```diff
@@ -41,80 +41,120 @@ def all_cached_data
...
     results.fill_in_nil_registrations
     results.count_cumulative_registrations
-    results.count_adjusted_registrations
+    results.count_adjusted_registrations_with_ltfu
+
+    if with_exclusions
+      results.count_adjusted_registrations
+    else
+      results.adjusted_registrations = results.adjusted_registrations_with_ltfu
+    end
```
prabhanshuguptagit, line 57: The adjusted registrations are defaulted to `adjusted_registrations_with_ltfu` if the flipper flag is off.

app/views/reports/regions/show.html.erb
```diff
@@ -304,4 +355,7 @@
 <div id="data-json" style="display: none;">
   <%= raw @data.to_json %>
 </div>
-<script src="https://cdn.jsdelivr.net/npm/chart.js@2.8.0"></script>
\ No newline at end of file
+<script src="https://cdn.jsdelivr.net/npm/chart.js@2.8.0"></script>
+<script>
+  window.withLtfu = <%= @with_ltfu %>;
```
prabhanshuguptagit, line 360: Is there a better way to pass this flag from the controller to JS?

app/models/latest_blood_pressures_per_patient_per_quarter.rb
prabhanshuguptagit: Added in 2ec7c97.

app/views/reports/regions/show.html.erb
prabhanshuguptagit: #2039 namespaces this JS code, will address passing this as a parameter to the reports function in that PR.

app/models/reports/result.rb
```diff
@@ -93,10 +98,12 @@ def last_key(key)
...
     [:period_info, :earliest_registration_period,
-      :registrations, :registrations_with_exclusions,
-      :adjusted_registrations, :cumulative_registrations,
+      :registrations, :denominator_registrations,
+      :ltfu_patients, :cumulative_registrations,
+      :adjusted_registrations, :adjusted_registrations_with_ltfu,
       :missed_visits, :missed_visits_rate,
-      :controlled_patients, :controlled_patients_rate,
+      :controlled_patients, :controlled_patients_with_ltfu,
+      :controlled_patients_rate, :controlled_patients_with_ltfu_rate,
```
harimohanraj89, line 106: I'm sensing some simplification in this area.
* Aren't controlled patients and controlled patients with LTFU the same? If you're controlled, you've had a BP in the last 3 months, you cannot be LTFU.
* The same would apply for uncontrolled patients
* The total patients (denominator) is what would switch between with or without LTFU

Hmm, here's the amorphous idea in my head. We store the following datapoints in the result. Each point would be an array of 24 numbers for the last 24 months.
1. Dead patients
2. LTFU patients
3. Not LTFU patients
4. Controlled patients
5. Uncontrolled patients

1 + 2 + 3 should be the total registered, right?

4 / (1 + 2 + 3) is our current control rate

4 / 3 is the new exclusioned control rate

5 / (1 + 2 + 3) is our current control rate

5 / 3 is the new exclusioned control rate

(1 + 2 +3) - (4 + 5) is current missed visit count, and that / (1 + 2 + 3) is the current missed visit rate

(3) - (4 + 5) is the new missed visit count, and that / (3) is the new missed visit rate.

Are these ideas already implemented? Do I have incorrect assumptions? I'd love to have you walk me through this 😅

harimohanraj89 (changes requested):
One round of questions and feedback below. I'd like to pair with you to dig into the queries themselves, @prabhanshuguptagit . Let's do that soon.

app/controllers/api/v3/analytics/user_analytics_controller.rb
```diff
@@ -4,12 +4,16 @@ class Api::V3::Analytics::UserAnalyticsController < Api::V3::AnalyticsController
   layout false
 
   def show
-    @user_analytics = UserAnalyticsPresenter.new(current_facility)
+    @user_analytics = UserAnalyticsPresenter.new(current_facility, with_exclusions: report_with_exclusions?)
```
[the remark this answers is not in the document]
prabhanshuguptagit: Yes, this handles the graph at the bottom and the cohort report graph.

app/controllers/reports/regions_controller.rb
```diff
@@ -40,10 +40,11 @@ def index
   end
 
   def show
-    @data = Reports::RegionService.new(region: @region, period: @period).call
+    @data = Reports::RegionService.new(region: @region, period: @period, with_exclusions: report_with_exclusions?).call
```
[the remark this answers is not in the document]
prabhanshuguptagit: `with_exclusions` as in "with exclusions applied". Does `exclusions_applied:` sound better?

app/models/concerns/patient_reportable_matview.rb
```diff
@@ -0,0 +1,24 @@
...
+  included do
+    def self.refresh
+      Scenic.database.refresh_materialized_view(table_name, concurrently: true, cascade: false)
+    end
+
+    scope :with_hypertension, -> { where("medical_history_hypertension = ?", "yes") }
+    scope :excluding_dead, -> { where.not(patient_status: :dead) }
+
+    scope :ltfu_as_of, ->(date) do
+      where.not("bp_recorded_at > ? AND bp_recorded_at < ?", date.to_date - TIME_TO_CONSIDER_LTFU, date.to_date)
```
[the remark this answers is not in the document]
prabhanshuguptagit: Thanks. Fixed in 2b462666. The `BETWEEN` version of this query was a lot slower, hence the range comparisons.

app/controllers/reports/regions_controller.rb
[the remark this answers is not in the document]
harimohanraj89: Nah I was just confused I think. `with_exclusions` makes sense.

app/models/concerns/patient_reportable_matview.rb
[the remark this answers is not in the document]
harimohanraj89: Oh darn. That sucks.

app/services/reports/region_cache_warmer.rb
```diff
@@ -24,36 +24,15 @@ def call
...
-
-        notify "starting facility caching"
-        Statsd.instance.time("region_cache_warmer.facilities") do
-          cache_facilities
-        end
-        Statsd.instance.time("region_cache_warmer.with_exclusions.facilities") do
-          cache_facilities(with_exclusions: true)
+            RegionService.call(region: region, period: period, with_exclusions: true)
+            Statsd.instance.increment("region_cache_warmer.with_exclusions.#{region.region_type}.cache")
+          end
```
harimohanraj89, line 35: Nice. We'll still get individual stats on facility cache times vs state cache times, etc, right?

spec/services/reports/region_cache_warmer_spec.rb
```diff
@@ -38,10 +38,9 @@ def refresh_views
 
   it "warms the cache for all regions" do
     _facilities = FactoryBot.create_list(:facility, 5, block: "Block 1", facility_group: facility_group_1)
-
-    expect(Reports::RegionService).to receive(:call).with(hash_including(region: instance_of(Region))).exactly(4).times
-    expect(Reports::RegionService).to receive(:call).with(hash_including(region: instance_of(FacilityGroup))).exactly(2).times
-    expect(Reports::RegionService).to receive(:call).with(hash_including(region: instance_of(Facility))).exactly(10).times
+    # we don't cache the Root or Organization regions
+    regions_to_cache = Region.count - 2
```
harimohanraj89, line 42: This assertion is a bit hand-wavy 😅  I agree, it's awkward to assert specific regions. Could we instead add a `expect().not_to receive().with(Org/Root region)`? That will confirm that the 2 missing cache calls are for the root and org regions.

harimohanraj89 (approved):
A few additional comments, but nothing serious. I'm excited to see this thing go live. Let's ensure we do the following thoroughly before we ship:
* Find out how the caching behavior changes when the code's shipped
* Find out how the caching behavior changes when the feature flag is toggled
* Ensure the new denominators are consistent across all reports and myfacilities pages (lets' get the team's help to click test through this)

prabhanshuguptagit:
> Find out how the caching behavior changes when the code's shipped
> Find out how the caching behavior changes when the feature flag is toggled

When we deploy this branch, none of the `with_exclusions` reports will be cached. Cache for existing reports will remain untouched. Once we have deployed this branch, we should ideally wait overnight (so the cache warmer can run for `with_exclusions`) before enabling the `report_with_exclusions` flag.

PR #2004, opened 2021-01-21
Auto deploy when on release tag and previous envs have passed

rsanheim:
Opening this PR to start the discussion about making production delivery easier and more automated. We are releasing more, and the 'click in Semaphore' step is annoying and more burdensome as our build times have increased.

This turns on auto_promote for release tags for prod envs if the
previous stages have passed.  The only other manual step here for
delivery would be the CHANGELOG piece, which we can automate next.

The deploying developer still needs to verify releases to the best of
their ability, of course.

@timcheadle @harimohanraj89 @simpledotorg/server-developers  I think this is a worthwhile change - do you think? One additional piece that would be nice would be more clear notifications for when deploys start and when they complete, right now the Semaphore notifications leave a bit to be desired.


harimohanraj89 (approved):
I consciously left production deployments manual when we started to take a conservative approach. Now that we've done this for a while, I feel confident that we can auto-promote 👍 

In fact, if this makes us feel nervous (ie, we want to test releases in Demo before clicking the big prod button), that is indicative of other problems in our process. For example, why did we merge the PR in the first place if we weren't confident about it?

All in all, let's do it 🚀

PR #2005, opened 2021-01-22
Make sure we find correct model via the region slug

rsanheim:
Story:  [ch2397](https://app.clubhouse.io/simpledotorg/story/2397/some-districts-aren-t-loading-for-users-with-region-reports-turned-off)

We need to make sure we find FacilityGroups and Facilities if region_reports is
disabled, and find Regions if region_reports is enabled. Otherwise slug
mismatches can result in records not being found, which ends up
returning authorization errors to users.

We also need to duck type source in some of the source models to make
sure the download links still work correctly.

This will all be resolved once all users are in the world of region
reports, and are list views _only_ show Regions and never FGs or
Facilities...so the links will have the correct slugs.

### DIFF
--- a/app/controllers/reports/regions_controller.rb
+++ b/app/controllers/reports/regions_controller.rb
@@ -119,7 +119,12 @@ class Reports::RegionsController < AdminController
     current_admin.region_access(memoized: true).accessible_region?(region, action)
   end
 
+  def region_reports_enabled?
+    current_admin.feature_enabled?(:region_reports)
+  end
+
   helper_method :accessible_region?
+  helper_method :region_reports_enabled?
 
   def download_filename
     time = Time.current.to_s(:number)
@@ -208,9 +213,5 @@ class Reports::RegionsController < AdminController
   def percentage(numerator, denominator)
     return 0 if denominator == 0 || numerator == 0
     ((numerator.to_f / denominator) * 100).round(2)
-  end
-
-  def region_reports_enabled?
-    current_admin.feature_enabled?(:region_reports)
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/reports/regions_controller.rb
```diff
@@ -163,11 +168,19 @@ def find_region
         scope = current_admin.accessible_facilities(:view_reports)
         FacilityDistrict.new(name: report_params[:id], scope: scope)
       when "district"
-        current_admin.accessible_district_regions(:view_reports).find_by!(slug: report_params[:id])
+        if region_reports_enabled?
+          current_admin.accessible_district_regions(:view_reports).find_by!(slug: report_params[:id])
+        else
+          current_admin.accessible_facility_groups(:view_reports).find_by!(slug: report_params[:id])
+        end
```
rsanheim, line 175: here is the issue - we need to use the FG slug when region reports are off, because those are the links users will be following.

app/controllers/reports/regions_controller.rb
```diff
@@ -163,11 +168,19 @@ def find_region
...
       when "block"
         current_admin.accessible_block_regions(:view_reports).find_by!(slug: report_params[:id])
       when "facility"
-        current_admin.accessible_facility_regions(:view_reports).find_by!(slug: report_params[:id])
+        if region_reports_enabled?
+          current_admin.accessible_facility_regions(:view_reports).find_by!(slug: report_params[:id])
+        else
+          current_admin.accessible_facilities(:view_reports).find_by!(slug: report_params[:id])
```
rsanheim, line 182: same here with facilities...

PR #2006, opened 2021-01-22
Region based reports: Search bar in reports

claudiovallejo:
**Story card:** [ch2409](https://app.clubhouse.io/simpledotorg/story/2409/add-search-bar-to-reports-pages)

## Because

Adding reports search bar to Reports pages for faster navigation between reports.

## This addresses

* Creates partial for search `<script>`
* Appends search script partial into search partial
* Adds search partial to reports pages

**Mobile view**
image

**Desktop view**
image


app/views/reports/regions/_search_script.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    search(input);
+    return false;
+  })
+  form.on("keyup", function(event) {
+    search(event.target);
+    return false;
+  })
+
+  const async = false
+  new AdminAccess("facility-access").initialize(async)
```
rsanheim, line 15: @claudiovallejo You'll want to add a guard here to only create `AdminAccess` if `facility-access` is on the page.  Otherwise this will break.

app/views/reports/regions/index.html.erb
```diff
@@ -21,24 +21,6 @@
...
-          form.on("keyup", function(event) {
-            search(event.target);
-            return false;
-          })
-
-          const async = false
-          new AdminAccess("facility-access").initialize(async)
-        </script>
-
       <% else %>
```
rsanheim, line 24: I may be missing something, but doesn't the index page now need the search partial as well?

rsanheim (changes requested):

app/views/reports/regions/_search_script.html.erb
```diff
@@ -0,0 +1,16 @@
+<script>
```
ssrihari, line 1: Since this is all static, I would suggest moving this javascript to a js file.

app/views/reports/regions/_search_script.html.erb
ssrihari: We can move these last two lines back to the index page, since the access stuff is specific to that page.

app/views/reports/regions/index.html.erb
ssrihari: This page does render the search partial, above on [line 8](https://github.com/simpledotorg/simple-server/pull/2006/files#diff-73451de3a663ee58032c74f69fb20b378538fd8e388f7b1cf79b1c04b8241ba7R8).

app/views/reports/regions/_search_script.html.erb
claudiovallejo: Good suggestion — we'll resolve this in a future PR when we overhaul our assets pipeline to webpack.

rsanheim (approved):

claudiovallejo:
Closing PR because semaphore tests are strangely failing. Created a new PR #2027

PR #2010, opened 2021-01-26
My Facilities Improvements: Remove new "Registrations" tables

claudiovallejo:
**Story card:** [ch2423](https://app.clubhouse.io/simpledotorg/story/2423/remove-registrations-page)

## Because

The new "Registrations" table displays rates that rely on "OPD load" — and the definition of OPD load is still being determined by leadership.

## This addresses

* Hides the "Registrations" page for users with access to the `:my_facilities_improvements` feature flag
* Redirects users with access to the `:my_facilities_improvements` feature flag to the "Home" overview page
* Updates registration controller and view to show old "Registration" tables for users without access to feature flag

### DIFF
--- a/app/controllers/my_facilities_controller.rb
+++ b/app/controllers/my_facilities_controller.rb
@@ -70,6 +70,7 @@ class MyFacilitiesController < AdminController
   def bp_not_controlled
     unless current_admin.feature_enabled?(:my_facilities_improvements)
       redirect_to my_facilities_overview_path(request.query_parameters)
+      return
     end
 
     @facilities = filter_facilities
@@ -87,7 +88,6 @@ class MyFacilitiesController < AdminController
 
     if current_admin.feature_enabled?(:my_facilities_improvements)
       redirect_to my_facilities_overview_path(request.query_parameters)
-      return
     else
       registrations_query = RegistrationsQuery.new(facilities: @facilities,
                                                    period: @selected_period,
--- a/app/controllers/my_facilities_controller.rb
+++ b/app/controllers/my_facilities_controller.rb
@@ -88,7 +88,6 @@ class MyFacilitiesController < AdminController
 
     if current_admin.feature_enabled?(:my_facilities_improvements)
       redirect_to my_facilities_overview_path(request.query_parameters)
-      return
     else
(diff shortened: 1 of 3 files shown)


app/views/my_facilities/registrations.html.erb
```diff
@@ -1,5 +1,3 @@
-<% if current_admin.feature_enabled?(:my_facilities_improvements) %>
-  <%= render "registrations_details_v2" %>
-<% else %>
+<% unless current_admin.feature_enabled?(:my_facilities_improvements) %>
```
kpethtel, line 1: Do we need this guard here? Isn't it handled by the controller?

app/controllers/my_facilities_controller.rb
```diff
@@ -87,17 +87,7 @@ def registrations
...
-
-      @facilities.each do |facility|
-        @data_for_facility[facility.name] = Reports::RegionService.new(region: facility, period: @period).call
-        @scores_for_facility[facility.name] = Reports::PerformanceScore.new(region: facility,
-                                                                            reports_result: @data_for_facility[facility.name],
-                                                                            period: @period)
-      end
-
-      @facilities_by_size = @facilities.group_by { |facility| facility.facility_size }
+      redirect_to my_facilities_overview_path(request.query_parameters)
```
rsanheim, line 90: You'll want to add a `&& return` here just for safety. I forget if Rails 5.x still requires that, but it shouldn't hurt.

app/views/my_facilities/registrations.html.erb
claudiovallejo: That's a great point - thanks @kpethtel !

app/controllers/my_facilities_controller.rb
```diff
@@ -87,17 +87,8 @@ def registrations
...
+      return
```
rsanheim, line 91: Oh I see, our linter sees that the redirect ends up being the last line executed because if the `if/else` here.  So maybe you don't need the explicit return. 🤷‍♂️

claudiovallejo:
Closing PR because this is addressed in #2013

PR #2012, opened 2021-01-27
Drug stock report

ssrihari:
**Story card:** [ch2223](https://app.clubhouse.io/simpledotorg/story/2223/add-drug-stock-reports-to-dashboard)

## Because

We want to display stock reports based on the Drug Stock data we're receiving

## This addresses

- Adds a calculation class to model the patient days formulas per state
- Calculate and display patient days numbers per drug category
- Display drug stock information
- Add coefficients for currently known state protocols
- Cache report, bust cache when data is entered
- Add a trait to the Protocol factory, to create protocol drugs with rxnorm_codes

Screenshot 2021-01-29 at 4 57 25 PM


app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -14,11 +14,19 @@ def index
     @facilities = filter_facilities
       .includes(facility_group: :protocol_drugs)
       .where(protocol_drugs: {stock_tracked: true})
+
+    render && return if @facilities.empty?
+    # handle no facilities
```
harimohanraj89, line 19: I think the code's fairly clear. We can remove this comment.

app/models/drug_stock.rb
```diff
@@ -7,9 +7,15 @@ class DrugStock < ApplicationRecord
   validates :received, numericality: true, allow_nil: true
   validates :for_end_of_month, presence: true
 
+  def self.latest_for_facilities(facilities, for_end_of_month)
+    DrugStock.select("DISTINCT ON (facility_id, protocol_drug_id) *")
+      .includes(:protocol_drug)
+      .where(facility: facilities.pluck(:id), for_end_of_month: for_end_of_month)
```
harimohanraj89, line 13: `for_end_of_month: for_end_of_month` Looks like `for_end_of_month` is a full-fledged time stamp. Is this check _too_ precise? Is it prone to breaking if, say, we change the timezone that the dashboard operates in?

app/queries/drug_stocks_query.rb
```diff
@@ -0,0 +1,119 @@
...
+    new_patient_coefficient = coefficients[:drug_categories][drug_category][:new_patient_coefficient]
+    estimated_patients = patient_count * coefficients[:load_factor] * new_patient_coefficient
+    {stocks_on_hand: stocks_on_hand,
+     patient_count: patient_count,
+     load_factor: coefficients[:load_factor],
+     new_patient_coefficient: new_patient_coefficient,
+     estimated_patients: estimated_patients,
+     patient_days: (stocks_on_hand.map { |stock| stock[:stock_on_hand] }.reduce(:+) / estimated_patients).floor}
+  rescue
+    # either drug stock is nil, or drug is not in formula
```
harimohanraj89, line 98: Can we manually report the exception to Sentry? We should know what is going wrong, especially if we're doing a blanket `rescue`.

app/queries/drug_stocks_query.rb
```diff
@@ -0,0 +1,119 @@
+class DrugStocksQuery
+  def initialize(facilities:, for_end_of_month:)
+    @facilities = facilities
+    @for_end_of_month = for_end_of_month
+    # assuming that all facilities on the page have the same protocol
+    @protocol = @facilities.first.protocol
+  end
+
+  def protocol_drugs_by_category
+    @drug_categories ||= @protocol.protocol_drugs
```
harimohanraj89, line 10: Let's name the caching var the same as the method for consistency.

app/queries/drug_stocks_query.rb
```diff
@@ -0,0 +1,119 @@
...
+      .where(stock_tracked: true)
+      .order(:name, :dosage)
+      .sort_by {|protocol_drug| [protocol_drug.name, protocol_drug.dosage.to_i]}
+      .group_by(&:drug_category)
+      .sort_by {|(drug_category, _) | drug_category}
+      .to_h
+  end
+
+  def call
+    patient_counts = Patient.where(assigned_facility: @facilities.map(&:id)).group(:assigned_facility).count
```
harimohanraj89: You shouldn't need to `map(&:id)` here.

harimohanraj89 (commented):
Some initial thoughts.

app/controllers/my_facilities/drug_stocks_controller.rb
ssrihari: Ha no, this was a TODO for us 😁 . Removed after handling in f9e2efdcbde68ce341d1b767c98bf1c3feb5d701.

app/models/drug_stock.rb
ssrihari: Good point, hadn't thought about timezones, and haven't really looked at how we handle it. Will check and revert back. We are setting this value in the controller, so we're quite in control of it either way.

app/queries/drug_stocks_query.rb
ssrihari: Yup, that was the plan here. We're planning on extracting the patient formula calculation to the model space, will pick this up along with that.

app/queries/drug_stocks_query.rb
ssrihari: Ah, this must've slipped while refactoring. Will fix, thanks!

app/queries/drug_stocks_query.rb
ssrihari: We tried that first, and I don't remember the exact error we ran into. @vkrmis?

vkrmis: I think we were doing something slightly different then, `assigned_facility: @facilities` should work

ssrihari: `PG::AmbiguousColumn: ERROR:  column reference "id" is ambiguous` is what I get when trying this.

app/queries/drug_stocks_query.rb
ssrihari: Fixed in 3f3b63c61c00a1a0595608fb9ce543e6bb995796.

app/models/drug_stock.rb
ssrihari: We've changed the column type here to a date now with 134a75e731ed83a6b218b05aec352202a8c99876, so we shouldn't have a tz issue with this anymore. More over, the dates are chosen based on a dropdown during creation, and not based on relative time, so we needn't handle user specific timezones I think.

app/helpers/my_facilities_helper.rb
```diff
@@ -18,4 +18,43 @@ def opd_load(facility, selected_period)
...
+      <% end %>
+      <span class='math'>
+        /<%= report[:patient_count] %>*<%= report[:load_coefficient] %>*<%= report[:new_patient_coefficient] %>
+      </span>
+      Patients
+      <br>
+      <span class='math'>=<%= report[:patient_days] %></span>
+      Patient days
+    EOF
+    tooltip_template.result(binding)
```
ssrihari, line 38: Is there a better way to do this where the ERB isn't in plain text? Don't want to read it from a file either.

spec/controllers/my_facilities/drug_stocks_controller_spec.rb
```diff
@@ -69,7 +69,7 @@
       let(:params) {
         {
           facility_id: facility.id,
-          for_end_of_month: Date.today.strftime("%B %Y")
+          for_end_of_month: Date.today.strftime("%b-%Y")
```
ssrihari, line 72: We will extract `"%b-%Y"` into a constant in another PR since it will change a lot of files in this one.

app/queries/drug_stocks_query.rb
```diff
@@ -0,0 +1,103 @@
...
+    @for_end_of_month = for_end_of_month
+    # assuming that all facilities on the page have the same protocol
+    @protocol = @facilities.first.protocol
+    @state = @facilities.first.state
+  end
+
+  def protocol_drugs_by_category
+    @protocol_drugs_by_category ||= @protocol.protocol_drugs
+      .where(stock_tracked: true)
+      .order(:name, :dosage)
```
harimohanraj89, line 15: Why order if we're sorting right under?

app/queries/drug_stocks_query.rb
```diff
@@ -0,0 +1,103 @@
...
+      .sort_by(&:sort_key)
+      .group_by(&:drug_category)
+      .sort_by { |(drug_category, _)| drug_category }
```
harimohanraj89, line 18: What does this sort achieve? Looks like we're making the final result a hash anyway, so does the ordering matter?

app/queries/drug_stocks_query.rb
```diff
@@ -0,0 +1,103 @@
...
+      @facilities.map(&:id).sort,
+      @for_end_of_month,
+      @protocol.id,
+      @state,
+      CACHE_VERSION
+    ].join("/")
+  end
+
+  def call
+    Rails.cache.fetch(cache_key,
```
harimohanraj89, line 34: Have we considered caching at the per-facility level? It's more cache keys, but each cache key is pretty neatly scoped, and may set us up nicely for future iterations like "show me all large facilities", or "show me all facilities in a block", etc.

app/queries/drug_stocks_query.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    drug_stocks = facility_drug_stocks.group_by { |drug_stock| drug_stock.protocol_drug.drug_category }
+    protocol_drugs_by_category.each do |(drug_category, _protocol_drugs)|
+      stocks = stocks_by_rxnorm_code(drug_stocks[drug_category])
+      patient_days = Reports::PatientDaysCalculation.new(@state, @protocol, drug_category, stocks, patient_count).calculate
+      next if patient_days.nil?
+      facility_report[drug_category] = patient_days.merge(drug_stocks: drug_stocks_by_drug_id(drug_stocks[drug_category]))
+    end
+    facility_report
+  end
```
harimohanraj89, line 70: This method is pretty dense, I'm having a tough time understanding what's going on here. Can you post an example payload? I'm curious if we can work backwards from there to craft a more readable way to express these operations.

app/queries/drug_stocks_query.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def drug_stock_totals
+    DrugStock.latest_for_facilities(@facilities, @for_end_of_month).group_by(&:protocol_drug).map { |(protocol_drug, drug_stocks)|
+      [protocol_drug, drug_stocks&.map(&:in_stock)&.compact.sum]
```
harimohanraj89, line 74: Agree with Standard. The plain `.sum` at the end is dangerous.

harimohanraj89 (changes requested):
Initial round of questions from me.

app/helpers/my_facilities_helper.rb
ssrihari: This is rendered a few times for every row, and we might have a few thousand rows at a time. Having gone the partial route with the form per row, we saw a lot of latency (10s of seconds) added to rendering time of the page.

In addition, this is actually a string that goes into a data field in html.

Will still give the partial a try to see if the latency impact is worth it. cc @vkrmis

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -38,24 +41,20 @@ def create
...
     end
-
-    if drug_stocks.empty?
-      redirect_to report_url_with_filters
-    elsif drug_stocks.all?(&:valid?)
-      redirect_to report_url_with_filters, notice: "Saved drug stocks"
-    elsif drug_stocks.any?(&:invalid?)
-      redirect_to report_url_with_filters, alert: "Something went wrong, Drug Stocks were not saved."
-    end
+    redirect_to redirect_url(force_cache: true), notice: "Saved drug stocks"
```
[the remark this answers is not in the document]
ssrihari: The expected behaviour here is that modifying the drug stock (by creating them) should reflect in the report page they just came from. Hence the deliberate cache busting on saving new drug stocks.

app/queries/drug_stocks_query.rb
ssrihari: Oh good catch :) Missed this while refactoring.

app/queries/drug_stocks_query.rb
ssrihari: This is the order in which the column groups, and columns are [displayed on the page](https://github.com/simpledotorg/simple-server/pull/2012/files/f6879780cc577fb49e4a5b74b17367da74c5da77#diff-d7918cfff1a26d2435b9e7511eb9fd48a0f1244a0106e31cd8d445c0b89d3204R22). We use this method as the canonical way to iterate over drug categories and drugs throughout this report as well.

The hash preserves the sorted order as per spec, I think. Can explicitly choose an ordered hash if we think that's better.

app/queries/drug_stocks_query.rb
ssrihari: This report is generally not very expensive, since the drug stocks per facility per month aren't going to be that many and the patient days calculation per facility isn't too intense. So we'll probably get very little out of caching it at a facility level. This is also why we didn't think too much about busting the cache anytime a drug stock is added for a facility in the list.

It's still somewhat useful to cache at a block level for example (that's the default view) since there's totalling and there might be a few thousand facilities in some blocks. And because we generally want the page loads to be quick.

app/queries/drug_stocks_query.rb
ssrihari: There are a couple of examples [in the spec](https://github.com/simpledotorg/simple-server/pull/2012/files/f6879780cc577fb49e4a5b74b17367da74c5da77#diff-e82c91761e53671a526dd257d939fad4211ce7f669ec69a2a02cdeebdbf7d1fcR36) that would help here I think.

Here's one way to think about this. There are two hierarchies:
1. Protocol -> Protocol Drug Categories -> Protocol Drugs (columns)
2. Facility -> Protocol Drug Categories -> Drug Stocks [and patient days] (rows)

We need to traverse through both these for every row, since there might be protocol drugs that don't have drug stocks in that facility, etc.

I do agree that the method is dense, however. Totally open to ideas. We could extract a method for each drug category so it's easier to read perhaps.

app/queries/drug_stocks_query.rb
ssrihari: Fixed in f31505e40e228389038f3b069073479562337d12.

app/queries/drug_stocks_query.rb
ssrihari: Fixed in f31505e40e228389038f3b069073479562337d12.

app/models/protocol_drug.rb
```diff
@@ -19,4 +19,8 @@ class ProtocolDrug < ApplicationRecord
   def assign_id
     self.id = SecureRandom.uuid
   end
+
+  def sort_key
+    [name, dosage.to_i]
```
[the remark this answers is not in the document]
ssrihari: Fixed in f31505e40e228389038f3b069073479562337d12.

app/models/protocol.rb
```diff
@@ -10,6 +10,12 @@ class Protocol < ApplicationRecord
 
   auto_strip_attributes :name, squish: true, upcase_first: true
 
+  def as_json
+    protocol_json = super
```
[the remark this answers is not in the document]
ssrihari: Fixed in f31505e40e228389038f3b069073479562337d12.

app/helpers/my_facilities_helper.rb
harimohanraj89: Neat. If performance sucks, then let's leave it like this and add a comment here explaining why this isn't a partial.

app/queries/drug_stocks_query.rb
harimohanraj89: Oh I see, the sort by drug category sorts by the enum order? Nice. I was under the impression that it would sort by category name, putting ARB drugs first.

ssrihari: Oh it is ordered by name, and ARB is indeed first. We verified this with @dburka as well that this is okay.

app/models/reports/patient_days_calculation.rb
```diff
@@ -0,0 +1,81 @@
+module Reports
+  class PatientDaysCalculation
+    def initialize(state, protocol, drug_category, stocks_by_rxnorm_code, patient_count)
```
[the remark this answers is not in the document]
ssrihari: Fixed in ddd80ed775fd8d7f41c2de181e7cb4c8d97da3ce.

app/queries/drug_stocks_query.rb
ssrihari: Extracted in 9877b110588afbc3cabb398af064bf59ab7b5967. We feel this reads easier. We've swept some of the complexity under the carpet of another function, but this is perhaps good enough for now.

Given that the data structure here is quite nested, we could probably use a better abstraction that would return normalised data for each row. We can go there the next time we visit this, or when we look at the consumption report perhaps.

-- @vkrmis and @ssrihari

app/helpers/my_facilities_helper.rb
ssrihari: We measured the performance, and it looks like partials take twice as long to render. However, the difference for about a hundred facilities is in hundreds of milliseconds, so it's okay to keep the tooltip in the partial for now. We can move it into a helper if we notice it getting slower with prod data.

Measurements in ms (page, view times):
- with partial, soft refresh: 847, 562
- with partial, hard refresh: 1219, 697
- without partial, hard refresh: 702, 356
- without partial, soft refresh: 770, 321

ssrihari: Fixed in 3ea0e0ee7465eced766cfb21f091eb5960559ec6.

app/controllers/my_facilities/drug_stocks_controller.rb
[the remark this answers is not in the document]
harimohanraj89: Ah, understood. 👍

app/queries/drug_stocks_query.rb
harimohanraj89: Cool cool 👍 The cache keys become unwieldy but we don't need worry about that, right? We never interact with them directly.

app/queries/drug_stocks_query.rb
harimohanraj89: Surprisingly, the method extraction actually cleans up the readability quite nicely! ✨ The nested structure of facility -> drug category -> stocks is more evident now. Nice.

app/queries/drug_stocks_query.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    drug_stocks = facility_drug_stocks.group_by { |drug_stock| drug_stock.protocol_drug.drug_category }
+    protocol_drugs_by_category.each do |(drug_category, _protocol_drugs)|
+      facility_report[drug_category] = drug_stock_for_category(drug_category, drug_stocks, patient_count)
+    end
+    facility_report
+  end
+
+  def drug_stock_for_category(drug_category, drug_stocks, patient_count)
+    patient_days = Reports::PatientDaysCalculation.new(
```
harimohanraj89, line 69: I stared at this method for a little bit wondering why we return _nothing_ if there is no patient days information to provide. I then realized, this service puts together the full drug stock info - how many of each drug left + how many patient days that comes to.

Can we rename the service to something like `Reports::DrugStockCalculation` to make these responsibilities more evident?

lib/seed/runner.rb
```diff
@@ -35,6 +35,8 @@ def call
       result = FacilitySeeder.call(config: config)
       total_counts[:facility] = result&.ids&.size || 0
       UserSeeder.call(config: config)
+      announce "Seeding drug stocks..."
+      seed_drug_stocks
```
harimohanraj89, line 39: The `FacilitySeeder` sometimes does not create new facilities, if enough facilities exist already. I think it should be ok for us to create extra drug stock records, since we only surface "most recent per month" anyway. Just wanted to flag here to be sure.

lib/seed/runner.rb
```diff
@@ -69,6 +71,15 @@ def seed_patients(progress)
       results
     end
 
+    def seed_drug_stocks
+      Facility.all.each do |facility|
+        user = facility.users.first
+        facility.protocol.protocol_drugs.where(stock_tracked: true).each do |protocol_drug|
+          FactoryBot.create(:drug_stock, facility: facility, user: user, protocol_drug: protocol_drug)
```
harimohanraj89, line 78: Sanity-check - let's ensure this factory call doesn't create extra tangential records like admins, organizations, etc.

spec/factories/protocols.rb
```diff
@@ -3,5 +3,61 @@
...
+          build(:protocol_drug,
+            name: "Chlorthalidone",
+            dosage: "12.5 mg",
+            rxnorm_code: "331132",
+            stock_tracked: true,
+            drug_category: "hypertension_diuretic",
+            protocol_id: id)
+        ]
+      }
+    end
```
harimohanraj89, line 61: This is very nice. If you're feeling up to it, wanna replace the code in `ProtocolSeeder` with a call to this factory? Optional.

spec/models/reports/patient_days_calculation_spec.rb
```diff
@@ -0,0 +1,220 @@
...
+        state: state,
+        protocol: protocol,
+        drug_category: drug_category,
+        stocks_by_rxnorm_code: stocks_by_rxnorm,
+        patient_count: patient_count
+      ).calculate
+      expect(result).to eq(patient_days: "error")
+    end
+
+    it "sentry" do
```
harimohanraj89, line 179: it "reports exceptions"? For a more natural reading :)

ssrihari: haha! we wrote this down as a placeholder while listing all tests to write. will fix.

spec/queries/drug_stocks_query_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+  let!(:drug_category) { "hypertension_ccb" }
+  let!(:stocks_by_rxnorm) {
+    {"329528" => 10000, "329526" => 20000, "316764" => 10000, "316765" => 20000, "979467" => 10000}
+  }
+  let!(:punjab_drug_stock_config) {
+    {"load_coefficient" => 1,
+     "drug_categories" =>
+        {"hypertension_ccb" => {"new_patient_coefficient" => 1.4, "329528" => 1.2, "329526" => 2},
+         "hypertension_arb" => {"new_patient_coefficient" => 0.37, "316764" => 1, "316765" => 2, "979467" => 1},
+         "hypertension_diuretic" => {"new_patient_coefficient" => 0.06, "316049" => 1, "331132" => 1}}}.with_indifferent_access
```
harimohanraj89, line 18: The test data can be simpler since we're mocking this anyway. `1`, `2`, `1.5`, easy coeffs.

app/queries/drug_stocks_query.rb
ssrihari: It actually returns the "calculation" of patient days, which includes all the factors that go into that equation and are then shown in the tooltip. The rest of the drug stock info on the report does not actually come from here.

However, we'll change around these lines with the consumption report since we'll reuse the class for that as well.

spec/models/reports/patient_days_calculation_spec.rb
ssrihari: Fixed in 3795213aa1a165ebc6ed76e3fb4b333014e240c6.

lib/seed/runner.rb
vkrmis: Yup, this should be fine, we'll just have some extra drug stock records for the month when its run.

lib/seed/runner.rb
vkrmis: Looks okay to me. `facility`, `user`, and `protocol_drug` are the only associations on `DrugStock`.
Since we're providing all three here, there shouldn't be any unnecessary records created.

spec/factories/protocols.rb
vkrmis: Replaced in a90cc9c99ae7bd755b20f4d5e432135bcb13a058

spec/queries/drug_stocks_query_spec.rb
vkrmis: We picked up these values from one of our state protocols. I think its okay to leave in, will give us realistic tables for sandbox data

PR #2014, opened 2021-01-28
More flexible caching approach for Reporting data

rsanheim:

## Because

Trying out a more granular / more flexible approach to grabbing per-period values from a cache.

The ControlRateService is caching too much data in one key, and also doesn't let us grab data for multiple regions at once.

## Explanation

This creates a [Repository](zero) layer that controllers or services can ask for particular data points from. The Repository is responsible for handling the caching and grabbing the data, so that calling code doesn't need to worry about that.

The only require the Repository has is that callers specify 1 to many regions and 1 to many periods that they are interested in...then they can ask for whatever data the Repository supports.

Currently that is just controlled, uncontrolled, an no bp measure counts.

The interface is shaping up nicely here, and it makes it **much** easier to grab _just_ the data you need in a cache MGETs, as opposed to grabbing giant chunks of data and then slicing out the numbers you care about.

### Before deploy

past 2 days of [reports traces](https://app.datadoghq.com/apm/traces?end=1612933891948&index=apm-search&paused=false&query=env%3Aproduction%20service%3Asimple_server%20operation_name%3Arack.request%20resource_name%3A%22Reports%3A%3ARegionsController%23show%22&saved_view=207987&start=1612761091948&storage=hot&streamTraces=true)

image


app/models/reports/repository.rb
```diff
@@ -0,0 +1,100 @@
...
+      else
+        periods.to_a
+      end
+    end
+
+    attr_reader :regions, :periods
+    attr_reader :regions_by_id
+    delegate :cache, :logger, to: Rails
+
+    class CacheKey
```
rsanheim, line 18: This handles extracting out pieces of the cache key that we need for the queries.

app/models/reports/repository.rb
```diff
@@ -0,0 +1,100 @@
...
+      end
+    end
+
+    def lookup_region(id)
+      @regions_by_id.fetch(id)
+    end
+
+    def controlled_patients_info
+      keys = cache_keys(:controlled_patients_info)
+      cached_results = cache.fetch_multi(*keys) { |key|
```
rsanheim, line 43: This does the Redis MGET -- any keys **not** found in the cache is yielded to the block.

app/models/reports/repository.rb
```diff
@@ -0,0 +1,100 @@
...
+      cached_results.each_with_object({}) do |(key, count), results|
+        cache_key = CacheKey.new(key)
+        results[cache_key.slug] ||= Hash.new(0)
+        results[cache_key.slug][cache_key.period] = count
+      end
+    end
+
+    def cache_keys(operation)
+      combinations = regions.to_a.map(&:cache_key_v2).product(periods.map(&:cache_key))
+      combinations.map { |region_key, period_key| [region_key, period_key, operation].join("/") }
```
rsanheim, line 73: We need to know all possible permutations of cache keys to get.

app/models/reports/repository.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    def controlled_patients_for(region, period)
```
rsanheim, line 76: All the below should probably be extracted to a new Query object, so that we can consolidate the logic from here and from the ControlRateService.

rsanheim (commented):
Some notes...

app/models/reports/repository.rb
prabhanshuguptagit: Been thinking about this for a while: We should port the query objects to support the new `Period` interface.

prabhanshuguptagit (commented):
This looks p good! I think being able to slice through regions and periods **consistently** is going to be a huge win. Mostly some minor thoughts on refactoring the cache logic. Apologies for throwing in a couple of large picture-y questions, mostly curious about what other ways we can achieve this interface.

app/models/reports/repository.rb
rsanheim: yeah, the old query has a few problems that prevented me from reusing it:

* it doesn't support regions
* it doesn't support the Period model
* it assumes you want to start from current period and always retrieve the last n periods
* it applies a `recorded_at` where clause to monthly_patients that we don't want 

I think we should extract the new queries for reuse and move away from the old ones.

app/models/reports/repository.rb
```diff
@@ -0,0 +1,100 @@
+module Reports
+  class Repository
```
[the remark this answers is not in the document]
rsanheim: > * Have numbers be first class entities (Think `Reports::ControlRate`, `Reports::Registrations`). These can be independently cached and the controller can pull numbers straight from them. I think this might make it easier to trace numbers from view to the query (which is a bit hard right now). This has some overlap with dismantling `result.rb`.

Yup, I think something like this definitely makes sense.

app/models/reports/cache_key_v2.rb
```diff
@@ -0,0 +1,7 @@
+module Reports
+  module CacheKeyV2
+    def cache_key_v2
```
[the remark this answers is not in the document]
rsanheim: yeah I'm just sketching this API out right now, I mostly want to distinguish it from the existing `cache_key` methods which are different and shouldn't be changed right now.

app/models/reports/repository.rb
```diff
@@ -0,0 +1,68 @@
...
+    attr_reader :with_exclusions
+
+    delegate :cache, :logger, to: Rails
+
+    def controlled_patients_count
+      items = cache_keys(:controlled_patients_count)
+      cached_results = cache.fetch_multi(*items, force: force_cache?) { |item|
+        control_rate_query.controlled(item.region, item.period, with_exclusions: with_exclusions).count
+      }
+      cached_results.each_with_object({}) do |(item, count), results|
```
kpethtel, line 26: Might be worth extracting this into a method, given that we do it three times.

app/models/reports/repository.rb
```diff
@@ -0,0 +1,100 @@
+module Reports
+  class Repository
+    def initialize(regions, periods:)
+      @regions = Array(regions)
```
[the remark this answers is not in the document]
rsanheim: This will enforce that for us -- if you pass in an array it remains an array.  If you pass in one region, it makes it into an array.

I'd like clients to be able to pass either either and not have to worry about, as there are lots of case where we want to do both.

app/models/reports/repository.rb
```diff
@@ -0,0 +1,100 @@
...
+      @regions_by_id = @regions.map { |r| [r.id, r] }.to_h
+
+      @periods = if periods.is_a?(Period)
+        Range.new(periods, periods).to_a
+      else
+        periods.to_a
+      end
```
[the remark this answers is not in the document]
rsanheim: Same here, we definitely want a single period for things like my facilities pages.

rsanheim (commented):
more notes

app/queries/control_rate_query.rb
```diff
@@ -0,0 +1,44 @@
...
+      .where(assigned_facility_id: region.facilities)
+      .where(year: quarter.year, quarter: quarter.number)
+      .where("patient_recorded_at >= ? and patient_recorded_at <= ?", cohort_quarter.beginning_of_quarter, cohort_quarter.end_of_quarter)
+      .order("patient_id, bp_recorded_at DESC, bp_id")
+  end
+
+  def bp_monthly_query(region, period, with_exclusions)
+    control_range = period.blood_pressure_control_range
+    # We need to avoid the default scope to avoid ambiguous column errors, hence the `with_discarded`
+    # Note that the deleted_at scoping piece is applied when the SQL view is created, so we don't need to worry about it here
```
kpethtel, line 34: I'm not sure I fully understand this comment, but wouldn't `unscoped` also avoid the default scope?

spec/queries/control_rate_query_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+      expect(may).to eq(1)
+      expect(june).to eq(1)
+      expect(july).to eq(1)
+      expect(august).to eq(0)
+    end
+  end
+
+  context "quarterly period" do
+    it "returns counts for patients registered in the previous quarter" do
```
kpethtel, line 81: Does this really mean it returns counts for the _previous_ quarter? It seems like this test shows that querying for Q4 does not return the count for registrations in Q3.

kpethtel (approved):
This is a real mountain of work. Very impressive.

PR #2017, opened 2021-01-28
Migrate to sentry-ruby

rsanheim:
**Story card:** [ch2462](https://app.clubhouse.io/simpledotorg/story/2462/upgrade-to-sentry-ruby-for-better-error-tracking)

Migrate to sentry-ruby, which replaces sentry-raven. Full docs on the migration benefits and
process here: https://docs.sentry.io/platforms/ruby/migration/

I've also added a power-user only tool to force errors in the app or in via a background job. This allows us to verify Sentry is configured correctly, as well as any related alerts (ie slack) around errors.

image


app/controllers/api/v3/sync_controller.rb
```diff
@@ -69,15 +69,12 @@ def params_with_errors(params, errors)
   def capture_errors(params, errors)
     return unless errors.present?
 
-    Raven.capture_message(
-      "Validation Error",
-      logger: "logger",
```
rsanheim, line 74: This logger attributes go away - we weren't really using this in with the old sentry anyways.

app/jobs/tracer_job.rb
```diff
@@ -1,8 +1,11 @@
-class TracerJob < ApplicationJob
```
rsanheim, line 1: We can't use ActiveJob for this...

app/jobs/tracer_job.rb
```diff
@@ -1,8 +1,11 @@
...
-  queue_as :default
-  self.queue_adapter = :sidekiq
+class TracerJob
+  include Sidekiq::Worker
+  sidekiq_options retry: false # job will be discarded if it fails
```
rsanheim, line 3: ...because we don't want failed jobs here to retry, because these are expected errors.

app/controllers/admin/error_traces_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+      redirect_to admin_error_traces_path, notice: "Background job error tracer submitted"
+    else
+      raise Boom, "Error test created by #{current_admin.full_name}"
+    end
+  end
+
+  private
+
+  def require_power_user
+    authorize { current_admin.power_user? }
```
rsanheim, line 25: Ideally this would be 'technical staff' or something like that, but we don't have a user access level like that.  Power user is the best we can do.

rsanheim (commented):
notes for reviewers

app/controllers/api/v3/sync_controller.rb
timcheadle, line 74: Does `sentry-ruby` automatically know what the logger is?

timcheadle (approved):

PR #2018, opened 2021-01-28
Skip region in the data tables if there is no data

rsanheim:
Fixes https://sentry.io/organizations/resolve-to-save-lives/issues/2183745572/?project=1217715

**Story card:** [ch2455](https://app.clubhouse.io/simpledotorg/story/2455/region-report-breaking-in-ihci-thane)

Fixes an error where I think a newly created region doesn't have any data for the most recent period, so the values are coming back nil here.

### DIFF
--- a/app/models/reports/result.rb
+++ b/app/models/reports/result.rb
@@ -43,6 +43,8 @@ module Reports
       @data
     end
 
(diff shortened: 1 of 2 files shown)


kpethtel (approved):
Nice use of delegation.

PR #2020, opened 2021-01-29
Disallow MO users from block-syncing

kitallis:
**Story card:** [ch2456](https://app.clubhouse.io/simpledotorg/story/2456/users-can-pick-mos-for-teleconsults-outside-their-blocks-causing-the-mos-to-not-find-the-patients)

## Because

Currently, nurses can pick MOs for teleconsults outside their blocks. This will cause the MOs to not find the patients if they both have block-sync enabled.

## This addresses

Make an exception for users who can teleconsult when determining if the user can block-sync.


harimohanraj89 (commented):
Slap a spec in here and it's good to go.

harimohanraj89 (approved):

PR #2029, opened 2021-02-02
Numerator and Denominator copy

claudiovallejo:
**Story card:** [ch2491](https://app.clubhouse.io/simpledotorg/story/2491/numerator-and-denominator-copyhelper)

## Because

BP controlled, BP not controlled, and Missed visits rates appear in the "Home" and "Reports" sections of our Dashboard but use different copy. This PR creates a `CopyHelper` to create copy constants and reuse the same copy across our Dashboard.

## This addresses

* Creates `CopyHelper`
* Creates constants for BP controlled, BP not controlled, and Missed visits numerators
* Creates a constant for the denominator of the above rates
* Uses `CopyHelper` to display numerator and denominator copy in the "Home" "BP controlled", "BP not controlled", and "Missed visits" sections
* Uses `CopyHelper` to display numerator and denominator copy in the "Reports" "Overview" section

## Screenshots

**Reports: Overview cards**
image

**Reports: Visit of all hypertensive patients**
image

**Reports: Child tables**
image

**Home: BP controlled**
image

**Home: BP not controlled**
image

**Home: Missed visits**
image


app/helpers/copy_helper.rb
```diff
@@ -0,0 +1,6 @@
+module CopyHelper
+  BP_CONTROLLED_NUMERATOR_COPY = "Patients with BP <140/90 at their most recent visit in the last 3 months"
```
rsanheim, line 2: This is a good start to consolidating all this, but you may want to move this to Rails builtin i18n library -- check out https://guides.rubyonrails.org/i18n.html#abstracting-localized-code for an example.

That would also make it easy to translate to other languages in the future if necessary.

claudiovallejo: Niceee that's so cool! Thanks @rsanheim! Will delete the `CopyHelper` module and add the numerator and denominator copy to the `en.yml` file

kpethtel (approved):
Nice work moving the copy into localization!

PR #2031, opened 2021-02-03
Turn off regular seed job + misc seed fixes

rsanheim:
**Story card:** [ch2493](https://app.clubhouse.io/simpledotorg/story/2493/seed-data-was-cleared-on-sandbox-and-staging)

* disables regular purge job
* creates a couple regular admin users
* some other misc cleanup


ssrihari (approved):

lib/seed/user_seeder.rb
```diff
@@ -11,15 +13,36 @@ def initialize(config:)
...
+      create_dashboard_admins
+      create_mobile_users
+    end
+
+    private
+
+    def create_dashboard_admins
+      unless EmailAuthentication.exists?(email: "power_user@simple.org")
+        FactoryBot.create(:admin, :power_user, full_name: "Power User", email: "power_user@simple.org", password: config.admin_password)
+      end
```
harimohanraj89, line 35: What happens if a `power_user@simple.org` exists, but they are not a power user? I know it's unlikely, but more robustness here means we won't have to babysit this process as often.

lib/seed/user_seeder.rb
```diff
@@ -11,15 +13,36 @@ def initialize(config:)
...
+
+      unless EmailAuthentication.exists?(email: "cvho@simple.org")
+        fg_1, fg_2 = *FacilityGroup.take(2)
+        user = FactoryBot.create(:admin, :manager, full_name: "CVHO", email: "cvho@simple.org")
+        user.accesses.create! resource_id: fg_1.id, resource_type: "FacilityGroup"
+        user.accesses.create! resource_id: fg_2.id, resource_type: "FacilityGroup"
+      end
```
harimohanraj89, line 42: Same here. Can we address the situation where a "cvho@simple.org" exists, but doesn't have the right permissions? A common pitfall would be - a "cvho@simple.org" exists, but has access records pointing to obsolete/deleted facilities.

Additionally, looks like we can use the `grant_access` method to _reset_ a user's permissions pretty efficiently.

```ruby
facility_groups = FacilityGroup.take(2)
facilities = facility_groups.flat_map(&:facilities)
cvho = User.find_by(email: "cvho@simple.org:") || FactoryBot.create(...)
power_user.grant_access(cvho, facilities)
```
What do you think?

lib/seed/user_seeder.rb
```diff
@@ -11,15 +13,36 @@ def initialize(config:)
...
+    end
```
harimohanraj89, line 43: As Claudio had recommended, can we add a few additional admins here, please?

* STS - manage access to one facility group (or one block)
* District official - view-reports-only access to one district
* Medical officer - view-all access to once facility

.env.seed.large
```diff
@@ -1,3 +1,5 @@
+SEED_ADMIN_PASSWORD="Trying to fix hypertension."
```
harimohanraj89, line 1: How about `trying to fix hypertension` to maintain consistency with review apps? Eg. Daniel may have built up some muscle memory.

lib/seed/user_seeder.rb
```diff
@@ -11,15 +13,36 @@ def initialize(config:)
...
+    private
+
+    def create_dashboard_admins
+      unless EmailAuthentication.exists?(email: "power_user@simple.org")
+        FactoryBot.create(:admin, :power_user, full_name: "Power User", email: "power_user@simple.org", password: config.admin_password)
+      end
+
+      unless EmailAuthentication.exists?(email: "cvho@simple.org")
+        fg_1, fg_2 = *FacilityGroup.take(2)
+        user = FactoryBot.create(:admin, :manager, full_name: "CVHO", email: "cvho@simple.org")
```
harimohanraj89, line 39: Let's create this with `config.admin_password` as well. It would be useful to be able to log in to these "user profiles" easily on SBX.

harimohanraj89 (changes requested):
Some suggestions from me.

lib/seed/user_seeder.rb
```diff
@@ -11,15 +13,36 @@ def initialize(config:)
...
 
     def call
+      create_dashboard_admins
+      create_mobile_users
+    end
+
+    private
+
+    def create_dashboard_admins
+      unless EmailAuthentication.exists?(email: "power_user@simple.org")
```
harimohanraj89, line 33: How about `admin@simple.org` for consistency with the existing initial power user?

config/schedule.rb
```diff
@@ -35,10 +35,6 @@ def local(time)
   runner "Reports::RegionCacheWarmer.call"
 end
 
-every :month, at: local("04:00 am"), roles: [:seed_data] do
-  rake "db:purge_and_reseed"
-end
```
harimohanraj89, line 40: I'd say keep this. Keeping the data current (seeing sensible data for the most recent month) is important.

rsanheim: The problem is it will blow away all the data and then regen, including any changes or tweaks made manually or any manual user accounts.  I think we can remove this and work on https://app.clubhouse.io/simpledotorg/story/2496/create-seed-data-since-the-last-seed-job-ran over the next month or so to get updated data functionality.

WDYT?

lib/seed/user_seeder.rb
rsanheim: I actually removed that user, and went with emails to match the access level for explicitness and consistency.

config/schedule.rb
harimohanraj89: That is ideal, I agree, as long as we can do that within the month 😄  Actually, yeah, let's keep it off. We can re-run manually if we don't get to ch2496 before March 1, and the data starts getting stale.

lib/seed/user_seeder.rb
rsanheim: We can definitely make this more idempotent, but I'd like to do that in future PRs around this.  Creating some stock users with minimal idempotency is an improvement over our existing setup where don't have any stock admin users. 😁 

Right now Sandbox seed data is empty and I want to fast track this so we can get it fixed.

lib/seed/user_seeder.rb
rsanheim: I added a bit more smarts to make sure the CVHO has access to two districts.

Looks like `grant_access` only knows about facility_ids, but we want to grant access at the facility_group level, so I went with a slightly different approach.

lib/seed/user_seeder.rb
rsanheim: Yup, added those. 👍

PR #2039, opened 2021-02-04
Add LTFU trends and patient breakdown to Reports

prabhanshuguptagit:
**Story card:** [ch2233](https://app.clubhouse.io/simpledotorg/story/2233/add-new-chart-for-all-htn-patient-breakdown-in-details-tab)

## Because
We want to show the LTFU trend and patient breakdown in the details page.

image

## This addresses
- This adds the two charts to the details page. 
- Adds a `PatientBreakdownService` which also caches the results.
- Updates the copy on region reports and moves the footnotes to tooltips.
image
- Changes the registrations card to show actual registration numbers instead of assigned patient numbers

Note: There's a big chunk of whitespace changes due to some JS refactoring, please hide them to make reviewing easier.


app/controllers/reports/regions_controller.rb
```diff
@@ -61,6 +61,12 @@ def details
     @dashboard_analytics = @region.dashboard_analytics(period: @period.type,
                                                        prev_periods: 6,
                                                        include_current_period: true)
+    @chart_data = {
+      patient_breakdown: PatientBreakdownService.call(region: @region),
+      ltfu_trend: Reports::RegionService.new(region: @region,
```
rsanheim, line 66: Looks like we should really make the LTFU data a separate piece so we don't need to retrieve all the other region related data with it. I suppose that will be a lot easier once we get https://github.com/simpledotorg/simple-server/pull/2014 merged.

app/models/reports/result.rb
```diff
@@ -184,6 +185,8 @@ def calculate_period_info(range)
         bp_control_range = period.blood_pressure_control_range
         hsh[period] = {
           name: period.to_s,
+          start_date: period.start_date.strftime(DATE_FORMAT),
+          end_date: period.end_date.strftime(DATE_FORMAT),
```
rsanheim, line 189: This will always be the same as `bp_control_end_date`, right? it makes sense to have it as a separate thing, just confirming.

app/models/reports/result.rb
```diff
@@ -198,16 +201,31 @@ def registrations_for_percentage_calculation_with_ltfu(period)
...
-        return registrations_for_percentage_calculation_with_ltfu(period)
+    def registrations_for_percentage_calculation_absolute(period)
+      if quarterly_report?
+        self[:cumulative_registrations][period.previous] || 0
+      else
+        cumulative_registrations_for(period)
       end
+    end
 
+    def registrations_for_percentage_calculation_denominator(period)
```
rsanheim, line 212: This name is a bit confusing - maybe `adjusted_registrations_for_percentage calculation` ?

app/models/reports/result.rb
```diff
@@ -198,16 +201,31 @@ def registrations_for_percentage_calculation_with_ltfu(period)
       end
     end
 
-    def registrations_for_percentage_calculation(period, key)
-      if key == :controlled_patients_with_ltfu
-        return registrations_for_percentage_calculation_with_ltfu(period)
+    def registrations_for_percentage_calculation_absolute(period)
```
rsanheim, line 204: This name is a bit confusing - maybe `cumulative_registrations_for_percentage calculation` ?

app/controllers/reports/regions_controller.rb
```diff
@@ -61,6 +61,12 @@ def details
     @dashboard_analytics = @region.dashboard_analytics(period: @period.type,
                                                        prev_periods: 6,
                                                        include_current_period: true)
+    @chart_data = {
+      patient_breakdown: PatientBreakdownService.call(region: @region),
```
rsanheim, line 65: I think we need to pass in `@period` here...

app/services/patient_breakdown_service.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def initialize(region:)
+    @region = region
+    @facilities = region.facilities
+  end
+
+  attr_reader :region
+  attr_reader :facilities
+
+  def call
```
rsanheim, line 16: Pretty sure we need to handle other possible `@period`s here, right?

rsanheim: Hm, now I'm confused:

* the actual details page does NOT have a date dropdown.
* the details page in the mockup does.

guess we should chat w/ @claudiovallejo about it.

app/views/reports/regions/details.html.erb
```diff
@@ -1,4 +1,5 @@
 <%= render "header", period_selector: false %>
+<%= render "reports/regions/patient_breakdown_charts" %>
```
rsanheim, line 2: minor thing, but can you throw a new line after each render here?

app/services/patient_breakdown_service.rb
```diff
@@ -0,0 +1,43 @@
+class PatientBreakdownService
```
rsanheim, line 1: Can you add some specs for this?

app/models/reports/result.rb
prabhanshuguptagit: Yeah I wasn't sure about the naming here. I'm thinking of refactoring the percentage calculation so we don't have to do the `if else` based on the key at all.

app/services/patient_breakdown_service.rb
prabhanshuguptagit: We decided on showing only the current period's data for the first iteration, without a datepicker. Once this is out, we can figure out a way to switch periods for the LTFU graph and figure out how to show the pie chart for previous periods (or if it is possible to do it).

app/models/reports/result.rb
prabhanshuguptagit: Yep, the ltfu graph shows patients recorded until _this_ month, so I had to add a separate `start_date` and `end_date` to show in the labels and description.

app/services/patient_breakdown_service.rb
rsanheim: Sounds good.

app/models/reports/result.rb
rsanheim: yeah these names are tough.  I think prepending the registration type (ie cumulative vs adjusted) could help if we have all these different methods going on.

I suppose another possible interface is:

```ruby
def registrations_for(period, type:)
  raise ArgumentError unless type.in?([:adjusted, :cumulative])
```

Something like that.  Not sure if that would end up better or worse.

prabhanshuguptagit: I've put the various denominators in a [hash for now](https://github.com/simpledotorg/simple-server/pull/2039/files#diff-2dd4f8a2ef5cd4e627113d439678eed40a6ad1adcc5d5de9f242dfe2e53459ddL198).

kpethtel (commented):
I'm not familiar enough with this work to feel comfortable giving an approval. But I do have a question. One thing I don't understand is why we use i18n localization for some strings and not for others, even within the same file. Do we have any guidelines for when to localize strings?

prabhanshuguptagit:
> why we use i18n localization for some strings and not for others

@kpethtel We have very recently started using locale files for the dashboard part of things. This is happening as an ongoing effort by Claudio to consolidate all our dashboard strings into one place. Anything on the admin dashboard isn't "localised" to any language though, it just leverages locale files and is all English.

We localize only app side strings to different languages. That includes:
- Progress Tab (this is rendered on the server and shown on the app as a webview)
- SMS reminders/notifications
- The android project

app/views/reports/regions/_child_data_tables.html.erb
```diff
@@ -1,277 +1,404 @@
...
+        <div class="mb-2 pt-12px">
+          <p class="us-none fs-14px c-transparent">
+            <strong>Denominator:</strong> <%= t("denominator_copy") %>
+          </p>
+        </div>
       </div>
     </div>
   </div>
-</div>
+<% end %>
```
rsanheim, line 404: This view is getting really hard to review now - can you break it out into two partials, one for each state of the feature flag?

spec/services/patient_breakdown_service_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  let!(:ltfu_patient) { create(:patient, :hypertension, recorded_at: 2.years.ago, assigned_facility: facility) }
+  let!(:not_ltfu_patient) { create(:patient, :hypertension, assigned_facility: facility) }
+  let!(:ltfu_transferred_patient) { create(:patient, :hypertension, recorded_at: 2.years.ago, status: :migrated, assigned_facility: facility) }
+  let!(:not_ltfu_transferred_patient) { create(:patient, :hypertension, status: :migrated, assigned_facility: facility) }
+  let!(:dead_patient) { create(:patient, :hypertension, status: :dead, assigned_facility: facility) }
+  let!(:not_hypertensive_patient) { create(:patient, :without_hypertension, assigned_facility: facility) }
+  let!(:blood_pressures) {
+    [create(:blood_pressure, recorded_at: 1.day.ago, patient: not_ltfu_patient),
+      create(:blood_pressure, recorded_at: 1.day.ago, patient: not_ltfu_transferred_patient)]
+  }
```
rsanheim, line 14: You can create all these models right in the `it` block itself ..there is just one test here after all.  No reason to scatter all this state into `let!`.

rsanheim (commented):
The core logic looks sound here.

I agree w/ @kpethtel that the views are getting incredibly unwieldy now...this PR is very difficult to review w/ all the feature flag branching.

rsanheim (approved):
👍 The partials definitely have helped out here w/ following what is happening in the views. Its a definite improvement until we can start to refactor more towards [view_components](https://github.com/github/view_component).

PR #2041, opened 2021-02-04
2482 my facilities dashboard stats

kpethtel:
Story card: ch2482

Because
We need facility statistics by size and month in the facilities dashboard.

This addresses


spec/services/facility_stats_service_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+  let(:supervisor) { create(:admin, :manager, :with_access, resource: organization) }
+  let(:facility_group) { create(:facility_group, organization: organization) }
+  let!(:small_facility1) { create(:facility, name: 'small1', facility_size: 'small', facility_group: facility_group) }
+  let!(:small_facility2) { create(:facility, name: 'small2', facility_size: 'small', facility_group: facility_group) }
+  let!(:medium_facility1) { create(:facility, name: 'medium1', facility_size: 'medium', facility_group: facility_group) }
+  let!(:medium_facility2) { create(:facility, name: 'medium2', facility_size: 'medium', facility_group: facility_group) }
+  let!(:large_facility1) { create(:facility, name: 'large1', facility_size: 'large', facility_group: facility_group) }
+  let!(:large_facility2) { create(:facility, name: 'large2', facility_size: 'large', facility_group: facility_group) }
+  let(:all_facilities) { [small_facility1, small_facility2, medium_facility1, medium_facility2, large_facility1, large_facility2] }
+  let(:period) { Period.month(today) }
```
rsanheim, line 19: This is why the test data is getting returned - our Periods always start at the beginning of months (or quarters).  The middle of the month date is throwing things off. - 

Changing this to `let(:period) { Reports::RegionService.default_period }` should get you data coming back.

spec/services/facility_stats_service_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      large_controlled.each { |patient| create(:blood_pressure, :under_control, patient: patient, facility: patient.assigned_facility) }
+    end
+    Timecop.freeze(four_months_ago) do
+      medium_controlled = create_list(:patient, 2, full_name: 'medium_uncontrolled', assigned_facility: medium_facility1)
+
+      medium_controlled.each { |patient| create(:blood_pressure, :under_control, patient: patient, facility: patient.assigned_facility) }
+      small_controlled.each { |patient| create(:blood_pressure, :under_control, patient: patient, facility: patient.assigned_facility, user: supervisor) }
+      small_uncontrolled.each { |patient| create(:blood_pressure, :hypertensive, patient: patient, facility: patient.assigned_facility, user: supervisor) }
+      large_uncontrolled.each { |patient| create(:blood_pressure, :hypertensive, patient: patient, facility: patient.assigned_facility) }
+    end
```
rsanheim, line 48: This is a _ton_ of test data that most of the tests don't use - I would recommend creating just the test data you need inside the specific specs. It will make for more clear specs and will also be much faster.

rsanheim (commented):
Found the issue and left a comment.

Also, can you start using standardrb to lint things ? It will make the code easier to review if its consistent w/ our style.

It will auto fix most things for you:

```
gem install standardrb
standardrb --fix
```

app/controllers/my_facilities_controller.rb
```diff
@@ -113,4 +92,13 @@ def force_cache?
   def report_with_exclusions?
     current_admin.feature_enabled?(:report_with_exclusions)
   end
+
+  def process_facility_stats(facilities, type)
+    stats_service = FacilityStatsService.new(accessible_facilities: @accessible_facilities, retain_facilities: facilities,
```
rsanheim, line 97: What is the `retain_facilities` argument for here?  Is that really the `filtered_facilities` ?  Or maybe `selected_facilities` ?

app/controllers/my_facilities_controller.rb
```diff
@@ -36,39 +36,18 @@ def index
...
-
-    @data_for_facility = {}
-
-    @facilities.each do |facility|
-      @data_for_facility[facility.name] = Reports::RegionService.new(region: facility, period: @period).call
-    end
-
-    @facilities_by_size = @facilities.group_by { |facility| facility.facility_size }
+    facilities = filter_facilities
+    process_facility_stats(facilities, "controlled_patients")
```
rsanheim, line 40: One overall piece of feedback: I think the new service is doing too much and too tightly coupled to the RegionsService.
My suggestion - change your new service to take the `facility_data` from RegionsService and just do the math on the data...let the calling code worry about how to retrieve that data.  So for the `bp_controlled` action for example:

```ruby
def bp_controlled
  @data_for_facility = {}		
  @facilities.each do |facility|		
     @data_for_facility[facility.name] = Reports::RegionService.new(region: facility, period: @period).call		
  end
  @stats_by_size = FacilityStatsService.new(@data_for_facility)
end
```

This kind of approach would hopefully make it easier to rip out or change the underlying implementation of RegionsService, which is coming 🔜 with https://github.com/simpledotorg/simple-server/pull/2014.

app/services/facility_stats_service.rb
```diff
@@ -0,0 +1,71 @@
...
+    @rate_numerator = rate_numerator
+    @rate_name = "#{rate_numerator}_rate"
+    @periods = ending_period.downto(5)
+    @facilities_data = {}
+    @stats_by_size = {}
+  end
+
+  def call
+    facilities.each do |facility|
+      facility_data = Reports::RegionService.new(region: facility, period: ending_period).call
```
rsanheim, line 17: With the above proposal, this call would go away and would become input to the service object.

app/services/facility_stats_service.rb
```diff
@@ -0,0 +1,71 @@
...
+    retain_facilities.find { |rf| rf.name == facility.name }
+  end
+
+  def add_size_section(size)
+    stats_by_size[size] = size_data_template
+  end
+
+  def size_data_template
+    periods.reverse.each_with_object({}) do |period, hsh|
+      hsh[period] = month_data_template
```
rsanheim, line 59: Inlining this would be more clear I think, so you could see the shape of the data in one method.

app/services/facility_stats_service.rb
```diff
@@ -0,0 +1,71 @@
...
+    add_size_section(size) unless stats_by_size[size]
+    periods.each do |period|
+      current_period = stats_by_size[size][period]
+      current_period[rate_numerator] += facility_data.dig(rate_numerator, period) || 0
+      current_period["adjusted_registrations"] += facility_data["adjusted_registrations"][period]
+      current_period["cumulative_registrations"] += facility_data["cumulative_registrations"][period]
+    end
+  end
+
+  def add_control_rates
```
rsanheim, line 39: Is this always control rates?  I think sometimes it can be `missed_visits` ...so maybe `calculate_percentages` ?

spec/services/facility_stats_service_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+      medium_facilities = setup_for_size("medium", december - 4.months)
+      large_facilities = setup_for_size("large", december - 3.months)
+      refresh_views
+      all_facilities = small_facilities + medium_facilities + large_facilities
+      subject = FacilityStatsService.new(accessible_facilities: all_facilities, retain_facilities: all_facilities,
+                                         ending_period: period, rate_numerator: "controlled_patients")
+      subject.call
+
+      # all numbers except cumulative_registrations appear in data 3 months after they're recorded
+      expect(stats_of_type(subject.stats_by_size["small"], "controlled_patients")).to eq [0, 0, 0, 2, 0, 0]
```
rsanheim, line 79: Would `subject.stats_by_size["small"].values` handle this for you? I don't think you need the helper method.

spec/services/facility_stats_service_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+      accessible_facilities = small_facilities + large_facilities
+      subject = FacilityStatsService.new(accessible_facilities: accessible_facilities, retain_facilities: large_facilities,
+                                         ending_period: period, rate_numerator: "controlled_patients")
+      subject.call
+      expect(subject.stats_by_size.keys).to match_array(["small", "large"])
+      expect(subject.facilities_data.keys).to match_array(large_facilities.map(&:name))
+    end
+
+    it "processes data for controlled_patients" do
+      subject = FacilityStatsService.new(accessible_facilities: [small_facility1], retain_facilities: [small_facility1],
```
rsanheim, line 107: I would suggest adding a class `call` method that forwards args to the initializer, then you can just do 

```
result = FacilityStatsService.call(accessible_facilities: [small_facility1], retain_facilities: [small_facility1]....)
```

We follow that type of pattern quite a bit, see https://github.com/simpledotorg/simple-server/blob/8014f0f5da0187a66202a1d3475f642db0357a0a/app/services/reports/region_service.rb#L10-L12 for example.

app/controllers/my_facilities_controller.rb
```diff
@@ -113,4 +92,13 @@ def force_cache?
   def report_with_exclusions?
     current_admin.feature_enabled?(:report_with_exclusions)
   end
+
+  def process_facility_stats(facilities, type)
+    stats_service = FacilityStatsService.new(accessible_facilities: @accessible_facilities, retain_facilities: facilities,
+                                             ending_period: @period, rate_numerator: type)
+    stats_service.call
+    @data_for_facility = stats_service.facilities_data
+    @stats_by_size = stats_service.stats_by_size
```
rsanheim, line 101: If the FacilityStatsService just returns the size aggregation, then it will only have one return value and these pieces will be more clear.

app/controllers/my_facilities_controller.rb
```diff
@@ -36,39 +36,18 @@ def index
...
-
-    @data_for_facility = {}
-
-    @facilities.each do |facility|
-      @data_for_facility[facility.name] = Reports::RegionService.new(region: facility, period: @period).call
-    end
-
-    @facilities_by_size = @facilities.group_by { |facility| facility.facility_size }
+    facilities = filter_facilities
+    process_facility_stats(facilities, "uncontrolled_patients")
```
rsanheim, line 45: The numerator arg should probably be a symbol - `:uncontrolled_patients` for example.

rsanheim (commented):
Looking good for a first pass ! 👍  Left some feedback, lemme know if you have questions.

app/services/facility_stats_service.rb
kpethtel: This needs to be renamed. My first pass through this I was working with control rates and named some things too specifically. Thanks for catching this.

spec/services/facility_stats_service_spec.rb
kpethtel: Yeah, that's a common pattern in ruby. Sounds good.

app/controllers/my_facilities_controller.rb
kpethtel: Yes, either filtered_facilities or selected_facilities would work. The idea of fetching general stats for all accessible facilities but returning specific data for only visible facilities was a change of scope that occurred on Thursday and I shoehorned it in a bit.

app/controllers/my_facilities_controller.rb
kpethtel: Will do.

spec/services/facility_stats_service_spec.rb
kpethtel: Unfortunately, that doesn't quite work because the shape of the data is 
```
{ size: { period: { 'controlled_patients': 0, ...}}}
```
But I've removed the helper anyhow. It wasn't really improving readability.

app/controllers/my_facilities_controller.rb
```diff
@@ -113,4 +89,17 @@ def force_cache?
...
   end
+
+  def process_facility_stats(type)
+    display_facilities = filter_facilities
+    @data_for_facility = {}
+    data_for_stats = {}
+    @accessible_facilities.each do |facility|
+      result = Reports::RegionService.new(region: facility, period: @period).call
+      @data_for_facility[facility.name] = result if display_facilities.include?(facility)
+      data_for_stats[facility.name] = result
```
rsanheim, line 100: Okay, now I understand the change here wrt filtering and such.

This is going to add a **substantial** amount of processing time. Consider:

#### Before
* we find what facilities a user has `view_reports` access to, apply the filters from the UI (for size, district, etc), and then grab numbers for that subset via RegionService

#### After
* we find what facilities a user has `view_reports` access to, grab numbers for **all** of them via RegionService, and then apply filters for things we want to display in the view

So for a power user, this could be going from ~100/200 calls to RegionService to well over a thousand. I would expect this to be a problem with a production-sized dataset.

spec/services/facility_stats_service_spec.rb
```diff
@@ -0,0 +1,119 @@
+require "rails_helper"
+
+RSpec.describe FacilityStatsService do
+  let(:small_facility) { create(:facility, name: "small1", facility_size: "small") }
+  let(:december) { Date.parse("1-12-2020").beginning_of_month }
```
rsanheim, line 5: You are calling this let var `december`, but the period is for January.  

I would recommend `Date.parse("12-01-2019")` for this if you really want December.

kpethtel: Whoops.

app/controllers/my_facilities_controller.rb
kpethtel: Yeah, this change was requested late last week. I may have made a bad assumption about what was being queried when you go to http://localhost:3000/my_facilities/bp_controlled. I looked at how filtered facilities were being set, but I should have done some manual testing and asked more questions. The default filtered facilities uses some fallbacks, like using the first facility group instead of all facility groups, which significantly decreases the scope. I'll talk to Claudio about which facilities we really care about. It may still be all accessible and you may be right that it could be a problem. Maybe we only care about all facilities within the current group.

rsanheim: > Maybe we only care about all facilities within the current group.

👍 Yeah that sounds like it a more reasonable group to grab the data for.

Any page that grabs report data for _all_ facilities is going to have a bad time. We don't have good summary tables or caching strategies in place for that sort of thing (yet).

app/controllers/concerns/my_facilities_filtering.rb
```diff
@@ -18,6 +18,10 @@ def filter_facilities
       facilities_by_size(filtered_facilities)
     end
 
+    def facilities_for_group
+      facilities_by_facility_group(@accessible_facilities)
```
rsanheim, line 22: Instead of introducing a new method, can you call this directly inline in the controller?  This method isn't really carrying its weight ...and long term I think we should be moving away from this module based filtering approach.  

Something like this in the controller:

```
facilities_for_totals = facilities_by_facility_group(@accessible_facilities)
```

This would make it more clear that we are constructing totals based on a superset of what is being filtered and shown in the view.

app/services/facility_stats_service.rb
```diff
@@ -0,0 +1,62 @@
+class FacilityStatsService
+  attr_reader :stats_by_size
+
+  def initialize(facilities:, ending_period:, rate_numerator:)
```
rsanheim, line 4: Just `period` is fine here for this argument name as opposed to `ending_period` - it matches the conventions we have elsewhere also.

spec/controllers/my_facilities_controller_spec.rb
```diff
@@ -7,6 +7,8 @@
 RSpec.describe MyFacilitiesController, type: :controller do
   let(:facility_group) { create(:facility_group) }
   let(:supervisor) { create(:admin, :manager, :with_access, resource: facility_group) }
+  let(:facility) { create(:facility, facility_group: facility_group) }
+  let(:december) { Period.month("December 2020") }
```
rsanheim, line 11: 👍

rsanheim (approved):
Looks good - just some minor feedback here.

I'm going to approve to unblock you - please feel free to merge once you are happy with things and tests pass / etc.

app/controllers/concerns/my_facilities_filtering.rb
kpethtel: Yeah, sounds good.

PR #2043, opened 2021-02-05
Fix child tables reports

prabhanshuguptagit:
**Story card:** [ch2503](https://app.clubhouse.io/simpledotorg/story/2503)

## Because

We found that reports for a district were breaking in IHCI. [Sentry reference](https://sentry.io/organizations/resolve-to-save-lives/issues/2190893958/?referrer=slack)

## This addresses
This happens because some facilities in the district have a different name from their regions. Specifically, these facilities have spaces in the name which were created before we added `auto_squish` to any models.

This breaks the reports because:
- The cache warmer caches `RegionService` [by facility](https://github.com/simpledotorg/simple-server/blob/7a0167ef36c77595d779390b77bff3f7b4eedf15/app/services/reports/region_cache_warmer.rb#L76), but the controller fetches reports by regions.
- A facility and it's region have the [same cache key](https://github.com/simpledotorg/simple-server/blob/7a0167ef36c77595d779390b77bff3f7b4eedf15/app/models/facility.rb#L104).
- When we request a `Result` object for a facility's region, we actually get the result object for the facility.
- The view however tries to iterate over regions and does not find any data if the names of the facility and it's region are different.

This skips putting results into a hash by the region's name and then looking them up in the view. This ensures that we display whatever results are fetched (from DB or cached) as is.

### DIFF
--- a/app/controllers/reports/regions_controller.rb
+++ b/app/controllers/reports/regions_controller.rb
@@ -46,10 +46,10 @@ class Reports::RegionsController < AdminController
     @adjusted_registration_date = @data[:adjusted_registrations].keys[-4]
 
     @children = @region.reportable_children(region_reports_enabled: region_reports_enabled?)
-    @children_data = @children.each_with_object({}) { |child, hsh|
-      hsh[child.name] = Reports::RegionService.new(region: child,
-                                                   period: @period,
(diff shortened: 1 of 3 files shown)


vkrmis (approved):
lgtm

PR #2045, opened 2021-02-05
Seed script fixes - Protocols and Organizations

harimohanraj89:
**Story card:** [ch2543](https://app.clubhouse.io/simpledotorg/story/2543/create-protocols-and-protocol-drugs)

## Because

We were not creating any protocols, which would cause syncs to fail.

## This addresses

* Seeds a protocol with standard HTN drugs for now
* Also addresses a rogue factory that created throwaway organizations unnecessarily.

### DIFF
--- a/lib/seed/facility_seeder.rb
+++ b/lib/seed/facility_seeder.rb
@@ -68,7 +68,7 @@ module Seed
       end
 
       announce "Creating protocol and protocol drugs..."
-      protocol = Seed::ProtocolSeeder.call
+      protocol = Seed::ProtocolSeeder.call(config: config)
 
       announce "Creating #{number_of_facility_groups} FacilityGroups..."
 
--- a/lib/seed/facility_seeder.rb
+++ b/lib/seed/facility_seeder.rb
@@ -37,6 +37,27 @@ module Seed
 
     delegate :scale_factor, :stdout, to: :config
 
+    def call
+      Region.root || Region.create!(name: "India", region_type: Region.region_types[:root], path: "india")
+
+      if number_of_facility_groups <= FacilityGroup.count
+        announce "Not creating FacilityGroups or Facilities, we already have max # (#{number_of_facility_groups}) of FacilityGroups"
+        return
+      end
+
+      announce "Creating protocol and protocol drugs..."
+      protocol = Seed::ProtocolSeeder.call(config: config)
+
+      announce "Creating #{number_of_facility_groups} FacilityGroups..."
+
+      state_results = create_state_regions
+      facility_group_results = create_facility_groups(protocol)
+      district_region_results = create_district_regions(state_results, facility_group_results)
+      create_block_regions(district_region_results)
+      facility_results = create_facilities(facility_group_results)
+      create_facility_regions(facility_results)
+    end
+
     def number_of_facility_groups
       config.number_of_facility_groups
     end
@@ -58,28 +79,19 @@ module Seed
       config.rand_or_max(1..max_number_facilities_per_block)
     end
 
-    def call
-      Region.root || Region.create!(name: "India", region_type: Region.region_types[:root], path: "india")
-      organization = Seed.seed_org
+    def organization
+      @organization ||= Seed.seed_org
+    end
 
-      if number_of_facility_groups <= FacilityGroup.count
-        announce "Not creating FacilityGroups or Facilities, we already have max # (#{number_of_facility_groups}) of FacilityGroups"
-        return
-      end
-
-      announce "Creating protocol and protocol drugs..."
-      protocol = Seed::ProtocolSeeder.call(config: config)
-
-      announce "Creating #{number_of_facility_groups} FacilityGroups..."
-
-      # Create state Regions
+    def create_state_regions
       state_names = Seed::FakeNames.instance.states.sample(number_of_states)
       states = number_of_states.times.each_with_index.map { |i|
         FactoryBot.build(:region, :state, id: nil, name: state_names[i], parent_path: organization.region.path)
       }
-      state_results = Region.import(states, returning: [:path])
+      Region.import(states, returning: [:path])
+    end
 
-      # Create FacilityGroups
+    def create_facility_groups(protocol)
       district_names = Seed::FakeNames.instance.districts.sample(number_of_facility_groups)
       facility_groups = number_of_facility_groups.times.each_with_index.map { |i|
         FactoryBot.build(:facility_group,
@@ -91,10 +103,11 @@ module Seed
           protocol_id: protocol.id,
           state: nil)
       }
-      fg_result = FacilityGroup.import(facility_groups, returning: [:id, :name], on_duplicate_key_ignore: true)
+      FacilityGroup.import(facility_groups, returning: [:id, :name], on_duplicate_key_ignore: true)
+    end
 
-      # Create district Regions
-      district_regions = fg_result.results.map { |row|
+    def create_district_regions(state_results, facility_group_results)
+      district_regions = facility_group_results.results.map { |row|
         facility_group_id, facility_group_name = *row
         attrs = {
           id: nil,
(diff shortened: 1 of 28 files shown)


lib/seed/protocol_seeder.rb
```diff
@@ -0,0 +1,49 @@
...
+      {name: "Chlorthalidone", dosage: "12.5 mg", rxnorm_code: "331132", drug_category: :hypertension_diuretic, stock_tracked: true}
+    ].freeze
+
+    def call
+      announce "Creating #{protocol_name}..."
+      protocol = Protocol.create!(name: protocol_name, follow_up_days: 28)
+
+      announce "Creating protocol drugs for #{protocol_name}..."
+      PROTOCOL_DRUGS.each do |attributes|
+        ProtocolDrug.create!(**attributes, protocol: protocol)
```
ssrihari, line 39: We could import the above list of hashes directly, I think: https://github.com/zdennis/activerecord-import#hashes. Not that it's particularly important for performance, but it's more consistent.

lib/seed/facility_seeder.rb
```diff
@@ -84,6 +88,7 @@ def call
           generating_seed_data: true,
           name: district_names[i],
           organization_id: organization.id,
+          protocol_id: protocol.id,
           state: nil)
       }
```
ssrihari, line 93: While we're here, and if we have the time, can we refactor the call method so it's easier to read? Just extract functions where we have comments right now, even?

rsanheim:
> We were not creating any protocols, which would cause syncs to fail.

I'm confused as to why the non-existence of a model like this would cause sync failure.

Is that because of a bug in the server or the client?  What would it take to make our sync more robust for this sort of case?

spec/factories/users.rb
```diff
@@ -69,7 +69,6 @@
     device_updated_at { Time.current }
     sync_approval_status { User.sync_approval_statuses[:denied] }
     email_authentications { build_list(:email_authentication, 1, email: email, password: password) }
-    organization
```
rsanheim, line 72: Ah, good catch, thanks for this.

rsanheim (commented):
Adding some more spec coverage around the org / seed issue in https://github.com/simpledotorg/simple-server/pull/2046

lib/seed/facility_seeder.rb
harimohanraj89: I like this idea. I'll do this now.

spec/factories/users.rb
harimohanraj89: Ooh neat, and associated specs too. I'll merge `master`, thanks!

lib/seed/protocol_seeder.rb
ssrihari: Resolved [elsewhere](https://simpledotorg.slack.com/archives/CFHMC60P5/p1612531890243700?thread_ts=1612524592.239700&cid=CFHMC60P5).

lib/seed/facility_seeder.rb
harimohanraj89: I just took a stab at this, and there's enough local temporary info (like `state_results`) that are being passed around that I think it's ok to leave as is. The alternative is passing these bits of info into the extracted methods, or saving them in state. Both seem a bit awkward to me.

What do you think?

ssrihari: If it's not easy to do now with this PR we can leave it for later I guess. I don't mind passing around some variables, I could also take a stab at this separately.

spec/lib/seed/facility_seeder_spec.rb
```diff
@@ -1,6 +1,12 @@
 require "rails_helper"
 
 RSpec.describe Seed::FacilitySeeder do
+  it "creates a protocol" do
```
kpethtel, line 4: Could also be worth confirming that a Protocol has been created in the db, similar to the facilities/groups test immediately following this.

harimohanraj89: The protocol creation in the DB is tested in the `protocol_seeder_spec`, so this spec just tests the right collaboration with that class. I usually take this approach to avoid double-testing the same implementations over and over. What do you think?

kpethtel: Sounds good to me. I'm reviewing this with little context, and I generally will agree with you about testing boundaries.

harimohanraj89:
@rsanheim 

> I'm confused as to why the non-existence of a model like this would cause sync failure.
> Is that because of a bug in the server or the client? What would it take to make our sync more robust for this sort of case?

Yup, the underlying cause is: Simple Server's data constraints and validation don't match Simple App's. To be super robust, we'd make protocol a required attribute of facility groups, ensure that all existing facility groups in prod have a protocol, and ship the change. This is, of course, a much more heavyhanded approach.

harimohanraj89:
@ssrihari I took a stab at extracting the FacilitySeeder sections into methods. Lemme know if this is what you had in mind, thanks!

lib/seed/facility_seeder.rb
ssrihari: The changes look good @harimohanraj89!

spec/lib/seed/facility_seeder_spec.rb
harimohanraj89: No worries at all, keep em coming 👌

spec/lib/seed/protocol_seeder_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    Seed::ProtocolSeeder.call(config: Seed::Config.new)
+
+    protocol = Protocol.find_by(name: "#{org_name} Protocol")
+    expect(protocol).to be_present
+    expect(protocol.protocol_drugs.pluck(:name)).to include(
+      "Amlodipine",
+      "Telmisartan",
+      "Losartan",
+      "Hydrochlorothiazide",
+      "Chlorthalidone"
```
rsanheim, line 15: 👍

lib/seed/protocol_seeder.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    PROTOCOL_DRUGS = [
+      {name: "Amlodipine", dosage: "5 mg", rxnorm_code: "329528", drug_category: :hypertension_ccb, stock_tracked: true},
+      {name: "Amlodipine", dosage: "10 mg", rxnorm_code: "329526", drug_category: :hypertension_ccb, stock_tracked: true},
+      {name: "Telmisartan", dosage: "40 mg", rxnorm_code: "316764", drug_category: :hypertension_arb, stock_tracked: true},
+      {name: "Telmisartan", dosage: "80 mg", rxnorm_code: "316765", drug_category: :hypertension_arb, stock_tracked: true},
+      {name: "Losartan", dosage: "50 mg", rxnorm_code: "979467", drug_category: :hypertension_arb, stock_tracked: true},
+      {name: "Hydrochlorothiazide", dosage: "12.5 mg", rxnorm_code: "316047", drug_category: :hypertension_diuretic, stock_tracked: true},
+      {name: "Hydrochlorothiazide", dosage: "25 mg", rxnorm_code: "316049", drug_category: :hypertension_diuretic, stock_tracked: true},
+      {name: "Chlorthalidone", dosage: "12.5 mg", rxnorm_code: "331132", drug_category: :hypertension_diuretic, stock_tracked: true}
```
rsanheim, line 30: Do you think these will remain pretty static over time? 

I'm wondering if we should extract to YAML or something if that happens ...lots of magic numbers / names here.

rsanheim (approved):
Dig the refactoring, 🚢 !

lib/seed/protocol_seeder.rb
harimohanraj89: They should be pretty static, the medication protocols in India are fairly solid. I'd be ok leaving as is.

ssrihari: These should be pretty static over years, but it's probably still worth extracting them into a YAML so it's a declarative reference for us through time.

PR #2046, opened 2021-02-05
Ensure we dont create unnecessary orgs in User seeds

rsanheim:
**Story card:** [ch2543](https://app.clubhouse.io/simpledotorg/story/2543/create-protocols-and-protocol-drugs)

Adding some more specs around a bug @harimohanraj89 found in https://github.com/simpledotorg/simple-server/pull/2045.

I _think_ the user#organization association can go away completely, since we now can determine org relationship via user_access....but we probably want to make sure all these users are still associated with the one seed org until that happens.

### DIFF
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -68,7 +68,6 @@ FactoryBot.define do
     device_updated_at { Time.current }
     sync_approval_status { User.sync_approval_statuses[:denied] }
     email_authentications { build_list(:email_authentication, 1, email: email, password: password) }
-    # organization
     role { "power user" }
     access_level { :power_user }
 
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -7,7 +7,6 @@ FactoryBot.define do
     end
 
     full_name { Faker::Name.name }
-    organization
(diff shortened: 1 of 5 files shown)


spec/lib/seed/runner_spec.rb
```diff
@@ -20,6 +20,7 @@
       .and change { BloodPressure.count }.by(expected_bps)
       .and change { Encounter.count }.by(expected_bps)
       .and change { Observation.count }.by(expected_bps)
+      .and change { Organization.count }.by(1)
```
rsanheim, line 23: This ensures the whole seed process only creates the one main seed org.

spec/factories/users.rb
```diff
@@ -69,7 +68,6 @@
     device_updated_at { Time.current }
     sync_approval_status { User.sync_approval_statuses[:denied] }
     email_authentications { build_list(:email_authentication, 1, email: email, password: password) }
-    organization
```
rsanheim, line 72: These removals will result in less unnecessary orgs getting created throughout all our tests...I bet it shaves some time off CI.

PR #2047, opened 2021-02-05
Make region_reports the mainline code and remove feature flag

rsanheim:
**Story card:** [ch2550](https://app.clubhouse.io/simpledotorg/story/2550/remove-region-reports-feature-flag)

Region reports is shipped everywhere, so we can make that feature the mainline code.

### DIFF
--- a/app/services/reports/region_cache_warmer.rb
+++ b/app/services/reports/region_cache_warmer.rb
@@ -31,7 +31,6 @@ module Reports
             Statsd.instance.increment("region_cache_warmer.states.cache")
           end
         end
-
       }
       notify "finished", duration: duration
     ensure
--- a/app/services/reports/region_cache_warmer.rb
+++ b/app/services/reports/region_cache_warmer.rb
@@ -24,25 +24,14 @@ module Reports
           return
         end
 
-        notify "starting facility_group caching"
+        notify "starting region caching"
         Statsd.instance.time("region_cache_warmer.states") do
-          cache_states
+          Region.where.not(region_type: ["root", "organization"]).find_each(batch_size: BATCH_SIZE) do |region|
+            RegionService.call(region: region, period: period)
+            Statsd.instance.increment("region_cache_warmer.states.cache")
+          end
         end
 
-        notify "starting facility_group caching"
-        Statsd.instance.time("region_cache_warmer.facility_groups") do
-          cache_facility_groups
-        end
-
-        notify "starting block caching"
-        Statsd.instance.time("region_cache_warmer.blocks") do
-          cache_blocks
-        end
-
-        notify "starting facility caching"
-        Statsd.instance.time("region_cache_warmer.facilities") do
-          cache_facilities
-        end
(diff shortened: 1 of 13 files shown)


app/services/reports/region_cache_warmer.rb
```diff
@@ -24,24 +24,12 @@ def call
...
-
-        notify "starting block caching"
-        Statsd.instance.time("region_cache_warmer.blocks") do
-          cache_blocks
-        end
-
-        notify "starting facility caching"
-        Statsd.instance.time("region_cache_warmer.facilities") do
-          cache_facilities
+          Region.where.not(region_type: ["root", "organization"]).find_each(batch_size: BATCH_SIZE) do |region|
```
prabhanshuguptagit, line 29: ✨

app/views/admin/facility_groups/_form.html.erb
```diff
@@ -1,6 +1,11 @@
 <div class="row">
   <div class="col-md-6">
     <%= bootstrap_form_with(model: [:admin, facility_group], local: true, autocomplete: "off", label_errors: true, html: {id: "facility-group-form"}) do |form| %>
+      <% if @facility_group.errors.any? %>
+        <div id="errors">
+          <%= form.label "Saving the FacilityGroup failed, please fix the following errors: #{@facility_group.errors.full_messages.to_sentence}", class: "text-danger" %>
```
prabhanshuguptagit, line 6: 👾  
```suggestion
          <%= form.label "Saving the Facility Group failed, please fix the following errors: #{@facility_group.errors.full_messages.to_sentence}", class: "text-danger" %>
```

PS What changed about this page?

app/views/reports/regions/index.html.erb
```diff
@@ -2,50 +2,29 @@
...
-              <div class="access-tree__section-header">
-                <%= org.name %>
-              </div>
-
-              <%= render Reports::RegionTreeComponent.new(parent: org, children: children) %>
+      <div id="facility-access">
+        <% @accessible_regions.each do |org, children| %>
+          <div class="card organization show region-index">
+            <div class="access-tree__section-header">
+              <%= org.name %>
```
prabhanshuguptagit, line 15: What will this look like in Bangladesh?

prabhanshuguptagit (approved):
Looks great👌🏼  Approving to unblock.

kpethtel (approved):
👍

app/views/admin/facility_groups/_form.html.erb
rsanheim: It wasn't showing errors properly when the Facility Group failed saving...I wanted to get the error message into the view _somehow_, and this was the simplest way to do it.

app/views/reports/regions/index.html.erb
rsanheim: It should be the same, because we restrict to reportable children, which only returns district and facility there.

PR #2053, opened 2021-02-10
My Facilities: Facility Size Subtotals UI

claudiovallejo:
**Story card:** [ch2483](https://app.clubhouse.io/simpledotorg/story/2483/create-ui-to-show-facility-size-trends?vc_group_by=day)

## Because

Program managers need to see 6-month trend data for all facilities of the same size. This helps better evaluate overall performance between facilities.

## This addresses

* Adds a new section that displays the current rate (and 6-month change) for all facilities within a facility size
* Adds a new "total" row to each facility size table
* Both of these features are fully responsive
* **Bug:** Fixed breaking title + "last updated" copy on mobile (on mobile they pile up, on desktop they float to opposite ends)

**Mobile screenshot**
image

**Desktop screenshot**
image


app/views/my_facilities/bp_controlled.html.erb
```diff
@@ -1,122 +1,167 @@
...
-        <strong>Numerator:</strong> <%= t("numerator_copy.bp_controlled") %>, <strong>Denominator:</strong> <%= t("denominator_copy") %>
-      </p>
-      <div class="d-flex fw-wrap mb-16px">
+      <div class="d-flex fw-wrap mb-16px ph-16px">
         <%= render "shared/my_facilities_filters" %>
       </div>
-      <% if @facility_sizes.empty? %>
-        <%= render "shared/my_facilities_table_empty_state" %>
-      <% end %>
-      <% @display_sizes.each do |size| %>
```
kpethtel, line 24: I believe changing this from display_sizes to facility_sizes has created a bug. If you select a size from the right-most dropdown, you see all sizes in the facility group.

kpethtel (changes requested):
Let me know if you have questions about the bug I commented on. It should be as simple as changing to display_sizes. The facility_sizes object is created in the MyFacilitiesFiltering module and is set to all sizes within the current facility group and zone, which is why I had to create the display_sizes object.

kpethtel (approved):
👍

PR #2055, opened 2021-02-10
Ethiopia needs state reports

harimohanraj89:
Quickfix for [ch2595](https://app.clubhouse.io/simpledotorg/story/2595/reports-tree-bugfix)

See detailed problem in story.


vkrmis (approved):

rsanheim:
I'm guessing this was intended as quick fix?  This would introduce all the redundant regions in ET that I think we don't displayed.

I like the approach you outlined in the clubhouse card...I don't think it would be much more work:

> stop rendering state/block links if extended region reports is off

PR #2056, opened 2021-02-10
2482 add total registrations

kpethtel:
**Story card:** ch2482

## Because

We need to expose the number of hypertensive patients within the facility group. 

## Comments

This branch was branched off of another branch that's been merged to master. I don't think git will have any issues squashing this neatly. Also, many of the changes in this PR are style changes, like moving away from strings as hash key accessors.


spec/services/facility_stats_service_spec.rb
```diff
@@ -5,23 +5,6 @@
   let(:december) { Date.parse("12-01-2020").beginning_of_month }
   let(:period) { Period.month(december) }
 
-  def setup_for_size(size, month = december)
```
kpethtel, line 8: I'm removing this for a few reasons. 
1) As I get more accustomed to our communal style, I see that we mostly set up our tests within individual tests rather than setting up complex structures for testing. 
2) This was too rigid and didn't allow enough variation.
3) This doesn't work with the tests I needed to set up for testing the registrations count, so this was only being used in one test.

spec/services/facility_stats_service_spec.rb
```diff
@@ -39,81 +22,136 @@ def facilities_data(facilities)
...
         periods << periods.last.previous
       }
       expect(small.keys).to match_array(periods)
     end
 
     it "accurately tallies stats for facilities by size and period" do
-      small_facilities = setup_for_size("small", december - 5.months)
-      medium_facilities = setup_for_size("medium", december - 4.months)
-      large_facilities = setup_for_size("large", december - 3.months)
+      small_facility1 = create(:facility, name: "small_1", facility_size: "small")
```
kpethtel, line 40: In addition to moving the data setup out of the helper method, I've also added some more variety of data (i.e., different numbers of patients for each facility size).

spec/services/facility_stats_service_spec.rb
```diff
@@ -39,81 +22,136 @@ def facilities_data(facilities)
...
       expect(stat_keys).to match_array(expected_keys)
       expect(stat_values).to match_array([0, 0, 0, 0])
     end
+
+    it "sets total_registered_hypertensive_patients for each size" do
+      small_facility
+      month = december - 4.months
+      medium_facility1 = create(:facility, name: "medium_1", facility_size: "medium")
+      medium_facility2 = create(:facility, name: "medium_2", facility_size: "medium")
+      create_list(:patient, 2, :hypertension, full_name: "medium_uncontrolled",
```
kpethtel, line 144: The biggest challenge I faced in this ticket was figuring out how to get patient data that showed up in the `facility.registered_hypertension_patients`. For reasons not clear to me yet, setting up the test data like I did above (make facility, make patients for facility, add blood pressures for patients) results in `facility.registered_hypertension_patients` being empty. Conversely, setting up patients with traits as I've done here results in the patient data not appearing in the RegionService controlled/uncontrolled results (I believe this was true even if I also created blood pressure readings). The problem may be my lack of understanding of how that data gets into the controlled/uncontrolled results, but I suspect there's something missing with the factories. Ideally, I would prefer to use these factories with traits, because they set up a medical history, which results in less tedious data setup (no need to create BP readings a month after setting up the patient).

rsanheim (approved):
Heya pushed up a spec refactoring to help with some speed with a more explicit object graph. It was easier to show with code rather than doing another round of PR feedback for that, hope that is okay.

Overall things look good here, ship away when ready 👍 .

app/services/facility_stats_service.rb
```diff
@@ -28,25 +31,32 @@ def add_facility_stats(facility_data)
     size = facility_data.region.source.facility_size
     add_size_section(size) unless stats_by_size[size]
     periods.each do |period|
-      current_period = stats_by_size[size][period]
+      current_period = stats_by_size[size][:periods][period]
       current_period[rate_numerator] += facility_data.dig(rate_numerator, period) || 0
-      current_period["adjusted_registrations"] += facility_data["adjusted_registrations"][period]
+      current_period[:adjusted_registrations] += facility_data[:adjusted_registrations][period]
```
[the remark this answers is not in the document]
kpethtel: The totals are already available from the data_for_facility, so I ended up not adding them to this PR at all. The plan has changed a bit since this PR was started. It was originally about adding those totals, which I had wrongly thought needed to be aggregated by size. Now this PR is more about removing unnecessary data fetching. The previous plan was to show stats for all sizes, but now we only want stats for the currently displayed sizes. So I've removed the extra data fetching. Also, I obviously rewrote a bunch of specs. This is not ideal, but this project is a bit in the air right now.

PR #2057, opened 2021-02-10
Handle nil period bug when a user requests reports before any registrations in a Region

rsanheim:
**Story card:** [ch1784](https://app.clubhouse.io/simpledotorg/story/1784/nil-error-in-some-reports)

### DIFF
--- a/app/models/period.rb
+++ b/app/models/period.rb
@@ -57,6 +57,16 @@ class Period
     (three_months_ago..end_date)
   end
 
+  alias_method :bp_control_range, :blood_pressure_control_range
+
+  def bp_control_range_start_date
+    bp_control_range.begin.next_day.to_s(:day_mon_year)
+  end
+
+  def bp_control_range_end_date
+    bp_control_range.end.to_s(:day_mon_year)
+  end
+
   def quarter?
     type == :quarter
   end
--- a/app/models/reports/result.rb
+++ b/app/models/reports/result.rb
@@ -167,11 +167,10 @@ module Reports
     DATE_FORMAT = "%-d-%b-%Y"
     def calculate_period_info(range)
       self.period_info = range.each_with_object({}) do |period, hsh|
-        bp_control_range = period.blood_pressure_control_range
         hsh[period] = {
           name: period.to_s,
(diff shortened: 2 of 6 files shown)


app/models/period.rb
```diff
@@ -57,6 +57,16 @@ def blood_pressure_control_range
...
 
+  alias_method :bp_control_range, :blood_pressure_control_range
+
+  def bp_control_range_start_date
+    bp_control_range.begin.next_day.to_s(:day_mon_year)
+  end
+
+  def bp_control_range_end_date
+    bp_control_range.end.to_s(:day_mon_year)
+  end
```
rsanheim, line 68: We use these begin / end dates enough that its worth exposing them right off the Period.

app/views/reports/regions/_visit_details_card.html.erb
```diff
@@ -26,23 +26,23 @@
                   <span data-missed-visits-patients="<%= number_with_delimiter(@data.last_value(:missed_visits)) %>">
                     <%= number_with_delimiter(@data.last_value(:missed_visits)) %>
                   </span>
-                  patients with no visit from 
-                  <span data-period-start="<%= @data.last_value(:period_info)["bp_control_start_date"] %>">
-                    <%= @data.last_value(:period_info)["bp_control_start_date"] %>
+                  patients with no visit from
+                  <span data-period-start="<%= @period.bp_control_range_start_date %>">
+                    <%= @period.bp_control_range_start_date %>
```
rsanheim, line 31: Instead of relying on the values in the `@data`, we can grab all these dates from the `@period` itself. We know the controller will set that period based on either the default current period (ie current month), or the period the user selects. So this is safer and more explicit.

app/models/reports/result.rb
```diff
@@ -85,7 +85,7 @@ def to_s
     end
 
     def last_value(key)
-      self[key].values.last || 0
```
rsanheim, line 88: I never liked this hacky default here, and we can remove it now.  

Its much better to know if there is actually a value or if its nil.

rsanheim (commented):
Notes for reviewers

kpethtel (approved):
👍

PR #2059, opened 2021-02-11
Grab just the data we need for child data tables via Repository

rsanheim:
Working towards some new pieces here that can pull out just the data we need for specific periods, instead of pulling back a giant set of data for all the child tables.


app/controllers/reports/regions_controller.rb
```diff
@@ -37,10 +37,21 @@ def show
     @with_ltfu = with_ltfu?
 
     @children = @region.reportable_children
+
+    repository = Reports::Repository.new(@children, periods: @period, with_exclusions: report_with_exclusions?)
```
rsanheim, line 41: The repository lets us grab the data for _just_ the period we are interested in for all the children regions.

app/controllers/reports/regions_controller.rb
```diff
@@ -37,10 +37,21 @@ def show
...
     @children_data = @children.map { |child|
-      Reports::RegionService.new(region: child,
-                                 period: @period,
-                                 with_exclusions: report_with_exclusions?).call
+      fetcher = repository.for_region_and_period(child, @period)
```
rsanheim, line 43: Not sure about this interface yet, but it makes it a bit nicer to grab the one specific datapoint we need for a region and period.

app/models/reports/repository.rb
```diff
@@ -17,13 +17,69 @@ def initialize(regions, periods:, with_exclusions: false)
...
+    #    region_slug: { period: value, period: value }
+    # }
+    def assigned_patients_count
+      full_assigned_patients_counts.each_with_object({}) do |(item, result), results|
+        values = periods.each_with_object({}) { |period, region_result| region_result[period] = result[period] }
+        results[item.region.slug] = values
+      end
+    end
+
+    def full_assigned_patients_counts
```
rsanheim, line 42: Not sure on the name here...we need to grab the entire range of assigned patient counts so we can calculate the running totals

rsanheim (commented):
notes

app/models/reports/region_item.rb
```diff
@@ -0,0 +1,16 @@
+module Reports
+  class RegionItem
```
prabhanshuguptagit, line 2: Instead of a new `RegionItem`, can we just use `Item` with `@period = Time.current`? I'm assuming this is because we want to do all time things such as `cumulative_registrations`.

app/controllers/reports/regions_controller.rb
prabhanshuguptagit: Is the alternative here to do `repository[child][@period]`? I think it would be simpler to just use hash lookup semantics if possible, makes it easier to trace things from views to the data.

prabhanshuguptagit (commented):
I haven't dug into the assigned patient changes, will do it once that's wrapped up. I would try and see if it's possible to do this neatly without `RegionAndPeriodFetcher` but overall this makes sense.

app/controllers/reports/regions_controller.rb
rsanheim: Yup, the fetcher piece may be overkill.  I have a simpler approach in mind for this that I'll try.

app/models/reports/region_item.rb
rsanheim: Correct, for some of the Regions we want to grab _all_ periods of data back to the start of that Region's history.  For those, it doesn't make sense to have period inside the item.

app/models/reports/region_entry.rb
```diff
@@ -0,0 +1,16 @@
+module Reports
+  class RegionEntry
```
kpethtel, line 2: I like the change from `RegionItem` to `RegionEntry` because it didn't feel right for `RegionItem` to be the same as `Item` but with less period specificity. If anything, it left me thinking `Item` should be renamed to `RegionItem` and this renamed to `RegionPeriodItem`.

app/controllers/reports/regions_controller.rb
kpethtel: I think there are a lot of options for how we can go about accessing the data. The approach Prabhanshu mentions is very straight-forward and avoids the additional layer of logic. The combination of having a method that returns an object that references and acts upon the called object, along with the use of method missing, definitely adds some cognitive overhead. And it's not clear that we're getting additional value out of it yet.

app/models/reports/repository.rb
```diff
@@ -18,49 +20,115 @@ def initialize(regions, periods:, with_exclusions: false)
...
+    #    region_slug: { period: value, period: value }
+    # }
+    def assigned_patients_count
+      full_assigned_patients_counts.each_with_object({}) do |(entry, result), results|
+        values = periods.each_with_object({}) { |period, region_result| region_result[period] = result[period] }
+        results[entry.region.slug] = values
+      end
+    end
+
+    def full_assigned_patients_counts
```
kpethtel, line 45: Should these methods be memoized? Pulling from the cache may be very efficient, but I would guess there's still some inefficiency here.

app/models/reports/region_item.rb
prabhanshuguptagit: If the `Reports::Item` has a `@period= Time.current` we _could_ ask it to store `cumulative_registrations` as of today. I think we can optimise for one less interface if it's not too much of a stretch.

rsanheim: > Could you try and see if it's possible to do this neatly without `RegionAndPeriodFetcher` and doing plain hash lookups?

Yup, I removed the fetcher and am doing a normal hash lookup now.

rsanheim:
This is ready for a thorough review. 

I don’t plan to merge until @prabhanshuguptagit has his LTFU work merged, and I’m sure this will require some changes after that to account for his changes. But the core ideas here are all solid.

app/controllers/reports/regions_controller.rb
```diff
@@ -34,10 +34,21 @@ def show
     @with_ltfu = with_ltfu?
 
     @children = @region.reportable_children
+
+    repo = Reports::Repository.new(@children, periods: @period, with_exclusions: report_with_exclusions?)
```
rsanheim, line 38: The Repository here grabs just the data for the `@period` we want for the children data, as opposed to RegionService which will always grab many months of data. More importantly, Repository caches at the region/period level, while RegionService caches the [entire range of data](https://github.com/simpledotorg/simple-server/blob/21abf030107f6751c7285e9cd4467355a4c2341a/app/services/control_rate_service.rb#L32-L34) available for any region.

app/controllers/reports/regions_controller.rb
```diff
@@ -34,10 +34,21 @@ def show
...
+      slug = child.slug
+      {
+        region: child,
+        controlled_patients: repo.controlled_patients_count[slug][@period],
+        controlled_patients_rate: repo.controlled_patient_rates[slug][@period],
+        uncontrolled_patients: repo.uncontrolled_patients_count[slug][@period],
+        uncontrolled_patients_rate: repo.uncontrolled_patient_rates[slug][@period],
+        missed_visits: repo.missed_visits_count[slug][@period],
+        missed_visits_percentage: repo.missed_visits_rate[slug][@period],
+        registrations: repo.assigned_patients_count[slug][@period],
```
rsanheim, line 49: @prabhanshuguptagit I'm thinking this should actually be registration data now?

prabhanshuguptagit: Yep should be registration counts.

app/controllers/reports/regions_controller.rb
prabhanshuguptagit: [nit] `children_data_repository = ...`?

app/controllers/reports/regions_controller.rb
```diff
@@ -34,10 +34,21 @@ def show
...
+        cumulative_patients: repo.cumulative_assigned_patients_count[slug][@period]
+      }
```
prabhanshuguptagit, line 51: Does this hash also need other `result` things like `period_info` etc.?

app/models/reports/repository.rb
```diff
@@ -1,64 +1,143 @@
...
+      cached_query(__method__) do |entry|
+        controlled = uncontrolled_patients_count[entry.region.slug][entry.period]
+        total = cumulative_assigned_patients_count[entry.region.slug][entry.period]
+        percentage(controlled, total)
+      end
+    end
+
+    smart_memoize def missed_visits_count
+      cached_query(__method__) do |entry|
+        no_bp_measure_query.call(entry.region, entry.period, with_exclusions: with_exclusions)
```
prabhanshuguptagit, line 99: To calculate missed visits we need to do
```
missed_visits = registrations - visited_without_bp_taken - controlled - uncontrolled
```

Not sure if the right thing is being called here. We should name the result of `no_bp_measure_query` as `visited_without_bp_taken` in any case.

app/models/reports/repository.rb
```diff
@@ -1,64 +1,143 @@
...
+    # because it is easier and fast enough to do so. We still return _just_ the periods the Repository was created with
+    # to conform to the same interface as all the other queries here.
+
+    # Returns a Hash in the shape of:
+    # {
+    #    region_slug: { period: value, period: value },
+    #    region_slug: { period: value, period: value }
+    # }
+    smart_memoize def assigned_patients_count
+      full_assigned_patients_counts.each_with_object({}) do |(entry, result), results|
```
prabhanshuguptagit, line 39: Is it possible to do a `group_by_period` and fill in `nils` after? That might be faster.

prabhanshuguptagit (dismissed):
I think this still needs to get some numbers right (missed/no bp visits, registrations vs assigned etc.) but looks mostly solid 👌🏼 .

app/models/reports/repository.rb
```diff
@@ -1,64 +1,143 @@
...
     attr_reader :with_exclusions
 
     delegate :cache, :logger, to: Rails
 
-    def controlled_patients_count
-      cached_query(:controlled_patients_count) do |item|
-        control_rate_query.controlled(item.region, item.period, with_exclusions: with_exclusions).count
+    # Uses Memery to memoize a method, but takes into account our force_cache setting. If force_cache is true,
+    # a caller is asking for all values to be retrieved fresh from the database, so we want to skip memoization.
+    def self.smart_memoize(method)
```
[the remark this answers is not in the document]
kpethtel: I'm not sure I understand this either. If I understand this correctly, when we specify force_cache, the method will not be memoized and we will request the data from the DB each time the method is called (which can be multiple times per request). If we don't specify force_cache, we will memoize and grab from the cache. So it seems like this will improve performance for the case where we use the cache but not for the case where we use the DB. Which seems like it will make the less performant case less performant. Perhaps I'm misunderstanding.

kpethtel (approved):

kpethtel (changes requested):
Sorry, I didn't mean to approve. I was just leaving that comment.

app/models/reports/repository.rb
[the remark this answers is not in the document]
rsanheim: The memoization is _just_ for the scope fo the request, for cases where we are in a tight loop and grabbing counts for different regions / periods over and over.  

> If we don't specify force_cache, we will memoize and grab from the cache. So it seems like this will improve performance for the case where we use the cache but not for the case where we use the DB. Which seems like it will make the less performant case less performant. Perhaps I'm misunderstanding.

This is correct, and how it is intended to work.  `force_cache` means to bust the cache and grab fresh values, its only used by us for debugging or when priming the cache.  All caching must be bypassed for that to work, including memoization.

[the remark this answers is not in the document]
rsanheim: side note - probably could use a better name for the `force_cache` piece ...maybe `force_fetch` ? `bust_cache` ?

app/controllers/reports/regions_controller.rb
rsanheim: No, the children tables don't need it.

I want to move away from the complexity of Result and just use plain hashes like this, so I'm just returning only the key/values we need.

app/models/reports/repository.rb
```diff
@@ -1,64 +1,143 @@
...
+    attr_reader :regions
     attr_reader :with_exclusions
 
     delegate :cache, :logger, to: Rails
 
-    def controlled_patients_count
-      cached_query(:controlled_patients_count) do |item|
-        control_rate_query.controlled(item.region, item.period, with_exclusions: with_exclusions).count
+    # Uses Memery to memoize a method, but takes into account our force_cache setting. If force_cache is true,
+    # a caller is asking for all values to be retrieved fresh from the database, so we want to skip memoization.
```
kpethtel, line 24: After our conversation on slack, I think it might be worth adding some additional comments like:
```
force_cache is only used when warming the cache and do not want to memoize in that case
```

app/models/reports/repository.rb
```diff
@@ -1,64 +1,143 @@
...
+    smart_memoize def cumulative_assigned_patients_count
+      full_assigned_patients_counts.each_with_object({}) do |(region_entry, patient_counts), totals|
+        region_slug = region_entry.region.slug
+        totals[region_slug] = Hash.new(0)
+        next if patient_counts.empty?
+        full_range = (patient_counts.keys.first..periods.end)
+        full_range.each do |period|
+          previous_total = totals[region_slug][period.previous]
+          current_amount = patient_counts[period] || 0
+          totals[region_slug][period] += previous_total + current_amount
```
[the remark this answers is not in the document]
rsanheim: 👍

app/models/reports/repository.rb
```diff
@@ -1,64 +1,143 @@
...
+      combinations = regions.to_a.product(periods.to_a)
+      combinations.map { |region, period| Reports::RegionPeriodEntry.new(region, period, calculation, with_exclusions: with_exclusions) }
+    end
+
+    def percentage(numerator, denominator)
+      return 0 if denominator == 0 || numerator == 0
+      ((numerator.to_f / denominator) * 100).round(PERCENTAGE_PRECISION)
     end
 
     def no_bp_measure_query
```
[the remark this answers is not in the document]
rsanheim: sure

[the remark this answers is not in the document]
rsanheim: actually I'm gonna go with good ole' constructor creation for these.

app/controllers/reports/regions_controller.rb
rsanheim: this is done

app/models/reports/repository.rb
rsanheim: not sure I follow - this is grabbing just the subset of assigned patient counts the caller is asking for here.

kpethtel (approved):
There's a lot going on here. Great work.

app/models/reports/repository.rb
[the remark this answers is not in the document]
rsanheim: > side note - probably could use a better name for the `force_cache` piece ...maybe `force_fetch` ? `bust_cache` 

see https://github.com/simpledotorg/simple-server/pull/2146

app/models/reports/repository.rb
rsanheim: you may be interested in https://github.com/simpledotorg/simple-server/pull/2146

app/models/reports/repository.rb
rsanheim: Yup, you're right.  I've fixed this up to have the correct name now.

app/services/reports/region_service.rb
```diff
@@ -29,7 +29,7 @@ def initialize(region:, period:, with_exclusions: false, months: MAX_MONTHS_OF_D
 
     def call
       result = ControlRateService.new(region, periods: range, with_exclusions: with_exclusions).call
-      result.visited_without_bp_taken = repository.no_bp_measure_count[region.slug]
+      result.visited_without_bp_taken = repository.missed_visits_count[region.slug]
```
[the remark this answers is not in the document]
rsanheim: Fixed this up.

prabhanshuguptagit (approved):
Looks solid. Thanks, this PR has been hanging around for a while now 🙏🏼 

What are your thoughts on the future for `Repository`? If (and when) we plan to move away from `RegionService`, there's a bunch of methods this will need to support that aren't here yet:
- Cumulative registrations
- LTFU patient counts
- Controlled patients with/without LTFU
- Misc like `period_info`

We'll just add them to `Repository` over time? I'll have to add some more numbers to the dashboard (LTFU variants of uncontrolled rate, missed visits), wondering if it's a good time to add them here instead of `result.rb`.

PR #2060, opened 2021-02-11
Abort if git errors out in bin/release

rsanheim:
Saw this happen locally when I accidentally duplicated an existing tag name.

The script should abort and not continue.

(no story)


vkrmis (approved):
lgtm

PR #2061, opened 2021-02-11
Patient line list LTFU status

prabhanshuguptagit:
**Story card:** [ch2530](https://app.clubhouse.io/simpledotorg/story/2530/add-lost-to-follow-up-status-in-patient-line-list)

## Because

We are making changes to the patients we show on our dashboards. [PRD](https://docs.google.com/document/d/1OD28cMu1N8W44ObJVtf7513V7vhZR_zfjSmVWs4OymE/edit#) We want to update the line list to show the correct status.

## This addresses

- Updates the “Status” column to have values:
  - Active
  - Lost to follow-up
  - Transferred out


harimohanraj89 (approved):
Let's do a performance test of this in a large facility group before we ship it.

app/exporters/patients_exporter.rb
```diff
@@ -191,4 +192,14 @@ def latest_blood_sugar_type(patient_summary)
     patient_summary.latest_blood_sugar_type.presence &&
       BLOOD_SUGAR_TYPES[patient_summary.latest_blood_sugar_type]
   end
+
+  def status(patient_summary)
+    patient_status = if patient_summary.ltfu?
+      :ltfu
+    else
+      patient_summary.status
+    end
```
harimohanraj89, line 201: This kind of tree makes me wonder if we should compute the status, including LTFU, when populating the view.

Is that too much business logic inside view generation? Does patient summary have too many touchpoints to confidently adjust the values we put inside the status?

prabhanshuguptagit: This _could_ sit in the view too. The information we'll have to put in the view would be:
- what's the `LTFU_PERIOD`
- the status field in the CSV can also be `Lost to follow-up`. If we want to split the column again to `status` and `ltfu-ness`, this might need a change.

wydt? This is isn't very expensive to calculate in ruby and the win there is we have some flexibility about these two things.

harimohanraj89: Hmm, yeah. We may not know enough just yet. Let's leave it as is.

prabhanshuguptagit:
Testede on perf, there isn't a noticeable difference in performance.

PR #2065, opened 2021-02-11
Add a Rachet to start decreasing our spec's reliance on let!

rsanheim:
`let!` should almost never be used in specs, as it leads to lots of
implicit creation of test objects as well as making tests harder to read
and maintain.  Good alternatives are inline setup and/or regular ruby methods if you have lots of complicated object setup to consolidate.

This adds a rachet to help us in slowly decreasing the usage of let!. It
will fail if you add more let! usage, and also tells you to lower the
rachet count down if you have removed our usage of let!.

See https://thoughtbot.com/blog/lets-not and https://thoughtbot.com/blog/my-issues-with-let for more context on why let! is bad some alternatives to use.

**Story card:** [chXXXX](URL)


spec/rachets/lets_not_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "spec_helper"
+
+RSpec.describe "Rachet down usage of let!" do
+  expected_usages = 698
+
+  it "ratchets down let!" do
+    command = %{grep -rn "let\!(:" spec}
```
timcheadle, line 7: may want to add an `\s` in front of the regex to avoid matching methods that are like `inlet!`

timcheadle (approved):

kpethtel (approved):
Clever way of handling this.

PR #2066, opened 2021-02-12
2531: add registered hypertension counts to my facilities views

kpethtel:
**Story card:** https://app.clubhouse.io/simpledotorg/story/2531

## Because

The data in the my facilities views is incorrect. 

## This addresses

This changes the views in my facilities to use registered hypertensive patients per region/facility.


claudiovallejo (approved):
Front-end changes look good to me 🔥

PR #2068, opened 2021-02-15
LTFU exclusions performance fixes

prabhanshuguptagit:
**Story card:** [ch1966](https://app.clubhouse.io/simpledotorg/epic/1866/denominator-exclusions)

## Because

The LTFU queries add a lot of overhead to the cache warmer, we noticed cache warmer with the exclusions query to be taking around [9 hours on the perf environment](https://app.datadoghq.com/logs?cols=host%2Cservice%2Cenv&event=AQAAAXekqrP-6Ey9ogAAAABBWGVrcXJiMkFBQWp5M09xbE9ISFFnQUQ&from_ts=1612791824109&index=&live=true&messageDisplay=inline&query=finished&stream_sort=desc&to_ts=1613396624109)

## This addresses

This makes some improvements to the LTFU query which improves it's worst case performance by ~50%. This also parallelizes the cache warmer to run a separate job per region. 

The jobs take roughly ~4-5 mins combined to finish when running on a prod sized data set (~6k regions tested on perf).

### DIFF
--- a/app/jobs/region_cache_warmer_job.rb
+++ b/app/jobs/region_cache_warmer_job.rb
@@ -1,0 +1,16 @@
+class RegionCacheWarmerJob
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :default
+
+  def perform(region_id, period_attributes)
+    region = Region.find(region_id)
+    period = Period.new(period_attributes)
+
+    Reports::RegionService.call(region: region, period: period)
+    Statsd.instance.increment("region_cache_warmer.#{region.region_type}.cache")
+
+    Reports::RegionService.call(region: region, period: period, with_exclusions: true)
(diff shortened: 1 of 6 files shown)


app/jobs/region_cache_warmer_job.rb
```diff
@@ -0,0 +1,33 @@
+class RegionCacheWarmerJob
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :default
```
rsanheim, line 4: This all seems fine, but this is going to completely saturate the default queue with hundreds of jobs while this runs...would it make more sense to put this in the `low` queue? The cache should still be warmed from the previous run, so these aren't really high or even 'default' priority.

ssrihari: The audit logs clog up the low queue in multiples of 100k. Might actually not be so bad to keep this at a slightly higher priority.

PR #2073, opened 2021-02-16
2092: prevent search filters from resetting when overdue patient stat…

kpethtel:
…us is updated

**Story card:** https://app.clubhouse.io/simpledotorg/story/2092

## Because
There's a bug in the overdue patients index in which the search filters reset to their default value after a 

## This addresses
Fixes that bug, updates tests to avoid `let!`, adds new tests.


app/controllers/appointments_controller.rb
```diff
@@ -10,9 +10,9 @@ def index
     authorize { current_admin.accessible_facilities(:manage_overdue_list).any? }
 
     @search_filters = index_params[:search_filters] || []
-    # We have to check to see this is the first page load where we want to apply default search filters. This
-    # is the case where the form is _not_ submitted and there are no search filters passed in.
-    if @search_filters.blank? && !index_params[:submitted]
+    # only apply default search filters on navigating to this page
+    # after initial page load, params will at least include a facility
+    if index_params.keys.empty?
```
kpethtel, line 15: I decided to move away from the use of `params[:submitted]` because after the initial page load, the page will always append a facility to the query params. After some thought, I concluded that this route without any query params should probably behave identically to "first page load" anyhow, so it seems like the simpler solution to me.

app/controllers/appointments_controller.rb
```diff
@@ -38,10 +38,12 @@ def index
 
   def update
     call_result = appointment_params[:call_result].to_sym
+    search_filters = appointment_params[:search_filters]&.split || []
```
kpethtel, line 41: As best I can tell, the only way to include an array in a hidden field is as a space separated string, so we have to split here.

app/controllers/appointments_controller.rb
```diff
@@ -38,10 +38,12 @@ def index
...
 
     if set_appointment_status_from_call_result(@appointment, call_result)
-      redirect_to appointments_url(params: {facility_id: selected_facility_id, page: page}),
-        notice: "Saved. #{@appointment.patient.full_name} marked as \"#{call_result.to_s.humanize}\""
+      redirect_to appointments_url(
+        params: {facility_id: selected_facility_id, page: page, search_filters: search_filters}
+      ), notice: "Saved. #{@appointment.patient.full_name} marked as \"#{call_result.to_s.humanize}\""
```
rsanheim, line 46:
```suggestion
      appt_params = {facility_id: selected_facility_id, page: page, search_filters: search_filters}
      notice = %Q{Saved. #{@appointment.patient.full_name} marked as "#{call_result.to_s.humanize}"}
      redirect_to appointments_url(params: appt_params), notice: notice
```

spec/rachets/lets_not_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "spec_helper"
 
 RSpec.describe "Rachet down usage of let!" do
-  expected_usages = 705
+  expected_usages = 694
```
rsanheim, line 4: 🎉

rsanheim (approved):
One suggestion, otherwise looks good to merge.

PR #2075, opened 2021-02-16
Fix report links for BD

rsanheim:
reports besides district and facility should be linkable in countries
that don't have extened region reports

Fixes [ch2652](https://app.clubhouse.io/simpledotorg/story/2652/reports-broken-in-bd-argumenterror-unsupported-region-type-state-for-legacy-children)

Fixes https://sentry.io/organizations/resolve-to-save-lives/issues/2210183543/?project=1793711


rsanheim:
Merging to get this fixed in prod.

PR #2077, opened 2021-02-17
Add missing indexes on matviews

prabhanshuguptagit:
**Story card:** -

## Because

Our materialised views aren't indexed by `patient_id`. This mainly hurts the newly added LTFU queries. This is a follow up to #2068.

## This addresses

This adds a `patient_id` index to the latest BP materialized views.


harimohanraj89 (approved):
These indices make sense to me.

db/migrate/20210217111153_add_patient_id_index_to_bp_mat_views.rb
```diff
@@ -0,0 +1,18 @@
...
+      name: "index_latest_bp_per_patient_per_months_patient_id",
+      algorithm: :concurrently
+    add_index :latest_blood_pressures_per_patient_per_quarters,
+      :patient_id,
+      name: "index_latest_bp_per_patient_per_quarters_patient_id",
+      algorithm: :concurrently
+    add_index :latest_blood_pressures_per_patients,
+      :patient_id,
+      name: "index_latest_bp_per_patient_patient_id",
+      algorithm: :concurrently
```
ssrihari, line 16: Do we know that these indexes are being used? At some point (not now surely), the cost of having these indexes will add up.

ssrihari (approved):

db/migrate/20210217111153_add_patient_id_index_to_bp_mat_views.rb
ssrihari: Use https://github.com/nilenso/postgresql-monitoring#index_usage

PR #2079, opened 2021-02-17
Bug: My Facilities table sorting

claudiovallejo:
**Story card:** [ch2687](https://app.clubhouse.io/simpledotorg/story/2687/fix-my-facilities-subtotal-table-sorting)

## Because

The "Total" row in "My Facilities" should always be the first row. Currently, changing the default sort method changes the "Total" row's placement in the table.

## This addresses

* Adds `data-default-method="none"` to the "Total" row in each view


kpethtel (approved):
👍

PR #2081, opened 2021-02-17
2115 simplify password requirements

kpethtel:
**Story card:** https://app.clubhouse.io/simpledotorg/story/2115

## Because

Our current password scheme is difficult to use and lacks a definitive, intuitive idea of what constitutes an acceptable password. 

## This addresses

This changes our password validations from an entropy based scheme to a more commonly understood list of requirements. 

## Testing

Test this by:
- signing in as a user with an old password that wouldn't match the current scheme (lacks numbers, for example) to confirm that the change doesn't cause any backwards compatibility issues
- creating a password for a newly invited admin
- resetting a password for an existing admin


spec/models/email_authentication_spec.rb
```diff
@@ -1,47 +1,89 @@
...
-  it "requires a non blank password" do
-    auth = build(:email_authentication, password: "")
-    expect(auth).to_not be_valid
-    expect(auth.errors.messages[:password]).to eq [weak_password_error]
-  end
+  describe "password" do
+    it "can't be nil" do
+      auth = build(:email_authentication, password: nil)
+      expect(auth).to_not be_valid
+      # expect(auth.errors.messages[:password]).to include "can't be blank"
```
kpethtel, line 13: All of these commented out error message tests are intentional. They will be made live when the UI is changed to handle multiple error messages.

spec/models/email_authentication_spec.rb
```diff
@@ -1,47 +1,89 @@
...
-    auth = EmailAuthentication.find(auth.id)
-    auth.email = "newemail@example.com"
-    expect(auth).to be_valid
-    auth.save!
-    # make sure a weak password change is prevented
-    auth.password = "passw0rd"
-    expect(auth).to_not be_valid
-    expect(auth.errors.messages[:password]).to eq [weak_password_error]
+  describe "updating" do
+    it "does not allow password to be an nil" do
```
kpethtel, line 59: I'm specifically testing this case because there is no model validation for presence of password on update. That is inherent to the design, since we need to allow it to be nil in order to update the email without the password. A nil password is guarded against on update by either devise or the postgres.

.env.development
```diff
@@ -120,6 +120,6 @@ OWNER_EMAILS='Owner <owner@test.com>, Another Owner <another-owner@test.com>'
 PROGRAM_INCEPTION_DATE=2018-01-01
 SUPERVISOR_EMAILS='Supervisor <supervisor@test.com>'
 
-SEED_GENERATED_ADMIN_PASSWORD='baby monkey honey badger star'
```
kpethtel, line 123: I'm not sure if there's a reason two of the seed passwords are different than the rest. I've made them all the same for simplicity, but I can make changes if there's a reason to keep them different.

rsanheim: Sounds good -- this particular password was a holdover from earlier seed code.

When you do ship this, please make sure to let every dev/designer know that their seed users will have different passwords the next time they do a reseed.

config/locales/en.yml
```diff
@@ -8,7 +8,7 @@ en:
   errors:
     messages:
       password:
-        password_strength: "Please choose a stronger password with at least 8 characters. Try a mix of letters, numbers, and symbols."
+        password_strength: "Password must be 10 characters long and contain a lower case letter, an upper case letter, and a number."
```
rsanheim, line 11:
```suggestion
        password_strength: "Password must at least 10 characters long and contain a lower case letter, an upper case letter, and a number."
```

spec/models/email_authentication_spec.rb
rsanheim: I'm not sure I follow -- wouldn't these model validations pass right now, because you have added the new validations above.  How does this depend upon the UI?

kpethtel: Currently, the validation messages get wiped out (if there's more than one of them) by this `after_validation` in the EmailAuthentication model.
```
  # We only want to display one error message to the user, so if we get multiple
  # errors clear out all errors and present our nice message to the user.
  def strip_unnecessary_errors
    if errors[:password].any? && errors[:password].size > 1
      errors.delete(:password)
      errors.add(:password, I18n.translate("errors.messages.password.password_strength"))
    end
  end
  ```
If you turn that validation off and have multiple error messages, you get this:
image
Not terrible but not ideal. I'm sure I could fix that in the view, but I figured I would leave it alone for now. I want to have multiple error messages to enable the password requirements checklist. So my plan was to just leave the error message overwriting until I have the UI changes for handling multiple errors.

.env.development
kpethtel: 👍

spec/models/email_authentication_spec.rb
rsanheim: I think removing that `after_validation` is fine, because the `nice message` it presents won't match these new validation rules: `Please choose a stronger password with at least 8 characters. Try a mix of letters, numbers, and symbols."`. Plus its better to make sure these specs are really running when they get merged, just to be sure our password rules are all good.

If we have ugly (but accurate) error messages displayed to a new user until you have the UI piece shipped, that seems fine to me.

kpethtel: Works for me. I did update that "nice message" to be correct, but I don't think that changes anything. It seems the more important point is perhaps that we can live with an ugly error message on a low-use page for a few days.

kpethtel: Works for me. I did update that "nice message" to be correct, but I don't think that changes anything. It seems the more important point is perhaps that we can live with an ugly error message on a low-use page for a few days.

rsanheim (approved):
looks good!

PR #2082, opened 2021-02-18
Merge duplicate patients

ssrihari:
**Story card:** [ch2665](https://app.clubhouse.io/simpledotorg/story/2665/merge-patient-attributes-and-entities-according-to-the-merge-logic)

## Because

[PRD](https://docs.google.com/document/d/1TItK2-kpQj91lNmOhzmq2CP1J15AB5knceobwZOIKoA/edit#) We want to merge duplicate patients. There are two steps to this problem:
- Identifying duplicate patients
- Merging these patients and their associated data

This PR tackles the second step.

## This addresses

This adds a `MergeDuplicatePatients` service that can merge a list of given patients (potentially duplicate). The exact merge logic is outlined [here](https://docs.google.com/document/d/1TItK2-kpQj91lNmOhzmq2CP1J15AB5knceobwZOIKoA/edit#heading=h.q1zwm2eg5vgc). TLDR:
- We create a new patient.
- We "copy over" all associated data for the patients to the new patient. The copying mechanism is whatever makes sense for each associated resource.
- The old duplicate patients are soft deleted.


app/services/merge_duplicate_patients.rb
```diff
@@ -0,0 +1,194 @@
+class MergeDuplicatePatients
```
harimohanraj89, line 1: How about `MergePatients`, because we can pass any arbitrary collection of patients in here? This class isn't responsible for identifying the dupes.

app/services/merge_duplicate_patients.rb
```diff
@@ -0,0 +1,194 @@
...
+      create_phone_numbers(new_patient)
+      create_patient_business_identifiers(new_patient)
+      create_encounters_and_observables(new_patient)
+      create_appointments(new_patient)
+      create_teleconsultations(new_patient)
+      mark_as_merged(new_patient)
+      new_patient.reload
+    end
+
+    Rails.logger.info "Merged patients #{@patients.pluck(:id)} into patient #{new_patient.id}"
```
harimohanraj89, line 28: What happens when the transaction above is rolled back because of a problem? Does this line of code still execute, or does the process halt with an exception?

harimohanraj89: Additionally, should we add a log _before_ we start the merge, so we can quickly identify which patients are involved if a merge _fails_?

db/migrate/20210224120024_add_merged_into_to_patient.rb
```diff
@@ -0,0 +1,6 @@
+class AddMergedIntoToPatient < ActiveRecord::Migration[5.2]
+  def change
+    add_column :patients, :merged_into, :uuid
```
harimohanraj89, line 3: Consider `merged_into_patient_id`? I like the idea of the foreign key having the `<model>_id` suffix. 

Additionally, shall we set up the associations necessary to say things like `patient.merged_to_patient` and `patient.merged_from_patients`?

spec/factories/encounters.rb
```diff
@@ -54,3 +54,21 @@ def updated_encounters_payload(existing_encounter)
...
+    patient: observable.patient,
+    observable: observable,
+    facility: observable.facility,
+    encountered_on: encountered_on)
+
+  create(:observation,
+    encounter_id: encounter_id,
+    observable: observable,
+    observable_type: observable.class.name,
+    user: observable.user)
```
harimohanraj89, line 73: 🙌

spec/factories/patients.rb
```diff
@@ -121,3 +122,31 @@ def updated_patient_payload(existing_patient)
...
+  create(:blood_pressure, :with_encounter, :critical, recorded_at: visited_at, facility: facility, patient: patient, user: user)
+  create(:blood_sugar, :fasting, :with_encounter, recorded_at: visited_at, facility: facility, patient: patient, user: user)
+  create_list(:prescription_drug, 2, :protocol, device_created_at: visited_at, facility: facility, patient: patient, user: user)
+  create_list(:prescription_drug, 2, device_created_at: visited_at, facility: facility, patient: patient, user: user)
+  patient.appointments.where("device_created_at < ?", visited_at).update_all(status: :visited)
+  create(:appointment, status: :scheduled, device_created_at: visited_at, scheduled_date: 1.month.after(visited_at), creation_facility: facility, facility: facility, patient: patient, user: user)
+  create(:teleconsultation, patient: patient, facility: facility, requester: user, medical_officer: user, requested_medical_officer: user)
+end
+
+def add_some_visits(patient, visit_count, facility: patient.registration_facility, user: patient.registration_user)
```
harimohanraj89, line 137: I think we can remove `some` from the method name 😅

harimohanraj89: If we wanna take inspiration from some of the factorybot patterns, like `create_list`, we could update the method signature to accept the count first, and make everything else a named arg:

`add_visits(visit_count, patient:, facility:, user:)`

spec/services/merge_duplicate_patients_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+    context "marks patients as merged" do
+      it "soft deletes the merged patients" do
+        patients = create_duplicate_patients.values
+
+        expect(patients.first).to receive(:discard_data).and_call_original
+        expect(patients.last).to receive(:discard_data).and_call_original
+
+        described_class.new(patients).merge
+
+        expect(patients.map(&:discarded?)).to all be true
```
harimohanraj89, line 212: TIL `to all be`

app/services/merge_duplicate_patients.rb
```diff
@@ -0,0 +1,194 @@
...
+      full_name: latest_patient.full_name,
+      gender: latest_patient.gender,
+      reminder_consent: latest_patient.reminder_consent,
+      recorded_at: earliest_patient.recorded_at,
+      registration_facility: earliest_patient.registration_facility,
+      registration_user: earliest_patient.registration_user,
+      device_created_at: earliest_patient.device_created_at,
+      device_updated_at: earliest_patient.device_updated_at,
+      assigned_facility: latest_patient.assigned_facility,
+      address: create_address
```
harimohanraj89, line 49: I love how this reads.

app/services/merge_duplicate_patients.rb
```diff
@@ -0,0 +1,194 @@
...
+    }.merge(age_and_dob)
```
harimohanraj89, line 50: If you're feeling fancy
```ruby
{
  # ...
  address: create_address,
  **age_and_dob
}
```

I like how keeping it inside the hash makes it clear that we're just setting some more attributes.

app/services/merge_duplicate_patients.rb
```diff
@@ -0,0 +1,194 @@
...
+    Patient.create!(attributes)
+  end
+
+  def age_and_dob
+    if @patients.map(&:date_of_birth).any?
+      {
+        date_of_birth: latest_available_property(:date_of_birth),
```
harimohanraj89, line 57: s/property/attribute?

app/services/merge_duplicate_patients.rb
```diff
@@ -0,0 +1,194 @@
...
+  def create_cloned_records!(patient, klass, records)
+    records.map do |record|
+      create_cloned_record!(patient, klass, record)
+    end
+  end
+
+  def create_cloned_record!(patient, klass, record)
+    klass.create!(
+      copyable_attributes(record)
+        .merge(id: SecureRandom.uuid, patient_id: patient.id)
```
harimohanraj89, line 185: I like how generalizable this was.

db/migrate/20210224120024_add_merged_into_to_patient.rb
harimohanraj89: Hmm... I feel like we can do better with the naming here too 🤔  I'm thinking something like `patient.duplicate_patients`, and `patient.original_patient`, but that's not _quite_ right. It's not an original, it's a _new_ one that is now the source of truth/consolidated result. Are there better words?

harimohanraj89 (changes requested):
My feedback below! Very important thing to test is - do merged patients sync correctly to device.

app/services/merge_duplicate_patients.rb
prabhanshuguptagit: The process would halt with an exception. We are planning on capturing this like [here](https://github.com/simpledotorg/simple-server/pull/2104/files#diff-d0645edf5c6ace2983e2040dc09f061206d993d2aa4b60e13945dc13213f701aR15) and reporting it.

spec/services/merge_duplicate_patients_spec.rb
prabhanshuguptagit: 💅

db/migrate/20210224120024_add_merged_into_to_patient.rb
prabhanshuguptagit: `merged_to_patient` and `merged_from_patients` seem like good names actually.

app/services/merge_duplicate_patients.rb
prabhanshuguptagit: Went from `attribute` to `property` in 7f7fcc9 since this method can also lookup associated records and `attribute` is predominantly being used for fields on a patient. wdyt?

app/services/merge_duplicate_patients.rb
ssrihari: We already have a different `MergePatientService` :/ How about `DedupPatients` or something that is not similar to the "merge" terminology?

db/migrate/20210224120024_add_merged_into_to_patient.rb
ssrihari: `merged_into_patient_id` seems good. I thought about the associations, and figured we'll add them when we need them. I suspect we might not actually need that in code. Only in one-off analytics perhaps.

app/services/merge_duplicate_patients.rb
prabhanshuguptagit: I was thinking along `CombinePatientsWithData` for the same reason.

app/services/merge_duplicate_patients.rb
ssrihari: Heh, @prabhanshuguptagit mentioned this, and I looked up the double splat at the time as well. I felt like keeping merge there was easier to read 🤷‍♀️  Not hardcore rubyist enough I guess 😁

PR #2083, opened 2021-02-18
Throttle cache warmer job

prabhanshuguptagit:
**Story card:** -

## Because

We made region cache warmer run in parallel via sidekiq. This chokes the DB since a lot of connections get opened by the jobs and the queries don't release them quickly enough.

## This addresses

This adds a concurrency limit to the job. Tested on perf, this manages to warm prod sized data with peak 60% CPU on a `t3.small` RDS.

### DIFF
--- a/app/jobs/region_cache_warmer_job.rb
+++ b/app/jobs/region_cache_warmer_job.rb
@@ -1,7 +1,9 @@
 class RegionCacheWarmerJob
   include Sidekiq::Worker
+  include Sidekiq::Throttled::Worker
 
   sidekiq_options queue: :default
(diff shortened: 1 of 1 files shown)


harimohanraj89 (approved):
Ah nice, we can throttle concurrency, instead of time to queue. Love it. How long will production cache warming take with this change?

prabhanshuguptagit:
@harimohanraj89 This takes roughly 5-6 minutes on perf, so should be something under that in prod.

harimohanraj89:
That's incredifast 👌  Godspeed.

PR #2084, opened 2021-02-18
2117 bootstrap show password

kpethtel:
**Story card:** https://app.clubhouse.io/simpledotorg/story/2117

## Because

The password confirmation process does not fit our new approach to validating passwords on the fly, so we're giving users the users the ability to see their password to reduce errors.


app/views/email_authentications/passwords/edit.html.erb
```diff
@@ -1,12 +1,14 @@
...
 <h2>Change your password</h2>
 
 <div class="row">
   <div class="col-md-6">
     <%= bootstrap_form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :put }, label_errors: false) do |f| %>
       <%= f.hidden_field :reset_password_token %>
 
-      <%= f.password_field :password, autofocus: true, autocomplete: "off" %>
-      <%= f.password_field :password_confirmation, autocomplete: "off", label: "Confirm new password" %>
+      <%= f.password_field :password, id: "p  assword", autofocus: true, autocomplete: "off", "data-toggle": "password" %>
```
rsanheim, line 11: Some strange spacing here in the `id`.

app/views/email_authentications/invitations/edit.html.erb
```diff
@@ -1,3 +1,6 @@
+<link rel="stylesheet" type="text/css" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
```
rsanheim, line 1: Is this the full bootstrap CSS?  Or a limited set that you just need for this...we already have bootstrap in the app wide bundle.

app/views/email_authentications/passwords/edit.html.erb
kpethtel: woops. i saw that in the DOM and didn't know where it was coming from. must have hit the space key

app/views/email_authentications/invitations/edit.html.erb
kpethtel: I don't know, but the eye asset it's looking for does not appear to be accessible without adding it. I'd speculate there's a better way of handling this, but the jquery function does hard code the icon name.

PR #2089, opened 2021-02-19
Extract patient count queries; use from ContralRateService

rsanheim:

Breaking out some smaller pieces from ControlRateService.  This will make it safer and easier to merge #2059 soon.

### DIFF
--- a/app/services/control_rate_service.rb
+++ b/app/services/control_rate_service.rb
@@ -11,6 +11,7 @@ class ControlRateService
     @facilities = region.facilities
     @periods = periods
     @report_range = periods
+    @period_type = @report_range.begin.type
     @quarterly_report = @report_range.begin.quarter?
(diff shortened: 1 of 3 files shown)


app/services/control_rate_service.rb
```diff
@@ -11,6 +11,7 @@ def initialize(region, periods:, with_exclusions: false)
     @facilities = region.facilities
     @periods = periods
     @report_range = periods
+    @period_type = @report_range.begin.type
```
kpethtel, line 14: Does it matter at all that you're using `begin` here and `end` below? I would guess all the periods are the same type, so probably not.

kpethtel (approved):

app/services/control_rate_service.rb
rsanheim: Yup, they will be all the same type.

PR #2091, opened 2021-02-19
Drug stock progress tab

rsanheim:
**Story card:** [ch2224](https://app.clubhouse.io/simpledotorg/story/2224/add-drug-stock-form-to-progress-tab-behind-a-feature-flag)

Starting w/ just the static UI from @dburka ....


app/controllers/api/drug_stocks_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+          :in_stock,
+          :protocol_drug_id]
+    )
+  end
+
+  def set_for_end_of_month
+    @for_end_of_month ||= if params[:for_end_of_month]
+      Date.strptime(params[:for_end_of_month], "%b-%Y").end_of_month
+    else
+      Date.today.end_of_month
```
kpethtel, line 38: Does this need to be Date.current? Seems like this could cause an edge case around the end of the month.

spec/controllers/api/drug_stocks_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
+require "rails_helper"
+
+RSpec.describe Api::DrugStocksController, type: :controller do
+  before do
```
kpethtel, line 4: It seems like you're turning the flag on for all tests within the seed data, so is this just to ensure that the tests work when the user hasn't reseeded recently? Also, I don't see any uses of the feature flag in the code. Is the code behind this endpoint feature flagged?

spec/controllers/api/drug_stocks_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      }
+
+      expect {
+        post :create, params: params
+        expect(response).to be_successful
+      }.to change { DrugStock.count }.by(1)
+      expect(JSON.parse(response.body)).to eq({"status" => "OK"})
+    end
+
+    it "sends error messages for invalid saves" do
```
kpethtel, line 43: If I'm reading the controller correctly, we roll back the transaction if any records are invalid. It might be worth a test to confirm that no records are created if any of them are invalid.

kpethtel (approved):
Looks good for a draft. Just a few thoughts.

kpethtel:
Closing because this is covered by another PR now.

PR #2094, opened 2021-02-22
User settings

harimohanraj89:
**Story card:** [ch2752](https://app.clubhouse.io/simpledotorg/story/2752/user-setting-to-stop-receiving-approval-emails)

## Because

Some users downgrade their access because they don't want email spam.

## This addresses

Allow dashboard users to opt out of approval emails. If opted out, they will no longer receive:
* New registration approval emails
* PIN reset approval emails

The new field is `users.receive_approval_emails`, and defaults to `true` to avoid changing current behavior.

The checkbox has been added to the admin management screen so folks like Praveen can turn emails off for Dr. Kiran, etc. We can easily extend this concept to a "Settings" page, but I've avoided doing that for now. Maybe after we have 2 or 3 settings worth exposing.

I played around with the idea of a separate table to capture these settings, but `has_one` relationships kinda suck. Ensuring data integrity, managing nested form resources, all of this was extremely painful, and you don't really gain much anyway. Similar to how we wish medical histories were just extra fields on `Patient`, I've stuck the new setting directly onto the `users` table.

## Screenshots

Screen Shot 2021-02-22 at 7 59 44 PM


db/schema.rb
```diff
@@ -730,6 +731,80 @@
      FROM (appointments ap
        JOIN facilities f ON ((ap.facility_id = f.id)));
   SQL
+  create_view "latest_blood_pressures_per_patient_per_months", materialized: true, sql_definition: <<-SQL
```
prabhanshuguptagit, line 734: Let's discard these reordering changes.

spec/controllers/admins_controller_spec.rb
```diff
@@ -112,8 +112,8 @@
       let(:organization) { create(:organization) }
       let(:facility_group) { create(:facility_group, organization: organization) }
       let(:facilities) { create_list(:facility, 2, facility_group: facility_group) }
-      let(:manager) { create(:admin, :manager, :with_access, resource: organization) }
-      let(:power_user) { create(:admin, :power_user) }
+      let(:manager) { create(:admin, :manager, :with_access, resource: organization, receive_approval_notifications: true) }
```
prabhanshuguptagit, line 115: Let's add a `receive_approval_notifications { true }` to the users factory also.

app/views/admins/_form.html.erb
```diff
@@ -50,6 +50,17 @@
...
+        <%= form.check_box(
+                :receive_approval_notifications,
+                {
+                    checked: admin.receive_approval_notifications,
+                    id: :receive_approval_notifications,
+                    label: "Receive approval emails"
+                }
+            ) %>
+      </div>
+
```
prabhanshuguptagit, line 63: Shall we also allow setting this field from the Invite admin page? It might be a convenient place for the inviter (eg. Praveen) to set the field for other folks in advance.

app/models/user.rb
```diff
@@ -60,6 +60,7 @@ def app_capabilities
     class_name: "Facility",
     join_table: "facilities_teleconsultation_medical_officers"
   has_many :accesses, dependent: :destroy
+  has_one :settings, class_name: "UserSettings"
```
prabhanshuguptagit, line 63: 👾

app/models/user.rb
```diff
@@ -232,10 +234,6 @@ def self.requested_sync_approval
     where(sync_approval_status: :requested)
   end
 
-  def has_role?(*roles)
```
prabhanshuguptagit, line 235: Curious, what was this 🤔 ?

db/migrate/20210220065508_add_receive_approval_notifications_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddReceiveApprovalNotificationsToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :receive_approval_notifications, :boolean, null: false, default: true
```
prabhanshuguptagit, line 3: Need to exclude this field from sync APIs. I think we only send users in the teleconsult MOs endpoint.

prabhanshuguptagit (changes requested):
Looks great and simple ✨ . Not adding a settings page sounds right, lets run this by daniel/claudio once too.

app/views/admins/_form.html.erb
harimohanraj89: Oh, I mistakenly thought this form would address both, and I didn't need to make any further controller updates. I'll go update the invitations controller to accept this param.

app/models/user.rb
harimohanraj89: A quick `blame` points to an initial permissions system. Looks like when we used to use `User#role` more formally, this was a convenience method for authorization checks. Search for `has_role?` in [this PR](https://github.com/simpledotorg/simple-server/pull/643/files)

We don't use `role` formally anymore, and this method isn't invoked anywhere anymore, so I've trashed i.t

db/migrate/20210220065508_add_receive_approval_notifications_to_users.rb
harimohanraj89: Ah, interesting. Will do. Should our API specs fail when unexpected fields are added to the response?

harimohanraj89: Our "transformer" sends a whitelist of attributes, so we are all good. I've updated the transformer spec as well to perform a strict check. If something gets in there accidentally, the spec should fail.

app/views/email_authentications/invitations/new.html.erb
```diff
@@ -9,6 +9,9 @@
       <%= form.text_field :full_name, id: :full_name, required: true, autocomplete: "off", label: "Full name *", placeholder: "Full name" %>
       <%= form.text_field :email, id: :email, required: true, label: "Email *", placeholder: "Email" %>
       <%= form.text_field :role, id: :role, required: true, help: "CVHO, STS, State Official etc.", label: 'Job title *', placeholder: "Job title" %>
+      <div class="mt-3 mb-3">
+        <%= form.check_box :receive_approval_notifications, checked: true, value: "1", label: "Receive approval notifications" %>
```
prabhanshuguptagit, line 13: This helper can take a `checked` value and an `unchecked` value like in https://github.com/simpledotorg/simple-server/blob/4a3148e37bf12b615635ec9cae0ab547bf3febb5/app/views/admin/facility_groups/_form.html.erb#L83

Don't need the `== "1"` check if we do this.

prabhanshuguptagit (approved):
:shipit:

app/views/email_authentications/invitations/new.html.erb
harimohanraj89: I just discovered that `true, false`, is actually incorrect 😓  I think we need to specify `"true", "false"` instead. The facility group form operates on Javascript so this isn't an issue, but the Rails form doesn't submit the unchecked value if we set it to `false`. Check out why [here](https://github.com/rails/rails/blob/44a9aedd7b8d65517b15bbbb7729f3f16991e23f/actionpack/lib/action_view/helpers/form_helper.rb#L1100).

In any case, I'll give `"true", "false"` a whirl!

harimohanraj89: It works great 🚀

harimohanraj89: Actually, we don't need to set values at all! "0" and "1" are designed to be assigned to boolean AR models already 🤦  #magic

PR #2097, opened 2021-02-23
Bug: Dashboard period range

claudiovallejo:
**Story card:** [ch2756](https://app.clubhouse.io/simpledotorg/story/2756/dashboard-period-range)

## Because

The dates displayed in the Reports "Overview" and "Details" cards are inaccurate. For a month period of `1-Jul-2020` the `blood_pressure_control_range` will return the range of `30-May-2020..31-Jul-2020`.

## This addresses

* Sets `blood_pressure_control_range.begin` to be the end of the month, three months ago
* Creates a new `@period` method, `bp_control_range_registration_date`, which equals `blood_pressure_control_range.begin`
* Updates the "Visit details of all hypertension patients" "patients registered till" date to `@period.bp_control_range_registration_date`
* Updates the "BP controlled", "BP not controlled", and "Missed visits" "patients registered till" date to `@period.bp_control_range_registration_date`

## How to test this PR

**Test with `report_with_exclusions` flag `disabled` and `enabled`:**
_Go through the checklist below for a state, district, block, or facility report_

_In the Reports "Overview" section with `Feb-2021` period selected..._
- [x] The "BP controlled" card's "patients registered till" date is `30-Nov-2020`
- [x] Hover over `Jan-2021` in the "BP controlled" graph. The "patients registered till" date is `31-Oct-2020`
- [x] The "BP not controlled" card's "patients registered till" date is `30-Nov-2020`
- [x] Hover over `Jan-2021` in the "BP not controlled" graph. The "patients registered till" date is `31-Oct-2020`
- [x] The "Missed visits" card's "patients registered till" date is `30-Nov-2020`
- [x] Hover over `Jan-2021` in the "Missed visits" graph. The "patients registered till" date is `31-Oct-2020`
- [x] The "Missed visits" "patients registered till" date in "Visit details of all assigned hypertension patients" card is `30-Nov-2020`
- [x] Hover over the `Jan-2021` bar chart in the "Visit details of all assigned hypertension patients" card. The "patients registered till" date for "Missed visits" is `31-Oct-2020`
- [x] The "Visit but no BP taken" "patients registered till" date in "Visit details of all assigned hypertension patients" card is `30-Nov-2020`
- [x] Hover over the `Jan-2021` bar chart in the "Visit details of all assigned hypertension patients" card. The "patients registered till" date for "Visit but no BP taken" is `31-Oct-2020`
- [x] The "BP not controlled" "patients registered till" date in "Visit details of all assigned hypertension patients" card is `30-Nov-2020`
- [x] Hover over the `Jan-2021` bar chart in the "Visit details of all assigned hypertension patients" card. The "patients registered till" date for "BP not controlled" is `31-Oct-2020`
- [x] The "BP controlled" "patients registered till" date in "Visit details of all assigned hypertension patients" card is `30-Nov-2020`
- [x] Hover over the `Jan-2021` bar chart in the "Visit details of all assigned hypertension patients" card. The "patients registered till" date for "BP controlled" is `31-Oct-2020`

_In the Reports "Overview" section with `Sep-2020` period selected..._
- [x] The "BP controlled" card's "patients registered till" date is `30-Jun-2020`
- [x] Hover over `Aug-2020` in the "BP controlled" graph. The "patients registered till" date is `31-May-2020`
- [x] The "BP not controlled" card's "patients registered till" date is `30-Jun-2020`
- [x] Hover over `Aug-2020` in the "BP not controlled" graph. The "patients registered till" date is `31-May-2020`
- [x] The "Missed visits" card's "patients registered till" date is `30-Jun-2020`
- [x] Hover over `Aug-2020` in the "Missed visits" graph. The "patients registered till" date is `31-May-2020`
- [x] The "Missed visits" "patients registered till" date in "Visit details of all assigned hypertension patients" card is `30-Jun-2020`
- [x] Hover over the `Aug-2020` bar chart in the "Visit details of all assigned hypertension patients" card. The "patients registered till" date for "Missed visits" is `31-May-2020`
- [x] The "Visit but no BP taken" "patients registered till" date in "Visit details of all assigned hypertension patients" card is `30-Jun-2020`
- [x] Hover over the `Aug-2020` bar chart in the "Visit details of all assigned hypertension patients" card. The "patients registered till" date for "Visit but no BP taken" is `31-May-2020`
- [x] The "BP not controlled" "patients registered till" date in "Visit details of all assigned hypertension patients" card is `30-Jun-2020`
- [x] Hover over the `Aug-2020` bar chart in the "Visit details of all assigned hypertension patients" card. The "patients registered till" date for "BP not controlled" is `31-May-2020`
- [x] The "BP controlled" "patients registered till" date in "Visit details of all assigned hypertension patients" card is `30-Jun-2020`
- [x] Hover over the `Aug-2020` bar chart in the "Visit details of all assigned hypertension patients" card. The "patients registered till" date for "BP controlled" is `31-May-2020`


app/views/reports/regions/show.html.erb
```diff
@@ -364,7 +364,7 @@
                   <p class="m-0px c-black c-print-black">
                     total <%= "registration".pluralize(@data.last_value(:cumulative_registrations)) %><br>till
                     <span data-registrations-period-end="<%= @period.bp_control_range_end_date %>">
-                      <%= @period.bp_control_range_end_date %>
+                      <%= Date.today.to_s(:day_mon_year) %>
```
rsanheim, line 367: This won't be correct if they have chosen a date in the past from the dropdown.

app/views/reports/regions/_visit_details_card.html.erb
```diff
@@ -43,7 +43,7 @@
                   </span>
                   patients registered till
                   <span data-period-start="<%= @period.bp_control_range_start_date %>">
-                    <%= @period.bp_control_range_start_date %>
+                    <%= @period.blood_pressure_control_range.begin.to_s(:day_mon_year) %>
```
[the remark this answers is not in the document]
claudiovallejo: @rsanheim the `bp_control_range_start_date` doesn't return the right date. It returns `blood_pressure_control_range.begin.next_day.to_s(:day_mon_year)`. Should I create a new method to return the denominator start date?

app/views/reports/regions/show.html.erb
claudiovallejo: That's a good point @rsanheim. should I check if the period is the current period? If it isn't I set the `end_date`, if it is, I set the current date?

app/views/reports/regions/_visit_details_card.html.erb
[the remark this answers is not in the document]
claudiovallejo: I created a new method, `bp_control_range_registrations_date`

app/views/reports/regions/show.html.erb
claudiovallejo: I decided to use a conditional to check the period selected to show either the current date or the end of the month of the selected date

app/views/reports/regions/_visit_details_card.html.erb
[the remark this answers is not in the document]
rsanheim: Oh right, that new method makes sense. 👍

app/views/reports/regions/show.html.erb
```diff
@@ -364,7 +364,11 @@
                   <p class="m-0px c-black c-print-black">
                     total <%= "registration".pluralize(@data.last_value(:cumulative_registrations)) %><br>till
                     <span data-registrations-period-end="<%= @period.bp_control_range_end_date %>">
-                      <%= @period.bp_control_range_end_date %>
+                      <% if @period.value.strftime("%m") == Date.today.strftime("%m") %>
+                        <%= Date.today.to_s(:day_mon_year) %>
+                      <% else %>
+                        <%= @period.bp_control_range_end_date %>
+                      <% end %>
```
rsanheim, line 371: I think this can be simplified a bit ...its probably worth moving to a helper method as well:

```ruby
module DashboardHelper
  def registrations_end_date
    if @period.value.month == Date.current.month
      <%= Date.current.to_s(:day_mon_year)
    else
      @period.bp_control_range_end_date
    end
  end
```

Note that you probably want to use `Date.current`, and not `Date.today`, because of [timezones](https://thoughtbot.com/blog/its-about-time-zones).

kpethtel (approved):
I did the testing except test 3 because we talked in slack and you said the feature flag tests aren't necessary, and I don't see any errors (except the one date that says 2021 in your test steps, which I assume is a typo). But I'd recommend having @harimohanraj89 review it since you worked on it with him and I'm not as familiar with where else the problem could be seen.

app/views/reports/regions/show.html.erb
claudiovallejo: Thanks @rsanheim!

app/helpers/dashboard_helper.rb
```diff
@@ -68,4 +68,12 @@ def six_month_rate_change(facility, rate_name)
   def facility_size_six_month_rate_change(facility_size_data, rate_name)
     facility_size_data[@period][rate_name] - facility_size_data[@start_period][rate_name] || 0
   end
+
+  def registrations_end_date
+    if @period.value.month == Date.current.month
```
harimohanraj89, line 73: `.month` just gives you a number between 1 and 12, so this method will behave unexpectedly if we look at, for example, last February's report today.

Can we, instead, do something like this:

```ruby
if @period.end_date > Date.current
  Date.current.to_s(:day_mon_year)
else
  @period.bp_control_range_end_date
end
```

app/models/reports/result.rb
```diff
@@ -180,9 +180,10 @@ def calculate_period_info(range)
         hsh[period] = {
           name: period.to_s,
           start_date: period.start_date.to_s(:day_mon_year),
-          ltfu_since_date: period.start_date.advance(months: -12).to_s(:day_mon_year),
+          ltfu_since_date: period.end_date.advance(months: -12).to_s(:day_mon_year),
```
harimohanraj89, line 183: This is an unrelated fix, right? Nice catch.

app/views/reports/regions/_patient_breakdown_charts.html.erb
```diff
@@ -41,8 +41,8 @@
                   <%= number_with_delimiter(@chart_data[:ltfu_trend].last_value(:cumulative_assigned_patients)) %>
                 </span>
                 patients registered till
-                <span data-registrations-period-end="<%= @chart_data[:ltfu_trend].last_value(:period_info)["start_date"] %>">
-                  <%= @chart_data[:ltfu_trend].last_value(:period_info)["start_date"] %>
+                <span data-registrations-period-end="<%= @period.bp_control_range_end_date %>">
+                  <%= @period.bp_control_range_end_date %>
```
harimohanraj89, line 45: Does this ensure that the "registered until" date on the breakdown charts always matches the one in the top-4 cards?

harimohanraj89 (dismissed):

harimohanraj89:
I found the following when clicktesting the review app.

**Today's date is not shown in LTFU cards for current month**

The LTFU cut-off (ticked in green) looks good to me. Should we show today's date (eg. Feb 25, 2021) in the red-underline places though?

Screen Shot 2021-02-25 at 11 52 50 AM

---

app/models/reports/result.rb
prabhanshuguptagit: This should be the `start_date`, no? The LTFU cards show data as of the beginning of the current month (`1-Feb-2021` instead `28-Feb-2021`).

prabhanshuguptagit: Should this be `period.bp_control_end_date.advance...`? The LTFU patients we shown are as of `1 Feb 2021` and not `28 Feb 2021`. We want the end of the previous month here, i.e `31 Jan 2021` I think.


Let's also rename this to something like `ltfu_bp_not_taken_since_date` since this name suggests these were lost to follow up an year ago, which means they didn't have a BP since 2 years.😅

harimohanraj89: @prabhanshuguptagit can you help confirm which date the query's using? Once we know, we can make sure:
* It's the intended date
* It's consistent with the copy

prabhanshuguptagit: The query returns patients ltfu as of `period.start_date`. So if we are looking at the chart in february (say Feb 15th):
- Patients breakdown is as of Feb 1
- Last data point on LTFU chart is as of Feb 1

app/assets/javascripts/common/reports.js
```diff
@@ -410,21 +410,29 @@ Reports = function () {
...
         } else {
           label = mostRecentPeriod;
         }
         const period = data.periodInfo[label];
         const cumulativeRegistrations = data.cumulativeRegistrations[label];
         const monthlyRegistrations = data.monthlyRegistrations[label];
 
+        const setRegistrationEndDate = dataPointIndex == totalRegistrationDataPoints || tooltip.dataPoints == undefined
+          ? registrationsEndDate
+          : period.bp_control_end_date;
```
prabhanshuguptagit, line 431: When is `tooltip.dataPoints` undefined? Is it in the case of no data: Do we separately need to handle that case? Trying to figure out if this can be simplified.

prabhanshuguptagit (commented):
@claudiovallejo I click tested the region reports which look great. Left some thoughts about the details page.

app/models/reports/result.rb
harimohanraj89: Thanks for the clarification @prabhanshuguptagit .

@claudiovallejo , that means the change you've made above makes the copy inconsistent with the query. Let's:
* Revert this change for now, to ensure the copy is accurate
* Find out what the desired LTFU behavior is (if we haven't already)
  * Should the Jan 2021 report show an LTFU breakdown as of Jan 1, or Jan 31?
  * If we're looking at the current month (Feb 2021), is it ok for our LTFU query to run from Feb 29, 2020 - Feb 26 (today) 2021?
* Update the copy and query together if needed.

app/models/period.rb
```diff
@@ -53,12 +53,16 @@ def to_quarter_period
...
   # For example, for a month period of July 1st 2020, this will return the range of April 30th..July 31st.
   def blood_pressure_control_range
-    three_months_ago = end_date.advance(months: -3)
+    three_months_ago = end_date.advance(months: -3).end_of_month
     (three_months_ago..end_date)
   end
 
   alias_method :bp_control_range, :blood_pressure_control_range
 
+  def bp_control_range_registration_date
```
prabhanshuguptagit, line 62: Is this the same as `bp_control_range_start_date` but different by one day? I'm wondering if the change on line 56 is all we need we keep using `bp_control_range_start_date`. These two are _very_ similar, I might be missing something.

prabhanshuguptagit (approved):
I click tested these on the review app, looking good👌🏼 . One last question, approving to unblock.

app/models/reports/result.rb
claudiovallejo: Thanks @harimohanraj89 - I reverted the changes and will discuss desired LTFU behavior with Reena and Daniel :)

app/assets/javascripts/common/reports.js
claudiovallejo: Thanks! I removed this change as part of the LTFU revert :)

app/models/period.rb
claudiovallejo: That's right - `bp_control_range_registration_date` is just one day before `bp_control_range_start_date`. I'm creating a new method to make it clear that `bp_control_range` has 3 dates we care about:
* The registration end date
* The start of the BP control range
* The end of the BP control range

How would you implement this?

prabhanshuguptagit: Ah I see. That makes sense now. Maybe `bp_control_registrations_until_date` is a better name since this date is not part of a range really. This is good though 👍🏼.

claudiovallejo: Thanks @prabhanshuguptagit I'll update the name and then merge :)

PR #2098, opened 2021-02-23
Add a feature flag for LTFU toggle

prabhanshuguptagit:
**Story card:** [ch2227](https://app.clubhouse.io/simpledotorg/story/2227/exclude-dead-patients-from-all-necessary-queries)

## Because

We want to allow alpha users of denominator exclusions (folks at ICMR and WHO) to be able to see the LTFU toggle, even though they aren't power users.
image

## This addresses

This adds another feature toggle for the LTFU toggle that will be enabled for the alpha users. We can probably revert to allowing powers users when #2094 is merged.

### DIFF
--- a/app/views/reports/regions/show.html.erb
+++ b/app/views/reports/regions/show.html.erb
@@ -25,7 +25,7 @@
               BP controlled
             </h3>
           <% end %>
-          <% if current_admin.feature_enabled?(:report_with_exclusions) && current_admin.power_user? %>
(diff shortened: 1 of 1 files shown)


harimohanraj89 (approved):
> We can probably revert to allowing powers users when #2094 is merged.

Do you mean, certain users like Dr. P are purposefully keeping their account at a "viewer" level, so they don't get email spam?

prabhanshuguptagit:
> Do you mean, certain users like Dr. P are purposefully keeping their account at a "viewer" level

Yep, praveen changed their access to viewer only because of the email spam.

PR #2099, opened 2021-02-23
2118 live password validation

kpethtel:
**Story card:** [ch2118](https://app.clubhouse.io/simpledotorg/story/2118)

## Because

We're adding new live password validation as the user types. 

## This addresses

This is a stepping stone toward a new password validation process. 

## Testing
Currently, this isn't really wired up to anything in the app, but it can be tested by going to the password resetting partial:
app/views/email_authentications/passwords/edit.html.erb

Then adding this between the <script> tags. 

```
const validator = new PasswordValidation();
validator.initialize();
```

Then log out of the app and click "forgot password". You can get the reset link either from the rails logs or from mailcatcher. From that link, open the inspector and watch the console. You should see a bunch of logging to indicate what the current password is, when a timer is set or canceled, and what the response is from the back end.


config/routes.rb
```diff
@@ -137,6 +137,8 @@
     path: "email_authentications",
     controllers: {invitations: "email_authentications/invitations"}
 
+  post "email_authentications/validate", to: "email_authentications/password_validations#create"
```
kpethtel, line 140: I'm open to feedback on the url path. I left "password" out of the url because I didn't see any reason to draw attention to it. Security through obscurity and all.

app/assets/javascripts/common/password_validation.js
```diff
@@ -0,0 +1,53 @@
+PasswordValidation = function() {
```
kpethtel, line 1: This is still a work in progress. The next ticket will require significant changes to this script. The console logging is intentional because it makes debugging and testing easier. It will be removed in the next ticket.

app/assets/javascripts/common/password_validation.js
```diff
@@ -0,0 +1,53 @@
...
+  const DebounceTimeout = 500;
```
kpethtel, line 2: The timeout exists to avoid hitting the back end with every keystroke. 500ms seemed like a reasonable amount of time to wait between keystrokes, but I may change this in the next ticket if it feels too slow or fast.

ssrihari:
@kpethtel Can we do this validation client side? Using a regex, perhaps?

While keeping the validation server side seems more future-proof, and will not require changes on the front-end, I wonder if the additional complexity here is worth the trade-off. Additional complexity = fair amount of javascript, debouncing, ajax, new API on admin, etc.

If this discussion has been had elsewhere, please feel free to just point me to it.

kpethtel:
> @kpethtel Can we do this validation client side? Using a regex, perhaps?
> 
> While keeping the validation server side seems more future-proof, and will not require changes on the front-end, I wonder if the additional complexity here is worth the trade-off. Additional complexity = fair amount of javascript, debouncing, ajax, new API on admin, etc.
> 
> If this discussion has been had elsewhere, please feel free to just point me to it.

That's a good question. The tickets for this were already written when I got them and it's possible I misunderstood some of the requirements. In this instance, the ticket just had this title and no description: "Wire up live AJAX indicators". I talked to @rsanheim about this at the beginning of the project and asked if we would repeat the same regex validations on the front end and I believe he said the back end would be the source of truth. I recall talking to Rob about this ticket being a new endpoint with a debounce, but you have me questioning if this might have all been a failure to clarify exactly what the plan was. Rob, what do you think?

In response to the complexity issue... If we were to change this to front-end regex, we could get rid of the new validation endpoint, the ajax request, and the debounce, and replace it with a few lines of regex. There would still need to be some javascript for the regex and handling the live updating of the page elements (which is coming in the next ticket). This is definitely more complex but I don't think it's problematically complex. I'm less concerned with the complexity and more concerned with the possibility that it might be a less responsive user experience.

Let's try to come to a group understanding. Regardless of what we choose, this is a good reminder for me that I need to be sure to clearly state what I think the plan is and get group buy-in before I proceed with it.

rsanheim (approved):
Looks like a good start to me. I think the JS can be simplified a bit but we can discuss that on future PRs.

rsanheim:
> @kpethtel Can we do this validation client side? Using a regex, perhaps?
> 
> While keeping the validation server side seems more future-proof, and will not require changes on the front-end, I wonder if the additional complexity here is worth the trade-off. Additional complexity = fair amount of javascript, debouncing, ajax, new API on admin, etc.
> 
> If this discussion has been had elsewhere, please feel free to just point me to it.

We have discussed the alternate approaches earlier, though most of that was done over calls or in slack...so not the most discoverable I admit.

The JS here won't really be bad, rails makes this kind of Ajax stuff pretty easy and straightforward. I think its well worth the cost of some JS to keep all the auth logic on the server side -- that will make it much easier for us to tweak or alter password rules going forward.  And for client side javascript, its never "just a few lines of regex" =). We would also need to duplicate the error messages over to the client, and we have to wire up the UI indication whether we do it client or server side.

We've talked about adding some validation against very easily-guessed passwords, for example, and wiring up some conventions + JS/XHR for the initial rules should make that easy to do.

app/assets/javascripts/common/password_validation.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  this.handlePasswordInput = () => {
+    console.log("CHANGE", this.passwordInput.val());
+    this.setTimer();
+  }
+
+  this.setTimer = function() {
+    console.log("STARTING TIMER")
+    this.cancelTimer();
```
rsanheim, line 18: Some feedback for the follow up PR:

I would suggest using the same [debounce pattern](https://github.com/simpledotorg/simple-server/blob/e906b5efcfa7d28945fcd28f5e593efb149b3e58/app/assets/javascripts/common/regions_search.js#L73-L86) we have elsewhere in the app, I think its more clear and easier to follow than this.

Also, you can use ECMAscript's `=>` and the like to clean up some of the line noise in here.

PR #2102, opened 2021-02-25
English names for Ethiopia states

harimohanraj89:
**Story card:** [ch2782](https://app.clubhouse.io/simpledotorg/story/2782/add-dire-dawa-to-canonical-region-list)

## Because

Facility registry in Ethiopia is in English - so our canonical state names should be as well, to ensure future facility groups have matching states.

### DIFF
--- a/config/data/canonical_states.yml
+++ b/config/data/canonical_states.yml
@@ -51,5 +51,6 @@ Bangladesh:
   - Sylhet
 Ethiopia:
   - Addis Ababa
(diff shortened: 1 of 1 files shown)


prabhanshuguptagit (approved):

harimohanraj89:
We _might_ be adding Dire Dawa as its own region, like Addis. I'm confirming that now, and will update this PR accordingly.

PR #2103, opened 2021-02-25
Add more info to test failure

harimohanraj89:
**Story card:** [ch2783](https://app.clubhouse.io/simpledotorg/story/2783)

Adds some useful failure info to diagnose flapping spec


kpethtel (approved):
This seems like a reasonable approach.

PR #2104, opened 2021-02-25
Merge exact duplicates

ssrihari:
**Story card:** [chXXXX](URL)

## Because

## This addresses


app/services/merge_exact_duplicate_patients.rb
```diff
@@ -0,0 +1,56 @@
...
+                 distinct: duplicate_patient_ids.count},
+     merged: {total: duplicate_patient_ids.flatten.count - merge_failures.flatten.count,
+              distinct: duplicate_patient_ids.count - merge_failures.count,
+              total_failures: merge_failures.count,
+              distinct_failure: merge_failures.flatten.count}}
+  end
+
+  def report
+    Rails.logger.info(report_stats.to_json)
+    Rails.logger.info "Failed to merge patients #{merge_failures}"
```
harimohanraj89, line 40: The logging seems a bit convoluted to me. `merge_failures` is populated in the `rescue` block, reported in `handle_error`, but also reported here? Could we tighten this up with

```ruby
	    duplicate_patient_ids.map do |patient_ids|
      DeduplicatePatients.new(Patient.where(id: patient_ids)).merge
      report
    rescue => e
      # Bad data can cause our merge logic to breakdown in unpredictable ways.
      # We want to report any such errors and look into them on a per case basis.
      handle_error(e, patient_ids)
      report_failures
    end
```

harimohanraj89 (approved):
Tiny suggestion, but overall looks simple ✨

PR #2105, opened 2021-02-25
add a test instructions section to the PR template

kpethtel:
## Because

Adding a test instructions section to the PR template should remind people to add test instructions and help reviewers to do higher quality testing. 

## This addresses

This will hopefully improve the quality of our review process and result in fewer bugs. 

## Test instructions

None.


harimohanraj89 (approved):
I think this is a small and fine addition for now, but I want to discuss the onus of testing with the whole team. So far, we've operated on the principle of "onus of testing lies with author". If we're changing the responsibility of code reviewers, let's talk about it.

kpethtel:
> I think this is a small and fine addition for now, but I want to discuss the onus of testing with the whole team. So far, we've operated on the principle of "onus of testing lies with author". If we're changing the responsibility of code reviewers, let's talk about it.

I'm all for talking about it. And it's good to know where the burden now lies. That wasn't completely clear to me, and I sometimes felt uncomfortable approving a PR without testing it. But we have these nice review environments to test on. It seems like a waste not to use them. Should we add this to the agenda for the next server sync?

harimohanraj89:
Yes we should. Did we talk about it yesterday? :sweat_smile: I missed your PR comment somehow 🤦

PR #2107, opened 2021-02-26
2116 password validation checklist

kpethtel:
**Story card:** [ch2116](https://app.clubhouse.io/simpledotorg/story/2116/update-ui)

## Because

We want to give users live feedback on their password validity. 

## This addresses

This is the final piece in changing our password scheme. 

## Test instructions

1) invitations
- invite a new user through the admin dashboard
- get the invitation link from rails logs or mailcatcher
- go to the page and confirm that the checklist accurately reflects the current state of the password
- confirm that the form button becomes submittable when the password is valid and that submission works

2) changing password
- log out of simple and click "forgot password". enter your user's email address
- get the invitation link from rails logs or mailcatcher
- go to the page and confirm that the checklist accurately reflects the current state of the password
- confirm that the form button becomes submittable when the password is valid and that submission works


app/assets/javascripts/common/password_validation.js
```diff
@@ -1,35 +1,26 @@
...
-    this.timer = null;
+    this.timer = setTimeout(later, DebounceTimeout);
   }
 
   this.validatePassword = () => {
-    console.log("MAKING REQUEST")
-    const token = $("meta[name=csrf-token]").attr("content")
-    const url = "http://localhost:3000/email_authentications/validate"
+    const token = $("meta[name=csrf-token]").attr("content");
+    const url = "http://localhost:3000/email_authentications/validate";
```
harimohanraj89, line 23: Should this be `localhost`?

kpethtel: OMG. I can't believe I forgot to change that. Thank you for pointing this out. I'll fix it.

harimohanraj89: Don't sweat it, that's what code review is for 😄

app/assets/javascripts/common/password_validation.js
```diff
@@ -39,15 +30,41 @@ PasswordValidation = function() {
...
       }
+      this.updateChecklist(response);
+      this.updateSubmitStatus(response);
+    });
+  }
+
+  this.updateChecklist = (response) => {
+    Validations.forEach(validationName => {
+      response.includes(validationName) ? this.uncheckItem(validationName) : this.checkItem(validationName);
     });
```
rsanheim, line 49: I think this may be more clear to just do add/remove classes inline - this is all easy enough to follow and at roughly the same level of abstraction as the `updateSubmitStatus` function:
```suggestion
    Validations.forEach(validation => {
      if(response.includes(validation)) {
        $(`#$(validation)`).removeClass("completed")
      } else {
        $(`#$(validation)`).addClass("completed")  
      }
   });
```

app/assets/javascripts/common/password_validation.js
```diff
@@ -39,15 +30,41 @@ PasswordValidation = function() {
...
-      console.log(data)
-      this.timer = null;
+    }).done((data, status) => {
+      let response;
       if (status === "success") {
-        this.response = data["errors"];
+        response = data["errors"];
       } else {
-        this.response = null;
+        // if we don't get a response, mark all validations as failures
```
rsanheim, line 38: This means if the server returns an error because of some other failure, we will display it as a password error to the user?  It may be better to show a generic error to the error if that happens, because it would mean something is wrong on the server side (a 500 error for example).  Also, a stale session would result in this case, because the CSRF token would've expired...we could even just show some sort of generic error w/ the HTTP code so users could pass it along if they are trying to get help from support:

```
An error occurred while trying to validate your password (HTTP 500)
```

app/assets/javascripts/common/password_validation.js
```diff
@@ -39,15 +30,41 @@ PasswordValidation = function() {
         "X-CSRF-Token": token
       },
       data: {"password": password}
-    }).done(function(data, status){
-      console.log(status)
-      console.log(data)
-      this.timer = null;
+    }).done((data, status) => {
+      let response;
```
rsanheim, line 34: This is really `errors`, not the response...

app/assets/javascripts/common/password_validation.js
```diff
@@ -39,15 +30,41 @@ PasswordValidation = function() {
...
       if (status === "success") {
-        this.response = data["errors"];
+        response = data["errors"];
       } else {
-        this.response = null;
+        // if we don't get a response, mark all validations as failures
+        response = Validations;
       }
+      this.updateChecklist(response);
+      this.updateSubmitStatus(response);
```
rsanheim, line 42: ...and here, we are actually passing in an array of errors _from_ the response, not the response itself. Would be good to name this `errors` to be explicit about what data we are working with.

app/assets/javascripts/common/password_validation.js
```diff
@@ -1,35 +1,26 @@
...
     clearTimeout(this.timer);
-    this.timer = null;
+    this.timer = setTimeout(later, DebounceTimeout);
   }
 
   this.validatePassword = () => {
-    console.log("MAKING REQUEST")
-    const token = $("meta[name=csrf-token]").attr("content")
-    const url = "http://localhost:3000/email_authentications/validate"
+    const token = $("meta[name=csrf-token]").attr("content");
```
rsanheim, line 22: 👍 looks like the csrf token was pretty straight forward

app/assets/javascripts/common/password_validation.js
kpethtel: Done.

app/assets/javascripts/common/password_validation.js
kpethtel: Done.

app/assets/javascripts/common/password_validation.js
kpethtel: Done.

app/assets/javascripts/common/password_validation.js
kpethtel: I'll talk to Claudio about that since it will require some visual changes. Thankfully, I wrote a small ticket that's coming after this for the case when the reset link is expired, so we just need to add some new copy. 

As for the nature of the error... I think the CSRF expiring would be a very edge case since I think it would mean the person has been sitting on the password setting page for hours. But your point about handling the unknown possibilities better is a good one. TBH, I was working more on the assumption that the request above might fail due to connectivity issues, in which case the validations would just go back to "everything is invalid" and their next keystroke might go through and everything would work again. The downside to redirecting to an error page upon error here is that a single errant request effectively ends the attempt. I doubt this is a major concern.

rsanheim (approved):
I have not tested this locally, but code LGTM!  The error handling and validation stuff is straight forward and easy to follow.

PR #2109, opened 2021-03-02
Patient breakdown: Show current month data always

prabhanshuguptagit:
**Story card:** [ch2760](https://app.clubhouse.io/simpledotorg/story/2760)

## Because

If a user selects "Sep-2020" in the "Overview" tab, the data displayed in the "Details" section for the new LTFU cards also shows data for "Sep-2020". [Slack ref](https://simpledotorg.slack.com/archives/CFHKJ5WJY/p1614151896054500?thread_ts=1614117582.051200&cid=CFHKJ5WJY) We should only show data as of the current month for patient breakdowns currently.

## This addresses

This changes the LTFU query to always fetch data for the current month.


claudiovallejo (approved):
This looks good, thanks @prabhanshuguptagit !

PR #2110, opened 2021-03-02
Seed prescription drug history for each patient

timcheadle:
**Story card:** [ch2548](https://app.clubhouse.io/simpledotorg/story/2548/add-prescription-drugs-to-patients)

## Because

We need prescription drug history to test things like the treatment inertia and protocol adherence reports

## This addresses

This adds basic protocol drug histories for patients in our seed data

### DIFF
--- a/lib/seed/blood_pressure_seeder.rb
+++ b/lib/seed/blood_pressure_seeder.rb
@@ -80,7 +80,6 @@ module Seed
       counts[:blood_pressure] = result.ids.size
 
       encounters = []
-      prescription_drugs = []
       result.results.each do |row|
         bp_id, recorded_at, patient_id = *row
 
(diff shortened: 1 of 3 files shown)


lib/seed/blood_pressure_seeder.rb
```diff
@@ -94,7 +96,24 @@ def call
           updated_at: recorded_at,
           timezone_offset: 0
         }
+
+        prescription_drugs << {
+          id: SecureRandom.uuid,
+          name: "Amlodipine",
```
rsanheim, line 102: Looks like we need to extract the common drug names / dosages from the protocol factory into a seed object somewhere, and then you could grab a random one here easily.  

something like...
```
drug = Seed::ProtocolDrugs.instance.sample
name = drug.name
dosage = drug.dosage
xnorm_code = drug.code
```

lib/seed/blood_pressure_seeder.rb
```diff
@@ -108,8 +127,10 @@ def call
 
       encounter_result = Encounter.import(encounters)
       observation_result = Observation.import(observations)
+      prescription_drug_result = PrescriptionDrug.import(prescription_drugs)
       counts[:encounter] = encounter_result.ids.size
       counts[:observation] = observation_result.ids.size
+      counts[:prescription_drug] = prescription_drug_result.ids.size
```
rsanheim, line 133: You can assert on this in a spec to make sure these are created.  We have a `test` mode for seeding so that everything is deterministic, lemme know if you need help w/ that piece.

PR #2112, opened 2021-03-03
Update our deployment docs for how things work today

rsanheim:
This was a little out of date, and didn't mention the fact that sandbox gets auto deploys on master merges.


kpethtel (approved):
Thanks for clarifying. 👍

PR #2115, opened 2021-03-03
2698 reset expiration

kpethtel:
**Story card:** [ch2783](https://app.clubhouse.io/simpledotorg/story/2783/flapping-test-invitations-controller)

## Because

It's confusing for a user if they're trying to reset a password and they can't because their reset token is expired. This probably doesn't impact many users, but it did impact me a couple of times while I was working on password resets. 

## This addresses

Currently, if a user's password reset token is expired, the password reset page will not in any way inform them and will silently fail. 

## Testing

- It can be helpful to change the devise config `  config.reset_password_within = 6.hours` to `1.second` to ensure that your token is expired. 
- go to reset link
- enter valid password
- submission should route you to the error page

It would be a better user experience if the link went directly to the error page rather than requiring submission, but I don't think that's possible without subclassing the devise controller because the validated EmailAuthentication in the view doesn't have any reset token errors. Subclassing the controller wouldn't be terribly difficult but would probably require learning some devise internals and I didn't think it was worth it in this case.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -693,3 +693,40 @@ span.inactive {
 	height: 43px;
 	width: 141px;
 }
+
+.error-body {
+    background-color: #F0F2F5;
+    color: #2E2F30;
+    text-align: center;
+    font-family: Roboto, sans-serif;
+    margin: 0;
+    padding: 20px;
+    line-height: 1.4;
+
+    a:link,
+    a:hover,
+    a:visited {
+        font-weight: bold;
+        color: #0075EB;
+        letter-spacing: 2px;
+        text-decoration: none;
+    }
+
+    img {
+        width: 80px;
+        margin-bottom: 2em;
+    }
+
+    .dialog {
+        max-width: 33em;
+        margin: 4em auto 0;
+        background: #fff;
+        padding: 40px;
+        border-radius: 7px;
+        box-shadow: 0 2px 6px rgba(0,0,0,0.05);
+    }
+
+    .text {
+        padding-bottom: 2em;
+    }
+}
--- a/app/views/email_authentications/passwords/_expired_reset_token.html.erb
+++ b/app/views/email_authentications/passwords/_expired_reset_token.html.erb
@@ -1,0 +1,12 @@
+<body class="error-body">
+  <div class="dialog">
+    <div class="text">
+      <img src="http://simple.org/images/simple-wordmark-standard.svg">
+      <h1>Reset link expired</h1>
+      <p>
+        The password link expired. <%= link_to "Click here", new_email_authentication_password_path %> to request a new link.
+      </p>
+    </div>
+    <%= link_to "RETURN HOME", "/" %>
+  </div>
+</body>
--- a/app/views/email_authentications/passwords/edit.html.erb
+++ b/app/views/email_authentications/passwords/edit.html.erb
@@ -1,4 +1,4 @@
-<% unless resource.reset_password_period_valid? == false # explicitly check false because it can be nil %>
+<% unless resource.reset_password_period_valid? == false %>
   <h1><%= t("devise.password_creation.reset_header") %></h1>
 
   <div class="row">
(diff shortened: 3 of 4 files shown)


app/views/email_authentications/passwords/_expired_reset_token.html.erb
```diff
@@ -0,0 +1,12 @@
+<body class="error-body">
```
harimohanraj89, line 1: I don't think we should be adding a `body` tag here, we'll already have one from the layout.

app/assets/stylesheets/application.scss
```diff
@@ -692,4 +692,41 @@ span.inactive {
 	font-size: 16px;
 	height: 43px;
 	width: 141px;
-}
\ No newline at end of file
+}
+
+.error-body {
```
harimohanraj89, line 697: How much custom styling should we be doing? Are there reusable bits (like the white card on the login screen) that we can use instead?

I'm thinking something like our 404 page - https://api.simple.org/not/found - but we can leverage our existing Rails style components.

app/views/email_authentications/passwords/edit.html.erb
```diff
@@ -1,9 +1,10 @@
-<h1><%= t("devise.password_creation.reset_header") %></h1>
+<% unless resource.reset_password_period_valid? == false %>
```
harimohanraj89, line 1: Consider `<% if resource.reset_password_period_valid? %>`

app/views/email_authentications/passwords/edit.html.erb
```diff
@@ -21,17 +22,20 @@
...
-</script>
\ No newline at end of file
+  <script>
+    new PasswordVisibilityToggle();
+    const validator = new PasswordValidation();
+    validator.initialize();
+  </script>
+<% else %>
+  <%= render partial: "expired_reset_token" %>
+<% end %>
```
harimohanraj89, line 41: Rather than a big conditional in the view, I think overriding the controller is better - and fairly straightforward too.

We can introduce an `EmailAuthentications::PasswordsController` with:
```ruby
def edit
  if resource.reset_password_period_valid?
    super
  else
    render :expired_reset_token
  end
end
```

And we can convert the `expired_reset_token` partial into a view.

harimohanraj89 (changes requested):
Some thoughts below. After our approach has settled down, let's add some test coverage for this as well!

app/views/email_authentications/passwords/_expired_reset_token.html.erb
kpethtel: You are probably correct there. I should've made this a draft PR because I had planned to go over styling with Claudio today. The layout was causing some slight issues, though I don't think it was because of this body tag.

app/assets/stylesheets/application.scss
kpethtel: The styling is directly stolen from the 404 page. If we weren't working async, I would've asked you about moving that styling from the 404 page into the general css, but the fact that all of the static html pages keep their styling separate gave me pause and I didn't want to hold up work until I could get a response from you. I briefly went down a path of moving the styling out and creating a common layout to use for this and the 404 page but ran into issues with the embedded link and decided I was over-complicating things.

app/views/email_authentications/passwords/edit.html.erb
kpethtel: Yeah. I wanted to keep the main case first. But I do agree with you that unless/else is poor form.

app/views/email_authentications/passwords/edit.html.erb
kpethtel: Music to my ears. I thought I would get pushback on subclassing the controller, but now that I have some buy-in... You may have noticed my comment in the "testing" instructions. The current flow is not ideal because you can't know if the link is invalid until the use posts. This should (hopefully) allow me to fix that. I'll give it a try. Thanks!

kpethtel:
Closing in favor of a new branch.

PR #2116, opened 2021-03-04
Add CustomMedication model

vkrmis:
**Story card:** [ch2825](https://app.clubhouse.io/simpledotorg/story/2825/add-custom-medication-data-model)

## Because

We need to add a list of custom medications for the app to suggest. This is expected to be a relatively static list, backed by a config file rather than the database

## This addresses

Adds a CustomMedication model, uses [ActiveHash](https://github.com/zilkey/active_hash) to populate this model from a static YAML file

### DIFF
--- a/app/models/custom_medication.rb
+++ b/app/models/custom_medication.rb
@@ -1,0 +1,4 @@
+class CustomMedication < ActiveYaml::Base
+  set_root_path "config/data"
+  set_filename "custom_medications"
+end
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ end
 
 gem "dotenv-rails"
 
-gem 'active_hash', '~> 2.3.0'
+gem "active_hash", "~> 2.3.0"
 gem "active_record_union"
 gem "activerecord-import"
 gem "amazing_print"
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ end
 
 gem "dotenv-rails"
 
+gem 'active_hash', '~> 2.3.0'
 gem "active_record_union"
 gem "activerecord-import"
 gem "amazing_print"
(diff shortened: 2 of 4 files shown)


app/models/custom_medication.rb
```diff
@@ -0,0 +1,4 @@
+class CustomMedication < ActiveYaml::Base
```
rsanheim, line 1: I may be missing something - but why wouldn't we use our existing ProtocolDrug model along w/ some seed generation code that reads from the yml you have in this PR?

I'm sure that the med details or the 'canonical' meds will change over time, so putting them all into a static YML file feels short sighted when we have a domain model and the models in our domain for all this.

rsanheim: ...one follow up thought: it may be that these aren't "protocol drugs", but they are still "prescription drugs" that we can figure out how to model along side the protocol drugs. Maybe this is a subset of some new "prescription" model that we are still trying to figure out?  In any case, I think it would be worth exploring what improving our existing ActiveRecord based drug models look like rather than trying to load and update things from yaml.

vkrmis: @rsanheim These are specifically non-protocol drugs that are currently entered by hand. ProtocolDrugs are not a good fit to model this.

PRD is here: https://docs.google.com/document/d/1lL43EJUxmnr7dsFjv4kyNnEyiGGbD7__4HbD_EUqpZw/edit#

vkrmis: I agree that unifying our drug model so we can mark some drugs as protocol and others as custom seems like a better solution overall, but that'll likely be a significantly larger change.

vkrmis:
Closing in favour of a consolidated Drug model (new PR coming)

PR #2118, opened 2021-03-04
2783: update password generation in specs and invitation controller t…

kpethtel:
**Story card:** [ch2783](https://app.clubhouse.io/simpledotorg/story/2783/flapping-test-invitations-controller)

## Because

We recently updated password requirements but missed a couple of places where passwords are generated within the app. 

## This addresses

- password generation within spec factories. The algorithm currently within the factories would probably never fail but I made it programmatically impossible for it to fail. 
- password generation within the invitations controller. This was the real problem and would have resulted in 500 errors within the app. I believe the error rate was below 5% on a low volume endpoint, which is why it hasn't been an issue yet. 

## Testing

The CI suite should be good enough.


app/models/email_authentication.rb
```diff
@@ -24,4 +24,11 @@ class EmailAuthentication < ApplicationRecord
   validates :password,
     format: {with: /(?=.*\d)/, error: :needs_number, message: "must contain at least one number"},
     allow_nil: true
+
+  def self.generate_password
```
kpethtel, line 28: I moved this to more closely associate the logic with the model and make it less likely it would be missed again in the future. 
I'm open to other solutions for how to generate a guaranteed valid password. I didn't want to mess around with retries. I tried this 1,000,000 times (using alphanumeric instead of base64 to improve the odds and adding a random number to eliminate the missing number issue that was the most frequent failure) and still got 4 errors:
`SecureRandom.alphanumeric(24) + SecureRandom.random_number(1).to_s`

spec/factories/users.rb
```diff
@@ -53,7 +53,7 @@
   end
 
   sequence(:strong_password) do |n|
-    [Faker::Name.first_name, Faker::Name.last_name, Faker::Internet.domain_word, n].join("-")
```
kpethtel, line 56: This probably couldn't have failed the new validations, but I'd still prefer a programmatically impossible failure.

app/models/email_authentication.rb
harimohanraj89: No, I really like this. Perfect KISS example.

You had also suggested (in Slack I think) that this password doesn't need to be random. Is that true? Is it used so temporarily that it can be a static string, because it's overridden anyway? Or does that make us feel a bit too uncomfortable from a security standpoint?

kpethtel: If I said that, I don't think I meant to. But I did just test that theory and it looks like the temporary password is not usable until you accept the invitation by setting a setting a new password. Regardless, I don't think it's worth changing to a static password approach. Who knows what holes might exist in devise.

harimohanraj89: Yeahhh my worry as well .Especially since our codebase is open source.

harimohanraj89 (approved):
Love it 👍 Let's remove the extra logging from the `invitations_controller_spec` as well.

PR #2120, opened 2021-03-04
2698 subclass devise password expiration

kpethtel:
**Story card:** [ch2698](https://app.clubhouse.io/simpledotorg/story/2698)

## Because

For now this is just a draft PR.


app/controllers/email_authentications/passwords_controller.rb
```diff
@@ -0,0 +1,7 @@
+class EmailAuthentications::PasswordsController < Devise::PasswordsController
+  def edit
+    users_email_authentication = resource_class.with_reset_password_token(params["reset_password_token"])
```
kpethtel, line 3: It's tempting to make a pull request on devise for this change. I don't think it would be approved because it would probably cause trouble for existing apps, but it seems like the better flow to me.

harimohanraj89: It's tempting indeed 😅 You could always introduce a configurable option, and default it to the current behavior. That would leave existing apps untouched. I'm curious if folks have tried opening a PR for this before. It seems like a fairly common user flow - let the user know early that their attempt won't work.

app/views/email_authentications/passwords/expired_reset_token.html.erb
```diff
@@ -0,0 +1,12 @@
...
+    <div class="text">
+      <img src="http://simple.org/images/simple-wordmark-standard.svg">
+      <h1>Reset link expired</h1>
+      <p>
+        The password link expired. <%= link_to "Click here", new_email_authentication_password_path %> to request a new link.
+      </p>
+    </div>
+    <%= link_to "RETURN HOME", "/" %>
+  </div>
+</div>
```
harimohanraj89, line 12: I took a stab at re-using our existing components and styles for this page.

```erb
<div class="col-md-6 mx-auto">
    <div class="card">
        <div class="row">
            <div class="col">
                <h1>Reset link expired</h1>
            </div>
            <div class="col text-right login-logo">
                <%= logo_for_environment %>
            </div>
        </div>
        <div class="row">
          <div class="col">
            <p>
              The password link expired. <%= link_to "Click here", new_email_authentication_password_path %> to request a new link.
            </p>
          </div>
        </div>
    </div>
    <div class="actions mt-4" style="line-height: 2.5em;">
      <%= link_to "Return to home", "/" %>
    </div>
</div>
```

Fancy taking this for a spin? If you like it, we can remove the custom styles we've added. There may be more available components (either from Bootstrap or from our existing styles) that we can use. I'd run this by Claudio to get his 👍 before we ship a final version.

harimohanraj89 (approved):
Devise changes look good! Outstanding items are:
* Styles - I've left some thoughts and suggestions, I'll let you take a final call with Claudio
* Standard - Some standardrb deviations. You should be able to quickly fix em all with a `$ bundle exec standardrb --fix`

app/views/email_authentications/passwords/expired_reset_token.html.erb
kpethtel: I met with Claudio about this and he wanted to keep the 404 styling in the app, so we've worked on it and tried to match the 404 page as closely as could with alongside some changes he requested.

app/views/email_authentications/passwords/expired_reset_token.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="error-body">
+  <div class="dialog">
+    <img src="http://simple.org/images/simple-wordmark-standard.svg">
+    <div class="text">
+      <p class="title">Reset link expired</p>
+      <p>
+        The password link expired. <%= link_to "Click here", new_email_authentication_password_path %> to request a new link.
+      </p>
+    </div>
+    <%= link_to "RETURN TO LOGIN PAGE", "/", class: "return-link" %>
```
rsanheim, line 10: Should this be all caps?

spec/controllers/email_authentications/passwords_controller_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      auth = build(:email_authentication)
+      token = auth.send_reset_password_instructions
+      put :edit, params: {reset_password_token: token}
+      expect(response).to render_template "edit"
+    end
+
+    it "responds successfully when the reset token is expired" do
+      auth = build(:email_authentication)
+      token = auth.send_reset_password_instructions
+      auth.update_attribute(:reset_password_sent_at, 7.hours.ago)
```
rsanheim, line 19: Two suggestions here:

* can you use `update!` so that the save will raise if anything goes wrong?
* it isn't clear that `7.hours` would mark the token as expired, and that could become an issue if we change the expiration time later. Maybe something like:

```
expired_sent_time = Devise.config.reset_password_within + 1.hour
auth.update!(reset_password_sent_at: expired_sent_time)
```

rsanheim:
> A flaw in devise, as far as I can tell.

It sounds like we are possibly doing something to break the flow Devise has to redirect a user to the new password page for an expired token: https://github.com/heartcombo/devise/pull/4837

That was merged well over a year ago, and we definitely have a version that includes that fix. 

This fix still makes sense to me to improve the user experience, just wanted to point out that Devise seems to have behavior built in to handle this case.

spec/controllers/email_authentications/passwords_controller_spec.rb
kpethtel: Yeah, I should've thought to use the Devise config. Good point.

app/views/email_authentications/passwords/expired_reset_token.html.erb
kpethtel: I'll double check with Claudio but I believe that was the request.

kpethtel:
> > A flaw in devise, as far as I can tell.
> 
> It sounds our app may be doing something to break the flow Devise has to redirect a user to the new password page for an expired token: [heartcombo/devise#4837](https://github.com/heartcombo/devise/pull/4837)
> 
> That was merged well over a year ago, and we definitely have a version that includes that fix.
> 
> This fix still makes sense to me to improve the user experience, just wanted to point out that Devise seems to have behavior built in around this.

Interesting find. That devise PR doesn't do quite the same thing as I've done here. It changes behavior when you follow a reset link and enter a new password from "you get an error message" to "you get redirected with an error message". My work here tells you as soon as you follow the reset link that your token is expired. You don't have to enter a password and then find out your token was expired.

But I probably wouldn't have made this change if I had been getting any feedback in the app that the password was expired. But I wasn't getting any feedback from the app. I would try to reset the password and nothing would happen. That's what prompted me to file a ticket. 

As part of this work, I downloaded the devise source code and looked at the passwords controller. This conversation prompted me to look at it again, and the bug fix you found is not in the devise currently. So maybe it was reverted for some reason.

claudiovallejo (approved):
Tested locally and it looks great, thanks Kris! 🚀

PR #2121, opened 2021-03-04
2832: fix inconsistent specs

kpethtel:
**Story card:** [ch2832](https://app.clubhouse.io/simpledotorg/story/2832/fix-inconsistent-spec)

## Because

Using the import method here skips model validations, which causes the facility group to not enforce a unique slug and results in errors on save. 

## This addresses

Spec errors seen on CI. 

## Testing

I have run the CI tests several times and the only errors I encountered were from the flapping invitations controller specs.


harimohanraj89 (approved):
Ooh nifty. The `import` was bypassing the polite unique slug generation, and now we stop bypassing it 👍

PR #2125, opened 2021-03-08
2795 incorrect registrations count on overview

kpethtel:
**Story card:** [ch2795](https://app.clubhouse.io/simpledotorg/story/2795/registration-counts-are-inconsistent-across-details-and-overview-tabs)

## Because

The registrations count in the reports overview can be wrong. 

## This addresses

Currently, when there are no registrations for the current month, the last month's registrations data can appear in this month's place. This is because of the way that registrations and assigned patients are queried and grouped by month and then processed through the Region::Result object. 

## Test instructions

The bug only appears when there are no registrations in the current month, so it may be difficult to repro if you have re-seeded your db recently. 
- Turn on the  report_with_exclusions flipper flag.
- Go to the reports details page and look in "Patient registrations and follow-ups" monthly registered patients for the current month. If that's a dash, we would expect this bug to occur. 
- Go to the reports overview page and look at "Mar-2021 district performance". The "Patients registered in Mar-2021" column should be 0 if you saw a dash in the details page.


app/views/reports/regions/_child_data_tables_with_exclusions.html.erb
```diff
@@ -57,64 +57,61 @@
...
-              <%= number_with_delimiter(result["missed_visits"].values.last) %>
+              <%= number_with_delimiter(result.dig(:missed_visits, @period)) %>
             </td>
             <td class="ta-left">
-              <%= number_with_delimiter(result["cumulative_registrations"].values.last) %>
+              <%= number_with_delimiter(result.dig(:cumulative_registrations, @period)) %>
             </td>
             <td class="ta-left">
-              <%= number_with_delimiter(result["registrations"].values.last || 0) %>
+              <%= number_with_delimiter(result.dig(:registrations, @period)) %>
```
kpethtel, line 114: This line is the actual bug, but the problem fundamentally is that we are relying on positional data when the positions aren't actually pinned to the current month, which is why I've changed all of the positional data fetching in this file. I think a better fix would be to pin the queries to the requested period more firmly because this bug is likely to pop up elsewhere otherwise. But that is outside the scope of the current bug fix and would possibly cause bugs elsewhere if done without sufficient care.

app/views/reports/regions/_child_data_tables_with_exclusions.html.erb
```diff
@@ -57,64 +57,61 @@
...
-            <%= number_with_delimiter(@data.last_value(:cumulative_registrations)) %>
+            <%= number_with_delimiter(@data.dig(:cumulative_registrations, @period)) %>
           </td>
           <td class="ta-left">
-            <%= number_with_delimiter(@data.last_value(:registrations)) %>
+            <%= number_with_delimiter(@data.dig(:registrations, @period)) %>
           </td>
         </tr>
         <% data.each do |result| %>
-          <% next if result.dig("controlled_patients_rate", @period).nil? %>
```
kpethtel, line 85: I removed these (and the `|| 0` code below) because all of these except missed_visits_rate are all hashes that are instantiated to return nil instead of zero. Frankly, I'm not even sure why we want to skip to the next result if this is nil. That seems wrong to me. Does anyone know why that is?

app/views/reports/regions/_child_data_tables_with_exclusions.html.erb
```diff
@@ -57,64 +57,61 @@
...
           <td class="ta-left">
-            <%= number_with_delimiter(@data.last_value(:missed_visits)) %>
+            <%= number_with_delimiter(@data.dig(:missed_visits, @period)) %>
           </td>
           <td class="ta-left">
-            <%= number_with_delimiter(@data.last_value(:cumulative_registrations)) %>
+            <%= number_with_delimiter(@data.dig(:cumulative_registrations, @period)) %>
           </td>
           <td class="ta-left">
-            <%= number_with_delimiter(@data.last_value(:registrations)) %>
```
kpethtel, line 81: This line is the actual bug, but the problem fundamentally is that we are relying on positional data when the positions aren't actually pinned to the current month, which is why I've changed all of the positional data fetching in this file. I think a better fix would be to pin the queries to the requested period more firmly because this bug is likely to pop up elsewhere otherwise. But that is outside the scope of the current bug fix and would possibly cause bugs elsewhere if done without sufficient care.

app/views/reports/regions/_child_data_tables_with_exclusions.html.erb
rsanheim: There are cases where the data for a facility can be completely empty for new facilities. With the positional approach, we would hit errors because values would be completely empty, so values.last would blow up. 

With the direct dig approach, I think we will rule out those types of errors altogether. 👍🏼

rsanheim (approved):
:shipit:

PR #2128, opened 2021-03-08
Region performance card improvements

claudiovallejo:
**Story card:** [ch2839](https://app.clubhouse.io/simpledotorg/story/2839/region-performance-improvements)

## Because

The region performance card could use some improvements to make copy / data easier to read

## This addresses

* Adds a new header row
* Adds tooltips to header row titles
* Updates sub-header row titles
* Uses tooltips to display percent values

image

## Test instructions

**Note:** Turn the `report_with_exclusions` feature flag

* [ ] The last card in the "Overview" section has a white header row with "BP controlled", "BP not controlled", "Missed visits", and "Registrations" sections
* [ ] The last card in the "Overview" section has a sub-header row with "Facility", "Percent", "Total", "Percent", "Total", "Percent", "Total", "Total", "Mar-2021" column titles
* [ ] All "Percent" cells display a pill. When the user hovers over the pill the pill displays the numerator and denominator

### DIFF
--- a/app/views/reports/regions/_child_data_tables_with_exclusions.html.erb
+++ b/app/views/reports/regions/_child_data_tables_with_exclusions.html.erb
@@ -16,6 +16,21 @@
         <col>
       </colgroup>
       <thead>
+        <tr>
+          <th></th>
+          <th colspan="2">
+            BP controlled
+          </th>
+          <th colspan="2">
+            BP not controlled
+          </th>
+          <th colspan="2">
+            Missed visits
+          </th>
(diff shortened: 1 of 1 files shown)


claudiovallejo:
> Looks good! The only thing that looks off is that the little arrow on the hover tip points to the middle of the cell instead of the middle of the pill, but that's a very minor quibble.

Thanks for pointing that out @kpethtel I notice that there's another tooltip alignment bug in the headers. I'll open a separate PR to fix those.

PR #2129, opened 2021-03-09
Don't auto deploy to performance env

prabhanshuguptagit:
**Story card:** -

## Because
We shouldn't auto deploy to perf envs since we use it for testing branches on. Auto deploys can put fresh code on the boxes in between a test (which happened yesterday while we were testing patient merge).

## This addresses

This removes perf from semaphore config.

### DIFF
--- a/.semaphore/deploy.yml
+++ b/.semaphore/deploy.yml
@@ -41,26 +41,6 @@ blocks:
             - bundle install --deployment --path vendor/bundle
             - cache store
             - BRANCH=$SEMAPHORE_GIT_BRANCH bundle exec cap qa deploy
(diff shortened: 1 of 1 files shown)


rsanheim (approved):

harimohanraj89:
Can we disable the auto-deployment, instead of the deployment? That way we can still button-click-deploy to performance, but we retain control over it.

prabhanshuguptagit:
>Can we disable the auto-deployment, instead of the deployment? That way we can still button-click-deploy to performance, but we retain control over it.

We could, although we usually need to deploy a branch so a semaphore button wouldn't really help.

PS How would I go about doing this?

PR #2131, opened 2021-03-10
Add Drug model and API

vkrmis:
**Story card:** [chXXXX](URL)

## Because

## This addresses

## Test instructions

### DIFF
--- a/app/controllers/api/v4/drugs_controller.rb
+++ b/app/controllers/api/v4/drugs_controller.rb
@@ -1,0 +1,50 @@
+class Api::V4::DrugsController < Api::V4::SyncController
+  skip_before_action :current_user_present?, only: [:sync_to_user]
+  skip_before_action :validate_sync_approval_status_allowed, only: [:sync_to_user]
+  skip_before_action :authenticate, only: [:sync_to_user]
+  skip_before_action :validate_facility, only: [:sync_to_user]
+  skip_before_action :validate_current_facility_belongs_to_users_facility_group, only: [:sync_to_user]
+
+  def sync_to_user
+    __sync_to_user__("drugs")
+  end
+
+  private
+
+  def current_facility_records
+    []
+  end
+
+  def other_facility_records
+    time(__method__) do
+      Drug.all
+    end
+  end
+
+  def disable_audit_logs?
+    true
+  end
+
+  def transform_to_response(drug)
+    Api::V4::DrugTransformer.to_response(drug)
+  end
+
+  def response_process_token
+    {other_facilities_processed_since: processed_until(other_facility_records),
+     resync_token: resync_token}
+  end
+
+  def processed_until(records)
+    Time.current.strftime(APIController::TIME_WITHOUT_TIMEZONE_FORMAT) if records.any?
+  end
+
+  def force_resync?
+    resync_token_modified?
+  end
+
+  def records_to_sync
+    time(__method__) do
+      other_facility_records
+    end
+  end
+end
--- a/app/models/drug.rb
+++ b/app/models/drug.rb
@@ -1,0 +1,21 @@
+class Drug < ActiveYaml::Base
+  set_root_path "config/data"
+  set_filename "drugs"
+
+  CREATED_TIME ||= File.ctime("config/data/drugs.yml").in_time_zone("UTC")
+  UPDATED_TIME ||= File.mtime("config/data/drugs.yml").in_time_zone("UTC")
+
+  def updated_at
+    UPDATED_TIME
+  end
+
+  def created_at
+    CREATED_TIME
+  end
+
+  def deleted_at
+    if deleted == true
+      UPDATED_TIME
+    end
+  end
+end
--- a/app/models/drug.rb
+++ b/app/models/drug.rb
@@ -5,6 +5,22 @@ class Drug < ActiveYaml::Base
   CREATED_TIME ||= File.ctime("config/data/drugs.yml").in_time_zone("UTC")
   UPDATED_TIME ||= File.mtime("config/data/drugs.yml").in_time_zone("UTC")
 
+  CATEGORIES = {
+    hypertension_ccb: "Hypertension: CCB",
+    hypertension_arb: "Hypertension: ARB",
+    hypertension_diuretic: "Hypertension: Diuretic",
+    hypertension_ace: "Hypertension: ACE Inhibitor",
+    hypertension_other: "Hypertension: Other",
+    diabetes: "Diabetes",
+    other: "Other"
+  }.freeze
+
+  FREQUENCIES = {
+    once_a_day: "Once a day",
+    twice_a_day: "Twice a day",
+    thrice_a_day: "Thrice a day"
+  }.freeze
+
   def updated_at
     UPDATED_TIME
   end
--- a/app/schema/api/v4/models.rb
+++ b/app/schema/api/v4/models.rb
@@ -302,6 +302,25 @@ class Api::V4::Models
        required: %w[id password]}
     end
 
+    def drug
+      {type: :object,
+       properties: {
+         id: {"$ref" => "#/definitions/uuid"},
+         name: {"$ref" => "#/definitions/non_empty_string"},
+         category: {type: [:string, "null"], enum: Drug::CATEGORIES.keys << nil},
+         frequency: {type: [:string, "null"], enum: Drug::FREQUENCIES.keys << nil},
+         composition: {type: [:string, "null"]},
+         dosage: {type: [:string, "null"]},
+         rxnorm_code: {type: [:string, "null"]},
+         protocol: {type: :boolean},
+         common: {type: :boolean},
+         created_at: {"$ref" => "#/definitions/timestamp"},
+         updated_at: {"$ref" => "#/definitions/timestamp"},
+         deleted_at: {"$ref" => "#/definitions/nullable_timestamp"}
+       },
+       required: %w[id name created_at updated_at]}
+    end
+
     def definitions
       {timestamp: timestamp,
        uuid: uuid,
@@ -320,7 +339,9 @@ class Api::V4::Models
        facility_medical_officer: facility_medical_officer,
        facility_medical_officers: array_of("facility_medical_officer"),
        teleconsultation: teleconsultation,
-       teleconsultations: array_of("teleconsultation")}
+       teleconsultations: array_of("teleconsultation"),
+       drug: drug,
+       drugs: array_of("drug")}
     end
   end
 end
--- a/app/transformers/api/v4/drug_transformer.rb
+++ b/app/transformers/api/v4/drug_transformer.rb
@@ -1,0 +1,15 @@
+class Api::V4::DrugTransformer < Api::V4::Transformer
+  class << self
+    def to_response(drug)
+      drug
+        .as_json["attributes"]
+        .merge("protocol" => false,
+               "common" => true,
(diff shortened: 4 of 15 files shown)


app/controllers/api/v4/drugs_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  private
+
+  def current_facility_records
+    []
+  end
+
+  def other_facility_records
+    time(__method__) do
+      Drug.all
```
vkrmis, line 20: Need to handle limit and updated_at filtering here

app/models/drug.rb
```diff
@@ -0,0 +1,37 @@
+class Drug < ActiveYaml::Base
+  set_root_path "config/data"
+  set_filename "drugs"
+
+  CREATED_TIME ||= File.ctime("config/data/drugs.yml").in_time_zone("UTC")
+  UPDATED_TIME ||= File.mtime("config/data/drugs.yml").in_time_zone("UTC")
```
vkrmis, line 6: Sets the drugs' updated at to whenever the file was last updated. This will resync all drugs whenever the file is modified, but that should be okay since its not a large list.

app/models/drug.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  CATEGORIES = {
+    hypertension_ccb: "Hypertension: CCB",
+    hypertension_arb: "Hypertension: ARB",
+    hypertension_diuretic: "Hypertension: Diuretic",
+    hypertension_ace: "Hypertension: ACE Inhibitor",
+    hypertension_other: "Hypertension: Other",
+    diabetes: "Diabetes",
+    other: "Other"
+  }.freeze
```
vkrmis, line 16: An analogous enum exists in the ProtocolDrug class, but this class should probably contain the authoritative list.
Can make it an enum when this becomes an ActiveRecord backed class.

app/models/drug.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  FREQUENCIES = {
+    once_a_day: "Once a day",
+    twice_a_day: "Twice a day",
+    thrice_a_day: "Thrice a day"
+  }.freeze
```
vkrmis, line 22: Similar to the `CATEGORIES` map above, a version of this is in the PrescriptionDrug class, should get folded into this class instead. Also using these representations instead of "OD", "BD" etc, since those strings are not applicable everywhere (eg: Ethiopia), and need to be i18n-ized. Picked this clearer representation since we'll i18n-ize anyway.

app/models/drug.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def created_at
+    CREATED_TIME
+  end
+
+  def deleted_at
+    if deleted == true
+      UPDATED_TIME
+    end
+  end
```
vkrmis, line 36: These methods will allow us to look up `drug.updated_at` etc and get a useful value back.

app/schema/api/v4/models.rb
```diff
@@ -302,6 +302,25 @@ def activate_user
...
+       properties: {
+         id: {"$ref" => "#/definitions/uuid"},
+         name: {"$ref" => "#/definitions/non_empty_string"},
+         category: {type: [:string, "null"], enum: Drug::CATEGORIES.keys << nil},
+         frequency: {type: [:string, "null"], enum: Drug::FREQUENCIES.keys << nil},
+         composition: {type: [:string, "null"]},
+         dosage: {type: [:string, "null"]},
+         rxnorm_code: {type: [:string, "null"]},
+         protocol: {type: :boolean},
+         common: {type: :boolean},
```
vkrmis, line 316: Left out `stock_tracked` since that's only relevant server-side

spec/api/v4/drugs_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "swagger_helper"
+
+describe "Drugs v4 API", swagger_doc: "v4/swagger.json" do
+  path "/drugs/sync" do
+    get "Syncs drug data from server to device." do
+      tags "Drug"
+      parameter name: "HTTP_X_FACILITY_ID", in: :header, type: :uuid
```
vkrmis, line 7: Making the facility_id a part of this spec, since it'll become relevant when we switchover to using this API for all drug info. 
Haven't made it mandatory though, so requests without facility_id will still work for now. Maybe it should be mandatory anyway?

app/controllers/api/v4/drugs_controller.rb
```diff
@@ -0,0 +1,50 @@
+class Api::V4::DrugsController < Api::V4::SyncController
+  skip_before_action :current_user_present?, only: [:sync_to_user]
+  skip_before_action :validate_sync_approval_status_allowed, only: [:sync_to_user]
+  skip_before_action :authenticate, only: [:sync_to_user]
+  skip_before_action :validate_facility, only: [:sync_to_user]
+  skip_before_action :validate_current_facility_belongs_to_users_facility_group, only: [:sync_to_user]
```
ssrihari, line 6: Should we extract all this out into a "public API" mixin or something like that? Don't want facilities, protocols, and this drugs model to drift on this.

app/controllers/api/v4/drugs_controller.rb
ssrihari: If we return all the drugs every time, when does the app stop syncing?

app/models/drug.rb
prabhanshuguptagit: Can this just be a list, since the values in the hash don't represent anything important? I think `PrescriptionDrug` has a hash because it's an `enum` and we also use the values to display.

app/models/drug.rb
```diff
@@ -0,0 +1,38 @@
+class Drug < ActiveYaml::Base
```
prabhanshuguptagit, line 1: Shall we call this model `Medication`, given we have a chance to rename and this is going to be the long term model? This seems to be the FHIR equivalent of what we're modelling https://www.hl7.org/fhir/medication.html.

Aside: `PrescriptionDrug` = https://www.hl7.org/fhir/medicationrequest.html, although this rename might be too much 😅 .

app/transformers/api/v4/drug_transformer.rb
```diff
@@ -0,0 +1,14 @@
+class Api::V4::DrugTransformer < Api::V4::Transformer
+  class << self
+    def to_response(drug)
+      drug
+        .as_json["attributes"]
```
prabhanshuguptagit, line 5: [nit]
```suggestion
        .attributes.as_json
```

config/data/drugs.yml
```diff
@@ -0,0 +1,1025 @@
+---
+- id: e9865df7-39b5-4fb8-8ab3-354f87a5d369
+  name: Amlodipine
+  composition:
```
prabhanshuguptagit, line 4: Double checking, do we want to render these empty fields as `nil` or empty string in the JSON?

spec/api/v4/drugs_spec.rb
prabhanshuguptagit: I think non-mandatory should be fine, we seem to allow it for other resources too.

prabhanshuguptagit (changes requested):
Left some questions around sync but looking solid overall👌🏼

app/models/drug.rb
vkrmis: We could, I put in a hash here so we can turn it into an `enum` when this gets promoted to an AR backed model

app/controllers/api/v4/drugs_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    Api::V4::DrugTransformer.to_response(drug)
+  end
+
+  def response_process_token
+    {other_facilities_processed_since: processed_until(other_facility_records),
+     resync_token: resync_token}
+  end
+
+  def processed_until(records)
+    Time.current.strftime(APIController::TIME_WITHOUT_TIMEZONE_FORMAT) if records.any?
```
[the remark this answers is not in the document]
vkrmis: App stops syncing when payload size < batch size it asks for. 
This should always be true in the near to mid future (~130 vs 1000)

app/models/drug.rb
vkrmis: Hmm, that's a good call, will take a look at FHIR models

config/data/drugs.yml
vkrmis: Its a nullable string, so nil is better

app/controllers/api/v4/drugs_controller.rb
vkrmis: We're relying on an incidental factor here: The app defaults to a batch size of 1000, and there are only ~130 drugs. The app will assume sync has completed when it gets back < 1000 records for a sync request.

Potential failure scenarios:
- The Drug list grows to be > 1000 records, and the app either crashes when it gets back > 1000 records, or gets stuck in a sync loop as it keeps receiving the same records over and over. However, this list is very unlikely to grow so large before we move to an AR backed model, with the usual sync semantics. 
- The app changes its batch size to ~100 records, and runs into the same problem ^. We'll need to coordinate with the client to ensure this doesn't happen in the near future.

The reasons for returning all drugs here instead of using the simulated `updated_at` timestamps are:
- For a small batch size, the above problem still exists, since all the drug records share the same timestamp anyway.
- git (sometimes) rewrites the entire file, so the `created_at` and `updated_at` timestamps aren't always reliable. Given that the `updated_at` semantics don't reliably work, we probably shouldn't pretend that they do, and make this behaviour explicit instead by returning all drugs.

prabhanshuguptagit: Can we manually set a `VERSION` in the model and cursor based on that? We can send it in the process token, seems straightforward than using an `updated_at` at all. 

> We'll need to coordinate with the client to ensure this doesn't happen in the near future.

This is a little worrisome since in the bad case it causes sync loops for everyone. The limit is not set in stone, it's just a firebase config I think. If it's not much effort let's cursor it?

app/models/drug.rb
prabhanshuguptagit: Do we need `.updated_at` for anything other than the process token?

vkrmis: No, but it's worth adding timestamps to the data model now, so we can transition to a AR backed model seamlessly (We'd need a version bump to add in this field later).

app/controllers/api/v4/drugs_controller.rb
ssrihari: This might be worth addressing if it takes about half a day to fix. Else, we can choose to accept the tradeoff for now and switch to active record when the time comes. Wdyt?

prabhanshuguptagit: Sounds good. 👍🏼

prabhanshuguptagit (approved):
Approving to unblock.

app/models/drug.rb
vkrmis: Renamed 🙌

app/controllers/api/v4/drugs_controller.rb
vkrmis: Thought about this a bit more, I've decided to leave it as-is for now.

- Don't want to reinvent the wheel to implement sync-logic for a temporary implementation
- We've not actually changed the sync batch size for a long time, don't foresee it changing drastically (x10 reduction)
- We could fix it by just changing this to an AR backed model if necessary

Will make sure we test this in sandbox and demo envs so we don't break sync at launch

PR #2132, opened 2021-03-10
Count LTFU numbers as of end of period

prabhanshuguptagit:
**Story card:** [ch2860](https://app.clubhouse.io/simpledotorg/story/2860)

## Because

We currently exclude LTFU numbers as of the beginning of the month in our all time reports. This is bad math while calculating controlled/uncontrolled rates for the current month. Since the BPs that we count are until the end of month but LTFU patients are excluded _as of the beginning of the month_, it's possible we undercount a patient who is registered.

Showing end of month numbers everywhere also keeps things consistent with other graphs.

More detailed discussion [in this Slack thread](https://simpledotorg.slack.com/archives/C01GF5R35RB/p1614696084007400)

### LTFU chart 
This now shows exclusion numbers as of end of month. This means we are counting LTFU since end of month of 12 months ago.
#### Before
image
#### After
image

### Breakdown pie chart
This shows `Dead` patients as of today and the rest of the patients are LTFU as of the end of the month. The `Transferred` count inside each section is also as of today.
#### Before
image
#### After
image

### Region reports

The denominator now excludes LTFU patients as of the end of month instead of beginning of month.

### DIFF
--- a/app/assets/javascripts/common/patient_breakdown_reports.js
+++ b/app/assets/javascripts/common/patient_breakdown_reports.js
@@ -109,7 +109,7 @@ PatientBreakdownReports = function () {
         totalPatientsNode.innerHTML = reports.formatNumberWithCommas(totalPatients);
         periodStartNode.innerHTML = period.ltfu_since_date;
         registrationsNode.innerHTML = reports.formatNumberWithCommas(cumulativeAssignedPatients);
-        registrationsPeriodEndNode.innerHTML = period.start_date;
+        registrationsPeriodEndNode.innerHTML = period.bp_control_end_date;
       }
     };
 
--- a/app/models/concerns/patient_reportable.rb
+++ b/app/models/concerns/patient_reportable.rb
@@ -12,15 +12,15 @@ module PatientReportable
     scope :ltfu_as_of, ->(date) do
       joins("LEFT OUTER JOIN latest_blood_pressures_per_patient_per_months
              ON patients.id = latest_blood_pressures_per_patient_per_months.patient_id
-             AND #{sanitize_sql(["bp_recorded_at > ? AND bp_recorded_at < ?", date - LTFU_TIME, date])}")
-        .where("bp_recorded_at IS NULL AND patients.recorded_at < ?", date - LTFU_TIME)
+             AND #{sanitize_sql(["bp_recorded_at > ? AND bp_recorded_at < ?", (date - LTFU_TIME).end_of_month, date])}")
+        .where("bp_recorded_at IS NULL AND patients.recorded_at < ?", (date - LTFU_TIME).end_of_month)
         .distinct(:patient_id)
     end
 
     scope :not_ltfu_as_of, ->(date) do
       joins("LEFT OUTER JOIN latest_blood_pressures_per_patient_per_months
              ON patients.id = latest_blood_pressures_per_patient_per_months.patient_id")
-        .where("bp_recorded_at > ? AND bp_recorded_at < ? OR patients.recorded_at >= ?", date - LTFU_TIME, date, date - LTFU_TIME)
+        .where("bp_recorded_at > ? AND bp_recorded_at < ? OR patients.recorded_at >= ?", (date - LTFU_TIME).end_of_month, date, (date - LTFU_TIME).end_of_month)
         .distinct(:patient_id)
     end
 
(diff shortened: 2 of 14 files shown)


app/models/concerns/patient_reportable.rb
```diff
@@ -12,15 +12,15 @@ module PatientReportable
     scope :ltfu_as_of, ->(date) do
       joins("LEFT OUTER JOIN latest_blood_pressures_per_patient_per_months
              ON patients.id = latest_blood_pressures_per_patient_per_months.patient_id
-             AND #{sanitize_sql(["bp_recorded_at > ? AND bp_recorded_at < ?", date - LTFU_TIME, date])}")
-        .where("bp_recorded_at IS NULL AND patients.recorded_at < ?", date - LTFU_TIME)
+             AND #{sanitize_sql(["bp_recorded_at > ? AND bp_recorded_at < ?", (date - LTFU_TIME).end_of_month, date])}")
+        .where("bp_recorded_at IS NULL AND patients.recorded_at < ?", (date - LTFU_TIME).end_of_month)
```
prabhanshuguptagit, line 16: This is slightly unrelated: This fixes the problem we saw in February where `12.months.ago` is not the right date to display.

app/models/reports/result.rb
```diff
@@ -179,8 +179,7 @@ def calculate_period_info(range)
       self.period_info = range.each_with_object({}) do |period, hsh|
         hsh[period] = {
           name: period.to_s,
-          start_date: period.start_date.to_s(:day_mon_year),
-          ltfu_since_date: period.start_date.advance(months: -12).to_s(:day_mon_year),
```
prabhanshuguptagit, line 183: We don't need a `start_date` anymore. We only show end of month numbers everywhere now!

claudiovallejo (approved):
Looks good to me sir!

app/assets/javascripts/common/patient_breakdown_reports.js
```diff
@@ -109,7 +109,7 @@ PatientBreakdownReports = function () {
         totalPatientsNode.innerHTML = reports.formatNumberWithCommas(totalPatients);
         periodStartNode.innerHTML = period.ltfu_since_date;
         registrationsNode.innerHTML = reports.formatNumberWithCommas(cumulativeAssignedPatients);
-        registrationsPeriodEndNode.innerHTML = period.start_date;
+        registrationsPeriodEndNode.innerHTML = period.bp_control_end_date;
```
harimohanraj89, line 112: I wanna learn more about the JS implementation - where does this data go and sit? Is it globbed onto data attributes, and it powers the mouseover behavior of the charts?

app/models/concerns/patient_reportable.rb
harimohanraj89: Wow, nice one. This Feb, it would have shown Feb 28 2020, even though the end of that month was Feb 29 (leap year)? Amazing ✨

harimohanraj89:
Hey @prabhanshuguptagit I found a bug, not sure if it's related to our changes here or not.

When a past period is selected in the overview, and then you go to the details page, the past period is retained in the URL param. This is causing some wonky behavior in the LTFU chart _after_ you mouseover an old datapoint.

Check it out.

img

harimohanraj89:
I suspect it's unrelated - we're accidentally respecting the `period` param in the LTFU chart, even though we shouldn't yet. In any case, I'll let you decide whether we wanna address it now or punt for later.

app/assets/javascripts/common/patient_breakdown_reports.js
prabhanshuguptagit: [This ChartJS config](https://github.com/simpledotorg/simple-server/blob/2baa53212c1a027f352856793cb4d7f373506c42/app/assets/javascripts/common/patient_breakdown_reports.js#L87) allows us to pass it a function that gets called on the mouseover. The `innerHTML` in this function changes the numbers at the top of the card.
Screenshot 2021-03-11 at 1 36 53 PM

app/models/concerns/patient_reportable.rb
prabhanshuguptagit: Yep 😅

harimohanraj89:
Neato 👌 The copy bug is already on your radar, so feel free to roll in the fix here or later, and :shipit:

harimohanraj89 (approved):

PR #2133, opened 2021-03-10
Exclude BPs from deleted patients from matview

harimohanraj89:
**Story card:** [ch2855](https://app.clubhouse.io/simpledotorg/story/2855/negative-reports-numbers)

## Because

If a patient is discarded but their BPs are not, their BPs are counted in reports numerators, but they are not counted in denominators. This is bad math, and leads to wonky results like negative numbers.

## This addresses

Updating the latest BPs per patient per month matview to exclude undiscarded BPs that belong to discarded patients.

The matview already excludes discarded BPs, and we rely on this behavior to circumvent some nasty ActiveRecord business. I think our options to protect against this were:
* ~~**Better data validations on BPs and patients:**~~ Since BP and patient data are synced asynchronously to us, I chose not to enforce any data validation rules on BPs and patients themselves
* ~~**Update the control and uncontrol queries to exclude discarded patient data**~~: I also chose not to do this because 1) the default scopes have trouble operating on our views, and 2) our assumption that only undiscarded data lives in the matview may be more widespread
* **Update the matview to ignore BPs that belong to discarded patients**: I think this is earliest layer where we can keep our data clean, so I've gone ahead with this.

## Test instructions

* Find a patient with a recent (within the last 3 months, the more recent the better) BP
* Make a note of the patient's assigned facility
* View the report for the patient's assigned facility. Make a note of controlled, uncontrolled, visit-but-no-bp, missed visit, and registration numbers for the month
* Discard the patient, but not their BPs
* Refresh materialized views, force-refresh the reports caches
* View the report for the patient's assigned facility again. The registrations number should be 1 less. The controlled/uncontrolled BP number should be 1 less, depending on whether the patient's recent BP was controlled or uncontrolled.

### DIFF
--- a/db/migrate/20210310102235_update_latest_blood_pressures_per_patient_per_months_to_version_5.rb
+++ b/db/migrate/20210310102235_update_latest_blood_pressures_per_patient_per_months_to_version_5.rb
@@ -1,0 +1,11 @@
+class UpdateLatestBloodPressuresPerPatientPerMonthsToVersion5 < ActiveRecord::Migration[5.2]
+  def change
+    drop_view :latest_blood_pressures_per_patient_per_quarters, materialized: true, revert_to_version: 3
+    drop_view :latest_blood_pressures_per_patients, materialized: true, revert_to_version: 1
+
+    update_view :latest_blood_pressures_per_patient_per_months, materialized: true, version: 5, revert_to_version: 4
+
+    create_view :latest_blood_pressures_per_patient_per_quarters, materialized: true, version: 3
+    create_view :latest_blood_pressures_per_patients, materialized: true, version: 1
+  end
+end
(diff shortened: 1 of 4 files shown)


prabhanshuguptagit:
This looks solid👌🏼. I was hoping we would find a way in option 1 or 2 but this seems the most robust.

PS Are we going with "the reviewer has to try out the testing instructions"? I would like to not do that, its a fair bit of overhead to test as a responsibility while reviewing. Will x-post this on the slack thread.

harimohanraj89:
Yeah, I'm not sure either. Let's discuss at show-and-tell today 👍 https://simpledotorg.slack.com/archives/CFHMC60P5/p1614334997093700

I will go through the testing instructions myself for this PR.

harimohanraj89:
Tested 👍

prabhanshuguptagit (approved):

PR #2134, opened 2021-03-10
Remove focus from test group

harimohanraj89:
**Story card:** N/A

Remove a stray `focus` from a test group. This was used for dev purposes, should not have been merged.


vkrmis (approved):

rsanheim:
Oof, thanks for catching this. I'll look into adding a fail safe to loudly error if these make it to CI.

PR #2137, opened 2021-03-10
Exclude non web requests from Rack mini profiler

rsanheim:
[rack-mini-profiler](https://github.com/MiniProfiler/rack-mini-profiler) will sometimes gum up the works when you are
developing against simple-server from the android app. We shouldn't have
it sending its own requests for API or webview requests.

no story, this just makes our dev life easier


kpethtel (approved):
👍

PR #2140, opened 2021-03-10
2838 assigned patients detail page

kpethtel:
**Story card:** [ch2838](https://app.clubhouse.io/simpledotorg/story/2838/add-total-assigned-patients-to-reports-details-page)

## Because

We're adding assigned patients to the regions detail view.

## This addresses

Users' desire to see assigned patients data. 

## Test instructions

Both with and without the report_with_exclusions flag on:
- Look at reports details page for a region and confirm it matches mocks. 
- This does not apply to individual facilities
- also confirm that tables in the app look correct in general. The one line of css changed here can impact other tables. 

## Deployment note
This will require busting the cache for the `district_analytics_query.rb `.


claudiovallejo (approved):
Looks good to me 👌  thanks for fixing the breaking period header titles!

spec/queries/district_analytics_query_spec.rb
```diff
@@ -227,4 +262,15 @@
       end
     end
   end
+
+  describe "#total_assigned_patients" do
+    it "returns total assigned patients sorted by facility id" do
+      patient = create(:patient, :hypertension, assigned_facility: facility_1)
```
harimohanraj89, line 268: You can drop the variable assignment to squash the warning here.

harimohanraj89, line 268: You can drop the variable assignment to squash the warning here.

app/helpers/dashboard_helper.rb
```diff
@@ -43,6 +43,12 @@ def format_period(period, value)
     period == :month ? value.strftime("%b-%Y") : quarter_string(value)
   end
 
+  def multiline_format_period(period, value)
+    formatted_period = format_period(period, value)
+    shortened_period, year = formatted_period.split("-")
+    shortened_period + "-\n" + year
```
harimohanraj89, line 49: The newline here doesn't introduce a newline in the markup, right? The newline would be interpreted as a space in the markup, and the word-wrap behavior would kick in? Shall we just introduce a space here?

Consider something like `format_period(period, value).gsub("-", "- ")`

app/queries/district_analytics_query.rb
```diff
@@ -34,6 +35,22 @@ def results
     results.inject(&:deep_merge)
   end
 
+  def total_assigned_patients
+    @total_assigned_patients ||=
+      Patient
+        .with_hypertension
+        .joins(:assigned_facility)
```
harimohanraj89, line 42: We don't need to join the full facilities table here. We should be able to do:
`.where(assigned_facility: facilities).group(:assigned_facility_id)`

app/views/reports/regions/details.html.erb
```diff
@@ -307,33 +308,41 @@
...
               </th>
             <% end %>
           </tr>
           <tr class="sorts" data-sort-method="thead">
             <th class="row-label sort-label sort-label-small ta-center" data-sort-default>
               Facilities
             </th>
+            <th class="row-label sort-label sort-label-small ta-center" data-sort-method="number">
+              <% if current_admin.feature_enabled?(:report_with_exclusions) %>
+                Total assigned<br>
```
harimohanraj89, line 324: Is the `br` necessary? It seems awkward to include it at this particular location. I'd rather have the word-wrapping do its thing if possible.

harimohanraj89 (changes requested):

app/queries/district_analytics_query.rb
kpethtel: TBH, I was just imitating the total_registered_patients method below. Your suggestion does seem to yield the same results. Thank you!

app/views/reports/regions/details.html.erb
kpethtel: I was just imitating line 332 below and trying to match the mocks. I'll agree that it isn't necessary and probably just makes our styling and css more finicky. I'll change it and line 332.

app/helpers/dashboard_helper.rb
kpethtel: This is a good suggestion, thank you. I knew something felt off about using split but I didn't think of gsub. You're probably also right about the newline. I'm not sure how the view would have interpreted that. But I think you're correct that it's unnecessary in this case. This table is already cramped for space so I don't think it would ever appear on a single line.

kpethtel: After talking to Claudio, we decided to use the carriage return after all because if we just add a space and the table has space for the full date, it would appear as "MON- YEAR" which seems worse than:
MON-
YEAR

app/queries/district_analytics_query.rb
harimohanraj89: Ha! Room for optimization everywhere 😂

PR #2141, opened 2021-03-10
Extract DrugStocksCreator service object; fix bug when params are empty

rsanheim:
Fixes a bug for drug stocks creation and refactors to a service object.

**Story card:** [ch2888](https://app.clubhouse.io/simpledotorg/story/2888/error-creating-drug-stocks-nomethoderror-undefined-method-map-for-nil-nilclass)


rsanheim:
Going to merge this as it fixes a bug the android team was running into when working on this.

ssrihari:
Would be good to move the tests for the transaction semantics from the usages to a spec file for this service. Also, why do we prefer a service object here instead of a class method on DrugStock?

PR #2145, opened 2021-03-12
Schedule an exact merge of duplicate patients every night

ssrihari:
**Story card:** [ch2685](https://app.clubhouse.io/simpledotorg/story/2685/schedule-the-merge-job-to-run-nightly-and-load-test-it)

Note: this PR is based off an [earlier PR](https://github.com/simpledotorg/simple-server/pull/2119) with some refactoring changes. Please merge that one before merging this one.


harimohanraj89:
Yo @ssrihari if you want to get a head start on reviewing this, wanna rebase this against #2119?

ssrihari:
> Yo @ssrihari if you want to get a head start on reviewing this, wanna rebase this against #2119?

I think Github messes up the change list if the target branch changes. It's okay if this gets delayed a bit. However, it's only a single commit: https://github.com/simpledotorg/simple-server/pull/2145/commits/184646258d0a026b8733b2f7ccd3438f2765efef, if you want to review anyway.

harimohanraj89:
No rush, we'll wait 😅

app/services/patient_deduplication/deduplicator.rb
```diff
@@ -0,0 +1,212 @@
...
+    end
+
+    def mark_as_merged(new_patient)
+      @patients.map do |patient|
+        patient.update!(
+          merged_into_patient_id: new_patient.id,
+          merged_by_user_id: @user_id
+        )
+      end
+      @patients.map(&:discard_data)
```
rsanheim, line 44: Our discard_data calls only `discard` methods, not `discard`.  That means we could do all the merges, and then attempt to discard, which will fail silently due to validation issues. We could easily end up in a situation where we have the new merged records, and some of the old invalid non-discarded records still hanging out.

I think we need to do a `discard!` for all the old dupe records to make sure we don't end up in a really bad state here.  It looks like [discard](https://github.com/jhawthorn/discard) supports the usual AR bang semantics for this.

app/controllers/admin/deduplicate_patients_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    duplicate_patient_ids = PatientDeduplication::Strategies.identifier_match(limit: DUPLICATE_LIMIT)
+    @duplicate_count = duplicate_patient_ids.count
+    @patients = Patient.where(id: duplicate_patient_ids.sample)
+  end
+
+  def merge
+    authorize { current_admin.power_user? }
+
+    duplicate_patients = Patient.where(id: params[:duplicate_patients])
+    PatientDeduplication::Deduplicator.new(duplicate_patients, user: current_admin).merge
```
rsanheim, line 17: This is going to do a bunch of work in the scope of a request, especially if the dupe patients have many BP's or encounters. We probably want to move this to a background job.

rsanheim (dismissed):

app/controllers/admin/deduplicate_patients_controller.rb
prabhanshuguptagit: We think this is an ok trade-off to keep the process simple. We've seen most merges happen within a few seconds (even large ones) and the we would also like to give the user feedback about the merge actually succeeding/failing.

app/services/patient_deduplication/deduplicator.rb
prabhanshuguptagit: The `update!` above should take care of any failing `patient` validations. For associated models, the only change we make is to update the `deleted_at`. This should be okay? (unless the data was already bad).

I think we should separately make `discard_data` use bang methods if we want to fail since it's ubiquitously used. wdyt?

app/controllers/admin/deduplicate_patients_controller.rb
harimohanraj89: Given it's only 1 patient group at a time, I guess that's ok. The complexity only scales with how _long_ the two dupes have been around, and how many BPs they have, etc.

app/services/patient_deduplication/deduplicator.rb
ssrihari: Created a [card on clubhouse](https://app.clubhouse.io/simpledotorg/story/2958/handle-invalid-records-in-patient-discard-data) to handle this later.

PR #2146, opened 2021-03-13
Rename force_cache to bust_cache

rsanheim:
[ch2928](https://app.clubhouse.io/simpledotorg/story/2928/rename-force-cache-to-bust-cache)

The name force_cache has led to some confusion. What it really means is
"bypass the cache, grab fresh values, and fill the cache with that" - so
I think bust_cache is a more clear name for that.

Also extract a module for this behavior, and make sure we setup the bust_cache RequestStore var for all admin controller actions.


rsanheim:
This is pretty much a straight rename, and I verified the `bust_cache` query string still works in development.

PR #2147, opened 2021-03-15
Log deduplication metrics

prabhanshuguptagit:
**Story card:** 

## This addresses

We want to push some metrics about Patient Deduplication to statsd that we can look at in a datadog dashboard.

### DIFF
--- a/app/controllers/admin/deduplicate_patients_controller.rb
+++ b/app/controllers/admin/deduplicate_patients_controller.rb
@@ -15,12 +15,12 @@ class Admin::DeduplicatePatientsController < AdminController
 
     duplicate_patients = Patient.where(id: params[:duplicate_patients])
     deduplicator = PatientDeduplication::Deduplicator.new(duplicate_patients, user: current_admin)
+    merged_patient = deduplicator.merge
 
-    if deduplicator.errors
-      redirect_to admin_deduplication_path, notice: deduplicator.errors
+    if deduplicator.errors.present?
       PatientDeduplication::Stats.report("manual", duplicate_patients.count, 0, duplicate_patients.count)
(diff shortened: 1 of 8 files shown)


app/services/patient_deduplication/runner.rb
```diff
@@ -31,6 +31,10 @@ def handle_error(e, patient_ids)
     def report_summary
       Rails.logger.info(report_stats.to_json)
       Rails.logger.info "Failed to merge patients #{merge_failures}"
+
+      Statsd.instance.count("PatientDeduplication.total_processed", duplicate_patient_ids.flatten.count)
+      Statsd.instance.count("PatientDeduplication.total_merged", duplicate_patient_ids.flatten.count - merge_failures.flatten.count)
+      Statsd.instance.count("PatientDeduplication.total_failures", merge_failures.flatten.count)
```
ssrihari, line 37: Can we add this into the dashboard controller as well? Perhaps we can use a tag to differentiate automatic vs manual while keeping the metric names the same? https://docs.datadoghq.com/developers/metrics/dogstatsd_metrics_submission/#metric-tagging

ssrihari: Addressed in adcac5155eac57195301b3094c36a21b1c0f1c96.

app/controllers/admin/deduplicate_patients_controller.rb
```diff
@@ -18,8 +18,10 @@ def merge
 
     if deduplicator.errors
       redirect_to admin_deduplication_path, notice: deduplicator.errors
+      PatientDeduplication::Stats.report("manual", duplicate_patients.count, 0, duplicate_patients.count)
     else
       merged_patient = deduplicator.merge
```
ssrihari, line 23: Should we be handling tx errors here?

PR #2150, opened 2021-03-16
Show drug stock report on success page (mobile app)

rsanheim:
Shows patient days report after drug stock is successfully submitted on mobile.

**Story card:** [ch2931](https://app.clubhouse.io/simpledotorg/story/2931/show-drug-stock-report-on-success-page-in-android)

### Screen shots

image

image

image


app/controllers/webview/drug_stocks_controller.rb
```diff
@@ -69,9 +77,13 @@ def safe_params
 
   def set_for_end_of_month
     @for_end_of_month ||= if params[:for_end_of_month]
-      Date.parse(params[:for_end_of_month])
+      Date.parse(params[:for_end_of_month]).end_of_month
```
ssrihari, line 80: oof. good catch.

ssrihari: ohey, didn't we have this method extracted to a module somewhere? precisely so we don't run into this kind of fragmentation issues?

.env.development
```diff
@@ -100,7 +100,7 @@ SECRET_KEY_BASE=b90c991f1987f5be558af41e1fdc73741adbef2285f30a5bd500dcd7804b2ef1
 SIMPLE_APP_SIGNATURE=
 SIMPLE_SERVER_ENV=development
 SIMPLE_SERVER_HOST='localhost:3000'
-SIMPLE_SERVER_HOST_PROTOCOL='http'
+SIMPLE_SERVER_HOST_PROTOCOL=https
```
rsanheim, line 103: See https://github.com/simpledotorg/simple-server/pull/2150/commits/48b1a68c880851b1f8b46bcaf5ad3da20cc52db6

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -28,7 +28,7 @@ def new
 
   def create
     DrugStocksCreator.call(current_admin, @facility, @for_end_of_month, drug_stocks_params[:drug_stocks])
-    redirect_to redirect_url(bust_cache: true), notice: "Saved drug stocks"
+    redirect_to redirect_url, notice: "Saved drug stocks"
```
rsanheim, line 31: We shouldn't be using the cache busting query string in our application logic, as its brittle and also will expose it regular users.  I've updated the cache key for drug stocks to get invalidated when new drug stocks are created, which decouples it from any sort of controller logic.

app/queries/drug_stocks_query.rb
```diff
@@ -187,6 +185,7 @@ def drug_stocks_cache_key
     [
       "#{self.class.name}#drug_stocks",
       @facilities.map(&:id).sort,
+      @latest_drug_stocks.cache_key,
```
rsanheim, line 188: @ssrihari @vkrmis This will correctly invalidate the cache when new drug stocks are created, so we don't do any `bust_cache` redirection to handle it. See https://bigbinary.com/blog/activerecord-relation-cache-key for some context on how this works.

app/controllers/webview/drug_stocks_controller.rb
rsanheim: There were enough differences I was finding between the web view side and the admin side that extracting did not make sense (yet).

rsanheim (commented):
Notes for @simpledotorg/server-developers.

app/queries/drug_stocks_query.rb
ssrihari: Thanks for doing this, @rsanheim. We had made a mental note to fix this later on. We recognise that the redirection was more a hack than a fix.

kpethtel (approved):
LGTM.

PR #2154, opened 2021-03-17
Change link for drug stock progress-tab - Part 2

harimohanraj89:
**Story card:** [ch2929](https://app.clubhouse.io/simpledotorg/story/2929/handle-navigation-and-from-drug-stock-success-page-in-the-android-app)

Follow-up to #2153 , adds the deeplink to the back button on the submission form as well.


app/views/webview/drug_stocks/new.html.erb
```diff
@@ -16,14 +16,14 @@
...
+
         <style>
           label.sr-only {display:none;}
           .form-group {padding-left:0;padding-top:4px;padding-bottom:32px;}
         </style>
-      
+
         <div class="progress-contents">
-          <a href="#" onclick="closeWindow('drug-stock-form', 'progress-start'); return false" class="help-back">
+          <a href="simple://progress-tab" title="Go Back" class="help-back">
```
prabhanshuguptagit, line 26: Do we need to do both or just the new one will do?

prabhanshuguptagit (approved):

app/views/webview/drug_stocks/new.html.erb
harimohanraj89: Do you mean both Back buttons? We want to attach the functionality to both. Whether you click "Back" on the drug stuck submission form page, or on the success page, you should be directed back to the top-level progress tab.

prabhanshuguptagit: I was wondering if the same button click triggers both the "clicks" (dumb thought). This makes sense 👍🏼

rsanheim (approved):
Ah yup, this makes sense 👍

PR #2168, opened 2021-03-18
YAMLize help screen

harimohanraj89:
**Story card:** [ch2968](https://app.clubhouse.io/simpledotorg/story/2968/convert-help-locales-to-yml)

## Because

The current HTML locale format of the help screens makes it difficult to:
* Introduce dynamic elements to the help page
* Introduce country-specific elements to the help page
* Manage translations in Transifex when large structural changes are made to the help screen

## This addresses

Extracting all Help text into a YAML-based locale file.

## Test instructions

* Turn on feature flag `yaml_help`
* Ensure the Help screen looks good and is formatted correctly in English only.

## After this PR

After this PR is merged, next steps will be:
* Making sure the new locale keys are imported into Transifex automatically
* Making sure most of the new keys are auto-translated (we have turned on the "Memory Lookup" feature in Transifex)
* Ask GPI to fill in the missing translations
* Accept the PRs that add the new locale keys to all other languages
* Turn on the feature flag
* Once confident, remove the feature flag and old HTML locales completely.


app/views/api/v3/help/show.html.erb
```diff
@@ -10,8 +10,495 @@
   <%= inline_stylesheet 'help.css' %>
   <%= inline_js "standalone/help.js" %>
 </head>
-
-<%= t('api.help.body_html',
+<% if Flipper.enabled?(:yml_help) %>
```
harimohanraj89, line 13: If feature flag is enabled, do the new fancy stuff. Else, do the usual HTML locale thing (at the very bottom)

kpethtel (approved):
This is nice. Tested it locally. Initially didn't think there was any visual difference until I started following the links. My one piece of feedback is that it would be nice to have comments breaking up the sections of the html file like we do here:
https://github.com/simpledotorg/simple-server/blob/master/app/views/api/v2/analytics/user_analytics/show.html.erb
These large html files are very hard to read at a glance, so I personally find it helpful to break them up into clearly defined sections that reflect the user experience.

harimohanraj89:
@kpethtel Great suggestion. I've added some helpful comments to demarcate the sections, and it's a game changer.

PR #2169, opened 2021-03-18
2937 create experiment

kpethtel:
**Story card:** [ch2937](https://app.clubhouse.io/simpledotorg/story/2937/create-new-experiment-and-appointment-models)

## Because

We are adding a new framework for testing the effectiveness of different types and cadences of appointment reminders.

## This addresses

This adds Experiment and AppointmentReminder models.

## Test instructions

Specs will be enough for this.


Gemfile
```diff
@@ -10,6 +10,7 @@ end
 gem "dotenv-rails"
 gem "rails", "5.2.4.5"
 
+gem "active_hash", "~> 2.3.0"
```
kpethtel, line 13: I used this version to match https://github.com/simpledotorg/simple-server/pull/2131/files
So we shouldn't get any conflicts.

config/data/experiments.yml
```diff
@@ -0,0 +1,22 @@
+---
+- id: 123
+  active: true
```
kpethtel, line 3: I'm also considering adding an experiment end date, which will make it easier to ensure that the experiment only runs for 30 days. When appointments are being scheduled, in addition to checking to see if the patient's experiment feature flag is on, it can also check to confirm that the experiment is still running.

db/migrate/20210316154530_create_appointment_reminders.rb
```diff
@@ -0,0 +1,12 @@
+class CreateAppointmentReminders < ActiveRecord::Migration[5.2]
```
kpethtel, line 1: I spent far too much time thinking about whether this should also have a reference to patients. Ultimately, I don't think we need `Patient.appointment_reminders` so I left it out. It's essentially a has_many_through appointments relationship already, so it's easily discoverable.

app/models/experiment.rb
```diff
@@ -0,0 +1,11 @@
+class Experiment < ActiveYaml::Base
+  set_root_path "config/data"
+  set_filename "experiments"
+  field :id
+  field :active
+  field :variations
+
+  def bucket_size
```
kpethtel, line 8: If this seems a little lonely, more methods are coming here.

kpethtel:
Temporarily closing this PR while I consider some changes.

PR #2170, opened 2021-03-19
Dont show drug stock form for community facilities

rsanheim:
**Story card:** [ch2962](https://app.clubhouse.io/simpledotorg/story/2962/only-show-the-drug-stock-form-in-the-mobile-app-for-non-community-facilities)

We only want to intake drug stocks for facilities larger than community -- this small change implements that for the form on the progress tab for mobile.


app/views/api/v3/analytics/user_analytics/show.html.erb
```diff
@@ -111,13 +111,15 @@
...
   <!-- drug stock -->
   <% if current_user.feature_enabled?(:drug_stocks_mobile) %>
-    <div class="card">
-        <h3>Drug stock report</h3>
-        <p style="text-align: left; padding: 6px 0;">Please submit your monthly hypertension drug stock report for end of <%= @for_end_of_month.to_s(:mon_year) %></p>
-        <%= link_to "Submit Drug Stock", new_webview_drug_stock_url(user_id: current_user.id,
-          access_token: current_user.access_token,
-          facility_id: current_facility.id), class: "button" %>
-    </div>
+    <% if current_facility.facility_size != "community" %>
```
harimohanraj89, line 114: How do you feel about `unless`? 😅

harimohanraj89 (changes requested):
Should we add a matching check on the controller too? Does an errant community clinic report submission mess up our numbers? Eg. would it distort a district's drug consumption reports, causing the per-facility numbers in the dashboard to not add up?

app/views/api/v3/analytics/user_analytics/show.html.erb
rsanheim: I dig it. 👍

PR #2175, opened 2021-03-22
Drug stock dashboard fixes

vkrmis:
**Story card:** [ch2984](https://app.clubhouse.io/simpledotorg/story/2984/exclude-community-facilities-from-the-drug-stock-reports)

## Because

Drug stock reports are not relevant to community sized facilities

## This addresses

* Excludes community sized facilities from the Drug stock report pages
* Also fixes a css width issue with tooltips, they can now grow as large as necessary to accommodate drugs with long names
* Removes `Download` button from the Drug Stock report page, since this isn't implemented yet

## Test instructions

* There shouldn't be any community sized facilities in the drug stock report pages
* Long drug names shouldn't break the hover tooltip on drug stock reports
* There shouldn't be a download button on either the Drug Stock or Drug Consumption report pages


app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -77,4 +79,15 @@ def drug_stocks_enabled?
...
+  def populate_facility_sizes
+    @facility_sizes = @accessible_facilities
+      .where(facility_group: @selected_facility_group, zone: @selected_zones)
+      .pluck(:facility_size)
+      .uniq
+      .compact
+      .sort
+      .reject { |size| size == "community" }
+    @facility_sizes = sort_facility_sizes_by_size(@facility_sizes)
+  end
```
ssrihari, line 92: Where is this used?

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -37,8 +37,10 @@ def create
 
   def drug_report
     @facilities = filter_facilities
+      .where.not(facility_size: :community)
```
ssrihari, line 40: Should this move into the drug stocks query, so that it's captured no matter where it is run from?

ssrihari (approved):

app/controllers/my_facilities/drug_stocks_controller.rb
vkrmis: It overrides the `populate_facility_sizes` method in `MyFacilitiesFiltering`

ssrihari: Ah okay, it doesn't very clear to me. Your call though.

Why not do uniq (distinct), compact (where not null), sort (order by), and reject (where not) in sql? Not for performance as much for consistency, and extensibility. Relations are easier to compose in rails than lists :/

ssrihari: And yes, if we move this into drug stock query this confusion goes away too.

app/controllers/my_facilities/drug_stocks_controller.rb
vkrmis: Don't think so. The query should just operate on the data its passed, any filtering should happen at the call site.

ssrihari: Why? Is there a case where it makes sense to report on community facilities as well?

app/controllers/my_facilities/drug_stocks_controller.rb
vkrmis: > Ah okay, it doesn't very clear to me. Your call though.
> 
> Why not do uniq (distinct), compact (where not null), sort (order by), and reject (where not) in sql? Not for performance as much for consistency, and extensibility. Relations are easier to compose in rails than lists :/

This is fair, I did it this way to be consistent with how `MyFacilitiesFiltering` does it currently. Can rewrite it in both places I guess. 
This method is really just trying to fetch a list of sizes `["large", "medium", "small", "community"]` that should be displayed in the size drop down, so the user can select a size to filter by. We're excluding community from this list so that the user can't select it in the drop down.

vkrmis: > And yes, if we move this into drug stock query this confusion goes away too.

This has nothing to do with the Query, just populates options for the drop down facility size filter.

app/controllers/my_facilities/drug_stocks_controller.rb
vkrmis: It totally might end up that way, there's a degree of uncertainty about how this feature will end up working. 
Regardless, it shouldn't be the QueryObject's responsibility to do filtering when we have a concern dedicated to doing this.

app/controllers/my_facilities/drug_stocks_controller.rb
ssrihari: Gotcha. My bad, this is fine then! 👍

app/controllers/my_facilities/drug_stocks_controller.rb
ssrihari: I'm not sure about the difference in concerns here, but the fact that we filter these to show on the filter anyway overrides the perspective I had. The way you've written it makes sense from a reuse angle as well. 👍

PR #2177, opened 2021-03-23
2937 appointment reminder experiment

kpethtel:
**Story card:** [ch2937](https://app.clubhouse.io/simpledotorg/story/2937/create-new-experiment-and-appointment-models)

## Because

We are introducing a method for testing the effectiveness of different types of appointment reminders in bringing patients back to clinics.

## This addresses

Creates models for that testing framework.

## Test instructions

The CI tests should suffice. This is just laying the groundwork.


db/migrate/20210322132613_create_reminder_experiments.rb
```diff
@@ -0,0 +1,10 @@
+class CreateReminderExperiments < ActiveRecord::Migration[5.2]
+  def change
+    create_table :reminder_experiments do |t|
```
kpethtel, line 3: Not using UUID for this model (and ReminderTemplate) because the id will need to be human usable through the command line and UUIDs are not great in that regard.

spec/models/appointment_reminder_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+
+describe AppointmentReminder, type: :model do
+  describe "associations" do
+    it { should belong_to(:appointment) }
+    it { should belong_to(:patient) }
+    xit { should belong_to(:reminder_template) }
```
kpethtel, line 7: This is temporarily disabled because I had issues getting the factory working for reasons I didn't figure out. I think it was somehow related to the use of the Experiment namespace. Any ideas?

db/migrate/20210322132613_create_reminder_experiments.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.boolean :active, null: false
```
kpethtel, line 4: In addition to `active`, I may also add a "type" field because there are two types of experiments. One for "active" patients and one for "inactive" patients and they behave differently. I may need to come up with better names for those groups or for this boolean, because it would be confusing to have an "active" type and an "active" bool. We need this boolean to indicate which experiment to look up when an appointment is created.

app/models/experiment/reminder_experiment.rb
```diff
@@ -0,0 +1,7 @@
+module Experiment
```
kpethtel, line 1: The stated possibility that we may do other experiments led me to put these in a namespace and include the experiment domain (reminders) in the name to avoid squatting on the name "experiment".

db/migrate/20210322132613_create_reminder_experiments.rb
vkrmis: We could use a different human-useable unique field if that's necessary. It might be valuable to keep UUIDs for:
- consistency (its the default for almost all our tables)
- globally uniqueness (hypothetically - comparing data across countries)

db/migrate/20210322132613_create_reminder_experiments.rb
vkrmis: We could also use a `state` enum with `["active", "inactive"]` allowed states instead of an `active` boolean. Will give us the ability to easily add more states if that becomes necessary.

app/models/experiment/reminder_experiment.rb
vkrmis: > It's very possible we'll never create more experiment types.
Agree, but imo we should try to model `Experiment` as a first class AR backed Model. We'll have to make a tradeoff on:
- keep `Experiment` as generic as possible
- avoid adding too much modelling bloat at this stage

db/migrate/20210322132613_create_reminder_experiments.rb
kpethtel: You make a good point about global uniqueness. I hadn't considered it. I still haven't gotten used to the fact that we have multiple unrelated deployments. 

Maybe I'll add a unique friendly name or something. UUID is also manageable, but it could increase the likelihood of human error.

db/migrate/20210322132613_create_reminder_experiments.rb
kpethtel: Not a bad idea.

app/models/experiment/reminder_experiment.rb
kpethtel: Not sure I follow. Are you suggesting changing `Experiment::ReminderExperiment` to just `Experiment`? Is the idea that we could later use single table inheritance if we add more models?

db/migrate/20210322132613_create_reminder_experiments.rb
rsanheim: > Maybe I'll add a unique friendly name or something

Yup, we have FriendlyId already in use in a bunch of other spots for `slug`, so that would work fine.

db/migrate/20210322225751_create_reminder_templates.rb
```diff
@@ -0,0 +1,11 @@
+class CreateReminderTemplates < ActiveRecord::Migration[5.2]
+  def change
+    create_table :reminder_templates do |t|
+      t.integer :experiment_group, null: false
+      t.string :message, null: true
+      t.integer :appointment_offset, null: true
+      t.references :reminder_experiment, null: false, foreign_key: true
+      t.timestamps
```
rsanheim, line 8: `null: false`

db/migrate/20210322132613_create_reminder_experiments.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.date :start_date, null: true
+      t.date :end_date, null: true
+      t.timestamps
```
rsanheim, line 7: `null: false`

db/migrate/20210322225751_create_reminder_templates.rb
```diff
@@ -0,0 +1,11 @@
+class CreateReminderTemplates < ActiveRecord::Migration[5.2]
+  def change
+    create_table :reminder_templates do |t|
+      t.integer :experiment_group, null: false
+      t.string :message, null: true
```
rsanheim, line 5: I think we probably want to enforce `null: false` here...it wouldn't make sense to create a template without a message.

db/migrate/20210322225751_create_reminder_templates.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.integer :appointment_offset, null: true
```
rsanheim, line 6: Does `null: true` make sense here?

db/migrate/20210323141407_create_appointment_reminders.rb
```diff
@@ -0,0 +1,13 @@
+class CreateAppointmentReminders < ActiveRecord::Migration[5.2]
+  def change
+    create_table :appointment_reminders, id: :uuid do |t|
+      t.date :remind_on, null: false
+      t.string :status, null: false
+      t.string :message, null: false
+      t.references :reminder_template, null: true, foreign_key: true
+      t.references :patient, type: :uuid, null: false, foreign_key: true
+      t.references :appointment, type: :uuid, null: false, foreign_key: true
+      t.timestamps
```
rsanheim, line 10: `null: false`

spec/models/appointment_reminder_spec.rb
rsanheim: Can you paste the log of what was going on here?

rsanheim (commented):
Some thoughts on the schema.

spec/models/appointment_reminder_spec.rb
kpethtel: It was definitely a namespacing issue. I'm not sure if there's an easy way around that with this test method. I temporarily moved reminder template out of experiment and found that the test then fails with: 
```
Expected AppointmentReminder to have a belongs_to association called reminder_template (and for the record to fail validation if :reminder_template is unset; i.e., either the association should have been defined with `required: true`, or there should be a presence validation on :reminder_template)
```
So I think this is just a bad test and I should remove it.

db/migrate/20210322225751_create_reminder_templates.rb
kpethtel: I'll add that to all of these. Looking at our current migrations, very few of them enforce this constraint. Perhaps they should.

app/models/experiment/reminder_experiment.rb
vkrmis: I didn't have a specific solution in mind, but I like that approach.

db/migrate/20210322225751_create_reminder_templates.rb
kpethtel: The reason I left message nullable is perhaps a code smell. I was initially going to create an experiment_group (bucket) model that would belong_to reminder_experiment and have many reminder_templates. But it would have contained no data of its own. At most it would need an integer identifier starting at zero for use with modulo operations in the bucketing process. It would just be a join table between experiment and reminder template. This felt like an unnecessary abstraction, so I decided to include experiment_group on reminder_template that can be used for bucketing. 

Because an experiment group can have multiple message templates, that means message templates can look like this:
```
MessageTemplate
experiment_group: 0
message: appointment in 3 days
appointment_offset: -3

MessageTemplate
experiment_group: 0
message: you're late
appointment_offset: 3


MessageTemplate
experiment_group: 1
message: nil
appointment_offset: nil
```

The two things to note are that 
1) there can be multiple reminder_templates with the same experiment group
2) there is a message template representing the control group, which gets no message. And as I type this all out, I feel like I'm convincing myself that perhaps I should make ExperimentGroup a first class citizen. One of the reasons I chose to leave it out is that it felt over-architected to have a an ExperimentGroup that, at most, has an integer identifier to use for matching with modulos. And also because there will be a very complex query at the end of this project for fetching all of the data and I didn't see a good reason to add yet another join.

db/migrate/20210322225751_create_reminder_templates.rb
kpethtel: I believe it does because experiments for non-active patients will not use an offset. They will just be scheduled over the next 30 days.

rsanheim: > They will just be scheduled over the next 30 days.

Scheduled how? Is there an offset hidden there we could capture here?

db/migrate/20210322225751_create_reminder_templates.rb
rsanheim: > The reason I left message nullable is perhaps a code smell. I was initially going to create an experiment_group (bucket) model that would belong_to reminder_experiment and have many reminder_templates. But it would have contained no data of its own.

I think it does capture an important piece of the model though - it captures the intersection of:

* patients
* an experiment
* the reminders to be sent for that experiment

The patients we can capture via some sort of Flipper mechanism (we can discuss that in a sync today), but for Flipper to have a handle on the group we want some notion of that group.  So I think it makes sense to make the ExperimentGroup a first class model.

Then the null: true code smell here goes away.

db/migrate/20210322225751_create_reminder_templates.rb
kpethtel: I just checked the original PRD, and I'm mistaken. They should use the offset, because some of them will get a cascade of messages. Assuming that I also make ExperimentGroup an AR model, this should be required.

app/models/experiment/reminder_experiment.rb
ssrihari: > in the name to avoid squatting on the name "experiment".

I don't think any other model should use the name experiment. It should refer to one thing ideally: experiments. And experiments should be separate from what we're doing inside them.

An Experiment should incorporate the follow concerns:
- Have a unique ID. If we have 2 experiments running at the same time, this ID would differentiate them.
- Be immutable: can't perform updates except to activate and inactivate once only
- Treatments (the right term for "buckets"): what they are, and what the traffic split between them is
- Start and end times, if we ever do scheduling
- The consistent hashing algorithm: can be different for different experiments, but perhaps they can be passed in.

We shouldn't [complect](https://www.merriam-webster.com/dictionary/complect) this concept with Reminders, which are how the Treatments are _executed_. This allows experiments to be reusable across others concerns.

app/models/appointment_reminder.rb
```diff
@@ -0,0 +1,11 @@
+class AppointmentReminder < ActiveRecord::Base
```
ssrihari, line 1: [non blocking] I presume we have discussed why this can't be a `Communication`. Can someone please point me to it?

ssrihari: Asking because we originally considered appointment reminder as the name for this sort of thing, and then picked Communication from FHIR since it was pertinent.

kpethtel: No one has talked to me about what a Communication is at all. I'm aware of the model and it will be involved in this epic, but it was going to be used in the same way that it currently is. AppointmentReminder was going to essentially take the `Appointment#remind_on` functionality, but I wasn't planning to change Communication significantly. Let's discuss this today. I would love to use fewer models if viable, but Communication's current schema doesn't seem viable for this because it lacks remind_on and message attributes.

kpethtel:
Closing for now due to feedback.

db/migrate/20210325162125_create_experiments.rb
```diff
@@ -0,0 +1,14 @@
+class CreateExperiments < ActiveRecord::Migration[5.2]
+  def change
+    create_table :experiments, id: :uuid do |t|
+      t.string :lookup_name, null: false
```
rsanheim, line 4: This could just be name?

app/models/experimentation/experiment.rb
```diff
@@ -0,0 +1,19 @@
...
+    has_many :treatment_buckets
+
+    validates :lookup_name, presence: true, uniqueness: true
+    validates :state, presence: true
+    validates :experiment_type, presence: true
+
+    enum state: {
+      inactive: "inactive",
+      active_selection: "active_selection",
+      active_preselected: "active_preselected"
```
rsanheim, line 12: What does active_preselected mean?

app/models/experimentation/experiment.rb
```diff
@@ -0,0 +1,19 @@
+module Experimentation
+  class Experiment < ActiveRecord::Base
+    has_many :treatment_buckets
+
+    validates :lookup_name, presence: true, uniqueness: true
+    validates :state, presence: true
+    validates :experiment_type, presence: true
+
+    enum state: {
```
rsanheim, line 9: We probably also want to have a `completed` state so we can easily grab finished experiments for data collection.

db/migrate/20210325162125_create_experiments.rb
kpethtel: Totally. In fact, it was name until yesterday. I changed it to make clear that the name is primarily to enable easier lookups. But I'm happy with name.

app/models/experimentation/experiment.rb
```diff
@@ -0,0 +1,21 @@
...
+      new: "new",
+      selecting: "selecting",
+      live: "live",
+      complete: "complete"
+    }, _prefix: true
+    enum experiment_type: {
+      current_patient_reminder: "current_patient_reminder",
+      stale_patient_reminder: "stale_patient_reminder"
```
rsanheim, line 18: Name feedback: I think `current_patient` and `inactive_patient` may be more clear.  Or perhaps `stale_patient` if you want to avoid "inactive".

spec/factories/experimentation/reminder_templates.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :reminder_template, class: Experimentation::ReminderTemplate do
+    message { "Your appointment is in three days" }
+    appointment_offset { -3 }
```
rsanheim, line 4: Lets not force negative numbers here...we could use `offset.days.from_now` in the code, and I think we can avoid having negatives like this.

app/models/experimentation/treatment_group.rb
```diff
@@ -0,0 +1,11 @@
+module Experimentation
+  class TreatmentGroup < ActiveRecord::Base
+    belongs_to :experiment
+    has_many :reminder_templates
+    has_many :treatment_group_memberships
+    has_many :patients, through: :treatment_group_memberships
+
+    validates :index, presence: true, numericality: {greater_than_or_equal_to: 0}
```
rsanheim, line 8: Index should also be unique within the experiment, correct?

rsanheim (approved):
Some naming feedback.

Otherwise this looks good, and I believe it is safe to merge to get these models into master. Nothing is really wired here until the next PR comes w/ ExperimentControl, right?

app/models/experimentation/treatment_group.rb
kpethtel: Yes. I was planning to write more model validations while I write a script for setting up experiments from a hash.

spec/factories/experimentation/reminder_templates.rb
kpethtel: Not sure I follow. The offset is relative to the appointment, not relative to now.

app/models/experimentation/experiment.rb
kpethtel: I'm open to change this. Daniel has also been calling them "dusty" patients, but inactive is probably the most immediately understandable word choice.

app/models/appointment_reminder.rb
```diff
@@ -0,0 +1,16 @@
+class AppointmentReminder < ApplicationRecord
+  belongs_to :appointment
+  belongs_to :patient
+  belongs_to :experiment, class_name: "Experimentation::Experiment", optional: true
```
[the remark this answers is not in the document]
kpethtel: We definitely don't need this. I believe I added it because I expected it would make some sql queries simpler, but I agree that it's redundant information.

app/models/experimentation/experiment.rb
rsanheim: we can handle this in a follow up PR

app/models/experimentation/treatment_group.rb
rsanheim: follow up PR

db/migrate/20210325164437_create_appointment_reminders.rb
```diff
@@ -0,0 +1,15 @@
+class CreateAppointmentReminders < ActiveRecord::Migration[5.2]
+  def change
+    create_table :appointment_reminders, id: :uuid do |t|
+      t.date :remind_on, null: false
```
rsanheim, line 4: Should this be a timestamp?

rsanheim: Nope, because appointments are just a date as well...so the time a reminder is sent will be handled by our notification service.

PR #2178, opened 2021-03-24
Enable patient dedup UI for org managers

prabhanshuguptagit:
**Story card:** [ch2998](https://app.clubhouse.io/simpledotorg/story/2998/release-merge-ui-to-early-users-org-managers)

## Because
We want to preview the merge UI to WHO and ICMR folks before releasing it to CVHOs.

## This addresses

Allows org managers to view the page instead of power users. The patients are not scoped by permissions yet, this will be added in a follow up PR when we release to CVHOs. This is ok for the early release since we only have one org right now.


app/controllers/admin/deduplicate_patients_controller.rb
```diff
@@ -3,15 +3,15 @@ class Admin::DeduplicatePatientsController < AdminController
   DUPLICATE_LIMIT = 250
 
   def show
-    authorize { current_admin.power_user? }
+    authorize { current_admin.current_admin.accessible_organizations(:manage).any? }
```
kpethtel, line 6: Did you mean to duplicate `current_admin` here?

prabhanshuguptagit: Oh thanks 👾

PR #2181, opened 2021-03-24
Registration buffer reporting bug (and fix some other possible edge cases)

rsanheim:
**Story card:** [ch2861](https://app.clubhouse.io/simpledotorg/story/2861/reports-bug-missing-visits)

We are miscounting some patients where their registration date falls on the boundary of our three month 'registration buffer'.

For example:

Patient registers on December 31st.  An admin views a report in March, which _should_ include that patient in the control / uncontrolled counts for the current report. Currently the report does not include that patient for controlled counts, because the Period object returns a Date for the `recorded_at` where clause, and not a Time.  So we have a clause like this in the SQL:

```
AND (patient_recorded_at <= '2020-12-31')
```

which PostgreSQL converts to a timestamp on the fly (since recorded_at is a timestamp), and when that is done it ends up being `2020-12-31 00:00:00` ...i.e the beginning of the day.

This PR get consistent with the names `begin` and `end` to determine the begin and end of Periods, and then ensures that we return the beginning of the day timestamp or the end of the day timestamp, respectively, to fix this issue.

### DIFF
--- a/app/models/period.rb
+++ b/app/models/period.rb
@@ -59,8 +59,8 @@ class Period
   # Returns a range of dates that correspond to the 'control range' for this period.
   # For example, for a month period of July 1st 2020, this will return the range of April 30th..July 31st.
   def blood_pressure_control_range
-    three_months_ago = end_date.advance(months: -3).end_of_month
-    (three_months_ago..end_date)
+    three_months_ago = end_time.advance(months: -3).end_of_month
+    (three_months_ago..end_time)
   end
 
   alias_method :bp_control_range, :blood_pressure_control_range
@@ -94,14 +94,14 @@ class Period
   end
   alias_method :begin, :start_date
 
-  def end_date
+  def end
     if quarter?
       value.end
     else
       value.end_of_month.end_of_day
     end
   end
-  alias_method :end, :end_date
+  alias_method :end_time, :end
 
   def succ
     if quarter?
--- a/app/models/period.rb
+++ b/app/models/period.rb
@@ -85,14 +85,14 @@ class Period
     value.to_date
   end
 
-  def start_date
+  def begin
     if quarter?
       value.begin
     else
       value.beginning_of_month.beginning_of_day
     end
   end
-  alias_method :begin, :start_date
+  alias_method :start_time, :begin
 
   def end
     if quarter?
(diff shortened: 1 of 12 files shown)


app/queries/control_rate_query.rb
```diff
@@ -41,8 +41,8 @@ def bp_monthly_query(region, period, with_exclusions)
       .for_reports(with_exclusions: with_exclusions)
       .select("distinct on (latest_blood_pressures_per_patient_per_months.patient_id) *")
       .where(assigned_facility_id: region.facilities)
-      .where("patient_recorded_at < ?", control_range.begin) # TODO this doesn't seem right -- revisit this exclusion
```
kpethtel, line 44: I guess something wasn't quite right.

kpethtel (approved):
Nothing is jumping out at me. Looks good as far as I can tell.

app/services/no_bp_measure_service.rb
```diff
@@ -38,32 +38,32 @@ def cache_keys_for_period
 
   def execute_sql(period)
     return 0 if facility_ids.blank?
-    start_date = period.blood_pressure_control_range.begin
-    end_date = period.blood_pressure_control_range.end
-    registration_date = period.blood_pressure_control_range.begin
+    start_time = period.blood_pressure_control_range.begin
+    end_time = period.blood_pressure_control_range.end
+    registration_buffer = period.blood_pressure_control_range.begin
```
prabhanshuguptagit, line 43: Should this be using `adjusted_period` instead? This is confusing and I suppose it works because `blood_pressure_control_range.begin` and `adjusted_period.end` are the same. Should use the right method here and in `NoBPMeasureQuery `.

app/models/period.rb
```diff
@@ -50,21 +56,21 @@ def to_quarter_period
...
 
   alias_method :bp_control_range, :blood_pressure_control_range
 
   def bp_control_registrations_until_date
-    bp_control_range.begin.to_s(:day_mon_year)
+    adjusted_period.end.to_s(:day_mon_year)
   end
 
   def bp_control_range_start_date
-    bp_control_range.begin.next_day.to_s(:day_mon_year)
```
prabhanshuguptagit, line 67: ❤️

prabhanshuguptagit (commented):
Looks good except for a few subtle naming things. I will test this on perf and keep you posted.

PR #2185, opened 2021-03-26
Add LTFU toggles for uncontrolled and missed visits cards

prabhanshuguptagit:
**Story card:** [ch2936](https://app.clubhouse.io/simpledotorg/story/2936/add-ltfu-toggle-for-power-users-to-bp-uncontrolled-and-missed-visits-in-reports-overview)

## This addresses

image


app/views/reports/regions/show.html.erb
```diff
@@ -1,117 +1,53 @@
...
+              <%= render "definition_tooltip",
+                         definitions: { "Numerator" => t("bp_controlled_copy.numerator"),
+                                        "Denominator" => t(control_graph_denominator_copy, region_name: @region.name) } %>
             </div>
+            <%= render "reports/regions/ltfu_toggle", enabled: @with_ltfu %>
           <% else %>
             <h3 class="mb-0px w-75">
               BP controlled
             </h3>
           <% end %>
```
claudiovallejo, line 27: @prabhanshuguptagit maybe I'm missing something, but I don't think the if/else is necessary right. We use the same title, and the ltfu_toggle partial is conditionally rendered already.

claudiovallejo: If the if/else isn't necessary, this comment applies to all other card titles that use it.

app/views/reports/regions/show.html.erb
```diff
@@ -325,134 +211,60 @@
...
             <h3 class="mb-0px mr-8px">
               Registrations
             </h3>
             <%= render "definition_tooltip",
                        definitions: { "Total registered patients" => t("registered_patients_copy.total_registered_patients", region_name: @region.name),
                                       "Monthly registered patients" => t("registered_patients_copy.monthly_registered_patients", region_name: @region.name) } %>
           </div>
+        <% else %>
+          <h3 class="mb-0px w-75">
+            Monthly and cumulative registrations
```
claudiovallejo, line 231: @prabhanshuguptagit this is title should be removed, it's old and should always show "Registrations" if we're now showing registrations instead of assigned patients.

claudiovallejo (commented):
Code and UI looks good @prabhanshuguptagit! Left a few comments and only UI feedback is that toggle is clicked, because the spans are empty, the text jumps left and right. Maybe we can add a min-width to the spans to avoid this? Other ideas on how to solve this?

app/views/reports/regions/show.html.erb
prabhanshuguptagit: The if-else toggles the new tooltip and some CSS. The toggle partial isn't conditionally rendered, so we need it for that too.

app/views/reports/regions/show.html.erb
prabhanshuguptagit: This will get cleaned up in the feature flag code cleanup, will remove in a follow up PR.

PR #2210, opened 2021-03-30
Finish migrating control rate service to using Repository

rsanheim:
**Story card:** [ch3205](https://app.clubhouse.io/simpledotorg/story/3205/migrate-all-of-controlrateservice-to-use-repository)

An earlier PR (https://github.com/simpledotorg/simple-server/pull/2059) started the move to Repository and helped flesh out the API and prove that its an improvement.

The goal w/ this PR is to complete the move for regions#show.

See also the related ADR in https://github.com/simpledotorg/simple-server/pull/2328.


prabhanshuguptagit:
@rsanheim Just seeing this, heads up that there might be some hairy conflicts with https://github.com/simpledotorg/simple-server/pull/2334. If this PR is close to ready, I'll wait for this to be merged before I cleanup.

rsanheim (commented):
going to pause on this until `with_exclusions` is entirely gone

kpethtel (approved):
This is very hard code to review and I'm sure it was very hard to work on. I can't claim to fully grok it in a big picture sense, but it looks good to me. 🚀

app/models/reports/repository.rb
```diff
@@ -63,42 +76,46 @@ def self.smart_memoize(method)
       end
     end
 
+    alias_method :adjusted_patient_counts_without_ltfu, :adjusted_patient_counts
```
prabhanshuguptagit, line 79: @rsanheim We should just call this `adjusted_patients_counts`, since it is the default now. The other one already implies that it is the `with_ltfu` version.

PR #2329, opened 2021-04-01
experiment spike

vkrmis:
This is a non functional prototype, branched off of [2937-appointment-reminder-experiment](https://github.com/simpledotorg/simple-server/tree/2937-appointment-reminder-experiment)

Adds `ExperimentNotification::Worker` and `ExperimentNotificationService` which are meant to be extremely similar to their `Appoinment` counterparts.


app/models/experimentation/experiment.rb
```diff
@@ -13,9 +13,15 @@ class Experiment < ActiveRecord::Base
       live: "live",
       complete: "complete"
     }, _prefix: true
-    enum experiment_type: {
-      current_patient_reminder: "current_patient_reminder",
-      stale_patient_reminder: "stale_patient_reminder"
-    }, _prefix: true
```
vkrmis, line 19: Dropping the `experiment_type` and replacing it with `fetch_message` and `queueable_message_exists?` below. Since we're modelling and recording treatment groups and membership, these can just be looked up as necessary.

app/services/appointment_notification_service.rb
```diff
@@ -16,6 +16,7 @@ def initialize(appointments:, days_overdue: 3)
 
   def send_after_missed_visit
     eligible_appointments = appointments.eligible_for_reminders(days_overdue: days_overdue)
+    # Exclude patients enrolled in an experiment
```
vkrmis, line 20: We'll need to be able to figure out whether an appointment is part of an experiment or not, so we can skip sending the usual 3 days overdue message. We could do this through a method that calculates based on enrollment and dates, or alternatively, add a nullable `experiment_id` column on appointments.

kpethtel:
Closing this PR because this ship has sailed.

PR #2332, opened 2021-04-01
2938 experiment controls 3

kpethtel:
**Story card:** [ch2938](https://app.clubhouse.io/simpledotorg/story/2938)

## Because

We need a mechanism for starting experiments.


app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+                appointments.scheduled_date >= ?)", date)
+        .order(Arel.sql("random()"))
+        .to_a
+
+      total_days.times do
+        daily_patients = eligible_patients.pop(10_000)
+        break if daily_patients.empty?
+        daily_patients.each do |patient|
+          group = experiment.group_for(patient.id)
+          appointment = patient.appointments.where(status: "scheduled").last
```
kpethtel, line 59: I'm not sure if this will be reliable. I had initially planned to find the last scheduled or cancelled appointment within the eligibility range, but given that these patients are selected based on encounter data and not patient data, I can't be sure what appointment data will exist. I need some appointment to attach to the reminders. It might be safer just to create an appointment for this purpose.

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    def start_stale_patient_experiment(name, total_days)
+      experiment = Experimentation::Experiment.find_by!(name: name, experiment_type: "stale_patient_reminder")
+      date = Date.current
+      eligibility_start = (date - 365.days).beginning_of_day
+      eligibility_end = (date - 35.days).end_of_day
+
+      experiment.state_selecting!
+
+      eligible_patients = patient_pool
```
kpethtel, line 43: If anyone knows more elegant approaches to this sql, I'm open to suggest. The general gist here is:
patients who had an encounter during the date range, haven't had an encounter after the date range, and don't have any appointments in the future. That last part may not be necessary–hard to know without knowing more about how appointments are scheduled–but I thought it wise to ensure that a patient can't be eligible for both experiment types.

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    protected
+
+    def patient_pool
+      Patient.from(Patient.with_hypertension, :patients)
+        .contactable
+        .where("age >= ?", 18)
+        .includes(treatment_group_memberships: [treatment_group: [:experiment]])
+        .where(["experiments.end_date < ? OR experiments.id IS NULL", 14.days.ago]).references(:experiment)
```
rsanheim, line 76: 14.days could become a constant, maybe `LAST_EXPERIMENT_BUFFER` ?

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    def schedule_reminders(patient, appointment, experiment, group, schedule_date)
+      group.reminder_templates.each do |template|
+        remind_on = schedule_date + template.appointment_offset.days
+        AppointmentReminder.create!(
```
rsanheim, line 82: This could hang off the experiment to make things more clear:

```
experiment.appointment_reminders.create! ....
```

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+      eligible_patients = patient_pool
+        .joins(:encounters)
+        .where("encounters.device_created_at BETWEEN ? AND ?", eligibility_start, eligibility_end)
+        .where("NOT EXISTS (SELECT 1 FROM encounters WHERE encounters.patient_id = patients.id AND
+                encounters.device_created_at > ?)", eligibility_end)
+        .left_joins(:appointments)
+        .where("NOT EXISTS (SELECT 1 FROM appointments WHERE appointments.patient_id = patients.id AND
+                appointments.scheduled_date >= ?)", date)
+        .order(Arel.sql("random()"))
+        .to_a
```
rsanheim, line 52: We should definitely not be doing `to_a` on a large relation of patients like this.  Lets discuss this during a pair session tomorrow.

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      total_days.times do
+        daily_patients = eligible_patients.pop(10_000)
```
rsanheim, line 55: Another constant we can extract here.

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
+class InvalidExperiment < StandardError; end
+
+class ExperimentControlService
+  class << self
+    def start_current_patient_experiment(name, days_til_start, days_til_end)
+      raise ArgumentError, "Start date must be before end date" if days_til_end < days_til_start
+
+      existing_experiment = Experimentation::Experiment.where(experiment_type: "current_patient_reminder", state: ["selecting", "live"])
+      raise InvalidExperiment, "A current patient experiment is already in progress" if existing_experiment.any?
```
rsanheim, line 9: This could also be enforced on the model...and probably would be good to do for added safety.

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      experiment = Experimentation::Experiment.find_by!(name: name, experiment_type: "current_patient_reminder")
+      experiment_start = days_til_start.days.from_now.beginning_of_day
+      experiment_end = days_til_end.days.from_now.end_of_day
+
+      experiment.state_selecting!
```
rsanheim, line 15: something like `start_selecting!` would be more clear here I think. 

We also may want to bring in a proper state machine gem, let's zoom about it.

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      experiment_patients = patient_pool
```
rsanheim, line 17: We should apply some percentage to patient_pool here, at a minimum so we can do a slow rollout of this and monitor in SBX and later production.

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+        .joins(:appointments)
+        .where("appointments.status = ?", "scheduled")
+        .where("appointments.scheduled_date BETWEEN ? AND ?", experiment_start, experiment_end)
+        .order(Arel.sql("random()"))
+
+      experiment_patients.each do |patient|
```
rsanheim, line 23: `find_each` here to loop through batches would probably be good

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+        group = experiment.group_for(patient.id)
+        appointment = patient.appointments
```
rsanheim, line 25: How about we had a scope or method somewhere to make this easier:

```
patient.appointments.scheduled_between(experiment_start, experiment_end)
```

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+          .where(status: "scheduled")
+          .where("appointments.scheduled_date BETWEEN ? AND ?", experiment_start, experiment_end).first
```
rsanheim, line 27: Why just the first? If we get multiple w/I the experiment time, wouldn't we want reminders for all?

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+        schedule_reminders(patient, appointment, experiment, group, appointment.scheduled_date)
```
rsanheim, line 28: We are passing a lot of state around here....could we get this down to perhaps just `patient, appointment, group` ? I think we can derive the rest from that...

rsanheim (commented):
Looks like a good start! There is a bunch going on here, lets pair on this some to try to tease apart some of the abstractions and discuss some of the feedback I'm leaving.

app/services/experiment_control_service.rb
kpethtel: That was the original plan, but Daniel mentioned that the plan was for experiments to include all eligible patients in a nation. I'll re-add that.

app/services/experiment_control_service.rb
kpethtel: That's a valid point. I'd been told a patient should only have one scheduled appointment, but I should still code on the assumption that that might not be true.

app/services/experiment_control_service.rb
kpethtel: Agreed. Performance was something I wanted your input on for this script in general. I'm sure we can just select ranges or something rather than popping.

app/services/experiment_control_service.rb
kpethtel: We can definitely derive experiment, but schedule_date does not come from appointment for stale patients. But that could change depending on another open question. In a comment above, I had said that grabbing the last appointment for a stale patient might not be reliable and that it might be better to create an appointment for this purpose. If we created an appointment for the date the patient's reminders should start, we would not need an additional argument for schedule date.

app/services/experiment_control_service.rb
kpethtel: That sounds like a good idea. I'll work on that.

app/services/experiment_control_service.rb
kpethtel: Experiment does not currently have a relationship with appointment reminders. I added the experiment id to appointment reminder at the last minute as part of the other PR but forgot to add the `has_many` to the experiment. But you suggested getting rid of the explicit relationship altogether here:
`https://github.com/simpledotorg/simple-server/pull/2177/files#r603542076`

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
+class InvalidExperiment < StandardError; end
+
+class ExperimentControlService
+  class << self
+    def start_current_patient_experiment(name, days_til_start, days_til_end)
+      raise ArgumentError, "Start date must be before end date" if days_til_end < days_til_start
```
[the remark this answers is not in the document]
kpethtel: Yeah, that makes sense.

app/services/experiment_control_service.rb
kpethtel: If I'm adding the date validations to the model, I will have to update the dates here along with the state to validate the dates. So the name of the method (which is just an enum method) becomes irrelevant, but we still might want a state machine.

app/services/experiment_control_service.rb
kpethtel: You're making good points about model validations. As I said in the other PR, I was planning to write a script for adding experiment data soon and intended to add more validations at that time, but it looks like some of it is necessary here.

app/services/experiment_control_service.rb
```diff
@@ -0,0 +1,94 @@
...
+        Experimentation::TreatmentGroupMembership.create!(treatment_group: group, patient: patient)
+      end
+
+      experiment.update!(state: "live", start_date: experiment_start, end_date: experiment_end)
+    end
+
+    def start_stale_patient_experiment(name, total_days)
+      experiment = Experimentation::Experiment.find_by!(name: name, experiment_type: "stale_patient_reminder")
+      date = Date.current
+      eligibility_start = (date - 365.days).beginning_of_day
```
kpethtel, line 38: These numbers should probably also be configurable.

kpethtel:
Need to make a major change.

PR #2334, opened 2021-04-02
Cleanup flipper flag for exclusions 

prabhanshuguptagit:
**Story card:** [ch3001](https://app.clubhouse.io/simpledotorg/story/3001/cleanup-usage-of-flipper-flag-report-with-exclusions-after-release)

## Because

We have released denominator exclusions to all envs and can now remove the `with_exclusions` feature flag.

## This addresses

This makes all the reports exclude dead patients by default. The LTFU exclusion is also applied as a default in the appropriate places. It also removes the `with_exclusions` optional param that was passed in all the query classes.


kpethtel (approved):
This will make the html much easier to read. Looks good.

app/models/reports/repository.rb
```diff
@@ -149,14 +147,12 @@ def self.smart_memoize(method)
...
+    # Repository only returns "excluding LTFU" rates.
+    # This only powers queries for children regions which do not require both variants of control rates, unlike Result.
+    # As we deprecate Result and shift to Repository, a Repository object should be able to return both rates.
     def denominator(region, period)
-      if with_exclusions
-        cumulative_assigned_patients_count[region.slug][period.adjusted_period] - ltfu_counts[region.slug][period]
-      else
-        cumulative_assigned_patients_count[region.slug][period.adjusted_period]
-      end
+      cumulative_assigned_patients_count[region.slug][period.adjusted_period] - ltfu_counts[region.slug][period]
```
prabhanshuguptagit, line 155: @rsanheim This method now only return the `with_exclusions` version right now, since we don't use the included version anywhere. When we decide to pull rates from this class, we'll need to do something like https://github.com/simpledotorg/simple-server/blob/8949d9ef94765ef0813f25518ac1f3f4e21c0c58/app/models/reports/result.rb#L218

app/models/concerns/patient_reportable.rb
```diff
@@ -25,18 +25,14 @@ module PatientReportable
...
-            .excluding_dead
-        end
+    scope :for_reports, ->(exclude_ltfu_as_of: nil) do
+      if exclude_ltfu_as_of
+        with_hypertension
+          .excluding_dead
+          .not_ltfu_as_of(exclude_ltfu_as_of)
       else
         with_hypertension
+          .excluding_dead
```
rsanheim, line 35: This would be more clear as something like this:

```ruby
 scope :for_reports, ->(exclude_ltfu_as_of: nil) do
    query = with_hypertension.excluding_dead
    query.not_ltfu_as_of(exclude_ltfu_as_of) if exclude_ltfu_as_of
    query
end
```

app/queries/district_analytics_query.rb
```diff
@@ -39,7 +39,7 @@ def results
   def total_assigned_patients
     @total_assigned_patients ||=
       Patient
-        .with_hypertension
+        .for_reports
```
rsanheim, line 42: 👍 I guess there were quite a few places where we were missing the dead exclusions.

app/queries/no_bp_measure_query.rb
```diff
@@ -12,7 +15,7 @@ def call(region, period, with_exclusions: false)
     registration_date = period.blood_pressure_control_range.begin
 
     Patient
-      .for_reports(with_exclusions: with_exclusions, exclude_ltfu_as_of: period.end_time)
+      .for_reports(exclude_ltfu_as_of: period.end_time)
```
rsanheim, line 18: I'm thinking the `for_reports` should be responsible for doing the `end_time` for periods...that way callers don't need to worry about it.

app/models/concerns/patient_reportable.rb
rsanheim: Maybe we do `exclude_ltfu_as_of.end_time` if all callers are consistently passing in a Period?  Or we could gate that on a `respond_to?` to be safe.

app/views/reports/regions/_child_data_tables_with_exclusions.html.erb
```diff
@@ -1,204 +0,0 @@
-<div class="card mb-0">
```
rsanheim, line 1: 🔥 🔥 🔥 🔥 🔥 🔥

rsanheim (commented):
Love to see all this consolidated 👍 .

rsanheim (approved):
Meant to approve (with the minor feedback) in that last review.

So here is an approval.

app/queries/district_analytics_query.rb
prabhanshuguptagit: Yep, found this in `CohortAnalyticsQuery` and `DistrictAnalyticsQuery`.

app/models/concerns/patient_reportable.rb
prabhanshuguptagit: The `end_time` is a query-specific clause in the two places `exclude_ltfu_as_of` used currently. "Which LTFU patients does the query want to remove?". I think we should let the query make that decision, keeps the scope decoupled.

app/views/reports/regions/_facility_tables.html.erb
```diff
@@ -1,277 +0,0 @@
-<h3 class="mt-48px mb-8px fs-28px c-black">
```
prabhanshuguptagit, line 1: These 4 tables were displayed in the facilities page at the bottom earlier, we don't show them anymore.

app/jobs/patient_list_download_job.rb
```diff
@@ -15,12 +15,7 @@ def perform(recipient_email, model_type, params, with_medication_history: false,
       raise ArgumentError, "unknown model_type #{model_type.inspect}"
     end
 
-    patients =
-      if with_exclusions
-        Patient.excluding_dead.where(id: model.assigned_patients)
-      else
-        model.assigned_patients
-      end
+    patients = model.assigned_patients.excluding_dead
```
harimohanraj89, line 18: This is slightly different from the existing implementation `Patient.excluding_dead.where(id: model.assigned_patients)`.

Are the two queries the same? Are there performance implications?

app/services/cohort_service.rb
```diff
@@ -48,11 +47,7 @@ def default_range
   end
 
   def cache_key(period)
-    if @with_exclusions
-      "#{self.class}/#{region.cache_key}/#{period.cache_key}/with_exclusions"
-    else
-      "#{self.class}/#{region.cache_key}/#{period.cache_key}"
-    end
+    "#{self.class}/#{region.cache_key}/#{period.cache_key}"
```
harimohanraj89, line 50: ~When we ship, our cache will temporarily hold the "without exclusions" value in this key. We'll need to warm the cache right after shipping?~

I love our cache versioning ❤️

app/views/my_facilities/missed_visits.html.erb
```diff
@@ -16,11 +16,7 @@
         <p class="c-grey-dark c-print-black">
           <strong>Numerator:</strong> <%= t("missed_visits_copy.numerator") %>
           <strong>Denominator:</strong>
-          <% if current_admin.feature_enabled?(:report_with_exclusions) %>
-            <%= t("denominator_excluding_ltfu_copy", region_name: "each facility") %>
-          <% else %>
-            <%= t("denominator_without_exclusions_copy", region_name: "each facility") %>
-          <% end %>
+          <%= t("denominator_excluding_ltfu_copy", region_name: "each facility") %>
```
harimohanraj89, line 19: Can we rename to `denominator_copy` now?

app/views/my_facilities/missed_visits.html.erb
```diff
@@ -16,11 +16,7 @@
         <p class="c-grey-dark c-print-black">
           <strong>Numerator:</strong> <%= t("missed_visits_copy.numerator") %>
           <strong>Denominator:</strong>
-          <% if current_admin.feature_enabled?(:report_with_exclusions) %>
-            <%= t("denominator_excluding_ltfu_copy", region_name: "each facility") %>
-          <% else %>
-            <%= t("denominator_without_exclusions_copy", region_name: "each facility") %>
```
harimohanraj89, line 22: Can this YML key be removed from the locale now?

app/jobs/patient_list_download_job.rb
prabhanshuguptagit: The newer query makes for simpler SQL.
```
> puts f.assigned_patients.excluding_dead.to_sql
SELECT "patients".* FROM "patients"
WHERE "patients"."deleted_at" IS NULL
AND "patients"."assigned_facility_id" = '10949769-acb5-4786-8c5c-9109f5f3f509'
AND "patients"."status" != 'dead'


> puts Patient.excluding_dead.where(id: f.assigned_patients).to_sql
SELECT "patients".* FROM "patients" 
WHERE "patients"."deleted_at" IS NULL
AND "patients"."status" != 'dead'
AND "patients"."id" IN (SELECT "patients"."id" FROM "patients" WHERE "patients"."deleted_at" IS NULL AND "patients"."assigned_facility_id" = '10949769-acb5-4786-8c5c-9109f5f3f509')
```

It's also a tad faster.
```
[12] pry(main)> Benchmark.ms { f.assigned_patients.excluding_dead }
=> 0.486999997519888
[13] pry(main)> Benchmark.ms { Patient.excluding_dead.where(id: f.assigned_patients) }
=> 0.8010000019567087
```

app/views/my_facilities/missed_visits.html.erb
prabhanshuguptagit: Addressed in 4c9b2f3.

app/views/my_facilities/missed_visits.html.erb
prabhanshuguptagit: Addressed in 6f099c3

app/jobs/patient_list_download_job.rb
harimohanraj89: Suite.

harimohanraj89 (approved):
Looking slick ✨

PR #2336, opened 2021-04-02
Drug stock form refresh

rsanheim:
Auto update the form (via javascript) when the user picks a different month.

Also fixes an issue in the report where we don't have drugs in stock but still need to show a row .

**Story card:** [ch2996](https://app.clubhouse.io/simpledotorg/story/2996/ensure-the-drug-stock-form-refreshes-with-latest-data-if-month-is-changed)


app/views/webview/drug_stocks/new.html.erb
```diff
@@ -77,5 +74,16 @@
...
   </body>
+
+  <script type="text/javascript" charset="utf-8">
+  window.addEventListener("load", function() {
+    monthSelect = document.getElementById("for_end_of_month");
+    monthSelect.addEventListener("change", function(e) {
+      window.location.href = location.origin + location.pathname + location.search + "&for_end_of_month=" + encodeURIComponent(monthSelect.value);
+    });
+  })
+  </script>
```
rsanheim, line 87: @msasikanth Do you know if adding javascript here will require any changes to the Android side?

We could make this work without JS, it would just require some re-working of the date selection UI and wouldn't be as nice.

PR #2387, opened 2021-04-06
Make org wide reports viewable behind a feature flag

rsanheim:
**Story card:** [ch3020](https://app.clubhouse.io/simpledotorg/story/3020/organization-wide-reports)

## Because

Some countries (like ET and BD) could benefit from seeing _all_ the data available in their country in dashboard reports.

## This addresses

Adds org wide reports behind a feature flag `organization_reports`.

### DIFF
--- a/app/controllers/reports/regions_controller.rb
+++ b/app/controllers/reports/regions_controller.rb
@@ -161,6 +161,9 @@ class Reports::RegionsController < AdminController
     report_scope = report_params[:report_scope]
     @region ||= authorize {
       case report_scope
+      when "organization"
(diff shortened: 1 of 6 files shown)


app/controllers/reports/regions_controller.rb
```diff
@@ -161,6 +161,9 @@ def find_region
     report_scope = report_params[:report_scope]
     @region ||= authorize {
       case report_scope
+      when "organization"
```
kpethtel, line 164: In the specs, I see a series of tests like, "retrieves block data". This should probably have some similar specs.

kpethtel (changes requested):

kpethtel (approved):
👍

PR #2330, opened 2021-04-01
ETL Prototyping

harimohanraj89:
**Story card:** TBD
## Because

Enter the reason for raising this PR...

## This addresses

Enter the details of what all this covers...

## Test instructions

Enter detailed instructions for how to test this PR...

PR #2388, opened 2021-04-07
ETL Pipeline ADR

harimohanraj89:
**Story card:** [ch3150](https://app.clubhouse.io/simpledotorg/story/3150/create-an-adr-for-reporting-etl-pipeline)

An ADR for the architectural changes in the upcoming Reporting ETL Pipeline.


doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,90 @@
...
+                                 (latest BP for each patient as of every month)   (as of every month, how many BPs were
+                                                                                   controlled and recent)
+```
+
+The reporting tables will be implemented as database views within the Simple database.
+
+We will use Metabase for the reports.
+* Easy-to-use interface
+* Easy to grant access to stakeholders, and controlled access to ourselves if necessary
+* Dynamic queries (eg. pick your own date range)
```
prabhanshuguptagit, line 59: Are there any cons to metabase? Let's specify if we're aware of any.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,90 @@
...
+```
+ --------------------             ------------------------------------             ------------------
+| Application tables |    -->    | Reporting tables as database views |    -->    | Metabase reports |
+ --------------------             ------------------------------------             ------------------
+eg. blood_pressures              eg. blood_pressures_over_time                    eg. monthly control rate query
+                                 (latest BP for each patient as of every month)   (as of every month, how many BPs were
+                                                                                   controlled and recent)
+```
+
+The reporting tables will be implemented as database views within the Simple database.
```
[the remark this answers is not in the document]
prabhanshuguptagit: Let us specify the granularity/shape of these tables also:
- On the time axis, we want to slice by `month`
- On the region axis, we want to slice by `facility`

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,90 @@
...
+## Status
+
+Proposed
+
+## Consequences
+
+* We will have more nightly cron tasks to refresh the reporting views. We need to monitor the performance of these
+  refreshes closely to identify when database views are no longer a viable option for us.
+* We can start thinking of migrating existing dashboard reports to consume data from the reporting tables.
+* We will need to maintain and evolve the reporting schema as reporting requirements change.
```
prabhanshuguptagit, line 90: [nit] This is kind of understood, no? 😛

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,90 @@
...
+                                                                                   controlled and recent)
+```
+
+The reporting tables will be implemented as database views within the Simple database.
+
+We will use Metabase for the reports.
+* Easy-to-use interface
+* Easy to grant access to stakeholders, and controlled access to ourselves if necessary
+* Dynamic queries (eg. pick your own date range)
+
```
prabhanshuguptagit, line 60: Looking at #2330, it seems like `calendar_month` will be a centrepiece for how we handle time. Should we include some details about what our time abstraction might look like?

prabhanshuguptagit (commented):
This is a good start. We can probably add some more details around:

- how the ETL pipeline will affect our current approach to reporting as a whole.
- the shape of the ETL tables: what columns to query by, granularity, axes etc.

-- vkrm & prabhanshu

doc/arch/014-reporting-etl-pipeline.md
harimohanraj89: Haha I meant it's and additional moving piece to maintain 😄

doc/arch/013-repository-layer-for-metrics.md
```diff
@@ -2,8 +2,8 @@
 
 ## Context
 
-We generate many reports for the Simple dashboard and the mobile app. Retreving all the data for these reports in a 
-consistent and performant manner is a challenge.  We need to aggregate and filter a large amount of data across multiple reports, and its important for user confidence that we have the same numbers reported regardless of what report they are on.
+We generate many reports for the Simple dashboard and the mobile app. Retrieving all the data for these reports in a
```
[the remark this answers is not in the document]
harimohanraj89: No worries 😅  I'll address the conflicts if #2328 goes in first.

doc/arch/014-reporting-etl-pipeline.md
harimohanraj89: Specified in the long-term section below.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+Simple's application schema is optimized for application transactions like registering patients, recording BPs,
+updating patient info, or tracking overdue patients. As such, reporting queries grow complex. Asking questions like
+"latest BP in a month", "prescribed drugs as of a date", "current step of treatment protocol" involve complex clauses
+in the SQL queries.
+
+This complexity has a tangible impact on the quality of our reports. We risk data inconsistency as complex queries are
+more error-prone when reused in multiple places.
+
+
+### 2. Coupling
```
ssrihari, line 19: I feel the first point is talking about complexity introduced through coupling of the transactional schema with the reporting schema. 

Would it be better to keep this second one about performance?

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+
+### 1. Complexity
+
+Simple's application schema is optimized for application transactions like registering patients, recording BPs,
+updating patient info, or tracking overdue patients. As such, reporting queries grow complex. Asking questions like
+"latest BP in a month", "prescribed drugs as of a date", "current step of treatment protocol" involve complex clauses
+in the SQL queries.
+
+This complexity has a tangible impact on the quality of our reports. We risk data inconsistency as complex queries are
+more error-prone when reused in multiple places.
```
ssrihari, line 16: I think we should mention the consolidation of questions, and research into the reporting schema done in https://github.com/simpledotorg/simple-etl. There's a lot of context and rationale there for this decision.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+### 2. Coupling
+
+When the application schema is used for reporting also, there is a tight coupling between application requirements and
+reporting requirements. Changes to the schema to optimize for reporting may have a negative impact on complexity in the
+application, and vice versa.
+
+In addition to complexity, performance is also a growing concern as we scale. Heavy reporting operations may strain the
+application's ability to write data to the database. Conversely, heavy application activity may impact our ability to
+generate reports in a performant manner. We currently use heavy layers of caching to mitigate these performance
+concerns, but the less caching we _need_ to do, the better.
```
ssrihari, line 28: > but the less caching we _need_ to do, the better.

I think it's worth elaborating on _why_ since this is our current approach.

Caching inherently comes with the added complexity of cache invalidation, which is touted to be one of the harder problems in computer science. Here are some reasons I can think of:
- We're past the point of using simple caching, and we're layering complex abstractions on it.
- We have a few use cases like drug stock, merging duplicate patients, etc where caching is not always an option.
- Our data can be changed retroactively (new data in the past, and updates to measurements), so cache invalidation can potentially become complex.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+
+Simple's dashboard reports are currently generated by queries on the application database's tables. This has several
+disadvantages as we start to scale.
+
+### 1. Complexity
+
+Simple's application schema is optimized for application transactions like registering patients, recording BPs,
+updating patient info, or tracking overdue patients. As such, reporting queries grow complex. Asking questions like
+"latest BP in a month", "prescribed drugs as of a date", "current step of treatment protocol" involve complex clauses
+in the SQL queries.
```
ssrihari, line 13: > involve complex clauses in the SQL queries.

I would phrase this as: involve joining across multiple tables and using significant amount of ruby code to transform the data into reports.

While some queries are complex, it is the amount of ruby code in my opinion that amounts to complexity. Further, we're actually adding significantly complex queries to create these schemas :) So we're actually not shying away from complex SQL queries no? We want overall report creation: sql + ruby to be simpler.

doc/arch/014-reporting-etl-pipeline.md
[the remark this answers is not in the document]
ssrihari: We do say materialized above. However, I think we should call out that materialized or not is mostly just based on performance requirements. It's perfectly legitimate for us to create non materialized reporting schema in the future to simplify reports.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+Proposed
+
+## Consequences
+
+* We will have more nightly cron tasks to refresh the reporting views. We need to monitor the performance of these
+  refreshes closely to identify when database views are no longer a viable option for us.
+* We can start thinking of migrating existing dashboard reports to consume data from the reporting tables. Some good
+  candidates to start migrating could possibly be:
+  * Registrations: It is a simple report and easy to verify
+  * Lost-to-follow-up breakdowns: It tests the more complex aspects of the reporting tables, but is fairly isolated.
```
ssrihari, line 149: I feel (not strongly) like this should be an _intended_ consequence, and should move into the decision section. But it's cool if we want to try it out first, and then commit to it ✌️

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+## Alternatives and constraints
+
+Apart from the proposed approach of reporting tables as database views, there are some alternatives that were
+considered.
+
+### Reporting tables in a separate database
+
+We could continue to generate the reporting tables by hand, but place them in a separate database. This would greatly
+alleviate performance problems. However, database performance is a secondary concern. Furthermore, the proposed approach
+will still set us up nicely to extract the reporting tables to a separate database in the future if necessary.
```
ssrihari, line 124: Also, we would have to build and maintain actual ETL pipelines for this approach.

doc/arch/014-reporting-etl-pipeline.md
ssrihari: Glad to see [2018 decisions](https://github.com/simpledotorg/simple-server/blob/master/doc/arch/004-reporting.md#metabase) standing 😁 . It does seem like we volleyed back to it though.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+The "prescription drugs over time" table will have records like:
+* As of Jan 2020, what drugs were prescribed to Ashish?
+* As of Feb 2020, what drugs were prescribed to Ashish?
+* As of Mar 2020, what drugs were prescribed to Ashish?
+* ...and so on
+
+Since these tables are pre-computing questions like "What was the most recent BP as of X month", the reports queries
+become much simpler. Instead of dealing with this question, the reports query can instead fetch all records for a
+particular month, and the "most recent" information will be readily available in the record.
+
```
ssrihari, line 79: I think it might be useful to link to a question being answered via the reporting schema here. Perhaps to the other open PR, and a metabase question?

ssrihari (commented):
This looks great, overall! 🙌  Mostly added some comments on clarifying things, and adding some more details.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+application, and vice versa.
+
+In addition to complexity, performance is also a growing concern as we scale. Heavy reporting operations may strain the
+application's ability to write data to the database. Conversely, heavy application activity may impact our ability to
+generate reports in a performant manner. We currently use heavy layers of caching to mitigate these performance
+concerns, but the less caching we _need_ to do, the better.
+
+### Priorities and constraints
+
+The upcoming changes to our reporting pipeline are quite new to us, so the priority is the ability to iterate.
```
rsanheim, line 32: The more I think about the "new" ETL we are talking about, the more I realize it isn't really new...its mostly a refined, better approach to the materialized views we already have in the app and have been using for a long time now. So in some ways, its already an iteration on a previous approach that has worked out well. That may be worth mentioning here...we aren't switching to some brand new ETL architecture in a separate database, we are instead working towards reporting views in our existing DB that are modeled better and more directly useful for the reports we need today.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,90 @@
...
+Performance is a secondary concern, since we are not yet at a scale where we need to worry about it too much, and our
+caching strategy works well today.
+
+## Decision
+
+We will introduce a set or reporting tables, implemented as database views, to start powering our reporting
+requirements. This reporting schema will transform the data in our application tables into a more reports-friendly
+format, exposing convenient building blocks to build reports with straightforward SQL queries.
+
+Once the reporting tables are populated, we will use Metabase to create and maintain the SQL queries that run against
```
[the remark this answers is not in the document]
rsanheim: It isn't clear to me how Metabase plays into things here...my understanding is that Metabase is useful for one-off reports or as a 'staging ground' to iterate on complex SQL queries before they make there way into the app, whether that's in Query objects or views.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+* As of Mar 2020, what drugs were prescribed to Ashish?
+* ...and so on
+
+Since these tables are pre-computing questions like "What was the most recent BP as of X month", the reports queries
+become much simpler. Instead of dealing with this question, the reports query can instead fetch all records for a
+particular month, and the "most recent" information will be readily available in the record.
+
+### Calendar Table
+
+Time, down to the month, is an important dimension of all reporting tables. As such, a calendar table will be created to
```
rsanheim, line 82: Would be good mention that we are explicitly **not** handing per week or per day queries here...that is a good constraint to call out.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+occurrence in data warehouse design ([Here's one such example](https://www.mssqltips.com/sqlservertip/4054/creating-a-date-dimension-or-calendar-table-in-sql-server/)).
+
+The important elements of our calendar will be:
+* Cardinal month (1-12)
+* Year
+* Quarter: Quarterly analysis is common, so precomputing will be beneficial.
+* A unique date, like first of the month: This helps uniquely identify or group by each record in the calendar table on
+  a single column. `GROUP BY (year, month)` is more awkward.
+
+The calendar table will usually be joined with Simple's application tables to populate the denormalized reporting
```
rsanheim, line 93: Will the calendar table assume that queries using it specify their time zone...I'm assuming UTC? Maybe a good thing to call out here.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+There's a high up-front cost and complexity to using a third-party tool. Additionally, there may be hidden constraints
+in such tools that we don't want to deal with while we're in an iterative stage of this effort.
+
+## Status
+
+Proposed
+
+## Consequences
+
+* We will have more nightly cron tasks to refresh the reporting views. We need to monitor the performance of these
```
rsanheim, line 144: One more consequence to list here is that there will be a temptation to add business logic to the reporting views that often would be better located in the application itself.  For example, we have encoded the idea of `risk_level` into some mat views before as [hardcoded SQL](https://github.com/simpledotorg/simple-server/blob/13299d2bc34312f49eff15249864a8a2615c78c3/db/views/patient_summaries_v01.sql#L35-L48) that generates the views. You can imagine if this were something that stakeholders and product wanted to iterate on, it would be more difficult to change and validate as raw SQL as compared to ruby logic.

One other example that has come up is answering the question of "what protocol step should a patient be on?" ...I can imagine a 20 line SQL query that can answer that question...but I don't think I'd want to write it or maintain it, at least not to start. 😁  That is the sort of 'core business logic' that I think belongs squarely in the ruby domain, at least for awhile while we iron out the rough edges and make sure it matches the health care folks want.

There is a tradeoff between encoding things into SQL as compared to Ruby ...Ruby is easier to change and iterate on, whereas SQL can consolidated and reduce duplication, but is often harder to change and test. Something worth considering and pointing out here.

doc/arch/014-reporting-etl-pipeline.md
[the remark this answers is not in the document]
rsanheim: I think the idea behind the "load" part of ETL is that there is something concrete and already "loaded" that future callers can query from...hence the materialized part. We are basically using materialized views as a shortcut to bypass some of the overhead of a real ETL pipeline.

Creating views that just load data from our actual application database every time doesn't really feel like an ETL at all, so I think stating "we are using materialized views" would be totally fair here.

doc/arch/014-reporting-etl-pipeline.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Reporting tables in a separate database
+
+We could continue to generate the reporting tables by hand, but place them in a separate database. This would greatly
+alleviate performance problems. However, database performance is a secondary concern. Furthermore, the proposed approach
+will still set us up nicely to extract the reporting tables to a separate database in the future if necessary.
+
+### Rollup summary tables
+
+Rollup summary tables were proposed in [an earlier ADR](011-rollup-summary-tables-for-dashboards.md). The crux of the
```
rsanheim, line 128: I think marking that ADR as rejected would be fine here, or in whatever PR marks this as accepted.

rsanheim (approved):
I approve this ADR with some feedbacks! 😸

doc/arch/014-reporting-etl-pipeline.md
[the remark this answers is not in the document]
harimohanraj89: Oops, I forgot to clean up the copy here. Yes, you are right - I will make it clear throughout the ADR that we _can_ start using Metabase more easily to _prototype_ new reports.

kpethtel (approved):

harimohanraj89:
Closing this until we're ready to resume ETL work.

PR #2389, opened 2021-04-07
2938 experiment controls 3

kpethtel:
**Story card:** [ch2938](https://app.clubhouse.io/simpledotorg/story/2938/start-active-patient-experiment)

## Because

We need the ability to start experiments. 

## Note

This is foundational work for the experiment project and is not likely to be the final version of the code. The code in this PR should have no impact on the app at present.


rsanheim (approved):
As discussed in slack, this looks good to merge to serve as a basis for us dividing up work into smaller stories 👍 .

Also, this has other handy pieces we want, like stricter validations and the seed data to make it easier to start playing with experiments in dev.

PR #2390, opened 2021-04-07
Rename experiment types to 'current_patients' and 'stale_patients'

rsanheim:
"active" and "stale" is a bit more intuitive for these, and also we can
leave off the "reminder" part because we know these experiments are
_only_ about AB testing reminders to get patients back to care.

No story for this, this is a name we talked about quite a bit during pairing and I wanted to propose this before moving forward on other things.

### DIFF
--- a/app/models/experimentation/experiment.rb
+++ b/app/models/experimentation/experiment.rb
@@ -7,14 +7,14 @@ module Experimentation
     validates :state, presence: true
     validates :experiment_type, presence: true
     validate :date_range, if: proc { |experiment| experiment.start_date_changed? || experiment.end_date_changed? }
-    validate :one_live_experiment_per_type
+    validate :one_active_experiment_per_type
 
     enum state: {
       new: "new",
       selecting: "selecting",
-      live: "live",
+      running: "running",
       complete: "complete"
-    }, _prefix: true
+    }, _suffix: true
 
     enum experiment_type: {
       active_patients: "active_patients",
@@ -28,11 +28,11 @@ module Experimentation
 
     private
 
-    def one_live_experiment_per_type
-      existing = self.class.where(state: ["live", "selecting"], experiment_type: experiment_type)
+    def one_active_experiment_per_type
+      existing = self.class.where(state: ["running", "selecting"], experiment_type: experiment_type)
       existing = existing.where("id != ?", id) if persisted?
       if existing.any?
-        errors.add(:state, "you cannot have multiple selecting or live experiments of type #{experiment_type}")
+        errors.add(:state, "you cannot have multiple active experiments of type #{experiment_type}")
       end
     end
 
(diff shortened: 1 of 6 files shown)


app/models/experimentation/experiment.rb
```diff
@@ -15,9 +15,10 @@ class Experiment < ActiveRecord::Base
       live: "live",
       complete: "complete"
     }, _prefix: true
+
     enum experiment_type: {
-      current_patient_reminder: "current_patient_reminder",
-      inactive_patient_reminder: "inactive_patient_reminder"
+      active_patients: "active_patients",
+      stale_patients: "stale_patients"
```
kpethtel, line 21: I may be misremembering, but I had previously thought you didn't like the name stale and wanted "inactive" so I changed "stale_patient_reminder" to "inactive_patient_reminder". I initially went with "current" and "stale" because the experiment had an "active" state and I wanted to avoid overloading of word.

kpethtel (approved):

app/models/experimentation/experiment.rb
rsanheim: Good point, lemme push up a refinement on this.

app/models/experimentation/experiment.rb
```diff
@@ -7,17 +7,18 @@ class Experiment < ActiveRecord::Base
...
     validates :experiment_type, presence: true
     validate :date_range, if: proc { |experiment| experiment.start_date_changed? || experiment.end_date_changed? }
-    validate :one_live_experiment_per_type
+    validate :one_active_experiment_per_type
 
     enum state: {
       new: "new",
       selecting: "selecting",
-      live: "live",
+      running: "running",
```
rsanheim, line 15: This feels better to me than "live" ...

app/models/experimentation/experiment.rb
```diff
@@ -27,11 +28,11 @@ def group_for(uuid)
 
     private
 
-    def one_live_experiment_per_type
-      existing = self.class.where(state: ["live", "selecting"], experiment_type: experiment_type)
+    def one_active_experiment_per_type
```
rsanheim, line 31: Now an active experiment is one that is selecting or running.

app/models/experimentation/experiment.rb
```diff
@@ -7,17 +7,18 @@ class Experiment < ActiveRecord::Base
...
+    validate :one_active_experiment_per_type
 
     enum state: {
       new: "new",
       selecting: "selecting",
-      live: "live",
+      running: "running",
       complete: "complete"
-    }, _prefix: true
+    }, _suffix: true
```
rsanheim, line 17: Using the suffix form here reads better:
```
experiment.running_state?
experiment.selecting_state?
```

rsanheim (commented):
notes

app/models/experimentation/experiment.rb
kpethtel: Yeah, that does probably read better. We use prefix on all other enums throughout the app. Maybe it reads better for those uses.

kpethtel (approved):

PR #2407, opened 2021-04-08
Rename to ControlRateCohortQuery; speed up and clean up misc tests

rsanheim:
This matches the fact this particular query is doing a cohort based
control rate selection, which is different from the standard
ControlRateQuery.

This helps make our queries a little more clear as to what is what.

**Story card:** [ch3164](https://app.clubhouse.io/simpledotorg/story/3164/rename-cohort-based-bp-query)


claudiovallejo (approved):

spec/tasks/scripts/delete_organization_data_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "rails_helper"
 require "tasks/scripts/delete_organization_data"
 
-RSpec.describe DeleteOrganizationData do
+RSpec.describe DeleteOrganizationData, :skip do
```
kpethtel, line 4: Why are we skipping this?

spec/models/facility_spec.rb
```diff
@@ -53,20 +53,15 @@
...
-          create(:patient,
-            :without_hypertension,
-            assigned_facility: assigned_facility,
-            registration_facility: registration_facility)]
+      let(:assigned_facility) { create(:facility) }
+      let(:registration_facility) { create(:facility) }
+      before do
+        @assigned_patients = [create(:patient, assigned_facility: assigned_facility, registration_facility: registration_facility),
+          create(:patient, :without_hypertension, assigned_facility: assigned_facility, registration_facility: registration_facility)]
       end
```
prabhanshuguptagit, line 61: I feel that creating an instance var in a `before` is just another way of writing a `let!`. Moreover, instance variables are hard to follow for. I've avoided them: 
- Instance variables don't throw a `NameError` when they're not declared, they just silently evaluate to `nil`
- It's possible to share state across specs in [funky ways](https://tomdebruijn.com/posts/ruby-rspec-instance-variables/#modifying-instance-variable-values).

Unless there is a strong reason to move to using instance vars, we should stick with plain old `let!`s. I think when we scrap a `let!` away, the only places that code should move to are:
- inside the `it` block
- into a method that is called in the `it` block
- into a factory method that is called in the `it` block

imo these other Rspec helpers share the same problem that `let!` has and are worse in their own ways.

prabhanshuguptagit (commented):
Left some comments on the approach to rewording specs. Mainly:
- Global shared state
- `let` vs `let!`
- Instance vars vs `let!`

These come up as nit's every now and then but I think we should discuss and conclude on these :D

spec/models/materialized_patient_summary_spec.rb
```diff
@@ -1,279 +1,265 @@
...
+    _new_phone = create(:patient_phone_number, patient: patient, device_created_at: new_date)
+    _old_passport = create(:patient_business_identifier, patient: patient, device_created_at: old_date)
+    @new_bp = create(:blood_pressure, patient: patient, facility: facility, recorded_at: new_date, user: user, systolic: 110, diastolic: 70)
+    @blood_sugar = create(:blood_sugar, patient: patient, facility: facility, recorded_at: new_date, user: user,
+                                        blood_sugar_type: "random", blood_sugar_value: 100)
+    patient
+  end
+
+  attr_reader :new_bp
+  attr_reader :blood_sugar
```
prabhanshuguptagit, line 28: I guess this is a natural outcome once we use instance vars to avoid mutating them. This reads like we're adding behaviour to a test though. I haven't seen this pattern before but it doesn't feel idiomatic. wdyt?

app/queries/control_rate_cohort_query.rb
```diff
@@ -1,6 +1,7 @@
-# frozen_string_literal: true
-
-class BloodPressureControlQuery
+# A "Cohort" for this query is defined depending on the period:
+#   Month: For patients registered in a month, find BPs taken for them in the following two months
+# . Quarter: For patients registered in a quarter, find BPs tken for them in the following quarter
+class ControlRateCohortQuery
```
prabhanshuguptagit, line 4: [nit] `CohortControlRateQuery`?

spec/models/materialized_patient_summary_spec.rb
kpethtel: I do understand Rob's desire to avoid `let!` because they do tend to bloat tests over time as different developers add to large test files, but I also have always disliked using instance variables in rspec. You can find articles with good reasons to avoid it. In this particular case, I can see why Rob is doing it. The previous implementation had several lets that were not directly used, which does make it more difficult to understand what's really going on and what's important. So I can support creating those within the patient definition. But this does feel very non-idiomatic. Personally, I would prefer to keep `new_bp` and `blood_sugar` in lets and add them to the patient in a before block.

spec/models/address_spec.rb
```diff
@@ -10,8 +10,8 @@
   end
 
   describe ".has_street_address?" do
-    let!(:address_without_street_address) { create(:address, :no_street_address) }
-    let!(:address_with_street_address) { create(:address) }
+    let(:address_without_street_address) { create(:address, :no_street_address) }
+    let(:address_with_street_address) { create(:address) }
```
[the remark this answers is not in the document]
rsanheim: > Do we have a preference between `let` and `let!`? 

We do, its avoiding `let` or `let!` as much as possible, with `let!` being the worse of the two. 😁 Both of them have very real problems, with `let!`  being more severe. I've been trying to work to reduce our reliance on both as I can.

I added the [lets_not rachet](https://github.com/simpledotorg/simple-server/blob/5a21196f7db598389b12e539d27ca63761c163f3/spec/rachets/lets_not_spec.rb) as a first step to start to improve our test suite around this. That spec has links to Thoughtbot blog posts around why let is harmful. Their posts are about both `let` and `let!`, but I figured it would be good start with the biggest offender I've seen, which is the bang version.

I realize that there are some valid use cases for `let` with the amount of test data we need to setup. With `let!` though, you end up with magical creation at a distance for every spec, with dependency chains that are really difficult to detangle. The bang version also results in incurring excessive fixture overhead for every spec, which is important for a test suite as large as ours. Consider this contrived example, which is not too far from many specs we have in our suite:

```ruby
let!(:organization, name: "my-org")
let!(:facility, :facility_group: facility_group)
let!(:facility_group, organization: organization, user: admin)
let!(:patient_1, registration_facility: facility)
let!(:blood_pressure, :controlled, patient: patient_1)
let!(:admin, :admin)

context "patient BP control status" do
  before do
    facility.do_some_other_setup_thing
  end
  
  it "is controlled if recent BP is controlled" do
    expect(patient_1).to be_controlled
  end

  it "has no uncontrolled BPs" do
    expect(patient_1.blood_pressures.uncontrolled).to be_empty
 end

  # Imagine some other developer adds the two specs below after seeing that the test
  # fixtures above didn't really work, but also not wanting to refactor everything for their 
  # particular test.  This happens commonly with tests when their are many lines of
  # `let` / `let!` setup.
  it "is uncontrolled if recent BP is uncontrolled" do
    patient = create(:patient)
    blood_pressure = create(:blood_pressure, :hypertensive, patient: patient)
    expect(patient).to be_uncontrolled
  end

  it "has no BPs if patient has not had a BP recorded" do
    patient = create(:patient)
    expect(patient.blood_pressures).to be_empty
  end
```

Note the last two specs are creating their own fixtures that are completely separate from the `let!` created fixtures. Yes, they could be in their own context block, or they could be extracted to their own `lets`, but the core issues still exist -- the data under test is too far from the test code itself, and we end up creating we more test data then we need. These problems become even worse when you consider the case of nested contexts and nested `before` blocks.
In a real world test you may have to scroll 100 lines between the rows of setup and the particular spec...that doesn't make for a maintainable or useful spec long-term.

Here is an alternative with many less `lets`, with a single judicious use of a `let` that is not central to the test:

```
  let(:user) { create(:user, name: "some-test-user) }

  it "is controlled if recent BP is controlled" do
    patient = create(:patient, user: user)
    bp = create(:blood_pressure, :controlled, user: user)
     expect(patient).to be_controlled
  end

  it "has no uncontrolled BPs" do
    patient = create(:patient, user: user)
    bp = create(:blood_pressure, :controlled, user: user)
    expect(patient.blood_pressures.uncontrolled).to be_empty
 end

  it "is uncontrolled if recent BP is uncontrolled" do
    patient = create(:patient, user: user)
    bp = create(:blood_pressure, :controlled, user: user)
    expect(patient).to be_uncontrolled
  end

  it "has no BPs if patient has not had a BP recorded" do
    patient = create(:patient, user: user)
    expect(patient.blood_pressures).to be_empty
  end
```

Here the test setup is right by the assertion, so every spec is easy to read and follow. Data that is not core to the test is in a lazy `let` and used to reduce excessive object creation. its easy to add new specs with particular test data that is needed, and you don't need to read a ton of setup to try to decide if you should build on that or do your own thing.

I realize this is an overly simplified example, but we can use plain old ruby when we need more complex setup. Or we can look into why fixture data is so cumbersome, and try to tackle that problem. The blogposts referenced above go into detail about better approaches.

I hope that is helpful, lemme know if you wanna discuss in more detail on zoom or slack.

spec/models/facility_spec.rb
rsanheim: For context here, I refactored from a mess of `let!` to a slightly less messy solution which is at least more clear to follow involving the `before` created state. Some more details below:

> I feel that creating an instance var in a `before` is just another way of writing a `let!`. Moreover, instance variables are harder to follow:

> * Instance variables don't throw a `NameError` when they're not declared, they just silently evaluate to `nil`

Agreed, that is not good. 

> * It's possible to share state across specs in [funky ways](https://tomdebruijn.com/posts/ruby-rspec-instance-variables/#modifying-instance-variable-values).

That post is talking about `before(:context)`, which is not the same as a plain `before`.  `Before(:context)` is setup once per context and shared with all specs inside the context, whereas standard `before` is reset and run for every spec. Same as any regular test state. Note that a `before` with no args is the same as `before(:example)` - see docs here: https://github.com/rspec/rspec-core/blob/c6315d6e899796d3d0203dc8b656708a3ebca9a1/lib/rspec/core/hooks.rb#L14-L15. The different args to `before` and the default arg at play can make this confusing 😐.

I agree that other `before` setup _besides_ `example` should be used very minimally, I've only added them for some [very isolated cases](https://github.com/simpledotorg/simple-server/blob/5a21196f7db598389b12e539d27ca63761c163f3/spec/rails_helper.rb#L31-L34).

> Unless there is a strong reason to move to using instance vars, we should stick with plain old `let!`s. I think when we scrap a `let!`, the only places that code should move to are:

> * inside the `it` block
> * into a method that is called in the `it` block

I agree, both of the above options are often preferable to me. They take more time to refactor to though when you have a lot specs using similar test setup (like this spec did).

> * into a factory method that is called in the `it` block

I'd like to avoid more logic in our factories if possible, unless its really something that is widely shared across our test suite.

> imo these other Rspec helpers share the same problem that `let!` has and are worse in their own ways.

I agree that ivars and `attr_reader` in tests can look weird. For context, I was working to remove our reliance on `let!` as well as to improve our test speed. I agree that there could be more done here to refactor test setup to plain locals or regular ruby methods to make things more clear.  For this specific case, maybe a regular ruby method that creates the assigned patients would be more clear?

rsanheim (commented):
@prabhanshuguptagit thanks for the feedback! I left some some response ya. Lemme know if you'd like to discuss more on zoom or wherever.

spec/models/address_spec.rb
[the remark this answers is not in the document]
prabhanshuguptagit: That makes sense, and I'm onboard with "lets not" (`let` and `let!` both). I think as we reduce the ratchet number though, moving from `let!` to `let` does not offer a significant advantage. I don't have a strong opinion between between the two, except for the few places where a `let` can be confusing that I've mentioned.

Thanks for the detailed examples, super helpful :D

spec/models/facility_spec.rb
prabhanshuguptagit: Yeah I've relied on regular ruby methods since we started moving away from `let`. 

This looks good for a cleanup. I guess what I don't follow here is the `@assigned_patients` in `before`. It could also be a `let` like the other ones, no?

PS It'd be great if we can also follow something like "No instance vars in our tests" 😛  . wdyt?

rsanheim: > This looks good for a cleanup. I guess what I don't follow here is the `@assigned_patients` in `before`. It could also be a `let` like the other ones, no?

Yeah I agree the ivar also looks weird, it is a tradeoff between more deterministic setup and the ease of 'cleanliness' of a let block.  I could go either way on that.

PR #2411, opened 2021-04-09
2940 send appointment reminders

kpethtel:
**Story card:** [ch2940](https://app.clubhouse.io/simpledotorg/story/2940/create-process-for-sending-an-appointment-reminder?vc_group_by=day)

## Because

We need to be able to send daily appointment reminders.


app/jobs/appointment_reminders/send_reminder_job.rb
```diff
@@ -0,0 +1,73 @@
+class AppointmentReminders::SendReminderJob
+  include Rails.application.routes.url_helpers
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :high
+
+  # i think this should be based on country instead
+  DEFAULT_LOCALE = :en
```
kpethtel, line 8: I may be wrong to think this should be based on country. I think defaulting to English is in line with how the rest of the app works. My guess is that most patients have an address, so their languages will be properly set.

app/jobs/appointment_reminders/send_reminder_job.rb
```diff
@@ -0,0 +1,73 @@
...
+  def perform(reminder_id)
+    reminder = AppointmentReminder.includes(:appointment, :patient).find(reminder_id)
+    send_message(reminder)
+  end
+
+  private
+
+  def send_message(reminder)
+    notification_service = NotificationService.new
+    # i think this logic should be handled by the notification service
```
kpethtel, line 19: I'm planning to move the error handling into the NotificationService. I've looked at how it's used around the app and it seems safe to move error reporting to the class that would be generating the error.

app/jobs/appointment_reminders/send_reminder_job.rb
```diff
@@ -0,0 +1,73 @@
...
+    patient.address&.locale || DEFAULT_LOCALE
+  end
+
+  # this should only be india for now
+  def communication_type
+    type = CountryConfig.current[:name] == "India" ? :missed_visit_whatsapp_reminder : :missed_visit_sms_reminder
+    Communication.communication_types[type]
+  end
+
+  # perhaps abort if not found, but that shouldn't happen
```
kpethtel, line 62: Patients are selected for having mobile phone numbers, but you never know when a patient record might be changed. It's a very edge case.

app/jobs/appointment_reminders/send_daily_reminders_job.rb
```diff
@@ -0,0 +1,15 @@
+class AppointmentReminders::SendDailyRemindersJob < ApplicationJob
+  queue_as :high
+
+  def perform
+    # add feature flag?
```
kpethtel, line 5: While there shouldn't be any reminders to send in other countries, I think it probably makes sense to either feature flag this or guard based on country.

app/jobs/appointment_reminders/send_daily_reminders_job.rb
```diff
@@ -0,0 +1,15 @@
...
+    # given that the cron runs in the middle of the night, i would imagine this will always
+    # result in messages being sent the same day, but we will need to be certain of that
+    # either by carefully scheduling the cron or by moving away from next_messaging_time
```
kpethtel, line 8: We'll just want to ensure we send these same day.

kpethtel:
Realized I want to add a few more things to this before getting feedback.

app/jobs/appointment_reminders/send_daily_reminders_job.rb
```diff
@@ -0,0 +1,16 @@
+class AppointmentReminders::SendDailyRemindersJob < ApplicationJob
+  queue_as :high
+
+  def perform
+    # add feature flag?
```
kpethtel, line 5: We can start a feature flag now or wait until we add the cron for this.

app/jobs/appointment_reminders/send_reminder_job.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+
+      Communication.create_with_twilio_details!(
+        appointment: reminder.appointment,
+        appointment_reminder: reminder,
+        twilio_sid: response.sid,
+        twilio_msg_status: response.status,
+        communication_type: communication_type
+      )
+    rescue Twilio::REST::TwilioError => e
```
kpethtel, line 43: I would really prefer to move this error handling to the NotificationService, but I'm leaving it alone for now.

app/jobs/appointment_reminders/send_reminder_job.rb
```diff
@@ -0,0 +1,98 @@
...
+    I18n.t(
+      reminder.message,
+      assigned_facility_name: reminder.appointment.facility.name,
+      patient_name: reminder.patient.full_name,
+      appointment_date: reminder.appointment.scheduled_date,
+      locale: patient_locale(reminder.patient)
+    )
+  end
+
+  # i'm not convinced this is good enough
```
kpethtel, line 58: The reason I'm not convinced this is good enough is that this is the logic in address:
```
  STATE_TO_LOCALE = {
    punjab: "pa-Guru-IN",
    maharashtra: "mr-IN",
    karnataka: "kn-IN"
  }

  def locale
    STATE_TO_LOCALE.fetch(state_to_sym, :en)
  end
```
That doesn't seem to support all options.

db/migrate/20210412172435_add_appointment_reminder_to_communication.rb
```diff
@@ -0,0 +1,5 @@
+class AddAppointmentReminderToCommunication < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :communications, :appointment_reminder, type: :uuid, foreign_key: true
```
kpethtel, line 3: Theoretically this should be null false but that would invalidate all existing communications.

app/controllers/api/v3/twilio_sms_delivery_controller.rb
```diff
@@ -6,11 +6,21 @@ def create
...
 
-    if communication_type == "missed_visit_whatsapp_reminder" && twilio_message.unsuccessful?
-      AppointmentNotification::Worker.perform_at(Communication.next_messaging_time, appointment_id, "missed_visit_sms_reminder")
+    if twilio_message.unsuccessful?
+      communication = twilio_message.communication
+      communication_type = communication.communication_type
+      appointment_id = communication.appointment_id
+      appointment_reminder_id = communication.appointment_reminder_id
+
+      if communication.appointment_reminder_id && communication.appointment_reminder.next_communication_type
```
rsanheim, line 16: This kind of branching logic would not be necessary if we migrating the existing `AppointmentNotification::Worker` code to use AppointmentReminders. It will consolidate and simplify how we handle reminders in the app, which I think would be a win.

app/jobs/appointment_reminders/send_reminder_job.rb
```diff
@@ -0,0 +1,103 @@
+class AppointmentReminders::SendReminderJob
+  include Rails.application.routes.url_helpers
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :high
+
+  def perform(reminder_id)
```
rsanheim, line 7: From scanning this job it duplicates lots of logic from the existing notification worker.  Another reason to migrate the existing code first.

app/models/communication.rb
```diff
@@ -3,6 +3,7 @@ class Communication < ApplicationRecord
   include Hashable
 
   belongs_to :appointment
+  belongs_to :appointment_reminder, optional: true
```
rsanheim, line 6: One more time =) ...if we migrate the existing system to use appointment_reminder, this can at least be enforced as required going forward, and I think we could actually remove the direct relationship to appointment (as appointment reminder gives us that instead.

PR #2412, opened 2021-04-09
Extract / refactor some parts of patient selection

rsanheim:
This is a refactoring to pave the way for **Story card:** [ch3155](https://app.clubhouse.io/simpledotorg/story/3155/experiments-should-send-reminders-to-inactive-patients)

PR #2413, opened 2021-04-09
Select patients with old visits for the "stale" experiment

rsanheim:
**Story card:** [ch3155](https://app.clubhouse.io/simpledotorg/story/3155/experiments-should-send-reminders-to-stale-patients)
**Related PRs** follow up to https://github.com/simpledotorg/simple-server/pull/2412

## Because

An inactive patient is one whose last visit was in between 35 and 365 days ago.

A visit in is one of an appointment created, a BP taken, a blood sugar taken, or a prescription drug was refilled. We can model this query off of the existing NoBpMeasureQuery, which captures visits in a particular time period without a BP taken. We can instead include BPs taken into that query to capture a visit.

We want these patients to be bucketed into experiment groups to try different reminders to get them back to care. These reminders are not tied at all to an Appointment.

We will want the created reminders for these patients to be cancelled if the patient does come back for care...this is handled by story 2942 .

We want to rate limit how many reminders are sent per day, so that we don't overwhelm clinics with thousands and thousands of appointments per day. We may want to bucket patients every day, so that we can capture that timestamp to output for the followup analysis.

## This addresses

Adding the new query, some new tests, refactoring.


app/models/experimentation/experiment.rb
```diff
@@ -22,7 +22,7 @@ class Experiment < ActiveRecord::Base
     }, _prefix: true
 
     def self.candidate_patients
-      Patient.from(Patient.with_hypertension, :patients)
+      Patient.with_hypertension
```
rsanheim, line 25: I think this was leftover from an earlier version of this query that was hacking around default scope things?

app/services/experimentation/stale_patient_selection.rb
```diff
@@ -1,27 +1,58 @@
...
+      return [] if candidate_ids.empty?
+
+      parameters = {
+        candidate_ids: candidate_ids,
+        hypertension: "yes",
+        range_start: eligible_range.begin,
+        range_end: eligible_range.end,
+        start_date: start_date
+      }
+      sql = GitHub::SQL.new(<<~SQL, parameters)
```
rsanheim, line 31: I'm using SQL here to make sure we have the logic correct first.  It is easier to do these types of queries in plain SQL compared to fighting with AR complexity for things like sub selects and such.  We can always refactor to AR later if we want.

Also, we may need raw SQL if we use the table sampling method we had discussed to grab random ids.

spec/services/experiment_control_service_spec.rb
```diff
@@ -361,13 +283,13 @@
     it "only schedules reminders for PATIENTS_PER_DAY * number of days" do
       stub_const("ExperimentControlService::PATIENTS_PER_DAY", 1)
       patient1 = create(:patient, age: 80)
-      create(:encounter, patient: patient1, device_created_at: 100.days.ago)
+      create(:blood_pressure, patient: patient1, device_created_at: 100.days.ago)
       patient2 = create(:patient, age: 80)
-      create(:encounter, patient: patient2, device_created_at: 100.days.ago)
+      create(:blood_pressure, patient: patient2, device_created_at: 100.days.ago)
```
rsanheim, line 288: All these test fixtures now rely on visits, which means an actual BP recorded or drug or RX, etc.  So we can't rely on creating encounters to setup our test data anymore.

rsanheim (commented):
Notes for reviewers

app/models/experimentation/experiment.rb
kpethtel: It may be related to the phone number default scope. I recall the problem was a "distinct" that didn't play well with the rest of the query.

app/services/experiment_control_service.rb
```diff
@@ -43,29 +41,26 @@ def start_current_patient_experiment(name, days_til_start, days_til_end, percent
...
+      eligible_ids = Experimentation::StalePatientSelection.call(start_date: start_date)
       eligible_ids.shuffle!
 
+      schedule_date = start_date
       total_days.times do
-        daily_ids = eligible_ids.pop(PATIENTS_PER_DAY)
+        daily_ids = eligible_ids.pop(patients_per_day)
         break if daily_ids.empty?
-        schedule_date = start_date
         # TODO: remove references to appointment after removing appointment dependency
```
kpethtel, line 58: I think we can remove this comment now. I wrote that when we were considering making appointment optional for reminders.

kpethtel (approved):
Looks good.

PR #2414, opened 2021-04-12
 ADR: Online patient lookup

ssrihari:


doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
...
+< @mobile-devs: please add details on possible approaches here >
+
+Note that the retention time period is not strict. i.e,
+it is okay for the record to stay on the phone for 2h if the retention
+period is 1h. We can perform deletions at a convenient time as long as
+the deletion happens deterministically.
+
+_Alternatively_: we can choose to hard-code the retention period on
+the app.
+
```
ssrihari, line 51: @msasikanth @vinaysshenoy need your help to hash this section out! 🙇

msasikanth: To begin with, I think we can go with a hard-coded (or remote config option) retention period on the app. That way we can remove the patient once the retention period is finished during our regular sync. A similar logic should apply for the retention period sent from the server, we remove the user from the local DB once the retention period is finished during our regular sync interval. I think that interval is deterministic, WDYT? 

Also, we won't remove the patient if they click on the sync indicator to avoid any confusion for them, since patients will be removed from the recent patient's list when they manually sync.

ssrihari: > or remote config option

If it's not very different in terms of implementation, I'd prefer keeping this config on the API for reasons mentioned in the ADR.

> once the retention period is finished during our regular sync interval. I think that interval is deterministic, WDYT?

Yup, this works quite well. What is the current sync interval?

> since patients will be removed from the recent patient's list when they manually sync.

As long as we continue to keep the record until the retention period is completed, we should be good. I feel it might be simpler to just apply this rule of `max(sync-time, retention-period-expires-time)` irrespective of how the sync is initiated. But, your call ✌️ .

msasikanth: > What is the current sync interval?

Currently it's 15 minutes

harimohanraj89: I wouldn't worry too much about the recent patients list. Unless you're in the last 3 patients that a nurse saw, the nurse isn't going to notice a disappearing patient.

If we're really worried about folks noticing it, or relying on the recent patient list to find older patients, then we should adjust our retention period accordingly - eg. make it 1 day instead of 1 hour.

doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
...
+}
+````
+
+_Alternatively_: we can try to repurpose the existing audit logs for
+this, while capturing all the information.
+
+### Rate limiting
+We currently rate limit our authentication endpoints using
+[rack-attack](https://github.com/rack/rack-attack). We will do the
+same for this API.
```
harimohanraj89, line 117: Let's add a note here why rate limiting this endpoint is important? That'll inform what rate limits to choose. 

We have some good starting points discussed in the PRD, I'll let you decide whether you want to include the 60rpm figure into the ADR as well.

doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
...
+it is okay for the record to stay on the phone for 2h if the retention
+period is 1h. We can perform deletions at a convenient time as long as
+the deletion happens deterministically.
+
+_Alternatively_: we can choose to hard-code the retention period on
+the app.
+
+### API contract
+For the request, we will use the endpoint `GET`:
+https://api.simple.org/api/v4/patient/identifier/, where `identifier`
```
harimohanraj89, line 54: `patients`, since we're returning a list? I'm not sure we can nail down a strictly "RESTful" URL for this endpoint 😅 

The endpoint itself is a simple change to make while developing, so finding the right URL isn't a blocker for the ADR, but it is a blocker for shipping the endpoint 👍

doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
...
+is any valid patient business identifier.
```
harimohanraj89, line 55: How will the client specify the identifier type? eg. BP passport vs National Health ID. If we have thoughts about it, shall we put it in the ADR, since it's covering the API contract?

doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
...
+with, but can later vary depending on the state, country or other
+factors.
+
+We will implement temporary retention in the app with this
+feature. With every patient retrieved via the lookup API, we will:
+
+- ensure the data is synced to the server
+- delete the data after the retention time period has passed
+
+< @mobile-devs: please add details on possible approaches here >
```
harimohanraj89, line 42: @msasikanth I think it's important to think through and write down how the app will _use_ the retention info in the API as a part of this ADR's review process.

doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
...
+
+In the response, will return a _list of patients_ that have that
+identifer. Note that it is possible for more than a single patient to
+have the same identifier. The response contract will be similar to
+[the API used in the BP Passport
+App](https://api.simple.org/api-docs#tag/Patient/paths/~1patient/get):
+
+````
+{"retention": {
+    "type": "temporary",
```
harimohanraj89, line 65: Is this redundant if a duration is provided below? Or is it included anyway for the client's ergonomics?

doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
...
+  user_id: user.id,
+  facility_id: current_facility.id,
+  identifier: identifier,
+  patient_ids: [patient_id],
+  time: Time.current
+}
+````
+
+_Alternatively_: we can try to repurpose the existing audit logs for
+this, while capturing all the information.
```
harimohanraj89, line 112: Separate audit log makes sense. A nurse looking up a specific patient is conceptually different enough from a nurse fetching their block's patients in bulk.

harimohanraj89 (changes requested):
My thoughts.

doc/arch/015-online-patient-lookup.md
ssrihari: I didn't want to add the specifics here, since it's already a part of the PRD.

The way I was thinking about it was: Rate limiting is a feature requirement, and is rationalised in the PRD. The value will be a config that will also be discussed and decided in the PRD. The ADR provides a rationale only for the implementation approach.

Wdyt? I don't mind x-posting what's in the PRD if that's helpful context, and we needn't be super strict about this.

doc/arch/015-online-patient-lookup.md
ssrihari: Yeah, `patients` makes sense. I specifically put the endpoint and API design here because I wanted feedback on it 😁 . Thanks!

doc/arch/015-online-patient-lookup.md
ssrihari: I intended for the identifier to be of any type (and hence unspecified), since even the android app might not be in a good place to discern it. It's also possible for us to have the same UUID across Simple BP Passports and another ID.

When we return the patients, they'll have the identifier and its type in the payload in case the app needs it.

Good call though, I'll add this to the ADR.

doc/arch/015-online-patient-lookup.md
ssrihari: Yeah, this redundancy is intended to make the API more readable, and usable. It's simpler to understand a "permanent" type rather than a "infinite" retention period. Easier to read for humans, and simpler to write the parser in code. Explicit vs implicit.

doc/arch/015-online-patient-lookup.md
ssrihari: Cool, I'll mark this part rejected.

doc/arch/015-online-patient-lookup.md
ssrihari: @msasikanth Feel free to push a commit onto this PR. If not, you can leave a comment here and I can pull that into the changes too.

doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
+# Online Patient Lookup
+
+## Status
+
+Proposed on 2020-04-12
+
+## Context
+
+Want want to ensure that a patient’s medical records within the state
```
kpethtel, line 9: I'm guessing this should say "we want".

doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Decision
+### Data retention
+There are two broad kinds of data retention on the app at a facility:
+1. Permanent: patient is within sync criteria: the same block, has an
+   appointment, or is assigned to the facility
+2. Temporary: patient is not within the sync criteria
+
+In order to have a better control of the retention, we will send this
+information from the server. This will be a static number to begin
```
kpethtel, line 32: "This will be a static number". I'm not sure what this refers to. What will be a static number? How long the data lives on the device?

doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
...
+In the response, will return a _list of patients_ that have that
+identifer. Note that it is possible for more than a single patient to
+have the same identifier. The response contract will be similar to
+[the API used in the BP Passport
+App](https://api.simple.org/api-docs#tag/Patient/paths/~1patient/get):
+
+````
+{"retention": {
+    "type": "temporary",
+    "duration_seconds": 3600
```
prabhanshuguptagit, line 66: Can we calculate the time at which this should be deleted server-side? This might make it easier for the app to track when to delete the record. Currently the app will need to keep a separate `synced_at` timestamp.

doc/arch/015-online-patient-lookup.md
kpethtel: If I understand this correctly, it sounds like there are at least two types of valid business identifiers: BP passport and National Health ID. And it sounds like the client doesn't know which one it is, so we search for both and return matches for both. So duplicate matches would be very rare, correct? I think it would be good to get these details into the ADR because it would not have been clear what the "business identifiers" are or why there can be multiple matches.

doc/arch/015-online-patient-lookup.md
```diff
@@ -0,0 +1,121 @@
...
+    "business_identifiers": []
+  }]
+}
+````
+
+_Alternatively_: We could use a v5 prefix to disambiguate with the
+Passport app lookup API more clearly.
+
+### Access restrictions
+All data that can be sync'd to facilities in the State can be returned
```
kpethtel, line 93: This is a bit unclear to me. What is "all data that can be synced to the facilities in the state"? What is the limitation? The next sentence seems to imply that the limitation is that data is completely siloed by state, but it's not clear to me.

kpethtel (changes requested):
Good work. I don't have any concerns about the plan, but I left a few questions where I thought the ADR was unclear about details.

doc/arch/015-online-patient-lookup.md
ssrihari: haha, the mind just glosses over such typos 😂  thanks, will fix

doc/arch/015-online-patient-lookup.md
ssrihari: Ah this is in (forward) reference to the retention period sent in the API. I'll make that clearer.

doc/arch/015-online-patient-lookup.md
ssrihari: There can also be multiple matches because of duplicate registrations, which is actually most of what we've actually seen. Collisions across identifier types are just hypothetical atm. I'll call out both of these in the ADR.

doc/arch/015-online-patient-lookup.md
ssrihari: Yeah, a bit of this context comes from the PRD. I'll try to rephrase this to be clearer. Perhaps like so: "Access for this API will be restricted to the state that the user is registered in. Trying to lookup a patient that resides outside the state will return 403".

As I write this out, I realise there's an edge case where a patient has travelled across states, and the identifier refers to 2 patient records in different states. In this case, I think we might return only the one patient belonging to the user's state.

doc/arch/015-online-patient-lookup.md
ssrihari: While this is a good idea, sharing timestamps across a distributed setting (server, android app) is usually not a fool proof concept. Using a relative value (like duration) often does the trick well enough.

And in this case, I think the `synced_at` value might already be stored in the app? @msasikanth

doc/arch/015-online-patient-lookup.md
ssrihari: @msasikanth I've tried to write out what the implementation details should look like. Please have a look, and suggest changes.

doc/arch/015-online-patient-lookup.md
ssrihari: Added a reference to the PRD accordingly.

doc/arch/015-online-patient-lookup.md
ssrihari: Changed this accordingly.

doc/arch/015-online-patient-lookup.md
ssrihari: Added details around the type of identifier not being specified in the API.

doc/arch/015-online-patient-lookup.md
ssrihari: Done.

doc/arch/015-online-patient-lookup.md
ssrihari: Fixed.

doc/arch/015-online-patient-lookup.md
ssrihari: Clarified this and added the forward reference.

doc/arch/015-online-patient-lookup.md
ssrihari: Added the above.

doc/arch/015-online-patient-lookup.md
msasikanth: @ssrihari I will capture the initial mobile implementation details in the ADR tomorrow and for anything else that is missed, will get to it once we started work on it and have more details.

doc/arch/015-online-patient-lookup.md
ssrihari: Spoke to @msasikanth over a call, we will write this section out later.

doc/arch/015-online-patient-lookup.md
ssrihari: Discussed and addressed in 5fc36c6e533b7fec018b832d1b8df28d0e5a8fae.

PR #2418, opened 2021-04-12
Remove DeleteOrganizationData

rsanheim:
This script would not work today as is, because it doesn't take into
Regions or any other new models that have been added since this was run.

We can always revive this from git history if we need to get it back
into shape for use.

(no story, just cleanup)


harimohanraj89 (changes requested):
According to the spec failures, we're still trying to load this thing `lib/tasks/hard_delete.rake`. Let's remove that dangling reference.

You can dismiss my review and merge once done :thum

PR #2419, opened 2021-04-12
Fix Patient#current_age (and flakey spec!)

rsanheim:
We've had a flakey spec failure in the patient_exporter_spec that has
been bugging out for a long time now...see an exmample of the failure
here: https://simple.semaphoreci.com/jobs/393a5bee-61bb-4094-9097-756671673bd3

This was actually a real bug that just happens to pop up randomly based
on the current date and the date patients get from the factory.

Our prevoius current_age calculation did not take into account months
when returning the age based on date_of_birth, so we would get
intermittent off by one errors.  This fixes that, so the spec should be
stable now.


app/models/patient.rb
```diff
@@ -188,7 +188,7 @@ def high_risk?
 
   def current_age
     if date_of_birth.present?
-      Date.current.year - date_of_birth.year
+      ((Time.zone.now - date_of_birth.to_time) / 1.year).floor
```
kpethtel, line 191: I remember looking at the current age method before I even started working at simple and thinking, "That doesn't seem right. You can't know someone's age just be subtracting years."

kpethtel (approved):
Nice catch.

app/models/patient.rb
ssrihari: Intuitively I felt there should be a better solution to get an interval's value in years. But apparently this is the best we can do 😁 . Glad we got this fixed.

PR #2420, opened 2021-04-13
Allow deduplicating by all identifiers types

prabhanshuguptagit:
**Story card:** [ch3105](https://app.clubhouse.io/simpledotorg/story/3105/allow-deduplication-for-all-bp-passport-types-not-just-simple-bp-passport)

## Because

The Deduplication UI and auto merge only dedupe patients for `simple_bp_passports`. We added this clause to explore dedupe only in IHCI initially. It can now be expanded to all identifier types.

## This addresses

Changes the dedupe strategy to pick up all identifier types.


app/services/patient_deduplication/strategies.rb
```diff
@@ -37,7 +37,6 @@ def duplicate_identifiers
           .joins(:patient)
           .select("identifier, array_agg(patient_id) as patient_ids")
           .where.not(identifier: "")
-          .where(identifier_type: simple_bp_passport)
```
harimohanraj89, line 40: If we're dropping this clause, then we should add `identifier_type` to the `group` clause below for accuracy.

harimohanraj89 (approved):
Looks great. Slapping that approval on here 👍 

Do we need test coverage for this change? I'll let you add the necessary tests and :shipit:

PR #2425, opened 2021-04-15
3189 convert to appointment reminder

kpethtel:
**Story card:** [chXXXX](URL)

## Because

## This addresses

## Test instructions

### DIFF
--- a/app/controllers/api/v3/twilio_sms_delivery_controller.rb
+++ b/app/controllers/api/v3/twilio_sms_delivery_controller.rb
@@ -7,10 +7,11 @@ class Api::V3::TwilioSmsDeliveryController < ApplicationController
 
     twilio_message.update(update_params)
     communication_type = twilio_message.communication.communication_type
-    appointment_id = twilio_message.communication.appointment_id
 
     if communication_type == "missed_visit_whatsapp_reminder" && twilio_message.unsuccessful?
-      AppointmentNotification::Worker.perform_at(Communication.next_messaging_time, appointment_id, "missed_visit_sms_reminder")
+      appointment_reminder_id = twilio_message.communication.appointment_reminder_id
+
+      AppointmentNotification::Worker.perform_at(Communication.next_messaging_time, appointment_reminder_id, "missed_visit_sms_reminder")
     end
 
     head :ok
--- a/app/jobs/appointment_notification/worker.rb
+++ b/app/jobs/appointment_notification/worker.rb
@@ -9,17 +9,14 @@ class AppointmentNotification::Worker
   def perform(appointment_reminder_id, communication_type)
     reminder = AppointmentReminder.includes(:appointment, :patient).find(appointment_reminder_id)
     return if reminder.appointment.previously_communicated_via?(communication_type)
-    check_for_errors(reminder)
+    if reminder.status != "scheduled"
+      report_error("scheduled appointment reminder has invalid status")
+      return
+    end
     send_message(reminder, communication_type)
   end
 
   private
-
-  def check_for_errors(reminder)
-    if reminder.status != "scheduled"
-      report_error("scheduled appointment reminder has invalid status")
-    end
-  end
 
   def send_message(reminder, communication_type)
     notification_service = NotificationService.new
--- a/app/jobs/appointment_notification/worker.rb
+++ b/app/jobs/appointment_notification/worker.rb
@@ -6,11 +6,11 @@ class AppointmentNotification::Worker
 
   DEFAULT_LOCALE = :en
 
-  def perform(appointment_reminder_id)
+  def perform(appointment_reminder_id, communication_type)
     reminder = AppointmentReminder.includes(:appointment, :patient).find(appointment_reminder_id)
     return if reminder.appointment.previously_communicated_via?(communication_type)
     check_for_errors(reminder)
-    send_message(reminder)
+    send_message(reminder, communication_type)
   end
 
   private
@@ -19,16 +19,13 @@ class AppointmentNotification::Worker
     if reminder.status != "scheduled"
       report_error("scheduled appointment reminder has invalid status")
     end
-    if reminder.next_communication_type.nil?
-      report_error("scheduled appointment reminder does not have a next communication type")
-    end
   end
 
-  def send_message(reminder)
+  def send_message(reminder, communication_type)
     notification_service = NotificationService.new
 
     begin
-      response = if reminder.next_communication_type == "missed_visit_whatsapp_reminder"
+      response = if communication_type == "missed_visit_whatsapp_reminder"
         notification_service.send_whatsapp(
           phone_number(reminder.patient),
           appointment_message(reminder),
@@ -41,20 +38,20 @@ class AppointmentNotification::Worker
           callback_url
         )
       end
-      create_communication(reminder, response)
+      create_communication(reminder, communication_type, response)
       mark_reminder_sent(reminder)
     rescue Twilio::REST::TwilioError => e
       report_error(e)
     end
   end
 
-  def create_communication(reminder, response)
+  def create_communication(reminder, communication_type, response)
     Communication.create_with_twilio_details!(
       appointment: reminder.appointment,
       appointment_reminder: reminder,
       twilio_sid: response.sid,
       twilio_msg_status: response.status,
-      communication_type: reminder.next_communication_type
+      communication_type: communication_type
     )
   end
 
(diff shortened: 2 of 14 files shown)


app/controllers/api/v3/twilio_sms_delivery_controller.rb
```diff
@@ -7,10 +7,11 @@ def create
 
     twilio_message.update(update_params)
     communication_type = twilio_message.communication.communication_type
-    appointment_id = twilio_message.communication.appointment_id
 
     if communication_type == "missed_visit_whatsapp_reminder" && twilio_message.unsuccessful?
-      AppointmentNotification::Worker.perform_at(Communication.next_messaging_time, appointment_id, "missed_visit_sms_reminder")
+      appointment_reminder_id = twilio_message.communication.appointment_reminder_id
```
kpethtel, line 12: IDK how long twilio might wait before returning a failed response, so it's possible some of these may have an appointment_reminder_id of nil and the worker will fail. The failed types are failed, undelivered, and unknown. I would guess failed and undelivered happen quickly.  We should probably do a 1off query to determine what the turnaround time is on failed messages, but my guess is that's it's typically very quick.

app/jobs/appointment_notification/worker.rb
```diff
@@ -6,50 +6,77 @@ class AppointmentNotification::Worker
 
   DEFAULT_LOCALE = :en
 
-  def perform(appointment_id, communication_type, locale = nil)
-    appointment = Appointment.find_by(id: appointment_id)
-    unless appointment
-      logger.warn "Appointment #{appointment_id} not found, skipping notification"
+  def perform(appointment_reminder_id, communication_type)
+    reminder = AppointmentReminder.includes(:appointment, :patient).find(appointment_reminder_id)
+    return if reminder.appointment.previously_communicated_via?(communication_type)
```
kpethtel, line 11: This will have to change for the experiment because we can send multiple of the same type. I have a version of that logic on another branch. But I think this should work for non-experiment reminders.

app/jobs/appointment_notification/worker.rb
```diff
@@ -6,50 +6,77 @@ class AppointmentNotification::Worker
...
+  def send_message(reminder, communication_type)
+    notification_service = NotificationService.new
 
     begin
-      notification_service = NotificationService.new
-
       response = if communication_type == "missed_visit_whatsapp_reminder"
-        notification_service.send_whatsapp(patient_phone_number, message, callback_url)
+        notification_service.send_whatsapp(
+          phone_number(reminder.patient),
```
rsanheim, line 27: This method extraction doesn't buy us anything, and actually obscures the logic of how a phone number is grabbed from the patient.

app/services/appointment_notification_service.rb
```diff
@@ -16,17 +16,32 @@ def initialize(appointments:, days_overdue: 3)
...
 
   private
 
+  def create_appointment_reminder(appointment)
+    AppointmentReminder.create!(
+      appointment: appointment,
+      patient: appointment.patient,
+      remind_on: appointment.remind_on,
+      status: "pending",
+      message: "sms.appointment_reminders.#{communication_type}"
```
rsanheim, line 42: Does the `sms` prefix make sense in this key anymore?  Because this also could be for WhatsApp right?

spec/rachets/lets_not_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "spec_helper"
 
 RSpec.describe "Rachet down usage of let!" do
-  expected_usages = 581
+  expected_usages = 577
```
rsanheim, line 4: ✨

app/jobs/appointment_notification/worker.rb
```diff
@@ -6,50 +6,77 @@ class AppointmentNotification::Worker
...
       end
-
-      Communication.create_with_twilio_details!(
-        appointment: appointment,
-        twilio_sid: response.sid,
-        twilio_msg_status: response.status,
-        communication_type: communication_type
-      )
+      create_communication(reminder, communication_type, response)
+      mark_reminder_sent(reminder)
```
rsanheim, line 39: Extracting a method here to send a message directly to the object feels unnecessary.

app/jobs/appointment_notification/worker.rb
```diff
@@ -6,50 +6,77 @@ class AppointmentNotification::Worker
...
+        )
       end
-
-      Communication.create_with_twilio_details!(
-        appointment: appointment,
-        twilio_sid: response.sid,
-        twilio_msg_status: response.status,
-        communication_type: communication_type
-      )
+      create_communication(reminder, communication_type, response)
```
rsanheim, line 38: The creation of the communication and the update of the reminder should be wrapped in a transaction.

app/jobs/appointment_notification/worker.rb
```diff
@@ -6,50 +6,77 @@ class AppointmentNotification::Worker
...
+      communication_type: communication_type
+    )
+  end
 
-  def appointment_message(appointment, communication_type, locale)
+  def mark_reminder_sent(reminder)
+    reminder.status_sent!
+  end
+
+  def appointment_message(reminder)
```
kpethtel, line 59: I think you could make an argument for this logic living within appointment reminder.

app/services/appointment_notification_service.rb
```diff
@@ -16,17 +16,32 @@ def initialize(appointments:, days_overdue: 3)
 
   def send_after_missed_visit
     eligible_appointments = appointments.eligible_for_reminders(days_overdue: days_overdue)
+    next_messaging_time = Communication.next_messaging_time
 
     eligible_appointments.each do |appointment|
+      # i don't believe this is the best way to control this. combined with the fact that
+      # we grab all appointments with remind_on before today, I suspect this means we're
+      # resending the same reminders because they failed the first time.
+      # leaving it in place for now in an effort to change as little of the current process as possible.
```
rsanheim, line 25: Yeah, this is where we could do a tiny PR to add some logging here to see if we have repeated old notifications going out and failing everyday.

rsanheim (changes requested):
Some feedback and questions.

app/services/appointment_notification_service.rb
kpethtel: Agreed. That is true for "TwilioSmsDeliveryDetail" too. And, if we really want to make changes here, I don't think there are any differences between any of the sms and whatsapp translations. I think it was a premature optimization.

app/services/appointment_notification_service.rb
kpethtel: I've made a 1off request that I'm hoping sheds light on that question.

app/jobs/appointment_notification/worker.rb
kpethtel: The communication and twilio detail creation are wrapped in a transaction within `Communication.create_with_twilio_details`, but I'll wrap

kpethtel:
> Some feedback and questions.

There are two comments here I didn't address in code. 
1) adding logging to determine if there's a bug resulting in resending notifications. I think the conversation here confirms that there is. AFAICT the only other use case we have for Communications is for sending OTP codes and those are one-time only, so I doubt they're showing up in these results. https://simpledotorg.slack.com/archives/C013HS4N9U3/p1618500349023600
2) changing the naming conventions for the locales. I'm open to changing those, but changing the "sms" namespace within locales would also impact the OTP messages. I think it would be better to make those changes in a subsequent PR after confirming that this code works. 
We can talk about this today if you like, as well as about how to test this code before sending it to prod. I assume we'll want to set up an appointment in sandbox/demo and confirm through the logs that a reminder would have sent.

app/jobs/appointment_notification/worker.rb
rsanheim: I agree with your argument.  Want to move it over? 😁

rsanheim (approved):
LGTM. Approving to unblock with one minor note of feedback.

There are also a couple linter things to cleanup.

rsanheim:
> 2\. hanging the naming conventions for the locales. I'm open to changing those, but changing the "sms" namespace within locales would also impact the OTP messages. I think it would be better to make those changes in a subsequent PR after confirming that this code works.

I agree doing it in a future PR makes sense, no need for more refactoring in this PR.

app/jobs/appointment_notification/worker.rb
kpethtel: I think I'll do that as part of the subsequent "make this work for experiment reminders" ticket because this one is blocking things.

PR #2426, opened 2021-04-15
Clean up unnecessary notification Org based feature flag

rsanheim:
If appointments are enabled for the country (via `APPOINTMENT_REMINDERS` feature toggle) we send to all Orgs in that country.

There is no reason for an additional level of feature flagging here.

@harimohanraj89 confirmed in Slack we don't need this anymore, as its a hold over from when we used to have two orgs in India.  That is no longer the case... https://simpledotorg.slack.com/archives/C01FMV58MEC/p1617903248049300?thread_ts=1617901116.046800&cid=C01FMV58MEC


kpethtel (approved):
Makes sense to me. I've heard a couple of times now that we've moved to a single org, so maybe I should've thought about this.

PR #2427, opened 2021-04-15
3193: assign patients to groups completely at random rather than usin…

kpethtel:
…g deterministic hash

**Story card:** [ch3193](https://app.clubhouse.io/simpledotorg/story/3193/shift-away-from-deterministic-hash-for-patient-bucketing-strategy?vc_group_by=day&ct_workflow=all)

## Because

We initially designed experiments to bucket patients according to a deterministic hash in order to more easily track which bucket they're in. That plan was flawed because it left it very difficult to know who was in the control group. We solved that problem by adding group memberships, which negated the need for determinism. And because the deterministic hash was based solely on the patient's uuid, it would always have the same result, which could result in a patient always being in the same treatment group if experiments have the same groups in the same order, which seems very possible.


ssrihari:
> because it left it very difficult to know who was in the control group.

I'm interested in understanding this better. Is there somewhere I can read more about this? Happy to get on a call as well if that's better. All non blocking, of course.

kpethtel:
> > because it left it very difficult to know who was in the control group.
> 
> I'm interested in understanding this better. Is there somewhere I can read more about this? Happy to get on a call as well if that's better. All non blocking, of course.

You would always know which bucket a person would be in if they were in the experiment. The problem is that you wouldn't know who is in the experiment. The selection criteria are not static. They change over time. The sql for selecting patients is somewhat complex and wouldn't produce the same results a month later. For non-control patients, you can infer who was in the experiment and which group they were in based on which appointment reminders they have. But the control group doesn't have any appointment reminders, so you can't infer. That's why we added a group membership join table and why we no longer need deterministic hashing.

ssrihari:
> https://github.com/simpledotorg/simple-server/pull/2427#issuecomment-821135172

Thanks @kpethtel, that makes sense 👍

app/models/experimentation/experiment.rb
```diff
@@ -29,9 +29,8 @@ def self.candidate_patients
         .where(["experiments.end_date < ? OR experiments.id IS NULL", ExperimentControlService::LAST_EXPERIMENT_BUFFER.ago]).references(:experiment)
     end
 
-    def group_for(uuid)
-      hash = Zlib.crc32(uuid) % treatment_groups.length
-      treatment_groups.find_by(index: hash)
+    def random_treatment_group
+      treatment_groups[rand(0..treatment_groups.count - 1)]
```
rsanheim, line 33: `treatment_groups.sample` would be simpler here I think

kpethtel: Didn't think of that. Good idea. I'll do that.

PR #2430, opened 2021-04-16
Cleanup block sync feature flagging

prabhanshuguptagit:
**Story card:** [ch2563](https://app.clubhouse.io/simpledotorg/story/2563/remove-all-block-level-sync-feature-flag-related-code)

## Because

Block level sync has been rolled out to all users. We can cleanup the branching around conditionally syncing block data.

## This addresses

Cleans up feature flagging for block level sync. Block level syncing will be the mainline code now.

- [x] Add a wiki/manual on how to switch to a different sync region. Maybe an ADR? -- Added an ADR in #2439.
- [x] Test on perf with real data
- [x] Test with an older build (Pre-block level sync)


app/controllers/api/v3/sync_controller.rb
```diff
@@ -14,7 +14,6 @@ def __sync_to_user__(response_key)
     records = records_to_sync
 
     AuditLog.create_logs_async(current_user, records, "fetch", Time.current) unless disable_audit_logs?
-    log_block_level_sync_metrics(response_key)
```
harimohanraj89, line 17: Is some form of this logging still valuable? If our regular sync logging has enough detail (sync region ID for every sync request), then we should be good.

app/controllers/api_controller.rb
```diff
@@ -44,6 +44,7 @@ def current_sync_region
   end
 
   def block_level_sync?
+    return false unless current_user
```
harimohanraj89, line 47: You could tack on a `current_user &&` on the next line if you want 😅 

I'll let you pick whichever seems more readable/natural 👍

app/controllers/api_controller.rb
```diff
@@ -44,6 +44,7 @@ def current_sync_region
...
     current_user.block_level_sync? && requested_sync_region_id == current_block.id
```
harimohanraj89, line 48: This still exists because some users (telemed officers) are not eligible for block syncing, regardless of the feature flag, right?

spec/models/user_spec.rb
```diff
@@ -339,28 +339,8 @@
...
-
-    it "is true when the flipper flag is on" do
-      expect(user.block_level_sync?).to eq true
-    end
-
-    it "is false when the flipper flag is off" do
-      Flipper.disable(:block_level_sync, user)
-      expect(user.block_level_sync?).to eq false
-    end
-
```
harimohanraj89, line 363: We should add another test here for `is truen when user is not enabled for teleconsultation`

harimohanraj89 (approved):
Looks like a clean break.

app/controllers/api/v3/sync_controller.rb
prabhanshuguptagit: We get sufficient information through the headers. This was mainly feeds the [datadog dashboard](https://app.datadoghq.com/dashboard/5gh-pzf-66k/block-level-sync-india-production), which I think we can remove now (haven't looked at it in a _long time_).

app/controllers/api_controller.rb
prabhanshuguptagit: I actually switched from `current_user&.` to `current_user &&` and then to this 😛 . I tried to state this explicitly, in a similar style as the `current_sync_region` method.

app/controllers/api_controller.rb
prabhanshuguptagit: Yes, just for telemed officers.

app/controllers/api/v3/facilities_controller.rb
```diff
@@ -26,7 +26,7 @@ def disable_audit_logs?
   def transform_to_response(facility)
     Api::V3::FacilityTransformer
       .to_response(facility)
-      .merge(sync_region_id: sync_region_id(facility))
+      .merge(sync_region_id: facility.block_region_id)
```
ssrihari, line 29: Should this be a config? For small districts, we could sync the full district right? It can also be different for different countries. I don't know if this fits within the gamut of this PR though.

app/controllers/api_controller.rb
ssrihari: In that case, can we remove `current_user.block_level_sync?` and replace it with just `!can_teleconsult?` and move that condition into the `current_sync_region` function? If `block_level_sync` is being deprecated as a concept, with this PR, we should see all instances of `block_level_sync` in the codebase gone.

app/controllers/api/v3/facilities_controller.rb
prabhanshuguptagit: As of now we block sync everywhere. Moving to config would be easy when the need arises.

app/controllers/api_controller.rb
prabhanshuguptagit: We will continue to do both kind of syncs even after this PR. The `block_level_sync` concept still exists _and_ is tied to a user.  I avoided putting the logic in the controller since the selection criteria for block sync is a bit arbitrary (can the user teleconsult?) and subject to change.

As an alternative, we add a `district_level_sync?` on `User`, since block level sync is the default now. Added in c2295bca, wdyt?

app/controllers/api/v3/facilities_controller.rb
harimohanraj89: It very well could be, but I agree - this probably isn't the PR to make that change 😄

ssrihari: cool 👌

PR #2432, opened 2021-04-16
Easier configurable regions for reports

rsanheim:
**Story card:** [chXXXX](URL)

No story yet

We need a better way to configure what regions show up in what country.  This will allow things like org level reports or more nuanced levels to display for each country.


harimohanraj89 (approved):
Yup, this is definitely the right direction.

Would the final form be - config stored in DB, configurable through dashboard?

kpethtel (approved):
Any reason we don't test reportable_region for Ethiopia?

PR #2436, opened 2021-04-19
Convert to Repository for DistrictAnalyticsQuery numbers

rsanheim:
this make up the details tab of reports, so to keep things consistent we
should grab them from the same repository that backs up other parts of
the app..

**Story card:** [ch2625](https://app.clubhouse.io/simpledotorg/story/2625/registration-counts-on-details-and-overview-don-t-match) and [ch3343](https://app.clubhouse.io/simpledotorg/story/3343/convert-to-repository-for-districtanalyticsquery-numbers)

## Because

We are trying to consolidate more queries to one consistent path

## This addresses

Moving part of the 'analytics queries' to repository

## Test instructions


app/views/reports/regions/_new_registrations_table.html.erb
```diff
@@ -24,6 +24,7 @@
         <th class="row-label"><strong>Total</strong></th>
         <th class="row-label" colspan="6"><strong>New registrations</strong></th>
       </tr>
+      <% raise "used?" %>
```
rsanheim, line 27: Pretty sure this view is unused...but wanted to verify with folks before removing.  Couldn't find it in click testing or via this raise.

app/views/reports/regions/details.html.erb
```diff
@@ -49,14 +50,14 @@
                 Total registered<br>
                 patients
             </th>
-            <% dates_for_periods(@period.type, 6, include_current_period: @show_current_period).each do |date| %>
+            <% @period_range.each do |period| %>
```
rsanheim, line 53: All these period changes cleaned up the views nicely.

app/views/reports/regions/details.html.erb
```diff
@@ -67,16 +68,18 @@
...
+                <%= number_or_dash_with_delimiter(@repository.registration_counts[@region.region.slug][period]) %>
               </td>
             <% end %>
-            <% dates_for_periods(@period.type, 6, include_current_period: @show_current_period).each do |date| %>
+            <% @period_range.each do |period| %>
               <td class="row-total ta-center">
-                <%= number_with_delimiter(dash_if_zero(analytics_totals(@dashboard_analytics, :follow_up_patients_by_period, date))) %>
+                <%= number_with_delimiter(dash_if_zero(analytics_totals(@dashboard_analytics, :follow_up_patients_by_period, period.value))) %>
+                ||
+                <%= number_or_dash_with_delimiter(@repository.hypertension_follow_ups(group: "blood_pressures.user_id")[@region.region.slug][period]) %>
```
rsanheim, line 82: I was verifying numbers locally with lots of manual QA by displaying both -- I'l remove the legacy numbers before merging this.

app/models/patient.rb
```diff
@@ -105,23 +104,6 @@ class Patient < ApplicationRecord
   validates_associated :address, if: :address
   validates_associated :phone_numbers, if: :phone_numbers
 
-  def self.follow_ups_with(model_name, period, time_column: "recorded_at", at_region: nil, current: true, last: nil)
```
rsanheim, line 108: I was struggling with figuring out how to handle follow ups, and I realized extracting to its own object would be the best way. Ideally the api for these sorts of queries will be much more constrained in the future, but right now too many code paths access the scopes above to remove them.

rsanheim (commented):
some notes...

app/views/reports/regions/_new_registrations_table.html.erb
harimohanraj89: Looks unused to me too 🤷

app/views/reports/regions/details.html.erb
```diff
@@ -49,14 +50,14 @@
                 Total registered<br>
                 patients
             </th>
-            <% dates_for_periods(@period.type, 6, include_current_period: @show_current_period).each do |date| %>
+            <% @period_range.each do |period| %>
               <th class="row-label sort-label sort-label-small" data-sort-method="number">
-                <%= format_period(@period.type, date) %>
+                <%= period %>
```
harimohanraj89, line 55: Oof, nice. `to_s` kicks in as we'd like?

app/views/reports/regions/details.html.erb
```diff
@@ -87,16 +90,16 @@
                   <%= link_to resource.send(:full_name), send(:admin_user_path, resource, period: @period) %>
                 </td>
                 <td class="ta-center">
-                  <%= number_with_delimiter(dash_if_zero(@dashboard_analytics.dig(resource.id, :total_registered_patients))) %>
+                  <%= number_or_dash_with_delimiter(@dashboard_analytics.dig(resource.id, :total_registered_patients)) %>
```
harimohanraj89, line 93: Technically, it's `number_with_delimiter_or_dash`. While we're consolidating this, shall we name the method on what it does, not how it does it? Eg. `format_indicator`?

config/initializers/time_and_date_formats.rb
```diff
@@ -1,6 +1,9 @@
 Time::DATE_FORMATS[:mon_year] = "%b-%Y"
 Date::DATE_FORMATS[:mon_year] = "%b-%Y"
 
+Date::DATE_FORMATS[:mon_year_multiline] = lambda { |date| date.strftime("%b-%Y").tr("-", "\n") }
+Time::DATE_FORMATS[:mon_year_multiline] = lambda { |date| date.strftime("%b-%Y").tr("-", "\n") }
```
harimohanraj89, line 5: Silly question, can we not put the newline char directly into a string literal?

`Date::DATE_FORMATS[:mon_year_multiline] = "%b\n%Y"?

app/views/reports/regions/details.html.erb
harimohanraj89: We're directly grabbing data from the repository here, whereas we're delegating that to the "analytics query" for registration info. Is that inconsistent, using the repository directly as well as through that query class? Can we add the necessary `@dashboard_analytics.hypertension_follow_ups(group: :something)` to the analytics query classes and keep our access consistently through `@dashboard_analytics`?

harimohanraj89 (changes requested):
My thoughts!

Additionally, how isolated are the codepaths we're touching to just districts? Can we get away with updating `DistrictAnalyticsQuery` without having to touch `FacilityAnalyticsQuery` at the same time?

app/views/reports/regions/details.html.erb
rsanheim: yup, that's the default for ERB `<%= `

rsanheim:
@harimohanraj89 to answer this question:

> We're directly grabbing data from the repository here, whereas we're delegating that to the "analytics query" for registration info. Is that inconsistent, using the repository directly as well as through that query class? Can we add the necessary @dashboard_analytics.hypertension_follow_ups(group: :something) to the analytics query classes and keep our access consistently through @dashboard_analytics?

I found it easier to grab things directly from the repository, rather than going through the multiple layers of service objects that the old code paths were using. The interface ended up being more consistent, and I'm less worried about breaking other code paths by going this route.

app/views/reports/regions/details.html.erb
rsanheim: I wanted to keep this consistent with the Rails view helper `number_with_delimiter` because our little helper is essentially just a wrapper around it.

harimohanraj89:
Ah I see - so we're moving _towards_ using the repository directly, but not retiring the existing `DistrictAnalyticsQuery` completely in one fell swoop. 👍

harimohanraj89 (approved):
This is the right direction, but it sounds like we've still got some work to do in terms of:
* Consistent use of the repository
* Retiring legacy "query" classes like `DistrictAnalyticsQuery`

Can we roughly story up the new few big refactors necessary to see this to completion?

app/models/facility.rb
```diff
@@ -145,15 +145,11 @@ def source
   # ----------------
 
   def hypertension_follow_ups_by_period(*args)
-    patients
-      .hypertension_follow_ups_by_period(*args)
-      .where(blood_pressures: {facility: self})
+    patients.hypertension_follow_ups_by_period(*args).where(blood_pressures: {facility: self})
```
vkrmis, line 148: [nitpick] I find this less readable than the multiline version

app/queries/follow_ups_query.rb
```diff
@@ -0,0 +1,54 @@
+class FollowUpsQuery
+  def initialize(region, period_type, group_by: nil)
+    @region = region
+    @period_type = period_type
+    @group_by = group_by
+    @formatter = lambda { |v| @period_type == :quarter ? Period.quarter(v) : Period.month(v) }
+  end
+
+  def hypertension
```
vkrmis, line 9: This is partially replicated in the `with` method, can we just call `with` instead?

spec/queries/follow_ups_query_spec.rb
```diff
@@ -0,0 +1,248 @@
...
+      # another follow up in the same month but another day
+      create(:blood_pressure,
+        :with_encounter,
+        patient: hypertensive_patient,
+        facility: current_facility,
+        user: current_user,
+        recorded_at: second_follow_up_date)
+    end
+
+    describe ".follow_ups" do
```
vkrmis, line 126: Since this is testing `Patient.follow_ups_by_period`, shouldn't these be in `Patient` instead?
Alternatively, we could test a call to `FollowUpsQuery` here.

app/queries/follow_ups_query.rb
vkrmis: Can we do one of:
- Keep all the `FollowUpsQuery` scopes in Patient and remove `.hypertension` here
- Move all these ^ scopes to `FollowUpsQuery`

Doing both is a bit confusing.

app/models/reports/repository.rb
```diff
@@ -106,6 +110,42 @@ def initialize(regions, periods:)
...
+    # .     user_id_1 => cumulative_count,
+    # .     user_id_2 => cumulative_count,
+    # .  }
+    # }
+    #
+    memoize def cumulative_registration_counts_by_user
+      registration_counts_by_user.each_with_object({}) do |(slug, period_counts), totals|
+        totals[slug] = {}
+        # collect all the user ids in a region that we need to count for
+        user_ids = period_counts.each_with_object(Set.new) { |(period, counts), sum| sum.merge(counts.keys) }
```
vkrmis, line 134: Can we rename:
- `counts` to `user_counts`
- `sum` to `user_ids`

vkrmis: Can this be simplified to something like `user_ids = period_counts.values.map(&:keys).flatten.to_set`

app/models/reports/repository.rb
```diff
@@ -209,7 +256,7 @@ def denominator(region, period)
     #     region_2_slug: { period_1: value, period_2: value }
     #   }
     #
-    def cached_query(calculation, &block)
+    def region_period_cached_query(calculation, &block)
```
vkrmis, line 259: Have we added other `_cached_queries`? Wondering why we're being more specific here. Is `cached_query` not descriptive enough?

app/views/reports/regions/details.html.erb
```diff
@@ -67,36 +68,40 @@
...
+                %>
               </td>
             <% end %>
           </tr>
           <% current_admin.accessible_users(:view_reports).order(:full_name).each do |resource| %>
             <% if @dashboard_analytics[resource.id].present? %>
               <tr>
                 <td class="row-title">
-                  <%= link_to resource.send(:full_name), send(:admin_user_path, resource, period: @period) %>
+                  <%= link_to resource.full_name, admin_user_path(resource, period: @period) %>
```
vkrmis, line 91: ❤️

vkrmis (commented):
Some overall thoughts:
- There's some follow ups logic that's distributed between `Patient` and `FollowUpsQuery`, can we consolidate these in one place?
- Is this a good time to scrap `DistrictAnalyticsQuery`?, it doesn't do much that `Repository` doesn't give us directly. 


Prabhanshu & vkrm

spec/queries/district_analytics_query_spec.rb
```diff
@@ -72,11 +77,11 @@
         #
         Timecop.travel(month + 2.months) do
           patients_1.each do |patient|
-            create(:blood_pressure, patient: patient, facility: facility_1)
+            create(:blood_pressure, patient: patient, facility: facility_1, user: user)
           end
 
           patients_2.each do |patient|
-            create(:blood_pressure, patient: patient, facility: facility_2)
+            create(:blood_pressure, patient: patient, facility: facility_2, user: user)
```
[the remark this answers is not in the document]
kpethtel: I assume because the factories were making unique users per bp reading. This is definitely the kind of thing that makes factorybot hard to use efficiently.

kpethtel (approved):
Really nice!

PR #2437, opened 2021-04-20
Add india national health id as an identifier type

ssrihari:
**Story card:** [ch3220](https://app.clubhouse.io/simpledotorg/story/3220)

## Because

We need to support india national health ids as identifier types for future app versions


prabhanshuguptagit (approved):

harimohanraj89:
@ssrihari did we confirm that it's ok to send NHIDs to older app versions? Will they ignore them gracefully?

ssrihari:
> @ssrihari did we confirm that it's ok to send NHIDs to older app versions? Will they ignore them gracefully?

@harimohanraj89 Yup. We've made some good decisions in the past 😁  See [related ADR](https://github.com/simpledotorg/simple-android/blob/master/doc/arch/006-forward-compatibility-for-enums.md), and [slack conversation](https://simpledotorg.slack.com/archives/C01T2MYJB97/p1618899070015100).

harimohanraj89:
Toit 😎

PR #2438, opened 2021-04-20
Add readme for patient data transfer script

prabhanshuguptagit:
**Story card:** [ch2497](https://app.clubhouse.io/simpledotorg/story/2947/chikmagalur-facility-cleanup)

## This addresses

This adds a doc for the patient data transfer script added in #2424.


doc/wiki/move-patient-data-between-facilities.md
```diff
@@ -0,0 +1,18 @@
...
+Users of Simple might need a facility's data to be transferred to another facility in some circumstances:
+- Patients were accidentally created in the wrong facility
+- A facility needs to be deleted and merged with another facility
+
+This script can assist with transferring patient data in such cases.
+```bash
+bundle exec cap india:production deploy:rake task=data_fixes:move_data_from_source_to_destination_facility[<source-facility-id>,<destination-facility-id>]
+```
+
+### Caveats
```
harimohanraj89, line 12: Can we first note a couple points on what the task _does_ transfer?

doc/wiki/move-patient-data-between-facilities.md
```diff
@@ -0,0 +1,18 @@
...
+
+- This does not transfer users from the source facility. In case users also need to be moved to the destination
+facility, it can be done manually via the `Edit User` UI or from console.
+- In case the source facility needs to be deleted, make sure the users also switch their facility in the app.
```
ssrihari, line 16: > make sure the users also switch their facility in the app.

Worth writing about how one would go about doing this.

doc/wiki/move-patient-data-between-facilities.md
```diff
@@ -0,0 +1,18 @@
+# Moving patient data from one facility to another
```
ssrihari, line 1: There was another doc or section in another doc about this right? Should we link to it? Or deprecate that?

doc/wiki/move-patient-data-between-facilities.md
```diff
@@ -0,0 +1,18 @@
...
+
+Users of Simple might need a facility's data to be transferred to another facility in some circumstances:
+- Patients were accidentally created in the wrong facility
+- A facility needs to be deleted and merged with another facility
+
+This script can assist with transferring patient data in such cases.
+```bash
+bundle exec cap india:production deploy:rake task=data_fixes:move_data_from_source_to_destination_facility[<source-facility-id>,<destination-facility-id>]
```
ssrihari, line 9: Perhaps replace `india:production` with something generic? Don't want to accidentally run with that.

doc/wiki/move-patient-data-between-facilities.md
```diff
@@ -0,0 +1,18 @@
...
+```
+
+### Caveats
+
+- This does not transfer users from the source facility. In case users also need to be moved to the destination
+facility, it can be done manually via the `Edit User` UI or from console.
+- In case the source facility needs to be deleted, make sure the users also switch their facility in the app.
+If there is a delay between the transfer and the soft deletion, its possible that a user creates more data in between which will hamper the deletion.
+Ideally the transfer and deletion should be atomic.
```
harimohanraj89, line 18: The caveats are _just_ gnarly enough for me to feel like I need a "Ideal workflow" section right below this. Can we add one? Something like:

```
1. Notify CVHO to move all users to destination facility in dashboard
2. Run script
3. Log into dashboard and delete old facility
```

harimohanraj89 (changes requested):
Some suggestions from me!

ssrihari (commented):
Looks good! Also note that we should move this functionality into the dashboard if we're using this rake task often.

doc/wiki/move-patient-data-between-facilities.md
prabhanshuguptagit: That one is for moving a user's data, we'll need them both.

harimohanraj89 (approved):
My feedback's addressed 🚀

PR #2441, opened 2021-04-21
Fix flaky patient summary spec

prabhanshuguptagit:
[ch3526](https://app.clubhouse.io/simpledotorg/story/3526/fix-flaky-patient-line-list-spec)

## Because

The patient summary spec fails randomly due to mismatch in the age. It only fails on semaphore, not locally. This is a follow up to #2419.

## This addresses

The flakiness only appears when the patient doesn't have an DoB. In that case we calculate the age using the `age` and `age_updated_at` fields in SQL. The SQL calculation is off by one from our `current_age` method randomly. 

This removes the DoB randomization in patient factory. The off by one bug would still exist.

PS We shouldn't randomize attributes in factories as a rule of thumb since it has led to to confusing specs in the past.

## Update

This was happening due to this behaviour of `travel_to`:
```
the usec for the time passed will be set to 0 to prevent rounding errors with external services, like MySQL (which will round instead of floor, leading to off-by-one-second errors).
```
[Link](https://api.rubyonrails.org/v5.1.7/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel_to)

Since `travel_to` drops the microsecond from timestamps,`travel_to Time.current` could move you to upto one millisecond in the past. In this case, if the `age_updated_at` is within the same millisecond as `Time.current`, `Time.current - age_updated_at` will be negative. This results in `current_age` to be 1 less than the matview.

This sets the `age_updated_at` to a time in the past, to avoid the rounding error by `travel_to`.

This is a spec problem, not a feature issue.


kpethtel (approved):
I agree with you that we shouldn't be randomizing this factory, but I think we should just remove this `transient` block altogether and remove the condition from these two lines:
```
    date_of_birth { rand(18..80).years.ago if has_date_of_birth? }
    age { rand(18..100) unless has_date_of_birth?
```

ssrihari:
> The SQL calculation is off by one from our current_age method randomly.

We need to fix this.

I'm almost tempted to leave this randomly failing test as is as a reminder that this is still broken. If we fix the broken test this way, then we will have an illusion of a passing test suite when we still have a known bug. Can we write a different failing test perhaps?

> PS We shouldn't randomize attributes in factories as a rule of thumb since it has led to to confusing specs in the past.

I like the randomness, actually. It elevates us slightly from writing purely example based tests to using a slightly wider range of test data. The randomness has clearly helped us catch a bug in this case.

If the confusion is costly in terms of our time to debug etc, it makes sense to keep it deterministic. However, this has been around since the time @govindkrjoshi and I added it at the beginning of the project, and hasn't really costed us that much right? Please do cmiiw.

prabhanshuguptagit:
> We need to fix this.

Yeah, I agree. Although as I was looking at this, I remembered that we spent a fair bit of time trying to address this in the [PR this spec was introduced](https://github.com/simpledotorg/simple-server/pull/1950) but had no luck. I'm ok with letting this flake until we can fix this.

We should still remove the factory randomisation though. Apologies for putting it as a `rule of thumb because it's hurt in the past`. I think mainly:
- Some randomised attributes at the model can cause a big affect down the line. As a slightly exaggerated example, lets randomise a patient's hypertensive-ness between `true` and `false`. This is probably good for tests at the model level. At the reports layer though, it will cause some metric to have a value of `0` or `10` depending on the sampling. The reports specs become hard to write (and read).
- This is only a personal experience, I don't have data to back this up: I've found it easy to fall for a false sense of coverage with this kind of minor randomisation. Testing for these attributes explicitly in the tests forces to think about these edge cases a little more actively.

harimohanraj89:
In my experience, the randomness of test data is not worth the edge cases like this it flushes out. I've found that smarter tests that hit the boundaries are usually better.

Eg., instead of randomizing age and dob, we should have two deterministic tests that take a patient that's 30yrs364days old, and 31yrs1days old.

I'd be in favor of deterministic-izing these specs. That unfortunately means we'd also wanna fix that off-by-one right now as well. Think we can figure that one out here?

ssrihari:
I am okay with removing the randomisation here. I think the benefits are small, when weighed wrt your valid experience writing/reading related reporting tests. We would lose the ability to pick up on bugs such as this one, but that's a fair trade off.

That said,

> The reports specs become hard to write (and read).

I feel like creating test data for reports where date is an important aspect would have date specified explicitly anyway. Or we _can_ just specify the age/dob in a `create_list` invocation, and that wouldn't be too bad.

> false sense of coverage with this kind of minor randomisation

At least when we added it, it was intended to catch exactly this kind of a bug. We would still add specific tests for the edge cases where we could think of them, these aren't mutually exclusive.

ssrihari:
> I've found that smarter tests that hit the boundaries are usually better.

That's assuming we're smart enough to think of all the boundaries the first time we write these tests :). As programmers, we're biased having written the code that is being tested, and are often overconfident of how well it works. We aren't inherently incentivised enough to think of all the edge cases.

That's the case for generative testing anyway. Obviously this `rand` in the factory is not a generative test, but it provides a tiny amount of the benefits. I was hoping to take that further at some point.

Just to be clear, I also think it's not worth keeping the `rand` here. I'm just indulging in related conversation :)

harimohanraj89:
Fair points indeed. Yup, the base factory probably isn't the place to add the randomization for generative tests 😅 

We'll have some work to do here before we can merge in the fix 😄

spec/exporters/patients_exporter_spec.rb
```diff
@@ -156,19 +149,19 @@
   end
 
   describe "#csv" do
-    let(:now) { Time.current }
-    let(:patient_batch) { Patient.where(id: patient.id) }
```
prabhanshuguptagit, line 160: This is a refactor to simplify the assertions.

spec/factories/patients.rb
```diff
@@ -4,17 +4,13 @@
...
     id { SecureRandom.uuid }
     gender { Seed::Gender.random_gender }
     full_name { common_names[gender.to_sym].sample + " " + common_names[gender.to_sym].sample }
     status { Patient::STATUSES[0] }
-    date_of_birth { rand(18..80).years.ago if has_date_of_birth? }
-    age { rand(18..100) unless has_date_of_birth? }
-    age_updated_at { Time.current }
+    date_of_birth { nil }
+    age { rand(18..100) }
+    age_updated_at { 1.second.ago }
```
prabhanshuguptagit, line 13: This is the actual fix ✨

config/initializers/filter_parameter_logging.rb
```diff
@@ -40,4 +40,4 @@ def self.filter
   end
 end
 
-Rails.application.config.filter_parameters += [:password, :age, ParameterFiltering.filter]
+Rails.application.config.filter_parameters += [:password, ParameterFiltering.filter]
```
prabhanshuguptagit, line 43: This is an unrelated fix that was surfaced by a breaking spec. A full fix will follow up in https://github.com/simpledotorg/simple-server/pull/2519

harimohanraj89: So this PR will start logging ages?

prabhanshuguptagit: Yes. Undid this change in 905a39c, so we can deal with it separately.

spec/exporters/patients_exporter_spec.rb
```diff
@@ -156,19 +149,19 @@
...
-
     it "generates a CSV of patient records" do
-      skip "intermittent test blocking deploys"
-      travel_to now do
+      Timecop.freeze do
```
prabhanshuguptagit, line 153: `Timecop.freeze` does not do any rounding of the timestamp.

harimohanraj89:
:shipit: :shipit: :shipit:  🚢 🚢 🚢

PR #2445, opened 2021-04-22
Ensure latest version of Capistrano-passenger

rsanheim:
otherwise deploys may break

See also https://github.com/capistrano/passenger/pull/54


kpethtel (approved):
Thanks for doing this!

PR #2450, opened 2021-04-23
Require metrics module

prabhanshuguptagit:
**Story card:** -

## Because

https://sentry.io/organizations/resolve-to-save-lives/issues/2357938469/?project=1793711&referrer=slack

## This addresses

Loads the metrics module.

### DIFF
--- a/config/initializers/00_datadog.rb
+++ b/config/initializers/00_datadog.rb
@@ -15,3 +15,4 @@ Datadog.configure do |c|
 end
 
 require "statsd"
+require "metrics"


vkrmis (approved):

rsanheim:
Thanks for the fix! I'm guessing what happened is that because Sidekiq has its own loading process, it did not pick up Metrics via autoloading early enough.

I know I've seen or written a spec in the past to catch this kind of thing, probably worth doing at some point.

PR #2455, opened 2021-04-27
Updated 'Assigned patients breakdown' copy

claudiovallejo:
**Story card:** [ch3332](https://app.clubhouse.io/simpledotorg/story/3332/update-assigned-patients-breakdown-copy)

## Because

The "Assigned patients breakdown" card copy is confusing for the following reasons:

1. The title is not accurate
2. The number displayed in the subtitle is not useful
3. It's not clear what "Transferred out" means

## This addresses

Card copy is much clearer and "Transferred out" copy is now displayed in tooltip and footnoes

## Test instructions

* Updated card title
* Updated card subtitle
* Created new "Transferred out" definition in `en.yml`
* Added "Transferred out" definition to card tooltip
* Added "Transferred out" definition to footnotes section
* Removed hyphen from "Transferred-out" copy in chart tooltip

### DIFF
--- a/app/views/reports/regions/_patient_breakdown_charts.html.erb
+++ b/app/views/reports/regions/_patient_breakdown_charts.html.erb
@@ -78,7 +78,8 @@
           </h3>
           <%= render "definition_tooltip",
                      definitions: { "Patients under care" => t("patients_under_care_copy"),
-                                    "Lost to follow-up" => t("lost_to_follow_up_copy.reports_card_subtitle") } %>
+                                    "Lost to follow-up" => t("lost_to_follow_up_copy.reports_card_subtitle"),
+                                    "Transferred out" => t("transferred_out_patients_copy")} %>
         </div>
         <p class="mb-0px c-grey-dark mr-lg-12px">
           Hypertension patients who have died, been lost to follow-up, and transferred out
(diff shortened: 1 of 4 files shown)


rsanheim (approved):
LGTM

app/views/reports/regions/_patient_breakdown_charts.html.erb
```diff
@@ -74,14 +74,16 @@
...
           <%= render "definition_tooltip",
                      definitions: { "Patients under care" => t("patients_under_care_copy"),
-                                    "Lost to follow-up" => t("lost_to_follow_up_copy.reports_card_subtitle") } %>
+                                    "Lost to follow-up" => t("lost_to_follow_up_copy.reports_card_subtitle"),
+                                    "Transferred out" => t("transferred_out_patients_copy")} %>
         </div>
-        <p class="mb-0px c-grey-dark mr-lg-12px"><%= number_with_delimiter(@chart_data[:patient_breakdown][:total_patients]) %>
-          hypertension patients assigned to <%= @region.name %> as of <%= @chart_data[:ltfu_trend].last_value(:period_info)["bp_control_end_date"] %></p>
+        <p class="mb-0px c-grey-dark mr-lg-12px">
+          Hypertension patients who have died, been lost to follow-up, and transferred out
```
kpethtel, line 85: I would say "or transferred out" instead of "and". Otherwise it sounds like they did all three, which would be impressive.

kpethtel (approved):

app/views/reports/regions/_patient_breakdown_charts.html.erb
claudiovallejo: Good call, thanks @kpethtel !

PR #2456, opened 2021-04-28
Patient import

harimohanraj89:
**Story card:** [ch3325](https://app.clubhouse.io/simpledotorg/story/3325/implement-patient-import)

## Because

We want to be able to import patient data in Ethiopia for facilities that are using the hypertension tracker spreadsheet. The import will let them start using Simple instead of the spreadsheets without losing the patient data they've collected so far.

## This addresses

A patient import page in the dashboard that accepts a spreadsheet in a standard format, and imports the patient data within. Only accessible to power users.

## Implementation details

This PR re-uses a lot of the same tooling as our API to ensure that the data is consistent. 
* API Validators
* API Transformers

Additionally, a few moving parts have been introduced in the `PatientImport` namespace:
* `SpreadsheetTransformer`: Parse the uploaded spreadsheet into standard-format params
* `Validator`: Validate parsed info using API validators, surface errors
* `Importer`: Import the parsed info using API transformers and associated services
* `ImportUser`: Exposes a standard bot user that performs the import

## Test instructions

## Todo

* [x] Detect existing patients with matching business identifier, and merge them
* [x] Add navigation to the patient import screen from dashboard menu, and from facility management page
* [x] Render success and failure messaging
* [x] Audit trail for imported patients
* [x] Test all the things


app/controllers/admin/patient_imports_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  def create
+    authorize { current_admin.power_user? }
+
+    facility = Facility.find(params[:facility_id])
+
+    data = read_xlsx_or_csv_file(params[:patient_import_file])
+    params = PatientImport::SpreadsheetTransformer.call(data, facility: facility)
+    errors = PatientImport::Validator.new(params).errors
+
+    if errors.values.flatten.any?
```
prabhanshuguptagit, line 17: Will be neater if we could expose this from the validator:

`PatientImport::Validator.new(params).valid?`

spec/controllers/admin/patient_imports_controller_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+        }.join(", ")
+
+        expect(basic_patient_1_bps).to eq("160/90, 145/89")
+
+        # Patient 1 prescription drugs
+        basic_patient_1_prescription_drugs = basic_patient_1.prescription_drugs.map { |pd|
+          "#{pd.name} #{pd.dosage}"
+        }.join(", ")
+
+        expect(basic_patient_1_prescription_drugs).to eq("Amlodipine 10 mg")
```
prabhanshuguptagit, line 86: Nice spec😎

prabhanshuguptagit (commented):
Looking good ✨ . Do we have an approach for idempotent updates to BPs etc? 

I'll test more and post updates.

app/controllers/admin/patient_imports_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  include FileUploadable
+
+  def new
+    authorize { current_admin.power_user? }
+  end
+
+  def create
+    authorize { current_admin.power_user? }
+
+    facility = Facility.find(params[:facility_id])
```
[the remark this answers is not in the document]
harimohanraj89: Yup absolutely. The facility's selected from the dropdown, so if it doesn't exist, something terrible has gone wrong and this should definitely error out.

app/services/patient_import/import_user.rb
```diff
@@ -0,0 +1,36 @@
+class PatientImport::ImportUser
+  IMPORT_USER_PHONE_NUMBER = "0000000001"
+
+  def self.find_or_create
+    find || create
```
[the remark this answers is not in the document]
harimohanraj89: Lol, right?

app/services/patient_import/import_user.rb
```diff
@@ -0,0 +1,36 @@
...
+    user = User.new(
+      full_name: "import-user",
+      organization_id: Organization.take,
+      device_created_at: Time.current,
+      device_updated_at: Time.current
+    )
+
+    phone_number_authentication = PhoneNumberAuthentication.new(
+      phone_number: IMPORT_USER_PHONE_NUMBER,
+      password: "#{rand(10)}#{rand(10)}#{rand(10)}#{rand(10)}",
```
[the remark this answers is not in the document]
harimohanraj89: This loses out on password < 1000. Do you think that's ok?

app/services/patient_import/spreadsheet_transformer.rb
```diff
@@ -0,0 +1,304 @@
...
+      patient_id = patient_id(row)
+      business_identifier_id = SecureRandom.uuid
+      phone_number_id = SecureRandom.uuid
+      address_id = SecureRandom.uuid
+      medical_history_id = SecureRandom.uuid
+
+      {
+        patient: {
+          id: patient_id,
+          registration_facility_id: registration_facility_id,
```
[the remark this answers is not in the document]
harimohanraj89: Is it being set to `nil` right now?! 😱 Yup, gotta do that.

spec/controllers/admin/patient_imports_controller_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+    context "with invalid file" do
+      it "raises validation errors" do
+        admin = create(:admin, :power_user)
+        sign_in(admin.email_authentication)
+
+        facility = create(:facility)
+        patient_import_file = fixture_file_upload("files/patient_import_invalid_test.xlsx", "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")
+        params = {patient_import_file: patient_import_file, facility_id: facility.id}
+
+        expect { post :create, params: params }.not_to change { Patient.count }
```
rsanheim, line 111: Would probably be good to verify we get some sort of helpful error message here as well...how much detail will the user see about why the import failed?

app/services/patient_import/spreadsheet_transformer.rb
```diff
@@ -0,0 +1,304 @@
...
+    def patient_status(row)
+      row[:died] == "yes" ? :dead : :active
+    end
+
+
+
+    def medication(name:, patient_id:, created_at:)
+      medication_record = medication_by_name_dosage_and_frequency(name) || medication_by_name_and_dosage(name)
+
+      return nil unless medication_record.present?
```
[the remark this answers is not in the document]
harimohanraj89: IT's not as bad as it sounds, because the XLS files we hand out have validations in these data fields. It's also incredibly easy to use if it's a single field, instead of separate fields for each ☝️ 

This line actually catches cells that still have `None` in them 😅 and also, yes, ignores any invalid data.

app/services/patient_import/spreadsheet_transformer.rb
```diff
@@ -0,0 +1,304 @@
...
+      PatientBusinessIdentifier.find_by(
+        identifier_type: row[:identifier_type],
+        identifier: row[:identifier]
+      )&.patient_id || SecureRandom.uuid
+    end
+
+    def business_identifier_metadata_version(row)
+      case row[:identifier_type]
+      when "simple_bp_passport"
+        "org.simple.bppassport.meta.v1"
```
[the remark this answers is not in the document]
harimohanraj89: Yeah, the BP passport meta tag is not `simple_bp_passport` 😭

spec/services/patient_import/importer_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        request_metadata: {
+          request_facility_id: facility.id,
+          request_user_id: PatientImport::ImportUser.find_or_create.id
+        }
+      ).and_call_original
+      expect(importer).to receive(:merge_encounter_observation).with(:blood_pressures, hash_including(property: "bp1 property"))
+      expect(importer).to receive(:merge_encounter_observation).with(:blood_pressures, hash_including(property: "bp2 property"))
+      expect(MedicalHistory).to receive(:merge).with(hash_including(property: "medical history property"))
+      expect(PrescriptionDrug).to receive(:merge).with(hash_including(property: "pd1 property"))
+      expect(PrescriptionDrug).to receive(:merge).with(hash_including(property: "pd2 property"))
```
rsanheim, line 43: Its not clear to me that this spec provides value, especially considering how brittle it will be with this many mocks. Any changes to the above objects or call sequence will result in this spec failing.

Is there a way we could test this more from an outside-in perspective, with less mocking? Perhaps verify that the merge creates the correct output after running the import, with a few patients where two should be merged and one should not?

db/migrate/20210428092832_create_patient_import_logs.rb
```diff
@@ -0,0 +1,13 @@
+class CreatePatientImportLogs < ActiveRecord::Migration[5.2]
+  def change
+    create_table :patient_import_logs, id: :uuid do |t|
+      t.belongs_to :user, null: false, type: :uuid
+
+      t.uuid :record_id
+      t.string :record_type
```
rsanheim, line 7: If I'm reading things right, this log will be basically be capturing:

```
user x imported patient Jane Doe
user x imported BP 1 for Jane Doe
user x imported rx drug for Jane Doe
```

and on and on for possibly thousands of records, correct?  

Is there a reason we can't send this data dog instead?  It looks like it would only be capturing UUIDs, a class name, and timestamps...do we really need to store this in the database for (basically) permeant storage?

app/services/patient_import/spreadsheet_transformer.rb
```diff
@@ -0,0 +1,304 @@
...
+    def localized_frequency(medication)
+      return nil unless medication.frequency.present?
+      I18n.translate("helpers.label.drug.frequency.#{medication.frequency}", locale: CountryConfig.current[:dashboard_locale])
+    end
+
+    def medication_by_name_and_dosage(name)
+      return unless name.present?
+
+      Medication.all.to_a.find do |medication|
+        name.gsub(/\s+/, "") == "#{medication.name}#{medication.dosage}".gsub(/\s+/, "")
```
rsanheim, line 286: Who is going to make sure medications in the spreadsheet actually match the values we have in our dataset?  This seems like it could be pretty difficult to keep in sync.

app/services/patient_import/spreadsheet_transformer.rb
```diff
@@ -0,0 +1,304 @@
...
+
+      user.phone_number_authentications = [phone_number_authentication]
+      user.sync_approval_denied("bot user for import")
+      user.save!
+
+      user
+    end
+
+    def patient_status(row)
+      row[:died] == "yes" ? :dead : :active
```
rsanheim, line 249: This should probably use the enums themselves from the Patient object to ensure things stay in sync.

app/services/patient_import/importer.rb
```diff
@@ -0,0 +1,80 @@
...
+        import_medical_history(medical_history_params)
+        import_prescription_drugs(prescription_drugs_params)
+
+        results.push(
+          patient: patient_params,
+          medical_history: medical_history_params,
+          blood_pressures: blood_pressures_params,
+          prescription_drugs: prescription_drugs_params
+        )
+      end
```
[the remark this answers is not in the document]
rsanheim: Same question as @ssrihari - what are the expected sizes for import spreadsheets? We probably need to batch up rows if is anything over a few hundred, given the amount of work going on here.  One malformed cell will also cause the entire transaction blow up, so error handling will be tricky. 😟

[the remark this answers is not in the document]
rsanheim: Actually, will failed results from all the `merge` calls result in a rollback at all? If we end up just doing a `save` or `update` down the stack, I think things will silently fail.

app/controllers/admin/patient_imports_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    data = read_xlsx_or_csv_file(params[:patient_import_file])
+    params = PatientImport::SpreadsheetTransformer.call(data, facility: facility)
+    errors = PatientImport::Validator.new(params).errors
+
+    if errors.values.flatten.any?
+      @errors = errors
+      render :new
+    else
+      results = PatientImport::Importer.call(params: params, facility: facility, admin: current_admin)
+      redirect_to new_admin_patient_import_url, notice: "Successfully imported #{results.count} patients to #{facility.name}"
```
rsanheim, line 22: It may be good to also display the number of total rows from the spreadsheet, especially so the user running things will see how many rows got skipped / ignored due to any errors.

app/controllers/admin/patient_imports_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    data = read_xlsx_or_csv_file(params[:patient_import_file])
+    params = PatientImport::SpreadsheetTransformer.call(data, facility: facility)
+    errors = PatientImport::Validator.new(params).errors
+
+    if errors.values.flatten.any?
+      @errors = errors
+      render :new
+    else
+      results = PatientImport::Importer.call(params: params, facility: facility, admin: current_admin)
```
rsanheim, line 21: Consider moving to a background job, and email when done? This is going to block a web request and could very easily time out for any large spreadsheet.

rsanheim (commented):
Some notes below.

I have two higher level concerns:

* who is going to make sure the spreadsheet matches the expected format? There are a lot of assumptions about the format of the data, and getting things massaged correctly could be a lot of work.
* What is the purpose of the audit trail ? Is it primarily to debug any errors when importing? If so, I think datadog would be a better target for that log.

db/migrate/20210428092832_create_patient_import_logs.rb
prabhanshuguptagit: We can log these I think. Although datadog lets us search through only the last month.

app/services/patient_import/importer.rb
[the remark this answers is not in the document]
harimohanraj89: There are 100-200 patients per spreadsheet.

The two options to commit to are:
1. All or nothing
1. Import what you can, report the failures well

The validations that we are doing before import should in theory take care of all problems with the data itself. How, then, can the import fail? Things like timeouts, network blips to DB, and of course edge cases we haven't thought about. I've just confirmed that Rob's suspicion about `merge` failing silently is true. In that case, I'm starting to lean towards approach 2.

Only power users (us) are going to be using this, so I'd like to prep each spreadsheet, upload it to ET Demo, verify if any records failed and why, and then upload to prod.

Additionally, we should be mindful not to open up this dashboard page to other users. We're making a lot of technical decisions here that are hinged upon the assumption that this tool is for our eyes only.

spec/controllers/admin/patient_imports_controller_spec.rb
harimohanraj89: That's a good idea. I could use `assigns` but I recall that being discouraged. Is that the case? If so, I can assert something on the response body itself.

app/controllers/admin/patient_imports_controller.rb
harimohanraj89: Oof, I like that a log.

app/services/patient_import/import_user.rb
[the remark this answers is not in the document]
prabhanshuguptagit: Ah right. Should be ok. Otherwise we can move this into a method.

app/controllers/admin/patient_imports_controller.rb
harimohanraj89: That sounds like a good idea. 150 patients was taking over 30 seconds to process.

I will also disable retries on the job, since the import is not designed to be idempotent.

app/controllers/admin/patient_imports_controller.rb
harimohanraj89: Great idea. I will make the `results` more useful. We should know:
* How many total patients I just tried to import
* How many were imported
* Which rows were not imported.

This will also be easier to craft when it's an email via background job 😉

app/services/patient_import/import_user.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def self.find
+    PhoneNumberAuthentication.find_by(phone_number: IMPORT_USER_PHONE_NUMBER)&.user
+  end
+
+  def self.create
+    user = User.new(
+      full_name: "import-user",
+      organization_id: Organization.take,
```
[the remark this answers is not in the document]
harimohanraj89: Looks like the difference is in ordering. `first` is deterministic. Sure, sounds good!

app/services/patient_import/import_user.rb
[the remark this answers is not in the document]
harimohanraj89: Extracted to method.

app/services/patient_import/spreadsheet_transformer.rb
```diff
@@ -0,0 +1,304 @@
...
+
+    def registration_facility_id
+      facility.id
+    end
+
+    def import_user
+      @import_user = PatientImport::ImportUser.find_or_create
+    end
+
+    def create_import_user
```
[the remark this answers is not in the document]
harimohanraj89: Yup. Extracted to `ImportUser` but forgot to delete here.

app/services/patient_import/spreadsheet_transformer.rb
harimohanraj89: It's not great, but we (Meg primarily) manages the spreadsheet that we share to facilities. The medication fields have strict Excel data validations on them. If we update our `medications.yml` list, we need to notify Meg.

Is there an automatic Github thing we can set up to remind us to do this?

spec/services/patient_import/importer_spec.rb
harimohanraj89: Yeah we can switch this spec to be more integration-y.

app/models/patient_import_log.rb
```diff
@@ -0,0 +1,4 @@
+class PatientImportLog < ApplicationRecord
+  belongs_to :user
+  belongs_to :record, polymorphic: true
+end
```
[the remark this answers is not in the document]
harimohanraj89: Writing this to our existing `AuditLog`s now. It's quite neat.

app/services/patient_import/spreadsheet_transformer.rb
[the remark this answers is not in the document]
harimohanraj89: Deleted.

db/schema.rb
```diff
@@ -1310,4 +1320,104 @@
   add_index "latest_blood_pressures_per_patients", ["bp_id"], name: "index_latest_blood_pressures_per_patients", unique: true
   add_index "latest_blood_pressures_per_patients", ["patient_id"], name: "index_latest_bp_per_patient_patient_id"
 
+  create_view "calendar_months", sql_definition: <<-SQL
```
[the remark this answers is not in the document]
harimohanraj89: Ripped out.

harimohanraj89:
@prabhanshuguptagit We're _not_ addressing idempotent BPs, we're stopping at idempotent patient records by business ID. BPs, and drugs, etc, are a slippery slope, and we're choosing to avoid supporting them in favor of the dev who's doing the import verifying which records made it, and which ones didn't.

app/services/patient_import/importer.rb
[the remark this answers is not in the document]
harimohanraj89: I've leaned into this idea with my latest commits.
* Every patient gets a separate job queued
* Every patient gets its operations wrapped in a DB transaction inside the job - This way we never import a patients BPs, but not their drugs, for example
* If a patient import fails because of transient issues, the job will retry and succeed
* If a patient import fails because the data is bad, and somehow got through our validations, then the job will stick around in Sidekiq, and the dev who ran the import can inspect the failed jobs there.

I think this is the simplest approach assuming that this is a _developer-only_ tool. What do you think?

app/services/patient_import/spreadsheet_transformer.rb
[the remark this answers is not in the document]
rsanheim: Cool, good to know.

spec/controllers/admin/patient_imports_controller_spec.rb
rsanheim: I think using either an `@errors` ivar or a `flash[:error]` would be fine.

app/controllers/admin/patient_imports_controller.rb
rsanheim: 👍

app/services/patient_import/spreadsheet_transformer.rb
rsanheim: If Meg gets a GitHub account, we could add that specific file to the CODEOWNERS file I've started over in https://github.com/simpledotorg/simple-server/pull/2467.  

Or we could have someone else as the 'owner' of that file who would ensure Meg gets a heads up when it is changed.

db/migrate/20210428092832_create_patient_import_logs.rb
rsanheim: > We can log these I think. Although datadog lets us search through only the last month.

I like the sound of that ...I'm assuming a month of history would be fine, as this would be debugged right after running?

app/services/patient_import/importer.rb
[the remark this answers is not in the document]
rsanheim: This sounds like a great approach to me, and will be much easier to troubleshoot when things come up.

db/migrate/20210428092832_create_patient_import_logs.rb
harimohanraj89: We're only capturing these logs for audit compliance. As such, I've changed it up to re-use the existing `AuditLog`  mechanism we have in place. Imported records will be reported in the audit logs as an "import" by ~"import bot user"~ "admin who uploaded the spreadsheet"

app/services/patient_import/spreadsheet_transformer.rb
harimohanraj89: I'll add us to the codeowners, and a comment to the YML file with instructions on who to notify.

app/controllers/admin/patient_imports_controller.rb
harimohanraj89: This point is now moot, since every patient gets their own job, and we'll be inspecting success of the import directly in Sidekiq.

app/controllers/admin/patient_imports_controller.rb
harimohanraj89: This point is now moot, since every patient gets their own job. 👍

spec/controllers/admin/patient_imports_controller_spec.rb
harimohanraj89: Donezo.

spec/services/patient_import/importer_spec.rb
harimohanraj89: Donezo.

app/services/patient_import/spreadsheet_transformer.rb
harimohanraj89: I've added a comment to the `medications.yml` file with instructions, and I'll add the file itself to the codeowners list in #2467 directly.

app/jobs/patient_import_job.rb
```diff
@@ -0,0 +1,5 @@
+class PatientImportJob < ApplicationJob
+  def perform(params:, facility:, admin:)
+    PatientImport::Importer.new(params: params, facility: facility, admin: admin).import
```
vkrmis, line 3: Do you think its worth adding a log here with `facility_id`, `patient_id`, and a param to indicate success/failure?
This will help us look up aggregate data for an input file via datadog.

vkrmis (approved):

app/services/patient_import/importer.rb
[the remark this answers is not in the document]
ssrihari: Agreed, this looks good! Some additional notes/observation:

>  merge failing silently

This seems like something we need to address regardless of which approach we take, ya?

> upload it to ET Demo

Um, would we upload real PHI to another environment? Maybe we can do it locally instead?

> We're making a lot of technical decisions here that are hinged upon the assumption that this tool is for our eyes only.

Let's call this out in a comment next to the permissions code?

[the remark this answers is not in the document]
ssrihari: Actually, if only power users should do this, should we just put it in a rake task? That way it's obvious we shouldn't (can't) allow others to use it.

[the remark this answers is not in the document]
ssrihari: Ah sorry I just remembered you mentioned convenience in this case is important, nvm re rake task. ✌️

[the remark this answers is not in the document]
harimohanraj89: Exactly re: rake task 😄 

Regarding PHI in a Demo environment, you're right. Demo might not be the right place, but a production-like environment will be important. I've been using the Heroku review app until now, which is great as it will die soon. Regarding sensitivity, we've definitely leaned towards "more secure, less aggressive testing" int he past and it bit us. I think the value of testing the actual PHI in, say, a review app is worth the extra exposure.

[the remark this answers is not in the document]
harimohanraj89: Re silent merges, yes. I think this is a hot topic for "sync api TLC". I'll add a note to https://app.clubhouse.io/simpledotorg/epic/3366/revisit-sync?ct_workflow=all

app/jobs/patient_import_job.rb
harimohanraj89: We're sticking to verification of aggregates by inspecting sidekiq queues and using the console. I think this should suffice for now. 

Actually, we're raising exceptions on failure, so we could confidently log successes along with audit logs. Yup, may be worth it to avoid having to trawl through audit logs to inspect import activity. Will do.

PR #2457, opened 2021-04-28
Disable reminders in the worker as well

rsanheim:
We want to disable closer to the source as well for appointment reminders, otherwise we could have cases where Twilio callbacks or re-running jobs still end up sending reminders.

### DIFF
--- a/spec/jobs/appointment_notification/worker_spec.rb
+++ b/spec/jobs/appointment_notification/worker_spec.rb
@@ -17,6 +17,7 @@ RSpec.describe AppointmentNotification::Worker, type: :job do
   let(:callback_url) { "https://localhost/api/v3/twilio_sms_delivery" }
 
   before do
+    Flipper.enable(:appointment_reminders)
     notification_response = double("NotificationServiceResponse")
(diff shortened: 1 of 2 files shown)


spec/jobs/appointment_notification/worker_spec.rb
```diff
@@ -17,6 +17,7 @@
   let(:callback_url) { "https://localhost/api/v3/twilio_sms_delivery" }
 
   before do
+    Flipper.enable(:appointment_reminders)
```
kpethtel, line 20: Do you need to disable it after the test?

kpethtel (approved):

spec/jobs/appointment_notification/worker_spec.rb
harimohanraj89: We use the DB for flipper store, so I think the flipper changes get rolled back after every spec.

ssrihari: The db tx rollback would take care of this, I think. I'm merging this now so we don't send out reminders today.

rsanheim: They are reset for every spec, as flipper is part of database state.

PR #2458, opened 2021-04-29
3327 who report

kpethtel:
**Story card:** [ch3327](https://app.clubhouse.io/simpledotorg/epic/3327)

## Because

The WHO wants reports in a specific format. 

## This addresses

This produces a csv with data for a district and its facilities. 

## Test instructions

- Go to the reports tab and navigate to a district
- click the download dropdown
- there should be a new option "Monthly district data"
- click that and accept the download
- expect it to be populated with data for the district and its facilities


app/views/reports/regions/who_report.csv.erb
```diff
@@ -0,0 +1,103 @@
+<%
```
kpethtel, line 1: I'm not sure if this being a view brings any value and would prefer it be in a ruby file, but I've followed the existing pattern.

app/controllers/reports/regions_controller.rb
```diff
@@ -86,16 +86,20 @@ def download
       raise ArgumentError, "Invalid Period #{@period} #{@period.inspect}"
     end
 
-    @cohort_analytics = @region.cohort_analytics(period: @period.type, prev_periods: 6)
     @dashboard_analytics = @region.dashboard_analytics(period: @period.type, prev_periods: 6)
 
     respond_to do |format|
       format.csv do
-        if @region.district_region?
+        if params[:type] == "who_report"
```
kpethtel, line 93: The fact that I needed to add a new parameter to the payload may indicate that we should just start a new endpoint for this report.

app/controllers/reports/regions_controller.rb
```diff
@@ -206,4 +210,36 @@ def percentage(numerator, denominator)
     return 0 if denominator == 0 || numerator == 0
     ((numerator.to_f / denominator) * 100).round(2)
   end
+
+  def set_data_for_who_report
+    @months = @period.downto(5).reverse
+    repo = Reports::Repository.new(@region, periods: @period)
+    slug = @region.slug
+    @region_data = Hash.new(0)
```
kpethtel, line 218: I may break this into a function to avoid the repetition.

app/controllers/reports/regions_controller.rb
rsanheim: Yeah, this should definitely be its own endpoint. I would suggest its own controller entirely, given this is focused on a making CSV downloads for the WHO.

app/controllers/reports/regions_controller.rb
```diff
@@ -206,4 +210,36 @@ def percentage(numerator, denominator)
     return 0 if denominator == 0 || numerator == 0
     ((numerator.to_f / denominator) * 100).round(2)
   end
+
+  def set_data_for_who_report
+    @months = @period.downto(5).reverse
+    repo = Reports::Repository.new(@region, periods: @period)
```
rsanheim, line 216: You can use the same repo for all the data here by doing:

```
regions = @region.facility_regions << @region
Repository.new(regions, periods: @period)
```

that will also reuse the same memorized cached, so better perf.

app/views/reports/regions/who_report.csv.erb
```diff
@@ -0,0 +1,103 @@
...
+    "Visited but BP not taken",
+    "Amlodipine",
+    "ARBs/ACE Inhibitors",
+    "Diuretic"
+  ].flatten
+
+  # DISTRICT ROW
+  dead = Patient.where(assigned_facility_id: @region.facilities.pluck(:id), status: "dead").count
+  registered_by_month = @months.map do |month|
+    @dashboard_analytics.sum { |_, data| data.dig(:registered_patients_by_period, month.value) || 0 }
```
rsanheim, line 36: You shouldn't need to use any of these numbers from `dashboard_analytics` ...these can all come from the repository numbers.

app/views/reports/regions/who_report.csv.erb
```diff
@@ -0,0 +1,103 @@
...
+  end
+  follow_up_by_month = @months.map do |month|
+    @dashboard_analytics.sum { |_, data| data.dig(:follow_up_patients_by_period, month.value) || 0 }
```
rsanheim, line 39: I'm working on adding follow up numbers to the Repo in https://github.com/simpledotorg/simple-server/pull/2436, so hopefully these can use the repo too soon.

rsanheim (commented):
Some initial thoughts.

app/controllers/reports/regions_controller.rb
kpethtel: Ah, thanks for the tip. I tried something similar but failed because the child facilities were an active record relation.

app/views/reports/regions/who_report.csv.erb
kpethtel: I see that now that you say it. I did the dashboard_analytics part first and didn't think about it again after.

app/views/reports/regions/who_report.csv.erb
```diff
@@ -0,0 +1,103 @@
...
+    end
+
+    matching_facility = @facilities_data.find{ |fac| fac[:region] == facility }
+    csv << [
+      index + 1,
+      @region.name,
+      facility.name,
+      facility.source.facility_type,
+      facility.source.block,
+      facility.source.blood_pressures.any?,
```
claudiovallejo, line 84: @kpethtel can we show "Active" if true and "Inactive" if false?

claudiovallejo (commented):
@kpethtel noticed that "Total patients under Care" in the spreadsheet doesn't match the number in the Dashboard

image

app/views/reports/regions/who_report.csv.erb
```diff
@@ -0,0 +1,103 @@
+<%
+  csv = CSV.generate(headers: true) do |csv|
+  csv << ["Facility Report #{@period.to_date.strftime("%B %Y")}"]
```
claudiovallejo, line 3: @kpethtel is it possible to add a new row, `section_labels` between the `title` and `month_labels` rows? Found a potential solution ([see in Stackoverflow](https://stackoverflow.com/questions/6156600/in-ruby-csv-how-to-write-a-blank-instead-of-to-a-file)):

```
section_labels = [
  nil,
  nil,
  nil,
  nil,
  nil,
  nil,
  nil,
  nil,
  nil,
  nil,
  nil,
  "New Registrations",
  nil,
  nil,
  nil,
  nil,
  nil,
  "Follow Up",
  nil,
  nil,
  nil,
  nil,
  nil,
  "Cumulative treatment outcome among patient under care",
  "Drug availability",
  nil,
  nil,
  nil
]

**Review by claudiovallejo — COMMENTED** [review]
@kpethtel also noticed that "Died" isn't working for facilities

image

File: app/views/reports/regions/who_report.csv.erb (hunk 7 above)
Comment (kpethtel) [reply app/views/reports/regions/who_report.csv.erb:84]: Yes. Will do.

**Comment by kpethtel** [comment]
> @kpethtel noticed that "Total patients under Care" in the spreadsheet doesn't match the number in the Dashboard
> 
> image

I'll try to sort that out.

**Comment by kpethtel** [comment]
> @kpethtel also noticed that "Died" isn't working for facilities
> 
> image

I'll try to fix that too.

File: app/views/reports/regions/who_report.csv.erb (hunk 8 above)
Comment (kpethtel) [reply app/views/reports/regions/who_report.csv.erb:3]: Yes, definitely. I don't think it will format correctly, but I'll definitely add a row there.

**Comment by claudiovallejo** [comment]
@kpethtel the "Total Patients under Care", "New Registration", and "Follow Up" fields aren't working for facilities

File: app/services/who_report_service.rb [hunk 9]
```diff
@@ -0,0 +1,148 @@
...
+  def header_row
+    month_labels = months.map { |month| month.value.strftime("%b-%Y") }
+    [
+      "#",
+      "District",
+      "Facility",
+      "Facility type",
+      "Block",
+      "Active/Inactive (Inactive facilities have 0 BP measures taken)",
+      "Estimated hypertension opulation",
```
claudiovallejo, line 49: @kpethtel small typo - should be "population"

app/services/who_report_service.rb
```diff
@@ -0,0 +1,148 @@
...
+      "Total assigned patients",
+      "Total lost to follow-up",
+      "Died",
+      "Total patients under care",
+      month_labels,
+      month_labels,
+      "BP controlled %",
+      "BP not controlled %",
+      "Missed visits %",
+      "Visit but no BP taken %",
```
claudiovallejo, line 60: @kpethtel since these fields will now show patient numbers, let's rename these to:
* Patients with BP controlled
* Patients with BP not controlled
* Patients with a missed visit
* Patients with a visit but no BP taken

I've updated the Epic's description :)

app/services/who_report_service.rb
```diff
@@ -0,0 +1,148 @@
...
+      "Facility",
+      "Facility type",
+      "Block",
+      "Active/Inactive (Inactive facilities have 0 BP measures taken)",
+      "Estimated hypertension opulation",
+      "Total registrations",
+      "Total assigned patients",
+      "Total lost to follow-up",
+      "Died",
+      "Total patients under care",
```
claudiovallejo, line 54: @kpethtel let's update to "Patients under care". I just noticed we don't include "Total" in the Dashboard. Sorry, this was my bad :D

app/services/who_report_service.rb
```diff
@@ -0,0 +1,148 @@
...
+      "#",
+      "District",
+      "Facility",
+      "Facility type",
+      "Block",
+      "Active/Inactive (Inactive facilities have 0 BP measures taken)",
+      "Estimated hypertension opulation",
+      "Total registrations",
+      "Total assigned patients",
+      "Total lost to follow-up",
```
claudiovallejo, line 52: @kpethtel can we rename to "Patients lost to follow-up"? It's a bit clearer (I've updated the Epic's description)

claudiovallejo (commented):
@kpethtel changes look great overall and thanks for adding the new row with column sub-headers 👌 

The numbers in the "Treatment outcomes of patients under care" don't reflect what's on the Dashboard (showing 0s in the spreadsheet).

Also, just as a reminder, the "Treatment outcomes of patients under care" should display patient numbers and not percentages.

app/services/who_report_service.rb
```diff
@@ -0,0 +1,127 @@
+class WhoReportService
+  attr_reader :region, :period, :months, :repo, :dashboard_analytics
+  def initialize(region, period)
+    @region = region
+    @period = period
+    @months = period.downto(5).reverse
+    regions = region.facility_regions.to_a << region
```
rsanheim, line 7: I would suggest teaching FacilityDistrict how to return `facility_regions` to match so this will work...that will it will match the "Region-ish" interface.

```
class FacilityDistrict
  def facility_regions
    Region.facility_regions.where(source_id: facility_Ids)
  end
```

app/controllers/reports/regions_controller.rb
```diff
@@ -101,6 +101,29 @@ def download
...
+    @period = Period.new(type: params[:period], value: Date.current)
+    unless @period.valid?
+      raise ArgumentError, "Invalid Period #{@period} #{@period.inspect}"
+    end
+    unless @period.month?
+      raise ArgumentError, "Period must be month"
+    end
+
+    csv = WhoReportService.new(@region, @period).report
+    report_date = @period.to_date.strftime("%B-%Y").downcase
```
rsanheim, line 117: You can use our custom [date/time format](https://github.com/simpledotorg/simple-server/blob/master/config/initializers/time_and_date_formats.rb) for this.

```
  @period.to_s(:month_year).downcase
```

app/services/who_report_service.rb
kpethtel: Sounds good 👍

app/models/facility_district.rb
```diff
@@ -28,6 +28,14 @@ def reportable_children
     children
   end
 
+  def facility_regions
+    # this should really be Region.facility_regions.where(source_id: facility_ids)
+    # change while fixing https://app.clubhouse.io/simpledotorg/story/3380/facilitydistrict-should-be-initialized-with-name-not-slug
```
kpethtel, line 33: I don't want to fix the bug as part of this because this needs to get out soon and I don't want to risk breaking anything else.

spec/services/monthly_district_data_service_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  describe "#report" do
+    it "produces valid csv data" do
+      result = service.report
+      expect {
+        CSV.parse(result)
+      }.not_to raise_error
+    end
+
+    it "provides accurate numbers for the district" do
```
kpethtel, line 26: These are poor specs. I got hit by another last minute issue and the data setup for these tests would have taken me a while to figure out.

app/services/monthly_district_data_service.rb
```diff
@@ -0,0 +1,126 @@
...
+      patients.status_dead.count,
+      patients.excluding_dead.not_ltfu_as_of(period.end).count,
+      registered_by_month,
+      follow_up_by_month,
+      repo.controlled_patients_count.dig(region.slug, period) || 0,
+      repo.uncontrolled_patients_count.dig(region.slug, period) || 0,
+      repo.missed_visits.dig(region.slug, period) || 0,
+      repo.visited_without_bp_taken.dig(region.slug, period) || 0,
+      Array.new(3, nil)
+    ].flatten
```
ssrihari, line 91: I normally prefer to model these as hashes because it's easier to read key->value than understand the sequence in which the values appear. Less chances of human errors that way.

We could factor in the sequence of columns and their flattening when writing to file with `values_at` or something.

spec/services/monthly_district_data_service_spec.rb
ssrihari: Related to the [previous comment](https://github.com/simpledotorg/simple-server/pull/2458/files#r625529751), I would suggest having an intermediate representation of the report in a hash that we can test easily for correctness. The CSV test could then just ensure that the value/header mapping is correct.

spec/services/monthly_district_data_service_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      expect(find_in_csv(csv, facility_row_index, "Total assigned patients")).to eq("1")
+      expect(find_in_csv(csv, facility_row_index, "Lost to follow-up patients")).to eq("0")
+      expect(find_in_csv(csv, facility_row_index, "Died all-time")).to eq("0")
+      expect(find_in_csv(csv, facility_row_index, "Patients under care")).to eq("1")
+      expect(csv[facility_row_index][12..24].uniq).to eq(["0"])
+      expect(find_in_csv(csv, facility_row_index, "Patients with BP controlled")).to eq("0")
+      expect(find_in_csv(csv, facility_row_index, "Patients with BP not controlled")).to eq("0")
+      expect(find_in_csv(csv, facility_row_index, "Patients with a missed visits")).to eq("0")
+      expect(find_in_csv(csv, facility_row_index, "Patients with a visit but no BP taken")).to eq("0")
+      expect(csv[facility_row_index][30..31].uniq).to eq([nil])
```
ssrihari, line 69: In addition to checking the correctness of individual values, I think it would be valuable to start adding the "property tests" we spoke about recently. It would help ensuring that the data in the reports is integral.

Some example properties for the report here:
- Patients under care = "Patients with BP controlled" + "Patients with BP not controlled" + "Patients with a missed visits" + "Patients with a visit but no BP taken"
- Registrations = Patients under care + ltfu + dead

app/controllers/reports/regions_controller.rb
```diff
@@ -101,6 +101,34 @@ def download
     end
   end
 
+  def monthly_district_data_report
+    # re-implementing part of the find_region method with a modification as a temporary workaround for this bug:
+    # https://app.clubhouse.io/simpledotorg/story/3380/facilitydistrict-should-be-initialized-with-name-not-slug
```
ssrihari, line 106: I'm guessing we're adding a workaround now because of time constraints. Just noting that we should be fixing the root cause soon. Such workarounds have a tendency to stick around for a long time despite our best intentions.

kpethtel: I intend to fix it. I don't want to fix it as part of this PR because of time constraints and because I think it would be better to do it separately rather than roll it in with unrelated work. It's easier to track and to revert that way.

app/services/monthly_district_data_service.rb
kpethtel: That's not a bad idea. I suppose you just take the values from the hash at the end.

PR #2461, opened 2021-05-03
WHO - report FacilityDistricts

rsanheim:
/cc @kpethtel


app/models/facility_district.rb
```diff
@@ -17,6 +17,10 @@ def facility_ids
     facilities.pluck("facilities.id")
   end
 
+  def facility_regions
+    Region.facility_regions.where(source_id: facility_ids)
+  end
```
ssrihari, line 22: Where is this function used?

rsanheim:
Not sure this is needed anymore.

PR #2463, opened 2021-05-04
Use locale to translate 'facility group' in UI

timcheadle:
**Story card:** [ch3223](https://app.clubhouse.io/simpledotorg/story/3223/update-new-facility-group-edit-facility-group-view)

## Because

Users don't know what a "facility group" is, so we should show them what they represent (district, zone, upazila, etc.)

## This addresses

This localizes a "facility group" and updates all relevant views to reflect the change.

*Note*: I excluded the CSV facility upload from this because it's tricky to localize headers

## Test instructions

1. Go into `.env.development` and set `DEFAULT_COUNTRY=BD`
2. Go through the Facilities and Organizations admin views on the dashboard
3. Ensure that "facility group" shows up as "upazila" in all relevant places


config/locales/en_BD.yml
```diff
@@ -1,4 +1,5 @@
 en_BD:
+  facility_group: "upazila"
```
rsanheim, line 2: @claudiovallejo Facility groups are equivalent to districts in BD, not upazillas, correct?

dburka: @rsanheim You shared this matrix on Slack, which looks correct to me.

```
| codebase Name    | India        | Bangladesh   | Ethiopia     | Generic name
|—————————+———————+———————+———————|
| Organization     | Organization | Organization | Organization | Level 1
| State            | State        | District     | Region       | Level 2
| District         | District     | Upazila      | Zone         | Level 3
| Block (aka Zone) | Block        | Union        | Woreda       | Level 4
| Facility         | Facility     | Facility     | Facility     | Level 5
```

app/views/admin/facilities/_form.html.erb
```diff
@@ -18,7 +18,7 @@
           required: true %>
       <small class="form-text text-muted mb-3">
         If you don't see a block in the list,
```
claudiovallejo, line 20: @timcheadle this should say "If you don't see a ${block} in the list". Block should be localized too :)

app/views/admin/facilities/index.html.erb
```diff
@@ -8,7 +8,7 @@
 
     <% if current_admin.accessible_organizations(:manage).any? || current_admin.power_user? %>
       <%= link_to new_admin_facility_group_path, class: "btn btn-sm btn-success" do %>
-        <i class="fas fa-plus mr-1"></i> Facility group
+        <i class="fas fa-plus mr-1"></i> <%= t("facility_group").capitalize %>
```
claudiovallejo, line 11: @timcheadle this would be clearer if it said "Add" before the name of the facility group. Example "+ Add Upazila" instead of "+ Upazila".

app/views/admin/facilities/index.html.erb
```diff
@@ -28,12 +28,12 @@
...
         <div class="row">
           <div class="col-8">
-            <h3><span class="heading-label">Facility group</span><%= facility_group.name %></h3>
+            <h3><span class="heading-label"><%= t("facility_group").capitalize %></span><%= facility_group.name %></h3>
           </div>
 
           <% if current_admin.accessible_facility_groups(:manage).find_by_id(facility_group) %>
             <div class="col-4 text-right">
-              <strong><%= link_to 'Edit group', edit_admin_facility_group_path(facility_group), class: "btn btn-sm btn-outline-primary" %></strong>
+              <strong><%= link_to "Edit #{t("facility_group")}", edit_admin_facility_group_path(facility_group), class: "btn btn-sm btn-outline-primary" %></strong>
```
claudiovallejo, line 36: Would suggest capitalizing the facility group to keep it consistent with the "+ Add Upazila" button at the top of the page

app/views/admin/facility_groups/_form.html.erb
```diff
@@ -56,7 +55,7 @@
                     <i class="fas fa-trash-alt text-grey"
                        data-toggle="tooltip"
                        data-placement="right"
-                       title="Contains <%= block.facility_regions.count %> facility(s)"></i>
+                       title="Cannot delete blocks with facilities"></i>
```
claudiovallejo, line 58: @timcheadle "blocks" should be localized. The title should be "Cannot delete ${blocks} with facilities".

config/locales/en_BD.yml
claudiovallejo: @rsanheim @dburka Facility groups are equivalent to Upazilas in Bangladesh. When adding a new facility to an Upazila, the user should be able to select a Union.

app/views/admin/facilities/index.html.erb
dburka: Plus means "Add". Let's keep this simple and consistent. "+ District" is good.

app/views/admin/facilities/index.html.erb
dburka: No. We should be using "Sentence case" all over the dashboard @claudiovallejo

app/views/admin/facilities/_form.html.erb
dburka: Fixing.

app/views/admin/facilities/index.html.erb
claudiovallejo: @dburka it's not consistent in other places, and "+ District" is not as clear as "+ Add district". Should we also update the "Add a facility" button to say "+ facility"?

dburka: Yes. I just updated the "+ Facility" button as well.

app/views/admin/facility_groups/_form.html.erb
dburka: Fixing by changing the copy.

config/locales/en_BD.yml
rsanheim: Thanks for the clarification, I should've just referred to that chart =).

app/views/admin/facility_groups/_form.html.erb
```diff
@@ -25,7 +24,7 @@
           organizations.pluck(:name, :id),
           {},
           {disabled: action_name == "edit"}) %>
-      <%= form.text_field :name, id: :facility_name, required: true, help: "Usually the name of a District", autocomplete: "off" %>
+      <%= form.text_field :name, id: :facility_name, required: true, autocomplete: "off" %>
```
harimohanraj89, line 27: Is the ID used? If not, we can axe/fix it. It should be `facility_group_name`, yes?

app/views/admin/facility_groups/_form.html.erb
```diff
@@ -41,7 +40,7 @@
       <div class="mt-3">
         <label><%= I18n.t("helpers.label.facility.zone").tr("* ", "").pluralize %></label>
         <div class="input-group mb-2">
-          <input type="text" class="form-control" id="new-block-name" placeholder="Add new <%= I18n.t("helpers.label.facility.zone").tr("* ", "") %>">
+          <input type="text" class="form-control" id="new-block-name" placeholder="Add new <%= I18n.t("helpers.label.facility.zone").downcase.tr("* ", "") %>">
```
harimohanraj89, line 43: If we're not getting the Rails-automagic label from the specific location of this locale `helpers.label.<model>.<attr>`, then shall we use top-level `facility_zone` locale instead? The way I see it now, the top-level locales you've added are the generic keys we should use across the dashboard, and the `helpers.` keys are added _just_ to leverage some Rails magic. 

What do you think? I want to be fairly mindful of how we're using which keys, since we're introducing a little redundancy with this PR.

harimohanraj89 (approved):
Some minor suggestions. Pre-emptive approval to unblock you 👍 

Additionally, do we need to revisit any of the copy on the state or district reports? Or are we restricting this PR to only administrative pages?

app/views/admin/facilities/upload.html.erb
```diff
@@ -11,30 +11,30 @@
...
 <h3>Instructions:</h3>
-<ol>
-  <li>
+<ol class="mb-5">
+  <li class="mt-3">
     Download and fill in
     <%= link_to 'this template', '/documents/upload_facilities.csv' %>
     to create multiple facilities via file upload.
   </li>
-  <li>Organizations for the facilities must already exist.
```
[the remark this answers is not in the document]
dburka: Yes, exactly @harimohanraj89. It's technically a correct instruction, but it's hard to imagine an instance when someone is bulk uploading facilities in place where an Org doesn't exist. I removed it to make this simpler.

app/views/admin/facility_groups/_form.html.erb
dburka: Leaving as-is so we don't accidentally break anything.

app/views/admin/facility_groups/_form.html.erb
timcheadle: Agree, we should have high level localization and, if needed, make the helpers use *those*. Will clean that up, perhaps in another PR.

PR #2464, opened 2021-05-04
Fix dashboard start of month bug

prabhanshuguptagit:
**Story card:** [ch3362](https://app.clubhouse.io/simpledotorg/story/3362)

## Because

We saw zeroes and negative numbers in region reports on May 1. This got fixed when cache was warmed the next day.

## This addresses

On May 1st 4AM IST (when the cache warmer runs), system time is 11:30PM UTC. The "current period" is still April, so the last period defaults to April. No results are populated for May in the cache. The requests come in IST time zone the next day for May. The cache returns no data for May and reports break.

This sets the default period to current month **in the reporting time zone**.

The fix is in ed3f050. The remaining commits are a refactor to consolidate `DEFAULT_ANALYTICS_TIME_ZONE` declaration.


app/services/reports/region_service.rb
```diff
@@ -4,7 +4,7 @@ class RegionService
 
     # The default period we report on is the current month.
     def self.default_period
-      Period.month(Date.current.beginning_of_month)
+      Period.month(Date.current.in_time_zone(Period::ANALYTICS_TIME_ZONE))
```
harimohanraj89, line 7: Do we need `beginning_of_month`?

harimohanraj89 (changes requested):
Can we add a test for this scenario? It can be on the `region_service`, perhaps.

app/services/reports/region_service.rb
rsanheim: > Do we need `beginning_of_month`?

yes we definitely want beginning of month here.

rsanheim (commented):
This fixes one part of the problem, but I think we want to also fix the front end side where it can properly handle no values for a particular month without returning NaN. That case could still happen for some edge-case type scenarios I think.

Non-blocking for this PR, though, as it could be its own PR.

I do think we want the begin of month fix for `Period.default`.

app/services/reports/region_service.rb
prabhanshuguptagit: `Period.month` already does a `beginning_of_month`. ex. Period.month(Date.today) = 1st May. It's not _required_, no?

harimohanraj89: Ah nice. Yup, if `Period.month` takes care of `beginning_of_month`, then we should be good 🆗

PR #2465, opened 2021-05-04
3392: scope monthly district data report to currently selected month

kpethtel:
**Story card:** [ch3392](https://app.clubhouse.io/simpledotorg/story/3392/add-ability-to-make-reports-starting-on-past-months?ct_workflow=all&vc_group_by=day)

## Because

The recently created monthly district data csv followed the pattern established by the existing csv report downloads, which scope the report to the current month. We would like to be able to create these reports for any month. 

## Test instructions
- Go to the reports tab and navigate to a district
- change the month dropdown to a different month
- click the download tab
- there should be a new option "Monthly district data"
- click that and accept the download
- expect it to be populated with data for the district and its facilities, scoped to the selected month


app/services/monthly_district_data_service.rb
```diff
@@ -65,62 +65,66 @@ def header_row
   end
 
   def district_row
-    complete_registration_counts = repo.complete_registration_counts.find { |k, _| k.slug == region.slug }.last
```
kpethtel, line 68: The changes to this file are in response to a suggestion by @ssrihari to use hashes instead of arrays to make it easier to keep track of the csv contents.

app/controllers/reports/regions_controller.rb
```diff
@@ -117,7 +117,8 @@ def monthly_district_data_report
         raise ActiveRecord::RecordNotFound, "unknown report_scope #{report_scope}"
       end
     }
-    @period = Period.month(Date.current)
+    period_end = params[:period_end] || Date.current
+    @period = Period.month(period_end)
```
rsanheim, line 121: This can just be something like this:

```
@current_period = Period.month(params[:period] || Date.current)
```

The fact that it is the end of the range of data is implicit through all these controlled actions.

rsanheim (commented):
one bit of feedback, and some standardrb fixes needed here

spec/services/monthly_district_data_service_spec.rb
```diff
@@ -3,11 +3,26 @@
 RSpec.describe MonthlyDistrictDataService do
   let(:organization) { FactoryBot.create(:organization) }
   let(:facility_group) { create(:facility_group, organization: organization) }
-  let(:facility) { create(:facility, facility_group: facility_group) }
```
kpethtel, line 6: These specs were pretty weak so I tried to improve them. It's difficult to get the data set up correctly to test everything and didn't seem worth the time, but this is still an improvement I think.

spec/services/monthly_district_data_service_spec.rb
```diff
@@ -24,49 +39,75 @@ def find_in_csv(csv_data, row_index, column_name)
...
       expect(csv[facility_row_index][30..31].uniq).to eq([nil])
     end
+
+    it "scopes the report to the provided period" do
+      old_period = Period.month("July 2018")
+      result = described_class.new(region, old_period).report
+      csv = CSV.parse(result)
+      column_headers = csv[2]
+      expect(column_headers[12]).to eq("Feb-2018")
+      expect(column_headers[17]).to eq("Jul-2018")
```
rsanheim, line 110: Is there any significance to the magic numbers 12 and 17 here?  It may be more clear here to do something like this:

```
  first_month_index = 12
  last_month_index = 17
  expect(column_headers[first_month_index]).to eq("Feb-2018")
  expect(column_headers[last_month_index]).to eq("Feb-2018")
```



  
  column_headers[month_header_start_index].upto(months_of_date).with_index { |

rsanheim (approved):
Couple small bits of feedback, along w/ an approval to unblock.

app/controllers/reports/regions_controller.rb
```diff
@@ -117,7 +117,7 @@ def monthly_district_data_report
         raise ActiveRecord::RecordNotFound, "unknown report_scope #{report_scope}"
       end
     }
-    @period = Period.month(Date.current)
+    @period = Period.month(params[:period_end] || Date.current)
```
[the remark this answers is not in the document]
kpethtel: The reason I named it something else is that it actually represents something different. Elsewhere in the downloads, it's used to define the type of period not the value: `period: :month`. In this case, it's being used to pass along the period value. I was trying to disambiguate it, but I'll agree that period_end isn't a very good name. `report_date` or something similar might have been better. Regardless, I'll change it back to period because I don't think it should have much impact.

PR #2467, opened 2021-05-05
Start a CODEOWNERS files for request review on reporting queries

rsanheim:
This adds a CODEOWNERS file that should request review from @claudiovallejo for any changes to the `app/queries` directory. This is covers the reporting queries, as well as other unrelated queries...which isn't perfect but I think that's fine to start as see how this works. 

See https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
for docs and examples

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,8 @@
 # Ensure Claudio gets requested review for reporting queries
 /app/queries/ @@claudiovallejo
 
-# See docs and examples for this file at 
+# Ensure Hari gets requested review canonical medications
+/config/data/medications.yml @harimohanraj89
+
+# See docs and examples for this file at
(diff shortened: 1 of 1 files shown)


claudiovallejo (approved):
Thanks for doing this @rsanheim!!

.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# Ensure Claudio gets requested review for reporting queries
+/app/queries/ @@claudiovallejo
```
harimohanraj89, line 2: `@claudiovallejo`, single `@`?

harimohanraj89 (changes requested):

harimohanraj89:
@rsanheim I've added myself to `/config/data/medications.yml` as discussed [here](https://github.com/simpledotorg/simple-server/pull/2456#discussion_r626453872).

harimohanraj89 (approved):

PR #2468, opened 2021-05-05
Send changelog to slack from release script

prabhanshuguptagit:
**Story card:** [chXXXX](URL)

## Because
We have to manually copy paste the changelog to slack after every release.

## This addresses
`bin/release` now sends the changelog to #releases.

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -49,7 +49,8 @@ end
 puts "CHANGELOG for #{release_tag}"
 puts "---------"
 puts changelog
-puts "SIMPLE_SERVER_DEPLOYMENT_NOTIFICATIONS_WEBHOOK is not present, notification will be skipped." unless ENV['SIMPLE_SERVER_DEPLOYMENT_NOTIFICATIONS_WEBHOOK']
+puts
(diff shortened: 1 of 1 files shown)


bin/release
```diff
@@ -47,6 +49,7 @@ end
 puts "CHANGELOG for #{release_tag}"
 puts "---------"
 puts changelog
+puts "SIMPLE_SERVER_DEPLOYMENT_NOTIFICATIONS_WEBHOOK is not present, notification will be skipped." unless ENV['SIMPLE_SERVER_DEPLOYMENT_NOTIFICATIONS_WEBHOOK']
```
harimohanraj89, line 52: Wanna do this [here](https://github.com/simpledotorg/simple-server/pull/2468/files#diff-ba4e74f7fdeafe09fab0dd8bd69b66578025911d8ea9403bf70182183fdd6600R79) instead?

Basically, if token's not there, we can say "Token missing, please publish the changelog in slack..."

harimohanraj89 (approved):
One nit, ship it!

PR #2471, opened 2021-05-06
Monthly district reports for FacilityDistrict

prabhanshuguptagit:
**Story card:** [chXXXX](URL)

## Because

Monthly district reports are breaking for `FacilityDistrict`s.
image

https://simpledotorg.slack.com/archives/C013HS4N9U3/p1620285572126300

## This addresses

We lookup the district region by slug for facility districts, however the facility district links use the district's name.
Eg. `https://api.simple.org/reports/regions/facility_district/Bathinda`. This fixes the bug.


harimohanraj89 (approved):
Can we add a controller test for this case as well?

PR #2472, opened 2021-05-06
Drop unused views

prabhanshuguptagit:
**Story card:** [ch2443](https://app.clubhouse.io/simpledotorg/story/2443/cleanup-unused-materialized-views)

## Because

These views were originally created for using in metabase queries, but we don't use it anywhere.

```
bp_drugs_views
bp_views
follow_up_views
overdue_views
patient_first_bp_views
patients_blood_pressures_facilities
```

## This addresses

Cleans up these views.

This is more or less a revert of https://github.com/simpledotorg/simple-server/pull/547


harimohanraj89 (commented):
Should we keep the `db/views` SQL files? 🤔

prabhanshuguptagit:
> Should we keep the db/views SQL files? 🤔

I tried removing them but it breaks migrations when they're run from the beginning. One way to make it work would be to squash our migrations using something like https://github.com/jalkoby/squasher. Shall we do it? It'll let us drop a bunch of `v_01`, `v_02` mat views also.

PR #2473, opened 2021-05-06
3412: correct inaccurate numbers on monthly district data export

kpethtel:
**Story card:** [ch3412](https://app.clubhouse.io/simpledotorg/story/3412/monthly-district-data-export-numbers-are-not-correct)

## Because

There were a few issues with the first draft of the new monthly district data export. 
1) the follow-ups by month were not correct when creating reports for past months
2) the missed visits numbers were incorrect on prod. 
3) the assigned patient count did not reflect the denominator from the reports overview page
4) we also want to add an assigned patient count to reflect the number as of today

## This addresses

This addresses 1, 3, and 4. I was unable to replicate the missed visits issue locally. 

## Test instructions

The best way to test this is to print a district report and compare the numbers to those found in the report overview and details pages. Most of the numbers here are easily available in the reports.


rsanheim (commented):
Is there a functional type spec we could add for this?

kpethtel:
> Is there a functional type spec we could add for this?

Let's talk about that tomorrow. I agree that the specs for this are very weak, but I'm just trying to get this to Praveen ASAP.

prabhanshuguptagit (approved):

vkrmis (approved):
lgtm

PR #2474, opened 2021-05-07
Refresh missed_visits cache in repository

prabhanshuguptagit:
**Story card:** -

## Because

We noticed the missed visits number between the `Repository` and `Result` was not matching. This happens because the `Repository` missed visits cache doesn't get refreshed via the cache warmer and only refreshes after the TTL which is 7 days.

## This addresses

This calls the missed visits method in the cache warmer code path so it gets refreshed. We're adding this since it blocks this report: https://github.com/simpledotorg/simple-server/pull/2458. 

The long term fix is to remove the missed visit calculation from result and use the method in repository for `ControlRateService`.

### DIFF
--- a/app/services/reports/region_service.rb
+++ b/app/services/reports/region_service.rb
@@ -39,6 +39,10 @@ module Reports
       result.calculate_missed_visits_percentages(calc_range, with_ltfu: true)
       result.calculate_period_info(calc_range)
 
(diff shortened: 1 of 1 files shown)


vkrmis (approved):
Lets add a maintenance card so we don't forget to do this

prabhanshuguptagit:
https://app.clubhouse.io/simpledotorg/story/3443/move-missed-visit-calculation-from-result-to-repository

PR #2481, opened 2021-05-10
Replace "Protocols" copy with "Medications"

claudiovallejo:
**Story card:** [ch3456](https://app.clubhouse.io/simpledotorg/story/3456/replace-protocols-copy-with-medications)

## Because

Using "Protocol" is misleading. "Protocols" are customizable lists of medications that appear in the Simple App. Using "Medications" is much clearer.

## This addresses

Updates all "protocols" copy in the "Protocol medications" admin pages

## Test instructions

Check out [this Figma file](https://www.figma.com/file/9inSvRD4dXcMfXchlyDZVV/Dashboard-Production?node-id=5989%3A2973) for a full list of copy changes


app/views/admin/protocols/_form.html.erb
```diff
@@ -1,9 +1,9 @@
 <div class="row">
   <div class="col-md-6">
-    <%= bootstrap_form_with(model: [:admin, protocol], local: true, label_errors: true) do |form| %>
-      <%= form.text_field :name, id: :protocol_name %>
-      <%= form.number_field :follow_up_days, id: :protocol_follow_up_days, help: "Note: Usually 28 days to ensure patients aren't reminded to return on days when facilities are closed." %>
```
claudiovallejo, line 5: @rsanheim How can I make this field's title "Follow-up days"? I tried using a string, but I broke the form because it doesn't accept a value for the parameter "follow up days".

kpethtel (changes requested):
I think we should get Rob's opinion about this. It seems to me that this PR is a half-way solution. It's half way between only changing the front end and fully changing Protocol to MedicationList. I think we should pick one or the other. Either we only change the wording in the front end, or we change the protocol model to medication list.

kpethtel (approved):

app/views/admin/protocols/_form.html.erb
claudiovallejo: Thanks @kpethtel, adding a `label: ` field fixed it :)

vkrmis:
Can this be "Protocol Medications" instead? We've already added a `Medications` model that captures non protocol medications: 
- https://github.com/simpledotorg/simple-server/blob/master/app/models/medication.rb
- https://github.com/simpledotorg/simple-server/blob/master/config/data/medications.yml

Changing this copy to Medications, but leaving the underlying models as `Protocol` and `ProtocolDrug`, + keeping the `Medications` model will get pretty confusing imo.

claudiovallejo:
> Can this be "Protocol Medications" instead? We've already added a `Medications` model that captures non protocol medications:
> 
> * https://github.com/simpledotorg/simple-server/blob/master/app/models/medication.rb
> * https://github.com/simpledotorg/simple-server/blob/master/config/data/medications.yml
> 
> Changing this copy to Medications, but leaving the underlying models as `Protocol` and `ProtocolDrug`, + keeping the `Medications` model will get pretty confusing imo.

@vkrmis I don't think we should name it "Protocol Medications" - "Protocol" is not an appropriate name to use. We could use "Common Medications", "Medication List", or "Mobile App Medications". I used "Medication Lists" in parts of the code.

Let's pair when you and @ssrihari update the model/tables so we can align on a good name.

rsanheim:
> Can this be "Protocol Medications" instead? We've already added a `Medications` model that captures non protocol medications:

Maybe that `medications` model should be named `non_standard_medications` or something? I'm not sure on the name, but I agree it would be good to figure out names much closer to what our users and health care workers would expect for what we are representing here.

rsanheim (approved):

PR #2503, opened 2021-05-11
2940: schedule experiment appointment reminders

kpethtel:
**Story card:** [ch2940](https://app.clubhouse.io/simpledotorg/story/2940)

## Because

We need the appointment notification system to work for scheduled appointment reminders to enable A/B testing. 

## This addresses

The current appointment notification system is only designed to send notifications for appointments that are 3+ days overdue. We are changing the process to create an appointment reminder for those overdue appointments and changing the worker to send notifications based on appointment reminders instead. This will make it work for A/B experiments and for the existing process.


app/models/appointment.rb
```diff
@@ -65,6 +65,10 @@ def self.eligible_for_reminders(days_overdue: 3)
     overdue_by(days_overdue)
       .joins(:patient)
       .merge(Patient.contactable)
+      .includes(:communications)
```
kpethtel, line 68: As you may recall, there's a bug in how notifications currently work. We send the same messages daily because this query looks for appointments that are 3+ days overdue. Then it skips if the appointment has a successful twilio response. Which means any failed communications will attempt to send every day. 

This changes the behavior to only include appointments if they don't have either a communication or appointment reminder. That's an improvement because we shouldn't keep trying if the communication failed. It might occasionally be successful the second time, but more often it will just fail again, either because the person doesn't have whatsapp or the number is no longer valid. Any notification that has ever been attempted (and didn't fail due to a sidekiq or twilio client error) should have a communication.

We also exclude any appointments that have reminders because an appointment that has a reminder but no communication is part of an experiment and will be handled by the experiment cron.

app/jobs/appointment_notification/schedule_experiment_reminders.rb
```diff
@@ -0,0 +1,14 @@
...
+  queue_as :high
+
+  def perform
+    return unless Flipper.enabled?(:appointment_reminders)
+
+    reminders = AppointmentReminder.due_today
+    next_messaging_time = Communication.next_messaging_time
+    reminders.each do |reminder|
+      AppointmentNotification::Worker.perform_at(next_messaging_time, reminder.id)
+      reminder.status_scheduled!
```
rsanheim, line 11: We should probably set the state to `scheduled` first, otherwise there is a race condition where the job picks up the reminder before the status has been set.

app/jobs/appointment_notification/worker.rb
```diff
@@ -4,27 +4,28 @@ class AppointmentNotification::Worker
...
   end
 
-  def perform(appointment_reminder_id, communication_type)
+  def perform(appointment_reminder_id)
     metrics.increment("attempts")
     unless Flipper.enabled?(:appointment_reminders)
       metrics.increment("skipped.feature_disabled")
       return
     end
     reminder = AppointmentReminder.includes(:appointment, :patient).find(appointment_reminder_id)
```
rsanheim, line 17: Can you chain the `find` first and then includes? It reads better as the `find` is the more important part of the lookup here.

app/jobs/appointment_notification/worker.rb
```diff
@@ -4,27 +4,28 @@ class AppointmentNotification::Worker
...
-    if reminder.appointment.previously_communicated_via?(communication_type)
+    communication_type = reminder.next_communication_type
+    unless communication_type
       metrics.increment("skipped.previously_communicated")
       return
     end
-    if reminder.status != "scheduled"
-      report_error("scheduled appointment reminder has invalid status")
+
+    unless reminder.status == "scheduled"
```
rsanheim, line 24: Here is the possible race condition.  Also, this can use the question method:

```
reminder.status_scheduled?
```

app/models/appointment.rb
```diff
@@ -65,6 +65,10 @@ def self.eligible_for_reminders(days_overdue: 3)
...
+      .where(communications: {id: nil})
+      .includes(:appointment_reminders)
```
rsanheim, line 70: I think you want to use an explicit [left_joins](https://apidock.com/rails/ActiveRecord/QueryMethods/left_outer_joins) here, as `includes` doesn't always necessarily do a left outer.

app/models/appointment.rb
rsanheim: should be explicit `left_joins`

app/models/appointment_reminder.rb
```diff
@@ -14,4 +15,45 @@ class AppointmentReminder < ApplicationRecord
...
+
+  def preferred_communication_method
+    CountryConfig.current[:name] == "India" ? "missed_visit_whatsapp_reminder" : nil
+  end
+
+  def backup_communication_method
+    "missed_visit_sms_reminder"
+  end
+
+  # this is temporary. will be replaced by facility address localization
```
rsanheim, line 55: Can you explain more when this will go away? Or link up the clubhouse story in the comment with the follow on work?

app/services/notification_service.rb
```diff
@@ -24,19 +25,26 @@ def parse_phone_number(number)
...
-    client.messages.create(
+    @response = client.messages.create(
       from: sender_number,
       to: recipient_number,
       status_callback: callback_url,
       body: message
     )
+  rescue Twilio::REST::TwilioError => e
+    report_error(e)
+    @error = :exception
```
rsanheim, line 47: How about

```
  @exception = exception
  report_error(exception)
end
```

We may as well capture the actual exception here rather than a symbol.

spec/controllers/api/v3/twilio_sms_delivery_controller_spec.rb
```diff
@@ -97,49 +97,90 @@
...
+            fallback_time,
+            communication.appointment_reminder_id
+          )
+          expect(Statsd.instance).to receive(:increment).with("twilio_callback.missed_visit_whatsapp_reminder.failed")
 
-          set_twilio_signature_header(callback_url, params)
           post :create, params: params
         end
+
+        # it "schedules a fallback SMS if a whatsapp message failed" do
```
rsanheim, line 142: Why is all this commented out?

rsanheim (commented):
Some initial thoughts on first pass here.

app/jobs/appointment_notification/schedule_experiment_reminders.rb
kpethtel: That's a fair point. I didn't consider it because I expect this script to run hours before the notifications send, because that's how the current notifications work. But tbh I think it's worth considering changing that behavior. I don't see much advantage to scheduling the notifications several hours early. It may have some small performance impact but it could also leave us open to sidekiq/redis errors.

app/jobs/appointment_notification/worker.rb
kpethtel: AFAIK that's just not how the syntax works. https://stackoverflow.com/questions/3943554/how-to-do-find-with-includes-in-rails-3

app/jobs/appointment_notification/worker.rb
kpethtel: Done.

app/models/appointment.rb
kpethtel: Thanks for catching this.

app/models/appointment_reminder.rb
kpethtel: I'll link the ticket. It's a subsequent ticket in this epic and I released a draft of it a while ago.

app/services/notification_service.rb
kpethtel: WIll do.

spec/controllers/api/v3/twilio_sms_delivery_controller_spec.rb
kpethtel: This is an oopsie daisy. It was the lat thing I changed and I obviously failed to delete the old code.

app/jobs/appointment_notification/worker.rb
rsanheim: 👍

app/models/appointment_reminder.rb
```diff
@@ -14,4 +15,46 @@ class AppointmentReminder < ApplicationRecord
...
+  def preferred_communication_method
+    CountryConfig.current[:name] == "India" ? "missed_visit_whatsapp_reminder" : nil
+  end
+
+  def backup_communication_method
+    "missed_visit_sms_reminder"
+  end
+
+  # this is temporary. will be replaced as part of:
+  # https://app.clubhouse.io/simpledotorg/story/2639/respect-translations-for-existing-patient-reminder-messages
```
rsanheim, line 56: thanks 👍

app/jobs/appointment_notification/worker.rb
rsanheim: Got it.

rsanheim (approved):
Looks good.

PR #2507, opened 2021-05-11
Cache only periods with data

rsanheim:
**Story card:** [ch3490](https://app.clubhouse.io/simpledotorg/story/3490/cache-only-periods-with-data)

Currently we cache entries for periods even going back before a Region has data - i.e. before there were any patients registered. This is extra work, and produces many extra keys and objects.

We should only cache entries where there is any actual data to cut down on this.


app/models/reports/repository.rb
```diff
@@ -74,6 +74,7 @@ def initialize(regions, periods:)
     memoize def adjusted_patient_counts
       cumulative_assigned_patients_count.each_with_object({}) do |(entry, result), results|
         values = periods.each_with_object(Hash.new(0)) { |period, region_result|
+          next unless result.key?(period.adjusted_period)
```
rsanheim, line 77: We don't want to attempt the subtraction here if there aren't any patient counts.

app/models/reports/repository.rb
```diff
@@ -238,18 +239,23 @@ def denominator(region, period)
     #   }
     #
     def region_period_cached_query(calculation, &block)
+      results = regions.each_with_object({}) { |region, hsh| hsh[region.slug] = Hash.new(0) }
```
rsanheim, line 242: Ensure we always have an entry for every region, even if there is no data for that particular region.

app/models/reports/repository.rb
```diff
@@ -238,18 +239,23 @@ def denominator(region, period)
...
+    # Generate all necessary region period cache entries, only going back to the earliest
+    # patient registration date for Periods. This ensures that we don't create many cache entries
+    # with 0 data for newer Regions.
     def cache_entries(calculation)
-      combinations = regions.to_a.product(periods.to_a)
+      combinations = regions.each_with_object([]) do |region, sum|
+        earliest_period = earliest_patient_recorded_at_period[region.slug]
+        next if earliest_period.nil?
+        periods_with_data = periods.select { |period| period >= earliest_period }
+        sum.concat(periods_with_data.to_a.map { |period| [region, period] })
```
rsanheim, line 257: Here is where we build up the periods we actually care about.

rsanheim (commented):
some notes

app/models/reports/repository.rb
```diff
@@ -238,18 +239,23 @@ def denominator(region, period)
...
+      cached_results.each { |(entry, count)| results[entry.region.slug][entry.period] = count }
+      results
     end
 
+    # Generate all necessary region period cache entries, only going back to the earliest
+    # patient registration date for Periods. This ensures that we don't create many cache entries
+    # with 0 data for newer Regions.
     def cache_entries(calculation)
-      combinations = regions.to_a.product(periods.to_a)
+      combinations = regions.each_with_object([]) do |region, sum|
```
kpethtel, line 253: I'm not sure sum is the best name for the array. Maybe just "result" like we used above.

spec/models/reports/repository_spec.rb
```diff
@@ -363,6 +373,18 @@ def refresh_views
...
+    it "works for very old dates" do
+      facility_1 = create(:facility)
+      patient = create(:patient, registration_facility: facility_1, recorded_at: jan_2020.advance(months: -4))
+      create(:blood_pressure, :under_control, recorded_at: jan_2020.advance(months: -1), patient: patient, facility: facility_1)
+      refresh_views
+
+      ten_years_ago = patient.recorded_at.advance(years: -10).to_period
+      range = ten_years_ago..(ten_years_ago.advance(months: 12))
+      repo = Reports::Repository.new(facility_1, periods: range)
+      expect(repo.adjusted_patient_counts[facility_1.slug]).to eq({})
```
kpethtel, line 385: I'm not sure I understand why we expect this to be empty. Is it because the patient is no longer hypertensive or is now LTFU? The test does validate that it doesn't bomb with older facilities, but it doesn't really test much function.

kpethtel (approved):
A couple of minor comments but looks good.

app/models/reports/repository.rb
rsanheim: 👍

app/models/reports/repository.rb
```diff
@@ -6,25 +6,31 @@ class Repository
...
       end
       @period_type = @periods.first.type
       raise ArgumentError, "Quarter periods not supported" if @period_type != :month
+
+      @assigned_patients_query = AssignedPatientsQuery.new
+      @control_rate_query = ControlRateQuery.new
+      @earliest_patient_data_query = EarliestPatientDataQuery.new
+      @follow_ups_query = FollowUpsQuery.new
+      @no_bp_measure_query = NoBPMeasureQuery.new
+      @registered_patients_query = RegisteredPatientsQuery.new
```
rsanheim, line 22: Its now easier to scan all the dependent queries that are involved here.

spec/models/reports/repository_spec.rb
rsanheim: The result set would be. empty because the range we are selecting for (10 years ago) in the Repository is before the registration date of the patient.

spec/queries/registered_patients_query_spec.rb
```diff
@@ -12,12 +14,23 @@
...
       create_list(:patient, 1, registration_facility: other_facility, registration_user: user2)
     end
-    result = RegisteredPatientsQuery.new.count(facility, :month)
+    result = described_class.new.count(facility, :month)
     expect(result[Period.month("December 1st 2020")]).to be_nil
     expect(result[Period.month("January 1st 2020")]).to eq(2)
     expect(result[Period.month("April 1st 2020")]).to eq(2)
   end
 
+  it "includes dead patients" do
```
rsanheim, line 23: @claudiovallejo @prabhanshuguptagit verifying this is true for the registered patients query.

PR #2508, opened 2021-05-12
Fix some dashboard copy

prabhanshuguptagit:
**Story card:** 

## Because

Some dashboard footnotes were incorrect.

## This addresses

- Fixes a region reports footnote
image
- Fixes a progress tab copy
- Fixes `undefined` percentage when no data is present
image


harimohanraj89 (approved):
Sounds good to me

PR #2510, opened 2021-05-12
DHIS2 aggregate data export (proof of concept)

timcheadle:
**Story card:** [ch3369](https://app.clubhouse.io/simpledotorg/story/3369/use-api-to-bulk-insert-aggregate-data)

## Because

We need to test using the DHIS2 API to bulk insert aggregate data for each facility, each month

## This addresses

This uses the `dhis2` Ruby gem that is relatively unmaintained but still works. It sends 9 basic indicators:

- Cumulative registrations
- Monthly registrations
- Cumulative assigned patients
- Cumulative assigned patients (adjusted, excludes <3 month registrations)
- Controlled patients
- Uncontrolled patients
- Missed visits
- Dead patients
- LTFU patients

## Test instructions

1. Copy the sandbox DB locally
2. run `rails dhis2:export`
3. Confirm the data was sent by querying the Pivot Table app: http://dhis2-sandbox.simple.org/dhis-web-pivot/


lib/tasks/dhis2.rake
```diff
@@ -0,0 +1,61 @@
+namespace :dhis2 do
+  desc "Export aggregate indicators for each facility to DHIS2"
+  task export: :environment do
+    data_elements_map = {
```
timcheadle, line 4: These are Bangladesh DHIS2 IDs (that I created). This whole thing is temporary; future iterations will have a separate data element config file

app/models/facility_business_identifier.rb
```diff
@@ -0,0 +1,12 @@
...
+  belongs_to :facility
+
+  enum identifier_type: {
+    dhis2_org_unit_id: "dhis2_org_unit_id"
+  }
+
+  validates :identifier, presence: true
+  validates :identifier_type, presence: true
+  validates :identifier_type, uniqueness: {scope: :facility_id}
+  validates :facility, presence: true
```
kpethtel, line 11: I'm not sure if you truly need this, given that the relationship is not optional, but it's not hurting anything.

lib/tasks/dhis2.rake
rsanheim: Can you include a comment indicating that these are hard coded to SBX values for now?

.env.development
```diff
@@ -122,3 +122,11 @@ SUPERVISOR_EMAILS='Supervisor <supervisor@test.com>'
...
 SEED_GENERATED_ACTIVE_USER_ROLE="Seed User | Active"
 SEED_GENERATED_INACTIVE_USER_ROLE="Seed User | Inactive"
+
+#
+# DHIS2 credentials
+#
+DHIS2_URL=
+DHIS2_USERNAME=
+DHIS2_PASSWORD=
+DHIS2_VERSION="2.28"  # This is a hack because the dhis2 gem only goes to 2.28
```
rsanheim, line 132: Not sure I understand this comment...because the gem we have locked is 3.3.1.  Can you expand on this for folks who aren't super familiar w/ how the gem and server interact?

app/models/facility_business_identifier.rb
timcheadle: This ensures you can't create a business identifier unless the facility actually exists

.env.development
timcheadle: This is the DHIS2 version installed on the server. The API is apparently tied to specific versions (?!)

We will need to update the gem in the future.

rsanheim:
```suggestion
DHIS2_VERSION="2.28"  # This is a hack because the current dhis2 gem only works against version 2.28 of the DHIS API
```

rsanheim (approved):

config/initializers/dhis2.rb
```diff
@@ -0,0 +1,6 @@
+Dhis2.configure do |config|
```
rsanheim, line 1: You could move this configuration into the task itself I think, as we don't need DHIS integration anywhere else.  Its pretty isolated.

rsanheim (commented):
need this to fix the build

PR #2511, opened 2021-05-14
Overdue bugfix

harimohanraj89:
**Story card:** [ch3498](https://app.clubhouse.io/simpledotorg/story/3498/user-unable-to-open-overdue-list-in-dashboard)

## Because

Sometimes discarded patients have undiscarded appointments, and our dashboard tries to render them in the overdue list and breaks.

## This addresses

Omits discarded patients from the PatientSummary **not**-materialized view.

Notes
* I've chosen not to prevent this data situation (discarded patient with undiscarded appointments) from arising because we have a very distributed data model with syncs coming from everywhere all the time. I think it's more valuable to be tolerant in this case, rather than introducing callbacks, validations, etc.
* I have verified that the only touchpoints of the PatientSummary view is for this overdue list. Patient linelists utilize a separate MaterializedPatientSummary view, and will be unaffected
* The MaterializedPatientSummary does a similar discarded exclusion, so I've done the same for consistency. The alternative would be to add the `deleted_at` field to the view, and do the scoping at query time.

Additional fun things

* Speaking of MaterializedPatientSummary, it would be great for that view to be re-written as `SELECT * FROM patient_summaries`, so that it stays consistent with the changes we're making to PatientSummary in the future.

## Test instructions

* Find an overdue patient in the dashboard.
* Discard them, not their appointments,
* Make sure the overdue list still works, and they're no longer there.

### DIFF
--- a/db/views/patient_summaries_v05.sql
+++ b/db/views/patient_summaries_v05.sql
@@ -1,0 +1,116 @@
+SELECT
+p.recorded_at,
+CONCAT(date_part('year', p.recorded_at), ' Q', EXTRACT(QUARTER FROM p.recorded_at)) AS registration_quarter,
+p.full_name,
+(
+    CASE
+      WHEN p.date_of_birth IS NOT NULL THEN date_part('year', age(p.date_of_birth))
+      ELSE p.age + date_part('years', age(NOW(), p.age_updated_at))
+    END
+) AS current_age,
+p.gender,
+p.status,
+p.assigned_facility_id AS assigned_facility_id,
+latest_phone_number.number AS latest_phone_number,
+addresses.village_or_colony AS village_or_colony,
+addresses.street_address AS street_address,
+addresses.district AS district,
+addresses.state AS state,
+reg_facility.name AS registration_facility_name,
+reg_facility.facility_type AS registration_facility_type,
+reg_facility.district AS registration_district,
+reg_facility.state AS registration_state,
+latest_blood_pressure.systolic AS latest_blood_pressure_systolic,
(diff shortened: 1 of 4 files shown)


db/views/patient_summaries_v05.sql
```diff
@@ -0,0 +1,116 @@
...
+ON latest_bp_passport.patient_id = p.id
+
+LEFT OUTER JOIN (
+    SELECT DISTINCT ON (patient_id) *
+    FROM appointments
+    ORDER BY patient_id, scheduled_date DESC
+) AS next_appointment
+ON next_appointment.patient_id = p.id
+LEFT OUTER JOIN facilities next_appointment_facility ON next_appointment_facility.id = next_appointment.facility_id
+WHERE p.deleted_at IS NULL;
```
prabhanshuguptagit, line 116: This is the only change in the definition?

prabhanshuguptagit (approved):
> it would be great for that view to be re-written as SELECT * FROM patient_summaries

Let's defo do that. I realised the same while figuring out the flaky spec. I don't think we'll be able to deprecate `patient_summaries` anytime soon as was originally intended.

db/views/patient_summaries_v05.sql
harimohanraj89: Correcto 👌

ssrihari:
@harimohanraj89 @prabhanshuguptagit My first thoughts to solve this would've been to mark incoming appointments belonging to discarded patients as discarded. Is there a reason to not do this?

Discarding a patient can't be "undone" from the app, so we should be in the clear here?

PR #2513, opened 2021-05-14
Cleanup old views and migrations

prabhanshuguptagit:
**Story card:** [ch3503](https://app.clubhouse.io/simpledotorg/story/3503/cleanup-old-unused-view-sql-files)

## Because

We have a bunch of old SQL view definitions sitting around in `db/views` that don't exist in the schema anymore. We have to keep them around since running migrations from the beginning requires them. 

## This addresses

Dropping our old migrations will allow us to cleanup `db/views` without leaving missing references. This requires us to start using `db:schema:load` instead of `db:migrate` when setting up a fresh database.
This makes 3 changes:
- clear out `db/migrate`
- clear out `db/data`
- cleanup the unused views from `db/views`

Another alternative is to squash our migrations into a single migration using something like https://github.com/jalkoby/squasher or doing https://naturaily.com/blog/remove-old-migrations-ruby-on-rails. We did not choose this because we’d have to run the squashed migration in prod and it's error prone.

## Test instructions

- Run on semaphore, review apps, SBX etc


harimohanraj89 (approved):
LGTM ✨

PR #2515, opened 2021-05-14
2639 add facility locales

kpethtel:
**Story card:** [ch2639](https://app.clubhouse.io/simpledotorg/story/2639/respect-translations-for-existing-patient-reminder-messages)

## Because

Our current message localization strategy is very limited and sends English messages to most patients. Many of those patients do not read English, so this will inevitably result in lower message effectiveness. We've chosen to localize according to the facility state, because that is more likely to be valid than the patient's address. 

The translations here are based on guidance from Praveen, captured in this document: https://docs.google.com/spreadsheets/d/1JbaJw34Hg9WTLvPDfBgLs7YjGLShmMwl8dmvqJPH86o/edit#gid=0

## Test instructions

I think the specs will suffice here. Messaging is currently turned off in prod and we will be doing some testing in the near future.


app/models/facility.rb
```diff
@@ -234,4 +231,39 @@ def localized_facility_size
     return unless facility_size
     I18n.t("activerecord.facility.facility_size.#{facility_size}", default: facility_size.capitalize)
   end
+
+  def locale
+    LOCALE_MAP.dig(country.downcase, state.downcase) || LOCALE_MAP.dig(country.downcase, "default") || LOCALE_MAP["default"]
+  end
+
+  LOCALE_MAP = {
```
kpethtel, line 239: @dburka I just want to run this by you. This is based on the info I got from Praveen. It's worth noting that there may be some pre-optimizations. Some of the states in Ethiopia listed here do not yet have facilities, but when we add facilities, messages for their patients will be localized. The localization protocol is in the method above:
first we try to localize by country and state; if the state isn't found, we use the country's default language; if the country isn't found, we use English. 

Does that sound right to you? I'm not sure what the process for adding a new country is like, but it might be worth thinking about how rollout to new countries should be handled.

kpethtel, line 239: We can extract this into a service object in the future if we ever need it for anything other than facility.

kpethtel:
@harimohanraj89 I would like you to review this because you seem to be the person who maintains this line in `application.rb`:
```
 config.i18n.available_locales = %w[en mr-IN pa-Guru-IN bn-BD kn-IN en_BD en_IN en_ET en-IND bn-IN hi-IN ta-IN te-IN am-ET om-ET so-ET ti-ET]
```
Do I need to be considering the country specific English translations like `en_BD`? Those seem to be specific to regional names for facility types and such.

app/models/facility.rb
dburka: It sounds good. You're right that we should definitely have a process in place not to miss setting a default language per country. For instance, if Sri Lanka starts soon.

rsanheim (approved):

PR #2517, opened 2021-05-16
Add BP Measures By Period to Repository

rsanheim:
**Story card:** [ch3666](https://app.clubhouse.io/simpledotorg/story/3666/add-bp-measures-per-period-to-repository)

## Because

We want these BP counts to come from the Repository:

image


app/services/patient_import/spreadsheet_transformer.rb
```diff
@@ -221,7 +221,7 @@ def import_user
     end
 
     def patient_status(row)
-      row[:died].presence&.downcase.in?(["yes", "y"]) ? Patient.statuses["dead"] : Patient.statuses["active"]
+      row[:died]&.downcase&.in?(["yes", "y"]) ? Patient.statuses["dead"] : Patient.statuses["active"]
```
timcheadle, line 224: Is this change meant for this PR?

timcheadle (approved):

app/queries/bp_measures_query.rb
```diff
@@ -0,0 +1,22 @@
...
+      .where(facility: region.facilities)
+    if group_by.present?
+      results = query.group(group_by).count
+      sum_groups_by_period(results)
+    else
+      query.count
+    end
+  end
+
+  def sum_groups_by_period(result)
```
harimohanraj89, line 15: We're flipping the data structure so the period is the first key, and the other grouping is the second? Should we remove the word `sum` from this method name?

app/queries/bp_measures_query.rb
```diff
@@ -0,0 +1,22 @@
+class BPMeasuresQuery
+  def count(region, period_type, diagnosis: :hypertension, group_by: nil)
+    query = BloodPressure
+      .joins(:patient).merge(Patient.with_hypertension)
+      .group_by_period(period_type, :recorded_at, {format: Period.formatter(period_type)})
+      .where(facility: region.facilities)
```
harimohanraj89, line 6: This code is ported from `ActivityService`? That _should_ ensure that these monthly numbers match the progress tab's monthly numbers. Should we use the repository in the progress tab right away to enforce consistency?

spec/queries/bp_measures_query_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      create(:blood_pressure, facility: facility, patient: patient, recorded_at: 2.months.ago, user: user_2)
+      expected = {
+        Period.month("January 2021") => {user_1.id => 1, user_2.id => 0},
+        Period.month("February 2021") => {user_1.id => 0, user_2.id => 0},
+        Period.month("March 2021") => {user_1.id => 0, user_2.id => 2}
+
+      }
+      expect(described_class.new.count(facility, :month, group_by: :user_id)).to eq(expected)
+    end
+  end
```
harimohanraj89, line 37: Learning from some recent mistakes, can we add some closer-to-the-boundary specs? Set the application timezone to India, create a BP on Jan 31, 1159pm IST, Feb 1, 1201am IST, make sure that each gets attributed to a separate month?

I know we have some conversations in flight about better ways to handle timezones correctly across the board, but these specs will:
* make sure we get it right in the meantime
* enable us to refactor our handling of timezones with confidence

app/services/patient_import/spreadsheet_transformer.rb
harimohanraj89: I guess it works. `"".downcase` should interpreted as "active".

app/views/reports/regions/_facility_details.html.erb
```diff
@@ -184,6 +184,8 @@
               <% @period_range.each do |period| %>
                 <td class="ta-center">
                   <%= number_with_delimiter(dash_if_zero(@dashboard_analytics.dig(resource.id, :bp_measures_by_period, period.value))) %>
+                  ||
+                  <%= number_with_delimiter(@repository.bp_measures_by_user.dig(@region.slug, period, resource.id)) %>
```
harimohanraj89, line 188: Reminder to remove the old value before we ship.

app/queries/bp_measures_query.rb
harimohanraj89: I guess it gets hard because this service only deals with hypertensive patients. 🤔 Hmm, how should handle this divergence?

harimohanraj89:
Whoops, meant to "request changes". Some minor feedback, as well as a major question about consistency with the progress tab.

app/services/patient_import/spreadsheet_transformer.rb
rsanheim: Yeah, it was, I was cleaning up a standardrb issue with this change.

app/queries/bp_measures_query.rb
rsanheim: > I guess it gets hard because this service only deals with hypertensive patients. 🤔 Hmm, how should handle this divergence?

Good question - I've been ignoring diabetes diagnosis for now, because just handling HTN is so much already.  I think we may want a new epic later to handle DM, whether its with a new repository or some objects that branch off of it.

app/queries/bp_measures_query.rb
rsanheim: Right, this is more inline w/ all the other data the Repo returns.  The structure is:

```
region1 => { period1 => xxx, period2 => yyy },
region2 => { period1 => xxx, period2 => yyy },
```

So for data that has an additional grouping, I'm going w/ the convention of:

```
region1 => { period1 => {group1 => x, group2 => y}, period2 => {group2 => z} },
region2 => { etc... }
```

The goal is that all data from Repository will follow this same sort of shape to stay consistent.

spec/queries/bp_measures_query_spec.rb
rsanheim: Yup will do.

PR #2519, opened 2021-05-17
Fix: Integers not being filtered from logs

prabhanshuguptagit:
**Story card:** [ch3530](https://app.clubhouse.io/simpledotorg/story/3530/fix-log-filtering)

## Because

We noticed while digging into a breaking spec on https://github.com/simpledotorg/simple-server/pull/2441 that our logs don't filter integer fields.
Screenshot 2021-05-17 at 6 00 24 PM

## This addresses

This should start filtering integer fields from logs. Approaches:
1. Find a way to write a lambda and add it to `Rails.application.config.filter_parameters` that can filter integers
2. List out all integer fields in `Rails.application.config.filter_parameters` separately like we do for `age`
3. Add a regex to `Rails.application.config.filter_parameters` that is the inverse of `ALLOWED_REGEX`

We've gone with 2 from above since we couldn't figure out a way to do 1 or 3. 
With 1, The lambda should mutate the filtered values in place, which is not possible for integers.
With 3, it isn't easy to write that can account for nested hashes since when we pass a regex it does a string match.

This does mean that whenever we add a new non-string field to the DB schema we should update `DISALLOWED_INTEGER_PARAMS `. It lets us stop leaking the existing fields though, which is better than the current situation.


config/initializers/filter_parameter_logging.rb
```diff
@@ -23,21 +23,13 @@ module ParameterFiltering
...
-          value.to_s.replace(SANITIZED_VALUE)
-        end
-      end
-    end
-  end
+  # Returns the lambda for attributes that are okay to leave in the logs
+  self.filter = lambda { |key, value| value.replace(SANITIZED_VALUE) unless key.match(ALLOWED_REGEX) }
 end
 
-Rails.application.config.filter_parameters += [:password, :age, ParameterFiltering.filter]
```
harimohanraj89, line 43: `password` is already in the list, so we don't need to add it again?

config/initializers/filter_parameter_logging.rb
```diff
@@ -23,21 +23,13 @@ module ParameterFiltering
...
-        when String
-          value.replace(SANITIZED_VALUE)
-        else
-          value.to_s.replace(SANITIZED_VALUE)
-        end
-      end
-    end
-  end
+  # Returns the lambda for attributes that are okay to leave in the logs
+  self.filter = lambda { |key, value| value.replace(SANITIZED_VALUE) unless key.match(ALLOWED_REGEX) }
```
harimohanraj89, line 32: The decision tree isn't required anymore because it doesn't need to handle integers anymore? Were there other data types that were being handled by that decision tree?

harimohanraj89 (changes requested):
Couple questions

config/initializers/filter_parameter_logging.rb
prabhanshuguptagit: The decision tree wasn't doing anything actually. Only the `String` case was doing the `[FILTERED]` replacement, everything else was being skipped.

config/initializers/filter_parameter_logging.rb
prabhanshuguptagit: Yes.

config/initializers/filter_parameter_logging.rb
harimohanraj89: Interesting. The "Is a string" check looks good. How about hashes? Is the `value.to_s.replace(SANITIZED_VALUE)` replacing sensitive hashes with `FILTERED`? 

Do help me understand how this stuff is working if I'm not getting it 😅

prabhanshuguptagit: The filter parameters API takes a lambda that accepts a `key` and a `value` and expects the lambda to mutate the value in place. It iterates over any hash and calls the lambda on each key value pair.

We were doing `value.to_s.replace(SANITIZED_VALUE)` earlier for non-strings, which just constructed a new string and replaced it. The original value was not mutated. Integers can't be mutated in place unlike strings.

Hope this helps.

https://api.rubyonrails.org/classes/ActionDispatch/Http/FilterParameters.html

harimohanraj89: Ha, makes sense to me now. 🚢

harimohanraj89 (approved):
🚀

PR #2520, opened 2021-05-17
update ngrok instructions in readme

kpethtel:
## Because

Filling a gap in documentation.


README.md
```diff
@@ -105,6 +105,15 @@ SIMPLE_SERVER_HOST=<HTTPS URL>
...
 
+In the server repo, open `app/views/api/manifests/show.json.jbuilder`. Change:
+```
+json.endpoint "#{ENV["SIMPLE_SERVER_HOST_PROTOCOL"]}://#{ENV["SIMPLE_SERVER_HOST"]}/api/"
+```
+to:
+```
+json.endpoint "<HTTPS URL>/api/"
+```
+
```
harimohanraj89, line 116: @kpethtel the steps above regarding `.env.development.local` is an alternative to this manifest change, and likely a cleaner solution. Had you done that when you tested? Should we add a line to the previous section that says "... and restart your server"?

harimohanraj89 (changes requested):

README.md
kpethtel: I had tried that. I think I had restarted my server after, but I'll try again.

kpethtel: I re-tested this. I still couldn't get it to work with the existing instructions. I did restart the server. But using the server-side fix without changing the android configs did work for me.

harimohanraj89 (approved):

PR #2521, opened 2021-05-17
Allow switching our Twilio client between test and production credentials

rsanheim:
**Story card:** [ch3517](https://app.clubhouse.io/simpledotorg/story/3517/add-ability-to-switch-between-prod-and-test-credentials-for-twilio?ct_workflow=all)

We can configure the Twilio client creds at the time the client is created, which will allow us to choose the correct credentials based on the SimpleServerEnv or other runtime criteria.

This requires https://github.com/simpledotorg/deployment/pull/332 to be deployed out to our environments.


app/services/notification_service.rb
```diff
@@ -1,9 +1,39 @@
...
 
-  attr_reader :response, :error
-  def client
-    @client ||= Twilio::REST::Client.new(twilio_account_sid, twilio_auth_token)
+  attr_reader :client
+  attr_reader :error
+  attr_reader :response
+  attr_reader :twilio_sender_number
+
+  def initialize
```
kpethtel, line 9: I think this would be simpler written as:
```
  attr_reader :error
  attr_reader :response
  attr_reader :twilio_sender_number

  def initialize
    @test_mode = !SimpleServer.env.production?
    @twilio_sender_number = ENV.fetch("TWILIO_PHONE_NUMBER")

    if @test_mode
      @twilio_account_sid = ENV.fetch("TWILIO_TEST_ACCOUNT_SID")
      @twilio_auth_token = ENV.fetch("TWILIO_TEST_AUTH_TOKEN")
    else
      @twilio_account_sid = ENV.fetch("TWILIO_ACCOUNT_SID")
      @twilio_auth_token = ENV.fetch("TWILIO_AUTH_TOKEN")
    end
  end

  def test_mode?
    @test_mode
  end

  def client
    @client ||= Twilio::REST::Client.new(@twilio_account_sid, @twilio_auth_token)
  end
```

kpethtel (approved):

harimohanraj89 (changes requested):
Why don't we sub out the env variable values with the test credentials, instead of introducing new environment variables? I guess it lets us switch between clients in a single environment, but should we encourage that?

rsanheim:
> Why don't we sub out the env variable values with the test credentials, instead of introducing new environment variables? I guess it lets us switch between clients in a single environment, but should we encourage that?

We need to be able to switch to the real credentials in certain situations, which is the same reason I want both clients available at run time.  Its for situations where we are running the normal batch runs on sandbox against test Twilio (so SMS never gets really sent), and then we want to verify things really work end to end.  So we switch to the prod client, run some selected notifications against some working India numbers to verify, and then switch back to the test client.

Its a bit more complex but I think worth it so we can verify notification changes in a non-prod environment.

PR #2522, opened 2021-05-17
Improve experiments / notifications system to more flexible

kpethtel:
**Story card:** [ch3514](https://app.clubhouse.io/simpledotorg/story/3514/can-we-run-this-as-an-experiment)

This now does a few things to make our notification system flexible enough to handle the med refill messaging we want to send out, as well as still allowing support for appointment reminders. It changes Notification to have a polymorphic relationship to a `subject`, which for Appt reminders would maintain the appointment relationship.

It also adds a backfill so that Communication data is copied over to Notification...this will allow us to remove redundant relationships in the future.  

There are also a bunch more renames going on to get things consistent here.

* [x] remove dependency on `Appointment` from `Communication`
* [x] rename `AppointmentReminder` to `Notification`
* [x] make `appointment` optional on `Notification`
* [x] rename any other related models / locals / etc
* [x] green build to verify all the things!
* [ ] verify we can send these notifications to Hari's phone on SBX


app/models/appointment.rb
```diff
@@ -148,7 +145,8 @@ def anonymized_data
   end
 
   def previously_communicated_via?(communication_type)
-    communications.latest_by_type(communication_type)&.attempted?
+    matching_notifications = notifications.includes(:communications).where(communications: {communication_type: communication_type})
+    matching_notifications.any? { |n| n.communication.attempted? }
```
rsanheim, line 149: I don't think this quite matches the previous logic ?

db/migrate/20210517215935_remove_appointment_from_communication_and_add_patient.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveAppointmentFromCommunicationAndAddPatient < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :communications, :appointment_id
+    add_reference :communications, :patient, null: true, foreign_key: true, type: :uuid
```
prabhanshuguptagit, line 4: Before we drop `appointment_id`, should we add a data migration to copy over `patient_id` for existing `communications`?

prabhanshuguptagit (commented):
Looking great so far 💯 .

app/models/appointment.rb
```diff
@@ -67,10 +66,8 @@ def self.eligible_for_reminders(days_overdue: 3)
     overdue_by(days_overdue)
       .joins(:patient)
       .merge(Patient.contactable)
-      .left_joins(:communications)
-      .where(communications: {id: nil})
-      .left_joins(:appointment_reminders)
-      .where(appointment_reminders: {id: nil})
+      .left_joins(:notifications)
```
kpethtel, line 69: We might have to make some changes to accommodate this. The reason I added the "where no communications" is to exclude all past appointments. Given that Notification is a new model, this will result in all past appointments receiving new notifications. This could also be fixed by changing
```
def self.overdue_by(number_of_days)
  overdue.where("scheduled_date <= ?", Date.current - number_of_days.days)
end
```
to only search for the scheduled date = date. But that would mean we might never send a reminder in some edge cases (notifications were turned off for a while, sidekiq dropped jobs).

app/models/appointment.rb
kpethtel: I think it achieves the same thing. Unrelated, but I think "attempted" is a poorly named method. It's defined as 
```
  def attempted?
    successful? || in_progress?
  end
  ```
  which means failures weren't attempted. So it's really more like "not failed".

File: config/locales/api/am_ET.yml [hunk 4]
```diff
@@ -14,7 +14,7 @@ am-ET:
       account_locked: "መለያዎ ለሚቀጥሉት %{minutes} ደቂቃዎች ተቆልፏል። እባክዎ ይቆዩ እና እንደገና ይሞክሩ።"
   sms:
     request_otp: "<#>%{otp}የእርስዎ ሲምፕል መተግበሪያ የማረጋገጫ የሚስጥር ቁጥር ነው\n%{app_signature}"
-    appointment_reminders:
+    notifications:
```
kpethtel, line 17: I'm the one who did this, but I'm not sure we needed to change these translations. They are actually appointment reminders.

db/migrate/20210517215935_remove_appointment_from_communication_and_add_patient.rb
kpethtel: We briefly considered that but didn't pursue it. Maybe we should, then we wouldn't have to make patient optional for the notification model.

kpethtel:
We should also consider renaming "reminder_template" to "notification_template" or something like that.

db/migrate/20210517215935_remove_appointment_from_communication_and_add_patient.rb
rsanheim: I've now added a backfill.

db/schema.rb
```diff
@@ -157,19 +139,16 @@
     t.datetime "deleted_at"
     t.string "detailable_type"
     t.bigint "detailable_id"
-    t.uuid "appointment_reminder_id"
-    t.index ["appointment_id"], name: "index_communications_on_appointment_id"
-    t.index ["appointment_reminder_id"], name: "index_communications_on_appointment_reminder_id"
+    t.uuid "notification_id"
+    t.uuid "appointment_id"
+    t.uuid "patient_id"
```
rsanheim, line 144: I no longer think we need `patient_id` on communication, because notifications has all that data.

app/models/appointment.rb
```diff
@@ -148,7 +145,11 @@ def anonymized_data
   end
 
   def previously_communicated_via?(communication_type)
-    communications.latest_by_type(communication_type)&.attempted?
+    latest_notification = notifications.includes(:communications)
+      .where(communications: {communication_type: communication_type})
+      .order(created_at: :desc)
+      .first
+    latest_notification&.communications&.any? { |c| c.attempted? }
```
kpethtel, line 152: I'd prefer this with a leading `!!` since this is a `?` method.

app/services/appointment_notification_service.rb
```diff
@@ -19,20 +19,20 @@ def send_after_missed_visit
...
 
   def create_appointment_reminder(appointment)
-    AppointmentReminder.create!(
-      appointment: appointment,
+    Notification.create!(
+      subject: appointment,
       patient: appointment.patient,
       remind_on: appointment.remind_on,
       status: "scheduled",
-      message: "sms.appointment_reminders.#{communication_type}"
```
kpethtel, line 35: I still think the locale keys are probably the only thing we should leave as is.

app/services/backfill_notifications.rb
```diff
@@ -0,0 +1,15 @@
+class BackfillNotifications
```
kpethtel, line 1: Nice. I think this will save us some pain.

db/migrate/20210518182324_make_notifications_polymorphic_to_subject.rb
```diff
@@ -0,0 +1,8 @@
+class MakeNotificationsPolymorphicToSubject < ActiveRecord::Migration[5.2]
+  def change
+    change_table :notifications do |t|
+      t.remove :appointment_id
+      t.references :subject, polymorphic: true, type: :uuid, null: true
```
kpethtel, line 5: Does this also need a `subject_type`? I suppose this works at least until we add a second subject type.

db/schema.rb
kpethtel: I think you're right.

lib/tasks/scripts/purge_users_data.rb
```diff
@@ -5,7 +5,7 @@ def self.perform
     # These are in a "valid" order so that we don't run into Foreign-Key violations
     models = [BloodPressure,
       BloodSugar,
-      AppointmentReminder,
```
kpethtel, line 8: Good catch.

spec/services/backfill_notifications_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+
+RSpec.describe BackfillNotifications do
```
kpethtel, line 3: I would suggest we delete these sometime in the future, since this script should only be used once and there's no reason to test it forever.

db/migrate/20210518182324_make_notifications_polymorphic_to_subject.rb
rsanheim: This adds the subject_id and subject_type automatically.

kpethtel: I see. Missed that in the schema.

app/services/backfill_notifications.rb
```diff
@@ -0,0 +1,15 @@
...
+  def self.call
+    Communication.find_each do |communication|
+      notification = communication.build_notification
+      notification.message = "sms.notifications.#{communication.communication_type}"
```
kpethtel, line 5: I think this needs to be changed since you changed the locales key back to "appointment_reminders".

db/data/20210518183334_create_notifications_for_communications.rb
```diff
@@ -0,0 +1,9 @@
+class CreateNotificationsForCommunications < ActiveRecord::Migration[5.2]
+  def up
+    BackfillNotifications.call
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
kpethtel, line 7: I don't think it provides any real value, but we could just destroy all notifications as a means of reversing this. But this is probably the better choice.

lib/seed/runner.rb
```diff
@@ -8,7 +8,7 @@ class Runner
     include ActiveSupport::Benchmarkable
     include ConsoleLogger
     SIZES = Facility.facility_sizes
-    FEATURE_FLAGS_ENABLED_BY_DEFAULT = %i[drug_stocks drug_stocks_mobile appointment_reminders]
+    FEATURE_FLAGS_ENABLED_BY_DEFAULT = %i[drug_stocks drug_stocks_mobile notifications]
```
kpethtel, line 11: This won't turn it on in Prod, right? I must not understand the relationship between these defaults and our prod flags.

spec/jobs/appointment_notification/worker_spec.rb
```diff
@@ -3,11 +3,16 @@
...
+      Flipper.enable(:notifications)
       allow(Statsd.instance).to receive(:increment).with(anything)
     end
 
-    let(:reminder) { create(:appointment_reminder, status: "scheduled", message: "sms.appointment_reminders.missed_visit_whatsapp_reminder") }
+    let(:notification) {
+      create(:notification,
+        subject: create(:appointment),
+        status: "scheduled",
+        message: "sms.notifications.missed_visit_whatsapp_reminder")
```
kpethtel, line 14: This should also probably be changed back to appointment_reminders.

kpethtel (commented):
Great work, very thorough!

kpethtel:
Unfortunately, I can't approve this because I opened it, but you have my approval.

app/services/backfill_notifications.rb
rsanheim: oh right, fixed

db/data/20210518183334_create_notifications_for_communications.rb
rsanheim: This is just the default that the data migrate tool gives us, because making them reversible is usually not worth the trouble

lib/seed/runner.rb
rsanheim: Correct, this is just seed data, so will only impact dev and sandbox (when we rerun seeds there).

spec/jobs/appointment_notification/worker_spec.rb
rsanheim: oh ya fixed

spec/services/backfill_notifications_spec.rb
rsanheim: 👍

rsanheim (approved):

PR #2535, opened 2021-05-19
Skip communication in backfill if no patient

rsanheim:
Saw this blow up the data backfill


rsanheim:
There is always one (on sandbox):

```
[5] pry(main)> Communication.select { |c| c.appointment.nil? || c.appointment.patient.nil? }.size
=> 1
[6] pry(main)>
```

PR #2538, opened 2021-05-20
3512 schedule medication reminders

kpethtel:
**Story card:** [ch3512](https://app.clubhouse.io/simpledotorg/story/3512/write-a-script-to-automate-sending-these-notifications-to-selected-patients-everyday)

## Because

We're adding a new type of experiment with different selection criteria and slightly different behavior. We need a way of starting the experiment. 

## Test instructions

Unit tests for now, but we will also test this on sandbox.


config/schedule.rb
```diff
@@ -35,6 +35,18 @@ def local(time)
...
 
+every :day, at: local("02:30 am").utc, roles: [:cron] do
+  if Flipper.enabled?(:live_experiment)
+    runner "ExperimentControlService.start_medication_reminder_experiment('name')" # needs the real experiment name
+  end
+end
+
+every :day, at: local("03:00 am").utc, roles: [:cron] do
+  if Flipper.enabled?(:live_experiment)
+    runner "AppointmentNotification::ScheduleExperimentReminders.perform_later"
```
kpethtel, line 46: I assume perform_later will work here. None of the other scheduled jobs directly use sidekiq. I'm not sure if there's a reason for that.

app/services/experiment_control_service.rb
```diff
@@ -67,11 +58,64 @@ def start_stale_patient_experiment(name, days_til_start, days_til_end, patients_
...
+      return unless Flipper.enabled?(:live_experiment)
+
+      experiment = Experimentation::Experiment.find_by!(name: name, experiment_type: "medication_reminder")
+      experiment.selecting_state!
+
+      eligible_ids = medication_reminder_patients(experiment)
+      if eligible_ids.any?
+        eligible_ids.shuffle!
+
+        daily_ids = eligible_ids.pop(patients_per_day)
```
kpethtel, line 81: I was thinking about maybe adding some logging here and to the other experiment controls. Perhaps log out which ids are being scheduled for the day. And maybe log out "no ids found" if there are no eligible ids for the day.

config/schedule.rb
rsanheim: I think we have been using ActiveJob where possible, so `perform_later` seems fine to me.

config/schedule.rb
```diff
@@ -35,6 +35,18 @@ def local(time)
   runner "PatientDeduplication::Runner.new(PatientDeduplication::Strategies.identifier_and_full_name_match).perform"
 end
 
+every :day, at: local("02:30 am").utc, roles: [:cron] do
+  if Flipper.enabled?(:live_experiment)
+    runner "ExperimentControlService.start_medication_reminder_experiment('name')" # needs the real experiment name
```
rsanheim, line 40: What is the significance of the name? If its just arbitrary, I would make it a constant and encapsulate it inside the experiment object.  None of the clients need to know about it or care.

```ruby
  ExperimentControlService.start_medication_reminder_experiment
...
class Experiment
  MEDICATION_REMINDER = "medication_reminder"
 
  def self.medication_reminder_experiment
    create_or_find_by(name: MEDICATION_REMINDER)
  end
```

app/jobs/appointment_notification/schedule_experiment_reminders.rb
```diff
@@ -2,7 +2,7 @@ class AppointmentNotification::ScheduleExperimentReminders < ApplicationJob
   queue_as :high
 
   def perform
-    return unless Flipper.enabled?(:notifications)
+    return unless Flipper.enabled?(:live_experiment)
```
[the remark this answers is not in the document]
kpethtel: Done.

app/services/experiment_control_service.rb
```diff
@@ -67,11 +58,64 @@ def start_stale_patient_experiment(name, days_til_start, days_til_end, patients_
...
+    end
+
+    # This is not a scientific experiment like active and stale patient experiments. It does not need or have human study
+    # approval and we are not excluding patients under 18 years old. We are leveraging the experimentation framework to
+    # request patients to visit clinics to pick up enough blood pressure medication to last the current covid spike.
+    # We're using the experimentation framework for two reasons:
+    # 1) we want most of the same functionality, including batched daily notification scheduling and the ability to track
+    #    results by leveraging the experimentation and notification models
+    # 2) this is an excellent opportunity to real-world test the experimentation framework
+    def start_medication_reminder_experiment(name, patients_per_day: PATIENTS_PER_DAY)
```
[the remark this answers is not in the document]
kpethtel: Done.

app/services/experiment_control_service.rb
```diff
@@ -67,11 +58,64 @@ def start_stale_patient_experiment(name, days_til_start, days_til_end, patients_
...
+
+      eligible_ids = medication_reminder_patients(experiment)
+      if eligible_ids.any?
+        eligible_ids.shuffle!
+
+        daily_ids = eligible_ids.pop(patients_per_day)
+        daily_patients = Patient.where(id: daily_ids)
+        daily_patients.each do |patient|
+          group = experiment.random_treatment_group
+          schedule_reminders(patient, nil, group, Date.current)
```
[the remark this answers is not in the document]
kpethtel: No longer a concern now that I've separated this into a separate service.

config/schedule.rb
kpethtel: I didn't do this because the experiment needs to be setup manually anyhow. Instead, I just got rid of the name argument and I'm just looking up the first medication reminder experiment. If we end up creating more than one medication reminder experiment, we can address the issue at that time.

app/models/experimentation/experiment.rb
```diff
@@ -6,6 +6,7 @@ class Experiment < ActiveRecord::Base
     validates :name, presence: true, uniqueness: true
     validates :state, presence: true
     validates :experiment_type, presence: true
+    validates :experiment_type, uniqueness: true, if: proc { |experiment| experiment.experiment_type == "medication_reminder" }
```
kpethtel, line 9: It seemed to me that the "find_or_create" approach you suggested was essentially enforcing uniqueness of the medication reminder type, so I chose to explicitly enforce it instead. I did this partly because there's no value in the "create" behavior because a newly created experiment would not be useful because it would not have treatment groups or reminder templates. It would prevent errors but not result in a usable experiment. This approach may not be ideal either. LMK if you have better ideas or want to brainstorm.

config/schedule.rb
```diff
@@ -35,6 +35,18 @@ def local(time)
   runner "PatientDeduplication::Runner.new(PatientDeduplication::Strategies.identifier_and_full_name_match).perform"
 end
 
+every :day, at: local("02:30 am").utc, roles: [:cron] do
```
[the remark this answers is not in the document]
rsanheim: Yeah maybe set this at 1 am and the second job at 4 am to have a bit more leeway?

app/services/experimentation/medication_reminder_service.rb
```diff
@@ -0,0 +1,68 @@
...
+    PATIENTS_PER_DAY = 10_000
+
+    class << self
+      def schedule_daily_notifications(patients_per_day: PATIENTS_PER_DAY)
+        return unless Flipper.enabled?(:experiment)
+
+        experiment = Experimentation::Experiment.find_by!(state: [:new, :running], experiment_type: "medication_reminder")
+        experiment.selecting_state!
+
+        eligible_ids = medication_reminder_patients(experiment)
```
rsanheim, line 19: We should probably pass through the `patients_per_day` to the `medication_reminder_patients` method to apply the limit there in SQL -- no reason to pass back thousands of patients if we just need 100 for the first batch.

rsanheim (approved):

app/models/experimentation/experiment.rb
rsanheim: This seems fine to me for now.

app/services/experimentation/medication_reminder_service.rb
kpethtel: I think we didn't do that in the A/B because we wanted to randomize before limiting. But that doesn't really matter here, so you're probably right about that.

PR #2541, opened 2021-05-20
3557: feature flap whatsapp notifications

kpethtel:
**Story card:** [ch3557](https://app.clubhouse.io/simpledotorg/story/3557/feature-flag-whatsapp-in-new-notification-system)

## Because

I made a mistake when developing this. It should be feature flagged, not geographical. 

## This addresses

We may need to send medical reminders without whatsapp, which will be easier if we can simply turn off the whatsapp feature flag in India.

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -41,9 +41,7 @@ class Notification < ApplicationRecord
     if preferred_communication_method && !previously_communicated_by?(preferred_communication_method)
       return preferred_communication_method
     end
-    unless previously_communicated_by?(backup_communication_method)
-      return backup_communication_method
-    end
+    return backup_communication_method unless previously_communicated_by?(backup_communication_method)
     nil
   end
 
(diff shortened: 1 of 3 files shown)


app/models/notification.rb
```diff
@@ -54,7 +52,7 @@ def previously_communicated_by?(method)
   end
 
   def preferred_communication_method
-    CountryConfig.current[:name] == "India" ? "missed_visit_whatsapp_reminder" : nil
+    Flipper.enabled?(:whatsapp_appointment_reminders) ? "missed_visit_whatsapp_reminder" : nil
```
rsanheim, line 55: This was the name of the feature flag before, correct?

rsanheim (approved):

app/models/notification.rb
kpethtel: This is still used in the appointment notification service.

kpethtel: I do think we need to separate our flags more carefully. We've already separated out "notifications" into a separate flag, and this should probably be separated into simply a "whatsapp" flag.

PR #2542, opened 2021-05-21
3563: add translations for covid lockdown medication reminders

kpethtel:
**Story card:** [ch3563](https://app.clubhouse.io/simpledotorg/story/3563/add-medication-reminder-translations)

## Because

We want medication reminders to be localized.


rsanheim (approved):
I can't speak to the translations, but the yaml changes look good 👍

PR #2544, opened 2021-05-24
Revert data migration to backfill notifications

prabhanshuguptagit:
**Story card:** [ch3585](https://app.clubhouse.io/simpledotorg/story/3585/backfill-notifications-from-communications)

## Because

The data migration to backfill `Notification` goes through all `Communication`s sequentially which is very slow. The data migration would be expected to take a few days. 

## This addresses

The migration is not required for `Notification` to be released. This drops the backfill data migration to unblock master. We should do the backfill at some point in a way that works for prod – IHCI has ~10m communications. 

Maintenance story:
https://app.clubhouse.io/simpledotorg/story/3585/backfill-notifications-from-communications


harimohanraj89:
@prabhanshuguptagit do we need to rollback the migration on your local machine? I expected to see the `data_schema.rb` file reverted as well.

prabhanshuguptagit:
@harimohanraj89 The `data_schema` was never committed to master, so we're good. No need to revert it.

db/schema.rb
```diff
@@ -141,6 +141,7 @@
     t.string "detailable_type"
     t.bigint "detailable_id"
     t.uuid "notification_id"
+    t.index ["appointment_id"], name: "index_communications_on_appointment_id"
```
prabhanshuguptagit, line 144: This is a previously uncommitted schema change.

db/schema.rb
```diff
@@ -130,7 +130,7 @@
   end
 
   create_table "communications", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
-    t.uuid "appointment_id", null: false
+    t.uuid "appointment_id"
```
harimohanraj89, line 133: As discussed offline, this is because a previous migration didn't check in its update to `schema.rb`

harimohanraj89 (approved):
:shipit:

PR #2545, opened 2021-05-24
Add twilio test whatsapp number

prabhanshuguptagit:
**Story card:** 

## Because

Twilio has different default phone numbers for their sandbox whatsapp and sandbox SMS accounts.

## This addresses

Lets `notification_service` pick the right phone number for Twilio Sandbox.


timcheadle (changes requested):
LGTM. Also need to update `spec/services/notification_service_spec.rb` for whatsapp tests.

timcheadle (approved):

kpethtel (approved):

PR #2546, opened 2021-05-24
3457: provide mechanism for terminating experiments early

kpethtel:
**Story card:** [ch3457](https://app.clubhouse.io/simpledotorg/story/3457/create-mechanism-for-ending-experiment-early)

## Because

We would like a quick and easy mechanism for ending experiments early in case we realize there's a problem.


app/services/experiment_control_service.rb
```diff
@@ -55,6 +55,16 @@ def schedule_daily_stale_patient_notifications(name, patients_per_day: PATIENTS_
...
 
+    def terminate_experiment(name)
+      experiment = Experimentation::Experiment.find_by!(name: name)
+      experiment.complete_state!
+
+      notifications = experiment.notifications.where(status: ["pending", "scheduled"])
+      notifications.find_each do |notification|
+        notification.status_cancelled!
+      end
+    end
```
ssrihari, line 66: - Where do we expect this to be called?
- Will synchronously cancelling 1000s of notifications be slow, and do we need to think about adding a tx around this?
- Should we consider using the experiment's state to stop sending out notifications, and then marking them cancelled asynchronously?

kpethtel: 1) my expectation is that this would be called from the rails console if we ever need to end an experiment early. 
2) it might be a bit slow. I'm not sure what a transaction would do here. Are you suggesting we'd roll back all the changes if one of them fails? I think it would make more sense to raise a loud error, let someone fix the problem, then run it again from where they left off. 
3) i think a guard in the worker against the experiment being complete is a good idea. It's already guarded against the notification being cancelled and the feature flag being off, but there are edge cases where those two guards might not be sufficient. For example, if this job takes a long time and we have two experiments running simultaneously with the same feature flag. I'll add that. 
We could cancel the notifications async, but I'm not sure it's worth the additional complexity. As we saw today, background jobs are much harder to debug.

app/models/patient.rb
```diff
@@ -44,6 +44,8 @@ class Patient < ApplicationRecord
   has_many :appointments
   has_many :notifications
   has_many :treatment_group_memberships, class_name: "Experimentation::TreatmentGroupMembership"
+  has_many :treatment_groups, through: :treatment_group_memberships, class_name: "Experimentation::TreatmentGroup"
```
kpethtel, line 47: This is not related to the ticket. It's something I found very helpful while doing something else, so I decided to add it.

harimohanraj89 (changes requested):

app/models/notification.rb
```diff
@@ -46,6 +46,8 @@ def localized_message
   end
 
   def next_communication_type
+    # guarding against experiment state to prevent race condition
+    return nil if status_cancelled? || experiment&.cancelled_state?
```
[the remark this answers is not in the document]
kpethtel: Yes. Explained in another discussion thread.

app/services/experiment_control_service.rb
kpethtel: This PR keeps getting put on the back burner. Seems like we're all good with points 1 and 2. So to address point 3... 

After Srihari's initial feedback, I decided to add the "cancelled" state option to the experiment model. That allows us to guard against sending notifications if the experiment is cancelled, which I believe handles Srihari's concerns that the job may take a while and could result in us sending notifications unintentionally because we hit the "abort" button but it couldn't cancel all the scheduled notifications before they could send. 

Maybe this helps to clarify. We have two basic kinds of experiments (not counting medication reminders). 
- current patients: 
  - start and end dates: don't really do anything. They're mostly used for historical reference.
  - `experiment.state_completed?` has no impact on the experiment. Can't have more than one "running" experiment of the same type at a time, so you do need to change an experiment to complete or canceled in order to run another active patient experiment.
  - `experiment.state_cancelled?` prevents scheduled notifications from being sent
- stale patients: 
  - start and end dates: used to guard against selecting more patients for the experiment. Selection only occurs if we're within the date range. End date is also used to mark the experiment as complete (see this PR, it's new functionality).
  - `experiment.state_completed?` daily selection does not occur if state is complete
  - `experiment.state_cancelled?` prevents scheduled notifications from being sent and also prevents daily selection from occurring. 
  
Does that clear anything up? It's all complicated by the fact that we have two different experiment types that run in very different ways.

spec/services/experiment_control_service_spec.rb
```diff
@@ -316,5 +315,51 @@
...
+      create(:blood_pressure, patient: patient1, device_created_at: 100.days.ago)
+
+      expect {
+        ExperimentControlService.schedule_daily_stale_patient_notifications(experiment.name)
+      }.to change { experiment.reload.state }.from("new").to("complete")
+      expect(Experimentation::TreatmentGroupMembership.count).to eq(0)
+    end
+  end
+
+  describe "self.abort_experiment" do
```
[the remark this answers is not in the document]
kpethtel: I always use `#` for instance methods but hadn't noticed that `.` is the convention for class methods. I'll make that change later because it's in multiple places. Fewer keystrokes to do nothing right now. Thanks for the tip, though.

app/services/experiment_control_service.rb
harimohanraj89: @kpethtel Understood, so cancelling an experiment needs to put an _immediate_ end to messages that were going to go out as part of the experiment, even if the `Notification`s have already been scheduled for the future.

I think this approach is 👌

harimohanraj89 (approved):

PR #2547, opened 2021-05-25
fix notification localization when notification has no subject

kpethtel:
**Story card:**

## Because

We made a mistake when refactoring appointment reminders into notifications and making appointment a polymorphic relationship and didn't handle the case when subject is nil. I'm fixing this in the ugliest way possible to ensure that we come up with a better long-term fix. Tim and I were talking earlier today for the need to refactor some of our communications modeling, and I think this ties into that issue. We need to make notifications less tied to the text within the notification.

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -32,6 +32,14 @@ class Notification < ApplicationRecord
         appointment_date: subject.scheduled_date,
         locale: subject.facility.locale
       )
+    when nil
+      # this is a temporary fix. we need to figure out a better way of making the message more flexible
+      I18n.t(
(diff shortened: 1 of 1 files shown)


ssrihari (approved):
Lets create a card for the permanent fix, and add it to the upcoming iteration so it doesn't slip.

prabhanshuguptagit:
Story for permanent fix: 
https://app.clubhouse.io/simpledotorg/story/3606/make-notifications-less-tied-to-the-text-in-the-notification

harimohanraj89:
Added to iteration 36.

PR #2548, opened 2021-05-25
Make appointment optional from communication

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/3566/deployment-plan-send-medication-reminders-in-india

## Because

Appointment communication was made optional earlier in the DB but got missed in the model.

## This addresses

Makes appointment - communication relation optional in the model.


harimohanraj89 (approved):
Pending specs, this change looks good.

PR #2550, opened 2021-05-25
backfill test for creating communication with twilio detail

kpethtel:
## Because

This creates a test for changes that were made this morning while trying to get medication reminders to send.


prabhanshuguptagit (approved):
👌🏼

PR #2552, opened 2021-05-26
Make notification scheduled again if it's communication fails

prabhanshuguptagit:
**Story card:** [ch3633](https://app.clubhouse.io/simpledotorg/story/3633/fix-sms-fallback)

## Because

We set a notification's status to `sent` once a communication is sent for it. When a whatsapp communication fails to be delivered, we queue a fallback job to send an SMS. However the fallback job makes sure that the notification's status is `scheduled` before sending the message. Since the failed communication does, the fallback job does not send the SMS. 

## This addresses

When a communication fails, this marks the notification back as `scheduled` so the fallback job can send an SMS.

### DIFF
--- a/app/controllers/api/v3/twilio_sms_delivery_controller.rb
+++ b/app/controllers/api/v3/twilio_sms_delivery_controller.rb
@@ -13,10 +13,10 @@ class Api::V3::TwilioSmsDeliveryController < ApplicationController
     metrics.increment(event)
 
     reminder = twilio_message.communication.notification
-    communication_type = reminder&.next_communication_type
-    if twilio_message.unsuccessful? && communication_type
-      notification = twilio_message.communication.notification
-      AppointmentNotification::Worker.perform_at(Communication.next_messaging_time, notification.id)
+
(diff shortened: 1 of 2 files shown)


harimohanraj89 (commented):
Can we write an integration-y spec, probably in that controller, that sets up a failed whatsapp message, and ensures that an SMS is _sent_ (not just queued)?

app/controllers/api/v3/twilio_sms_delivery_controller.rb
```diff
@@ -13,10 +13,10 @@ def create
     metrics.increment(event)
 
     reminder = twilio_message.communication.notification
-    communication_type = reminder&.next_communication_type
-    if twilio_message.unsuccessful? && communication_type
-      notification = twilio_message.communication.notification
```
kpethtel, line 18: Weird, I hope the duplicated line here was from the model renaming last week.

kpethtel (approved):

prabhanshuguptagit:
Added spec.

PR #2556, opened 2021-05-28
3641: log prescription drug creation when patient is part of medicati…

kpethtel:
…on reminder experiment

**Story card:** [ch3641](https://app.clubhouse.io/simpledotorg/story/3641/log-when-a-prescription-drug-is-created-for-medication-reminder-experiment-patients)

## Because

We want to be able to track the success rate of medication reminders.

### DIFF
--- a/app/models/prescription_drug.rb
+++ b/app/models/prescription_drug.rb
@@ -47,11 +47,11 @@ class PrescriptionDrug < ApplicationRecord
   # this code should only be temporary, as it's part of monitoring a one-off experiment
   # https://app.clubhouse.io/simpledotorg/story/3642/remove-unnecessary-medication-reminder-code
   def log_medication_reminder_success
-    experiment_membership = patient.treatment_group_memberships.find {|membership| membership.treatment_group.experiment.experiment_type == "medication_reminder" }
+    experiment_membership = patient.treatment_group_memberships.find { |membership| membership.treatment_group.experiment.experiment_type == "medication_reminder" }
     return unless experiment_membership
     experiment = experiment_membership.treatment_group.experiment
     notification = patient.notifications.find_by(experiment_id: experiment.id)
-    communication = notification.communications.find {|communication| communication.successful? }
+    communication = notification.communications.find { |communication| communication.successful? }
     return unless communication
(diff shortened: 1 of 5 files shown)


app/models/prescription_drug.rb
```diff
@@ -39,4 +41,28 @@ def anonymized_data
      medicine_name: name,
      dosage: dosage}
   end
+
+  private
+
+  # this code should only be temporary, as it's part of monitoring a one-off experiment
+  # https://app.clubhouse.io/simpledotorg/story/3642/remove-unnecessary-medication-reminder-code
+  def log_medication_reminder_success
```
kpethtel, line 49: This is not my finest code and I'm glad it's temporary. Hopefully it accomplishes our goal here.

app/models/prescription_drug.rb
```diff
@@ -39,4 +41,26 @@ def anonymized_data
...
+    experiment = patient.experiments.find_by(experiment_type: "medication_reminder")
+
+    return unless experiment
+    notification = patient.notifications.find_by(experiment_id: experiment.id)
+    communication = notification.communications.find { |communication| communication.successful? }
+    return unless communication
+    time_till_visit = device_created_at - communication.detailable.delivered_on
```
prabhanshuguptagit, line 56: 👌🏼

app/models/prescription_drug.rb
prabhanshuguptagit: If we don't want to rely on callbacks, a couple of other ways: 
- Log this in the prescription drug sync_from_user controller
- Run a nightly job that collects all prescription drugs synced in the day, and logs this 

I am leaning towards option 2 since it'll be isolated.

prabhanshuguptagit (commented):
The criteria for success looks good 💯 . Have a suggestion about the callback, approving to unblock.

prabhanshuguptagit:
As you mentioned on slack, since a patient can pick up more than one prescription drug in a day, aggregating this in SQL might be easier? A SQL query on metabase might be a better option here.

kpethtel:
Putting this on metabase instead.

PR #2560, opened 2021-06-01
Fix timezone inconsistencies

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/3660/fix-timezone-inconsistencies-in-current-reporting

* Wrap reports cache warming in the right time zone
* Fix boundaries of LTFU scopes, with rigorous test coverage

### Cache warming

We used to warm the cache in Sidekiq, and we explicitly [set the application's timezone](https://github.com/simpledotorg/simple-server/blob/master/config/initializers/sidekiq.rb#L1-L5) around all Sidekiq jobs, similar to how we do [for web requests](https://github.com/simpledotorg/simple-server/blob/master/app/controllers/reports/regions_controller.rb#L219-L226).

When we un-backgrounded the warming job recently (#2423), we inadvertently started warming our caches in UTC, which would place patient activity in the wrong day/time if it occurred between midnight and 530am IST. It's unlikely to happen in reality, but an inconsistency we should avoid. It is definitely prevalent in our Sandbox data, and makes it hard for us to compare new reporting work against existing dashboard data.

### LTFU Queries

These are tricky. There are two separate problems that come together on this one.
1. The LTFU query makes _date_ comparisons to include/exclude BPs from a 1-year window. It should instead use precise _times_ to include and exclude BPs.
2. If we use times throughout, the LTFU query has another tricky bit that needs care. It joins the `patients` app table with the `latest_blood_pressures_per_patient_per_months` matview table. We then compare timestamps from both tables against some cutoffs. Comparing cutoffs for the patients table is straightforward, but the matview timestamps are stored as timestamp-without-tz, and have been cast to the local reporting time. ie, a BP recorded at 6pm India time is stored as `18:00` without timezone in the matview. When Rails crafts a query to compare `bp_recorded_at` against our cutoff, it converts our cutoff to UTC and compares just the timestamps, but this is incorrect. Hence, we have to manually ensure that the BP cutoffs are shifted by the UTC amount before comparing against matview timestamps.

### Overarching concerns/patterns/conventions

This is, personally for me, the first foray into some gnarly timezone issues that we'll probably continue to face, and we'll need some strong conventions to follow to keep our data consistent and accurate. Here's a proposed convention that Srihari and I have been working on.

* Reporting tables: DB has local time, use ts w/o tz in Rails
* App tables: DB has UTC, use ts with tz in Rails
* When joining: ideally don't join app tables with reporting tables. If you do, ensure all TS are in reporting tz
* When querying reporting tables for durations: use computed durations, or computed months, not timestamps. ie, rely on fields like `month` and `year`, rather than using `recorded_at`, as much as possible. 

### Other inconsistencies/questions

The next big area we'll have to take a look at is how the Groupdate gem is applied to our matviews. If we're setting the groupdate gem's timezone to "Asia/Kolkata", and asking it to group data in the matview, is it incorrectly assuming the timestamps inside are in UTC, and shifting those times by _another_ 5.5 hours?

Does modifying the LTFU queries' boundaries in isolation here cause inconsistency against control rate data? Eg. Patients under care + LTFU + dead + transferred = total assigned patients? Or will this change propagate gracefully to the other queries?


harimohanraj89:
Oops not ready yet.

spec/models/patient_spec.rb
```diff
@@ -1,7 +1,9 @@
 require "rails_helper"
 
 describe Patient, type: :model do
+  let(:reporting_timezone) { CountryConfig.current[:time_zone] }
```
ssrihari, line 4: How is this config different from `ANALYTICS_TIME_ZONE`?

harimohanraj89: Oof, great question. It's not/shouldn't be. We've got:
* `ANALYTICS_TIME_ZONE` env var
* `CountryConfig.current[:time_zone]` (from country configurations, based on `DEFAULT_COUNTRY` env var)
* ...and `Period::ANALYTICS_TIME_ZONE`, which is CountryConfig with a default value

I'm not sure if we should use CountryConfig here, or Period here, but I'm fairly certain we can start phasing out `ANALYTICS_TIME_ZONE`.

spec/models/patient_spec.rb
```diff
@@ -199,6 +201,91 @@ def refresh_views
 
         expect(described_class.ltfu_as_of(Time.current)).not_to include(not_ltfu_patient)
       end
+
+      describe "timezone-specific boundaries" do
```
harimohanraj89, line 205: Do we need specs that include/exclude dead, migrated, transferred patients?

spec/models/patient_spec.rb
```diff
@@ -223,6 +310,91 @@ def refresh_views
...
+
+          expect(described_class.not_ltfu_as_of(beginning_of_month)).to include(not_ltfu_patient)
+          expect(described_class.not_ltfu_as_of(end_of_month)).to include(not_ltfu_patient)
+
+          expect(described_class.not_ltfu_as_of(beginning_of_month)).not_to include(ltfu_patient)
+          expect(described_class.not_ltfu_as_of(end_of_month)).not_to include(ltfu_patient)
+        end
+      end
     end
   end
```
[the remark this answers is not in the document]
prabhanshuguptagit: > Does modifying the LTFU queries' boundaries in isolation here cause inconsistency against control rate data

Since the LTFU exclusion happens at the topmost layer, the numbers should ideally add up (just by definition). Will be good to verify 😛

app/models/concerns/patient_reportable.rb
```diff
@@ -35,8 +43,9 @@ module PatientReportable
       end
     end
 
-    def self.ltfu_time_ago(date)
-      (date.to_date - LTFU_TIME).end_of_month
+    # Returns a plain timestamp to be interpreted in reporting time, for use with materialized views
+    def self.reporting_timestamp(time)
```
prabhanshuguptagit, line 47: Add this to `Period`?

spec/models/patient_spec.rb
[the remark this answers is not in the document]
harimohanraj89: That's great to hear. Yeah, I suspected, but we'll verify consistency before we ship anything of course.

prabhanshuguptagit:
> App tables: DB has UTC, use ts with tz in Rails

Are we thinking of changing the type for all app tables?

rsanheim (changes requested):
> This is, personally for me, the first foray into some gnarly timezone issues that we'll probably continue to face, and we'll need some strong conventions to follow to keep our data consistent and accurate. Here's a proposed convention that Srihari and I have been working on.
> 
> * Reporting tables: DB has local time, use ts w/o tz in Rails
> * App tables: DB has UTC, use ts with tz in Rails

😱 😱 Noooo, please no inconsistent timezones between two sets of tables. If that is already the case with old mat views, lets consider that 'legacy' and get things consistent w/ the new reporting schema.

> * When joining: ideally don't join app tables with reporting tables. If you do, ensure all TS are in reporting tz
> * When querying reporting tables for durations: use computed durations, or computed months, not timestamps. ie, rely on fields like `month` and `year`, rather than using `recorded_at`, as much as possible.

Somewhat of an aside: we actually made changes recently to make sure to [use timestamps](https://github.com/simpledotorg/simple-server/pull/2181) for some similar cases when using `recorded_at`, which solved this issue.

Anyway, I *strongly* disagree with having different conventions for what timezone we store in our database between app tables and reporting tables. It will be a source of subtle bugs when we pull data out and forget to use the correct Postgres timezone conversion methods, or especially for the joining case you mention that it sure to come up. 

The only way I could see a case for using different timezones is if our reporting database was actually a separate database altogether, in which case we wouldn't have to worry about joins and would have different code paths grabbing that data.

I realize we have "reporting tables" already in the current mat views and the new reporting schema, and it sounds like we _already_ have this mismatched timezones.  But for the _new_ schema we should get things consistent and use UTC everywhere to avoid confusing and more issues going forward.

I'll try to schedule a zoom soon to discuss this, as it would be great for all of us to get on the same page moving forward on this.

ssrihari:
We hear you @rsanheim, the consistent timezones in the DB tables is definitely a problem worth discussing. We won't solidify the convention we proposed until we're all on the same page.

These fixes are still relevant to the "legacy" materialised views as you mentioned, and we'd like to take this PR to completion for data consistency, and so that we can compare the new reporting schema results to the correct numbers.

-- @harimohanraj89 and @ssrihari

app/models/concerns/patient_reportable.rb
```diff
@@ -9,19 +9,27 @@ module PatientReportable
     scope :with_hypertension, -> { joins(:medical_history).merge(MedicalHistory.hypertension_yes).distinct }
```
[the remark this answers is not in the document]
ssrihari: Yup, that's right. This behaviour will be fixed with this PR.

app/models/concerns/patient_reportable.rb
ssrihari: We're not sure where this will go yet, Period isn't the right place because it doesn't really deal with timezones or "reporting". There are a few other things like decided which timezone, using that timezone, etc, in the reporting path that will need extraction as well, but it's not clear where they'll go yet.

We'll keep this here for now, and keep an eye out for when/where to extract these. This is a workaround, and is not meant to set the precedent for other places/contexts.

spec/models/patient_spec.rb
ssrihari: Fixed in 1aa47d9254236c297e059e239b7e53dd6056ddab. We use `Period::ANALYTICS_TIME_ZONE` everywhere.

spec/models/patient_spec.rb
ssrihari: Nope, verified that we test this in the control rate service.

spec/models/patient_spec.rb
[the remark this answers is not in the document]
ssrihari: ✅

app/models/concerns/patient_reportable.rb
```diff
@@ -35,8 +43,9 @@ module PatientReportable
       end
     end
 
-    def self.ltfu_time_ago(date)
-      (date.to_date - LTFU_TIME).end_of_month
+    # Returns a plain timestamp to be interpreted in reporting time, for use with materialized views
+    def self.reporting_timestamp(time)
+      time + time.utc_offset
```
rsanheim, line 48: I switched my console to IST to try and play out what will happen here, assuming the reporting date is today.

```
31] pry(main)> Time.zone = "Asia/Kolkata"
=> "Asia/Kolkata"
[32] pry(main)> time = Time.current
=> Thu, 03 Jun 2021 21:43:12 IST +05:30
[33] pry(main)> offset = time + time.utc_offset
=> Fri, 04 Jun 2021 03:13:12 IST +05:30
```

I _think_ that is what we want, but it still reads really weird to _add_ the utc_offset to our application time, which is already in the offset time.  Is there a better way we could express this?

app/models/concerns/patient_reportable.rb
```diff
@@ -35,8 +43,9 @@ module PatientReportable
       end
     end
 
-    def self.ltfu_time_ago(date)
-      (date.to_date - LTFU_TIME).end_of_month
+    # Returns a plain timestamp to be interpreted in reporting time, for use with materialized views
```
rsanheim, line 46: What is a "plain timestamp" ? Can we expand on this comment for future timezone struggles?

```
Our legacy materialized views store their timestamps in _local time_, so IST for India for example. Because of this we need to adjust our application provided timestamps by the UTC offset, otherwise Rails automatic timezone conversion do the wrong thing, as it assumes times from the database are all UTC.
```

After writing all that out, this would really be much more clear if we always let Postgres convert timezones from the legacy mat views to UTC, because then Rails would do the right thing.  That would probably slow down all these queries though, and require more careful SQL.

app/models/concerns/patient_reportable.rb
```diff
@@ -9,19 +9,27 @@ module PatientReportable
...
-             AND #{sanitize_sql(["bp_recorded_at > ? AND bp_recorded_at < ?", (date.to_date - LTFU_TIME).end_of_month, date])}")
-        .where("bp_recorded_at IS NULL AND patients.recorded_at < ?", (date.to_date - LTFU_TIME).end_of_month)
+             AND #{sanitize_sql(["bp_recorded_at > ? AND bp_recorded_at < ?", bp_start, bp_end])}")
+        .where("bp_recorded_at IS NULL AND patients.recorded_at <= ?", patient_registration_cutoff)
         .distinct(:patient_id)
     end
 
-    scope :not_ltfu_as_of, ->(date) do
+    scope :not_ltfu_as_of, ->(time) do
+      bp_start = reporting_timestamp((time - LTFU_TIME).end_of_month)
```
rsanheim, line 25: A better name for this method could be `legacy_matview_timestamp`, to be very clear when this should be used.

app/queries/control_rate_cohort_query.rb
```diff
@@ -80,7 +80,7 @@ def overall_controlled_bps_per_facility
   def overall_patients
     @overall_patients ||=
       Patient
-        .for_reports(exclude_ltfu_as_of: Date.today.end_of_month)
+        .for_reports(exclude_ltfu_as_of: Period.month(Date.today).end)
```
rsanheim, line 83: Mind adding a little factory method to Period to make this easier?

```ruby
Period.current.end
```

It could default to month period, because we use that in 95% of cases...we could add an optional arg version for quarters if we want:

```ruby
Period.current(:quarter).end
```

rsanheim: Oh, and this should use `end_time` to be more explicit here (its aliased right now but `end` should really go away at some point...).

app/queries/earliest_patient_data_query.rb
```diff
@@ -23,6 +23,7 @@ def call(region)
         AND (patients.assigned_facility_id in :facility_ids OR patients.registration_facility_id in :facility_ids)
     SQL
     return nil if sql.values.first.blank?
-    Time.zone.parse(sql.values.first)
+    # we store timestamps in UTC in the app tables
+    Time.find_zone("UTC").parse(sql.values.first).in_time_zone(Time.zone)
```
rsanheim, line 27: I think you can instead use [force_timezone](https://github.com/github/github-ds/blob/c5962e5f59ea97924278a01ea8d85edd29aa2386/lib/github/sql.rb#L143-L144) from GitHub::SQL... I believe it would be `force_timezone: :utc` win the constructor above.

app/services/reports/region_cache_warmer.rb
```diff
@@ -32,19 +32,21 @@ def call
     end
 
     def warm_region_cache(region)
-      RequestStore.store[:bust_cache] = true
+      Time.use_zone(Period::ANALYTICS_TIME_ZONE) do
```
rsanheim, line 35: 👍

spec/models/patient_spec.rb
```diff
@@ -199,6 +201,89 @@ def refresh_views
...
+        end
+
+        it "bp cutoffs for a year ago" do
+          # For any provided date in June in the local timezone, the LTFU BP cutoff is the end of June 30 of the
+          # previous year in the local timezone.
+          #
+          # Eg. For any date provided in June 2021, the cutoff is the June 30-Jul 1 boundary of 2020
+
+          long_ago = 5.years.ago
+          under_a_year_ago = Time.zone.local(2020, 7, 1, 0, 0, 1) # Beginning of July 1 2020 in local timezone
```
rsanheim, line 221: You could use iso8601 strings here ? It would make this setup much more clear, and it looks like parse works fine with a little whitespace to make it more readable:

comparison:


```ruby
[52] pry(main)> under_a_year_ago = Time.zone.local(2020, 7, 1, 0, 0, 1)
=> Wed, 01 Jul 2020 00:00:01 IST +05:30
[53] pry(main)> under_a_year_ago = Time.zone.parse("2020-07-01 T00:00:01 +05:30")
=> Wed, 01 Jul 2020 00:00:01 IST +05:30
```

rsanheim (commented):
Some feedback and questions.  thanks for adding the detailed specs, those will be very useful.

rsanheim:
> We're comparing timestamps incorrectly in several places because of local-cast timestamps in the matviews. I wonder if it's worth the effort to reverse these legacy decisions _right away_?

I thought a little more about this, and I wonder if maybe the better route is to just convert all legacy mat views to store as UTC and not deal with all this conversion.  It's not a dangerous data transition, since they are regenerating every night - it would just be a matter of slight changes to the legacy mat views, and it would also help with verifying correctness w/ the new pipeline.

I'll spike it out today just to see.

app/models/concerns/patient_reportable.rb
harimohanraj89: Yeah, it's weird because we add an _extra_ 5.5 hours, because we know Rails is going to accidentally _subtract_ 5.5 hours when it runs the query.

In any case, #2567 is the simpler approach.

app/queries/earliest_patient_data_query.rb
harimohanraj89: That's the elegant solution we were looking for ✨

harimohanraj89: Tried this a bit, but couldn't quite get it to do what we wanted. I'm going to stick to this explicit approach for now.
* Parse the timestamp, interpreting it as a UTC timestamp
* Give me that real time in IST (to maintain the contract of this query)

harimohanraj89: Extracted to https://github.com/simpledotorg/simple-server/pull/2570

app/services/reports/region_cache_warmer.rb
harimohanraj89: Extracted to https://github.com/simpledotorg/simple-server/pull/2569

harimohanraj89:
The changes in this PR have all been extracted to smaller PRs:

* #2567 
* #2569 
* #2570 

The only remaining info in this PR is the "TODO"s. I've verified that all the TODOs will no longer need fixing once the matview timestamps are updated.

ssrihari:
All smaller PRs have been merged, closing this.

PR #2561, opened 2021-06-01
Fix download name spec

harimohanraj89:
**Story card:** N/A

We were mistakenly asserting the full month name to be in the file download, but we only put the month abbreviation in. We missed this when we added the spec because we were in the month of... yeah... May when we wrote it. 

This spec started failing today. June 1.


rsanheim:
😆 oh the fun bugs we can create

PR #2562, opened 2021-06-01
Automate covid medication reminders

prabhanshuguptagit:
**Story card:** [ch3566](https://app.clubhouse.io/simpledotorg/story/3566/deployment-plan-send-medication-reminders-in-india)

## Because

We want to automate sending COVID medication reminders over the next few days.

## This addresses

Sends 20,000 medication reminders every day

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -49,6 +49,10 @@ every :monday, at: local("6:00 am"), roles: [:cron] do
   end
 end
 
+every :day, at: local("07:00 am"), roles: [:cron] do
+  rake "appointment_notification:covid_medication_reminders[20000]"
+end
+
 every 2.minutes, roles: [:cron] do
   runner "TracerJob.perform_async(Time.current.iso8601, false)"
 end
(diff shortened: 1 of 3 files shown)


config/schedule.rb
```diff
@@ -49,6 +49,10 @@ def local(time)
   end
 end
 
+every :day, at: local("07:00 am"), roles: [:cron] do
+  rake "appointment_notification:covid_medication_reminders[20000]"
```
rsanheim, line 53: The rake argument syntax always looks weird to me.

lib/tasks/scripts/covid_medication_reminders.rb
```diff
@@ -0,0 +1,10 @@
+class CovidMedicationReminders
+  def self.call(number_of_patients: 10_000)
+    if CountryConfig.current[:name] != "India" || !SimpleServer.env.production? || !Flipper.enabled?(:experiment)
+      raise "Cannot send Covid medication reminders in this env"
```
rsanheim, line 4: This will give a confusing error message if it hits the last branch for the feature flag being disabled...may be worth having its own message for that.

config/schedule.rb
kpethtel: It is weird. And it gives ambiguous errors in zsh.

PR #2564, opened 2021-06-02
Patient Import Enhancements

harimohanraj89:
**Story card:** 

[ch3681](https://app.clubhouse.io/simpledotorg/story/3681/don-t-break-if-medical-history-question-is-blank)
[ch3680](https://app.clubhouse.io/simpledotorg/story/3680/support-y-n-m-f)
[ch3496](https://app.clubhouse.io/simpledotorg/story/3496/skip-rows-that-don-t-have-a-regn-date)

* Skip row if registration date is missing
* Don't break if diagnosis answers are missing
* Recognize gender abbreviations
* Recognize Y/N as Yes No for "Died?"


app/services/patient_import/spreadsheet_transformer.rb
```diff
@@ -266,6 +266,23 @@ def medication_by_name_dosage_and_frequency(name)
...
+        "female"
+      when "t", "transgender"
+        "transgender"
+      else
+        value
+      end
+    end
+
+    def history(value)
+      value.presence&.downcase || "unknown"
```
kpethtel, line 283: The one thing that stands out to me is that we're using presence here and not in other places (like gender). The only difference I can think of is that `"".presence&.downcase` will equal nil and `""&.downcase` will be an empty string. If that's not meaningful, I'd suggest we make it more uniform.

spec/services/patient_import/spreadsheet_transformer_spec.rb
```diff
@@ -215,4 +215,48 @@
       expect(patient[:prescription_drugs].first).not_to have_key(:rxnorm_code)
     end
   end
+
+  context "for abbreviated genders" do
```
kpethtel, line 219: I don't see any tests for the `else` case.

app/services/patient_import/spreadsheet_transformer.rb
```diff
@@ -266,6 +266,23 @@ def medication_by_name_dosage_and_frequency(name)
...
+    def gender(value)
+      case value.downcase
+      when "m", "male"
+        "male"
+      when "f", "female"
+        "female"
+      when "t", "transgender"
+        "transgender"
+      else
+        value
```
kpethtel, line 278: Given that this is a data import, do we want to have an else case that allows the value to pass through? Will that result in unexpected values in the database?

spec/services/patient_import/spreadsheet_transformer_spec.rb
```diff
@@ -215,4 +215,48 @@
...
+    it "recognizes 't'" do
+      params = PatientImport::SpreadsheetTransformer.call(data, facility: facility)
+
+      patient = params.find { |p| p[:patient][:full_name] == "Gender Abbrev T" }.deep_symbolize_keys
+
+      expect(patient[:patient][:gender]).to eq("transgender")
+    end
+  end
+
+  context "for abbreviated died? status" do
```
kpethtel, line 245: These tests don't exactly match the code. The code converts "y" and "yes" to dead and everything else to active. We should add a test for "yes" and perhaps make it clear in tests that it doesn't matter if it says "N" or "Alive" or "Deceased", they all translate to "active"

kpethtel (approved):
A few thoughts, but nothing blocking. 👍

app/services/patient_import/spreadsheet_transformer.rb
harimohanraj89: Good call.

app/services/patient_import/spreadsheet_transformer.rb
harimohanraj89: That will let the bad data hit our model validations, which is already hooked up to provide feedback back to the user in a flash message.

spec/services/patient_import/spreadsheet_transformer_spec.rb
harimohanraj89: Great call. I'll add some more test data that contains all sorts of values 👍

rsanheim (approved):

PR #2566, opened 2021-06-03
Automatic COVID reminder fixes

prabhanshuguptagit:
**Story card:** [chXXXX](URL)

## Because

We don't want to send covid medication reminders on sundays.

## This addresses

The job doesn't run if day is sunday. It also silently returns on non prod envs instead of throwing and exception.

### DIFF
--- a/lib/tasks/scripts/covid_medication_reminders.rb
+++ b/lib/tasks/scripts/covid_medication_reminders.rb
@@ -1,10 +1,12 @@
 class CovidMedicationReminders
   def self.call(number_of_patients: 10_000)
-    if CountryConfig.current[:name] != "India" || !SimpleServer.env.production?
-      raise "Cannot send Covid medication reminders in this env"
+    raise "Experiments are not enabled in this env" unless Flipper.enabled?(:experiment)
+
+    if CountryConfig.current[:name] != "India" || ENV["SIMPLE_SERVER_ENV"] != "production"
+      return "Cannot send Covid medication reminders in this env"
     end
 
-    raise "Experiments are not enabled in this env" unless Flipper.enabled?(:experiment)
+    return if Date.today.sunday?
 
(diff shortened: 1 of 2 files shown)


spec/tasks/scripts/covid_medication_reminders_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  it "does not send notifications reminders if day is sunday" do
+    enable_flag(:experiment)
+    allow(CountryConfig.current).to receive("[]").with(:name).and_return("India")
+    allow(ENV).to receive(:[]).and_call_original
+    allow(ENV).to receive(:[]).with("SIMPLE_SERVER_ENV").and_return("production")
+
+    expect(Experimentation::MedicationReminderService).not_to receive(:schedule_daily_notifications)
+
+    Timecop.freeze("6 Jun 2021") do # Not a Sunday
```
kpethtel, line 36: I think you meant "this is a sunday". It might be easier to just set variables like:
```
let(:sunday) { "6 Jun 2021" }
let(:monday) { "7 Jun 2021" } 
```

lib/tasks/scripts/covid_medication_reminders.rb
```diff
@@ -1,10 +1,12 @@
 class CovidMedicationReminders
   def self.call(number_of_patients: 10_000)
-    if CountryConfig.current[:name] != "India" || !SimpleServer.env.production?
-      raise "Cannot send Covid medication reminders in this env"
+    raise "Experiments are not enabled in this env" unless Flipper.enabled?(:experiment)
```
kpethtel, line 3: I think we should move this back to where it was (after the guard for country and environment). This will raise errors in Ethiopia, Bangladesh, and Sandbox, which seems like unnecessary noise.

kpethtel (approved):
Approved, but I think we should move feature flag error before merging.

PR #2567, opened 2021-06-03
Keep timestamps as UTC when creating materialized views

rsanheim:
This is an alternative, simpler approach to https://github.com/simpledotorg/simple-server/pull/2560.  I think there is actually only one mat view we need to change (`LatestBloodPressuresPerPatientPerMonth`), because the others are derived from that or just straight selects w/o any timezone conversion.

This will remove a bunch of unnecessary conversion we would need to do
when comparing timestamps in our app, because Rails wants everything to
stored as UTC at the data layer.  And all our other tables do this
already, so consistency!

**Story card:** [ch3660](https://app.clubhouse.io/simpledotorg/story/3660/fix-timezone-inconsistencies-in-current-reporting)

### Details

This turned out to be pretty straightforward once I got the mat view update correct _and_ got the test data creation in order.  One thing that surprised me is that we don't set an application time zone **at all** in any of the API sync code paths - which means we default to UTC. I think that is fine and correct, but we should probably make it explicit.  I also _assume_ our PG TimeZone is UTC, but again, something we should confirm.

I'm assuming the timestamps get created correctly as UTC in transactional tables via timezones coming in from the JSON payloads as iso8601 format, so Rails will be able to parse the time zone from that and then convert to UTC when doing INSERTs or UPDATEs.  Someone from Nilseno should confirm that on prod though with some real requests.

One other gotcha I was struggling with is all the configuration state going on, and realizing that if you don't have something set up correctly you may have ActiveRecord / Postgres silently converting timezones on insert...which can really make it hard to know what is going on.  I added some comments and a spec to verify all of this and also to try to emulate the different time zones at play in the app ...where we have API in UTC and reporting in ITC.

### Further Fun Timezone reading

* https://phili.pe/posts/timestamps-and-time-zones-in-postgresql/
* https://medium.com/building-the-system/how-to-store-dates-and-times-in-postgresql-269bda8d6403
* https://makandracards.com/makandra/46009-working-with-or-without-time-zones-in-rails-applications
* https://stackoverflow.com/questions/5876218/difference-between-timestamps-with-without-time-zone-in-postgresql
* https://stackoverflow.com/questions/9571392/ignoring-time-zones-altogether-in-rails-and-postgresql/9576170#9576170
* https://dba.stackexchange.com/questions/59006/what-is-a-valid-use-case-for-using-timestamp-without-time-zone


spec/models/patient_spec.rb
```diff
@@ -1,8 +1,12 @@
 require "rails_helper"
 
 describe Patient, type: :model do
+  let(:reporting_timezone) { Period::ANALYTICS_TIME_ZONE }
   def refresh_views
+    original = ActiveRecord::Base.connection.execute("SELECT current_setting('TIMEZONE')").first["current_setting"]
+    ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{Period::ANALYTICS_TIME_ZONE}'")
     LatestBloodPressuresPerPatientPerMonth.refresh
+    ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE #{original}")
```
rsanheim, line 9: reset PG's time zone back to whatever it was originally, for safety

spec/models/patient_spec.rb
```diff
@@ -199,6 +203,98 @@ def refresh_views
...
+          # For any provided date in June in the local timezone, the LTFU BP cutoff is the end of June 30 of the
+          # previous year in the local timezone.
+          #
+          # Eg. For any date provided in June 2021, the cutoff is the June 30-Jul 1 boundary of 2020
+
+          long_ago = 5.years.ago
+
+          # We explicitly set the times in the reporting TZ here, but don't use the block helper because its a hassle w/
+          # all the local vars we need
+          timezone = Time.find_zone(reporting_timezone)
```
rsanheim, line 226: We get the zone object to use it explicitly, so we don't need to use the block helper here...the default time zone is UTC, which is what we want for the actual fixture creation below.

spec/models/patient_spec.rb
```diff
@@ -199,6 +203,98 @@ def refresh_views
...
+          over_a_year_ago = timezone.local(2020, 6, 30, 23, 59, 59) # End of June 30 2020 in local timezone
+          beginning_of_month = timezone.local(2021, 6, 1, 0, 0, 0) # Beginning of June 1 2021 in local timezone
+          end_of_month = timezone.local(2021, 6, 30, 23, 59, 59) # End of June 30 2021 in local timezone
+
+          not_ltfu_patient = create(:patient, recorded_at: long_ago)
+          ltfu_patient = create(:patient, recorded_at: long_ago)
+
+          create(:blood_pressure, patient: not_ltfu_patient, recorded_at: under_a_year_ago)
+          create(:blood_pressure, patient: ltfu_patient, recorded_at: over_a_year_ago)
+          with_reporting_time_zones do # We don't actually need this, but its a nice sanity check
```
rsanheim, line 237: The spec passes without this (which it should, since the time zone conversion will 'just work'), but its nice for a sanity check.

spec/models/latest_blood_pressures_per_patient_per_month_spec.rb
```diff
@@ -5,10 +5,63 @@ def refresh_views
...
+      #   1) populating materialized views
+      #   2) retrieving data to show in reports
+      # This is to verify that times are consistent and correct and respect Rails auto conversion to TimeWithZone
+      Time.use_zone(Period::ANALYTICS_TIME_ZONE) do
+        expect(Time.zone.name).to eq("Asia/Kolkata")
+        refresh_views_with_pg_set_to_reporting_time_zone
+        mat_view_row = described_class.find_by!(patient: ltfu_patient)
+
+        raw_recorded_at = described_class.connection.select_one("SELECT patient_recorded_at FROM #{described_class.table_name}")
+        expect(raw_recorded_at["patient_recorded_at"]).to eq("2018-01-15 12:00:00+00")
```
rsanheim, line 47: Still in UTC, which is good...

spec/models/latest_blood_pressures_per_patient_per_month_spec.rb
```diff
@@ -5,10 +5,63 @@ def refresh_views
...
+        expect(mat_view_row.bp_recorded_at).to eq(bp.recorded_at)
+
+        expect(ltfu_patient.recorded_at).to eq(timestamp_in_utc)
+        expect(ltfu_patient.recorded_at).to eq(timestamp_in_ist) # Note that these are equal, even with different time zones
+        expect(mat_view_row.patient_recorded_at).to eq(timestamp_in_utc)
+      end
+
+      mat_view_row = described_class.find_by!(patient: ltfu_patient)
+      expect(mat_view_row.patient_recorded_at).to eq(ltfu_patient.recorded_at)
+      expect(mat_view_row.bp_recorded_at).to eq(bp.recorded_at)
```
rsanheim, line 61: Just more sanity checks here...

spec/models/latest_blood_pressures_per_patient_per_month_spec.rb
```diff
@@ -5,10 +5,63 @@ def refresh_views
...
+    described_class.refresh
+    ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE #{original}")
+  end
+
   describe "Associations" do
     it { should belong_to(:patient) }
   end
 
+  describe "handling time zones" do
+    it "stores mat view timestamps in UTC to stay consistent with the source application tables" do
```
rsanheim, line 20: This spec is long and verbose, but I wanted to try and verify a bunch of scenarios _and_ emulate the different time zone configurations that happen in the app between record creation (UTC via the API) and reporting (IST via our before_actions).

app/models/concerns/patient_reportable.rb
```diff
@@ -9,19 +9,21 @@ module PatientReportable
...
     end
 
-    scope :not_ltfu_as_of, ->(date) do
+    scope :not_ltfu_as_of, ->(time) do
+      ltfu_at = (time - LTFU_TIME).end_of_month
       joins("LEFT OUTER JOIN latest_blood_pressures_per_patient_per_months
              ON patients.id = latest_blood_pressures_per_patient_per_months.patient_id")
         .where("bp_recorded_at > ? AND bp_recorded_at < ? OR patients.recorded_at >= ?",
-          ltfu_time_ago(date), date, ltfu_time_ago(date))
+          ltfu_at, time, ltfu_at)
```
rsanheim, line 26: These are basically the same fixes as https://github.com/simpledotorg/simple-server/pull/2560, with the exception that we don't need to any UTC offset math.

db/views/latest_blood_pressures_per_patient_per_months_v06.sql
```diff
@@ -0,0 +1,20 @@
+SELECT DISTINCT ON (patient_id, year, month)
+    blood_pressures.id AS bp_id,
+    blood_pressures.patient_id AS patient_id,
+    patients.registration_facility_id AS registration_facility_id,
+    patients.assigned_facility_id AS assigned_facility_id,
+    patients.status as patient_status,
+    blood_pressures.facility_id AS bp_facility_id,
+    blood_pressures.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' AS bp_recorded_at,
```
rsanheim, line 8: Here is how we maintain the timestamps as UTC when grabbing them from the application tables.  

We need to be very explicit here and tell PostgreSQL we are grabbing `bp_recorded_at` UTC and storing It as UTC. if we don't do this, the PSQL TimeZone setting may break things by silently converting.  I originally tried a straight import without any time zone conversion at all, but that was causing issues and also was scary with how much magical timezone conversion could be happening.

db/views/latest_blood_pressures_per_patient_per_months_v06.sql
```diff
@@ -0,0 +1,20 @@
...
+    patients.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' AS patient_recorded_at,
+    blood_pressures.systolic AS systolic,
+    blood_pressures.diastolic AS diastolic,
+    blood_pressures.deleted_at AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' AS deleted_at,
+    medical_histories.hypertension as medical_history_hypertension,
+    cast(EXTRACT(MONTH FROM blood_pressures.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE'))) as text) AS month,
```
rsanheim, line 14: This is the same as before, where we grab the `recorded_at` as UTC, convert to the reporting time zone (assuming its set in the PG session), and then grab the month / year / quarter from it.

rsanheim (commented):
Some notes for reviewers

spec/models/patient_spec.rb
harimohanraj89: The `LOCAL` in `SET LOCAL TIME ZONE` ensures that the change is only for the current transaction. Shall we rely on that behavior of Rspec to simplify this?

harimohanraj89 (approved):
I've pushed a commit to propagate your more tactical approach to the LTFU specs (just to save some time). This change is lookin good 👍

app/models/period.rb
```diff
@@ -1,6 +1,6 @@
 class Period
   REGISTRATION_BUFFER_MONTHS = 3
-  REPORTING_TIMEZONE = CountryConfig.current[:time_zone] || "Asia/Kolkata"
+  REPORTING_TIME_ZONE = CountryConfig.current[:time_zone] || "Asia/Kolkata"
```
ssrihari, line 3: Unrelated, but maybe defaulting to Asia/Kolkata isn't valid anymore?

db/views/latest_blood_pressures_per_patient_per_months_v06.sql
ssrihari: This shouldn't behave differently for app tables and reporting tables no?

>  PSQL TimeZone 

Are you referring to the psql client timezone?

spec/models/latest_blood_pressures_per_patient_per_month_spec.rb
```diff
@@ -5,45 +5,60 @@ def refresh_views
     described_class.refresh
   end
 
+  def refresh_views_with_pg_set_to_reporting_time_zone
+    original = ActiveRecord::Base.connection.execute("SELECT current_setting('TIMEZONE')").first["current_setting"]
+    ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{Period::REPORTING_TIME_ZONE}'")
+    described_class.refresh
```
ssrihari, line 11: Ideally this is baked into the `refresh` method. @harimohanraj89 mentioned that when he tried to do that it broke a bunch of specs. Since you might have additional context around this, would you like to take a shot at that @rsanheim? I'm okay with us going ahead with this though.

app/models/concerns/patient_reportable.rb
```diff
@@ -9,19 +9,21 @@ module PatientReportable
     scope :with_hypertension, -> { joins(:medical_history).merge(MedicalHistory.hypertension_yes).distinct }
     scope :excluding_dead, -> { where.not(status: :dead) }
 
-    scope :ltfu_as_of, ->(date) do
+    scope :ltfu_as_of, ->(time) do
+      ltfu_at = (time - LTFU_TIME).end_of_month
```
prabhanshuguptagit, line 13: `ltfu_at` and `ltfu_as_of` are very similar. How about
```suggestion
     last_bp_after = (time - LTFU_TIME).end_of_month
```

prabhanshuguptagit (approved):
✨ Approving to unblock

spec/models/patient_spec.rb
rsanheim: Are you suggesting we make this a global RSpec hook?  I'd rather not make this change across our entire test suite, as I don't _think_ it would match production.

db/views/latest_blood_pressures_per_patient_per_months_v06.sql
rsanheim: > This shouldn't behave differently for app tables and reporting tables no?

In the prod app today, it doesn't behave differently between app and reporting tables, because:

* application rows are written with the PSQL session time zone set to UTC and the Rails app time zone set to UTC (ie the default)
* reporting rows are written with the PSQL session time zone set to UTC (I believe we don't change this) and the app time zone set to REPORTING_TIME_ZONE (ie IST in India)

Given the differences in these two contexts, I wanted to be 100% explicit and clear when creating the materialized view.

app/models/period.rb
rsanheim: That would make sense - I think the alternative is to require this be set in the country config, else we raise an error and blow up on application boot.

spec/models/latest_blood_pressures_per_patient_per_month_spec.rb
rsanheim: I spent more time reading through various Postgres and Rails timestamp / timezone docs last week, and it still isn't clear to me when ActiveRecord will convert timezones on insert versus when it won't...so any large scale change to our `refresh` method would definitely require some production verification to make sure we are staying consistent and correct.

app/models/concerns/patient_reportable.rb
rsanheim: ya I struggled w name for this...I'll go w/ your suggestion.

spec/models/latest_blood_pressures_per_patient_per_month_spec.rb
ssrihari: I don't think ActiveRecord itself sets a time zone in postgresql, does it? It would change how it interprets timestamps going in and coming up, but in this case with running a `refresh` there isn't any argument or result set coming out. I think we should be safe.

PR #2569, opened 2021-06-04
Fix cache warming timezone

harimohanraj89:
**Story card:** https://app.clubhouse.io/simpledotorg/story/3660/fix-timezone-inconsistencies-in-current-reporting

Extraction of cache-warming fixes only from #2560 

We used to warm the cache in Sidekiq, and we explicitly [set the application's timezone](https://github.com/simpledotorg/simple-server/blob/master/config/initializers/sidekiq.rb#L1-L5) around all Sidekiq jobs, similar to how we do [for web requests](https://github.com/simpledotorg/simple-server/blob/master/app/controllers/reports/regions_controller.rb#L219-L226).

When we un-backgrounded the warming job recently (#2423), we inadvertently started warming our caches in UTC, which would place patient activity in the wrong day/time if it occurred between midnight and 530am IST. It's unlikely to happen in reality, but an inconsistency we should avoid. It is definitely prevalent in our Sandbox data, and makes it hard for us to compare new reporting work against existing dashboard data.

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -103,7 +103,7 @@ class Admin::UsersController < AdminController
   end
 
   def set_time_zone
-    Time.use_zone(Period::ANALYTICS_TIME_ZONE) { yield }
+    Time.use_zone(Period::REPORTING_TIMEZONE) { yield }
   end
 
   def user_params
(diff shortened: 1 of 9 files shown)


rsanheim (approved):
Looks great. 👍

PR #2571, opened 2021-06-06
3705: stale patient query should return distinct patients

kpethtel:
**Story card:** [ch3705](https://app.clubhouse.io/simpledotorg/story/3705/make-query-for-stale-patients-distinct)

## Because

We made a mistake in the original query. It should return distinct patient ids.


rsanheim (approved):
looks good except for the listing

PR #2584, opened 2021-06-07
Update rspec-rails / rspec

rsanheim:
(no-story, just keeping dependencies up to date)

Update `rspec-rails`, which updates downstream dependencies as well (`rspec-core` etc).

I've also moved it out of the main gem bundle, as we shouldn't ever have `rspec` or other test dependencies in our main application bundle.


kpethtel (approved):
👍

PR #2593, opened 2021-06-10
3840 Reporting schema test coverage

ssrihari:
**Story card:** [ch3840](https://app.clubhouse.io/simpledotorg/story/3840/add-unit-test-coverage-for-the-database-views)

This PR adds comprehensive unit test coverage to the generation of the reporting matviews themselves.

The general strategy of these specs is:
* Pick a matview field (eg. `htn_care_state`)
* Set up test data
* Refresh the view
* Assert stuff

Of note, the [timeline spec](https://github.com/simpledotorg/simple-server/pull/2593/files#diff-2261936879c265d5b13ee66967e40ea802e83118ee097a415c8e84e5df1d66aaR295) in the patient states spec is a nice "integration" test that ensures that the matviews massage our data into the expected form.

Miscellaneous things
* Un-materializes the `reporting_facilities` dimension table, consistent with the `reporting_months` table
* Fixes a timezone issue for encounter timestamps


db/schema.rb
```diff
@@ -690,6 +690,22 @@
...
+     FROM ((patients
+       JOIN facilities ON ((patients.registration_facility_id = facilities.id)))
+       JOIN medical_histories ON (((patients.id = medical_histories.patient_id) AND (medical_histories.hypertension = 'yes'::text))))
+    WHERE (patients.deleted_at IS NULL)
+    GROUP BY (date_part('doy'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('utc'::text, patients.recorded_at))))::text, (date_part('month'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('utc'::text, patients.recorded_at))))::text, (date_part('quarter'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('utc'::text, patients.recorded_at))))::text, (date_part('year'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('utc'::text, patients.recorded_at))))::text, patients.registration_facility_id, facilities.deleted_at;
+  SQL
+  add_index "patient_registrations_per_day_per_facilities", ["facility_id", "day", "year"], name: "index_patient_registrations_per_day_per_facilities", unique: true
```
harimohanraj89, line 707: Ordering issue.

db/schema.rb
```diff
@@ -1126,13 +1188,13 @@
       p.month,
       p.quarter,
       p.year,
-      timezone('UTC'::text, timezone('UTC'::text, GREATEST((e.encountered_on)::timestamp without time zone, pd.device_created_at, app.device_created_at))) AS visited_at,
+      timezone('UTC'::text, timezone('UTC'::text, GREATEST(e.encountered_on, pd.device_created_at, app.device_created_at))) AS visited_at,
```
harimohanraj89, line 1191: This is the encounter timestamp fix.

db/schema.rb
```diff
@@ -1070,7 +1058,81 @@
...
+      latest_blood_pressures_per_patient_per_months.month,
+      latest_blood_pressures_per_patient_per_months.quarter,
+      latest_blood_pressures_per_patient_per_months.year
+     FROM latest_blood_pressures_per_patient_per_months
+    ORDER BY latest_blood_pressures_per_patient_per_months.patient_id, latest_blood_pressures_per_patient_per_months.bp_recorded_at DESC, latest_blood_pressures_per_patient_per_months.bp_id;
+  SQL
+  add_index "latest_blood_pressures_per_patients", ["bp_id"], name: "index_latest_blood_pressures_per_patients", unique: true
+  add_index "latest_blood_pressures_per_patients", ["patient_id"], name: "index_latest_bp_per_patient_patient_id"
+
+  create_view "reporting_facilities", sql_definition: <<-SQL
```
harimohanraj89, line 1135: This is no longer materialized.

db/migrate/20210614090147_do_not_materialize_reporting_facilities.rb
```diff
@@ -0,0 +1,30 @@
...
+    drop_view :reporting_patient_visits_per_month, materialized: true
+    drop_view :reporting_patient_blood_pressures_per_month, materialized: true
+    drop_view :reporting_facilities, materialized: true
+
+    execute "SET LOCAL TIME ZONE '#{Rails.application.config.country[:time_zone]}'"
+
+    create_view :reporting_facilities, materialized: false
+    create_view :reporting_patient_blood_pressures_per_month, materialized: true
+    create_view :reporting_patient_visits_per_month, materialized: true
+    update_view :reporting_patient_visits_per_month, version: 2, revert_to_version: 1, materialized: true
```
harimohanraj89, line 13: This is shoehorned in here, but prevents from having to drop and recreate the patient states table another time.

spec/models/reporting/patient_states_per_month_spec.rb
```diff
@@ -0,0 +1,354 @@
...
+        # 5  months ago    uncontrolled bp taken
+        patient = create(:patient, recorded_at: two_years_ago)
+        create(:blood_pressure, :with_encounter, patient: patient, recorded_at: ten_months_ago, systolic: 139, diastolic: 89)
+        create(:prescription_drug, patient: patient, device_created_at: eight_months_ago)
+        create(:blood_pressure, :with_encounter, patient: patient, recorded_at: five_months_ago, systolic: 140, diastolic: 90)
+
+        described_class.refresh
+
+        with_reporting_time_zones do
+          expect(patient_states(patient).pluck(:months_since_registration)).to eq((0..24).to_a)
```
timcheadle, line 326: Clever 💯

spec/models/reporting/patient_visits_per_month_spec.rb
```diff
@@ -0,0 +1,81 @@
+require "rails_helper"
+
+RSpec.describe Reporting::PatientVisitsPerMonth, {type: :model, reporting_spec: true} do
+  describe "Associations" do
+    it { should belong_to(:patient) }
+  end
+
+  describe "the visit definition" do
+    it "considers a BP measurement as a visit" do
+      bp = create(:blood_pressure, :with_encounter, recorded_at: test_times[:now])
```
timcheadle, line 10: I'd love to eventually make `:with_encounter` just the default

spec/models/reporting/patient_visits_per_month_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        expect(visit.visited_at).to eq blood_sugar.recorded_at
+      end
+    end
+
+    it "considers a Prescription Drug creation as a visit" do
+      prescription_drug = create(:prescription_drug, device_created_at: test_times[:now])
+      described_class.refresh
+      with_reporting_time_zones do
+        visit = described_class.find_by(patient_id: prescription_drug.patient_id, month_date: test_times[:now])
+        expect(visit.visited_at).to eq prescription_drug.device_created_at
```
timcheadle, line 34: Wondering if we should alias `recorded_at` for these models. Otherwise we have to remember to use date A for this model, and date B for this model.

timcheadle (commented):
Solid work!

What would've helped a lot is to have a data dictionary describing the indicators and their possible values

spec/reporting_helpers.rb
```diff
@@ -0,0 +1,40 @@
...
+      long_ago: now - 5.years,
+      under_a_year_ago: timezone.local(2020, 7, 1, 0, 0, 1), # Beginning of July 1 2020
+      over_a_year_ago: timezone.local(2020, 6, 30, 23, 59, 59), # End of June 30 2020
+      month_string: "2021-06",
+      beginning_of_month: now, # Beginning of June 1 2021
+      over_three_months_ago: timezone.local(2021, 3, 31, 0, 0, 0), # End of March 2021
+      under_three_months_ago: timezone.local(2021, 4, 1, 0, 0, 0), # Beginning of April 2021
+      end_of_month: timezone.local(2021, 6, 30, 23, 59, 59) # End of June 30 2021
+    }
+    # Reference chart:
```
rsanheim, line 26: Not sure I understand this chart...what do the integer indices refer to?

spec/reporting_helpers.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+  end
+
+  def test_times
+    # We explicitly set the times in the reporting TZ here, but don't use the block helper because its a hassle w/
+    # all the local vars we need
+    timezone = Time.find_zone(Period::REPORTING_TIME_ZONE)
+    now = timezone.local(2021, 6, 1, 0, 0, 0)
+    {
+      now: now,
```
rsanheim, line 16: If we are using June 1 as the basis for these times, maybe change the helper to indicate that...otherwise its impossible to know what these times are when you are looking at a test in isolate. Maybe name it `june_2021`, so it would read like `june_2021[:now]` or `june_2021[:under_a_year_ago]`...

spec/factories/encounters.rb
```diff
@@ -56,7 +56,7 @@ def updated_encounters_payload(existing_encounter)
 end
 
 def associate_encounter(observable)
-  encountered_on = Encounter.generate_encountered_on(observable.recorded_at, 0)
+  encountered_on = Encounter.generate_encountered_on(observable.recorded_at, Time.find_zone(Period::REPORTING_TIME_ZONE).utc_offset)
```
rsanheim, line 59: 👍

rsanheim: this ensures encounters are created as UTC, correct? Maybe the `generate_encountered_on` method should be responsible for that...feels like this method could just pass in the `recorded_at` ?

app/models/reporting/patient_states_per_month.rb
```diff
@@ -0,0 +1,6 @@
+module Reporting
```
rsanheim, line 1: Its unfortunate we have another module named `Reporting` now when we already have `Reports`.  Could we just call this `Reports` to be consistent?  I think `set_table_name` will do the right thing for us...we can then use whatever module we want.

app/models/reporting/matview.rb
```diff
@@ -0,0 +1,10 @@
+module Reporting
+  class Matview < ActiveRecord::Base
+    def self.refresh
+      ActiveRecord::Base.transaction do
+        ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{Period::REPORTING_TIME_ZONE}'")
```
rsanheim, line 5: Do we have a helper for this? If not its probably time to create one - `with_reporting_time`

rsanheim (commented):
Some feedback, but overall these tests look good!

db/views/reporting_patient_visits_per_month_v02.sql
```diff
@@ -0,0 +1,64 @@
...
+FROM
+(
+    SELECT * FROM patients p
+    LEFT OUTER JOIN reporting_months cal
+    ON to_char(p.recorded_at AT TIME ZONE 'utc' AT TIME ZONE (SELECT current_setting('TIMEZONE')), 'YYYY-MM') <= cal.month_string
+) p
+-- Only fetch Encounters that happened on or before the selected calendar month
+-- We use year and month comparisons to avoid timezone errors
+LEFT JOIN LATERAL (
+    SELECT encountered_on::timestamp AT TIME ZONE (SELECT current_setting('TIMEZONE')) AT TIME ZONE 'UTC' as encountered_on, facility_id
```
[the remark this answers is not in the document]
ssrihari: >  Is this just to get the "timezone units" the same?

exactly! this is what the conversion looks like:
```sql
simple-server_development=# select '2020-01-01'::date::timestamp;
      timestamp
---------------------
 2020-01-01 00:00:00
(1 row)
```

spec/models/reporting/patient_states_per_month_spec.rb
```diff
@@ -0,0 +1,354 @@
...
+  def under_care_patient_ids(month_date: Date.current.beginning_of_month)
+    described_class
+      .where(htn_care_state: "under_care", month_date: month_date)
+      .pluck(:id)
+  end
+
+  context "indicators" do
+    describe "htn_care_state" do
+      it "marks a dead patient dead" do
+        dead_patient = create(:patient, status: Patient.statuses[:dead])
```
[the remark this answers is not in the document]
ssrihari: I could also just do `create(:patient, status: :dead)` to be less clunky. A trait is going far for this I feel. wdyt?

spec/models/reporting/patient_states_per_month_spec.rb
```diff
@@ -0,0 +1,354 @@
...
+        described_class.refresh
+        with_reporting_time_zones do
+          expect(described_class.where(htn_care_state: "dead").pluck(:id)).to include(dead_patient.id)
+        end
+      end
+
+      it "marks a patient registered more than 12 months ago with BP more than 12 months ago as ltfu" do
+        patient_registered_13m_ago = Timecop.freeze(13.months.ago) { create(:patient) }
```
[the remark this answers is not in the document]
ssrihari: This scopes the variables to the `freeze`d block, and we use a different scope when they're used as well.

spec/models/reporting/patient_states_per_month_spec.rb
```diff
@@ -0,0 +1,354 @@
...
+      .pluck(:id)
+  end
+
+  context "indicators" do
+    describe "htn_care_state" do
+      it "marks a dead patient dead" do
+        dead_patient = create(:patient, status: Patient.statuses[:dead])
+        described_class.refresh
+        with_reporting_time_zones do
+          expect(described_class.where(htn_care_state: "dead").pluck(:id)).to include(dead_patient.id)
```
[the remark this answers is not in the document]
ssrihari: Yup. When a patient is dead, they're marked dead in all months.

spec/models/reporting/patient_states_per_month_spec.rb
```diff
@@ -0,0 +1,354 @@
...
+
+        described_class.refresh
+        with_reporting_time_zones do
+          expect(ltfu_patient_ids).not_to include(patient_with_recent_bp.id)
+          expect(under_care_patient_ids).to include(patient_with_recent_bp.id)
+        end
+      end
+
+      context "ltfu tests ported from patient_spec.rb" do
+        it "bp cutoffs for a year ago" do
```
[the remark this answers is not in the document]
ssrihari: 👍  fixed in 008239d7e57efef896653c9b8513ef3ef47dcf46.

spec/models/reporting/patient_states_per_month_spec.rb
```diff
@@ -0,0 +1,354 @@
...
+      it "computes it correctly" do
+        patient_1 = create(:patient, recorded_at: test_times[:under_a_year_ago])
+        patient_2 = create(:patient, recorded_at: test_times[:over_a_year_ago])
+        patient_3 = create(:patient, recorded_at: test_times[:now])
+        patient_4 = create(:patient, recorded_at: test_times[:over_three_months_ago])
+
+        described_class.refresh
+        with_reporting_time_zones do
+          expect(described_class.find_by(id: patient_1.id, month_string: test_times[:month_string]).months_since_registration).to eq 11
+          expect(described_class.find_by(id: patient_2.id, month_string: test_times[:month_string]).months_since_registration).to eq 12
```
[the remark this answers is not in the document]
ssrihari: I see what you mean. `over_a_year_ago` could mean any number more than 12. I wonder if `just_over_a_year_ago` would be a better way to represent this. The intention is to test around the boundaries for potential timezone issues, so I'd still like to keep the example time under test. I'll consider renaming it though.

spec/models/reporting/patient_visits_per_month_spec.rb
ssrihari: The model itself doesn't expose any other `recorded_at` time, other than `visited_at`. But that's a good idea, to alias the fields for consistency!

ssrihari: Fixed in 46d1a40d.

spec/models/reporting/patient_states_per_month_spec.rb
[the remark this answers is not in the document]
harimohanraj89: I think ☝️ is a good middleground

spec/models/reporting/patient_states_per_month_spec.rb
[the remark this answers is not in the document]
harimohanraj89: We could consider `over_twelve_months_ago`.

spec/models/reporting/patient_states_per_month_spec.rb
```diff
@@ -0,0 +1,354 @@
...
+    it { should belong_to(:patient) }
+  end
+
+  def ltfu_patient_ids(month_date: Date.current.beginning_of_month)
+    described_class
+      .where(htn_care_state: "lost_to_follow_up", month_date: month_date)
+      .pluck(:id)
+  end
+
+  def under_care_patient_ids(month_date: Date.current.beginning_of_month)
```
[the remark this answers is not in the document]
ssrihari: Thanks Rob. I Aaree with the helpers obscuring it, I've inlined it with 9067d25c. I want to make the enum change as well because we'll likely use the values in calling locations, will do that in the following PR.

I'm wary of adding scopes at this time, however. I would like to keep the reporting query interface minimal, and not provide too many options to start with.

spec/reporting_helpers.rb
ssrihari: Ha. This was something we used to calculate relative months while writing the test, didn't mean to check it in. Removed it with 6530c43750ed9280ce265044cb5a9c4175451187.

spec/factories/encounters.rb
ssrihari: The `encountered_on` date is actually converted in local (reporting) time, as understood by the rails app. This change is just doing what we do in the controller path in the test path as well.

app/models/reporting/patient_states_per_month.rb
ssrihari: I was thinking about the github/scientist branching when I created a different namespace. I thought we could eventually just split across 2 different namespaces for the comparison. Also, "reporting" followed naturally from the table name prefixes.

There's also the existing matviews, and I didn't want to conflate them with the new tables. How would you feel about merging these two namespaces after the github/scientist experiment? We'll have a good sense of the final picture by then.

app/models/reporting/matview.rb
ssrihari: I'm actually hoping to refactor refreshes so that this is the _only_ place we'll need to use this. Can you think of another place we need to set the DB local time in the same way?

spec/models/reporting/patient_states_per_month_spec.rb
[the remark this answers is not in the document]
ssrihari: Fixed in e098c903.

spec/reporting_helpers.rb
ssrihari: Thanks for the naming suggestion! Fixed in e098c903.

spec/models/reporting/patient_states_per_month_spec.rb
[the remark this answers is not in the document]
ssrihari: Fixed in e098c9035ccd5c79528b738976693cd3fb3834f0.

spec/models/reporting/patient_visits_per_month_spec.rb
ssrihari: I think we could do this in a different PR. Added a [story](https://app.clubhouse.io/simpledotorg/story/3899/change-bp-and-bs-factories-to-create-encounters-by-default) for it.

spec/models/reporting/patient_states_per_month_spec.rb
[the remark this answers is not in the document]
harimohanraj89: I also think the specs are not the right drivers for this decision. Let's start using this model for actual reports first. Once we get a feel for what that looks like, let's let that drive the enum decision for us.

spec/factories/encounters.rb
ssrihari: In the old code, the provide timestamp `observable.recorded_at` is interpreted as observed in UTC, which results in us interpreting it as the wrong day in the reporting tables.

app/models/reporting/patient_states_per_month.rb
ssrihari: We'll rename the namespace to `ReportingPipeline` to make the intention very clear. Also created a [story](https://app.clubhouse.io/simpledotorg/story/3900/consolidate-reporting-model-namespaces) to merge the namespaces.

app/models/reporting/matview.rb
ssrihari: We use this while [refreshing the other matviews](https://github.com/simpledotorg/simple-server/blob/master/app/services/refresh_materialized_views.rb#L37) right now. Once we merge the two reporting namespaces, we'll be able to keep the refresh in literally one place. 

This is a pretty weird/obscure operation used in a specific context, and we don't want to propagate the use of it by adding a helper for it.

timcheadle (approved):

PR #2594, opened 2021-06-10
Remove rspec from production bundle, take two

rsanheim:
Second attempt from https://github.com/simpledotorg/simple-server/pull/2584, which broke deploys.  @ssrihari added back rspec in https://github.com/simpledotorg/simple-server/pull/2589 to unblock deploys.

Going to try again here while also keeping deploys working 😏 .

### Details

* move rspec-rails and rswag to dev/test bundle
* updated rspec-rails
* updated data-migrate
* update dogstatsd-ruby to 4.8.3 and lock to version 4 for right now. Looks like 4.8.3 could bring in some perf improvements to metrics submission 🐎. Version 5 has API breaking changes so that should def be its own PR. See the [changelog](https://github.com/DataDog/dogstatsd-ruby/blob/master/CHANGELOG.md) for details.


rsanheim:
Stale.

PR #2596, opened 2021-06-14
Lookup regions by slug in patient lists, not region-sources

ssrihari:
**Story card:** [ch3560](https://app.clubhouse.io/simpledotorg/story/3560/missing-linelists)


app/controllers/reports/patient_lists_controller.rb
```diff
@@ -15,11 +15,11 @@ def show
     recipient_email = current_admin.email
     download_params = case region_class
     when "facility_group"
-      {id: @region.id}
+      {id: @region.source_id}
```
harimohanraj89, line 18: Would this be a good time to stick to `@region.id`, and update the behavior of the `PatientListDownloadJob`?

harimohanraj89 (approved):
One suggestion for a refactor while we're here. Not a blocker 🚀

PR #2597, opened 2021-06-14
Feature flag drug stock for districts instead of users

prabhanshuguptagit:
**Story card:** [ch3321](https://app.clubhouse.io/simpledotorg/story/3321)

## Because

We want to enable drug stocks for a district at a time, instead of enabling on a per user basis.

## This addresses

Adds `Region` as a flipper actor and feature flags the drug stocks pages based on it.


harimohanraj89 (changes requested):
Looks great, can we add the relevant test coverage to this PR?

prabhanshuguptagit:
Did some refactoring, ready for re-review @harimohanraj89.

spec/models/user_spec.rb
```diff
@@ -349,4 +349,31 @@
...
+    it "has feature_enabled?" do
+      user = create(:user)
+      Flipper.enable(:a_flag, user)
+
+      expect(user.feature_enabled?(:a_flag)).to be true
+    end
+  end
+
+  describe "drug_stocks_enabled?" do
+    it "is true if any district accessible by the user is enabled" do
```
harimohanraj89, line 369: I'd love to one more spec that flips on a facility _outside_ the admin's access, and ensures the result is false.

spec/models/region_spec.rb
```diff
@@ -432,4 +432,19 @@
...
   end
+
+  describe "flipperable" do
+    it "has a flipper_id" do
+      region = create(:region, reparent_to: Region.root)
+
+      expect(region.flipper_id).to eq("Region;#{region.id}")
+    end
+
+    it "has feature_enabled?" do
```
harimohanraj89, line 443: Could we add more sad path coverage?

```ruby
describe "#feature_enabled?" do
  it "returns true when feature is enabled for region"
  it "returns false when feature is disabled for region"
end
```

harimohanraj89 (approved):
So sad path. :shipit:

PR #2599, opened 2021-06-14
Repository renames for clarity and concision

rsanheim:
This renames a bunch of the methods in Repository, removing `_counts` suffixes, and trying to get more consistent with some of the regularly used names across the app.

I also made one small change to grab LTFU from Repo in RegionService, instead of grabbing it directly from Patient.

[ch3668](https://app.clubhouse.io/simpledotorg/story/3668/grab-all-ltfu-data-from-repository-instead-of-regionservic)


app/controllers/reports/regions_controller.rb
```diff
@@ -39,15 +39,15 @@ def show
       slug = region.slug
       {
         region: region,
-        adjusted_patient_counts: repo.adjusted_patient_counts[slug],
-        controlled_patients: repo.controlled_patients_count[slug],
-        controlled_patients_rate: repo.controlled_patients_rate[slug],
-        uncontrolled_patients: repo.uncontrolled_patients_count[slug],
-        uncontrolled_patients_rate: repo.uncontrolled_patients_rate[slug],
+        adjusted_patient_counts: repo.adjusted_patients[slug],
```
kpethtel, line 42: Given that we've mostly dropped `_patients_count` like how we changed `controlled_patient_counts` to `controlled`, is there any value in keeping `_patients` on `adjusted_patients` and `cumulative_assigned_patients`?

app/controllers/reports/regions_controller.rb
```diff
@@ -39,15 +39,15 @@ def show
...
+        controlled_patients: repo.controlled[slug],
+        controlled_patients_rate: repo.controlled_rates[slug],
+        uncontrolled_patients: repo.uncontrolled[slug],
+        uncontrolled_patients_rate: repo.uncontrolled_rates[slug],
         missed_visits: repo.missed_visits[slug],
         missed_visits_rate: repo.missed_visits_rate[slug],
-        registrations: repo.registration_counts[slug],
-        cumulative_patients: repo.cumulative_assigned_patients_count[slug],
+        registrations: repo.monthly_registrations[slug],
```
kpethtel, line 49: These changes to `monthly_` are definitely an improvement. 👍

kpethtel (approved):
Looks good. I think we could consider dropping the `_patients` for consistency, but it's not something I feel strongly about.

app/controllers/reports/regions_controller.rb
rsanheim: Yeah, I can go back and forth on this.  I think the repository does (currently) return things other than just patient related data (like BP measure counts), so it may be a bit early to drop the `patients` suffix.  I do like that idea if we can split up the repository into a more modular interface.

PR #2602, opened 2021-06-15
Remove analytics from regions details

rsanheim:
**Story card:** [ch3447](https://app.clubhouse.io/simpledotorg/story/3447/retire-the-legacy-analyticsquery-objects)

Removing the call to `dashboard_analytics` from regions#details now that we don't need any of the data anymore, as it is all coming from the repository.

### DIFF
--- a/app/helpers/reports/regions_helper.rb
+++ b/app/helpers/reports/regions_helper.rb
@@ -21,6 +21,13 @@ module Reports::RegionsHelper
       .flatten.compact.sum
   end
 
+  def sum_bp_measures(repository, *keys)
+    slug, user_id = keys
+    repository.bp_measures_by_user.dig(slug)
+      .map { |period, user_counts| user_counts.dig(user_id) }
+      .flatten.compact.sum
+  end
+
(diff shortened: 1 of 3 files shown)


app/helpers/reports/regions_helper.rb
```diff
@@ -21,6 +21,13 @@ def sum_registration_counts(repository, *keys)
       .flatten.compact.sum
   end
 
+  def sum_bp_measures(repository, *keys)
```
kpethtel, line 24: I would probably just name slug and user_id as params and skip the splat/decompose. Looks like you followed the pattern of the above method, but I don't know why the above method is doing it either.

kpethtel (approved):
One comment, but LGTM.

app/helpers/reports/regions_helper.rb
rsanheim: I believe the idea was there are other keys we may split on.

I think I'll refactor this into the repository, so not too worried about these helpers being a little bit weird for right now.

PR #2603, opened 2021-06-16
Show current deploy ref in HTML header

rsanheim:
Helpful to be able to tell what is deployed on prod.  This will include a link to the commit on GitHub in the HTML header in comments. Only for power users

image

(no story)


kpethtel (approved):
This is nice!

timcheadle (approved):

PR #2604, opened 2021-06-16
Grab LTFU data from Repo; not RegionService

rsanheim:
Removes another LTFU call from RegionService to Repository, and also extracts the period info hash out of Result...so that we can get rid of that object soon.

**Story card:** [ch3668](https://app.clubhouse.io/simpledotorg/story/3668/grab-all-ltfu-data-from-repository-instead-of-regionservic)


app/models/reports/repository.rb
```diff
@@ -147,6 +147,13 @@ def initialize(regions, periods:)
       end
     end
 
+    memoize def ltfu_rates
+      region_period_cached_query(__method__) do |entry|
+        slug, period = entry.slug, entry.period
+        percentage(ltfu[slug][period], denominator(entry.region, period, with_ltfu: true))
```
rsanheim, line 153: This is not right, need to use the correct denominator here...

app/models/period.rb
```diff
@@ -92,6 +92,10 @@ def bp_control_range_end_date
     bp_control_range.end.to_s(:day_mon_year)
   end
 
+  def ltfu_since_date
+    public_send(:begin).advance(months: -12).end_of_month.to_s(:day_mon_year)
```
rsanheim, line 96: have to use `send` because `begin` is a keyword :/.

app/controllers/reports/regions_controller.rb
```diff
@@ -63,10 +67,17 @@ def details
       [@region, @region.facility_regions].flatten
     end
     @repository = Reports::Repository.new(regions, periods: @period_range)
+    chart_repo = Reports::Repository.new(@region, periods: chart_range)
```
prabhanshuguptagit, line 70: We could just use `@repository` too I think. It includes the `@region`'s report and `@period_range` is a superset of `chart_range`. Will make the controller easier to follow.

app/controllers/reports/regions_controller.rb
```diff
@@ -63,10 +67,17 @@ def details
...
 
+    ltfu_data = {
+      cumulative_assigned_patients: chart_repo.cumulative_assigned_patients[@region.slug],
+      ltfu_patients: chart_repo.ltfu[@region.slug],
+      ltfu_patients_rate: chart_repo.ltfu_rates[@region.slug],
+      period_info: chart_range.each_with_object({}) { |period, hsh| hsh[period] = period.to_hash }
+    }
```
prabhanshuguptagit, line 77: Could pull this into a method. `ltfu_chart_data`?

app/models/period.rb
```diff
@@ -92,6 +92,10 @@ def bp_control_range_end_date
     bp_control_range.end.to_s(:day_mon_year)
   end
 
+  def ltfu_since_date
```
prabhanshuguptagit, line 95: 💯

app/models/period.rb
```diff
@@ -181,6 +185,17 @@ def to_s(format = :mon_year)
     end
   end
 
+  # Returns a Hash with various Period related dates for eash consumption by the view
+  def to_hash
```
prabhanshuguptagit, line 189: `def info`? This is returning some view specific dates, `to_hash` suggests that this is a hash representation of the `Period` object.

app/models/reports/repository.rb
prabhanshuguptagit: I would prefer adding a separate method `ltfu_denominator`, or just calling `cumulative_monthly_registrations` directly here. This attempt to make the denominator intelligent wasn't great:

https://github.com/simpledotorg/simple-server/blob/6b4f8bd8fe9e5664b64589f3a165a9d20a054344/app/models/reports/result.rb#L167-L175

prabhanshuguptagit (approved):
Looks awesome :D

app/models/period.rb
prabhanshuguptagit: I've been meaning to revisit this name, `ltfu_since_date` is a little misleading. It suggests the patient has been ltfu since 12 months but they're actually ltfu since the `period`. They haven't _had a BP since 12 months ago_. 

Any ideas?

app/models/period.rb
rsanheim: 👍

app/models/period.rb
rsanheim: Yeah I struggled on the name for this.  `info` feels a little too general...maybe `full_info` ? I dunno. We can always rename away from `to_hash` if we have a more specific representation of hash we want to add in the future.

app/models/period.rb
rsanheim: I don't really have anything better for this right now.

app/controllers/reports/regions_controller.rb
rsanheim: I tried this, and the problem is that the range of data we need for the charts is much larger than the data we need for the tables (24 months vs 6).  So if we use the single repository, we end up pulling back a bunch more data for the children regions than we need.

This added at least 500 ms to a large district in my dev testing for the cached case, so pretty bad perf hit for regions w/ lots of data.

app/controllers/reports/regions_controller.rb
rsanheim: 👍

PR #2605, opened 2021-06-17
Continuous Deployment from master

rsanheim:
This changes our semaphore config to auto promost the master branch to
production for all environments, assuming CI passes and previous
workflows have passed.

When we move to this sort of continous deployment, it doesn't really
make sense to cut tags for every release, because basically every merge
ref will become a release.

I've kept the logic around deploying tags that start with `release-` for
backwards compatiblity, for times when we want to cut releases by hand
still.

We currently have alerts coming to `#alerts` and `#bots` in Slack as things progress through workflows - do we need anything else for good visibility into this?


harimohanraj89 (approved):
Codewise, this looks solid. Questions:
* Should we axe the `release` script?
* Wanna check with the team during show-and-tell (or later)?

rsanheim:
> Codewise, this looks solid. Questions:
> 
> * Should we axe the `release` script?

I think keep around for now in case we want to use it for specific special case releases?

> * Wanna check with the team during show-and-tell (or later)?

Yup I pinged all server-devs in my standup message from yesterday, and from previous discussions I think everyone is on board with giving this a shot.

timcheadle:
LGTM. I'm in favor of keeping `bin/release` around until this stabilizes. We should update the README to reflect this auto-promotion.

PR #2606, opened 2021-06-17
Track merges where the records have the same device_updated_at

rsanheim:
To get some more metrics about [ch3843](https://app.clubhouse.io/simpledotorg/story/3843/high-database-load-around-10-am-ist-time-june-16-17th)

I suspect we are touching many thousands of records that we don't need to in our merge code paths that are commonly hit by the sync API.


app/models/concerns/mergeable.rb
```diff
@@ -82,9 +82,12 @@ def update_existing_record(existing_record, attributes)
...
     end
 
-    def return_old_record(existing_record)
+    def return_old_record(existing_record, new_record)
       logger.debug "#{self} with id #{existing_record.id} is old, keeping existing."
       increment_metric(:old)
+      if existing_record.device_updated_at == new_record.device_updated_at
+        increment_metric(:same_device_updated_at)
+      end
       existing_record.touch
```
kpethtel, line 91: Were we considering doing something about this touch? I'm guessing you just want to log for now to see if this is the cause before we revert the touch.

kpethtel (approved):

app/models/concerns/mergeable.rb
rsanheim: Yup, just tracking for now to see what things look like tomorrow.

PR #2607, opened 2021-06-17
3760 imo invite

kpethtel:
**Story card:** [ch3760](https://app.clubhouse.io/simpledotorg/story/3760/invite-bangladesh-patients-to-follow-simple)

## Because

We are rolling out Imo in Bangladesh. Imo requires that patients subscribe to our channel before we can message them. This PR allows us to request that all users follow our Imo channel.


app/jobs/imo/invite_patient.rb
```diff
@@ -0,0 +1,29 @@
...
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :default
+
+  attr_reader :patient
+
+  def perform(patient_id)
+    @patient = Patient.find(patient_id)
+    result = client.invite
+    return if result == "failure"
```
kpethtel, line 11: This should only happen due to poorly structured requests, so I don't think it should cause an infinite loop of invitations. These failures will be logged via sentry, so we should know if failures are happening frequently.

app/jobs/imo/invite_unsubscribed_patients.rb
```diff
@@ -0,0 +1,21 @@
...
+    patients.each do |patient|
+      Imo::InvitePatient.perform_at(next_messaging_time, patient.id)
+    end
+  end
+
+  def patients
+    Patient
+      .contactable
+      .left_joins(:imo_authorization)
+      .where("imo_authorizations.last_invited_at < ? AND imo_authorizations.status != ? OR imo_authorizations.id IS NULL", 6.months.ago, "subscribed")
```
kpethtel, line 19: 6 months is a tentative value we discussed in weekly sync.

app/models/imo_authorization.rb
```diff
@@ -0,0 +1,12 @@
+class ImoAuthorization < ApplicationRecord
+  belongs_to :patient
+  validates :last_invited_at, presence: true
+  validates :status, presence: true
+
+  enum status: {
+    invited: "invited",
+    no_imo_account: "no_imo_account",
+    subscribed: "subscribed",
+    unsubscribed: "unsubscribed"
```
kpethtel, line 10: Unsubscribed isn't used yet, but I'm assuming that someone can subscribe and later unsubscribe, which is the case I want to catch here.

app/services/imo_api_service.rb
```diff
@@ -0,0 +1,71 @@
+class ImoApiService
+  IMO_USERNAME = ENV["IMO_USERNAME"]
```
kpethtel, line 2: These values need to be added to the deployment repo. There are no test values.

app/jobs/imo/invite_unsubscribed_patients.rb
```diff
@@ -0,0 +1,26 @@
+class Imo::InviteUnsubscribedPatients
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :default
+
+  REINVITATION_BUFFER = 6.months.freeze
```
kpethtel, line 6: This value is tentative and was proposed in NA/EU sync.

app/jobs/imo/invite_unsubscribed_patients.rb
```diff
@@ -0,0 +1,26 @@
...
+    return unless Flipper.enabled?(:imo_messaging)
+
+    next_messaging_time = Communication.next_messaging_time
+    patients.each do |patient|
+      Imo::InvitePatient.perform_at(next_messaging_time, patient.id)
+    end
+  end
+
+  def patients
+    Patient
```
kpethtel, line 18: I may need to exclude LTFU. I've asked Daniel.

app/services/imo_api_service.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  def process_response(response)
+    return "invited" if response.status == 200
+    if response.status == 400
+      parsed = JSON.parse(response.body)
+      error_type = parsed.dig("response", "type")
+      return "no_imo_account" if error_type == "nonexistent_user"
+    end
+    "failure"
+  end
```
rsanheim, line 53: I think we could refactor to something like this if my understanding of the API and control flow is correct:

```suggestion
  def process_response(response, path)
    case response.status
    when 200 then :invited
    when 400 then 
      if JSON.parse(response.body).dig("response", "type") == "nonexistent_user"
        :no_imo_account
      else
        raise Error.new("Unknown 400 error from IMO", path: path, response: response)
      end
    else
      raise Error.new("Unknown response error from IMO", path: path, response: response)
    end
```

app/services/imo_api_service.rb
```diff
@@ -0,0 +1,71 @@
...
+      description,
+      extra: {
+        api_path: api_path,
+        exception: exception.to_s,
+        response_status: response&.status,
+        body: response&.body&.to_s
+      },
+      tags: {type: "imo-api"}
+    )
+  end
```
rsanheim, line 70: This can go away. Capture_message is really for sending non-exception type error messages to Sentry.  Since we are using exceptions here for any case that is not invite accepted or nonexistent_user, we don't need to use capture_message.

You can set the tag with the block helper around the entire `invite` method:


```ruby
Sentry.with_scope do |scope|
  scope.set_tags(type: "imo-api")
  ``` do the things
end
```

see https://github.com/getsentry/sentry-ruby for the API details

app/jobs/imo/invite_patient.rb
rsanheim: If it does fail, it will go into the Sidekiq retry queue and retry some limited amount of times btw....it would not be infinite.  

That said, I don't think we should silently return if things fail.  I don't think we should ever get a generic `failure` if we handle things properly at the lower level.

I think these are the possible outcomes, lemme know what you think:

1. we send the invite and get back a valid response (`invited` or `no_imo_account`)
2. we send the invite and some sort of error happens at the API level ... the lower level job should raise and it will never get back here.  That may mean a timeout or some real error in IMO, or it may mean its some status comign back from IMO we need to add logic for to handle

If that is true, this all becomes

```
  patient = Patient.find(patient_id)
  result = ImpApiService.new(patient).invite
  ImoAuthorization.create!(patient: patient, status: result, last_invited_at: Time.current)
```

No need for instance state or a private method.

app/jobs/imo/invite_unsubscribed_patients.rb
```diff
@@ -0,0 +1,27 @@
...
+      .contactable
+      .not_ltfu_as_of(Time.current)
+      .left_joins(:imo_authorization)
+      .where(
+        "imo_authorizations.last_invited_at < ? AND imo_authorizations.status != ? OR imo_authorizations.id IS NULL",
```
rsanheim, line 23: Can you add some parentheses around the AND clause here to make this more explicit?

app/models/imo_authorization.rb
```diff
@@ -0,0 +1,12 @@
+class ImoAuthorization < ApplicationRecord
+  belongs_to :patient
+  validates :last_invited_at, presence: true
+  validates :status, presence: true
+
+  enum status: {
+    invited: "invited",
```
rsanheim, line 7: `invited` just means it has been sent, right?

app/models/imo_authorization.rb
```diff
@@ -0,0 +1,12 @@
+class ImoAuthorization < ApplicationRecord
```
rsanheim, line 1: Note that we won't have history here of past invites if we are just using one record per patient.

Maybe that is fine, but we should be very careful we aren't spamming users with multiple invites over time due to bugs or errors in our logic. With just the one record, it will make it a bit harder to track that.

app/models/imo_authorization.rb
rsanheim: I like the idea of being able to capture this, but I would recommend not adding a state that is not set or used anywhere yet.

app/jobs/imo/invite_unsubscribed_patients.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  sidekiq_options queue: :default
+
+  REINVITATION_BUFFER = 6.months.freeze
+
+  def perform
+    return unless Flipper.enabled?(:imo_messaging)
+
+    next_messaging_time = Communication.next_messaging_time
+    patients.each do |patient|
```
rsanheim, line 12: `find_each` here, otherwise we will get all contactable patients the first time in loaded into memory. 🤯 

Also, we can `pluck(:id)`, as that is the only thing we need here...not the whole object.

app/models/imo_authorization.rb
```diff
@@ -0,0 +1,12 @@
...
+  belongs_to :patient
+  validates :last_invited_at, presence: true
+  validates :status, presence: true
+
+  enum status: {
+    invited: "invited",
+    no_imo_account: "no_imo_account",
+    subscribed: "subscribed",
```
rsanheim, line 9: How do we know when they are subscribed?

app/services/imo_api_service.rb
```diff
@@ -0,0 +1,71 @@
...
+  IMO_USERNAME = ENV["IMO_USERNAME"]
+  IMO_PASSWORD = ENV["IMO_PASSWORD"]
+  BASE_URL = "https://sgp.imo.im/api/simple/"
+
+  class ImoApiService::HTTPError < HTTP::Error
+  end
+
+  attr_reader :phone_number, :recipient_name, :locale
+
+  def initialize(phone_number:, recipient_name:, locale:)
```
rsanheim, line 11: Lets pass in `patient` here, and let the service grab what it needs from the patient.

The API request is tightly coupled with the patient, so I think its best just to hand it the patient object and let the API grab whatever things it needs for the request.

spec/jobs/imo/invite_patient_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      before { Flipper.enable(:imo_messaging) }
+
+      it "raises an error when the patient id is not found" do
+        expect {
+          described_class.perform_async("fake")
+          described_class.drain
+        }.to raise_error(ActiveRecord::RecordNotFound)
+      end
+
+      it "creates an ImoAuthorization with 'invited' status when invitation is successful" do
```
rsanheim, line 27: `invitation is sent` ...successful could imply the user responded to it.

spec/jobs/imo/invite_patient_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+          described_class.drain
+        }.to change { ImoAuthorization.count }.from(0).to(1)
+        imo_auth = ImoAuthorization.last
+        expect(imo_auth.status).to eq("no_imo_account")
+        expect(imo_auth.patient_id).to eq(patient.id)
+      end
+
+      it "does not create an ImoAuthorization when invitation fails" do
+        patient = create(:patient)
+        allow_any_instance_of(ImoApiService).to receive(:invite).and_return("failure")
```
rsanheim, line 55: This could become `receive(:invite).and_raise("error")`

rsanheim (changes requested):
Hey, added some feedback around error handling and trying to ensure we get good visibility into any errors that come out of this.  Given the IMO API sounds like it is still under active development, we wanna make sure we cover our bases well here.

Lemme know if you'd like to pair on this at all on Monday.

app/services/imo_api_service.rb
```diff
@@ -0,0 +1,71 @@
...
+    @phone_number = phone_number
+    @recipient_name = recipient_name
+    @locale = locale
+  end
+
+  def invite
+    url = BASE_URL + "send_invite"
```
rsanheim, line 18: Lets add some basic metrics from the beginning here so we can create a dashboard to monitor all this.

```
Statsd.increment("imo.invites.attempts")
```

then later on when we get the response

```
Statds.increment("imo.invites.no_imo_account")
# or
Statds.increment("imo.invites.error")
```

rsanheim (changes requested):

app/jobs/imo/invite_unsubscribed_patients.rb
kpethtel: Yes, I had them at one point and must have removed them while testing the query. Totally agree that this needs them.

app/models/imo_authorization.rb
kpethtel: Yeah, I figured this question might come up. I felt like having a record per invitation was over-engineering. I'd rather just add some logging like `sending invitation to #{patient_id}` than keep a stack of records just for debugging purposes. Maybe I'll regret that, but I would like to keep the design very slim here if possible.

app/models/imo_authorization.rb
kpethtel: When we attempt to send an imo notification, we get a 200 if the user exists, but if the user hasn't yet subscribed, the body will be:
```    
"response": {
      "status": "failed",
      "message": "User has not subscribed",
      "error_code": "not_subscribed"
}
```
I'm not exactly sure what that body will be on a successful response because I can't test that yet and it's not in the API docs, but I'm guessing it will be different. So the general idea is that if a user's `ImoAuthorization#status` is `invited`, we will try to send them a notification via imo before sms. If the response tells us the message succeeded and doesn't tell us they're unsubscribed, we update the status to `subscribed`. Which prevents us from inviting them again.

app/models/imo_authorization.rb
kpethtel: Yeah, that's probably fair. It was more of a placeholder for when I can actually use their API. I was planning to test for the case of a user who subscribed then unsubscribes. I was thinking if they've subscribed, we try messaging them, find out they're no longer subscribed, we could have a separate "unsubscribed" status that prevents us from inviting them again. Maybe. Unclear if that's behavior we even want.

app/models/imo_authorization.rb
kpethtel: Yes.

app/services/imo_api_service.rb
kpethtel: The API is only really tied to patients in the sense that we don't currently have a reason to send imo messages to non-patients. I was trying to keep this as generic as possible for that reason. I find it unlikely we will ever want to use imo with users, but this still seems like reasonable separation of concerns.

app/services/imo_api_service.rb
```diff
@@ -0,0 +1,71 @@
+class ImoApiService
+  IMO_USERNAME = ENV["IMO_USERNAME"]
+  IMO_PASSWORD = ENV["IMO_PASSWORD"]
+  BASE_URL = "https://sgp.imo.im/api/simple/"
+
+  class ImoApiService::HTTPError < HTTP::Error
```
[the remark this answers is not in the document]
kpethtel: The error handling in this class was just copied from `exotel_api_service`. Happy to change it.

app/jobs/imo/invite_patient.rb
kpethtel: Yeah, I didn't fully think through the fact that sidekiq needs the errors to cause retries. That's just an oversight on my part. I'll refactor.

app/jobs/imo/invite_unsubscribed_patients.rb
kpethtel: I should definitely pluck. The find_each comment is irrelevant if I pluck, right? Are you saying the scope would load all contactable patients into memory before the rest of the query narrows it down? Not sure I follow that part.

rsanheim: It would load everything into memory after all those scopes are applied...which would still be tens of thousands of patients on prod I'm pretty sure.  So too much to load into memory.

👍  to pluck

rsanheim (approved):
Looks good

PR #2609, opened 2021-06-18
Sync - Don't touch "same" records

harimohanraj89:
**Story card:** [ch3843](https://app.clubhouse.io/simpledotorg/story/3843/high-database-load-around-10-am-ist-time-june-16-17th)

## Because

After launching #2606 , we discovered that most repetitive sync requests were touching the records, even though the record is most likely up-to-date on the client (since the timestamp matches exactly)

## This addresses

Stop `touch`ing these "identical" records.

## Assumptions/Consequences

* We assume that the same record updated by two different clients will never have the same `device_updated_at` timestamp
* We assume that the `.touch` was for the initiating client to re-fetch the existing record. We assume that this `touch` is used because sometimes the client's process token moves forward, even though it has not yet fetched latest updates for all records. Hence it _thinks_ it's up-to-date, but it's not. Touching the record will bump the records' timestamp to _after_ the client's process token again, forcing the client to refetch the existing record the next time it syncs.

### DIFF
--- a/app/models/concerns/mergeable.rb
+++ b/app/models/concerns/mergeable.rb
@@ -18,6 +18,8 @@ module Mergeable
         :discarded
       elsif new_record.device_updated_at > existing_record.device_updated_at
         :updated
+      elsif new_record.device_updated_at == existing_record.device_updated_at
+        :identical
       else
         :old
       end
@@ -36,8 +38,10 @@ module Mergeable
         create_new_record(attributes)
       when :updated
         update_existing_record(existing_record, attributes)
+      when :identical
+        return_identical_record(existing_record)
       when :old
-        return_old_record(existing_record, new_record)
+        return_old_record(existing_record)
       end
     end
 
@@ -82,12 +86,16 @@ module Mergeable
       existing_record
     end
 
-    def return_old_record(existing_record, new_record)
+    def return_identical_record(existing_record)
+      logger.debug "#{self} with id #{existing_record.id} is identical, doing nothing."
+      increment_metric(:identical)
+      existing_record.merge_status = :identical
+      existing_record
+    end
+
+    def return_old_record(existing_record)
       logger.debug "#{self} with id #{existing_record.id} is old, keeping existing."
       increment_metric(:old)
-      if existing_record.device_updated_at == new_record.device_updated_at
-        increment_metric(:same_device_updated_at)
-      end
       existing_record.touch
       existing_record.merge_status = :old
       existing_record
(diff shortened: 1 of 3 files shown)


spec/models/concerns/mergeable_spec.rb
```diff
@@ -36,31 +36,50 @@
     expect(Patient.count).to eq 1
   end
 
-  it "returns the existing record if input record is older" do
-    existing_patient = FactoryBot.create(:patient, address: FactoryBot.create(:address))
+  it "returns the existing record touchced if input record is older" do
```
rsanheim, line 39: `touched`

app/models/concerns/mergeable.rb
```diff
@@ -82,12 +86,16 @@ def update_existing_record(existing_record, attributes)
       existing_record
     end
 
-    def return_old_record(existing_record, new_record)
+    def return_identical_record(existing_record)
+      logger.debug "#{self} with id #{existing_record.id} is identical, doing nothing."
+      increment_metric(:identical)
+      existing_record.merge_status = :identical
```
rsanheim, line 92: Do we need to account for this new merge_status [over here](https://github.com/simpledotorg/simple-server/blob/865f3b4324b8941687ee823190a1d968586da0e3/app/services/merge_patient_service.rb#L35)?  I'm not sure if these paths ever intersect, but I imagine they _could_ sometime in the future, and we would probably want to ignore `identical` in the MergePatientService.

spec/models/concerns/mergeable_spec.rb
```diff
@@ -36,31 +36,50 @@
...
+    expect(Patient.count).to eq 1
+  end
+
+  it "returns the existing record untouched if input record is equally up-to-date" do
+    existing_patient = FactoryBot.create(:patient, address: FactoryBot.create(:address))
+    updated_at = existing_patient.updated_at
+
+    Patient.merge(existing_patient.attributes)
+
+    expect(Patient.find(existing_patient.id).updated_at.to_i).to eq updated_at.to_i
```
rsanheim, line 61: 👍

rsanheim (approved):
Looks good.  I agree w/ the slack conversation that we should definitely figure out _why_ these are happening so frequently to get to the root cause(s).

app/models/concerns/mergeable.rb
harimohanraj89: I took a look, and the current decision tree is still accurate. It's only concerned with whether or not to set the registration user and facility from the request, or keep the existing info. If the record's `identical`, it's an existing record, so it should keep its existing regn info. The `else` clause does that correctly. After that, the service invokes `Patient.merge`, which now is a no-op as of this PR.

PR #2610, opened 2021-06-18
Explicitly filter dead patients from overdue list

timcheadle:
**Story card:** [ch3761](https://app.clubhouse.io/simpledotorg/story/3761/overdue-list-does-not-explicitly-filter-out-dead-patients)

## Because

The overdue list depends on appointment status being `cancelled` to remove patients from the list. It does not explicitly exclude dead patients.

## This addresses

This explicitly filters out patients with status `dead`.

## Test instructions

1. Load the overdue list
2. Go into the console and mark the first patient that appears as `patient.status = :dead`
3. Make sure they disappear from the overdue list on reload

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -53,6 +53,8 @@ class Appointment < ApplicationRecord
     where(status: "scheduled")
       .where(arel_table[:scheduled_date].lt(Date.current))
       .where(arel_table[:remind_on].eq(nil).or(arel_table[:remind_on].lteq(Date.current)))
+      .joins(:patient)
+      .where.not(patients: {status: "dead"})
   end
 
   def self.overdue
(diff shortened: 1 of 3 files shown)


app/models/appointment.rb
```diff
@@ -53,6 +53,8 @@ def self.all_overdue
     where(status: "scheduled")
       .where(arel_table[:scheduled_date].lt(Date.current))
       .where(arel_table[:remind_on].eq(nil).or(arel_table[:remind_on].lteq(Date.current)))
+      .joins(:patient)
+      .where.not(patients: {status: "dead"})
```
rsanheim, line 57: This should be the symbol  - `status: :dead`

rsanheim (approved):
small suggestion, otherwise LGTM

PR #2612, opened 2021-06-21
Add meaningful page titles for reports

rsanheim:
Adds page titles to the report pages, as well as making the env name a shorter, more clear abbreviation for better tab readability.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -3,9 +3,8 @@ module ApplicationHelper
   STANDARD_DATE_DISPLAY_FORMAT = "%d-%^b-%Y"
 
   def page_title
-    str = get_title_for_environment
-    str << " |" << content_for(:title) if content_for?(:title)
-    str
+    title = content_for?(:title) ? content_for(:title) : I18n.t("admin.dashboard_title")
+    [env_prefix, title].compact.join(" ")
   end
 
   def bootstrap_class_for_flash(flash_type)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,6 +2,12 @@ module ApplicationHelper
   DEFAULT_PROGRAM_INCEPTION_DATE = Time.new(2018, 0o1, 0o1)
   STANDARD_DATE_DISPLAY_FORMAT = "%d-%^b-%Y"
 
+  def page_title
+    str = get_title_for_environment
+    str << " |" << content_for(:title) if content_for?(:title)
+    str
+  end
+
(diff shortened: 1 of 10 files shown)


app/helpers/simple_server_env_helper.rb
```diff
@@ -1,5 +1,6 @@
 module SimpleServerEnvHelper
-  CUSTOMIZED_ENVS = %w[development qa sandbox demo production].freeze
+  CUSTOMIZED_ENVS = %w[development demo qa sandbox production].freeze
```
harimohanraj89, line 2: How far reaching is this little constant? I'm wondering if we can easily add real support for our review envs as well while we're here, or whether that's a task for later.

app/views/layouts/application.html.erb
```diff
@@ -17,7 +17,7 @@
       </script>
     <% end %>
     <title>
-      <%= get_title_for_environment %>
+      <%= page_title %>
```
harimohanraj89, line 20: Ah interesting. We just said `Simple Dashboard` everywhere, and now this method lets us define additional content per-page. Toit.

app/helpers/application_helper.rb
```diff
@@ -2,6 +2,12 @@ module ApplicationHelper
   DEFAULT_PROGRAM_INCEPTION_DATE = Time.new(2018, 0o1, 0o1)
   STANDARD_DATE_DISPLAY_FORMAT = "%d-%^b-%Y"
 
+  def page_title
+    str = get_title_for_environment
+    str << " |" << content_for(:title) if content_for?(:title)
```
harimohanraj89, line 7: I wonder if we can do `[ENV] <content> | Simple Dashboard`, or even remove `Simple Dashboard` from the titles? I'm looking at Github page titles for reference, and they don't mention "Github" in the page titles - I guess we let the icons do the talking.

harimohanraj89 (approved):
Code looks solid. Just content suggestions from me.

We can ship this piecemeal if needed.

app/helpers/application_helper.rb
rsanheim: Yup, I'll add that for when there is an explicit title provided.

kpethtel (approved):
Nicely done.

app/views/layouts/application.html.erb
rsanheim: 👍

app/helpers/simple_server_env_helper.rb
rsanheim: Yeah we can always add that later, I'm guessing review falls into the `default` bucket for now.

PR #2601, opened 2021-06-15
Sidama support

harimohanraj89:
**Story card:** [ch3770](https://app.clubhouse.io/simpledotorg/story/3770/add-sidama-translations-to-simple)

## Because

We want to add Sidama support to Simple

* Adds API locale
* Adds notifications locale
* Uses locale key: `sid-ET`

PR #2613, opened 2021-06-21
3864 add notification purpose

kpethtel:
**Story card:** [ch3864](https://app.clubhouse.io/simpledotorg/story/3864/add-purpose-to-notifications)

## Because

Communication types like, "missed_visit_sms_reminder" conflate the mode of communication with the purpose of the communication. So we're simplifying `Communication#communication_type` to `sms` and `whatsapp` and moving the purpose to the notifications model.


app/models/notification.rb
```diff
@@ -12,36 +12,41 @@ class Notification < ApplicationRecord
...
     pending: "pending",
     scheduled: "scheduled",
     sent: "sent",
     cancelled: "cancelled"
   }, _prefix: true
+  enum purpose: {
+    covid_medication_reminder: "covid_medication_reminder",
+    experimental_appointment_reminder: "experimental_appointment_reminder",
+    missed_visit_appointment_reminder: "missed_visit_appointment_reminder"
+  }, _suffix: true
```
kpethtel, line 27: TBH, it's pretty clunky as either a prefix or suffix.

db/migrate/20210620144826_add_purpose_to_notifications.rb
```diff
@@ -0,0 +1,13 @@
+class AddPurposeToNotifications < ActiveRecord::Migration[5.2]
+  def up
+    add_column :notifications, :purpose, :string
+    Notification.find_each do |notification|
```
kpethtel, line 4: All notifications currently in the wild should be from the covid medication reminders.

db/migrate/20210620144826_add_purpose_to_notifications.rb
```diff
@@ -0,0 +1,13 @@
+class AddPurposeToNotifications < ActiveRecord::Migration[5.2]
```
kpethtel, line 1: It may have been smarter to do the migration in its own PR rather than wrapping it up with a bunch of other code changes.

db/migrate/20210620144826_add_purpose_to_notifications.rb
harimohanraj89: We could be more robust by grabbing `Notification.where(subject: nil)` only. We can also set `Notification.where.not(subject: nil)` to `missed_visit_reminder` as well for thoroughness.

db/migrate/20210620144826_add_purpose_to_notifications.rb
harimohanraj89: It's simple enough, I think we can include it here. It also avoids the window of time where the column exists, but the code's not setting it.

app/jobs/appointment_notification/worker.rb
```diff
@@ -41,22 +41,25 @@ def send_message(notification, communication_type)
       NotificationService.new
     end
 
-    if communication_type == "missed_visit_whatsapp_reminder"
+    case communication_type
+    when "whatsapp"
```
harimohanraj89, line 45: Should we keep the old types around for a while (`when "whatsapp", "missed_visit_whatsapp_reminder"`) to handle any jobs that are executed before the migration switches everyone's comms type? It would be a "transitional backwards compatibility", similar to the communication types hash below - https://github.com/simpledotorg/simple-server/pull/2613/files#diff-8bef04c7770083f39d473b82194bccd100e857238d53212f8a79d72731dbdcebR12-R14

harimohanraj89 (changes requested):
This is neat! Initial feedback below. Could you also give us an overview (in the PR desc) of what this whole ecosystem looks like after this change? What are the different parameters for any given message, etc?

app/jobs/appointment_notification/worker.rb
kpethtel: The reason I kept the types in the enum is because I want the records to continue to be valid and update-able. I didn't keep them here because there shouldn't be a case where we want to send any of the notifications that currently exist. However... I've been wrong many times before, and it's probably not a bad idea to keep there here to be safe. I'll re-add and add a note in the migration ticket to remove the references.

app/jobs/appointment_notification/worker.rb
```diff
@@ -41,22 +41,25 @@ def send_message(notification, communication_type)
...
+    when "sms"
       notification_service.send_sms(
         notification.patient.latest_mobile_number,
         notification.localized_message,
         callback_url
       ).tap do |response|
         metrics.increment("sent.sms")
       end
+    else
+      raise StandardError, "#{self.class.name} is not configured to handle communication type #{communication_type}"
```
rsanheim, line 64: This looks good, but can you create a custom Exception class to raise?  Maybe `UnknownCommunicationType`.

db/migrate/20210620144826_add_purpose_to_notifications.rb
kpethtel: This is definitely the more correct thing thing to do. I didn't do it because I believe the only notifications that exist on prod are for medication reminders. I believe @prabhanshuguptagit confirmed that last week. I've implemented your suggestion here, but  before we run this migration on prod we should confirm that all notifications without subjects are medication reminders and that either 1) there are no notifications with subjects or 2) all notifications with subjects are appointment reminders. You can use the Notification#message attribute to distinguish them. I've already confirmed that there are no notifications in the db in Bangladesh.

app/models/notification.rb
```diff
@@ -5,43 +5,46 @@ class Notification < ApplicationRecord
...
+    missed_visit_reminder: "missed_visit_reminder"
+  }, _suffix: true
 
   scope :due_today, -> { where(remind_on: Date.current, status: [:pending]) }
 
   def localized_message
-    case subject
-    when Appointment
+    case purpose
+    when "missed_visit_reminder", "experimental_appointment_reminder"
```
rsanheim, line 31: All of these strings for the `purpose` can be symbols...I'll just comment on this one but the others in this PR can also get changed over.  Since these are basically enums a symbol better represents that.

app/models/notification.rb
rsanheim: Maybe just leave suffix and prefix off? These are long strings as they are, don't think they need the prefix/suffix.

db/migrate/20210620144826_add_purpose_to_notifications.rb
```diff
@@ -0,0 +1,16 @@
...
+  def up
+    add_column :notifications, :purpose, :string
+    Notification.where(subject: nil).find_each do |notification|
+      notification.update_column(:purpose, "covid_medication_reminder")
+    end
+    Notification.where.not(subject: nil).find_each do |notification|
+      notification.update_column(:purpose, "missed_visit_reminder")
+    end
```
rsanheim, line 9: (blocking) You can use [update_all](https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/update_all) to do all these updates in two SQL statements.  No reason to iterate over every record, especially in a migration where we want things to complete quickly.

rsanheim (changes requested):
I like how this clarifies things.  Left some feedback, one item blocking.

config/locales/api/am_ET.yml
```diff
@@ -12,11 +12,11 @@ am-ET:
       expired_otp: "OTPዎ ጊዜው አልፏል። አዲስ ይጠይቁ።"
       invalid_password: "የይለፍ ቃልዎ አይዛመድም። ድጋሚ ይሞክሩ?"
       account_locked: "መለያዎ ለሚቀጥሉት %{minutes} ደቂቃዎች ተቆልፏል። እባክዎ ይቆዩ እና እንደገና ይሞክሩ።"
-  sms:
+  communications:
```
kpethtel, line 15: Hopefully this change doesn't cause confusion. There are no actual communication records for OTP  requests, but SMS is pretty obviously wrong, since it contains whatsapp translations and will probably also end up with Imo translations.

app/jobs/appointment_notification/worker.rb
harimohanraj89: Yeah let's leave the backwards compatibility in here for a day, and then rip it out.

db/migrate/20210620144826_add_purpose_to_notifications.rb
harimohanraj89: I think this reply was for the previous comment - https://github.com/simpledotorg/simple-server/pull/2613#discussion_r656844355

The less we have to keep in mind while shipping this, the better. Looks like we've updated the migration to operate on covid reminders and missed visit reminders separately. Nice 💯

harimohanraj89:
Heads up @kpethtel - There's one more translation file that needs updating now (once you've merged in the latest master) - `api/sid_ET.yml`. It was added in #2601 .

The rest of my feedback has been addressed. I'm leaving my blocking review here so you don't miss the Sidama file 😄 Feel free to dismiss my review once it's fixed!

app/models/notification.rb
kpethtel: That's a good idea. Thanks. Hadn't occurred to me to just leave it off.

db/migrate/20210620144826_add_purpose_to_notifications.rb
kpethtel: 🤦 
Funny thing is I noticed your comment persisted after I made the change and thought you had left it at the top of the relevant code block so it wouldn't disappear when I changed the line in question.

app/jobs/appointment_notification/worker.rb
kpethtel: Will do.

db/migrate/20210620144826_add_purpose_to_notifications.rb
kpethtel: Does that not lock the db? I thought that was the purpose in batching.

kpethtel:
> Heads up @kpethtel - There's one more translation file that needs updating now (once you've merged in the latest master) - `api/sid_ET.yml`. It was added in #2601 .
> 
> The rest of my feedback has been addressed. I'm leaving my blocking review here so you don't miss the Sidama file 😄 Feel free to dismiss my review once it's fixed!

Thanks for the heads up. I'll take care of it.

app/models/notification.rb
kpethtel: I do prefer symbols...

app/models/facility.rb
```diff
@@ -246,6 +246,7 @@ def locale
       "default" => "am-ET",
       "dire dawa" => "am-ET",
       "oromia" => "om-ET",
+      "sidama" => "sid-ET",
```
kpethtel, line 249: Not directly related to the PR. We added a new language and I'm adding it to our locales map.

app/models/notification.rb
kpethtel: Changing to symbols didn't work as expected. I think you can use symbols for enums like `purpose[:missed_visit_reminder]` but it seems the actual values are strings.

harimohanraj89 (approved):

db/migrate/20210620144826_add_purpose_to_notifications.rb
harimohanraj89: We have ~200k notifications? I think it's preferable to lock and update at once than running 200k queries. A good middle ground would be to grab the records in batches of, say, 10,000, instead of one by one. 

Rough pseudocode: `Thing.where(condition).find_in_batches(10000) { |batch| batch.update_all }`

kpethtel: Yeah, that seems reasonable. `find_each` uses `find_in_batches(1_000)`, so the only difference is batch size. I think the criticism Rob was initially making is that I should've been updating in batches rather than individually and I misread him as, "don't use find_each". I think using find_each with update_all satisfies his original request.

PR #2614, opened 2021-06-21
Remove ControlRateService entirely; replace with calls to RegionService

rsanheim:
Removing ControlRateService, which now is one unnecessary layer that causes confusion.

RegionService is the sort of 'god object' that returns the giant Result of data, and legacy clients can use that.  This gets us closer to cleaner Repository usage.


app/services/reports/region_service.rb
```diff
@@ -10,34 +10,48 @@ def self.call(*args)
...
+      result.controlled_patients_with_ltfu_rate = repository.controlled_rates(with_ltfu: true)[slug]
+      result.cumulative_assigned_patients = repository.cumulative_assigned_patients[slug]
+      result.cumulative_registrations = repository.cumulative_registrations[slug]
+      result.earliest_registration_period = repository.earliest_patient_recorded_at_period[slug]
+      result.ltfu_patients = repository.ltfu[slug]
+      result.ltfu_patients_rate = repository.ltfu_rates[slug]
+      result.registrations = repository.monthly_registrations[slug]
+      result.uncontrolled_patients = repository.uncontrolled[slug]
+      result.uncontrolled_patients_rate = repository.uncontrolled_rates[slug]
+      result.uncontrolled_patients_with_ltfu_rate = repository.uncontrolled_rates(with_ltfu: true)[slug]
```
rsanheim, line 42: Since we've been replacing things w/ calls to Repository, this was basically all that `ControlRateService` was doing...we were just passing hashes back and forth.  So we directly populate the Result object here now with the values from the repo.

spec/services/reports/region_service_spec.rb
```diff
@@ -35,6 +41,224 @@ def refresh_views
     end
   end
 
+  it "counts registrations and cumulative registrations" do
```
rsanheim, line 44: All these specs came over from ControlRateService, and I think its worth keeping them around for now just for a bit more confidence in the legacy things here.

timcheadle (approved):
looks solid!

kpethtel (approved):

PR #2588, opened 2021-06-08
Spike Reporting schema with toggle

rsanheim:
Spiking off of https://github.com/simpledotorg/simple-server/pull/2523 to add a seam to see how the new schema looks and performs in our reports.

**WIP**

PR #2617, opened 2021-06-22
Populate BP Controlled graph from reporting_schema_v2

rsanheim:
**Story card:** [ch3892](https://app.clubhouse.io/simpledotorg/story/3892/wire-up-controlled-uncontrolled-reports-from-pipeline-via-a-toggle)

This wires up the new reporting schema via science for controlled / uncontrolled counts only.

The science will _only_ run in dev/sandbox for right now, so that we can analyze the impact on performance first.
A power_user can pass in the query string of `v2=1` in any environment to grab control numbers from the the new schema, which bypasses science entirely. This is to do quick testing and verification of the charts.

I've also added a first pass of indexes on the patient_states table...these improved performance on our typical count queries by a 10-20x...dropping many of the queries from around somewhere around a second to 10-50 ms per query.

See also:
* earlier spike https://github.com/simpledotorg/simple-server/pull/2588
* test coverage PR for reporting schema https://github.com/simpledotorg/simple-server/pull/2593


app/models/reports/repository.rb
```diff
@@ -155,17 +163,46 @@ def initialize(regions, periods:)
     end
 
     memoize def controlled
```
kpethtel, line 165: So if I understand this correctly, we only compare the two on requests for v1? I assume this is because it's the more common code path and will provide the broadest coverage. 

But if that's the case, should we be raising on mismatches? Seems like that would potentially cause errors to users who aren't using the new v2 flag. I looked at `raise_on_mismatches` in the scientist docs, and it says to only use it in your test suite. So I'm unsure if I'm understanding this correctly.

app/models/reports/repository.rb
```diff
@@ -155,17 +163,46 @@ def initialize(regions, periods:)
...
+      return controlled_v2 if reporting_schema_v2?
+      science("controlled") do |e|
+        e.use { controlled_v1 }
+        e.try { controlled_v2 }
+        e.raise_on_mismatches = true
+      end
+    end
+
+    def controlled_v1
```
kpethtel, line 174: I would guess we can make these v1/v2 methods private. If you want to keep them close to the methods that call them, you can do the inline `private def`.

app/models/reports/repository.rb
```diff
@@ -155,17 +163,46 @@ def initialize(regions, periods:)
...
+      regions.each_with_object({}).each do |region, hsh|
+        hsh[region.slug] = control_rate_query_v2.controlled_counts(region, range: active_range(region)).tap { |hsh| hsh.default = 0 }
+      end
+    end
+
     memoize def uncontrolled
+      return uncontrolled_v2 if reporting_schema_v2?
+      science("uncontrolled") do |e|
+        e.use { uncontrolled_v1 }
+        e.try { uncontrolled_v2 }
```
kpethtel, line 190: The fact that you're not raising here makes me think the other one was left by oversight.

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,7 @@
+module Reports
```
kpethtel, line 1: I'm not sure I understand this. Is it intended to be a mixin?

app/queries/control_rate_query_v2.rb
```diff
@@ -0,0 +1,27 @@
+class ControlRateQueryV2
+  def controlled(region)
+    ReportingPipeline::PatientStatesPerMonth
```
kpethtel, line 3: This reads so nicely.

app/queries/control_rate_query_v2.rb
```diff
@@ -0,0 +1,27 @@
...
+      .where(hypertension: "yes", htn_care_state: "under_care")
+      .where(patient_assigned_facility_id: region.facilities)
+      .where("months_since_registration >= ?", Reports::REGISTRATION_BUFFER_IN_MONTHS)
+      .where(htn_treatment_outcome_in_last_3_months: :controlled)
+  end
+
+  def controlled_counts(region, range: nil)
```
kpethtel, line 10: Should range default to nil? It seems like a required parameter. It looks like it would error on `range.begin` if it were nil.

kpethtel (approved):
Several questions and comments, but it looks good!

app/models/reports/repository.rb
```diff
@@ -155,17 +163,46 @@ def initialize(regions, periods:)
...
+    def controlled_v1
       region_period_cached_query(__method__) do |entry|
         control_rate_query.controlled(entry.region, entry.period).count
       end
     end
 
+    memoize def controlled_v2
+      regions.each_with_object({}).each do |region, hsh|
+        hsh[region.slug] = control_rate_query_v2.controlled_counts(region, range: active_range(region)).tap { |hsh| hsh.default = 0 }
+      end
```
ssrihari, line 183: I was intending to avoid this per-region-entry query with v2. We have the ability to query for parent regions in v2, and that is more expressive, and more performant. In other words, 1 query for a block (or district) would be more efficient than 20 (or 200) queries for each facility in it.

Do we plan to explore that approach in a future iteration?

app/queries/control_rate_query_v2.rb
```diff
@@ -0,0 +1,27 @@
...
+    controlled(region).group_by_period(:month, :month_date, range: time_range, format: Period.formatter(:month)).count
+  end
+
+  def uncontrolled(region)
+    ReportingPipeline::PatientStatesPerMonth
+      .where("assigned_#{region.region_type}_region_id" => region.id)
+      .where(hypertension: "yes", htn_care_state: "under_care")
+      .where("months_since_registration >= ?", Reports::REGISTRATION_BUFFER_IN_MONTHS)
+      .where(htn_treatment_outcome_in_last_3_months: :uncontrolled)
+  end
```
ssrihari, line 21: I'm afraid of writing out the controlled and uncontrolled queries (then the missed-visits and visited-no-bp queries) as independent functions. There is an inherent coupling between them, and we should strive to maintain that.

For example, if the denominator of the query changes for one treatment outcome, it is important to change it in other places as well. Any divergence in these queries will likely introduce data variance on the dashboard which we are actively trying to avoid.

How do you feel about having a single "treatment outcome" query, that we can read any treatment outcome number from? I had spiked out one approach in the other PR: https://github.com/simpledotorg/simple-server/pull/2619/files#diff-88aaebae700a1b0c64581250b0d554092bd7e8adbc64a85fc427a0fad71a7024R88-R112

app/services/refresh_materialized_views.rb
```diff
@@ -39,37 +42,44 @@ def self.tz
...
+    LatestBloodPressuresPerPatientPerQuarter
+    BloodPressuresPerFacilityPerDay
+    PatientRegistrationsPerDayPerFacility
+    MaterializedPatientSummary
+  ].freeze
+  V2_MATVIEWS = %w[
+    ReportingPipeline::PatientBloodPressuresPerMonth
+    ReportingPipeline::PatientStatesPerMonth
+    ReportingPipeline::PatientVisitsPerMonth
+  ].freeze
```
ssrihari, line 57: Added notes about refreshes in the other PR.

spec/factories/blood_pressures.rb
```diff
@@ -32,6 +32,13 @@
       diastolic { rand(60..89) }
     end
   end
+
+  # We are making this as a new default BP factory to move to, because many of our tests
+  # will depend on the existance of an associated observation / encounter. We don't want to
+  # change the main factory as there would be too much test suite fallout.
+  factory :bp_with_encounter, parent: :blood_pressure do
+    with_encounter
+  end
```
ssrihari, line 41: This doesn't feel very different from using the trait to me. What is the path you're envisioning to creating encounters by default?

spec/models/patient_spec.rb
```diff
@@ -330,14 +322,6 @@ def with_reporting_time_zones(&blk)
...
 
       describe "timezone-specific boundaries" do
-        def with_reporting_time_zones(&blk)
-          Time.use_zone(reporting_timezone) do
-            Groupdate.time_zone = reporting_timezone
-            blk.call
-            Groupdate.time_zone = nil
-          end
-        end
-
```
ssrihari, line 340: Cool! Is this in favor of the reporting helper?

spec/models/reports/repository_spec.rb
```diff
@@ -412,4 +419,61 @@ def refresh_views
       expect(repo.missed_visits_with_ltfu_rates[slug]).to eq(legacy_results[:missed_visits_with_ltfu_rate])
     end
   end
+
+  context "v2" do
```
ssrihari, line 423: Are we replicating the v1 tests here? Can we instead put them in a loop so it's the exact same tests running the flag on/off? If v1 and v2 are functionally the same, we shouldn't need new tests no?

spec/rails_helper.rb
```diff
@@ -37,6 +37,14 @@ def common_org
     @common_org ||= Organization.find_or_create_by!(name: "Common Test Organization")
   end
 
+  def with_reporting_time_zone(&blk)
+    Time.use_zone(Period::REPORTING_TIME_ZONE) do
+      Groupdate.time_zone = Period::REPORTING_TIME_ZONE
+      blk.call
+      Groupdate.time_zone = nil
+    end
+  end
```
ssrihari, line 46: Do we really want this to be available outside of tests for reporting? If we need it, I feel like that reflects some lack of separation between transactional and reporting models.

ssrihari (changes requested):
The stuff around toggle and scientist looks good, Rob! I feel like we should leverage the v2 schema in a couple of ways for accuracy and performance. I've added more details in the comments.

app/models/reports/schema_v2.rb
rsanheim: removing for now

app/queries/control_rate_query_v2.rb
rsanheim: just fixed that up

app/queries/control_rate_query_v2.rb
rsanheim: Not sure I follow, this doesn't involved the denominator at all.  This is just the numerator.  

All the queries you mention use the same denominator currently in repository, and will continue to with the new matviews as well.

spec/factories/blood_pressures.rb
rsanheim: The goal is to have `with_encounter` be the default, but switch to that over time in multiple PRs.

app/controllers/reports/regions_controller.rb
```diff
@@ -164,6 +171,20 @@ def ltfu_chart_data(repo, range)
     }
   end
 
+  def check_reporting_schema_toggle
+    original = RequestStore[:reporting_schema_v2]
+    RequestStore[:reporting_schema_v2] = true if report_params[:v2]
+    yield
```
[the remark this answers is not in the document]
kpethtel: Shows you how often I write around actions.

app/models/reports/repository.rb
rsanheim: Yup, I think that is definitely worth exploring in the future.  My goal here is to get things correct first, and optimize after that.

app/queries/control_rate_query_v2.rb
rsanheim: For reference, here is the denominator used for calculating rates - https://github.com/simpledotorg/simple-server/blob/04ece50ae87a8885aa723e3b765e18ca4103ac15/app/models/reports/repository.rb#L278

app/models/reports/repository.rb
rsanheim: I looked into using Diffy briefly, and it didn't really give us anything useful out of the box...I'll leave the diffing for a future PR, as there are lots of options out there.

rsanheim (commented):
Okay, bunch of follow up here and some notes about what is going on.

I've also run the most build with mismatches on, so you can see the output for what sorts of failures we get here: https://simple.semaphoreci.com/workflows/9537cc73-085b-42c8-884b-e872da838c1c?pipeline_id=408e321f-e8d6-4139-b1b5-1d5690c999ab.

Note that lots of these are just things we need to fix in our tests, for example changing a `:blood_pressure` fixture to a `:bp_with_encounter` fixture.

For merging this, I'll turn the `raise_on_mismatches` back to ENV var only, and then we can start to move thru mismatches and test fixes in smaller PRs.

spec/models/reporting_pipeline/patient_states_per_month_spec.rb
```diff
@@ -343,50 +343,52 @@ def patient_states(patient, from: nil, to: nil)
...
-          expect(patient_states(patient, to: twelve_months_ago).pluck(:htn_care_state)).to all(eq("under_care"))
-          expect(patient_states(patient, from: twelve_months_ago, to: ten_months_ago).pluck(:htn_care_state)).to all(eq("lost_to_follow_up"))
-          expect(patient_states(patient, from: ten_months_ago).pluck(:htn_care_state)).to all(eq("under_care"))
-
-          expect(patient_states(patient, to: ten_months_ago).pluck(:htn_treatment_outcome_in_last_3_months)).to all(eq("missed_visit"))
-          expect(patient_states(patient, from: ten_months_ago, to: seven_months_ago).pluck(:htn_treatment_outcome_in_last_3_months)).to all(eq("controlled"))
-          expect(patient_states(patient, from: seven_months_ago, to: five_months_ago).pluck(:htn_treatment_outcome_in_last_3_months)).to all(eq("visited_no_bp"))
-          expect(patient_states(patient, from: five_months_ago, to: two_months_ago).pluck(:htn_treatment_outcome_in_last_3_months)).to all(eq("uncontrolled"))
-          expect(patient_states(patient, from: two_months_ago).pluck(:htn_treatment_outcome_in_last_3_months)).to all(eq("missed_visit"))
+        Timecop.freeze(june_2021[:now]) do
```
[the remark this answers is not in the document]
ssrihari: I'll take a shot at it.

app/models/reports/repository.rb
ssrihari: I'm okay with that as well. We should remember that the performance is sub-linear when working with multiple regions in a subquery, and evaluate the performance of the v2 reports after that change has been made.

Also, the correctness aspect might change (or require another verification) since the signature itself might change when we accommodate parent regions. That's part of the reason I am suggesting front-loading that work.

app/models/reports/repository.rb
```diff
@@ -155,17 +164,53 @@ def initialize(regions, periods:)
...
+    private def controlled_v1
       region_period_cached_query(__method__) do |entry|
         control_rate_query.controlled(entry.region, entry.period).count
       end
     end
 
+    private def controlled_v2
+      regions.each_with_object({}).each do |region, hsh|
+        if earliest_patient_recorded_at[region.slug].nil?
+          hsh[region.slug] = {}
```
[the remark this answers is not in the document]
ssrihari: I added another note about this somewhere. I'd like it if we can avoid `group_date` now that we have the necessary abstraction of grouped-times in the reporting tables already. Plain old `group(:month_date)` should suffice imo.

app/models/reports/repository.rb
ssrihari: Cool, yeah, diffing in another PR sounds good. We'll also know by then if it's necessary.

app/queries/control_rate_query_v2.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def controlled_counts(region, range: nil)
+    options = group_date_options(range)
+    controlled(region).group_by_period(:month, :month_date, options).count
+  end
+
+  def uncontrolled(maybe_region)
+    region = maybe_region.region
+    ReportingPipeline::PatientStatesPerMonth
+      .where("assigned_#{region.region.region_type}_region_id" => region.id)
```
ssrihari, line 19: Is this a typo? `region.region.region_type`?

More reason to keep it in one place 😜

app/queries/control_rate_query_v2.rb
ssrihari: Trying to illustrate what I mean by keeping the common/denominator in one place:
```ruby
  def common(maybe_region)
    region = maybe_region.region
    ReportingPipeline::PatientStatesPerMonth
      .where(hypertension: "yes", htn_care_state: "under_care")
      .where("assigned_#{region.region_type}_region_id" => region.id)
      .where("months_since_registration >= ?", Reports::REGISTRATION_BUFFER_IN_MONTHS)
  end
  
  def controlled(maybe_region)
    common(maybe_region).where(htn_treatment_outcome_in_last_3_months: :controlled)
  end

  def uncontrolled(maybe_region)
    common(maybe_region).where(htn_treatment_outcome_in_last_3_months: :uncontrolled)
  end
```

We should also be using the same query for the denominator in treatment outcome rates as well.

spec/models/reports/repository_spec.rb
ssrihari: Got it, thanks for the clarification!

spec/queries/control_rate_query_v2_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  let(:query) { ControlRateQueryV2.new }
+
+  let(:june_1_2018) { Time.parse("June 1, 2018 00:00:00+00:00") }
+  let(:june_1_2020) { Time.parse("June 1, 2020 00:00:00+00:00") }
+  let(:june_30_2020) { Time.parse("June 30, 2020 00:00:00+00:00") }
+  let(:july_2020) { Time.parse("July 15, 2020 00:00:00+00:00") }
+  let(:jan_2019) { Time.parse("January 1st, 2019 00:00:00+00:00") }
+  let(:jan_2020) { Time.parse("January 1st, 2020 00:00:00+00:00") }
+  let(:july_2018) { Time.parse("July 1st, 2018 00:00:00+00:00") }
+  let(:july_1_2020) { Time.parse("July 1st, 2020 00:00:00+00:00") }
```
ssrihari, line 16: This is the sort of use case that the `june_2021` helper was created for. I thought we could use a common set of known times and boundaries that we're interested in, in all our reporting tests as opposed to creating new times and boundaries in each test.

ssrihari (dismissed):

app/queries/control_rate_query_v2.rb
rsanheim: Sure, I can refactor towards something like that.

`group_date` gives us consistent time zone handling, which is a huge win, and also handles returning a full series with default values, which we also use.  Definitely not looking to move away from that for right now while changing the queries to point at new matviews across the board.

spec/models/reporting_pipeline/patient_states_per_month_spec.rb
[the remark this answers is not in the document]
rsanheim: yea lemme know if you wanna pair on it - I wonder if there is some schema flipping going on here between integer and floats for this.

spec/queries/control_rate_query_v2_spec.rb
rsanheim: I like the idea of having some common date helpers, but so far I prefer the readability of having the actual time stamps created in the test file itself.  Its simple and explicit and clear, and I don't have to do relative date math in my head.

Also, one thing we've ran into in the past is figuring out what timezone `now - 2.years` is operating in -- with a iso8601(ish) timestamp string like above, there is no confusion.  Its right there in the string that gets parsed.

db/migrate/20210702192822_add_indexes_to_patient_states.rb
```diff
@@ -0,0 +1,11 @@
+class AddIndexesToPatientStates < ActiveRecord::Migration[5.2]
+  def change
+    add_index :reporting_patient_states_per_month, [:assigned_facility_region_id], name: "patient_states_assigned_facility"
+    add_index :reporting_patient_states_per_month, [:assigned_block_region_id], name: "patient_states_assigned_block"
+    add_index :reporting_patient_states_per_month, [:assigned_district_region_id], name: "patient_states_assigned_district"
+    add_index :reporting_patient_states_per_month, [:assigned_state_region_id], name: "patient_states_assigned_state"
+    add_index :reporting_patient_states_per_month, [:assigned_organization_region_id], name: "patient_organization_assigned_state"
```
rsanheim, line 7: These cover the five ways we can filter reports on.

db/migrate/20210702192822_add_indexes_to_patient_states.rb
```diff
@@ -0,0 +1,11 @@
...
+    add_index :reporting_patient_states_per_month, [:month_date], name: "patient_states_month_date"
```
rsanheim, line 8: We are almost always grouping or selecting for a specific `month_date`.

db/migrate/20210702192822_add_indexes_to_patient_states.rb
```diff
@@ -0,0 +1,11 @@
...
+    add_index :reporting_patient_states_per_month, [:hypertension, :htn_care_state, :htn_treatment_outcome_in_last_3_months], name: "patient_states_care_state"
```
rsanheim, line 9: This compound index handles the treatment outcome along w/ the HTN = 'yes' and care_state = 'under_care' or LTFU.

rsanheim (commented):
Some notes on the new indexes.

PR #2620, opened 2021-06-25
Refactor DrugstockQuery

prabhanshuguptagit:
**Story card:** [ch3322](https://app.clubhouse.io/simpledotorg/story/3322/add-a-block-level-drug-stock-table)

## Because

We want to add block level and district level totals to drug stock and consumption reports. The query object in its current form is a bit complex and not amenable to these additions.

## This addresses

- Shifts drug stock totalling to SQL using `GROUP BY .. SUM` queries.
- Breaks the drug stocks reports hash into smaller methods at the top level
- Makes the methods in the query object simpler and reusable


app/models/reports/drug_stock_calculation.rb
```diff
@@ -148,5 +148,13 @@ def patient_days_coefficients(state)
...
       end
     end
+
+    def drug_attribute_sum_by_rxnorm_code(drug_stocks, attribute)
+      drug_stocks
+        .select { |drug_stock| drug_stock[attribute].present? }
+        .group_by { |drug_stock| drug_stock.protocol_drug.rxnorm_code }
+        .map { |rxnorm_code, drug_stocks| [rxnorm_code, drug_stocks.pluck(attribute).sum] }
+        .to_h
+    end
```
harimohanraj89, line 158: Ah, a new way to sum ay attribute based on the flat hash structure we have now.

app/queries/drug_stocks_query.rb
```diff
@@ -10,23 +12,18 @@ def initialize(facilities:, for_end_of_month:)
...
 
   def drug_stocks_report
     Rails.cache.fetch(drug_stocks_cache_key,
       expires_in: ENV.fetch("ANALYTICS_DASHBOARD_CACHE_TTL"),
       force: RequestStore.store[:bust_cache]) do
-      {all: drug_stock_totals,
-       facilities: drug_stock_report_for_facilities,
+      {total_patients: total_patients,
+       all_patient_days: all_patient_days,
+       all_drugs_in_stock: all_drugs_in_stock,
```
harimohanraj89, line 23: Can these three attributes be named the way way as the three following ones, without the `by_facility_id` suffix? That would be very consistent and easy to correlate.

app/queries/drug_stocks_query.rb
```diff
@@ -35,152 +32,123 @@ def drug_consumption_report
...
+  memoize def drug_categories
+    drugs.pluck(:drug_category).uniq
   end
 
-  def drug_stock_report_for_facilities
-    @stock_report_for_facilities ||= @facilities.each_with_object({}) { |facility, report|
-      report[facility.id] = drug_stock_for_facility(facility, drug_stocks[facility.id])
-    }
+  memoize def patient_counts_by_facility_id
+    Patient.where(assigned_facility_id: @facilities).group(:assigned_facility_id).count
```
harimohanraj89, line 60: I'd be very curious to see how the code would shake out if we were able to re-use the same codepaths to craft `total_patients` as well as `patient_counts_by_facility_id`. The idea of `Patient.where(assigned_facility_id: something)` and `.count` are common to both paths. The only change is the `group` clause.

It may not be worth it to extract that just yet, given we only have two cases right now (total and by-facility). So feel free to abandon this idea if it gets gnarly.

harimohanraj89: Same thought for the other pairs of methods in here.

harimohanraj89 (changes requested):
Some thoughts and suggestions.

app/queries/drug_stocks_query.rb
prabhanshuguptagit: Changed in dfe7952

app/queries/drug_stocks_query.rb
prabhanshuguptagit: 👍🏼  Will explore this when we add the block level totals.

harimohanraj89 (approved):

PR #2622, opened 2021-06-28
Run standardrb using a regular old CI job

rsanheim:
Our standardrb action broke again this morning, and it had no useful output at all.

We can run this directly on semaphore with a small job, and get better feedback when things go wrong. This also means it will always use the version we have in our Gemfile, which is a nice bonus.


kpethtel (approved):
👍

PR #2648, opened 2021-06-28
3864 add purpose to notification

kpethtel:
**Story card:** [ch3864](https://app.clubhouse.io/simpledotorg/story/3864/add-purpose-to-notifications)

## Because

The previous merge of this work included a problematic migration. We had to revert. This brings back the reverted code and fixes the migration.


rsanheim (approved):
👍

PR #2661, opened 2021-06-29
Update deduplication log index

prabhanshuguptagit:
**Story card:** -

## Because
Follow up to #2616. We added an index on the wrong column.

## This addresses

Drops the index added in #2616. Adds the correct index. Removes the accidentally committed temporary index from schema.


harimohanraj89 (approved):
Looks great. Let's confirm that the index is used before shipping 😅

PR #2666, opened 2021-06-29
Send mat view timings to Datadog

rsanheim:
[ch3963](https://app.clubhouse.io/simpledotorg/story/3963/instrumentation-of-new-mat-views)

## Because

We want to know how long each mat view refresh takes for prod environments...this wires up instrumentation for the existing views first.

### DIFF
--- a/app/services/refresh_materialized_views.rb
+++ b/app/services/refresh_materialized_views.rb
@@ -18,8 +18,17 @@ class RefreshMaterializedViews
     @logger ||= Rails.logger.child(class: self.class.name)
   end
 
+  def benchmark_and_statsd(operation)
+    name = "refresh_matviews.#{operation}"
+    benchmark(name) do
+      Statsd.instance.time(name) do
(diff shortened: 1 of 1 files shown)


timcheadle (approved):

app/services/refresh_materialized_views.rb
```diff
@@ -18,8 +18,17 @@ def logger
     @logger ||= Rails.logger.child(class: self.class.name)
   end
 
+  def benchmark_and_statsd(operation)
```
kpethtel, line 21: Is it an operation? Maybe we could call it "view"?

kpethtel (approved):

app/services/refresh_materialized_views.rb
rsanheim: Yeah, its slightly more general than just the view name because we also capture the timings of the entire run for all views.

PR #2667, opened 2021-06-29
Add v2 matviews to nightly refresh

rsanheim:
**Story card:** [ch3892](https://app.clubhouse.io/simpledotorg/story/3892/wire-up-controlled-uncontrolled-reports-from-pipeline-via-a-toggle)

In my dev environment, with a large dataset, the new mat views take 20 minutes to refresh...so its pretty long for dev, but not unexpected given this data set size:

```
}
[2021-06-29T22:11:45.408-05:00] DEBUG:    (308.9ms)  COMMIT
[2021-06-29T22:11:45.409-05:00] INFO: refresh_matviews.all_v2 (1388043.2ms)
{

[1] pry(main)> Patient.count
=> 516,441
[2] pry(main)> BloodPressure.count
=> 1,867,157
```


app/services/refresh_materialized_views.rb
```diff
@@ -39,37 +42,44 @@ def self.tz
...
+    LatestBloodPressuresPerPatientPerQuarter
+    BloodPressuresPerFacilityPerDay
+    PatientRegistrationsPerDayPerFacility
+    MaterializedPatientSummary
+  ].freeze
+  V2_MATVIEWS = %w[
+    ReportingPipeline::PatientBloodPressuresPerMonth
+    ReportingPipeline::PatientStatesPerMonth
+    ReportingPipeline::PatientVisitsPerMonth
+  ].freeze
```
ssrihari, line 57: @rsanheim We would only need:

```ruby
ReportingPipeline::PatientStatesPerMonth.refresh
```

The `cascade: true` in the scenic refresh function takes care of refreshing the dependent views.

app/models/reporting_pipeline/patient_blood_pressures_per_month.rb
```diff
@@ -0,0 +1,6 @@
+module ReportingPipeline
+  class PatientBloodPressuresPerMonth < Matview
+    self.table_name = "reporting_patient_blood_pressures_per_month"
+    belongs_to :patient
+  end
+end
```
ssrihari, line 6: Can we avoid adding this model until we actually use it in code? I was even trying to avoid adding the `Visits` model, but we wanted to unit test that finely. This model's unit tests are already covered in the `States` model's unit tests.

app/services/refresh_materialized_views.rb
```diff
@@ -39,37 +42,44 @@ def self.tz
...
       end
+      set_last_updated_at
+    end
+  end
 
-      benchmark_and_statsd("MaterializedPatientSummary") do
-        MaterializedPatientSummary.refresh
+  def refresh_v2
+    ActiveRecord::Base.transaction do
+      ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{tz}'")
```
ssrihari, line 76: `ReportingPipeline::Matview.refresh` already takes care of the transaction and the timezone setting. We can just wrap with metrics here.

ssrihari (changes requested):
The metrics/measurement looks good. Just added some notes about refreshing the new matviews.

app/services/refresh_materialized_views.rb
rsanheim: Cascade [doesn't work](https://github.com/scenic-views/scenic/issues/275), and it relies some [own SQL parsing and regexp](https://github.com/scenic-views/scenic/blob/3600a485797fe1dbf30d152cc60b6d2318f04c48/lib/scenic/adapters/postgres/refresh_dependencies.rb#L29) that I wouldn't really trust even if it did work for our case 😄 .  I'll make sure to remove all the `cascade: true` options for the new matviews.

I think its fine to know what our dependency chain is and declare it explicitly in the refresh object...its a pretty key part of these matviews and we shouldn't be refreshing them in isolation for most cases anyways.  New tests that rely on the matviews can call `refresh_v2` for their setup, the same way we had a repeated `refresh` call for the old ones.

app/models/reporting_pipeline/patient_blood_pressures_per_month.rb
rsanheim: I was doing that initially, but then I discovered the cascade issue in testing and using this all locally...and I didn't want to add a manual refresh call for just _one_ of our views.

app/services/refresh_materialized_views.rb
rsanheim: The outer transaction is to rollback if any part of this fails I think, which is why we are doing it w/ the other views as well.

The `tz` piece is also just for safety and consistency w/ the other refresh...the more I think about this the more I think we should just move all the refresh logic here and not have it in the model objects at all.  It doesn't make sense to refresh a matview in isolation, as they are all pretty tightly coupled w/ each other.

spec/models/reporting_pipeline/patient_states_per_month_spec.rb
```diff
@@ -19,7 +19,7 @@
         patient_registered_13m_ago = Timecop.freeze(13.months.ago) { create(:patient) }
         Timecop.freeze(13.months.ago) { create(:blood_pressure, patient: patient_registered_13m_ago) }
 
-        described_class.refresh
+        RefreshMaterializedViews.new.refresh_v2
```
kpethtel, line 22: Hopefully this isn't too much slower in specs than just refreshing the described class.

kpethtel (approved):

app/services/refresh_materialized_views.rb
ssrihari: I actually think it's better to express the dependency tree closer to the model where there's most context about it. Each view can express its own dependencies, that way the refresh logic is in one place. When we change it for tests, we change it for the job as well.

app/services/refresh_materialized_views.rb
ssrihari: Oh well. Thanks for catching that btw! 😅

app/models/reporting_pipeline/patient_blood_pressures_per_month.rb
ssrihari: We could also just call scenic's refresh with a given table name without creating a model. I was doing this before I saw the cascade option.

spec/models/reporting_pipeline/patient_states_per_month_spec.rb
```diff
@@ -5,6 +5,17 @@
...
+  around do |example|
+    # We need to enforce a known time for this test, otherwise we will have intermittent failures. For example,
+    # if we use live system time, many of these specs will fail after 18:30 UTC (ie 14:30 ET) when on the last day of a month,
+    # because that falls into the next day in IST (our reporting time zone). So to prevent confusing failures for
+    # developers or CI during North American afternoons, we freeze to a time that will be the end of the month for
+    # UTC, ET, and IST. Timezones! 🤯
+    Timecop.freeze("June 30 2021 5:30 UTC") do # June 30th 23:00 IST time
+      example.run
+    end
+  end
```
ssrihari, line 17: The issue here is in using relative times. This was always the case, perhaps the reporting tables make the issue more apparent.

My preference would be to use the absolute times for the test as in `june_2021` that we setup in the helper, as opposed to relative times.

app/services/refresh_materialized_views.rb
rsanheim: I mean, you can't really see the dependency tree unless you look at the SQL itself.  Abstractions on abstractions here.  😁 

> I actually think it's better to express the dependency tree closer to the model where there's most context about it. Each view can express its own dependencies, that way the refresh logic is in one place.

I can't think of a way to do that that would be more clear than a method that calls the `refresh` methods in the order we need, the dependency ordering is pretty clear there.  Suggestions welcome though.

spec/models/reporting_pipeline/patient_states_per_month_spec.rb
rsanheim: > My preference would be to use the absolute times for the test as in `june_2021` that we setup in the helper, as opposed to relative times.

Yeah, I definitely prefer absolute times as well. I think the point I was trying to make is that even with the `june_2021` helper, we need to establish an absolute time via `freeze`, or things will fail like this intermittently.  

This spec _does_ use `june_2021` helper btw, but that doesn't help when time is a frustrating continually advancing thing.  😃

app/services/refresh_materialized_views.rb
ssrihari: I just meant having the same list in the patient-state model itself. I'll raise a tiny PR for this perhaps.

spec/models/reporting_pipeline/patient_states_per_month_spec.rb
ssrihari: Hrm, I'm not sure I follow. What are the timestamps in concern that are advancing? Is it the time at which refresh runs that's the issue?

rsanheim: > Hrm, I'm not sure I follow. What are the timestamps in concern that are advancing? Is it the time at which refresh runs that's the issue?

The timestamps that are advancing is system time, i.e. `Time.current` ...without the `Timecop.freeze`, we would have intermittent failures in this spec depending on the date and time the spec is being run.

PR #2668, opened 2021-06-30
Use patients under care for drug stocks calculation

prabhanshuguptagit:
**Story card:** [3089](https://app.clubhouse.io/simpledotorg/story/3089/update-drug-stock-calculations)

## Because
We are currently using assigned patients for drug stock calculations. We should use [patients under care](https://docs.simple.org/what-we-report#patients-under-care) instead.

## This addresses
Updates the patient count calculation in drug stocks queries.


app/controllers/concerns/set_for_end_of_month.rb
```diff
@@ -1,11 +1,14 @@
 module SetForEndOfMonth
```
harimohanraj89, line 1: Not totally related to this work, but this module name seems _super_ generic. I'd have a hard time understanding what it's trying to do (and consequetially, I don't think I quite get what `@for_end_of_month` is for).

Any ideas for better names?

app/controllers/concerns/set_for_end_of_month.rb
```diff
@@ -1,11 +1,14 @@
...
   def set_for_end_of_month
+    # Defaults to current month if it's the last week of the month.
+    @show_current_month = ((Date.current.end_of_month - Date.current).to_i <= 7)
+
     @for_end_of_month ||= if params[:for_end_of_month]
```
harimohanraj89, line 6: Calling this with a `for` prefix seems weird. How about just `end_of_month`? Or `end_of_latest_month`?

harimohanraj89 (changes requested):

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -104,4 +106,8 @@ def populate_facility_sizes
       .reject { |size| size == "community" }
     @facility_sizes = sort_facility_sizes_by_size(@facility_sizes)
   end
+
+  def set_time_zone
+    Time.use_zone(Period::REPORTING_TIME_ZONE) { yield }
```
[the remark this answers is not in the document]
prabhanshuguptagit: This would touch a lot of unrelated controllers, will raise a separate PR for it right away.

app/controllers/concerns/set_for_end_of_month.rb
prabhanshuguptagit: `for_end_of_month` comes from the drug stock model. It indicates which month the drug stock report is being filed for. It _could_ be just `month` but we'd also want to change it in the model.

app/controllers/my_facilities/drug_stocks_controller.rb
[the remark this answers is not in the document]
harimohanraj89: It wouldn't necessarily. My suggestion is to move the method definition to ApplicationController, and set up the `before_action`s only in the two relevant controllers.

app/controllers/concerns/set_for_end_of_month.rb
harimohanraj89: Ohhh gotcha, this lines up with the model. Sweet.

app/controllers/my_facilities/drug_stocks_controller.rb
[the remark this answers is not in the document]
prabhanshuguptagit: Addressed in #2692

harimohanraj89 (approved):
Looks like most of the naming is consistent with the model. Looks great 🚀

PR #2683, opened 2021-07-01
Reporting schema enhancements

ssrihari:
**Story card:** [ch3928](https://app.clubhouse.io/simpledotorg/story/3928/enhancements-to-patient-states-table)

### List of changes (because it can be difficult to parse the sql)
- Add facility size and type (as needed in some adhoc reports, and a couple of dashboard reports)
- Add current age (computed from dob, age, and age_updated_at to disaggregate by age)
- Add other medical history fields (needed in some ihci reports)
- Add `quarter`, `quarter_string`, and `quarters_since_*` (for cohort reports)
- Remove `patient_` prefix from `patient_assigned_facility_id` and `patient_registration_facility_id`
- Add facility_ids for BPs, Encounters, Appointments and Prescription Drugs (needed for new follow up reports)
- Add IDs for related entities to allow adhoc joins with those tables
- Add treatment outcomes for monthly and quarterly cohorts

### Diff in patient states (not adding diff from other views for brevity)
```diff
diff --git a/db/views/reporting_patient_states_per_month_v02.sql b/db/views/reporting_patient_states_per_month_v03.sql
index 973da83a..ed2ab6ab 100644
--- a/db/views/reporting_patient_states_per_month_v02.sql
+++ b/db/views/reporting_patient_states_per_month_v03.sql
@@ -3,24 +3,43 @@ SELECT
     ------------------------------------------------------------
     -- basic patient identifiers
     p.id as patient_id,
+    p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' as recorded_at,
     p.status,
     p.gender,
     p.age,
     p.age_updated_at AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' AS age_updated_at,
     p.date_of_birth,
-    mh.hypertension as hypertension,
+    EXTRACT(YEAR
+        FROM COALESCE(
+            age(p.date_of_birth),
+            make_interval(years => p.age) + age(p.age_updated_at)
+        )
+    ) AS current_age,
 
     ------------------------------------------------------------
     -- data for the month of
+    cal.month_date,
     cal.month,
+    cal.quarter,
     cal.year,
-    cal.month_date,
     cal.month_string,
+    cal.quarter_string,
+
+    ------------------------------------------------------------
+    -- medical history
+    mh.hypertension as hypertension,
+    mh.prior_heart_attack as prior_heart_attack,
+    mh.prior_stroke as prior_stroke,
+    mh.chronic_kidney_disease as chronic_kidney_disease,
+    mh.receiving_treatment_for_hypertension as receiving_treatment_for_hypertension,
+    mh.diabetes as diabetes,
 
     ------------------------------------------------------------
     -- information on assigned facility and parent regions
-    p.assigned_facility_id AS patient_assigned_facility_id,
-    assigned_facility.facility_slug as assigned_facility_slug,
+    p.assigned_facility_id AS assigned_facility_id,
+    assigned_facility.facility_size as assigned_facility_size,
+    assigned_facility.facility_type as assigned_facility_type,
+    assigned_facility.facility_region_slug as assigned_facility_slug,
     assigned_facility.facility_region_id as assigned_facility_region_id,
     assigned_facility.block_slug as assigned_block_slug,
     assigned_facility.block_region_id as assigned_block_region_id,
@@ -33,8 +52,10 @@ SELECT
 
     ------------------------------------------------------------
     -- information on registration facility and parent regions
-    p.registration_facility_id AS patient_registration_facility_id,
-    registration_facility.facility_slug as registration_facility_slug,
+    p.registration_facility_id AS registration_facility_id,
+    registration_facility.facility_size as registration_facility_size,
+    registration_facility.facility_type as registration_facility_type,
+    registration_facility.facility_region_slug as registration_facility_slug,
     registration_facility.facility_region_id as registration_facility_region_id,
     registration_facility.block_slug as registration_block_slug,
     registration_facility.block_region_id as registration_block_region_id,
@@ -46,23 +67,43 @@ SELECT
     registration_facility.organization_region_id as registration_organization_region_id,
 
     ------------------------------------------------------------
-    -- latest visit info for the month
+    -- details of the visit: latest BP, encounter, prescription drug and appointment
+    bps.blood_pressure_id as blood_pressure_id,
+    bps.blood_pressure_facility_id AS bp_facility_id,
     bps.blood_pressure_recorded_at AS bp_recorded_at,
     bps.systolic,
     bps.diastolic,
-    visits.visited_at AS visited_at,
-    p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' as recorded_at,
 
+    visits.encounter_id AS encounter_id,
+    visits.encounter_facility_id AS encounter_facility_id,
+    visits.encounter_recorded_at AS encounter_recorded_at,
+
+    visits.prescription_drug_id AS prescription_drug_id,
+    visits.prescription_drug_facility_id AS prescription_drug_facility_id,
+    visits.prescription_drug_recorded_at AS prescription_drug_recorded_at,
+
+    visits.appointment_id AS appointment_id,
+    visits.appointment_creation_facility_id AS appointment_creation_facility_id,
+    visits.appointment_recorded_at AS appointment_recorded_at,
+
+    ------------------------------------------------------------
+    -- relative time calculations
 
     (cal.year - DATE_PART('year', p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')))) * 12 +
     (cal.month - DATE_PART('month', p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE'))))
     AS months_since_registration,
 
+    (cal.year - DATE_PART('year', p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')))) * 4 +
+    (cal.month - DATE_PART('quarter', p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE'))))
+    AS quarters_since_registration,
+
     visits.months_since_visit AS months_since_visit,
-    bps.months_since_bp_observation AS months_since_bp,
+    visits.quarters_since_visit AS quarters_since_visit,
+    bps.months_since_bp AS months_since_bp,
+    bps.quarters_since_bp AS quarters_since_bp,
 
     ------------------------------------------------------------
-    -- categorization
+    -- indicators
     CASE
         WHEN (bps.systolic IS NULL OR bps.diastolic IS NULL) THEN 'unknown'
         WHEN (bps.systolic < 140 AND bps.diastolic < 90) THEN 'controlled'
@@ -79,7 +120,7 @@ SELECT
 
           OR
 
-          bps.months_since_bp_observation < 12
+          bps.months_since_bp < 12
         ) THEN 'under_care'
         ELSE 'lost_to_follow_up'
         END
@@ -87,11 +128,27 @@ SELECT
 
     CASE
         WHEN (visits.months_since_visit >= 3 OR visits.months_since_visit is NULL) THEN 'missed_visit'
-        WHEN (bps.months_since_bp_observation >= 3 OR bps.months_since_bp_observation is NULL) THEN 'visited_no_bp'
+        WHEN (bps.months_since_bp >= 3 OR bps.months_since_bp is NULL) THEN 'visited_no_bp'
+        WHEN (bps.systolic < 140 AND bps.diastolic < 90) THEN 'controlled'
+        ELSE 'uncontrolled'
+        END
+        AS htn_treatment_outcome_in_last_3_months,
+
+    CASE
+        WHEN (visits.months_since_visit >= 2 OR visits.months_since_visit is NULL) THEN 'missed_visit'
+        WHEN (bps.months_since_bp >= 2 OR bps.months_since_bp is NULL) THEN 'visited_no_bp'
         WHEN (bps.systolic < 140 AND bps.diastolic < 90) THEN 'controlled'
         ELSE 'uncontrolled'
         END
-        AS htn_treatment_outcome_in_last_3_months
+        AS htn_treatment_outcome_in_last_2_months,
+
+    CASE
+        WHEN (visits.quarters_since_visit >= 1 OR visits.quarters_since_visit is NULL) THEN 'missed_visit'
+        WHEN (bps.quarters_since_bp >= 1 OR bps.quarters_since_bp is NULL) THEN 'visited_no_bp'
+        WHEN (bps.systolic < 140 AND bps.diastolic < 90) THEN 'controlled'
+        ELSE 'uncontrolled'
+        END
+        AS htn_treatment_outcome_in_last_quarter
 
 FROM patients p
 LEFT OUTER JOIN reporting_months cal
@@ -103,7 +160,6 @@ LEFT OUTER JOIN reporting_patient_blood_pressures_per_month bps
     ON p.id = bps.patient_id AND cal.month = bps.month AND cal.year = bps.year
 LEFT OUTER JOIN reporting_patient_visits_per_month visits
     ON p.id = visits.patient_id AND cal.month = visits.month AND cal.year = visits.year
-
 LEFT OUTER JOIN medical_histories mh
     ON p.id = mh.patient_id
     AND mh.deleted_at IS NULL

```


harimohanraj89:
Another idea for you for this PR: Consistent-ize the region field names. We've got an extra `patient_` prefix on a [few fields](https://github.com/simpledotorg/simple-server/blob/drug-stocks-refactor-queries/db/views/reporting_patient_states_per_month_v02.sql#L22)

db/migrate/20210702192823_enhance_reporting_schema.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    add_index :reporting_patient_states_per_month, [:assigned_facility_region_id], name: "patient_states_assigned_facility"
+    add_index :reporting_patient_states_per_month, [:assigned_block_region_id], name: "patient_states_assigned_block"
+    add_index :reporting_patient_states_per_month, [:assigned_district_region_id], name: "patient_states_assigned_district"
+    add_index :reporting_patient_states_per_month, [:assigned_state_region_id], name: "patient_states_assigned_state"
+    add_index :reporting_patient_states_per_month, [:assigned_organization_region_id], name: "patient_organization_assigned_state"
+    add_index :reporting_patient_states_per_month, [:month_date], name: "patient_states_month_date"
+    add_index :reporting_patient_states_per_month, [:hypertension, :htn_care_state, :htn_treatment_outcome_in_last_3_months], name: "patient_states_care_state"
+  end
+end
```
harimohanraj89, line 49: Can you remind me why we drop and recreate, instead of version bumping?

db/views/reporting_months_v02.sql
```diff
@@ -0,0 +1,13 @@
...
+    FROM generate_series('2018-01-01', now(), '1 month')
+)
+
+SELECT
+    month_date,
+    EXTRACT(MONTH from month_date) as month,
+    EXTRACT(QUARTER from month_date) as quarter,
+    EXTRACT(YEAR FROM month_date) as year,
+    to_char(month_date, 'YYYY-MM') as month_string,
+    to_char(month_date, 'YYYY-Q') as quarter_string
```
harimohanraj89, line 12: Suite.

db/views/reporting_facilities_v02.sql
```diff
@@ -0,0 +1,29 @@
...
+    district_regions.id AS district_region_id,
+    district_regions.name AS district_name,
+    district_regions.slug AS district_slug,
+    state_regions.id AS state_region_id,
+    state_regions.name AS state_name,
+    state_regions.slug AS state_slug,
+    org_regions.source_id AS organization_id,
+    org_regions.id AS organization_region_id,
+    org_regions.name AS organization_name,
+    org_regions.slug AS organization_slug
```
harimohanraj89, line 22: This is rich. ✨

db/views/reporting_patient_blood_pressures_per_month_v02.sql
```diff
@@ -0,0 +1,52 @@
...
+    (cal.month - DATE_PART('month', bp.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE'))))
+    AS months_since_bp,
+
+    (cal.year - DATE_PART('year', bp.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')))) * 4 +
+    (cal.quarter - DATE_PART('quarter', bp.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE'))))
+    AS quarters_since_bp
+
+FROM blood_pressures bp
+         -- Only fetch BPs that happened on or before the selected calendar month
+-- We use year and month comparisons to avoid timezone errors
```
harimohanraj89, line 43: Some indentation-fu required here.

db/views/reporting_patient_states_per_month_v03.sql
```diff
@@ -0,0 +1,173 @@
...
+    ------------------------------------------------------------
+    -- relative time calculations
+
+    (cal.year - DATE_PART('year', p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')))) * 12 +
+    (cal.month - DATE_PART('month', p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE'))))
+    AS months_since_registration,
+
+    (cal.year - DATE_PART('year', p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')))) * 4 +
+    (cal.month - DATE_PART('quarter', p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE'))))
+    AS quarters_since_registration,
```
harimohanraj89, line 98: The repetition here makes me wonder whether `_since_registration` is required in the BPs and Visits tables. 🤔 

If we're not sure, let's leave it as is.

db/views/reporting_patient_states_per_month_v03.sql
```diff
@@ -0,0 +1,173 @@
...
+        END
+        AS htn_treatment_outcome_in_last_2_months,
+
+    CASE
+        WHEN (visits.quarters_since_visit >= 1 OR visits.quarters_since_visit is NULL) THEN 'missed_visit'
+        WHEN (bps.quarters_since_bp >= 1 OR bps.quarters_since_bp is NULL) THEN 'visited_no_bp'
+        WHEN (bps.systolic < 140 AND bps.diastolic < 90) THEN 'controlled'
+        ELSE 'uncontrolled'
+        END
+        AS htn_treatment_outcome_in_last_quarter
```
harimohanraj89, line 151: This is the outcome in _this_ quarter, right? Can we rename to `in_this_quarter`, or `in_quarter`?

db/views/reporting_patient_states_per_month_v03.sql
```diff
@@ -0,0 +1,173 @@
...
+    visits.encounter_facility_id AS encounter_facility_id,
+    visits.encounter_recorded_at AS encounter_recorded_at,
+
+    visits.prescription_drug_id AS prescription_drug_id,
+    visits.prescription_drug_facility_id AS prescription_drug_facility_id,
+    visits.prescription_drug_recorded_at AS prescription_drug_recorded_at,
+
+    visits.appointment_id AS appointment_id,
+    visits.appointment_creation_facility_id AS appointment_creation_facility_id,
+    visits.appointment_recorded_at AS appointment_recorded_at,
```
harimohanraj89, line 87: From all these facilities, how do we ask "what's the facility that the patient last visited?"

harimohanraj89 (approved):
Several questions mainly!

db/migrate/20210702192823_enhance_reporting_schema.rb
ssrihari: We need to do this when other views depend on the view that we're updating. Scenic isn't intelligent enough to handle that. And in this PR, we're updating pretty much all the views including `reporting_months` and `reporting_facilities`.

db/views/reporting_patient_blood_pressures_per_month_v02.sql
ssrihari: Nice catch!

db/views/reporting_patient_states_per_month_v03.sql
ssrihari: It might be useful if we're using those tables in ad-hoc reports. It's useful to leave in there, ya.

db/views/reporting_patient_states_per_month_v03.sql
ssrihari: Ha! Yes, I missed this while renaming it to be similar to the other ones. Will rename to `in_quarter`.

db/migrate/20210702192823_enhance_reporting_schema.rb
harimohanraj89: Gotcha. Then the only thing that makes me nervous is remembering to recreate all the right indices. Is there a clever way we can capture these indices somewhere permanent?

db/views/reporting_patient_states_per_month_v03.sql
ssrihari: A dirty solution would be:
```sql
    CASE
        WHEN visits.visited_at = visits.prescription_drug_recorded_at THEN visits.prescription_drug_facility_id  
        WHEN visits.visited_at = visits.appointment_recorded_at THEN visits.appointment_creation_facility_id    
        ELSE visits.encounter_facility_id
        AS latest_facility_id
```

I was thinking about follow ups where we have to answer if the patient visited a given facility in a month. And that would be:
```sql
'given_facility' IN
array_remove(
        ARRAY[
            (CASE WHEN to_char(encounter_recorded_at, 'YYYY-MM') = month_string THEN encounter_facility_id),
            (CASE WHEN to_char(prescription_drug_recorded_at, 'YYYY-MM') = month_string THEN prescription_drug_facility_id),
            (CASE WHEN to_char(appointment_recorded_at, 'YYYY-MM') = month_string THEN appointment_facility_id)
        ],
    null)
```

Let me think of better solutions, or add these to the view. Good question though!

rsanheim (commented):
The linked story for this PR looks to be pretty focused on adding support for cohort reports, but the actual changes in this PR have a much wider scope than that. I know there is a lot we want to do w/ the reporting schema, but if we keep the PRs more focused it will help us get things live smoother and decrease some of our feedback cycles here. 😁 

I'm leaving some inline feedback on the diff.  Some questions on the motivation behind these changes:

> ### List of changes (because it can be difficult to parse the sql)
> * Add facility size and type (as needed in some adhoc reports, and a couple of dashboard reports)
> * Add current age (computed from dob, age, and age_updated_at to disaggregate by age)
👍 
> * Add other medical history fields (needed in some ihci reports)
What IHCI reports? Do you mean ad-hoc reports we generate for them?  Can we join the data from the matviews to medical_history if we need that additional info?
> * Add `quarter`, `quarter_string`, and `quarters_since_*` (for cohort reports)
👍 
> * Remove `patient_` prefix from `patient_assigned_facility_id` and `patient_registration_facility_id`
👍 
> * Add facility_ids for BPs, Encounters, Appointments and Prescription Drugs (needed for new follow up reports)
Can you link to a story for these reports? 
> * Add IDs for related entities to allow adhoc joins with those tables
> * Add treatment outcomes for monthly and quarterly cohorts

One tension we are facing here is optimizing these matviews for our existing reports in the dashboard and progress tab, versus adding all the things to make ad-hoc or one-off reports easier. Everything we add for ad-hoc reports that we *don't* need for our current dashboard reports is going add complexity and hurt performance for the reports we have in the app.

I think we should build these matviews for the reports in the app first and foremost, and if we need more of a 'kitchen-sink' type approach for ad-hoc reports perhaps those could be added to a newer, much wider matview that is more of a scratch-space to make ad-hoc reporting easier.

db/views/reporting_months_v02.sql
harimohanraj89: I think it's more valuable for this string to be used for queries and ordering, rather than display. Keeping it `YYYY-Q` makes it much easier to sort - `WHERE quarter_string >= '2020-3'`

db/views/reporting_patient_states_per_month_v03.sql
harimohanraj89: Haha yeah, my initial thought was - "do we need to precompute the "visit_facility_id" in the visits table itself? Maybe 🤔

db/migrate/20210702192823_enhance_reporting_schema.rb
ssrihari: > Scenic will recreate views for us

recreate indexes you mean? We still have to drop and re-create the dependent views since scenic doesn't track that.

db/views/reporting_patient_states_per_month_v03.sql
```diff
@@ -0,0 +1,173 @@
...
+    cal.quarter_string,
+
+    ------------------------------------------------------------
+    -- medical history
+    mh.hypertension as hypertension,
+    mh.prior_heart_attack as prior_heart_attack,
+    mh.prior_stroke as prior_stroke,
+    mh.chronic_kidney_disease as chronic_kidney_disease,
+    mh.receiving_treatment_for_hypertension as receiving_treatment_for_hypertension,
+    mh.diabetes as diabetes,
```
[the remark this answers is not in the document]
ssrihari: Adding additional columns, especially these tiny text ones, doesn't hurt our performance. The largest part time in the queries (if you've looked at the query plan for most of our analytics queries) is spent on finding the right rows to fetch. That happens via an index, usually. Once that happens, holding a few extra columns in memory is actually quite cheap.

Regarding the performance of this query itself, we already need to join with medical histories to fetch a few fields, so adding some more to the result is cheap.

Denormalisation is key in making queries simple in reporting. I've seen tables with over a hundred columns, and it makes a lot of sense when reporting on them. We have a few incoming adhoc reports that need this information and I see a possibility of those kinds of reports making it to the dashboard as well.

db/migrate/20210702192823_enhance_reporting_schema.rb
```diff
@@ -0,0 +1,49 @@
+class EnhanceReportingSchema < ActiveRecord::Migration[5.2]
+  def up
+    drop_view :reporting_patient_states_per_month, materialized: true
+    drop_view :reporting_patient_visits_per_month, materialized: true
+    drop_view :reporting_patient_blood_pressures_per_month, materialized: true
```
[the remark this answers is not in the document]
ssrihari: To someone new looking at these tables, there is no indication that this row will have one row per patient per month. I would expect a table called "Patient States" to have one row per patient, intuitively.

db/views/reporting_patient_states_per_month_v03.sql
```diff
@@ -0,0 +1,173 @@
...
+
+    ------------------------------------------------------------
+    -- information on assigned facility and parent regions
+    p.assigned_facility_id AS assigned_facility_id,
+    assigned_facility.facility_size as assigned_facility_size,
+    assigned_facility.facility_type as assigned_facility_type,
+    assigned_facility.facility_region_slug as assigned_facility_slug,
```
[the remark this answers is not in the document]
ssrihari: > they increase the view width dramatically

How so? And, how it does it affect us?

> don't really give us anything
> querying by slug is imprecise

We use slugs in our URLs, and look up regions primarily by slugs in our dashboard. If looking up by slug is imprecise, we should change our URLs as well. The motivation was that one could use region slugs to reach directly into the tables for reports, either from the dashboard or from metabase/elsewhere.

db/views/reporting_patient_states_per_month_v03.sql
```diff
@@ -0,0 +1,173 @@
...
+    visits.encounter_id AS encounter_id,
+    visits.encounter_facility_id AS encounter_facility_id,
+    visits.encounter_recorded_at AS encounter_recorded_at,
+
+    visits.prescription_drug_id AS prescription_drug_id,
+    visits.prescription_drug_facility_id AS prescription_drug_facility_id,
+    visits.prescription_drug_recorded_at AS prescription_drug_recorded_at,
+
+    visits.appointment_id AS appointment_id,
+    visits.appointment_creation_facility_id AS appointment_creation_facility_id,
```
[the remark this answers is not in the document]
ssrihari: > Why are we adding all this?

I mention this in the PR overview/description. This is needed for new follow up reports: https://app.clubhouse.io/simpledotorg/epic/3817/update-reports-follow-up-tables-in-dashboard

> going to slow down all the reporting queries we need to run for dashboard and progress tab

Why do you think so? Like I said in a previous comment, adding to a row's size doesn't impact reports by a lot. The entire patient states table is a maximum of 5G in production, which will likely be in memory of the RDS box. Adding a few ID columns will not affect the performance by much. The query time as far as I've seen, is mostly in filtering to the right rows.

db/views/reporting_patient_states_per_month_v03.sql
```diff
@@ -0,0 +1,173 @@
...
+    AS months_since_registration,
+
+    (cal.year - DATE_PART('year', p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')))) * 4 +
+    (cal.month - DATE_PART('quarter', p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE'))))
+    AS quarters_since_registration,
+
+    visits.months_since_visit AS months_since_visit,
+    visits.quarters_since_visit AS quarters_since_visit,
+    bps.months_since_bp AS months_since_bp,
+    bps.quarters_since_bp AS quarters_since_bp,
```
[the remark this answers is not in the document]
ssrihari: Yup, we need relative times in quarters just like we have them in months. We ask of things like "previous quarter", or "next quarter", so this would help in answering those kinds of questions.

ssrihari:
In addition to my comments, I feel like most of your concerns about performance would be addressed by the addition of the [facility-states-per-month](https://app.clubhouse.io/simpledotorg/story/3934/create-facility-states-per-month-mat-view) matview, which in contrast to these matviews will be closely built to accommodate the dashboard's needs.

ssrihari:
> but the actual changes in this PR have a much wider scope than that
> but if we keep the PRs more focused it will help us get things live smoother and decrease some of our feedback cycles here

Yeah, this one was an amalgamation of things that cropped up since the time I wrote that story. Feedback taken, I'll try to keep upcoming PRs more focused. @rsanheim @harimohanraj89

db/views/reporting_patient_states_per_month_v03.sql
ssrihari: @harimohanraj89 There could be multiple visits in a month across multiple facilities. I'm thinking of adding `visited_facility_ids`, wdyt?

rsanheim (approved):
👍 I'll update the indexes PR once this gets merged.

PR #2684, opened 2021-07-01
3944: rename NotificationService to TwilioApiService and allow it to …

kpethtel:

## Because

The NotificationService is tightly coupled with Twilio, so I'm renaming it to reflect that coupling. I'm also changing the way that errors are handled within the notification service to force sidekiq retries.


app/services/twilio_api_service.rb
```diff
@@ -1,8 +1,5 @@
-class NotificationService
```
kpethtel, line 1: I refactored this class a while ago to be more object oriented and to avoid having to rescue twilio errors outside of this class. Looks like I'm coming full circle. The reason I'm choosing to allow the errors to bubble up again is to better match the pattern @rsanheim recommended for handling Imo invitations.

app/services/twilio_api_service.rb
```diff
@@ -12,6 +9,14 @@ class NotificationService
   TWILIO_TEST_SMS_NUMBER = "+15005550006"
   TWILIO_TEST_WHATSAPP_NUMBER = "+14155238886"
 
+  class Error < StandardError
```
kpethtel, line 12: I'm open to leaving out this custom error class and letting sentry pick up the twilio errors.

app/jobs/appointment_notification/worker.rb
```diff
@@ -67,20 +67,6 @@ def send_message(notification, communication_type)
       raise UnknownCommunicationType, "#{self.class.name} is not configured to handle communication type #{communication_type}"
     end
 
-    log_info = {
```
kpethtel, line 70: Instead of capturing the error and logging here, I'm just letting the twilio service fail, which will prompt a retry.

app/services/twilio_api_service.rb
```diff
@@ -12,6 +9,14 @@ class NotificationService
...
+    attr_reader :exception_message
+    def initialize(message, exception_message: nil)
+      super(message)
+      @exception_message = exception_message
+    end
+  end
+
```
rsanheim, line 19: Don't need to define anything custom, plain ruby exceptions have all we need here.

```suggestion
  class Error < StandardError
  end
```

app/services/twilio_api_service.rb
```diff
@@ -86,19 +87,6 @@ def send_twilio_message(sender_number, recipient_number, message, callback_url =
...
-    Sentry.capture_message(
-      "Error while processing notification",
-      extra: {
-        exception: e.to_s
-      },
-      tags: {
-        type: "notification-service"
-      }
-    )
+    raise Error.new("Error while calling the Twilio API", exception_message: exception)
```
rsanheim, line 90: Ruby will wrap the original exception automatically for it and maintain that as the `cause` on our Error:

```suggestion
    raise Error, "Error while calling Twilio API"
  ```

**Review by rsanheim — APPROVED** [review]
Some minor feedback on the exception stuff, but otherwise LGTM

File: app/services/twilio_api_service.rb [hunk 6]
```diff
@@ -85,20 +81,7 @@ def send_twilio_message(sender_number, recipient_number, message, callback_url =
...
-    Sentry.capture_message(
-      "Error while processing notification",
-      extra: {
-        exception: e.to_s
-      },
-      tags: {
-        type: "notification-service"
-      }
-    )
+  rescue Twilio::REST::TwilioError
```
harimohanraj89, line 84: Yup I think we should definitely be letting the Twilio error pass through un-rescued. We might be losing contextual info that's attached to the Twilio error.

app/services/twilio_api_service.rb
```diff
@@ -85,20 +81,7 @@ def send_twilio_message(sender_number, recipient_number, message, callback_url =
...
-  def report_error(e)
-    Sentry.capture_message(
-      "Error while processing notification",
-      extra: {
-        exception: e.to_s
-      },
-      tags: {
-        type: "notification-service"
-      }
-    )
```
harimohanraj89, line 102: Awesome. Yeah I don't think we're gaining much by rescuing and manually reporting to Sentry here.

app/jobs/appointment_notification/worker.rb
harimohanraj89: I think the log is still useful. The Twilio error won't have context on what communication or notification ID, making it hard to diagnose when something does go wrong? I'd also be open to rescuing the exception, attaching this extra info to the exception, and re-raising the enriched exception. THis'll generate a rich error report that we can use to effectively diagnose. I'm not sure how easy it will be to attach context to the exception object itself though 😅

app/jobs/appointment_notification/worker.rb
```diff
@@ -67,20 +67,6 @@ def send_message(notification, communication_type)
...
-      failed: !!notification_service.failed?,
-      error: notification_service.error,
-      sender: medication_reminder_sms_sender || "default",
-      communication_type: communication_type,
-      notification_id: notification.id
-    }
-
-    logger.info log_info
-
-    return if notification_service.failed?
```
harimohanraj89, line 82: The big change in behavior is that notification failure now raises an exception. This will trigger a retry automatically as long as the notification service is being invoked in Sidekiq. Are there adverse consequences to this? What kinds of errors can we expect to see? Will we be retrying notifications in situations where it will never succeed? Eg. sending to an invalid phone number?

Does Twilio's exceptions give us some granular information (like discrete error codes) that we can use to selectively decide _when_ we want to retry (eg. network blip), and when we don't (eg. bad phone number)?

lib/tasks/scripts/message_patients.rb
```diff
@@ -41,16 +41,16 @@ def send_messages
...
-      notification_service = NotificationService.new
-      if whatsapp?
-        notification_service.send_whatsapp(phone_number, message)
-      elsif sms?
-        notification_service.send_sms(phone_number, message)
-      end
-      if notification_service.failed?
-        update_report(:exception, patient: patient)
-      else
+      notification_service = TwilioApiService.new
```
harimohanraj89, line 44: This is the pivotal point of decoupling. Very cool.

harimohanraj89 (changes requested):
The rename is solid ✨ but I want to think through the exception handling and logging a bit more. Comments below!

app/jobs/appointment_notification/worker.rb
kpethtel: I believe we only expect twilio errors on network errors, but that's largely guesswork. This code was all in place before I started working on it. When I found it, this worker was rescuing from the twilio error and logging. IMO, this class shouldn't need to know anything about the twilio error. That should all be handled from the twilio client. The twilio client is used in other places, so we had multiple `rescue TwilioError` in our code, which strikes me as a bad pattern. And it will only get worse when we add Imo, which is why I'm making changes. I think the other option is to have a `NotificationService` that handles the Twilio and Imo clients. We can pass some context into it and pass it into the error class. I don't think we should have this logging here, though. It would be better to have it all in one place (i.e., sentry log for the error) not spread out across Sentry and rails logs. I'll look into what would be causing twilio errors (i.e., bad number or just network error) and try to come up with a better solution.

harimohanraj89: The major point I wanna think through is - will this make our code retry jobs again and again that are never destined to pass. If we can find out by taking a look at the Twilio gem or API error codes, etc, it may help us understand better, and my concerns would be addressed 👍

kpethtel: I agree with that point, and I'll try to figure that out. We did have a bug that resent the same messages every day because they never succeeded, and I don't want to recreate the problem. My point is really more about keeping the error handling as close to the origin as possible and not spreading its logging across systems.

kpethtel: As best I can tell, Sentry only keeps logs for 90 days. Notifications have been turned off for some of that time, but I only see two instances of twilio errors. Both are for a bad number. I can try to avoid retries for this case, but it should also capture network errors. 
https://sentry.io/organizations/resolve-to-save-lives/issues/2443427939/?end=2021-07-02T10%3A51%3A54&environment=production&project=1793711&query=Error+while+processing+notification&start=2021-04-03T10%3A51%3A54&utc=true

app/services/twilio_api_service.rb
```diff
@@ -49,56 +56,55 @@ def test_client
...
       from: sender_number,
       to: recipient_number,
       status_callback: callback_url,
       body: message
     )
   rescue Twilio::REST::TwilioError => exception
-    @error = exception
-    report_error(exception)
+    # see the link above for a list of codes; the else case happens on network failures
+    if exception.respond_to?(:code)
```
kpethtel, line 93: I looked through the error codes and we probably don't want to retry on any of them and probably also don't want to alert Sentry on any of them (maaaybe on one of them, but I'm not sure it's worth the extra code complexity).

lib/tasks/scripts/message_patients.rb
```diff
@@ -41,16 +41,26 @@ def send_messages
...
-      end
-      if notification_service.failed?
+      notification_service = TwilioApiService.new
+      context = {
+        calling_class: self.class.name,
+        patient_id: patient.id,
+        communication_type: channel
+      }
+
+      begin
```
kpethtel, line 51: I can't say I like all this branching logic.

app/services/twilio_api_service.rb
```diff
@@ -49,56 +56,55 @@ def test_client
...
+    client.messages.create(
       from: sender_number,
       to: recipient_number,
       status_callback: callback_url,
       body: message
     )
   rescue Twilio::REST::TwilioError => exception
-    @error = exception
-    report_error(exception)
+    # see the link above for a list of codes; the else case happens on network failures
```
harimohanraj89, line 92: Which link? Shall we just put it here too?

app/services/twilio_api_service.rb
harimohanraj89: Let's put this knowledge into a comment here. eg. `Exceptions with codes are Twilio errors that will not succeed on retry. Log the error and swallow the exception.`

app/services/twilio_api_service.rb
harimohanraj89: Nah the extra context is useful. If we can't easily attach that context to the twilio error itself, then a custom error is probably more useful 👍

lib/tasks/scripts/message_patients.rb
```diff
@@ -41,16 +41,26 @@ def send_messages
...
+        response = if whatsapp?
+          notification_service.send_whatsapp(recipient_number: phone_number, message: message, context: context)
+        elsif sms?
+          notification_service.send_sms(recipient_number: phone_number, message: message, context: context)
+        end
+      rescue TwilioApiService::Error
```
harimohanraj89, line 57: Ah I see. In this script, we don't want to halt on exception, so we rescue and report. Neat.

app/services/twilio_api_service.rb
```diff
@@ -49,56 +56,55 @@ def test_client
...
-        type: "notification-service"
-      }
-    )
+  def log_error(e, sender_number, recipient_number, context)
+    logger.info({
+      class: self.class.name,
+      error: e,
+      sender: sender_number,
+      recipient: recipient_number,
+      context: context
```
harimohanraj89, line 107: Consider flattening the context keys directly into the hash. That'll make it easier to read and perhaps parse the logs too.

```
({
  class: self.class.name,
  error: e,
  sender: sender_number,
  recipient: recipient_number,
  **context
})
```

harimohanraj89 (approved):
Few last minor comments, but nothing blocking. The crux was nailing down the "retry" scenarios vs the "nope, this won't work. Report and move on" scenarios.

Looks great!

app/services/twilio_api_service.rb
kpethtel: I'll move it to the top of the file and explain this more clearly.

app/services/twilio_api_service.rb
kpethtel: I'll add this to a block comment at the top of the file along with the links.

app/services/twilio_api_service.rb
kpethtel: Not a bad idea... Will do.

PR #2687, opened 2021-07-01
Fix patient states spec by limiting the range we assert on

rsanheim:
Postgres builds its reporting_months  view using now(), which means we
will always get records up to the real current time.  We can't rely on
Timecop for that.

### DIFF
--- a/spec/models/reporting_pipeline/patient_states_per_month_spec.rb
+++ b/spec/models/reporting_pipeline/patient_states_per_month_spec.rb
@@ -365,9 +365,9 @@ RSpec.describe ReportingPipeline::PatientStatesPerMonth, {type: :model, reportin
 
             RefreshMaterializedViews.new.refresh_v2
 
-            # NOTE: we have to run this test for a range of up until the current timestamp, because we build this matview
-            # off a join with a calendar months table, which will always have records up until the actual current time.
-            # Timecop only impacts Ruby, and not the actual system time.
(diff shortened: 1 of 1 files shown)


spec/models/reporting_pipeline/patient_states_per_month_spec.rb
```diff
@@ -343,50 +343,56 @@ def patient_states(patient, from: nil, to: nil)
...
+            create(:blood_pressure, :with_encounter, patient: patient, recorded_at: ten_months_ago, systolic: 139, diastolic: 89)
+            create(:prescription_drug, patient: patient, device_created_at: eight_months_ago)
+            create(:blood_pressure, :with_encounter, patient: patient, recorded_at: five_months_ago, systolic: 140, diastolic: 90)
+
+            RefreshMaterializedViews.new.refresh_v2
+
+            # NOTE: we have to run some of these assertions for a range of up until the "current" frozen timestamp, because we build this matview
+            # off a join with the reporting_months view, which uses now() and so will always have records up until
+            # the actual current time.  Timecop only impacts Ruby, and not the actual system time.
+            expect(patient_states(patient, from: two_years_ago, to: now).pluck(:months_since_registration)).to eq((0...24).to_a)
```
kpethtel, line 371: Tricky stuff. Nice fix.

kpethtel (approved):

ssrihari:
🙌

PR #2688, opened 2021-07-02
Overdue list updates

prabhanshuguptagit:
**Story card:** [ch3973](https://app.clubhouse.io/simpledotorg/story/3973/patients-missing-from-overdue-list-download)

## Because

We don’t want to exclude any patients from the download. Any patient who hasn't visited their last appointment that is scheduled or cancelled should be in the overdue list.

All these patients should be included

- Patient has agreed to visit
- Nurse has set a remind_on date for the patient
- Patient has a cancelled appointment

## This addresses

- [x] Add all patients with a cancelled appointment to the list
- [x] Add all patients who have agreed to visit or need to be reminded on a date
- [x] Add an Appointment Status column
- [x] Add a Cancellation reason column
- [x] Add a Remind Patient on date column


app/models/appointment.rb
```diff
@@ -51,16 +51,26 @@ def self.between(start_date, end_date)
...
+
+  def self.all_overdue
+    passed_unvisited
+      .where(status: :scheduled)
+      .where(arel_table[:remind_on].eq(nil).or(arel_table[:remind_on].lteq(Date.current)))
+  end
+
   def self.overdue
-    all_overdue.where(arel_table[:scheduled_date].gteq(365.days.ago))
+    all_overdue.where("appointments.scheduled_date >= ?", 365.days.ago)
```
prabhanshuguptagit, line 73: The `overdue` scope is only used by the filters on the dashboard and can be done directly in `PatientSummaryQuery`. The names `overdue` and `all_overdue` are not clear. We think:
- Remove the `overdue` method
- Rename `all_overdue` to `overdue`
- No need for `last_year_unvisited` as well, do the filtering in `PatientSummaryQuery`

This is a larger change and can come in a follow up PR.

app/models/appointment.rb
```diff
@@ -51,16 +51,26 @@ def self.between(start_date, end_date)
...
-    where(status: "scheduled")
-      .where(arel_table[:scheduled_date].lt(Date.current))
-      .where(arel_table[:remind_on].eq(nil).or(arel_table[:remind_on].lteq(Date.current)))
+  def self.passed_unvisited
+    # Scheduled or cancelled appointments whose scheduled date has passed.
+    where.not(appointments: {status: :visited})
+      .where("appointments.scheduled_date < ?", Date.current)
       .joins(:patient)
       .where.not(patients: {status: :dead})
   end
```
prabhanshuguptagit, line 60: A new definition `passed_unvisited` is being added. It is a superset of `all_overdue`.

app/queries/patient_summary_query.rb
```diff
@@ -18,17 +18,27 @@ def self.label_for(filter)
...
-    result = if filters.include?("only_less_than_year_overdue")
-      relation.overdue
+    result = if @only_overdue
+      if filters.include?("only_less_than_year_overdue")
+        relation.overdue
+      else
+        relation.all_overdue
+      end
+    elsif filters.include?("only_less_than_year_overdue")
+      relation.last_year_unvisited
```
prabhanshuguptagit, line 37: The filtering should be applied/translated to scopes in the controller. Passing the filters as strings leaks in controller concerns. This is a larger change, can tackle in a follow up PR.

app/controllers/appointments_controller.rb
```diff
@@ -91,4 +83,13 @@ def download_filename
     facility_name = current_facility.present? ? current_facility.name.parameterize : "all"
     "overdue-patients_#{facility_name}_#{Time.current.to_s(:number)}.csv"
   end
+
+  def patient_summaries(only_overdue: true)
+    PatientSummaryQuery.call(
```
harimohanraj89, line 88: This interface and interaction is looking a lot slicker. Does this query only service overdue lists? Can we rename the query class?

app/models/appointment.rb
harimohanraj89: I feel like getting these names right is a big piece of this work and future improvements. We have three scopes in diminishing order:
1. All overdue patients regardless of their "latest communication status" - to power the download
2. Overdue patients that need to be contacted right away - to power the dashboard and app lists
3. (2) that are not more than 365 days overdue - to power some filtering on the dashboard

I'm leaning towards calling (1) Overdue, and then setting up _independent_ scopes that can be composed for (2) and (3). Eg.

`Appointment.overdue`
`Appointment.overdue.require_reminder`
`Appointment.overdue.within_one_year_overdue.require_reminder`

The names are awkward, but the concept is that we don't introduce _nested_ scopes, we introduce _independent_ scopes that are chained together, which may lend itself to more intuitive names.

app/models/appointment.rb
harimohanraj89: Ignoring my longer comment about overall naming, I think we can find something slightly better for this name in the short-term. Nit: `past_unvisited`? `missed` (this incorporates the condition of being in the past in the name itself. You can't miss something in the future)?

db/views/patient_summaries_v06.sql
```diff
@@ -40,6 +40,7 @@ greatest(0, date_part('day', NOW() - next_appointment.scheduled_date)) AS days_o
 next_appointment.id AS next_appointment_id,
 next_appointment.scheduled_date AS next_appointment_scheduled_date,
 next_appointment.status AS next_appointment_status,
+next_appointment.cancel_reason AS next_appointment_cancel_reason,
```
harimohanraj89, line 43: Interesting, this is the only missing field? That's awesome!

app/controllers/appointments_controller.rb
```diff
@@ -16,21 +16,13 @@ def index
...
-
-    if current_facility
-      @patient_summaries = @patient_summaries.where(assigned_facility_id: current_facility.id)
-    end
-    @patient_summaries = @patient_summaries.order(risk_level: :desc, next_appointment_scheduled_date: :desc, id: :asc)
-
     respond_to do |format|
       format.html do
-        @patient_summaries = paginate(@patient_summaries).includes(:next_appointment, patient: :appointments)
+        @patient_summaries = paginate(patient_summaries(only_overdue: true)).includes(:next_appointment, patient: :appointments)
```
harimohanraj89, line 21: `only_overdue` seems like a weird name for a flag that generates overdue lists 🤔 I really think we should pursue the idea of `requires_reminder`, `requires_call`, `due_for_call`, or something like that. In plain English, that's a fairly intuitive way to describe the difference between the dashboard and download, right? The dashboard only shows people you need to call, the download shows you everyone?

harimohanraj89 (changes requested):
Some scope structure suggestions, a lot of debate on naming 🤔 I know we're on a clock, but I think solidifying the names correctly here will have a pretty big impact, since we're introducing a new concept ("passed missed appt patients").

app/models/appointment.rb
prabhanshuguptagit: We originally landed on `past_unvisited`. Wanted to avoid `missed_unvisited` because "Missed visit" is a thing.

app/controllers/appointments_controller.rb
prabhanshuguptagit: Yeah, that's pretty much the main difference between dashboard and download. Maybe `all_patients_requiring_call` or `all_patients_requiring_follow_up` would be better here. What about `only_scheduled_overdue`?

app/models/appointment.rb
prabhanshuguptagit: That sounds good. The `overdue` methods have touch points all over that we wanted to avoid with this feature PR. 

> I'm leaning towards calling (1) Overdue
This would mean our "overdue" definition changes to be "scheduled appointment in the past" to "any appointment in the past" though.

harimohanraj89 (approved):
After much zoom discussion, we have come to the shared understanding that:
* We didn't have any better naming system that was intuitively better
* The team needs to have a conversation about what "overdue" means, before we can nail down the namings here.

In light of that uncertainty, and the urgency of this enhancement, let's ship this as is, and start a conversation about what "overdue" means with our team.

PR #2690, opened 2021-07-05
Change online lookup API to move identifier into body for security

ssrihari:
**Story card:** [ch4007](https://app.clubhouse.io/simpledotorg/story/4007/change-online-lookup-api-to-move-identifier-into-body-for-security)


app/schema/api/v4/models.rb
```diff
@@ -365,6 +372,7 @@ def definitions
         lookup_patient: lookup_patient,
         lookup_patients: array_of("lookup_patient"),
         medical_history: Api::V3::Models.medical_history,
```
prabhanshuguptagit, line 374: Is `medical_history` still being used?

spec/controllers/api/v4/patients_controller_spec.rb
```diff
@@ -49,6 +49,17 @@ def set_headers(user, facility)
       expect(response_patient[:appointments].count).to eq 2
     end
 
+    it "works on patients without a medical history" do
```
prabhanshuguptagit, line 52:
```suggestion
    it "sends a nil medical history for patients without a medical history" do
```

prabhanshuguptagit (approved):
Looks good 👌🏼

app/schema/api/v4/models.rb
ssrihari: Good catch! 🙇  Fixed in 84964b8e.

PR #2692, opened 2021-07-06
Consolidate set reporting time zone to ApplicationController

prabhanshuguptagit:
**Story card:** -

## Because

We have a bunch of different implementations of `set_time_zone` across controllers, even though they do the same thing:
- Set `Time.zone`
- Set `Groupdate.timezone`

Follow up to https://github.com/simpledotorg/simple-server/pull/2668#discussion_r662792047

## This addresses

Consolidates them all to `ApplicationController`. The `around_action` is still invoked from the individual controllers.


harimohanraj89 (approved):
Holy moly that's a lot of touchpoints. Looks like we were long overdue for this one 😅

app/controllers/application_controller.rb
```diff
@@ -9,6 +9,15 @@ def switch_locale(&action)
     I18n.with_locale(locale, &action)
   end
 
+  def set_time_zone
```
rsanheim, line 12: We can't put this in ApplicationController -- all our sync controllers in the API inherit from here, and setting our timezone to the reporting timezone could break the timezone conversion that happens when we create records from synced resources.

rsanheim (changes requested):
We talked about this before when dealing w/ time zones for the new matviews...we actually don't want to change the timezone away from UTC for our API requests (UTC is the default there).  At least not with a lot of testing to verify times get created correctly via sync.

app/controllers/application_controller.rb
rsanheim: Ohh, this just extracts the method and keeps the `before_action`s in place.  I think this is okay if we rename the action to `set_reporting_time_zone` or something like that...otherwise this almost seems like it should be the default time zone across all actions, which we definitely don't want.

harimohanraj89: Great call on the naming. Hehe your confusion is a good indicator that `set_time_zone` isn't the right name 😅

rsanheim: > Renamed to `set_reporting_time_zone` in [78bd51b](https://github.com/simpledotorg/simple-server/commit/78bd51bd95b1ec7701935b2761c200e85335415d)

awesome thanks

rsanheim (approved):

PR #2695, opened 2021-07-06
HOTFIX: Restore DHIS2 API calls

timcheadle:
## Because

The API call to DHIS2 was accidentally commented out for debugging and merged with `master`

## This addresses

This restores the API call


harimohanraj89 (approved):
Whoopsie.

PR #2696, opened 2021-07-06
DHIS2 feature flag and scheduling

timcheadle:
**Story card:** [ch3940](https://app.clubhouse.io/simpledotorg/story/3940/add-scheduled-task-to-send-dhis2-data-monthly)

## Because

DHIS2 export should be feature flagged and scheduled for export on the first of each month

## This addresses

- Adds a `dhis2_export` feature flag
- Schedules export at 4am local time on 1st of each month

## Test instructions

1. In the console: `Flipper.enable(:dhis2_export)`
2. On the command line: `whenever` (to confirm the cron job is scheduled).
3. In the console: `Flipper.disable(:dhis2_export)`
4. On the command line: `whenever` (to confirm the cron job is no longer scheduled)


config/schedule.rb
```diff
@@ -60,3 +60,9 @@ def local(time)
 every 30.minutes, roles: [:cron] do
   runner "RegionsIntegrityCheck.sweep"
 end
+
+every 1.month, at: local("4:00 am"), roles: [:cron]  do
```
kpethtel, line 64: I find it strange that we don't specify a day of the month, but we do the same above on the exotel job.

kpethtel (approved):

config/schedule.rb
timcheadle: Yep I confirmed it writes this:

```
0 4 1 * * /bin/bash -l -c 'cd /Users/tim/src/simple-server && RAILS_ENV=production bundle exec rake dhis2:export --silent >> /home/deploy/apps/simple-server/shared/log/cron.log 2>&1'
```

rsanheim (approved):

PR #2700, opened 2021-07-07
3944 imo notification action

kpethtel:

## Because

We need the ability to send notifications through Imo.

## Test instructions

If you want to get wild, you can install Imo on your phone, change a patient record to have your phone number, add the imo password from 1pass, and send yourself a notification through this class. I've done it locally and don't expect a reviewer to do that.


app/services/imo_api_service.rb
```diff
@@ -32,7 +27,35 @@ def invite
...
     response = execute_post(url, body: request_body)
-    process_response(response, url)
+    result = process_response(response, url, "invitation")
+
+    return if result.nil?
+
+    ImoAuthorization.create!(patient: patient, status: result, last_invited_at: Time.current)
+  end
+
+  def send_notification(patient, message)
```
kpethtel, line 37: I'm not currently guarding against the possibility that the patient doesn't have an ImoAuthorization. If they don't, we'll just get an error response from Imo and it will log to sentry, which will let us know that there may be a problem. The same thing is happening in the invite endpoint. I don't guard against the possibility that they've already been invited. If they have, we'll get an error response of some sort.

app/services/imo_api_service.rb
```diff
@@ -42,27 +65,55 @@ def execute_post(url, data)
       .basic_auth(user: IMO_USERNAME, pass: IMO_PASSWORD)
       .post(url, data)
   rescue HTTP::Error => e
-    raise Error.new("Error while calling the IMO API", path: url, exception_message: e)
+    raise Error.new("Error while calling the Imo API", path: url, exception_message: e)
   end
 
-  def process_response(response, url)
+  def process_response(response, url, action)
```
kpethtel, line 71: This is not pretty, but it's how Imo's api is shaped.

app/services/imo_api_service.rb
```diff
@@ -42,27 +65,55 @@ def execute_post(url, data)
...
     case response.status
-    when 200 then :invited
+    when 200
+      body = JSON.parse(response.body)
+      body_status = body.dig("response", "status")
+      return :invited if body_status == "success" && action == "invitation"
+      return :subscribed if body_status == "success" && action == "notification"
```
kpethtel, line 77: Imo just added invitation callbacks to their API doc, so this will probably change. Without that callback, the only way to know if they've subscribed is to send a message and see if it succeeds.

app/services/imo_api_service.rb
```diff
@@ -2,6 +2,7 @@ class ImoApiService
   IMO_USERNAME = ENV["IMO_USERNAME"]
   IMO_PASSWORD = ENV["IMO_PASSWORD"]
   BASE_URL = "https://sgp.imo.im/api/simple/".freeze
+  USER_FACING_URL = "https://www.nhf.org.bd".freeze
```
rsanheim, line 5: What is this URL for?  

It doesn't resolve with the `www`, though without the `www` it works.  We should probably remove that so its a valid address.

```
[~/src/simpledotorg/simple-server (rename-reporting-matviews)]> nslookup nhf.org.bd
Server:		172.16.0.1
Address:	172.16.0.1#53

Non-authoritative answer:
Name:	nhf.org.bd
Address: 203.191.33.185

[~/src/simpledotorg/simple-server (rename-reporting-matviews)]> nslookup www.nhf.org.bd
Server:		172.16.0.1
Address:	172.16.0.1#53

** server can't find www.nhf.org.bd: REFUSED

[~/src/simpledotorg/simple-server (rename-reporting-matviews)]>
```

app/services/imo_api_service.rb
```diff
@@ -13,13 +14,7 @@ def initialize(message, path: nil, response: nil, exception_message: nil)
...
   end
 
-  attr_reader :patient
-
-  def initialize(patient)
-    @patient = patient
-  end
-
-  def invite
+  def invite(patient)
```
rsanheim, line 17: How about `send_invitation` to be consistent with the other method here?

app/services/imo_api_service.rb
```diff
@@ -42,27 +65,55 @@ def execute_post(url, data)
...
   end
 
-  def process_response(response, url)
+  def process_response(response, url, action)
     case response.status
-    when 200 then :invited
+    when 200
+      body = JSON.parse(response.body)
+      body_status = body.dig("response", "status")
+      return :invited if body_status == "success" && action == "invitation"
```
rsanheim, line 76: The `action` is just what we, the callers, are attempting to do with IMO - right?  I can see how that is useful for the instrumentation below, but we don't really need it for the response handling.  

I would suggest returning `true` here for success, and let the calling method set the correct state based on that.  That way we won't need these two cases for the success case.  Just something like:

```suggestion
      return true if status == "success"
```

Then callers can do either

`ImoAuthorization.create(status: :invited)` or `imo_authorization.update(status: :subscribed)`  .  Or better yet, use the enum methods where possible...for example, `imo_authorization.subscribed!`

app/services/imo_api_service.rb
```diff
@@ -42,27 +65,55 @@ def execute_post(url, data)
...
+      return :subscribed if body_status == "success" && action == "notification"
+
+      case body.dig("response", "error_code")
+      when "not_subscribed"
```
rsanheim, line 80: Lets add a comment here explaining that we don't know if a patient is subscribed based on any proper API - we have to send a message and see if it succeeds.

spec/jobs/imo/invite_patient_spec.rb
```diff
@@ -24,43 +22,14 @@
...
-        imo_auth = ImoAuthorization.last
-        expect(imo_auth.status).to eq("invited")
-        expect(imo_auth.patient_id).to eq(patient.id)
-      end
-
-      it "creates an ImoAuthorization with 'no_imo_account' status when invited user has no Imo account" do
+      it "invites the patient to Imo" do
         patient = create(:patient)
-        allow_any_instance_of(ImoApiService).to receive(:invite).and_return(:no_imo_account)
+        imo_service = double
```
rsanheim, line 27: This can be an [instance_double](https://relishapp.com/rspec/rspec-mocks/v/3-10/docs/verifying-doubles/using-an-instance-double), which is safer to use for these sorts of mock based tests:

`imo_double = instance_double(ImoApiService)`

rsanheim (commented):
Some feedback on the control flow here, and some lower level things.

app/services/imo_api_service.rb
rsanheim: If someone has already been invited, will we continually attempt to re-invite them based on some sort of schedule? Should we be capturing invitation attempts on our end so we don't keep retrying patients who are not going to opt-in?

kpethtel: We are capturing that information in ImoAuthorization, which has a `last_invited_at` attribute.

app/services/imo_api_service.rb
kpethtel: I meant to rename `USER_FACING_URL` to something like, `REDIRECT_URL`. It's the url that the user is navigated to if they click on a certain part of the Imo invitation card. It's a required field and has to be whitelisted by Imo. The www version resolves for me. 
```
nslookup www.nhf.org.bd
Server:		192.168.0.1
Address:	192.168.0.1#53

Non-authoritative answer:
Name:	www.nhf.org.bd
Address: 203.191.33.185
```
Also works in browser. I wonder if it was just a blip.

app/services/imo_api_service.rb
kpethtel: The action is also being used by Statsd. I could remove the statsd logging or make it non-specific to the action. But I think returning true makes the logic more complex. The reason I didn't use the enum methods is because there are multiple return values from this function that can all be used to update the enum value. So if the user has no account, we update the ImoAuthorization with `:no_imo_account`. If you still think this should be changed, maybe we should zoom for a couple of minutes to avoid any confusion.

app/services/imo_api_service.rb
kpethtel: Will do.

app/services/imo_api_service.rb
kpethtel: Will do.

spec/jobs/imo/invite_patient_spec.rb
kpethtel: I tried:
```
imo_double = instance_double(ImoApiService)
expect(imo_service).to receive(:send_invitation).with(patient)
```
And several variations on that theme. Didn't work for me. Not sure why.

app/services/imo_api_service.rb
kpethtel: Actually, after working ahead a little, I've decided I will change this. The handling will have to change after the callbacks are added anyhow. If you don't mind turning a blind eye to this for the moment, I'll change it in the next PR, which is branched off of this.

app/services/imo_api_service.rb
rsanheim: Yup I bet it was some caching thing, it works for me tnow. 👍

spec/jobs/imo/invite_patient_spec.rb
rsanheim: oh, you would still have to stub it into place for it to work....something like

```
imo_service = instance_double(ImoApiService)
allow(ImoApiService).to receive(:new).and_return(imo_service)
expect(imo_service).to receive(:something).and_return(:something)
```

app/services/imo_api_service.rb
```diff
@@ -2,6 +2,8 @@ class ImoApiService
   IMO_USERNAME = ENV["IMO_USERNAME"]
   IMO_PASSWORD = ENV["IMO_PASSWORD"]
   BASE_URL = "https://sgp.imo.im/api/simple/".freeze
+  # this is where the patient is redirected to when they click on the invitation card details section
```
rsanheim, line 5: 👍

app/services/imo_api_service.rb
rsanheim: ah okay, great.

app/services/imo_api_service.rb
rsanheim: Okay, sounds good, I can follow along there.

rsanheim (approved):
🚢  lets zoom about the overall flow tomorrow or friday once you have more of those pieces in place in the follow up PR.  I think it would be helpful for me to have more context on the big picture here.

PR #2704, opened 2021-07-08
Fix flakey spec failing due to Groupdate and timezones

rsanheim:


spec/queries/bp_measures_query_spec.rb
```diff
@@ -27,9 +27,11 @@ def ist_zone
     facility = create(:facility)
     end_of_jan = ist_zone.parse("January 31st 23:59:59 IST")
     beg_of_feb = ist_zone.parse("February 1st 00:00:00 IST")
+
     create(:blood_pressure, facility: facility, recorded_at: end_of_jan, user: user_1)
     create(:blood_pressure, facility: facility, recorded_at: beg_of_feb, user: user_2)
-    Time.use_zone(ist_zone) do
+
+    with_reporting_time_zone do
```
rsanheim, line 34: This is the fix - we weren't using Groupdate for the query under test, so things would get sorted as UTC instead of IST.

app/models/reports/matview.rb
```diff
@@ -2,7 +2,9 @@ module Reports
   class Matview < ActiveRecord::Base
     def self.refresh
       ActiveRecord::Base.transaction do
-        ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{Period::REPORTING_TIME_ZONE}'")
+        tz = Rails.application.config.country[:time_zone]
+        ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{tz}'")
+        # ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{Period::REPORTING_TIME_ZONE}'")
```
kpethtel, line 7: Did you mean to leave this as a comment?

kpethtel (approved):
One question, but otherwise, it looks... good? Timezones... what can you do.

PR #2705, opened 2021-07-08
4084: render error page when user is not found by reset token

kpethtel:
**Story card:** [ch4084](https://app.clubhouse.io/simpledotorg/story/4084/error-when-login-token-is-not-found)

## Because
This fixes a sentry error we've recently seen: https://sentry.io/organizations/simple-ethiopia/issues/2499060149/?project=5496145&referrer=slack

Without digging too deeply into how Devise works, my best guess is that the token is very old and has been cleared from devise's cache. That's pure speculation. Regardless, showing the error page on whatever is going on here seems appropriate. If there's a deeper problem, hopefully it will be reported and we can fix it with more information.

### DIFF
--- a/app/controllers/email_authentications/passwords_controller.rb
+++ b/app/controllers/email_authentications/passwords_controller.rb
@@ -1,7 +1,7 @@
 class EmailAuthentications::PasswordsController < Devise::PasswordsController
   def edit
     users_email_authentication = resource_class.with_reset_password_token(params["reset_password_token"])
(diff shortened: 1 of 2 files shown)


harimohanraj89 (approved):
This seems great to me. If the token's old enough that it's cleared out, the error still makes sense. If the token's not found because the user was link-fudging, it's totally OK for them to see this error message anyway 😅

PR #2706, opened 2021-07-08
3944 add imo communication type

kpethtel:

## Because

We need to be able to send notifications via Imo.

## This addresses

This adds the imo communication type and adjusts the notifications worker to send imo messages. This does not changelib/tasks/scripts/message_patients.rb to use Imo. It's unclear if that's something we want. If it is, we should open a separate ticket because it's lower priority than the remaining work in the epic.


app/jobs/appointment_notification/worker.rb
```diff
@@ -38,14 +38,15 @@ def perform(notification_id)
   private
 
   def send_message(notification, communication_type)
-    notification_service = if notification.experiment&.experiment_type == "medication_reminder" && medication_reminder_sms_sender
+    notification_service = if communication_type == "imo"
+      ImoApiService.new
+    # given that we expect higher volume in the future, should this be default behavior?
+    elsif notification.experiment&.experiment_type == "medication_reminder" && medication_reminder_sms_sender
```
kpethtel, line 44: given that we expect higher volume in the future, should this be default behavior?

app/jobs/appointment_notification/worker.rb
```diff
@@ -54,29 +54,28 @@ def send_message(notification, communication_type)
...
-        metrics.increment("sent.sms")
-      end
+      )
+    when "imo"
+      notification_service.send_notification(notification.patient, notification.localized_message)
     else
       raise UnknownCommunicationType, "#{self.class.name} is not configured to handle communication type #{communication_type}"
     end
 
+    metrics.increment("sent.#{communication_type}")
```
harimohanraj89, line 78: Nice extraction 👍

app/models/communication.rb
```diff
@@ -60,6 +61,20 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
     end
   end
 
+  def self.create_with_imo_details!(appointment:, notification:, result:)
+    transaction do
```
harimohanraj89, line 65: Looks like the transaction in the Twilio method was because we were creating two records. Here we're only making one, so we can remove the transaction.

PS. Do we need an equivalent of `TwilioSmsDeliveryDetail`? I'm not saying we do, I'm genuinely unaware. What's that thing for?

app/models/communication.rb
```diff
@@ -60,6 +61,20 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
...
+      # create detailable here once detailable model exists
+      communication = Communication.create!(communication_type: "imo",
```
harimohanraj89, line 67: We can remove `Communication.`, redundant.

app/models/communication.rb
```diff
@@ -60,6 +61,20 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
...
+                                            detailable: nil,
+                                            appointment: appointment,
+                                            notification: notification,
+                                            device_created_at: DateTime.current,
+                                            device_updated_at: DateTime.current)
```
harimohanraj89, line 72: Extract `DateTime.current` into a `now` var, instead of invoking twice? If we're trying to emulate a device creating a communication, then we should ensure the timestamps match exactly.

app/models/communication.rb
```diff
@@ -60,6 +61,20 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
     end
   end
 
+  def self.create_with_imo_details!(appointment:, notification:, result:)
```
harimohanraj89, line 64: What are we doing with the `result`?

app/models/communication.rb
```diff
@@ -60,6 +61,20 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
...
+    transaction do
+      # create detailable here once detailable model exists
+      communication = Communication.create!(communication_type: "imo",
+                                            detailable: nil,
+                                            appointment: appointment,
+                                            notification: notification,
+                                            device_created_at: DateTime.current,
+                                            device_updated_at: DateTime.current)
+      logger.info class: self.class.name, msg: "create_with_imo_details", communication: communication.id,
+                  communication_type: "imo", appointment_id: appointment.id
```
harimohanraj89, line 74: Shall we add the notification ID here for completeness? We're accepting one in the method signature.

app/models/communication.rb
```diff
@@ -60,6 +61,20 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
...
+  def self.create_with_imo_details!(appointment:, notification:, result:)
+    transaction do
+      # create detailable here once detailable model exists
+      communication = Communication.create!(communication_type: "imo",
+                                            detailable: nil,
+                                            appointment: appointment,
+                                            notification: notification,
+                                            device_created_at: DateTime.current,
+                                            device_updated_at: DateTime.current)
+      logger.info class: self.class.name, msg: "create_with_imo_details", communication: communication.id,
```
harimohanraj89, line 73: Consider renaming the key `communication_id` for consistency.

app/models/notification.rb
```diff
@@ -65,7 +65,10 @@ def previously_communicated_by?(method)
   end
 
   def preferred_communication_method
-    Flipper.enabled?(:whatsapp_appointment_reminders) ? "whatsapp" : nil
+    return "whatsapp" if Flipper.enabled?(:whatsapp_appointment_reminders)
+    imo_status = patient.imo_authorization&.status
+    return "imo" if Flipper.enabled?(:imo_messaging) && imo_status.in?(["invited", "subscribed"])
+    nil
```
harimohanraj89, line 71: Food for thought - I see that this code assumes that both Whatsapp and IMO are never enabled together. If they were, this codepath would ignore IMO altogether. Should we allow `Notification` to be more flexible? Create a better communication method priority that can cascade from whatsapp to IMO to SMS, and detect which methods have already been attempted when choosing the next one?

I don't think this PR's the right place to _do_ that work, but wanted to run the idea by you here, as we now have _three_ available methods to contact patients :smile:

app/services/imo_api_service.rb
```diff
@@ -30,9 +30,10 @@ def send_invitation(patient)
     response = execute_post(url, body: request_body)
     result = process_response(response, url, "invitation")
 
-    return if result.nil?
+    return if result == :error
 
-    ImoAuthorization.create!(patient: patient, status: result, last_invited_at: Time.current)
+    status = result == :success ? :invited : result
```
harimohanraj89, line 35: Fun - does the invitation API send us a `success` response, that we're translating to something more semantic here?

harimohanraj89: Out of curiosity, what else can `result` be? Anything other than `success` and `error`?

app/services/imo_api_service.rb
```diff
@@ -52,11 +53,13 @@ def send_notification(patient, message)
...
-
-    unless patient.imo_authorization.status == result.to_s
-      patient.imo_authorization.update!(status: result)
+    # until we implement the invitation callback, the only way for us to know if the user
+    # has accepted our invitation is to send a notication to see if it succeeds
+    status = result == :success ? :subscribed : result
+    unless patient.imo_authorization.status == status.to_s || status == :error
+      patient.imo_authorization.update!(status: status)
     end
+    status
```
harimohanraj89, line 62: It's a bit strange that we're returning the _invitation_ status (eg. "subscribed") here. Can we keep that as an internal detail, and instead return the notification's original result?

app/services/imo_api_service.rb
```diff
@@ -52,11 +53,13 @@ def send_notification(patient, message)
...
 
-    return if result.nil?
-
-    unless patient.imo_authorization.status == result.to_s
-      patient.imo_authorization.update!(status: result)
+    # until we implement the invitation callback, the only way for us to know if the user
+    # has accepted our invitation is to send a notication to see if it succeeds
+    status = result == :success ? :subscribed : result
+    unless patient.imo_authorization.status == status.to_s || status == :error
+      patient.imo_authorization.update!(status: status)
```
harimohanraj89, line 60: It might be a good idea to extract this chunk into a semantic method, especially since this is a temporary "btw has this user accepted our invite?" hook. Perhaps something like `update_imo_subscription_status(result)`?

spec/models/notification_spec.rb
```diff
@@ -63,7 +63,33 @@
       end
     end
 
-    context "when WhatsApp flag is off" do
+    context "when imo flag is on" do
+      before { Flipper.enable(:imo_messaging) }
+
+      it "returns sms if patient can't receive Imo notifications" do
+        expect(notification.next_communication_type).to eq("sms")
```
harimohanraj89, line 70: Let's make an explicit ImoAuthorization with `status: :error`?

spec/models/communication_spec.rb
```diff
@@ -85,6 +85,19 @@
...
+  describe ".create_with_imo_details" do
+    it "creates a communication" do
+      patient = create(:patient)
+      appt = create(:appointment, patient: patient)
+      notification = create(:notification, subject: appt, patient: patient)
+      expect {
+        Communication.create_with_imo_details!(appointment: notification.subject,
+                                               notification: notification,
+                                               result: nil)
+      }.to change { Communication.count }.by(1)
```
harimohanraj89, line 97: Can we add more specific assertions to this spec around the details of the `Communication` record we just created?

harimohanraj89 (changes requested):
Feedback below!

app/models/communication.rb
kpethtel: There will be a `TwilioSmsDeliveryDetail`. That's why it's still wrapped in a transaction. It's in an upcoming ticket.

app/models/communication.rb
kpethtel: Fair point. I'll correct it in the twilio variation too.

app/models/communication.rb
kpethtel: It'll go into `TwilioSmsDeliveryDetail` in one of the upcoming tickets. I'm starting to think I should've made that a bit clearer.

harimohanraj89: Or we can add things like `result` into that subsequent PR only, rather than into this one. That woudl eliminate the need to explain the sequence of events too deeply everywhere.

app/models/communication.rb
kpethtel: Good catch. This was copy-pasted from the twilio method above. I'll correct it there too.

app/models/communication.rb
kpethtel: Yeah, this is another case of me blindly copy-pasting from the above method. I can correct it in both locations.

app/services/imo_api_service.rb
kpethtel: Sorry, this is one of those things that's just captured in my head. When I started work on this, there was no invitation callback, so the only way to know if the user was subscribed was to send a message and see if it succeeds. This changed last week when Imo added an invitation callback. So this logic will change in the near future. We will only send notifications to patients who have subscribed, so this logic will be removed. I left this here because I haven't been able to test the invitation callback yet to confirm that it works. I was not able to test that due to technical constraints on Imo's side. They've just unblocked me (I believe), so I will test it soon. Regardless, I suggest we leave this here for now and I'll change it while I'm adding the invitation callback logic.

app/services/imo_api_service.rb
kpethtel: Sorry, this is one of those things that's just captured in my head. When I started work on this, there was no invitation callback, so the only way to know if the user was subscribed was to send a message and see if it succeeds. This changed last week when Imo added an invitation callback. So this logic will change in the near future. We will only send notifications to patients who have subscribed, so this logic will be removed. I left this here because I haven't been able to test the invitation callback yet to confirm that it works. I was not able to test that due to technical constraints on Imo's side. They've just unblocked me (I believe), so I will test it soon. Regardless, I suggest we leave this here for now and I'll change it while I'm adding the invitation callback logic.

spec/models/notification_spec.rb
kpethtel: That might be wise. Lemme think about that a bit.

app/models/notification.rb
kpethtel: Yeah, that's a fair point. I assumed the two would never be on together because that is the current plan and attempting to enforce that shouldn't be handled here. The only thing we have to do to make your suggestion work is determine which communication method is higher priority.

app/services/imo_api_service.rb
kpethtel: It can also be `not_subscribed` which could happen if the patient subscribed then unsubscribed or `no_imo_account` which could happen if the patient deleted their imo.

kpethtel: There may be others in the future as we get to know this API better.

app/models/notification.rb
harimohanraj89: Yup. Definitely not the right PR to address those concerns, but cool follow-up work/improvements for us to document before we release this work.

app/services/imo_api_service.rb
harimohanraj89: Ah understood. Can we rename to `invitation_status` for clarity in that case? Minimize confusion about whether or not this is the status of the notification we just sent (sent, delivered, failed, etc).

kpethtel: I ended up removing this behavior. Adding the invitation callback is my next ticket anyhow.

spec/models/notification_spec.rb
kpethtel: I did add this. It only applies to the invitation job.

app/services/imo_api_service.rb
```diff
@@ -52,11 +51,10 @@ def send_notification(patient, message)
     response = execute_post(url, body: request_body)
     result = process_response(response, url, "notification")
 
-    return if result.nil?
-
-    unless patient.imo_authorization.status == result.to_s
+    if patient.imo_authorization.status != result.to_s && result.in?([:not_subscribed, :no_imo_account])
```
kpethtel, line 54: I'm no longer allowing this to update to "subscribed". My next ticket is the invitation callback, so it won't be necessary.

harimohanraj89 (approved):
This is looking slick. Some more follow-up comments from me but approval to unblock you. Feel free to get another round of synchronous review if you think it's required.

I want to make sure I understand the flow and branching correctly.

* `Notification` records are spawned for messages we wanna send patients
* The notification has that cascading preferred comms method ,and we queue up a worker to act on that notification
* Imo is now a potential comms method, and we select the Imo API service in those cases to send out the message
* We attempt to send the message. If the user's not subscribed, we make a note of that in the patient's Imo Auth record

Last thing I wanted to clarify, I know we have tooling to invite unsubscribed users periodically. Is that tooling being used yet? If so, can you point me to it? Or are we yet to hook it up to a cron task or something similar?

app/services/imo_api_service.rb
kpethtel: Correct. It also means that we won't even attempt to send a notification via Imo unless we know they've already subscribed.

spec/models/notification_spec.rb
kpethtel: I misread you. That's why I said, "This seems like an odd place for the comment." You're right and I'm surprised I didn't add a test for it.

kpethtel: Ok, I had to look at this again. So I did test it, just not as clearly as I could have. Got it. Sorry for the confusion. I read you as saying, "We should have an error state in ImoAuth." And I ended up adding an error state to imo auth. So now if we receive an unknown Imo error, we log it in sentry and add the ImoAuth with an error state to prevent continuous retries.

app/models/communication.rb
```diff
@@ -45,21 +46,35 @@ def self.latest_by_type(communication_type)
...
+  def self.create_with_imo_details!(appointment:, notification:)
+    now = DateTime.current
+    communication = create!(communication_type: "imo",
+                            detailable: nil,
+                            appointment: appointment,
+                            notification: notification,
+                            device_created_at: now,
+                            device_updated_at: now)
+    logger.info class: self.class.name, msg: "create_with_imo_details", communication_id: communication.id,
+                communication_type: "imo", appointment_id: appointment.id, notification_id: notification&.id
```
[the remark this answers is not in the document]
kpethtel: Actually, notification is that one that could be nil, not notification. This was an error on my part.

PR #2707, opened 2021-07-09
Drug stock add block rows

vkrmis:
**Story card:** [ch3322](https://app.clubhouse.io/simpledotorg/story/3322/add-a-block-level-drug-stock-table)

## Because

We want to display block level totals on the drug stock and consumption reports.

## This addresses

Adds block level numbers in the query object and the view.
image

This also adds a `Facilities subtotal` which is the same as the current `All` total. We will be adding district warehouse numbers to the `All` row in a follow up.


app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -59,9 +59,17 @@ def create_drug_report
       .includes(facility_group: :protocol_drugs)
       .where(protocol_drugs: {stock_tracked: true})
 
+    @blocks = Region
+      .block_regions
+      .joins("INNER JOIN regions facility_region ON regions.path @> facility_region.path")
+      .where(facility_region: {source_id: @facilities})
+      .distinct("regions.id")
```
harimohanraj89, line 66: I've seen some elegant ways to fetch blocks for a list of regions somewhere else in our codebase I think. Wanna plop this in #server for some quick ideas?

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -59,9 +59,17 @@ def create_drug_report
...
+
     @for_end_of_month_display = @for_end_of_month.strftime("%b-%Y")
-    render && return if @facilities.empty?
-    @query = DrugStocksQuery.new(facilities: @facilities, for_end_of_month: @for_end_of_month)
+    @query = DrugStocksQuery.new(facility_group: @selected_facility_group,
+                                 facilities: @facilities,
+                                 blocks: @blocks,
```
harimohanraj89, line 71: Why do we need to pass blocks in here? Can't the `DrugStocksQuery` figure this out? If it's only to expose it to the view, I'd rather do something like `@blocks = @query.blocks`. I think it's better to avoid room for inconsistent input.

app/queries/drug_stocks_query.rb
```diff
@@ -3,16 +3,17 @@ class DrugStocksQuery
 
   CACHE_VERSION = 1
 
-  def initialize(facilities:, for_end_of_month:)
-    @facilities = facilities
+  def initialize(facility_group:, facilities:, for_end_of_month:, blocks: Region.none)
+    @facilities = Facility.where(id: facilities)
     @for_end_of_month = for_end_of_month
-    # assuming that all facilities on the page have the same protocol
```
harimohanraj89, line 9: Instead of assuming, can we raise an error? That'll make sure our current codepaths are solid, and warn devs in the future if they use this thing differently.

harimohanraj89 (changes requested):
Some tiny feedback, but I'm starting to think about the overall approach here, now that we've got blocks in here too. Specifically, the things like `with_block_id`, and the almost-same repetition between methods is making me think that the `DrugStockQuery` is trying to do too much. Here are my thoughts for alternative ways to tackle this that might result in much more concise code that's easier to maintain going forward:

* DrugStockQuery groups by either facility, block, or district. We use three query objects to assemble the results we need for the page
* DrugStockQuery groups only by facility, but includes district and block information against each blob. We group and sum in memory (this is basically recreating the grouping and summing we do in SQL right now)

What do you think? I'm down to pull the trigger right away, or in a follow-up PR.

app/controllers/my_facilities/drug_stocks_controller.rb
prabhanshuguptagit: Since `@facilities` is a set of filtered facilities, we can't fetch blocks top down and we need to go from the facilities, bottom up. The elegant ways don't work for this. We tried a few different queries, out of which the current one is most performant:
```
Region.where(id: @facilities.with_block_region_id.pluck(:block_region_id).uniq)
```
```
Region.where(id:
 Region
   .where(source_id: @facilities.pluck(:id))
   .joins("INNER JOIN regions block_region ON block_region.path @> regions.path AND block_region.region_type = 'block'")
   .select("block_region.*")
   .distinct
)
```

harimohanraj89: Cooooolbeans, you're way ahead of me 🚀

app/queries/drug_stocks_query.rb
prabhanshuguptagit: We have removed this assumption now, we are explicitly passing in a facility group which can only have one protocol.

harimohanraj89: Yup makes sense. This comment is only relevant if we undo the FG and block arguments.

app/queries/drug_stocks_query.rb
```diff
@@ -3,14 +3,15 @@ class DrugStocksQuery
 
   CACHE_VERSION = 1
 
-  def initialize(facility_group:, facilities:, for_end_of_month:, blocks: Region.none)
+  def initialize(facilities:, for_end_of_month:, include_block_report: true)
```
harimohanraj89, line 6: This signature looks great.

app/queries/drug_stocks_query.rb
```diff
@@ -25,9 +26,9 @@ def drug_stocks_report
        patient_count_by_facility_id: patient_count_by_facility_id,
        patient_days_by_facility_id: patient_days_by_facility_id,
        drugs_in_stock_by_facility_id: drugs_in_stock_by_facility_id,
-       patient_count_by_block_id: patient_count_by_block_id,
-       patient_days_by_block_id: patient_days_by_block_id,
-       drugs_in_stock_by_block_id: drugs_in_stock_by_block_id,
+       patient_count_by_block_id: @include_block_report ? patient_count_by_block_id : {},
+       patient_days_by_block_id: @include_block_report ? patient_days_by_block_id : {},
+       drugs_in_stock_by_block_id: @include_block_report ? drugs_in_stock_by_block_id : {},
```
harimohanraj89, line 31: Any reason to include these keys if `include_block_report` is false? Could we set to `nil` and `compact` at the end?

app/queries/drug_stocks_query.rb
```diff
@@ -218,10 +222,28 @@ def drug_consumption_cache_key
...
       @state,
       CACHE_VERSION
     ].join("/")
   end
+
+  private
+
+  def set_facility_group
+    facility_group_ids = @facilities.pluck(:facility_group_id).uniq
+    throw "All facilities should belong to the same facility group." if facility_group_ids.count > 1
```
harimohanraj89, line 237: We should use `raise`. `throw` actually does some subtly different things in Ruby that we usually don't want.

harimohanraj89 (approved):
Looking slick. Couple of last thoughts :shipit:

app/queries/drug_stocks_query.rb
prabhanshuguptagit: TIL hashes can be compacted 🙌🏼

PR #2710, opened 2021-07-09
3762 imo invitations translations

kpethtel:
**Story card:** [ch3762](https://app.clubhouse.io/simpledotorg/story/3762/add-translations-for-invitations)

## Because

We would like Imo invitations to be in localized to the patient.


spec/services/imo_api_service_spec.rb
```diff
@@ -71,6 +74,15 @@
...
         }.to raise_error(ImoApiService::Error)
       end
+
+      it "raises an error when the patient's locale is not supported" do
+        stub_request(:post, request_url).with(headers: request_headers).to_return(status: 200, body: success_body)
+        allow(patient).to receive(:locale).and_return("en")
+
+        expect {
+          service.send_invitation(patient)
+        }.to raise_error
```
rsanheim, line 84: small thing here, but can you specify the error type so this is more specific?

rsanheim (approved):
one minor bit of feedback, otherwise looks good

PR #2713, opened 2021-07-09
Update to dogstatsd-ruby 5.2, resetting the client after fork

rsanheim:
This updates to the latest version of [dogstatsd-ruby](https://github.com/DataDog/dogstatsd-ruby), which changes how metrics are submitted and requires a bit more care around the client.

Quoting from the Dogstatd docs:

> Version v5.x of dogstatsd-ruby is using a companion thread for preemptive flushing, it brings better performances for application having a high-throughput of statsd metrics, but it comes with new pitfalls:

> * Applications forking after having created the dogstatsd instance: forking a process can't duplicate the existing threads, meaning that one of the processes won't have a companion thread to flush the metrics and will lead to missing metrics.
> * Applications creating a lot of different instances of the client without closing them: it is important to close the instance to free the thread and the socket it is using or it will lead to thread leaks.
> If you are using Sidekiq, please make sure to close the client instances that are instantiated. See this example on using DogStatsD-ruby v5.x with Sidekiq.

> If you are using Puma or Unicorn, please make sure to create the instance of DogStatsD in the workers, not in the main process before it forks to create its workers. See this comment for more details.

* https://github.com/DataDog/dogstatsd-ruby#v5x-common-pitfalls
* https://github.com/DataDog/dogstatsd-ruby/issues/179
* https://www.phusionpassenger.com/library/indepth/ruby/spawn_methods/#the-most-straightforward-and-traditional-way-direct-spawning

**Story card:** [ch4102](https://app.clubhouse.io/simpledotorg/story/4102/upgrade-dogstatsd-ruby-to-latest-version-s)


kpethtel (approved):
Configuration is hard to review, but this looks fine. Are there any concerns that this will cause issues?

harimohanraj89 (approved):
This looks good, and seems to cover all the pitfalls mentioned on the Statsd repo. If I understand correctly, the singleton `Statsd` class can spawn one instance per Passenger process.

Best way to verify this is to test in SBX. Have at it 🚀  Once it looks good there, #shipit

PR #2715, opened 2021-07-12
Unique indexes for reporting matviews

rsanheim:
This will allow us to refresh concurrently, which means concurrent
selects will still work.  Otherwise the entire mat view is locked out.

docs: https://www.postgresql.org/docs/9.4/sql-refreshmaterializedview.html


db/migrate/20210712223435_add_unique_indexes_to_matviews.rb
```diff
@@ -0,0 +1,6 @@
+class AddUniqueIndexesToMatviews < ActiveRecord::Migration[5.2]
+  def change
+    add_index :reporting_patient_blood_pressures, [:patient_id, :month_date], unique: true, name: "patient_blood_pressures_patient_id_month_date"
+    add_index :reporting_patient_visits, [:patient_id, :month_date], unique: true, name: "patient_visits_patient_id_month_date"
```
ssrihari, line 4: We should change the order of the multicolumn index to be `month_date` first, and `patient_id` second. This way, this index would be used in queries that filter by `month_date`. See the docs for more details on effect of leading columns in multicolumn indexes: https://www.postgresql.org/docs/10/indexes-multicolumn.html

ssrihari (dismissed):
Shall we also include the `concurrent: true` in the calls to `refresh` with this PR? That will ensure that this is tested as well.

db/migrate/20210712223435_add_unique_indexes_to_matviews.rb
ssrihari: We should also change the index on `patient_states` to be the same way.

rsanheim:
> Shall we also include the `concurrent: true` in the calls to `refresh` with this PR? That will ensure that this is tested as well.

I'm going to do that as a follow up PR and will test it locally - it helps us keep any PRs changing the schema focused and easier to review.

db/migrate/20210712223435_add_unique_indexes_to_matviews.rb
rsanheim: change ✨

kpethtel (approved):
👍

PR #2716, opened 2021-07-13
Fix date selection in drug stock webview

prabhanshuguptagit:
**Story card:** -

## Because

A regression got introduced in #2668 where the date format submitted from the drug stock webview doesn't match the format the controller was expecting.

## This addresses

Fixes the date format in the webview.


vkrmis (approved):
lgtm

PR #2717, opened 2021-07-13
Fix drug stock report in webview

prabhanshuguptagit:
**Story card:** -

## Because

We introduced a regression in #2668 that broke the webview drug stock report redirect.

## This addresses

This fixes the report lookup and the redirect.


vkrmis (approved):
lgtm

PR #2719, opened 2021-07-13
Drug stock redistribution

prabhanshuguptagit:
**Story card:** [ch4029](https://app.clubhouse.io/simpledotorg/story/4029/add-a-column-for-redistributed-drugs-to-drug-stock?vc_group_by=day&ct_workflow=all&cf_workflow=500000005)

## Because

We want to account for drug distribution in consumption calculation. 

`Drug consumption = Opening balance + Stock received - Closing balance - Redistributed stock`

## This addresses

- Adds the redistributed stock column to drug stock model
- Adds the column to various forms
- Uses the column in consumption calculation

image
image


app/assets/stylesheets/user_analytics.scss
```diff
@@ -746,4 +746,61 @@ select.form-control, input.form-control {
     box-shadow: 0 2px 1px rgba(0,0,0,.16);
     box-sizing: border-box;
   }
-}
\ No newline at end of file
+}
+
+.switch {
```
harimohanraj89, line 751: This is the "did you redistribute?" switch? Could we add a screenshot to PR desc?

app/controllers/webview/drug_stocks_controller.rb
```diff
@@ -21,7 +21,7 @@ def create
     DrugStocksCreator.call(user: current_user,
                            facility: @current_facility,
                            for_end_of_month: @for_end_of_month,
-                           drug_stocks_params: safe_params[:drug_stocks])
+                           drug_stocks_params: drug_stocks_params)
```
harimohanraj89, line 24: We went from passing `safe_params[:drug_stocks]` to `safe_params[:drug_stocks].values`, but I don't see any corresponding change in the behavior of the DrugStocksCreator. Is this intentional?

app/models/reports/drug_stock_calculation.rb
```diff
@@ -88,21 +89,23 @@ def patient_days_calculation
...
+    def consumption_calculation(opening_balance, received, redistributed, closing_balance)
       return {consumed: nil} if [opening_balance, received, closing_balance].all?(&:nil?)
 
       {
         opening_balance: opening_balance,
         received: received,
+        redistributed: redistributed,
         closing_balance: closing_balance,
-        consumed: opening_balance + received - closing_balance
+        consumed: opening_balance + received - (redistributed || 0) - closing_balance
```
harimohanraj89, line 100: This is the heart of the change here, right? I'm also a fan of the other method refactors going on in this class.

app/views/my_facilities/drug_stocks/new.html.erb
```diff
@@ -31,10 +31,13 @@
             <%= drug_stock_form.hidden_field :protocol_drug_id, value: protocol_drug.id %>
             <div class="form-row">
               <div class="col">
-                <%= drug_stock_form.number_field :received, class: :received, value: @drug_stocks[protocol_drug.id].try(&:received), hide_label: true, help: "Received during the month" %>
```
harimohanraj89, line 34: Were we using the `class: :received` for any styling?

spec/exporters/drug_consumption_report_exporter_spec.rb
```diff
@@ -16,11 +16,11 @@
...
-      "329528" => {in_stock: 10000, received: 2000},
-      "329526" => {in_stock: 20000, received: 2000},
-      "316764" => {in_stock: 10000, received: 2000},
-      "316765" => {in_stock: 20000, received: 2000},
-      "979467" => {in_stock: 10000, received: 2000}
+      "329528" => {in_stock: 10000, received: 2000, redistributed: 0},
+      "329526" => {in_stock: 20000, received: 2000, redistributed: 0},
+      "316764" => {in_stock: 10000, received: 2000, redistributed: 0},
+      "316765" => {in_stock: 20000, received: 2000, redistributed: 0},
+      "979467" => {in_stock: 10000, received: 2000, redistributed: 0}
```
harimohanraj89, line 23: Can this thing handle `redistributed: nil`? Does it need to?

spec/models/reports/drug_stock_calculation_spec.rb
```diff
@@ -223,4 +227,21 @@
       expect(instance.patient_days_coefficients("state_that_does_not_exist")).to eq nil
     end
   end
+
+  describe "#consumption" do
+    it "uses the redistribution numbers for consumption calculation" do
```
harimohanraj89, line 232: This might be too specific to the current PR. How about something like `calculates consumption using stock in hand, received, and redistributed`?

harimohanraj89 (changes requested):

app/models/reports/drug_stock_calculation.rb
prabhanshuguptagit: Yes 👍🏼

app/views/my_facilities/drug_stocks/new.html.erb
prabhanshuguptagit: No, this was unused.

spec/exporters/drug_consumption_report_exporter_spec.rb
prabhanshuguptagit: The factory sets `redistributed` to `rand(1..1000)`, so we're explicitly setting this to 0. We could consider removing the randomness from the factory, which I think is only important for seed data.

harimohanraj89: I'm less worried about test data, and more worried about actual `DrugStock` records having `redistributed: nil`, and whether this class can handle such records. I think we're expecting `redistributed: nil` because I saw calculations like `something - (redistributed || 0)`

app/controllers/webview/drug_stocks_controller.rb
prabhanshuguptagit: We've changed the form's structure in the view to let us split the form fields HTML without any JS. 

Earlier it was an array of protocol drug objects
```
[{protocol_drug_id: , in_stock:, received: }, ....]
```

Now it is a hash that looks like
```
{ 
  "0" => {protocol_drug_id: , in_stock:, received: }, 
  "1" => {protocol_drug_id: , in_stock:, received: }
   ...
}
```

spec/exporters/drug_consumption_report_exporter_spec.rb
prabhanshuguptagit: That's right 👍🏼 , we expect most drug stocks to have `redistributed: nil` since only a few facilities redistribute drugs and will fill the field in the form.

app/controllers/webview/drug_stocks_controller.rb
harimohanraj89: Ah understood 👍

spec/exporters/drug_consumption_report_exporter_spec.rb
harimohanraj89: Gotcha. Could we make one of these zeros into a `nil` to ensure that these things can work with such records?

harimohanraj89 (approved):
🚀

PR #2720, opened 2021-07-13
Fix reportable_children for org's in BD

rsanheim:
This fix is necessary for Organization reports to work on Bangladesh.

**Story card:** [ch4030](https://app.clubhouse.io/simpledotorg/story/4030/enable-org-wide-reports-on-bangladesh)

### DIFF
--- a/spec/models/region_spec.rb
+++ b/spec/models/region_spec.rb
@@ -51,10 +51,12 @@ RSpec.describe Region, type: :model do
     it "is everything for India" do
       expect(CountryConfig).to receive(:current).and_return(CountryConfig.for(:IN)).at_least(:once)
 
+      org = Seed.seed_org.region
       state = FactoryBot.create(:region, :state, reparent_to: Seed.seed_org.region)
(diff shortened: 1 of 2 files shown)


spec/models/region_spec.rb
```diff
@@ -70,6 +72,14 @@
       facility_region = facility.region
       expect(district.reportable_children).to contain_exactly(facility_region)
     end
+
+    it "works for Bangladesh organizations" do
+      allow(CountryConfig).to receive(:current).and_return(CountryConfig.for(:BD)).at_least(:once)
```
kpethtel, line 77: Does the `at_least(:once)` do anything here? Seems like it's more intended for use with `expect` than `allow`.

kpethtel (approved):

spec/models/region_spec.rb
rsanheim: Yup, it ensures that the code path we are calling is really getting hit.

If we don't have that, and someone changes the country config check, we could end up with false positives here.

PR #2723, opened 2021-07-14
4058: migrate imo auth to use UUID

kpethtel:
**Story card:** [ch4058](https://app.clubhouse.io/simpledotorg/story/4058/imo-invitation-callbacks)

## Because

I accidentally created ImoAuth to use integer IDs instead of uuids. Changing to use UUID like most other models.


db/migrate/20210714183848_change_imo_authorization_id_type_to_uuid.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeImoAuthorizationIdTypeToUuid < ActiveRecord::Migration[5.2]
```
kpethtel, line 1: I'm not sure this is really needed, but I noticed that I forgot to make this model use UUIDs.

rsanheim (approved):
makes sense to me to keep things consistent

PR #2724, opened 2021-07-15
Remove feature flag for DM in progress tab

harimohanraj89:
**Story card:** [ch4166](https://app.clubhouse.io/simpledotorg/story/4166/env-cleanup)

## Because

We don't need this feature flag anymore. Also, it's accidentally off in Ethiopia, and needs to be on. ([Slack ref](https://simpledotorg.slack.com/archives/CV9KNKY06/p1626261454029400))


prabhanshuguptagit (approved):
LGTM

PR #2725, opened 2021-07-15
Alphabetize env

harimohanraj89:
**Story card:** [ch4166](https://app.clubhouse.io/simpledotorg/story/4166/env-cleanup)

Alphabetize our .env files for objective consistency. I did a `cat | sort` at the end for good measure.

**Note:** I've also added some more removals from the `app.json` file into the same PR, because the removals don't overlap. Let me know if you want me to extract to a separate PR (it seemed like overkill for this situation)


prabhanshuguptagit (approved):
✨ Looks great. Can we also verify that these are now good to rely on as the canonical env files? All the vars from india production's .env post-https://github.com/simpledotorg/deployment/pull/348 should be here.

harimohanraj89:
The differences lie in the `ENABLE_` flags. Some of them are explicitly set to false, and some are just omitted. Apart from those, I'll compare the files to see if they're all consistent now.

I'll open a follow-up PR on `deployment` to alphabetize those too. Makes it a lot easier to compare.

PR #2726, opened 2021-07-15
Extract temporary retention duration seconds into a config

ssrihari:
**Story card:** [ch4177](https://app.clubhouse.io/simpledotorg/story/4177/extract-retention-period-on-the-server-to-an-env-config)


harimohanraj89 (changes requested):
We should either add this to [the required list of vars](https://github.com/simpledotorg/simple-server/blob/master/config/initializers/config.rb#L1), or fall back to a default. We don't wanna send zeros back, do we?

harimohanraj89 (approved):
Amaze! Let's ship the corresponding `deployment` PR first and then 🚀

rsanheim:
I checked the story but don't see any context -- why do we want this to be an env var?  Is it `86400` in all prod-ish environments?

harimohanraj89:
Good call. It's because we want to easily tune the duration down drastically in QA and SBX to speed up the testing cycle while we ship this work. @ssrihari I'll let you document this in the story/epic, and craft a cleanup story for later if you wanna rip out the env var after launch.

ssrihari:
@harimohanraj89 @rsanheim I've updated the story with details now. Apologies for not doing this earlier 🙇

rsanheim:
Thanks!

PR #2728, opened 2021-07-15
Fixes and unit tests for reporting pipeline enhancements

ssrihari:
Adding tests and some fixes around the [enhancements PR](https://github.com/simpledotorg/simple-server/pull/2683).
Please use the schema diff to see the differences in the actual SQL.

- Fix `quarters` computation: had made a stupid mistake of using months to calculate quarters :/
- Fix calculation of month to figure out facilities visited in a given month
- Add unit tests for:
    - `current_age`
    - `quarters_since_*`
    - new `htn_treatment_outcome_*` fields
    - `visited_facility_ids`
- Add a new factory to create blood sugars with encounters by default.

Note: for `visited_facility_ids`, we currently use only the *latest encounter* in the visits table. So, if a patient were to get a BP and Sugar measurement in different facilities in the same month, then they won't be counted in here. Luckily, [we don't have any such data in production](https://metabase.simple.org/question/320), and it seems reasonable that it would remain that way for a while. I think we should pick up this change when we have first class support for visits with encounters covering appointments and medicines as well.


spec/models/reports/patient_state_spec.rb
```diff
@@ -225,6 +244,151 @@
...
+        RefreshMaterializedViews.new.refresh_v2
+
+        with_reporting_time_zone do
+          expect(described_class.where(htn_treatment_outcome_in_last_2_months: "visited_no_bp", month_date: june_2021[:now]).pluck(:patient_id))
+            .to include(patient_bp_over_2_months.id, patient_with_no_bp.id)
+          expect(described_class.where(htn_treatment_outcome_in_last_2_months: "visited_no_bp", month_date: june_2021[:now]).pluck(:patient_id))
+            .not_to include(patient_bp_under_2_months.id)
+        end
+      end
+      it "controlled/uncontrolled if there is a BP measured in the last 2 months that is under/not under control" do
```
rsanheim, line 295: missing some spacing here

spec/models/reports/patient_state_spec.rb
```diff
@@ -225,6 +244,151 @@
...
+        patient_controlled = create(:patient, recorded_at: june_2021[:long_ago])
+        create(:blood_pressure, :with_encounter, patient: patient_controlled, recorded_at: june_2021[:now] - 1.month, systolic: 139, diastolic: 89)
```
rsanheim, line 297: we can use the factory trait here, its more expressive

```suggestion
        create(:blood_pressure, :with_encounter, :under_control, patient: patient_controlled, recorded_at: june_2021[:now] - 1.month)
```

spec/models/reports/patient_state_spec.rb
```diff
@@ -225,6 +244,151 @@
...
+
+        patient_uncontrolled = create(:patient, recorded_at: june_2021[:long_ago])
+        create(:blood_pressure, :with_encounter, patient: patient_uncontrolled, recorded_at: june_2021[:now] - 1.months, systolic: 140, diastolic: 90)
```
rsanheim, line 300: same here

```suggestion
        create(:blood_pressure, :with_encounter, :hypertensive, patient: patient_uncontrolled, recorded_at: june_2021[:now] - 1.months)
```

spec/models/reports/patient_state_spec.rb
```diff
@@ -225,6 +244,151 @@
...
+        RefreshMaterializedViews.new.refresh_v2
+
+        with_reporting_time_zone do
+          expect(described_class.where(htn_treatment_outcome_in_quarter: "visited_no_bp", month_date: june_2021[:now]).pluck(:patient_id))
+            .to include(patient_bp_in_last_quarter.id, patient_with_no_bp.id)
+          expect(described_class.where(htn_treatment_outcome_in_quarter: "visited_no_bp", month_date: june_2021[:now]).pluck(:patient_id))
+            .not_to include(patient_bp_in_this_quarter.id)
+        end
+      end
+      it "controlled/uncontrolled if there is a BP measured in this quarter that is under/not under control" do
```
rsanheim, line 369: missing new line

spec/models/reports/patient_state_spec.rb
```diff
@@ -225,6 +244,151 @@
...
+        patient_controlled = create(:patient, recorded_at: june_2021[:long_ago])
+        create(:blood_pressure, :with_encounter, patient: patient_controlled, recorded_at: june_2021[:now], systolic: 139, diastolic: 89)
+
+        patient_uncontrolled = create(:patient, recorded_at: june_2021[:long_ago])
+        create(:blood_pressure, :with_encounter, patient: patient_uncontrolled, recorded_at: june_2021[:now], systolic: 140, diastolic: 90)
```
rsanheim, line 374: both of these can use the traits

db/migrate/20210715120731_fix_quarter_computation_in_reporting_tables.rb
```diff
@@ -0,0 +1,41 @@
...
+    drop_view :reporting_patient_visits, materialized: true
+    drop_view :reporting_patient_blood_pressures, materialized: true
+
+    execute "SET LOCAL TIME ZONE '#{Rails.application.config.country[:time_zone]}'"
+
+    create_view :reporting_patient_blood_pressures, version: 2, materialized: true
+    create_view :reporting_patient_visits, version: 2, materialized: true
+    create_view :reporting_patient_states, version: 2, materialized: true
+
+    add_index :reporting_patient_states, [:assigned_block_region_id], name: :patient_states_assigned_block
```
rsanheim, line 34: Need to add the unique indexes back for the other two matviews as well to keep the `down` a true reverse of the `up`.

db/migrate/20210715120731_fix_quarter_computation_in_reporting_tables.rb
```diff
@@ -0,0 +1,41 @@
...
+    create_view :reporting_patient_states, version: 3, materialized: true
+
+    add_index :reporting_patient_blood_pressures, [:month_date, :patient_id], unique: true, name: "patient_blood_pressures_patient_id_month_date"
+    add_index :reporting_patient_visits, [:month_date, :patient_id], unique: true, name: "patient_visits_patient_id_month_date"
+    add_index :reporting_patient_states, [:assigned_block_region_id], name: :patient_states_assigned_block
+    add_index :reporting_patient_states, [:assigned_district_region_id], name: :patient_states_assigned_district
+    add_index :reporting_patient_states, [:assigned_facility_region_id], name: :patient_states_assigned_facility
+    add_index :reporting_patient_states, [:assigned_state_region_id], name: :patient_states_assigned_state
+    add_index :reporting_patient_states, [:hypertension, :htn_care_state, :htn_treatment_outcome_in_last_3_months], name: :patient_states_care_state
+    add_index :reporting_patient_states, [:month_date, :patient_id], name: :patient_states_month_date_patient_id, unique: true
```
rsanheim, line 20: Did you try using `update_view` here and starting w/ `patient_states` first, and working your way back?  That would mean scenic will maintain all your indexes and recreated them for you.

In any case, I suppose you can at least do `update_view` for patient_states, as nothing depends on it (yet), which would mean one less matview to have to copy/paste indexes for. 😁

db/views/reporting_patient_visits_v03.sql
```diff
@@ -0,0 +1,102 @@
...
+    app.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' AS appointment_recorded_at,
+
+    array_remove(
+        ARRAY[
+            (CASE WHEN to_char(e.recorded_at AT TIME ZONE 'utc' AT TIME ZONE (SELECT current_setting('TIMEZONE')), 'YYYY-MM') = month_string THEN e.facility_id END),
+            (CASE WHEN to_char(pd.recorded_at AT TIME ZONE 'utc' AT TIME ZONE (SELECT current_setting('TIMEZONE')), 'YYYY-MM') = month_string THEN pd.facility_id END),
+            (CASE WHEN to_char(app.recorded_at AT TIME ZONE 'utc' AT TIME ZONE (SELECT current_setting('TIMEZONE')), 'YYYY-MM') = month_string THEN app.creation_facility_id END)
+        ],
+        null
+    ) AS visited_facility_ids,
```
harimohanraj89, line 39: I might have missed this column in the initial enhancements PR, my bad. This field seems a bit gnarly to me. Would it only contain facility IDs if the visit happened _this_ month? Eg. If I last visited in May, would the June "visited_facility_ids" be empty?

Intuitively, I'd think that I'd have a `months_since_visit`, and a `visit_facility_id` that is the facility of my most recent visit, regardless of how long ago.

spec/models/reports/patient_state_spec.rb
```diff
@@ -11,12 +11,31 @@
...
+      end
+
+      it "determines the current age of the patient given their dob, and prefers it even when age is present" do
+        patient = create(:patient, date_of_birth: Date.new(1941, 6, 1), age: 58, recorded_at: 2.years.ago, age_updated_at: 2.years.ago)
+        described_class.refresh
+        with_reporting_time_zone do
+          expect(described_class.where(patient_id: patient.id).pluck(:current_age)).to all eq(80)
+        end
+      end
+    end
```
harimohanraj89, line 37: The tests make the behavior super clear. I would have expected the `current_age` of the patient to reflect the age _as of the reporting month_. Not as of time of refreshing the views. The same patient would have several state records, and you'd be able to see the age change from 58 to 59 to 60, etc.

spec/models/reports/patient_state_spec.rb
harimohanraj89: Also I think it's ok to split this into two examples. 4 outcomes, 4 specs.

spec/models/reports/patient_state_spec.rb
harimohanraj89: Same comment about splitting.

spec/models/reports/patient_state_spec.rb
```diff
@@ -369,16 +550,28 @@ def patient_states(patient, from: nil, to: nil)
...
             expect(patient_states(patient, to: ten_months_ago).pluck(:months_since_bp)).to all(be_nil)
             expect(patient_states(patient, from: ten_months_ago, to: five_months_ago).pluck(:months_since_bp)).to eq((0..4).to_a)
             expect(patient_states(patient, from: five_months_ago, to: now).pluck(:months_since_bp)).to eq((0...5).to_a)
 
+            expect(patient_states(patient, to: ten_months_ago).pluck(:quarters_since_bp)).to all(be_nil)
+            expect(patient_states(patient, from: ten_months_ago, to: eight_months_ago).pluck(:quarters_since_bp)).to all eq 0
+            expect(patient_states(patient, from: eight_months_ago, to: five_months_ago).pluck(:quarters_since_bp)).to all eq 1
+            expect(patient_states(patient, from: five_months_ago, to: two_months_ago).pluck(:quarters_since_bp)).to all eq 0
+            expect(patient_states(patient, from: two_months_ago, to: now).pluck(:quarters_since_bp)).to all eq 1
+
```
harimohanraj89, line 574: If we change the age calculation, I'd love to see timeline specs for age as well.

harimohanraj89 (changes requested):

db/migrate/20210715120731_fix_quarter_computation_in_reporting_tables.rb
ssrihari: Since we're also changing the lower matviews (views etc) in this one, we have to drop `patient_states` first anyway. So we're forced to do a create instead of an update.

@harimohanraj89 in the next PR, I'd like to try using `ALTER MATERIALIZED VIEW` like I suggested [on slack](https://simpledotorg.slack.com/archives/C01JJJM1NA1/p1626369609084900).

db/views/reporting_patient_visits_v03.sql
ssrihari: x-posting a comment from [elsewhere](https://simpledotorg.slack.com/archives/C01JJJM1NA1/p1626673495091300?thread_ts=1626431704.090700&cid=C01JJJM1NA1):

The follow up definitions in `patient_states` would be flawed regardless of the implementation of `visited_facility_ids`, so I’m in favour of removing this, and other follow up related columns as well _in the next PR_. I’ll know more after I deal with the one-off request to create follow up reports with the new definitions. 

- The current matviews hold only “_latest_” data in the month, which means if there are 2 BPs in 2 different facilities, we would only consider one of them in the reports, which is incorrect
- The views matview uses encounters, so it inherently suppresses either BP/Sugar reading to get us the latest information

My takeaway from all this is that “follow up” reports tell us about the facility’s activity/health, whereas the treatment-outcome etc tell us about the patient’s health. I.e, the follow up reports are not longitudinal patient reports, and we can’t use the patient-states table for it.

The current plan (in my head) is to create the follow up report elsewhere (a different matview perhaps) and add that information to the facility states matview.

spec/models/reports/patient_state_spec.rb
ssrihari: I too struggled a bit to choose which version of age we need in the reports, and ended up choosing this version because it is what we need according to some of the [one-off requests](https://app.clubhouse.io/simpledotorg/story/3967/ihci-annual-report-2021-follow-up) we have now.

We can totally add the rolling age as another column. Wdyt?

spec/models/reports/patient_state_spec.rb
```diff
@@ -11,12 +11,31 @@
     # because that falls into the next day in IST (our reporting time zone). So to prevent confusing failures for
     # developers or CI during North American afternoons, we freeze to a time that will be the end of the month for
     # UTC, ET, and IST. Timezones! 🤯
-    Timecop.freeze("June 30 2021 5:30 UTC") do # June 30th 23:00 IST time
+    Timecop.freeze("June 30 2021 5:30 UTC") do
```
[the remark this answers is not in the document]
ssrihari: Thanks rob, fixed it in 9fd44bedc28db74657a34b20faba397116da8459. It was also bothering me that 5:30 UTC isn't 23:00 IST 😂

spec/models/reports/patient_state_spec.rb
ssrihari: Fixed spacing in ed649e75367c82231d69e97799b2407b63c4b818 and split examples in c39ffe8d4ee95c936b28fa3ae8d4e168a3cc97e9.

spec/models/reports/patient_state_spec.rb
ssrihari: 💯  Thanks! Fixed in da2d6a8367bea0f872ed58601741ead5c2954a4b.

spec/models/reports/patient_state_spec.rb
ssrihari: Fixed in da2d6a8367bea0f872ed58601741ead5c2954a4b.

spec/models/reports/patient_state_spec.rb
ssrihari: Fixed spacing in ed649e7 and split examples in c39ffe8.

db/migrate/20210715120731_fix_quarter_computation_in_reporting_tables.rb
ssrihari: Phew, good catch! 😓  Fixed in 190189675e981e1065ab84e767cba7fb7619db00.

spec/models/reports/patient_state_spec.rb
harimohanraj89: I'd punt on it until we get a concrete requirement. Nobod's asking for age segmentation yet. Let's see what it looks like when we actually get that request.

rsanheim (approved):
Changes here look good!

PR #2730, opened 2021-07-15
4058 imo invitation callback

kpethtel:
**Story card:** [ch4058](https://app.clubhouse.io/simpledotorg/story/4058/imo-invitation-callbacks?vc_group_by=day&ct_workflow=all&cf_workflow=500000005)

## Because

Imo sends a callback to our invitations. We want to process those callbacks to ensure that we only send notifications to patients who've subscribed.


app/controllers/api/v3/imo_callbacks_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  private
+
+  def permitted_params
+    params.permit(:patient_id, :event)
+  end
+
+  def authenticate_request
+    authenticate_or_request_with_http_basic do |username, password|
```
rsanheim, line 31: Looks like it can be even simpler - see https://api.rubyonrails.org/v5.2.1/classes/ActionController/HttpAuthentication/Basic.html

```
   http_basic_authenticate_with name: "dhh", password: "secret", except: :index
```

rsanheim (approved):
one small thing that can be simpler, otherwise looks good

PR #2709, opened 2021-07-09
Update Datadog to 5.2.0 (latest)

rsanheim:
**Story card:** [ch4102](https://app.clubhouse.io/simpledotorg/story/4102/upgrade-dogstatsd-ruby-to-latest-version-s)

This PR can follow https://github.com/simpledotorg/simple-server/pull/2708

dogstatsd-ruby changelog - https://github.com/DataDog/dogstatsd-ruby/blob/master/CHANGELOG.md

PR #2732, opened 2021-07-15
One more datadog client update

rsanheim:
Apparently we need to update ddtrace to match the statsd collector now.

These two gems should really have some sort of hard dependancy on each
other.

https://github.com/DataDog/dd-trace-rb/issues/1491

re: #2709


rsanheim:
Going to merge this w/o a review because the DataDog docs warn against thread leaks (and mem leaks) if we aren't on the newest version here.

PR #2733, opened 2021-07-16
Drug stock district warehouse

prabhanshuguptagit:
**Story card:** [ch3321](https://app.clubhouse.io/simpledotorg/story/3321/support-district-level-drug-inventory)

## Because

image

## This addresses

Adds capability to add drug stocks for a district warehouse.
`DrugStockQuery` now returns block and district information in `drug_stocks_report`. The webview calls the required methods instead of asking for the entire hash.

Pending:
- [x] Add district row in download
- [x] Add district row for consumption and consumption download

Follow up work: https://app.clubhouse.io/simpledotorg/story/4083/cleanup-drop-facility-id-column-from-drug-stocks


app/queries/drug_stocks_query.rb
```diff
@@ -3,46 +3,60 @@ class DrugStocksQuery
 
   CACHE_VERSION = 1
 
-  def initialize(facilities:, for_end_of_month:, include_block_report: true)
```
harimohanraj89, line 6: Are we biting the bullet and always including block-level info? Let's capture that in the PR description, it's useful for reviewers.

app/queries/drug_stocks_query.rb
```diff
@@ -64,16 +78,16 @@ def drug_consumption_report
   end
 
   memoize def repository
-    Reports::Repository.new(Region.where(source: @facilities), periods: Period.month(@for_end_of_month))
+    Reports::Repository.new(Region.where(source: @facilities) + [@district], periods: Period.month(@for_end_of_month))
```
harimohanraj89, line 81: Can we extract this region math to a semantically named method?

harimohanraj89: Additionally, would accepting a list of facility _regions_ make this less awkward?

db/schema.rb
```diff
@@ -174,9 +173,9 @@
     t.datetime "deleted_at"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.uuid "region_id"
+    t.uuid "region_id", null: false
+    t.uuid "facility_id"
```
harimohanraj89, line 177: Do we keep this around to preserve older data? Can we get away with _not_ having to do that? A data migration plus dropping the column?

app/models/drug_stock.rb
```diff
@@ -32,4 +46,10 @@ def self.latest_for_facilities_cte(facilities, for_end_of_month)
     from(latest_for_facilities(facilities, for_end_of_month), table_name)
       .includes(:protocol_drug)
   end
```
harimohanraj89, line 48: If we can manage to drop the `facility` from drug stocks, we could remove these `latest_for_facilities*` methods too, right?

app/queries/drug_stocks_query.rb
```diff
@@ -3,46 +3,60 @@ class DrugStocksQuery
...
+     facilities_total_drugs_in_stock: facilities_total_drugs_in_stock,
+
+     patient_count_by_facility_id: patient_count_by_facility_id,
+     patient_days_by_facility_id: patient_days_by_facility_id,
+     drugs_in_stock_by_facility_id: drugs_in_stock_by_facility_id,
+
+     patient_count_by_block_id: patient_count_by_block_id,
+     patient_days_by_block_id: patient_days_by_block_id,
+     drugs_in_stock_by_block_id: drugs_in_stock_by_block_id,
+     last_updated_at: Time.now}
```
harimohanraj89, line 39: This is probably my favorite section. It's like the table of contents for this query object.

harimohanraj89 (changes requested):
I like the idea of rallying around regions consistently to power the district warehouse. It's an awkward feature to shoehorn in here, and this looks good. Most of my questions are around seeing that idea to completion.

app/queries/drug_stocks_query.rb
prabhanshuguptagit: We're trying to make a single call to repository to extract the patients under care number for both facilities and districts. The regions don't have anything in common, it's just a list of regions we want to lookup in the repository. A method like `repository_regions` doesn't seem very useful.

We can also make two repository instances: `facilities_repository` and a `district_repository`, although that doesn't buy us much semantically.

wdyt of 94df952?

db/schema.rb
prabhanshuguptagit: We want to do this eventually but it requires more refactoring across the view and the controller. We can do this iteratively, don't want to block releasing this on the cleanup.

app/models/drug_stock.rb
prabhanshuguptagit: Yes 👍🏼

app/queries/drug_stocks_query.rb
harimohanraj89: Oh this is only for patients under care? Interesting. Yeah that commit looks good 😅

db/schema.rb
harimohanraj89: Can you craft a story and link it here?

harimohanraj89 (approved):
Let's do this 😎 

Importantly, let's document the follow-up refactor in a story that we can get to.

PR #2734, opened 2021-07-16
Grab data from Facility states for Repository

rsanheim:


db/views/reporting_facility_states_v01.sql
```diff
@@ -0,0 +1,77 @@
...
+               count(distinct(patient_id)) FILTER (WHERE htn_treatment_outcome_in_last_3_months = 'uncontrolled' AND htn_care_state = 'lost_to_follow_up') as uncontrolled_lost_to_follow_up,
+
+               count(distinct(patient_id)) FILTER (WHERE htn_treatment_outcome_in_last_3_months = 'missed_visit' AND htn_care_state = 'under_care') as missed_visit_under_care,
+               count(distinct(patient_id)) FILTER (WHERE htn_treatment_outcome_in_last_3_months = 'missed_visit' AND htn_care_state = 'lost_to_follow_up') as missed_visit_lost_to_follow_up,
+
+               count(distinct(patient_id)) FILTER (WHERE htn_treatment_outcome_in_last_3_months = 'visited_no_bp' AND htn_care_state = 'under_care') as visited_no_bp_under_care,
+               count(distinct(patient_id)) FILTER (WHERE htn_treatment_outcome_in_last_3_months = 'visited_no_bp' AND htn_care_state = 'lost_to_follow_up') as visited_no_bp_lost_to_follow_up,
+
+               count(distinct(patient_id)) FILTER (WHERE htn_care_state = 'under_care') as patients_under_care,
+               count(distinct(patient_id)) FILTER (WHERE htn_care_state = 'lost_to_follow_up') as patients_lost_to_follow_up
```
ssrihari, line 38: Remove invalid fields from here.

app/models/reports/repository.rb
```diff
@@ -6,8 +6,8 @@ class Repository
     include Scientist
     PERCENTAGE_PRECISION = 0
 
-    def initialize(regions, periods:, reporting_schema_v2: false)
-      @regions = Array(regions)
+    def initialize(regions, periods:, reporting_schema_v2: true)
```
rsanheim, line 9: Changing this temporarily to true to see what falls out from CI.

app/models/reports/repository.rb
```diff
@@ -100,15 +100,28 @@ def reporting_schema_v2?
 
     alias_method :adjusted_patients, :adjusted_patients_without_ltfu
 
-    # Return the running total of cumulative assigned patient counts.
+    # Returns cumulative assigned patients from facility_states - this includes LTFU
+    private def cumulative_assigned_patients_query_v2(region)
+      Reports::FacilityState.for_facility(region).order(:month_date).pluck(:month_date, :assigned_patients)
```
rsanheim, line 105: I realized the better method name for this is `cumulative_assigned...`, to better match the conventions we have here in Repository. We probably want to change it in the matviews as well.

db/schema.rb
```diff
@@ -1449,4 +1449,91 @@
...
+      tout.missed_visit_lost_to_follow_up,
+      tout.visited_no_bp_under_care,
+      tout.visited_no_bp_lost_to_follow_up,
+      tout.patients_under_care,
+      tout.patients_lost_to_follow_up
+     FROM ((((registered_patients rp
+       LEFT JOIN assigned_patients ap ON (((rp.region_id = ap.region_id) AND (rp.month_date = ap.month_date))))
+       LEFT JOIN treatment_outcomes_in_last_3_months tout ON (((rp.region_id = tout.region_id) AND (rp.month_date = tout.month_date))))
+       JOIN reporting_facilities rf ON ((rp.region_id = rf.facility_region_id)))
+       JOIN reporting_months cal ON ((rp.month_date = cal.month_date)));
```
rsanheim, line 1535: These were the only changes here -- we have to use `LEFT JOIN` on assigned and treatment outcomes, otherwise we would lose records for brand new facilities where they _only_ have patients registered but no assigned patients. This happened in a bunch of tests, which admittedly may not represent how prod data looks...but we've seen enough strange edge cases happen when new facilities start that I wouldn't be surprised if this could be a real thing.

spec/controllers/reports/regions_controller_spec.rb
```diff
@@ -258,8 +258,8 @@ def refresh_views
     it "returns period info for current month" do
       today = Date.current
       Timecop.freeze(today) do
-        patient = create(:patient, registration_facility: @facility, recorded_at: today)
-        create(:blood_pressure, :under_control, recorded_at: today, patient: patient, facility: @facility)
+        patient = create(:patient, registration_facility: @facility, recorded_at: 2.months.ago)
+        create(:bp_with_encounter, :under_control, recorded_at: Time.current.yesterday, patient: patient, facility: @facility)
```
rsanheim, line 262: Bunch of specs need the smarter fixture that includes an encounter for things to work w/ the new views.

spec/models/reports/repository_spec.rb
```diff
@@ -112,7 +120,10 @@ def refresh_views
 
     it "gets controlled counts and rates for single region" do
       facilities = FactoryBot.create_list(:facility, 2, facility_group: facility_group_1).sort_by(&:slug)
+      other_facility = create(:facility)
       facility_1, facility_2 = facilities.take(2)
+      # TODO we shouldn't need to create a registered patient here to make FacilityStates return records...
+      _facility_1_registered_but_not_assigned = create(:patient, recorded_at: jan_2019, assigned_facility: other_facility, registration_facility: facility_1, registration_user: user)
```
rsanheim, line 126: Here was one other problem w/ the JOINs we have -- we need a registered patient early enough w/ a facility for the left join to work.  Maybe we could use a `FULL OUTER JOIN` across registered, assigned, and treatment outcomes to get all the things regardless of what happens first for a facility?  

Thinking thru a real world case of this: a facility starts at the end of the month, some patients get assigned at month end, and they don't get new registrations until the start of the next month.  The assigned numbers for the end of month would not be counted until registrations start happening.

spec/models/reports/repository_spec.rb
```diff
@@ -312,7 +324,7 @@ def refresh_views
     end
   end
 
-  context "caching" do
+  context "caching", skip: "not worrying about caching specs for now" do
```
rsanheim, line 327: We are doing straight SELECTs from facility_states, we probably don't need caching or if we do, it could be much less.

rsanheim (commented):
Notes on progress for @ssrihari and @simpledotorg/server-developers

app/models/reports/repository.rb
ssrihari: Renamed in fe2501e6968a8b23864e88fedb3cf2d379bc17bb.

db/schema.rb
ssrihari: Makes sense, @rsanheim. I've made a change with fe2501e6968a8b23864e88fedb3cf2d379bc17bb that ensures that we will have a row for every facility and month combination. Even if a bit redundant, I feel this is simple and creates a certainty on the rows present in the table.

spec/models/reports/repository_spec.rb
ssrihari: https://github.com/simpledotorg/simple-server/pull/2734#discussion_r672877926

app/models/reports/patient_state.rb
```diff
@@ -2,5 +2,14 @@ module Reports
...
     self.table_name = "reporting_patient_states"
     belongs_to :patient
+
+    def self.by_assigned_region(region_or_source)
+      region = region_or_source.region
+      where("assigned_#{region.region_type}_region_id" => region.id)
+    end
+
+    def self.by_month_date(month_date)
+      where(month_date: month_date)
```
ssrihari, line 12: This seems like an alias we could avoid?

app/models/reports/repository.rb
```diff
@@ -120,21 +133,45 @@ def reporting_schema_v2?
       end
     end
 
+    private def registered_patients_query_v2(region)
+      Reports::FacilityState.for_facility(region).order(:month_date).pluck(:month_date, :monthly_registrations)
```
ssrihari, line 137: I feel using a `where` directly is more transparent, and familiar. Wdyt? Oh and we drop the `Reports::` prefix since we're in the same module, ya?

```ruby
FacilityState
  .where(facility_region_id: region.id)
  .order(:month_date)
  .pluck(:month_date, :monthly_registrations)
```

app/models/reports/repository.rb
```diff
@@ -120,21 +133,45 @@ def reporting_schema_v2?
...
     memoize def complete_monthly_registrations
-      items = regions.map { |region| RegionEntry.new(region, __method__, period_type: period_type) }
-      cache.fetch_multi(*items, force: bust_cache?) { |entry|
-        registered_patients_query.count(entry.region, period_type)
-      }
+      if reporting_schema_v2?
+        regions.each_with_object({}) { |region, result|
+          result[region.slug] = registered_patients_query_v2(region).each_with_object(Hash.new(0)) { |(month_date, count), hsh|
+            hsh[Period.month(month_date)] = count
+          }
```
ssrihari, line 151: Is this worth extracting? Keying the results by `region.slug` and `month_date` would happen often enough (twice in this PR) right?

app/models/period.rb
```diff
@@ -50,6 +50,7 @@ def initialize(attributes = {})
     else
       value.to_date.beginning_of_month
     end
+    validate!
```
rsanheim, line 53: Fail fast if a Period is created w/o type or value (which I saw during some of this refactoring).

app/models/reports/repository.rb
```diff
@@ -16,112 +23,55 @@ def initialize(regions, periods:, reporting_schema_v2: false)
       @period_type = @periods.first.type
       @reporting_schema_v2 = reporting_schema_v2
       raise ArgumentError, "Quarter periods not supported" if @period_type != :month
+      @schema = if reporting_schema_v2?
```
rsanheim, line 26: This is the "schema" or maybe "data access object" that either grabs things from the v1 matviews or the new v2 ones.

app/models/reports/repository.rb
```diff
@@ -16,112 +23,55 @@ def initialize(regions, periods:, reporting_schema_v2: false)
...
+      :earliest_patient_recorded_at,
+      :earliest_patient_recorded_at_period,
+      :ltfu_rates,
+      :ltfu,
+      :monthly_registrations,
+      :uncontrolled_rates,
+      :uncontrolled
+    ]
+
+    delegate(*DELEGATED_METHODS, to: :schema)
```
rsanheim, line 71: All of these methods are implemented in both v1 and v2 schemas, hence the delegation.

Everything else below is either not retrievable yet from v2 (follows up by user), or is something that I haven't dug into yet in terms of how to grab it from the v2 schema.

app/models/reports/repository.rb
```diff
@@ -133,68 +83,6 @@ def reporting_schema_v2?
...
-      regions.each_with_object({}).each do |region, hsh|
-        if earliest_patient_recorded_at[region.slug].nil?
-          hsh[region.slug] = {}
-          next
-        end
-        hsh[region.slug] = control_rate_query_v2.uncontrolled_counts(region, range: active_range(region))
-      end
-    end
-
     memoize def missed_visits_without_ltfu
```
rsanheim, line 86: All the missed visits queries can be moved over to v2, but I'm going to that in a follow up PR, as this PR has gotten quite big already

app/models/reports/schema_v1.rb
```diff
@@ -0,0 +1,282 @@
+module Reports
+  class SchemaV1
```
rsanheim, line 2: This is the v1 queries copied over from Repository.

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,286 @@
+module Reports
+  class SchemaV2
```
rsanheim, line 2: Every query in this object will grab from the new reporting pipeline.

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,286 @@
...
+      }
+      cached_results.each_with_object({}) { |(region_entry, time), results| results[region_entry.slug] = time }
+    end
+
+    memoize def earliest_patient_recorded_at_period
+      earliest_patient_recorded_at.each_with_object({}) { |(slug, time), hsh| hsh[slug] = Period.new(value: time, type: @period_type) if time }
+    end
+
+    def earliest_patient_data_query_v2(region)
+      FacilityState.for_region(region).where("cumulative_registrations > 0 OR cumulative_assigned_patients > 0")
```
rsanheim, line 37: Finding the 'birth date' of a region.

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,228 @@
...
+      }
+    end
+
+    private def adjusted_patients_without_ltfu_query(region)
+      FacilityState.for_region(region)
+        .order(:month_date)
+        .group(:month_date)
+        .sum("patients_under_care::int")
+        .to_h(&period_hash)
+    end
```
rsanheim: Many of the queries now follow this same pattern, so we are duplicating lots of code w/ the exception of the field we are summing.

I'm going to consolidate these queries into one query for simplicity and performance in the next story / PR -- see https://app.clubhouse.io/simpledotorg/story/4310/grab-as-much-repository-schema-v2-data-w-one-query-as-possible.

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,228 @@
...
+        result[region.slug] = cumulative_assigned_patients_query_v2(region)
+      }
+    end
+
+    # Returns cumulative assigned patients from facility_states - this includes LTFU
+    private def cumulative_assigned_patients_query_v2(region)
+      FacilityState.for_region(region)
+        .order(:month_date)
+        .group(:month_date)
+        .sum("cumulative_assigned_patients::int")
```
rsanheim, line 83: We cast to `int` because otherwise the summed values are returned as floats...which doesn't matter much in the Reports (we show integers there)...but does make comparing and debugging in dev/test troublesome.  Hence the cast.

I'm not really sure why PSQL / Rails return the summed values as floats in the first place, but its not a big deal right now.

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,228 @@
...
+
+    memoize def uncontrolled
+      regions.each_with_object({}).each do |region, hsh|
+        if earliest_patient_recorded_at[region.slug].nil?
+          hsh[region.slug] = Hash.new(0)
+          next
+        end
+        hsh[region.slug] = control_rate_query_v2.uncontrolled_counts(region, range: active_range(region))
+      end
+    end
```
rsanheim, line 161: These need to be converted to use FacilityState ...future story here https://app.clubhouse.io/simpledotorg/story/4310/grab-as-much-repository-schema-v2-data-w-one-query-as-possible

rsanheim (commented):
This PR is looking good in dev with real world data sets when toggling on v2.   I'd like to get it merged very soon to avoid merge conflicts and to work w/ some smaller PRs for future iteration.

Notes and explanation in inline comments below.

app/models/reports/facility_state.rb
```diff
@@ -1,6 +1,12 @@
 module Reports
   class FacilityState < Matview
     self.table_name = "reporting_facility_states"
+
     belongs_to :facility
+
+    def self.for_region(region_or_source)
+      region = region_or_source.region
+      where("#{region.region_type}_region_id" => region.id)
+    end
```
ssrihari, line 10: Is it worth unit testing this? I would find those tests as valuable documentation for sure.

app/models/reports/patient_state.rb
ssrihari: Is this function used now?

app/models/reports/patient_state.rb
```diff
@@ -2,5 +2,14 @@ module Reports
   class PatientState < Matview
     self.table_name = "reporting_patient_states"
     belongs_to :patient
+
+    def self.by_assigned_region(region_or_source)
+      region = region_or_source.region
+      where("assigned_#{region.region_type}_region_id" => region.id)
+    end
```
ssrihari, line 9: Is this function used now?

app/models/reports/schema_v2.rb
ssrihari: Sweet! 🍬

app/models/reports/schema_v2.rb
ssrihari: Nice! I am trying to understand why it is important to treat facilities without any data differently. I.e, why can't we just show 0s in the graphs or tables, perhaps?

ssrihari: Also, can we inline this query?

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,230 @@
...
+
+    # Returns the earliest patient record for a Region from either assigned or registered patients. Note that this *ignores*
+    # the periods that are passed in for the Repository - this is the true 'earliest report date' for a Region.
+    memoize def earliest_patient_recorded_at
+      region_entries = regions.map { |region| RegionEntry.new(region, __method__) }
+      cached_results = cache.fetch_multi(*region_entries, force: bust_cache?) { |region_entry|
+        earliest_patient_data_query_v2(region_entry.region)
+      }
+      cached_results.each_with_object({}) { |(region_entry, time), results| results[region_entry.slug] = time }
```
ssrihari, line 29: Do we need to cache this?

app/models/reports/schema_v2.rb
ssrihari: I'll try to fix this in another PR. Created a story to track it here: https://app.clubhouse.io/simpledotorg/story/4311/ensure-integer-values-from-the-schema-are-returned-as-integers-not-floats

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,230 @@
...
+        .to_h(&period_hash)
+        .tap { |hsh| hsh.default = 0 }
+    end
+
+    # Returns registration counts per region / period
+    memoize def monthly_registrations
+      regions.each_with_object({}) { |region, result|
+        result[region.slug] = registered_patients_query_v2(region).slice(*periods.entries)
+      }
+    end
```
ssrihari, line 93: Would be cool to extract this as well :) Doesn't have to be in this PR though.

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,230 @@
...
+
+    def registered_patients_query_v2(region)
+      return {} if earliest_patient_recorded_at_period[region.slug].nil?
+      FacilityState.for_region(region)
+        .where("month_date >= ?", earliest_patient_recorded_at_period[region.slug].to_date)
+        .order(:month_date)
```
ssrihari, line 99: Why is it necessary to order the months? Also, is the order preserved when doing a `to_h`? Same question for the other queries that do the same.

spec/queries/drug_stocks_query_spec.rb
```diff
@@ -57,6 +57,7 @@
 
     before do
       allow_any_instance_of(Reports::DrugStockCalculation).to receive(:patient_days_coefficients).and_return(punjab_drug_stock_config)
+      RefreshMaterializedViews.call
```
ssrihari, line 60: Is this unrelated?

ssrihari (approved):
Looking good! Mostly minor comments, and stuff that could be fixed in the following PR as well.

app/models/reports/facility_state.rb
harimohanraj89: I'd love to have these tested with some straightforward specs, for sure.

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,230 @@
...
+        earliest_patient_data_query_v2(region_entry.region)
+      }
+      cached_results.each_with_object({}) { |(region_entry, time), results| results[region_entry.slug] = time }
+    end
+
+    memoize def earliest_patient_recorded_at_period
+      earliest_patient_recorded_at.each_with_object({}) { |(slug, time), hsh| hsh[slug] = Period.new(value: time, type: @period_type) if time }
+    end
+
+    private def earliest_patient_data_query_v2(region)
```
harimohanraj89, line 36: Does this method need to be suffixed with v2 if it's in the v2 schema?

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,230 @@
...
+      FacilityState.for_region(region).where("cumulative_registrations > 0 OR cumulative_assigned_patients > 0")
+        .minimum(:month_date)
+    end
+
+    # Adjusted patient counts are the patient counts from three months ago (the adjusted period) that
+    # are the basis for control rates. These counts DO include lost to follow up.
+    memoize def adjusted_patients_with_ltfu
```
harimohanraj89, line 43: Is this a good time to start computing this as the sum of control, uncontrol, missed, and visit-but-no-bp? One idea of this pipeline would be to avoid annoying in-memory manipulation such as this. We're building confidence that this'll add up by introducing better test coverage to the matviews, yes?

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,230 @@
...
+
+    # Return the running total of cumulative assigned patient counts. Note that this *includes* LTFU.
+    memoize def cumulative_assigned_patients
+      regions.each_with_object({}) { |region, result|
+        result[region.slug] = cumulative_assigned_patients_query_v2(region)
+      }
+    end
+
+    # Returns cumulative assigned patients from facility_states - this includes LTFU
+    private def cumulative_assigned_patients_query_v2(region)
```
harimohanraj89, line 79: Same question around v2 suffix.

app/models/reports/schema_v2.rb
harimohanraj89: Neat. This PR doesn't have to wait for that fix either, which is nice. We can rip out the typecasts when ready later.

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,230 @@
...
+    attr_reader :regions
+
+    delegate :cache, :logger, to: Rails
+
+    def initialize(regions, periods:)
+      @regions = regions
+      @periods = periods
+      @period_type = @periods.first.type
+      @control_rate_query_v2 = ControlRateQueryV2.new
+      @period_hash = lambda { |month_date, count| [Period.month(month_date), count] }
```
harimohanraj89, line 19: This is schwifty, I'm digging it.

app/models/reports/schema_v2.rb
```diff
@@ -0,0 +1,230 @@
...
+        }
+        results[entry] = values
+      end
+    end
+
+    # Adjusted patient counts are the patient counts from three months ago (the adjusted period) that
+    # are the basis for control rates. These counts DO NOT include lost to follow up.
+    memoize def adjusted_patients_without_ltfu
+      regions.each_with_object({}) { |region, result|
+        result[region.slug] = adjusted_patients_without_ltfu_query(region).slice(*periods.entries)
```
harimohanraj89, line 56: Should we be making use of `period.adjusted_period` here? I don't see where the 3 month shift is happening.

app/models/reports/schema_v2.rb
harimohanraj89: We limit the number of months shown on our graphs based on a region's birth date. If you've only got 5 months of data, no need to show 24 datapoints crowded together, we should 5 data points spaced apart (visually).

harimohanraj89 (commented):
Cursory comments, mostly wanted to take a look. I'll let @ssrihari give this thing thorough review 👍

app/models/reports/schema_v2.rb
rsanheim: Hopefully not - I'll do some measurements after consolidating queries, where I think the first step is to rip out the caching from v2.

app/models/reports/schema_v2.rb
rsanheim: yup, what @hari said 😄

PR #2735, opened 2021-07-19
4199: ensure that latest mobile number is localized to country

kpethtel:
**Story card:** [ch4199](https://app.clubhouse.io/simpledotorg/story/4199/ensure-that-all-phone-numbers-use-the-country-code)

## Because

Mobile numbers now need to be localized to country for Imo as well as Twilio, so I'm moving that functionality.


spec/models/patient_phone_number_spec.rb
```diff
@@ -103,4 +103,21 @@
...
+      phone_number = create(:patient_phone_number, number: "1234567890")
+      expect(phone_number.number_with_country_code).to eq("+911234567890")
+    end
+
+    it "does not change the number if it's already localized correctly" do
+      phone_number = create(:patient_phone_number, number: "+911234567890")
+      expect(phone_number.number_with_country_code).to eq("+911234567890")
+    end
+
+    it "changes the number to the current country code if it is localized to another country" do
```
kpethtel, line 118: Whether we should be doing it this way is debatable, but this is the current functionality we're using when sending messages via twilio.

app/models/patient_phone_number.rb
```diff
@@ -54,4 +54,10 @@ def update_whitelist_requested_at(time)
 
     exotel_phone_number_detail.update(whitelist_requested_at: time)
   end
+
+  def number_with_country_code
```
kpethtel, line 58: I think this should probably be happening as a validation on the model and "normalizing" the number on save. We should also probably not be overriding the country code is already present, but this PR is not intended to change existing functionality.

spec/models/patient_phone_number_spec.rb
```diff
@@ -103,4 +103,21 @@
         .from(0).to(1)
     end
   end
+
+  describe "#number_with_country_code" do
+    it "adds the country code when one is not present" do
```
rsanheim, line 108: Can you add some more test cases here ? What happens when its some sort of invalid phone number, like not enough digits or too many?

app/models/patient_phone_number.rb
rsanheim: I agree we should move this to the model eventually, but I think there is enough bad data out there w/ the phone numbers to make that premature right now.

spec/models/patient_phone_number_spec.rb
kpethtel: I added more tests. In the process, I found some behavior in the gem I don't totally understand regarding how it strips the raw national number. I'm guessing it may have something to do with what constitutes a "valid" number. If the abbreviation is set to US and you have a 9 digit number that starts with 1, it returns the number you gave it. With numbers of other lengths, it strips the leading 1. So "111-111-111" and "111-111-1111" both return nine ones. I don't think it's worth digging into further. The truth is the current process for processing phone numbers is imperfect and we're not going to fix it today.

rsanheim (approved):
Thanks for adding the specs

PR #2736, opened 2021-07-20
Add a metric to track number of online lookups

ssrihari:
**Story card:** [ch4211](https://app.clubhouse.io/simpledotorg/story/4211/create-a-dashboard-to-monitor-metrics-relating-to-performance-of-the-online-lookup-feature)

- Increment a metric for every patient looked up
- Add the retention type in the metric name
- Add state and user-id in tags as metadata

### DIFF
--- a/app/controllers/api/v4/patients_controller.rb
+++ b/app/controllers/api/v4/patients_controller.rb
@@ -17,8 +17,8 @@ class Api::V4::PatientsController < APIController
       json: Oj.dump({
         patients: patients.map do |patient|
           retention = retention(patient)
-          Api::V4::PatientLookupTransformer.to_response(patient, retention)
           Statsd.instance.increment("OnlineLookup.#{retention[:type]}", tags: [current_state.name, current_user.id])
+          Api::V4::PatientLookupTransformer.to_response(patient, retention)
         end
       }, mode: :compat),
       status: :ok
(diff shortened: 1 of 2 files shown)


app/controllers/api/v4/patients_controller.rb
```diff
@@ -15,7 +15,11 @@ def lookup
     trigger_audit_log(patients)
     render(
       json: Oj.dump({
-        patients: patients.map { |patient| Api::V4::PatientLookupTransformer.to_response(patient, retention(patient)) }
+        patients: patients.map do |patient|
+          retention = retention(patient)
+          Api::V4::PatientLookupTransformer.to_response(patient, retention)
```
harimohanraj89, line 20: We have to return this from the block.

ssrihari: Fixed in bcf610e 🤦

harimohanraj89: Classic 😅

app/controllers/api/v4/patients_controller.rb
```diff
@@ -15,7 +15,11 @@ def lookup
     trigger_audit_log(patients)
     render(
       json: Oj.dump({
-        patients: patients.map { |patient| Api::V4::PatientLookupTransformer.to_response(patient, retention(patient)) }
+        patients: patients.map do |patient|
+          retention = retention(patient)
+          Statsd.instance.increment("OnlineLookup.#{retention[:type]}", tags: [current_state.name, current_user.id])
```
harimohanraj89, line 20: I suspect `current_state` and `current_user` are guaranteed to be present for an authenticated request? If so, :100:

harimohanraj89 (approved):
Sweet sweet metrics.

PR #2738, opened 2021-07-20
Fix ordering in Recent BPs log

prabhanshuguptagit:
**Story card:** [ch4219](https://app.clubhouse.io/simpledotorg/story/4219)

## Because

Recent BP log ordering breaks for BPs added late at night in india timezone.
image

## This addresses

The ordering of recent BPs happens first by date (descending) and then by time of day (ascending). The date calculation in the ordering clause needs to be aware of the reporting time zone. Currently, a BP recorded at 12:30AM IST -> 7:30AM IST gets categorised into the previous day.

This makes the `DATE` ordering time zone aware. It also refactors the clause into a scope in `BloodPressure` since it's called by multiple models.


app/models/blood_pressure.rb
```diff
@@ -64,4 +64,8 @@ def anonymized_data
      bp_systolic: systolic,
      bp_diastolic: diastolic}
   end
+
+  def self.for_recent_bp_log
+    order(Arel.sql("DATE(recorded_at at time zone 'utc' at time zone '#{CountryConfig.current[:time_zone]}') DESC, recorded_at ASC"))
+  end
```
harimohanraj89, line 70: Such a niche use case, I don't think the model's the right place for this. It's not a very widely used concept to order the BPs this way. Can we extract this to a query object?

```
RecentBPLogQuery.for(user: @user)

class RecentBPLogQuery
  def self.for(scope: true)
    BloodPressure.where(scope).order(the_fancy_stuff)
  end
end
```

We should also add some tight unit test coverage around the timezone boundaries and ordering for this, regardless of where it lives.

harimohanraj89 (changes requested):
An organizational note, but also we should definitely add test coverage for this.

app/models/blood_pressure.rb
prabhanshuguptagit: That sounds fantastic👌🏼

prabhanshuguptagit: On second thought, a query object doesn't suit this use case since it just orders BPs more or less. Moved it to a view helper in 32e8df5

harimohanraj89:
That extraction seems fine to me. We should still add very precise test coverage.

app/helpers/reports/regions_helper.rb
```diff
@@ -36,4 +36,9 @@ def percentage_or_na(value, options)
   def cohort_report_type(period)
     "#{period.type.to_s.humanize}ly report"
   end
+
+  def recent_bp_log(blood_pressures)
+    blood_pressures
+      .order(Arel.sql("DATE(recorded_at at time zone 'utc' at time zone '#{CountryConfig.current[:time_zone]}') DESC, recorded_at ASC"))
```
ssrihari, line 42: Why do we need to cast timezones to sort? Also, why are we doing a date desc sort, and then again an asc sort?

ssrihari: Understood the desc date, asc time from the spec.

spec/helpers/regions_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        bp_2 = create(:blood_pressure, recorded_at: 10.minutes.ago)
+
+        expect(recent_bp_log(BloodPressure.all)).to eq([bp_1, bp_2])
+      end
+    end
+
+    context "respects the reporting timezone for ordering" do
+      it "BPs in different days in reporting time zone but same day in UTC are ordered descending" do
+        bp_1 = create(:blood_pressure, recorded_at: Time.parse("1 July 2021 11:30PM IST"))
+        bp_2 = create(:blood_pressure, recorded_at: Time.parse("2 July 2021 12:30AM IST"))
```
rsanheim, line 26: I think you probably want to use `Time.zone.parse` here, otherwise I'm not sure the times you create will actually respect the time zones you are providing.

app/helpers/reports/regions_helper.rb
harimohanraj89: @ssrihari it was very validating to know that someone went through the exact sequence of thoughts looking at this code 😂

spec/helpers/regions_helper_spec.rb
harimohanraj89: Screen Shot 2021-07-22 at 12 39 13 PM

Looks like both work, but I'd be more comfortable dealing with the ActiveSupport times anyway.

prabhanshuguptagit: @harimohanraj89 That probably worked because your local timezone is IST. It doesn't work on a UTC instance.

harimohanraj89: That's why I used EST timestamps as well 😅  But I agree, let's use `Time.zone`

Issue #2108, opened 2021-02-28
docker-compose configuration out of date


matisk0:
Unable to run simple-server with current docker-compose config, which still based on ``` ruby:2.5.1 ```

```
$ docker-compose up
```

```
Your Ruby version is 2.5.1, but your Gemfile specified 2.6.6
ERROR: Service 'server' failed to build: The command '/bin/sh -c bundle _1.17.3_ install' returned a non-zero code: 18
```

After setting specified ruby 2.6.6 version, docker-compose setup almost finished except the following error:
```
server_1    | Seeding Facilities: |==========================================================|
server_1    | ⭐️  Seed complete! Elasped time less than a minute ⭐️
server_1    | rails aborted!
server_1    | ArgumentError: invalid uri scheme ''
server_1    | /usr/local/bundle/gems/redis-4.1.4/lib/redis/client.rb:439:in `_parse_options'
server_1    | /usr/local/bundle/gems/redis-4.1.4/lib/redis/client.rb:83:in `initialize'
server_1    | /usr/local/bundle/gems/redis-4.1.4/lib/redis.rb:44:in `new'
server_1    | /usr/local/bundle/gems/redis-4.1.4/lib/redis.rb:44:in `initialize'

server_1    | Tasks: TOP => db:refresh_materialized_views
server_1    | (See full trace by running task with --trace)
```

Seems like REDIS_URL isn't set properly or missing.

benryder1988:
Thanks for raising the issue - I have some time this week and will look into it!

benryder1988:
This should now have been fixed in the following pull request:
https://github.com/simpledotorg/simple-server/pull/2739

rsanheim:
Fixed in #2739. Thanks!

PR #2739, opened 2021-07-21
Updating Docker configs for issue #2108

benryder1988:
**Story card:** https://github.com/simpledotorg/simple-server/issues/2108

## Because

As mentioned in the issue, the Dockerfile was out out date with the required Ruby version, and when the Ruby version was updated there was a Redis error.

Issue details: https://github.com/simpledotorg/simple-server/issues/2108

## This addresses

The Ruby version was updated in the Dockerfile, and the Redis bug fixed by updating the Docker Compose config. The README was updated with details on removing the volumes used by docker compose, and how to test with the Android App. 

## Test instructions

- Run the server with 'docker compose up'
- Capture the ngrok url from the logs
- Configure the Android App to use the ngrok URL
- Build and test the app

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,4 @@
-FROM ruby:2.5.1
-
-RUN curl https://deb.nodesource.com/setup_12.x | bash
-RUN curl https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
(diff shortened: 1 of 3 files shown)


rsanheim:
Thanks for this! I ran verified it all works w/ my docker setup here.

PR #2740, opened 2021-07-21
3586: set up production ab test

kpethtel:
**Story card:** [ch3586](https://app.clubhouse.io/simpledotorg/story/3586/test-in-sandbox)

## Because

We want to perform a test of the AB framework in production with real phone numbers to ensure end-to-end functionality.

## Implementation
- Ensure that `experiment` and `whatsapp_appointment_reminders` feature flags are on
- Find or create a User to enter into the seed script. I didn't add this to the script because I didn't want to create a test facility or add a user to a random facility. 
- Run seed script. Days till appointment should be at least 1 to allow for time to send notifications a day before the appointment. 
- Update patient phone numbers.
- Start experiment: ExperimentControlService.start_current_patient_experiment("production test", 0, 10)
  - the 0 and 10 shouldn't matter much because they're normally used to select patients. We just need to ensure that the experiment is active when we want messages to send. 
- assuming that all of this is done on Thursday before the end of the messaging window (4pm), we should be able to run the data setup script with `days_till_appointment` of 1, which will create the appointments for Friday. We'll already be past the cron job that schedules notifications for the day, but we can still send out notifications that day by running it manually:  `AppointmentNotification::ScheduleExperimentReminders.perfom_now`.


config/schedule.rb
```diff
@@ -50,7 +50,7 @@ def local(time)
 end
 
 every :day, at: local("07:00 am"), roles: [:cron] do
-  rake "appointment_notification:covid_medication_reminders[20000]"
```
kpethtel, line 53: This needs to be turned off regardless.

db/data/20210721181311_create_test_experiment.rb
```diff
@@ -0,0 +1,31 @@
...
+      control_group = experiment.treatment_groups.create!(description: "control")
+      treatment_group = experiment.treatment_groups.create!(description: "professional_request")
+      treatment_group.reminder_templates.create!(message: "notifications.set01.professional_request", remind_on_in_days: -1)
+      treatment_group.reminder_templates.create!(message: "notifications.set02.professional_request", remind_on_in_days: 0)
+      treatment_group.reminder_templates.create!(message: "notifications.set03.professional_request", remind_on_in_days: 1)
+
+      now = Time.current
+      user = User.last # we should find a user we can use or create a new one
+      patient = Patient.create!(id: SecureRandom.uuid, full_name: "Hari AB Tester", status: "active", device_created_at: now, device_updated_at: now, registration_user: user)
+      patient.phone_numbers.create!(id: SecureRandom.uuid, number: "this should not be in a public git", phone_type: "mobile", active: true, device_created_at: now, device_updated_at: now)
```
kpethtel, line 14: I don't want to add their phone numbers to git, so that may need to be a manual step. It would also be nice to have more than four users. It would be really sad if all four patients were put in the control group by chance.

db/data/20210721181311_create_test_experiment.rb
```diff
@@ -0,0 +1,31 @@
...
+      patient = Patient.create!(id: SecureRandom.uuid, full_name: "Srihari AB Tester", status: "active", device_created_at: now, device_updated_at: now, registration_user: user)
+      patient.phone_numbers.create!(id: SecureRandom.uuid, number: "this should not be in a public git", phone_type: "mobile", active: true, device_created_at: now, device_updated_at: now)
+      patient = Patient.create!(id: SecureRandom.uuid, full_name: "Pragati AB Tester", status: "active", device_created_at: now, device_updated_at: now, registration_user: user)
+      patient.phone_numbers.create!(id: SecureRandom.uuid, number: "this should not be in a public git", phone_type: "mobile", active: true, device_created_at: now, device_updated_at: now)
+
+      # create appointments during experiment window
+
+      ExperimentControlService.start_current_patient_experiment(experiment.name, 1, 5)
```
kpethtel, line 24: This might also be best as a manual step. It could even be added to `schedule.rb` for a specific date.

db/data/20210721181311_create_test_experiment.rb
```diff
@@ -0,0 +1,31 @@
...
+      experiment = Experimentation::Experiment.create!(name: "production test", experiment_type: "current_patients", state: "new")
+      control_group = experiment.treatment_groups.create!(description: "control")
+      treatment_group = experiment.treatment_groups.create!(description: "professional_request")
+      treatment_group.reminder_templates.create!(message: "notifications.set01.professional_request", remind_on_in_days: -1)
+      treatment_group.reminder_templates.create!(message: "notifications.set02.professional_request", remind_on_in_days: 0)
+      treatment_group.reminder_templates.create!(message: "notifications.set03.professional_request", remind_on_in_days: 1)
+
+      now = Time.current
+      user = User.last # we should find a user we can use or create a new one
+      patient = Patient.create!(id: SecureRandom.uuid, full_name: "Hari AB Tester", status: "active", device_created_at: now, device_updated_at: now, registration_user: user)
```
kpethtel, line 13: I'm not sure why some of our models don't automatically create uuids on save. I should look into it.

config/schedule.rb
```diff
@@ -50,7 +50,7 @@ def local(time)
...
+  AppointmentNotification::ScheduleExperimentReminders.perfom_now
```
rsanheim, line 53: I don't believe this will work, because this gets written to a plain old cron job via [whenever](https://github.com/javan/whenever).  You need to use `runner` to invoke Ruby like this.

db/data/20210721181311_create_test_experiment.rb
rsanheim: Huh, I didn't realize that was the case - I bet its because patients are often created on devices where they will already have a uuid created there.

db/data/20210721181311_create_test_experiment.rb
rsanheim: How about just put in a more-realistic fake number as a constant and wire that in as a phone number? Then whoever runs things can wire phone numbers manually.

db/data/20210721181311_create_test_experiment.rb
```diff
@@ -0,0 +1,31 @@
+class CreateTestExperiment < ActiveRecord::Migration[5.2]
+  def up
+    if CountryConfig.current_country?("India")
+      experiment = Experimentation::Experiment.create!(name: "production test", experiment_type: "current_patients", state: "new")
```
rsanheim, line 4: It may be worth pulling this out to a service object and calling it from this data migration -- that way you can easily run it locally and also add some tests to verify things.

kpethtel: That's probably a better idea, definitely easier to work with that.

config/schedule.rb
kpethtel: I'll make that change.

app/models/appointment.rb
```diff
@@ -40,6 +40,7 @@ class Appointment < ApplicationRecord
   validate :cancel_reason_is_present_if_cancelled
   validates :device_created_at, presence: true
   validates :device_updated_at, presence: true
+  validates :appointment_type, presence: true
```
kpethtel, line 43: I added this because I got database errors instead of validation errors when I tried saving an appointment.

app/services/experiment_control_service.rb
```diff
@@ -11,7 +11,12 @@ def start_current_patient_experiment(name, days_til_start, days_til_end, percent
 
       experiment.update!(state: "selecting", start_date: experiment_start.to_date, end_date: experiment_end.to_date)
 
-      eligible_ids = current_patient_candidates(experiment_start, experiment_end).shuffle!
+      eligible_ids = if name == "production test"
```
rsanheim, line 14: `production test` feels like it should be a constant.

lib/seed/production_experiment_test_seeder.rb
```diff
@@ -0,0 +1,28 @@
...
+    def self.call(user_id:, days_till_appointment: 1)
+      ExperimentSeeder.call(experiment_name: "production test", experiment_type: "current_patients")
+
+      now = Time.current
+      today = Date.current
+      user = User.find(user_id)
+      facility = user.registration_facility
+
+      patient = Patient.create!(id: SecureRandom.uuid, full_name: "Hari AB Tester", status: "active", device_created_at: now, device_updated_at: now, registration_user: user)
+      patient.phone_numbers.create!(id: SecureRandom.uuid, number: "8675309", phone_type: "mobile", active: true, device_created_at: now, device_updated_at: now)
```
rsanheim, line 12: constant for the fake number? In fact, we could probably just have an array of the `full_names` and loop thru them, it looks like everything is the same here...and then Nilenso can do the phone number updates manually afterwards.

rsanheim (approved):
Some feedback, but approving to unblock.

db/data/20210721181311_create_test_experiment.rb
harimohanraj89: Yup precisely. I don't know if we explored allowing the server to assign the UUID automatically if missing, though. Looks like we've taken the extreme approach of not even touching the thing.

app/services/experiment_control_service.rb
kpethtel: Fair. I was hoping this would only be temporary code, but we could leave it in as a long term testing option.

lib/seed/production_experiment_test_seeder.rb
kpethtel: That's also a valid point. I'll make that change.

lib/seed/production_experiment_test_seeder.rb
```diff
@@ -0,0 +1,24 @@
+module Seed
+  class ProductionExperimentTestSeeder
+    TEMPORARY_PHONE_NUMBER = "8675309"
```
rsanheim, line 3:
```suggestion
    TEMPORARY_PHONE_NUMBER = "0001234"
```

Might as well make sure this is really an invalid number, just in case.

rsanheim (approved):
LGTM, one minor thing re: fake phone number

lib/seed/production_experiment_test_seeder.rb
kpethtel: Done.

PR #2742, opened 2021-07-22
Empty seed type and better blocks

harimohanraj89:
**Story card:** [ch4205](https://app.clubhouse.io/simpledotorg/story/4205/optionally-disable-patient-data-generation-in-the-seed-generation-script)

## Because

The mobile team would like to spin up and spin down server instance rapidly to back their automated test suites. Re-using the QA environment over and over again is not great ([see epic](https://app.clubhouse.io/simpledotorg/epic/4222/isolate-android-integration-tests)).

## This addresses

Two big enhancements to our seeds to support mobile tests:
* Adds an `empty` seed type that creates facilities, but doesn't populate any patient data. The number of facilities, etc, matches the `small` seed type.
* Creates more than one block per facility group. Taking a look at the old code, it seems like we intended to create several blocks per facility group, but left those mechanisms unused. We ended up creating only 1 block per FG, and the mobile tests need at least one FG with two blocks.

## Test instructions

Empty seed type
* Reset your local DB
* Run `SEED_TYPE=empty rails db:seed`
* Facilities should be created, patient data should not

Environment-based check
* Reset your local DB
* Set `SIMPLE_SERVER_ENV=android_review`
* Run `rails db:seed`
* Facilities should be created, patient data should not

Block check
* Reset your local DB
* Run `rails db:seed`
* Facilities should be created, patient data should be created, most if not all facility groups should have more than one block


lib/seed/config.rb
```diff
@@ -4,6 +4,8 @@ class Config
     def initialize
       @type = ENV["SEED_TYPE"] ||
         case SimpleServer.env
+        when "android_review"
```
harimohanraj89, line 7: This is the `SIMPLE_SERVER_ENV` that the mobile test suite will set when it spins up a test server.

lib/seed/facility_seeder.rb
```diff
@@ -74,9 +71,8 @@ def weighted_facility_size_sample
       FACILITY_SIZE_WEIGHTS.max_by { |_, weight| rand**(1.0 / weight) }.first
     end
 
-    def facilities_per_block
-      max_number_facilities_per_block = (number_of_facilities_per_facility_group / MAX_FACILITIES_PER_BLOCK).ceil
-      config.rand_or_max(1..max_number_facilities_per_block)
+    def max_number_of_blocks_per_facility_group
+      2
```
harimohanraj89, line 75: I've put this into a method since it's likely a parameter we'll want to get fancy with soon. Let me know if you'd like me to just shove it back into the method below instead.

lib/seed/facility_seeder.rb
```diff
@@ -123,17 +119,25 @@ def create_district_regions(state_results, facility_group_results)
     end
 
     def create_block_regions(district_region_results)
-      block_count = district_region_results.ids.size
+      # Eagerly fetch block names to avoid duplicates
```
harimohanraj89, line 122: @rsanheim Is this reason accurate?

lib/seed/config.rb
prabhanshuguptagit: Will this be set somewhere in `app.json`?

lib/seed/facility_seeder.rb
```diff
@@ -74,9 +71,8 @@ def weighted_facility_size_sample
       FACILITY_SIZE_WEIGHTS.max_by { |_, weight| rand**(1.0 / weight) }.first
     end
 
-    def facilities_per_block
-      max_number_facilities_per_block = (number_of_facilities_per_facility_group / MAX_FACILITIES_PER_BLOCK).ceil
-      config.rand_or_max(1..max_number_facilities_per_block)
```
prabhanshuguptagit, line 79: Was this method unused?

lib/seed/facility_seeder.rb
prabhanshuguptagit: We should be able to move the sampling in the loop. `Region` can support it now, although there might be a higher rate of blocks with same names I suppose.

lib/seed/facility_seeder.rb
```diff
@@ -123,17 +119,25 @@ def create_district_regions(state_results, facility_group_results)
     end
 
     def create_block_regions(district_region_results)
-      block_count = district_region_results.ids.size
+      # Eagerly fetch block names to avoid duplicates
+      block_count = district_region_results.ids.size * max_number_of_blocks_per_facility_group
       block_names = Seed::FakeNames.instance.blocks.sample(block_count)
-      block_regions = district_region_results.results.each_with_index.map { |row, i|
+
+      block_counter = 0
```
prabhanshuguptagit, line 126: This took a while to understand 😅 , moving the sampling in should simplify (at the expense of fully distinct names).

prabhanshuguptagit (approved):
👌🏼 👌🏼

lib/seed/facility_seeder.rb
harimohanraj89: Yup, it was unused.

lib/seed/config.rb
harimohanraj89: It's being set manually with a `heroku config:set` CLI command in the Github action steps. I toyed with an idea to introduce a `.env.android-review` dotenv file, but didn't have luck with it yet.

lib/seed/facility_seeder.rb
harimohanraj89: Yup indeed. I'll do a test with a handful of block names and tons of duplication, and see if our reports still hold up.

harimohanraj89: Ah I've found the problem. It's not blocks with the same name, rather - the ActiveRecord import that we do isn't smart enough to assign unique slugs to blocks with the same name. I'm going to leave this as is.

prabhanshuguptagit: This is where the the slug generation happens, it should work out as long as they have a different uuid 🤔  https://github.com/simpledotorg/simple-server/blob/71db34c7219508ae4422db530d5e3b60874de457/app/models/region.rb#L89

lib/seed/config.rb
prabhanshuguptagit: And `heroku config:set` overrides `app.json` I am guessing.

lib/seed/facility_seeder.rb
harimohanraj89: Nah, ActiveRecord import bypasses these application-level hooks.

lib/seed/facility_seeder.rb
harimohanraj89: Closing the loop on this, the problem is not app code, but the bulk import. The bulk import bypasses app-level hooks, so two blocks with the same name dont' get assigned unique slugs, and end up clashing in the DB.

lib/seed/config.rb
harimohanraj89: Actually, `app.json` is only used for _review apps_, which spawn off of server PRs. The android repo is going to be creating its own heroku apps per _android_ PR without using review apps, so the `app.json` is completely ignored.

lib/seed/facility_seeder.rb
prabhanshuguptagit: Ah ok 👍🏼 . Non-blocking: Do you know where the slugs get generated in that case?

harimohanraj89: That's a good point. Clearly activerecord import does some weird middle ground where slug generation is done, but uniqueness isn't checked? Might make sense, right? The slug candidates are chosen based on records already persisted in the DB. When you're importing, nothing's persisted yet. 🤷 

It's weird enough that I'm gonna leave it alone for now 😅

PR #2743, opened 2021-07-22
Check if index exists before creating it

vkrmis:
**Story card:** none

## Because

The index creation fails if the index already exists

## This addresses

Check to see if there is already an index on the facilities column before attempting creation


vkrmis:
Closing in favour of:  https://github.com/simpledotorg/simple-server/pull/2744

PR #2745, opened 2021-07-22
Prevent parallel deploys via semaphore

prabhanshuguptagit:
**Story card:** -

## Because

Our semaphore pipeline always picks up the latest master to deploy irresepective of the tag it was triggered from.

Also if multiple deploys are triggered in parallel, we can have a race condition where an older deploy can end up overwriting a newer deploy.

## This addresses

- Fixes `semaphore.yml` to pickup the git SHA instead of the branch name (which was always `master`).
- Puts the deploy pipeline in a [queue](https://docs.semaphoreci.com/essentials/pipeline-queues) so there's only one deploy running at a time.

### DIFF
--- a/.semaphore/bangladesh_demo_deployment.yml
+++ b/.semaphore/bangladesh_demo_deployment.yml
@@ -15,7 +15,7 @@ blocks:
             - yarn install
             - bundle install --deployment --path vendor/bundle
             - cache store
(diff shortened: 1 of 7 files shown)


harimohanraj89 (approved):
Nice improvements 👍

rsanheim (approved):
Nice!

PR #2747, opened 2021-07-23
Cleanup FacilityDistrict

prabhanshuguptagit:
**Story card:** [ch3805](https://app.clubhouse.io/simpledotorg/story/3805/cleanup-facilitydistrict-code)

## Because

`FacilityDistrict` was used as a virtual-model to populate district reports. All the reports now use `Region` and we don't need the `FacilityDistrict` model around anymore.

## This addresses

Removes the model and its usages. Removes the `facility_district` region reports. 

- [x] Communicate to Praveen and Dr. Ganesh about removing https://api.simple.org/reports/regions/facility_district/Chennai, they might be using them: [Slack thread](https://simpledotorg.slack.com/archives/CK95QFJA2/p1626949404023000). If they still want a top level report for Chennai they can use the `Tamil Nadu` state report until we fix that hierarchy.
- [x] Click test on performance


harimohanraj89:
That's all the touchpoints? I thought there were more for some reason, not bad.

Can we confirm that normal district reports for Bathinda and Mansa are working in IHCI prod also?

prabhanshuguptagit:
> Can we confirm that normal district reports for Bathinda and Mansa are working in IHCI prod also?

The regular region reports work, yes. We've informed Dr. Bidisha to move away from the facility district report too.

app/controllers/reports/regions_controller.rb
```diff
@@ -118,20 +118,7 @@ def download
...
-      case report_scope
-      when "facility_district"
-        scope = current_admin.accessible_facilities(:view_reports)
-        FacilityDistrict.new(name: report_params[:id], scope: scope)
-      when "district"
-        current_admin.accessible_district_regions(:view_reports).find_by!(slug: report_params[:id])
-      else
-        raise ActiveRecord::RecordNotFound, "unknown report_scope #{report_scope}"
-      end
-    }
```
harimohanraj89, line 134: With this removal, can we also stop specifying a `report_scope` parameter somewhere?

prabhanshuguptagit: We still need to pass `report_scope` everywhere for differentiating between `district`, `facility` and `blocks`.

rsanheim:
These remaining usages are the actual `district` field on the Facility model, right?

```
[~/src/simpledotorg/simple-server (cleanup-facility-district)]> ag facility_district
app/exporters/patients_exporter.rb
124:      patient_summary.assigned_facility_district,

app/exporters/patients_with_history_exporter.rb
104:      patient_summary.assigned_facility_district,

app/views/admin/facilities/_form.html.erb
24:      <%= form.text_field :district, id: :facility_district, required: true %>

db/schema.rb
771:      assigned_facility.district AS assigned_facility_district,

db/views/materialized_patient_summaries_v01.sql
28:       assigned_facility.district                                                                                                AS assigned_facility_district,
[~/src/simpledotorg/simple-server (cleanup-facility-district)]>
```

app/controllers/reports/regions_controller.rb
rsanheim: > With this removal, can we also stop specifying a `report_scope` parameter somewhere?

We could get away from using `report_scope`, but it would mean changing our URLs to use Region slug instead of source slug...as the Region slug is truly unique, and then we could do a plain old `Region.find_by(slug: params[:slug])`.  That would definitely be a bigger change, and would break any old URLs that admins may have bookmarked where the Region slug does not match the source slug.

rsanheim (approved):
Really glad to see this go! FacilityDistrict was going to add complication to all the new reporting schema stuff, so this is great. ✨

PR #2749, opened 2021-07-23
Add Ethiopia DHIS2 data element IDs

timcheadle:
**Story card:** [ch4271](https://app.clubhouse.io/simpledotorg/story/4271/configure-data-element-ids-in-simple)

## Because

To send data to Ethiopia DHIS2, we need to map our core data elements to their associated DHIS2 IDs.

## This addresses

This adds the Ethiopian data element IDs and moves everything to `CountryConfig` so we can handle multiple countries.


harimohanraj89 (approved):
:shipit:

PR #2750, opened 2021-07-26
Drug stock improvements

prabhanshuguptagit:
**Story card:** [ch4303](https://app.clubhouse.io/simpledotorg/story/4303/cosmetic-improvements)

## Because

We got a few additional requests during a demo, mostly UI changes 

## This addresses

- [x] Default calculation to 0 for received
- [x] When filtering don’t show block rows, keep warehouse row
- [x] Order columns as CCB, ARB, Diuretic
- [x] Bug to fix: filters are not being preserved when another filter is selected
- [x] Bug to fix: Top level total rows (`All`) shouldn't be affected by filters


.env.development
```diff
@@ -5,6 +5,7 @@ APPOINTMENT_NOTIFICATION_HOUR_OF_DAY_START=14
 CALL_SESSION_REDIS_HOST=localhost
 CALL_SESSION_REDIS_POOL_SIZE=12
 CALL_SESSION_REDIS_TIMEOUT_SEC=1
+CUSTOM_DRUG_CATEGORY_ORDER=["hypertension_ccb", "hypertension_arb", "hypertension_diuretic"]
```
prabhanshuguptagit, line 8: We can also do `hypertension_ccb,hypertension_arb,hypertension_diuretic` and split(",") in code.

app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -92,6 +91,14 @@ def set_region
       end
   end
 
+  def blocks_to_display
+    if @selected_facility_sizes != @facility_sizes || @selected_zones != @zones
+      Region.none
+    else
+      @query.blocks.order(:name)
+    end
```
harimohanraj89, line 99: We could flip this conditional, it would read a bit better, I think.

`if sizes same and zones same, blocks.order(stuff)`

.env.development
vkrmis: Put this in `CountryConfig` instead

harimohanraj89 (approved):
Not much to add to this. Godspeed.

PR #2751, opened 2021-07-26
Make assigned_patients private

rsanheim:
Trying to reduce the surface area of Repository a bit, which will make it easier to move the public API to the new pipeline.

breaking this from #2734


kpethtel (approved):
I do like a good public interface.

PR #2753, opened 2021-07-27
Unit tests for facility states

ssrihari:


spec/models/reports/facility_state_spec.rb
```diff
@@ -5,16 +5,8 @@
...
   around do |example|
-    # We need to enforce a known time for this test, otherwise we will have intermittent failures. For example,
-    # if we use live system time, many of these specs will fail after 18:30 UTC (ie 14:30 ET) when on the last day of a month,
-    # because that falls into the next day in IST (our reporting time zone). So to prevent confusing failures for
-    # developers or CI during North American afternoons, we freeze to a time that will be the end of the month for
-    # UTC, ET, and IST. Timezones! 🤯
-    Timecop.freeze("June 30 2021 23:00 IST") do
-      example.run
-    end
+    freeze_time(example)
```
ssrihari, line 9: can we think of a better name?

spec/models/reports/facility_state_spec.rb
```diff
@@ -67,4 +59,101 @@
...
+      with_reporting_time_zone do
+        facility_state_june_2021 = described_class.find_by(facility: facility, month_date: june_2021[:now])
+
+        expect(facility_state_june_2021.missed_visit_lost_to_follow_up).to eq 1
+        expect(facility_state_june_2021.visited_no_bp_lost_to_follow_up).to eq 1
+        expect(facility_state_june_2021.patients_under_care).to eq 0
+        expect(facility_state_june_2021.patients_lost_to_follow_up).to eq 2
+      end
+    end
+  end
```
ssrihari, line 158: These tests only assert on the totalling functionality, and the exhaustive reporting tests continue to be in repository. We could start moving some of those tests here, potentially.

spec/models/reports/facility_state_spec.rb
```diff
@@ -67,4 +59,103 @@
...
+        ltfu_patient = create(:patient, recorded_at: june_2021[:over_12_months_ago], assigned_facility: facility)
+        create(:blood_pressure, patient: ltfu_patient, recorded_at: june_2021[:over_12_months_ago])
+        _different_facility_patient = create(:patient, recorded_at: june_2021[:long_ago])
+
+        RefreshMaterializedViews.new.refresh_v2
+        with_reporting_time_zone do
+          expect(described_class.find_by(month_date: june_2021[:now], facility: facility).under_care).to eq 1
+          expect(described_class.find_by(month_date: june_2021[:now], facility: facility).lost_to_follow_up).to eq 1
+          expect(described_class.find_by(month_date: june_2021[:now], facility: facility).dead).to eq 1
+          expect(described_class.find_by(month_date: june_2021[:now], facility: facility).cumulative_assigned_patients).to eq 3
```
ssrihari, line 78: @harimohanraj89 had flagged this earlier, should this count exclude dead patients @rsanheim?

spec/models/reports/facility_state_spec.rb
```diff
@@ -67,4 +59,103 @@
...
+      RefreshMaterializedViews.new.refresh_v2
+      with_reporting_time_zone do
+        facility_state_june_2021 = described_class.find_by(facility: facility, month_date: june_2021[:now])
+
+        expect(facility_state_june_2021.controlled_under_care).to eq 1
+        expect(facility_state_june_2021.uncontrolled_under_care).to eq 1
+        expect(facility_state_june_2021.missed_visit_under_care).to eq 1
+        expect(facility_state_june_2021.visited_no_bp_under_care).to eq 1
+        expect(facility_state_june_2021.patients_under_care).to eq 4
+        expect(facility_state_june_2021.patients_lost_to_follow_up).to eq 0
```
ssrihari, line 132: We should probably have better names for these 2 fields. We currently have
- `under_care`: number of assigned patients by care state
- `patients_under_care`: **adjusted** number of assigned patients by care state

Should we perhaps prefix all the adjusted fields with `adjusted_`?

spec/models/reports/facility_state_spec.rb
ssrihari: I've used `freeze_time_for_reporting_specs` for now, but still not convinced that's good.

spec/models/reports/facility_state_spec.rb
```diff
@@ -67,4 +59,103 @@
...
+        _under_care_patient = create(:patient, recorded_at: june_2021[:under_12_months_ago], assigned_facility: facility)
+        ltfu_patient = create(:patient, recorded_at: june_2021[:over_12_months_ago], assigned_facility: facility)
+        create(:blood_pressure, patient: ltfu_patient, recorded_at: june_2021[:over_12_months_ago])
+        _different_facility_patient = create(:patient, recorded_at: june_2021[:long_ago])
+
+        RefreshMaterializedViews.new.refresh_v2
+        with_reporting_time_zone do
+          expect(described_class.find_by(month_date: june_2021[:now], facility: facility).under_care).to eq 1
+          expect(described_class.find_by(month_date: june_2021[:now], facility: facility).lost_to_follow_up).to eq 1
+          expect(described_class.find_by(month_date: june_2021[:now], facility: facility).dead).to eq 1
```
harimohanraj89, line 77: Can we make 1, 2, and 3, of these patients respectively? Keeping everything as `1` increases the risk of false positives, what do you think?

spec/models/reports/facility_state_spec.rb
```diff
@@ -67,4 +59,103 @@
...
+
+  context "assigned patients by care states" do
+    describe "under_care, lost_to_follow_up, dead" do
+      it "computes the number of assigned patients correctly with each care state" do
+        facility = create(:facility)
+        _dead_patient = create(:patient, recorded_at: june_2021[:long_ago], assigned_facility: facility, status: "dead")
+        _under_care_patient = create(:patient, recorded_at: june_2021[:under_12_months_ago], assigned_facility: facility)
+        ltfu_patient = create(:patient, recorded_at: june_2021[:over_12_months_ago], assigned_facility: facility)
+        create(:blood_pressure, patient: ltfu_patient, recorded_at: june_2021[:over_12_months_ago])
+        _different_facility_patient = create(:patient, recorded_at: june_2021[:long_ago])
```
harimohanraj89, line 71: Can we make a deleted patient also, ensure it's excluded?

spec/models/reports/facility_state_spec.rb
harimohanraj89: Reposting my earlier concern: In our dashboard and terminology, "assigned" means "under care" + "LTFU". In some places of our reporting pipeline "assigned" means "under care" + "LTFU" + "dead". Let's nip this confusion in the bud.

spec/models/reports/facility_state_spec.rb
```diff
@@ -67,4 +59,103 @@
...
+      create(:blood_pressure, patient: patient_visited_no_bp, recorded_at: june_2021[:over_3_months_ago])
+
+      RefreshMaterializedViews.new.refresh_v2
+      with_reporting_time_zone do
+        facility_state_june_2021 = described_class.find_by(facility: facility, month_date: june_2021[:now])
+
+        expect(facility_state_june_2021.controlled_under_care).to eq 1
+        expect(facility_state_june_2021.uncontrolled_under_care).to eq 1
+        expect(facility_state_june_2021.missed_visit_under_care).to eq 1
+        expect(facility_state_june_2021.visited_no_bp_under_care).to eq 1
```
harimohanraj89, line 130: Same concern as above. Can we stagger the number of patients in each bucket?

spec/models/reports/facility_state_spec.rb
harimohanraj89: PS we can remove this assertion in favor of the `cumulative_assigned_patients` specs below, if you'd like.

spec/models/reports/facility_state_spec.rb
harimohanraj89: We definitely should. I'm not a big fan of the `adjusted` term. Ideally, I'd love to use a term like that means "eligible for control", or "without new people", or "controllable", or something. If we can't find the right word for that, then `adjusted` it is.

spec/models/reports/facility_state_spec.rb
```diff
@@ -67,4 +59,103 @@
...
+        expect(facility_state_june_2021.patients_under_care).to eq 4
+        expect(facility_state_june_2021.patients_lost_to_follow_up).to eq 0
+      end
+    end
+
+    it "computes totals for patients lost to follow up" do
```
harimohanraj89, line 136: I noticed that the assigned and LTFU specs are sort of timeline specs, whereas control and care state are not. For good coverage, can we add a single timeline spec that, perhaps, tracks one patient over time in a facility going from ineligible for control (recent regn), to controlled, to missed visit, to LTFU?

harimohanraj89 (changes requested):

spec/models/reports/facility_state_spec.rb
rsanheim: 👍

spec/models/reports/facility_state_spec.rb
rsanheim: Can we also add a patient who was registered over 12 months ago and never came in for a BP?  They would also be LTFU.

spec/models/reports/facility_state_spec.rb
rsanheim: Yep we should exclude `dead` - they are always excluded from any part of the reports where we should cumulative assigned:

image

spec/models/reports/facility_state_spec.rb
rsanheim: I think we should stick with `adjusted_` for now because it is the term we use all throughout the code base currently. Introducing a new term for this in the pipeline _without_ changing it everywhere in the code will lead to lots of confusion.

I'm open to any suggestions for better terms going forward, but we if we do come up with one we should do that rename across the app and the DB to avoid fragmentation.  FWIW, @claudiovallejo and I tried a few times to come up with a better term for this, and adjusted was the best we could come up with 🤷‍♂️ . We don't have a succinct term for it in the actual reports UI, which I think reflects how hard naming this is.

spec/models/reports/facility_state_spec.rb
```diff
@@ -67,4 +59,103 @@
       end
     end
   end
+
+  context "assigned patients by care states" do
+    describe "under_care, lost_to_follow_up, dead" do
+      it "computes the number of assigned patients correctly with each care state" do
+        facility = create(:facility)
+        _dead_patient = create(:patient, recorded_at: june_2021[:long_ago], assigned_facility: facility, status: "dead")
```
rsanheim, line 67: Mind adding a patient or two who is not diagnosed with HTN, just to be explicit and show that this matview only selects for hypertension?

spec/models/reports/facility_state_spec.rb
```diff
@@ -67,4 +59,103 @@
...
+      facility = create(:facility)
+
+      patient_controlled = create(:patient, assigned_facility: facility, recorded_at: june_2021[:long_ago])
+      create(:bp_with_encounter, :under_control, patient: patient_controlled, recorded_at: june_2021[:now] - 1.month)
+
+      patient_uncontrolled = create(:patient, assigned_facility: facility, recorded_at: june_2021[:long_ago])
+      create(:bp_with_encounter, :hypertensive, patient: patient_uncontrolled, recorded_at: june_2021[:now] - 1.months)
+
+      patient_missed_visit = create(:patient, assigned_facility: facility, recorded_at: june_2021[:long_ago])
+      create(:bp_with_encounter, patient: patient_missed_visit, recorded_at: june_2021[:over_3_months_ago])
```
rsanheim, line 113: Want to add another missed visit patient, one who had no BPs or visits done at all?

spec/models/reports/facility_state_spec.rb
ssrihari: Thanks for the suggestion, Rob. Added in 44db5ab22bc2ae9c15747ab6c66ca13bae0f3689.

spec/models/reports/facility_state_spec.rb
ssrihari: @harimohanraj89 I've added deleted patient/visit specs in `patient_states` and `patient_visits` tables with 7ab0ca193b173c097c2bf1e4d7dc223b2bccad46. At the facility state level the understanding is that we're not dealing with deleted patients at all, and there's nothing in the factility states matview query that deals with removing deleted patients. Do you think the test here would still be valuable?

ssrihari: @rsanheim The LTFU definitions are at the `patient_state` level, and are tested there: https://github.com/simpledotorg/simple-server/blob/master/spec/models/reports/patient_state_spec.rb#L62. I was mostly testing the counting/aggregation in facility states, which is its responsibility. Wdyt?

spec/models/reports/facility_state_spec.rb
ssrihari: Thanks for prodding this @harimohanraj89 🙇 Done in bc96f9f31bad66f3aead8dd904610c61192a47b7 and d45f60e5b808eece609704a01ebddb336a09716e.

spec/models/reports/facility_state_spec.rb
ssrihari: Done in 1735b493c47329439f21244bac76598aad456ebb. Thanks!

spec/models/reports/facility_state_spec.rb
ssrihari: Thanks @harimohanraj89 🙇 Done in bc96f9f31bad66f3aead8dd904610c61192a47b7 and d45f60e5b808eece609704a01ebddb336a09716e.

spec/models/reports/facility_state_spec.rb
ssrihari: Can we continue this conversation with https://github.com/simpledotorg/simple-server/pull/2754/files#r677140960? There's some more context there on naming, and I'm doing some renames there anyway.

harimohanraj89: Good to know we've thought about this loads already. Cool, in the absence of a better succinct term that we can propagate everywhere, let's stick to `adjusted` for now.

spec/models/reports/facility_state_spec.rb
ssrihari: Only the `cumulative_assigned_patients` would be a bit of a timeline spec since that's necessary to assert the "cumulative" aspect of it. The definition of care states is in patient_states, and we've already unit tested the care state changing through time there ya?

spec/models/reports/facility_state_spec.rb
harimohanraj89: Ah good point, we've got good coverage of deletion in the patient state specs. Hmm, I could go either way. On one hand, adding a delete patient here makes this spec independent of the underlying implementation of facility states. On the other, relying on the patient state specs means we're assuming that facility states will always use patient states. I think that's a strong enough assumption that we can avoid adding a deleted patient here.

spec/models/reports/facility_state_spec.rb
harimohanraj89: Oh I see now. Are we confident enough that we'll be grabbing the correct month's data from patient states and putting them in the correct month in this table?

spec/models/reports/facility_state_spec.rb
ssrihari: Thanks folks! Fixed this with 9084c509b.

harimohanraj89: Do we need the total number (under care + LTFU + dead) for anything? If so, are we ok with needing to add assigned + dead to get it?

spec/models/reports/facility_state_spec.rb
harimohanraj89: And yes, let's continue this conversation in the other PR 👍

harimohanraj89:
[One last question](https://github.com/simpledotorg/simple-server/pull/2753#discussion_r678020311) for me

spec/models/reports/facility_state_spec.rb
ssrihari: AFAIK, we don't include dead in total assigned. Which, prompts to me think about total registrations actually. Does that need to exclude dead patients as well?

harimohanraj89: That is correct, we should **not** exclude dead patients from registration counts.

rsanheim (approved):

PR #2754, opened 2021-07-27
Add support for monthly cohort reports

ssrihari:


db/schema.rb
```diff
@@ -1516,19 +1530,28 @@
...
+      adjusted_outcomes.patients_lost_to_follow_up AS adjusted_patients_lost_to_follow_up,
+      monthly_cohort_outcomes.controlled_under_care AS monthly_cohort_controlled_under_care,
+      monthly_cohort_outcomes.uncontrolled_under_care AS monthly_cohort_uncontrolled_under_care,
+      monthly_cohort_outcomes.missed_visit_under_care AS monthly_cohort_missed_visit_under_care,
+      monthly_cohort_outcomes.visited_no_bp_under_care AS monthly_cohort_visited_no_bp_under_care,
+      monthly_cohort_outcomes.missed_visit_lost_to_follow_up AS monthly_cohort_missed_visit_lost_to_follow_up,
+      monthly_cohort_outcomes.visited_no_bp_lost_to_follow_up AS monthly_cohort_visited_no_bp_lost_to_follow_up,
+      monthly_cohort_outcomes.patients_under_care AS monthly_cohort_patients_under_care,
+      monthly_cohort_outcomes.patients_lost_to_follow_up AS monthly_cohort_patients_lost_to_follow_up
+     FROM (((((reporting_facilities rf
```
ssrihari, line 1549: @harimohanraj89 @rsanheim thoughts on the column naming here?

harimohanraj89: I'm not quite happy with the choice of `adjusted`. I understand it's consistent with our repository keys etc, but I don't think it's a great name. What if we stuck to `treatment_outcome_in_last_3_months`? If you were registered recently, you don't get a treatment outcome because you haven't been in the program long enough. Does that make more sense?

ssrihari: Related conversation: https://github.com/simpledotorg/simple-server/pull/2753/files#r677114529

ssrihari: > What if we stuck to treatment_outcome_in_last_3_months?

Note that we need a prefix for the aggregate columns like `_controlled_under_care `.

harimohanraj89: Regarding the naming "adjusted", I think we've spent a lot of time thinking and haven't come up with a better name to represent the concept of "patients who have been registered long enough to be eligible for control rates". Let's stick to "adjusted" to represent that concept.

Regarding the use of "adjusted" for these columns, we could either:
1. Use the `adjusted_` prefix to denote that these patienst don't include recent registrations, or
1. Drop the suffix and lean into the commitment that all control/uncontrol numbers have a registration exclusion baked in
1. Add a separate semantic prefix like `treament_outcome_`

Thinking through this a little bit, I think I'm a fan of (1) because:
* The `htn_care_state` in the patient states table does _not_ do any registration exclusion. Every patient gets a care state even if they were registered this month.
* [This table](https://github.com/simpledotorg/simple-server/pull/2754/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3R1486-R1498) is adding the registration exclusion, and as devs we're fairly comfortable with the idea that `adjusted` means "excluding recent regns".

Earlier, I forgot about the `htn_care_state` that exists on patient states. And I think it would be a good idea to introduce unique and familiar naming here to make it obvious that these counts do additional regn filtering on top of htn care state.

What do you guys think?

ssrihari: I followed a very similar thought process when making this PR as well @harimohanraj89. `FacilityState` is closer piece to the dashboard than as an independent building block to be used by epis, etc. I went with (2), i.e no suffix earlier, but with the same fields for `monthly_cohort_` coming in it made sense to add it.

harimohanraj89: Very cool, that sounds groovy to me.

harimohanraj89 (approved):
The pattern looks good, and as we discussed, let's stick to `adjusted` 😅 

Took me a while to re-remember that line of thought after so many days. Thanks for your patience 😄

db/schema.rb
rsanheim: > Very cool, that sounds groovy to me.

Some things on this - apologies for late feedback.  All of this is non-blocking btw.

The `adjusted_` prefix on all the columns feels awkward here, and doesn't really match how we think of 'controlled' or 'uncontrolled' in terms of reports...whenever we report 'controlled' or 'uncontrolled', the registration adjustments is **always** assumed.  

That said, this should be an easy change to handle from the repository perspective, so I don't really have strong feelings either way.  From the application code perspective, we'll continue to use the term `controlled`, because controlled always assumes an adjusted denominator for rate calculations.  And that matches the domain from the user's perspective better as well.

I do think this will actually be more confusing to any end-users if they want to use this matview from metabase, because Praveen and team generally think in terms of controlled / uncontrolled, not "uncontrolled adjusted" or "adjusted controlled".

PR #2755, opened 2021-07-27
Add quarterly facility states table, and the cohort reports into it

ssrihari:
**Story card:** [ch4503](https://app.clubhouse.io/simpledotorg/story/4503/support-quarterly-cohort-reports)


spec/models/reports/quarterly_facility_state_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      RefreshMaterializedViews.new.refresh_v2
+      with_reporting_time_zone do
+        quarterly_facility_state_2021_q2 = described_class.find_by(facility: facility, quarter_string: "2021-2")
+
+        expect(quarterly_facility_state_2021_q2.quarterly_cohort_controlled).to eq 2
+        expect(quarterly_facility_state_2021_q2.quarterly_cohort_patients).to eq 12
+        expect(quarterly_facility_state_2021_q2.quarterly_cohort_visited_no_bp).to eq 2
+        expect(quarterly_facility_state_2021_q2.quarterly_cohort_uncontrolled).to eq 3
+        expect(quarterly_facility_state_2021_q2.quarterly_cohort_missed_visit).to eq 5
```
harimohanraj89, line 81: Could we order this as total, controlled, uncontrolled, visit-but-no-bp, missed-visit? We could also put total at the end.

spec/models/reports/quarterly_facility_state_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      patients_uncontrolled.each do |patient|
+        create(:bp_with_encounter, :hypertensive, patient: patient, recorded_at: this_quarter)
+      end
+
+      patients_missed_visit = create_list(:patient, 4, assigned_facility: facility, recorded_at: previous_quarter)
+      patients_missed_visit.each do |patient|
+        create(:bp_with_encounter, patient: patient, recorded_at: previous_quarter)
+      end
+
+      _patient_no_visit = create(:patient, assigned_facility: facility, recorded_at: previous_quarter)
```
harimohanraj89, line 60: Ah, no visit and missed visit above both contribute to the missed visit count. Could we clump these two together?

harimohanraj89 (dismissed):
Some thoughts. Overall, this was more straightforward than I thought it would be!

db/views/reporting_quarterly_facility_states_v01.sql
```diff
@@ -0,0 +1,34 @@
...
+quarterly_cohort_outcomes.uncontrolled AS quarterly_cohort_uncontrolled,
+quarterly_cohort_outcomes.missed_visit AS quarterly_cohort_missed_visit,
+quarterly_cohort_outcomes.visited_no_bp AS quarterly_cohort_visited_no_bp,
+quarterly_cohort_outcomes.patients quarterly_cohort_patients
+
+-- ensure a row for every facility and quarter combination
+FROM reporting_facilities rf
+INNER JOIN reporting_months cal
+-- only pick end of quarters, and current quarter in progress
+    ON (month::integer % 3 = 0 OR month_string = to_char(now(), 'YYYY-MM'))
```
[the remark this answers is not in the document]
ssrihari: We want to have a row for every quarter in this table, even if there is no data for it. The way `reporting_patient_states` is built ensures that now. However, I'd still like to keep this (seemingly redundant) clause so that we independently describe the structure we want in this table.

spec/models/reports/quarterly_facility_state_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        described_class.where(facility: facility)
+        .where("quarter_string >= ?", "2020-1")
+        .where("quarter_string <= ?", "2021-2")
+        .count
+      ).to eq 6
+
+      expect(described_class.where(quarter_string: "2021-1").count).to eq(Facility.count)
+      expect(described_class.find_by(facility: facility, quarter_string: "2021-2").quarterly_cohort_patients).to eq 1
+      expect(described_class.find_by(facility: facility, quarter_string: "2021-1").quarterly_cohort_patients).to eq 1
+      expect(described_class.find_by(facility: facility, quarter_string: "2020-4").quarterly_cohort_patients).to eq 1
```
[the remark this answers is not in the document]
ssrihari: Yup, fixed in e7da2251f. Note that the value on rows where there's no data is nil, not 0. `sum(nil, nil...)` in postgres is nil, I guess. We can coalesce this into 0s if we want to.

spec/models/reports/quarterly_facility_state_spec.rb
ssrihari: Yup, fixed in 9b9a23e2f.

PR #2757, opened 2021-07-27
2966: add translations aggregator script

kpethtel:
**Story card:** [ch2966](https://app.clubhouse.io/simpledotorg/story/2966/obtain-whatsapp-approval-for-all-localized-experiment-messages)

## Because

Adding a script Tim wrote for formatting and outputting all translations for upload to twilio for whatsapp template approval.


timcheadle (approved):
Looks good. We should also add instructions to this:

https://github.com/simpledotorg/simple-server/blob/master/doc/wiki/transifex-localization.md

PR #2759, opened 2021-07-28
Fix flaky drug stock spec

vkrmis:
**Story card:** n/a

## Because

We introduced a flaky spec in #2719 

The flakiness has two underlying causes:
- The `drug_stock` factory was creating unnecessary `protocol_drug` and `protocol` records, even when the association was provided.
- We were looking up `ProtocolDrug`s by `rxnorm_code`, which would occasionally find the extra record created above.

## This addresses

Prevent creation of these extra records, and also lookup drugs in the protocol we explicitly create.

### DIFF
--- a/spec/factories/drug_stocks.rb
+++ b/spec/factories/drug_stocks.rb
@@ -1,10 +1,10 @@
 FactoryBot.define do
   factory :drug_stock do
     id { SecureRandom.uuid }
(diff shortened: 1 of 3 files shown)


harimohanraj89 (approved):
Looks great to me. I agree, I don't see a reason to set the strategies explicitly.

PR #2760, opened 2021-07-28
Fix the start date in reporting months to be in UTC

ssrihari:
Noticed this while debugging an issue with the one-off reports in production.

```sql
FROM generate_series('2017-12-31 18:30:00+00'::timestamp with time zone, now(), '1 mon'::interval)
```

The `month_date`s in the `reporting_months` are off by a bit due to this, and it might explain the tiny difference we’re seeing in the numbers. See below the `month_date`s in production which should be the 1st of every month:
image (11)


db/migrate/20210728100307_fix_reporting_months_start_date.rb
```diff
@@ -0,0 +1,6 @@
+class FixReportingMonthsStartDate < ActiveRecord::Migration[5.2]
+  def up
+    execute "SET LOCAL TIME ZONE '#{Rails.application.config.country[:time_zone]}'"
+    replace_view :reporting_months, version: 3, revert_to_version: 2, materialized: false
```
ssrihari, line 4: We can use `replace_view` here since it's not a matview, and postgres lets us do this as long as the view's schema remains intact.

rsanheim (approved):
🆒 🚢

PR #2762, opened 2021-07-28
Use multiple Twilo senders for all non-IMO communication

rsanheim:
**Story card:** [ch4319](https://app.clubhouse.io/simpledotorg/story/4319/send-twilio-messages-using-multiple-sending-numbers)


prabhanshuguptagit (approved):
Looks good 👌🏼

app/jobs/appointment_notification/worker.rb
```diff
@@ -40,10 +40,8 @@ def perform(notification_id)
   def send_message(notification, communication_type)
     notification_service = if communication_type == "imo"
       ImoApiService.new
-    elsif notification.experiment&.experiment_type == "medication_reminder" && medication_reminder_sms_sender
-      TwilioApiService.new(sms_sender: medication_reminder_sms_sender)
     else
-      TwilioApiService.new
+      TwilioApiService.new(sms_sender: medication_reminder_sms_sender)
```
kpethtel, line 44: Should we move the number selection to the twilio service? It seems more specific to that than to the worker.

rsanheim: I think we should, but lets do it in a future PR...there are too many conditionals involved here and I don't want to break existing behavior that depends on the old single SMS number.

PR #2763, opened 2021-07-28
Track metrics about Twilio errors

rsanheim:
Part of [ch2943](https://app.clubhouse.io/simpledotorg/story/2943/as-a-developer-i-want-to-be-able-monitor-existing-experiments-and-reminders?vc_group_by=day&ct_workflow=all&cf_workflow=500000005)

We want to be able to track errors in addition to the attempted notification sends in one place (a Datadog dashboard).

This wires up the Twilio API client so it tracks metrics as well.

### DIFF
--- a/app/services/twilio_api_service.rb
+++ b/app/services/twilio_api_service.rb
@@ -12,6 +12,8 @@
 
 class TwilioApiService
   attr_reader :client
+  attr_reader :communication_type
+  attr_reader :metrics
   attr_reader :twilio_sender_sms_number
   attr_reader :twilio_sender_whatsapp_number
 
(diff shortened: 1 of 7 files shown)


kpethtel (commented):
Let's talk about this tomorrow. If we're passing the communication type in, I'd argue there's no longer any purpose in having separate `send_sms` and `send_whatsapp` methods. This class was written as though it would be a singular service used to send many messages by varied communication types, but we don't use it that. This implementation also doesn't differentiate between experiment notifications and other notifications, so it doesn't get the most accurate count for the dashboard. Another option could be to pass in an optional block that handles the logging.

app/services/twilio_api_service.rb
```diff
@@ -93,6 +97,7 @@ def send_twilio_message(sender_number, recipient_number, message, callback_url,
       body: message
     )
   rescue Twilio::REST::TwilioError => exception
+    metrics.increment("errors.#{communication_type}")
```
prabhanshuguptagit, line 100: Like @kpethtel mentioned, we can move these increments to `send_sms` and `send_whatsapp` so we don't have to change `TwilioApiService`'s signature. We could also pass in the `communication_type` to `send_twilio_message` as an argument.

rsanheim:
If we decide to do https://app.clubhouse.io/simpledotorg/story/4330/stop-rescuing-twilio-errors this will become much easier, so we should hold off on this until we decide on 4330

rsanheim:
Closing in favor of https://github.com/simpledotorg/simple-server/pull/2765

PR #2764, opened 2021-07-29
4318: do not raise an error when experiment is not found

kpethtel:
**Story card:** [ch4318](https://app.clubhouse.io/simpledotorg/story/4318/release-plan-for-first-full-ab-experiment)

## Because

This allows us to start current patient experiments without manual steps. We can add it to the schedule to start the experiment and remove the schedule code at our leisure.


rsanheim (approved):

kpethtel:
> Shall we do the same for `schedule_daily_stale_patient_notifications` ?

Yeah, I'll do that. As it is, it will work fine but start to error after the experiment is over.

PR #2765, opened 2021-07-29
4330: Better Twilio error handling

kpethtel:
**Story card:** [ch4330](https://app.clubhouse.io/simpledotorg/story/4330/stop-rescuing-twilio-errors) and 
[ch2943](https://app.clubhouse.io/simpledotorg/story/2943/as-a-developer-i-want-to-be-able-monitor-existing-experiments-and-reminders)
## Because

This is undoing a change I made because I had looked at a list of twilio errors and none of them looked like they should merit a retry. However, that list was incomplete and most of the errors we encounter are for things like rate limiting, which we do want to know about. This ensures that all twilio errors will be reported to sentry.

### DIFF
--- a/app/services/twilio_api_service.rb
+++ b/app/services/twilio_api_service.rb
@@ -21,12 +21,6 @@ class TwilioApiService
   delegate :logger, to: Rails
 
   class Error < StandardError
-    attr_reader :exception_message, :context
-    def initialize(message, exception_message:, context:)
-      super(message)
-      @exception_message = exception_message
-      @context = context
-    end
   end
 
   def initialize(sms_sender: nil)
@@ -86,6 +80,7 @@ class TwilioApiService
   private
 
   def send_twilio_message(sender_number, recipient_number, message, callback_url, context)
+    Sentry.set_tags(context)
     client.messages.create(
       from: sender_number,
       to: recipient_number,
@@ -93,6 +88,6 @@ class TwilioApiService
       body: message
     )
   rescue Twilio::REST::TwilioError => exception
-    raise Error.new("Error while calling Twilio API", exception_message: exception.to_s, context: context)
+    raise Error.new("Error while calling Twilio API: #{exception.message}")
   end
 end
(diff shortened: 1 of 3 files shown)


rsanheim:
I've adding metrics here as well to track messages sent and errors.

app/services/twilio_api_service.rb
```diff
@@ -71,12 +68,16 @@ def test_client
   end
 
   def send_sms(recipient_number:, message:, callback_url: nil, context: {})
+    self.communication_type = "sms"
```
rsanheim, line 71: This is hacky, but until we refactor how the Worker works with this service it will be fine.

app/services/twilio_api_service.rb
```diff
@@ -86,28 +87,16 @@ def send_whatsapp(recipient_number:, message:, callback_url: nil, context: {})
...
-  def log_error(e, sender_number, recipient_number, context)
-    logger.info({
-      class: self.class.name,
-      error: e,
-      sender: sender_number,
-      recipient: recipient_number,
-      **context
-    })
+    metrics.increment("#{communication_type}.errors")
+    raise Error.new("Error while calling Twilio API: #{exception.message}")
```
rsanheim, line 100: We don't need to add more state to the custom error ...the original error will be capture implicitly by ruby (and then we will see it in sentry).

rsanheim (approved):

app/services/twilio_api_service.rb
kpethtel: Any reason you prefer having he accessor and setting via self rather than using a reader and setting via instance variable?

app/services/twilio_api_service.rb
kpethtel: Interesting. I didn't know about this Sentry.set_tags.

app/services/twilio_api_service.rb
rsanheim: I think it is a bit cleaner and will make things a bit easier to refactor later.  Ideally we wouldn't have to set communication_type here at all, but instead could do it earlier in the initialize...but that will require some more refactoring.

PR #2767, opened 2021-07-29
Grab missed_visits and control rates from FacilityStates; consolidate to one query for SchemaV2

rsanheim:
**Story card:** [ch3934](https://app.clubhouse.io/simpledotorg/story/3934/create-facility-states-per-month-mat-view) and [ch4310](https://app.clubhouse.io/simpledotorg/story/4310/grab-as-much-repository-schema-v2-data-w-one-query-as-possible)

This PR ended up doing two main things, but they happened to go along together nicely:

* grab control_rates and missed_visits from FacilityStates for the Repository
* grab the summed values via one query per Region for FacilityStates

Along with this there was a bunch of nicely cleaned up code and a lot of dead code that I was able to remove. I also added spec coverage to some areas I found lacking as I went.

Please see the attached review for inline notes.  **NOTE:** This code is still only run for explicit usages of the `v2` flag in prod, we are not yet flipping things over to the reporting pipeline here.


app/models/reports/schema_v2.rb
```diff
@@ -141,23 +95,11 @@ def registered_patients_query_v2(region)
...
 
     memoize def uncontrolled
-      regions.each_with_object({}).each do |region, hsh|
-        if earliest_patient_recorded_at[region.slug].nil?
-          hsh[region.slug] = Hash.new(0)
-          next
-        end
-        hsh[region.slug] = control_rate_query_v2.uncontrolled_counts(region, range: active_range(region))
-      end
+      regions.each_with_object({}) { |region, hsh| hsh[region.slug] = sum(region, :uncontrolled_under_care) }
```
rsanheim, line 102: These came over nicely to the new single-query code, and the old PatientState query can be removed.

rsanheim (commented):
Notes for reviewers

app/models/reports/facility_state.rb
```diff
@@ -8,5 +8,9 @@ def self.for_region(region_or_source)
       region = region_or_source.region
       where("#{region.region_type}_region_id" => region.id)
     end
+
+    def period_month_date
+      Period.month(month_date)
+    end
```
ssrihari, line 14: how about just `period`? `facility_state.period` seems clear to me.

app/models/reports/schema_v2.rb
```diff
@@ -226,5 +186,38 @@ def percentage(numerator, denominator)
...
+    # oldest to newest - it also makes reading output in specs and debugging much easier.
+    memoize def facility_state_data(region)
+      calculations = FIELDS.map { |field| Arel.sql("COALESCE(SUM(#{field}::int), 0) as sum_#{field}") }
+
+      FacilityState.for_region(region)
+        .where("cumulative_registrations IS NOT NULL OR cumulative_assigned_patients IS NOT NULL")
+        .order(:month_date)
+        .group(:month_date)
+        .select(:month_date)
+        .select(*calculations)
```
ssrihari, line 220: Do we not want to filter to the given period here? As in, read only the months that the repository is initiated for?

app/models/reports/schema_v2.rb
```diff
@@ -226,5 +186,38 @@ def percentage(numerator, denominator)
...
+    def sum(region, field)
+      summed_field = "sum_#{field}"
+      facility_state_data(region)
+        .reject { |facility_state| Period.month(facility_state.month_date) < earliest_patient_recorded_at_period[region.slug] }
+        .select { |facility_state| facility_state.period_month_date.in?(periods) }
+        .to_h { |facility_state| [Period.month(facility_state.month_date), facility_state.public_send(summed_field)] }
+        .tap { |hsh| hsh.default = 0 }
+    end
+
+    delegate :sql, to: Arel
```
ssrihari, line 205: Do we need this since we're doing `Arel.sql` and it's not too verbose?

app/models/reports/schema_v2.rb
```diff
@@ -226,5 +186,38 @@ def percentage(numerator, denominator)
...
     end
+
+    # Grab a particular summed field for a region. We return data in the format of:
+    #   { region_slug => { period_1 => x, period_2 => y }, ...}
+    # to maintain consistency w/ the format callers expect from the Repository.
+    #
+    # Note that we do filtering on the result set to limit the returned amount of data to the data that callers are
+    # requesting via the `periods` argument the Repository was created with.
+    def sum(region, field)
+      summed_field = "sum_#{field}"
```
ssrihari, line 197: We use this in a couple of places and it feels like an important implementation detail. Can we extract a function for it?

app/models/reports/schema_v2.rb
```diff
@@ -176,15 +118,33 @@ def registered_patients_query_v2(region)
...
-    def active_range(region)
-      start = [earliest_patient_recorded_at_period[region.slug], periods.begin].compact.max
-      (start..periods.end)
+    def missed_visits_with_ltfu
+      missed_visits(with_ltfu: true)
     end
 
+    def missed_visits_with_ltfu_rates
+      missed_visits_rates(with_ltfu: true)
+    end
```
ssrihari, line 144: Should we do this for the `_without_` versions as well, and remove the defaults? I feel that would be simpler because it would be more consistent / symmetric. Not a blocker though.

app/models/reports/schema_v2.rb
```diff
@@ -176,15 +118,33 @@ def registered_patients_query_v2(region)
       end
     end
 
-    private
+    memoize def missed_visits(with_ltfu: false)
+      field = with_ltfu ? :missed_visit_lost_to_follow_up : :missed_visit_under_care
```
ssrihari, line 122: Nice, is this worth extracting? It would be the same for all treatment control fields no? It could even go into `FacilityState`: `facility_state.field(field_name, care_state)`?

app/models/reports/schema_v2.rb
```diff
@@ -176,15 +118,33 @@ def registered_patients_query_v2(region)
...
+      regions.each_with_object({}) { |region, hsh| hsh[region.slug] = sum(region, field) }
+    end
+
+    def missed_visits_rates(with_ltfu: false)
+      region_period_cached_query(__method__, with_ltfu: with_ltfu) do |entry|
```
ssrihari, line 127: Is this is for convenience, or do we need caching here?

app/models/reports/schema_v2.rb
```diff
@@ -52,85 +64,27 @@ def initialize(regions, periods:)
...
-    end
-
-    private def adjusted_patients_without_ltfu_query(region)
-      FacilityState.for_region(region)
-        .order(:month_date)
-        .group(:month_date)
-        .sum("patients_under_care::int")
-        .to_h(&period_hash)
-        .tap { |hsh| hsh.default = 0 }
+      regions.each_with_object({}) { |region, result| result[region.slug] = sum(region, :patients_under_care) }
```
ssrihari, line 67: Does it make sense to extract a `sum_by_region(regions, field)`?

ssrihari: Following this trail of thought: if we're running a single query to get _all_ the data that goes into the region report on a page, then can we just return a single hash map from the repository, instead of having functions? Will that be simpler?

This doesn't have to be done now, I'm just throwing a suggestion.

app/models/reports/schema_v1.rb
```diff
@@ -160,8 +160,9 @@ def initialize(regions, periods:)
     memoize def missed_visits_without_ltfu_rates
       region_period_cached_query(__method__) do |entry|
         slug, period = entry.slug, entry.period
-        visit_rates = controlled_rates[slug][period] + uncontrolled_rates[slug][period] + visited_without_bp_taken_rate[slug][period]
```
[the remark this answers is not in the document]
ssrihari: Oh what's the bug?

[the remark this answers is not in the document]
ssrihari: And have we verified that the v2 numbers are correct in the same way that we expect here?

app/models/reports/schema_v2.rb
```diff
@@ -3,19 +3,30 @@ class SchemaV2
...
+      cumulative_registrations
+      lost_to_follow_up
+      missed_visit_lost_to_follow_up
+      missed_visit_under_care
+      monthly_registrations
+      patients_under_care
+      uncontrolled_under_care
+      visited_no_bp_under_care
+      visited_no_bp_lost_to_follow_up
+    ].freeze
```
ssrihari, line 18: Should this go into `FacilityState`?

ssrihari: Also see renames in https://github.com/simpledotorg/simple-server/pull/2754.

app/models/reports/schema_v2.rb
ssrihari: 🙌  sweet!

app/models/reports/facility_state.rb
rsanheim: done

app/models/reports/schema_v2.rb
rsanheim: I'm debating going to an API that actually just has the indicator name with the `with_ltfu` argument, and dropping the `with` and `with_out` suffixes entirely.  My plan is to have a rename / refactor PR that lands after v2 is the only code path, as it will be easier to clean things up w/o supporting both versions.

app/models/reports/schema_v2.rb
rsanheim: > instead of having functions? Will that be simpler?

Yeah the Repository API could change to support something like that for sure, but it would mean changing a lot of callers (or having dual interfaces).  Maybe someday down the road 😁 .  I'm ready to get to V2 and let this be mostly stable for awhile until the dashboard refresh or other features require larger changes.

app/models/reports/schema_v1.rb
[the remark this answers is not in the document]
rsanheim: The bug was that the % wouldn't match the counts...going thru the new specs show that things look right (and consistent) between the two versions now.

app/models/reports/schema_v2.rb
rsanheim: Nah, not yet, I want to compare some performance numbers and query plans w/ both approaches before adding that to the `where` clause.

app/models/reports/schema_v2.rb
rsanheim: true, this was leftover - removed

app/models/reports/schema_v2.rb
rsanheim: 👍

app/models/reports/schema_v2.rb
rsanheim: No don't think so, these are just the fields that Repository cares about from FacilityState.

app/models/reports/schema_v2.rb
ssrihari: What's the trade-off though? As in, what do we lose in adding a where clause given that the repo is instantiated for a given period anyway?

PR #2770, opened 2021-07-30
Draft of stale patient notifications per day formula

rsanheim:
**Story card:** [ch4373](https://app.clubhouse.io/simpledotorg/story/4373/calculate-how-many-patients-we-can-send-to-per-day-for-the-stale-patient-experiment)

## Because

We want to have a number of patients per day that we add to the 'stale patient experiment' such that we stay under the WhatsApp rate limit. This script is an attempt at a formula for that.


rsanheim:
We probably can keep it even simpler than this and just pick a number under 5k and call it good, so closing.

PR #2771, opened 2021-08-01
4322: report to sentry when experiment is not found and make current …

kpethtel:
…patient experiments self-completing

**Story card:** [ch4322](https://app.clubhouse.io/simpledotorg/story/4322/add-both-active-and-stale-experiments-to-scheduler)

## Because

We want experiments to have fewer manual steps. With this approach, we can create an experiment and add it to the scheduler and forget about it until it's done and starts to complain to sentry.


rsanheim (approved):
We should discuss if there are any timezone conversions we need to worry about here, but we can pair on that and do a seperate PR if need be.

LGTM! 👍

PR #2774, opened 2021-08-02
Grab visited_without_bp_taken rates from repository

rsanheim:
No need to calculate these via result anymore...which should either fix the related Sentry error or at least make it easier to track down.

**Story card:** [ch4390](https://app.clubhouse.io/simpledotorg/story/4390/error-in-my-facilities-dashboard-undefined-method-each-with-object)

**Sentry** https://sentry.io/organizations/resolve-to-save-lives/issues/2540211592/?project=1217715&referrer=Clubhouse

### DIFF
--- a/app/services/reports/region_service.rb
+++ b/app/services/reports/region_service.rb
@@ -42,8 +42,8 @@ module Reports
       result.uncontrolled_patients_with_ltfu_rate = repository.uncontrolled_rates(with_ltfu: true)[slug]
 
       result.visited_without_bp_taken = repository.visited_without_bp_taken[region.slug]
-      result.calculate_percentages(:visited_without_bp_taken)
-      result.calculate_percentages(:visited_without_bp_taken, with_ltfu: true)
+      result.visited_without_bp_taken_rate = repository.visited_without_bp_taken_rate[region.slug]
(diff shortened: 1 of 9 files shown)


app/services/reports/region_service.rb
```diff
@@ -42,8 +42,8 @@ def call
       result.uncontrolled_patients_with_ltfu_rate = repository.uncontrolled_rates(with_ltfu: true)[slug]
 
       result.visited_without_bp_taken = repository.visited_without_bp_taken[region.slug]
-      result.calculate_percentages(:visited_without_bp_taken)
-      result.calculate_percentages(:visited_without_bp_taken, with_ltfu: true)
+      result.visited_without_bp_taken_rate = repository.visited_without_bp_taken_rate[region.slug]
+      result.visited_without_bp_taken_with_ltfu_rate = repository.visited_without_bp_taken_rate(with_ltfu: true)[region.slug]
```
rsanheim, line 46: This is the (hopeful) bug fix - we no longer need to recalc the percentage, as repository already has done that for us.

kpethtel (approved):
I can't claim to have a lot of knowledge of it, but it LGTM.

PR #2775, opened 2021-08-02
Set Sentry content in notification worker

rsanheim:
For more context on errors like
https://sentry.io/organizations/resolve-to-save-lives/issues/2539076354/?environment=production&project=1217715&query=is%3Aunresolved


kpethtel (approved):
👍

PR #2776, opened 2021-08-02
Don't depend on an appointment for stale patient notifications

rsanheim:
**Story card:** [ch4398](https://app.clubhouse.io/simpledotorg/story/4398/failing-notification-creation-nomethoderror-undefined-method-facility-for-nil-nilclass)

Also an intial step towards [ch4399](https://app.clubhouse.io/simpledotorg/story/4399/write-a-b-integration-specs)


app/models/notification.rb
```diff
@@ -28,20 +29,29 @@ class Notification < ApplicationRecord
...
+    when "experimental_appointment_reminder"
+      facility = subject&.facility || patient.assigned_facility
       I18n.t(
         message,
-        facility_name: patient.assigned_facility.name,
+        facility_name: facility.name,
         patient_name: patient.full_name,
-        locale: patient.assigned_facility.locale
+        appointment_date: subject&.scheduled_date,
+        locale: facility.locale
```
kpethtel, line 46: This is an edge case, but I'd argue that all of these locales should be `patient.locale`, which delegates to `patient.assigned_facility.locale`. If a patient ever had an appointment at a facility other than their assigned facility, they would get notifications in the usual language.

app/services/experiment_control_service.rb
```diff
@@ -4,6 +4,10 @@ class ExperimentControlService
   BATCH_SIZE = 100
 
   class << self
+    def logger
+      @logger ||= Rails.logger.child(class: name)
```
kpethtel, line 8: What is name here? As an aside, instance variables in class methods always throw me for a loop.

app/models/notification.rb
```diff
@@ -28,20 +29,29 @@ class Notification < ApplicationRecord
...
-        facility_name: subject.facility.name,
+        facility_name: patient.assigned_facility.name,
         patient_name: patient.full_name,
-        appointment_date: subject.scheduled_date,
-        locale: subject.facility.locale
+        locale: patient.assigned_facility.locale
       )
-    when "covid_medication_reminder"
+    when "experimental_appointment_reminder"
+      facility = subject&.facility || patient.assigned_facility
```
rsanheim, line 40: This is the actual bug fix - make sure we have a facility to determine the locale

app/services/experimentation/runner.rb
```diff
@@ -1,26 +1,38 @@
...
-
-  class << self
-    def start_current_patient_experiment(name:, percentage_of_patients: 100)
-      experiment = Experimentation::Experiment.find_by(name: name, experiment_type: "current_patients", state: [:new, :running])
-      unless experiment
-        Sentry.capture_message("Experiment #{name} not found and may need to be removed from scheduler.")
+# require_relative "stale_patient_selection"
+
+module Experimentation
+  class Runner
```
rsanheim, line 4: Moving this into Experimentation makes a bunch of other things more clear and easier to follow below...

app/services/experimentation/runner.rb
```diff
@@ -1,26 +1,38 @@
...
+        logger.info("Experiment #{name} is past its end_date of #{experiment.end_date} - completing.")
         experiment.complete_state!
         return
       end
 
-      return if experiment.state == "running"
+      if experiment.running_state?
+        logger.info("Experiment #{name} is a running current_patient experiment, nothing to do - exiting.")
+        return
+      end
```
rsanheim: Logging for operators to follow along

app/services/experimentation/runner.rb
```diff
@@ -41,25 +53,31 @@ def start_current_patient_experiment(name:, percentage_of_patients: 100)
...
 
+      logger.info("Finished scheduling notifications for #{name} experiment - marking as running and exiting.")
       experiment.running_state!
     end
 
-    def schedule_daily_stale_patient_notifications(name:, patients_per_day: PATIENTS_PER_DAY)
-      experiment = Experimentation::Experiment.find_by(name: name, experiment_type: "stale_patients", state: [:new, :running])
-      unless experiment
-        Sentry.capture_message("Experiment #{name} not found and may need to be removed from scheduler.")
+    def self.schedule_daily_stale_patient_notifications(name:, patients_per_day: PATIENTS_PER_DAY)
```
rsanheim, line 60: Changing to `self.` allows rails constant lookup to just work when we are inside a module like this

app/services/experimentation/runner.rb
```diff
@@ -74,39 +92,39 @@ def schedule_daily_stale_patient_notifications(name:, patients_per_day: PATIENTS
...
         notification.status_cancelled!
       end
+      logger.warn "Aborting experiment #{name} finished."
     end
 
-    protected
-
-    def current_patient_candidates(start_date, end_date)
-      Experimentation::Experiment.candidate_patients
+    def self.current_patient_candidates(start_date, end_date)
```
rsanheim, line 108: Operators can just run this directly now since its not protected.

lib/seed/experiment_seeder.rb
```diff
@@ -2,37 +2,35 @@ module Seed
...
-
-      cascade = experiment.treatment_groups.create!(description: "cascade")
-      cascade.reminder_templates.create!(message: "notifications.set02.basic", remind_on_in_days: 0)
-      cascade.reminder_templates.create!(message: "notifications.set03.basic", remind_on_in_days: 3)
+      transaction do
+        Experimentation::Experiment.stale_patients.create!(name: experiment_name, state: "new", start_date: start_date, end_date: end_date).tap do |experiment|
+          _control_group = experiment.treatment_groups.create!(description: "control")
+
+          single_group = experiment.treatment_groups.create!(description: "single_notification")
+          single_group.reminder_templates.create!(message: "notifications.set03.basic", remind_on_in_days: 0)
```
rsanheim, line 31: Just one message for stale patient treatment group.

spec/services/experimentation/runner_spec.rb
```diff
@@ -399,11 +401,43 @@
...
       }.to change { experiment.reload.state }.to("cancelled")
       expect(pending_notification.reload.status).to eq("cancelled")
       expect(scheduled_notification.reload.status).to eq("cancelled")
       expect(sent_notification.reload.status).to eq("sent")
     end
   end
+
+  # NOTE: putting a best attempt at an e2e test here for now, with the intention that we can
+  # pull it out to its own dedicated integration test file soon
+  describe "end to end testing" do
```
rsanheim, line 414: New end to end integration test to verify all the things

rsanheim (commented):
Ended up refactoring a bit more here, but I think its worth it to have things a bit easier to follow...I think we can change the runner to be an instance instead of class methods at some point, which would allow us to clean it up quite a bit more.  But that can come later.

kpethtel (approved):
🚢

app/models/notification.rb
rsanheim: Discussed this while on zoom w/ Kris, and deciding to ship this as is -- this sounds like a question more for Praveen about how this really should work.

PR #2779, opened 2021-08-03
Update BD production state names

prabhanshuguptagit:
**Story card:** [ch4072](https://app.clubhouse.io/simpledotorg/story/4072/update-bangladesh-production-districts)

## Because

The Bangladesh team would like to update the following:

- Rename Dhaka North district to Dhaka
- Delete Dhaka South
- Add Chattogram district

## This addresses

Updates `canonical_states.yml`. Will rename `Dhaka North` to `Dhaka` via console.


claudiovallejo (approved):

harimohanraj89:
What happens to facilities that have a "state" value of Dhaka North/South? I see you mentioned "Region hierarchy will be unaffected" in the story. Do we not need to rename/move around region records too?

prabhanshuguptagit:
Will manually rename the `Dhaka North` region to `Dhaka` and update the facility's states.

harimohanraj89:
Understood. Any reason not to roll that in as a data migration?

prabhanshuguptagit:
No particular reason, just trivial enough to do on console 😅

harimohanraj89:
If it's simple enough, I'd like to commit it into the repo as a data migration so that:
* We can track what we did
* We can run it across envs (not totally applicable in this case but a good practice)

Let me know if that makes sense!

PR #2803, opened 2021-08-03
Add comments on table and columns for reporting patient states

ssrihari:
This is in an effort to create documentation for the reporting schema, allowing epidemiologists to peruse the data and craft reports themselves.

I'll create a story with all the context soon, and reference it here.


app/models/reports/matview.rb
```diff
@@ -4,7 +4,29 @@ def self.refresh
       ActiveRecord::Base.transaction do
         ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{Period::REPORTING_TIME_ZONE}'")
         Scenic.database.refresh_materialized_view(table_name, concurrently: true, cascade: false)
+        add_comments
```
ssrihari, line 7: We don't really need to add comments with every refresh. I am looking for a way to abstract this away.

app/models/reports/matview.rb
```diff
@@ -4,7 +4,29 @@ def self.refresh
...
+    def self.add_comments
+      table_comments = YAML.load_file("config/schema_descriptions.yml")[table_name]
+      return unless table_comments.present?
+      add_table_description(table_name, table_comments["description"])
+      table_comments["columns"].each do |column_name, column_description|
+        add_column_description(column_description, column_name)
+      end
+    end
+
+    def self.add_table_description(table_name, table_description)
```
ssrihari, line 20: Rails doesn't have a way to change comments from what I can tell, some [older gems](https://github.com/pinnymz/migration_comments) do. I'll try them out.

I would also like the ability to throw an error if the yml file doesn't have the description for a given column, and we'd need to hand-roll that at least.

config/schema_descriptions.yml
```diff
@@ -0,0 +1,72 @@
...
+    year: "Year in YYYY format"
+    month_string: "String that represents a month, in YYYY-MM format"
+    quarter_string: "String that represents a month, in YYYY-Q format"
+    hypertension: "Yes if patient is hypertensive, no if not, and nil if the data is unavailable"
+    prior_heart_attack: "Yes if the patient has had a heart attack, no if not, nil if unknown"
+    prior_stroke: "Yes if the patient has had a stroke, no if not, nil if unknown"
+    chronic_kidney_disease: "Yes if the patient has had a chronic kidney disease, no if not, nil if unknown"
+    receiving_treatment_for_hypertension: "Yes if the patient was already receiving treatment for hypertension during registration"
+    diabetes: "Yes if patient is diabetic, no if not, and nil if the data is unavailable"
+    assigned_facility_id: "Facility ID to which the patient is currently assigned"
```
harimohanraj89, line 24: Consider "Facility ID of the patient's assigned facility"

config/schema_descriptions.yml
```diff
@@ -0,0 +1,72 @@
...
+    assigned_facility_size: "Size of the assigned facility"
```
harimohanraj89, line 25: Consider adding "patient's" everywhere. "...patient's assigned facility"

config/schema_descriptions.yml
```diff
@@ -0,0 +1,72 @@
+reporting_patient_states:
+  description: "Monthly summary of a patient's information and health indicators. This table has one row per patient, per month, from the month of the patient's registration."
+  columns:
+    patient_id: "ID of the patient"
+    recorded_at: "Time (in UTC) at which the patient was registered"
+    status: "active, dead, migrated, etc"
+    gender: "Gender of the patient"
+    age: "Age of the patient as entered by a nurse"
+    age_updated_at: "Time (in UTC) at which the age was last updated"
```
harimohanraj89, line 9: What convention can we use to denote a reference to another field? ie, "at which the `age` was last updated". It's not clear that "age" is referring to the previous field, I think.

config/schema_descriptions.yml
```diff
@@ -0,0 +1,72 @@
...
+    date_of_birth: "Date of birth of the patient"
+    current_age: "Computed age as of today, given age, age-updated-at and date-of-birth. This will have the same value for a given patient across all rows."
```
harimohanraj89, line 11: This one is a doozy. Hmm, how about something like:

> Patient's age as of today, based on `age`, `age-updated-at`, and `date-of-birth`. This will...

I'm also sensing that we have some restrictions on what characters we can use in these comments. Feel free to us `Age of the patient`. The confusion for me mostly stems from "given age".

config/schema_descriptions.yml
```diff
@@ -0,0 +1,72 @@
...
+    month_date: "Date at the beginning of the month for which this row has the latest information of a patient"
+    month: "Month (1-12) of year"
+    quarter: "Quarter (1-4) of year"
+    year: "Year in YYYY format"
+    month_string: "String that represents a month, in YYYY-MM format"
+    quarter_string: "String that represents a month, in YYYY-Q format"
```
harimohanraj89, line 17: "represents a quarter"

config/schema_descriptions.yml
```diff
@@ -0,0 +1,72 @@
...
+    hypertension: "Yes if patient is hypertensive, no if not, and nil if the data is unavailable"
```
harimohanraj89, line 18: Consider a more conceptual description:

> Has the patient been diagnosed with hypertension (yes/no/unknown). Nil if the data is unavailable.

Also, let's make sure to use the correct enum values with the right case as well.

harimohanraj89: Obv same for all medical history questions

config/schema_descriptions.yml
```diff
@@ -0,0 +1,72 @@
...
+    registration_block_region_id: "ID of the registration facility's block"
+    registration_district_slug: "Human readable ID of the registration facility's district"
+    registration_district_region_id: "ID of the registration facility's district"
+    registration_state_slug: "Human readable ID of the registration facility's state"
+    registration_state_region_id: "ID of the registration facility's state"
+    registration_organization_slug: "Human readable ID of the registration facility's organization"
+    registration_organization_region_id: "ID of the registration facility's organization"
+    blood_pressure_id: "ID of the latest blood pressure as of this month. Use this to join with the blood pressures table."
+    bp_facility_id: "ID of the facility at which the latest BP was observed as of this month"
+    bp_recorded_at: "Time (in UTC) at which the latest BP as of this month was recorded"
```
harimohanraj89, line 52: Some fragmentation of language across:
* BP vs blood pressure
* Observed vs recorded vs measured

Shall we consistent-ize this?

config/schema_descriptions.yml
```diff
@@ -0,0 +1,72 @@
...
+  columns:
+    patient_id: "ID of the patient"
+    recorded_at: "Time (in UTC) at which the patient was registered"
+    status: "active, dead, migrated, etc"
+    gender: "Gender of the patient"
+    age: "Age of the patient as entered by a nurse"
+    age_updated_at: "Time (in UTC) at which the age was last updated"
+    date_of_birth: "Date of birth of the patient"
+    current_age: "Computed age as of today, given age, age-updated-at and date-of-birth. This will have the same value for a given patient across all rows."
+    month_date: "Date at the beginning of the month for which this row has the latest information of a patient"
```
harimohanraj89, line 12: I wonder if a simpler description would be more effective for and epi audience. Eg. "The reporting month of this row". We can then utilize "the reporting month" as a consistent reference down below, eg. for the treatment outcomes and stuff.

config/schema_descriptions.yml
ssrihari: I've added single quotes around it, hope that's understood. I also wonder if we can remove this field if the `current_age` will suffice. This documentation has got me thinking about trimming some fields for clarity.

config/schema_descriptions.yml
ssrihari: Yup, makes sense, I've updated it.

config/schema_descriptions.yml
ssrihari: I've updated the definition here. But I like "this month" in other descriptions because it feels simpler, easier to relate to. Wdyt?

config/schema_descriptions.yml
ssrihari: 🤦 fixed

config/schema_descriptions.yml
ssrihari: Yeah, made this change :) Would be nice to have enums for the field so that we don't have to explain the data type in the description. Perhaps in a later update.

config/schema_descriptions.yml
ssrihari: Done.

config/schema_descriptions.yml
ssrihari: Added this. It's a bit awkward in some places: "patient's registration facility's state", but it's alright I think.

config/schema_descriptions.yml
ssrihari: Thanks! I've used "recorded" in all places, to also be consistent with the "recorded_at" field.

config/schema_descriptions.yml
harimohanraj89: Ha, that's a good sign. We can also choose to leave all the fields and descriptions as is, and publicize only a select few in whatever learning material we craft for epis.

config/schema_descriptions.yml
```diff
@@ -6,52 +6,52 @@ reporting_patient_states:
     status: "active, dead, migrated, etc"
     gender: "Gender of the patient"
     age: "Age of the patient as entered by a nurse"
-    age_updated_at: "Time (in UTC) at which the age was last updated"
+    age_updated_at: "Time (in UTC) at which the field 'age' was last updated"
     date_of_birth: "Date of birth of the patient"
-    current_age: "Computed age as of today, given age, age-updated-at and date-of-birth. This will have the same value for a given patient across all rows."
-    month_date: "Date at the beginning of the month for which this row has the latest information of a patient"
+    current_age: "Patient's age as of today, based on age, age-updated-at and date-of-birth. This will have the same value for a patient across all rows."
```
harimohanraj89, line 11: Let's wrap these field references in single quotes as well for consistency. Shall we use snake case as well?

config/schema_descriptions.yml
```diff
@@ -6,52 +6,52 @@ reporting_patient_states:
...
+    month_date: "The reporting month for this row, represented as the date at the beginning of the month"
```
harimohanraj89, line 12: Love this. 👍

config/schema_descriptions.yml
```diff
@@ -63,9 +63,9 @@ reporting_patient_states:
...
     months_since_visit: "Number of months since the patient's last visit. If a patient visited on 31st Jan, it would be 1 month since the visit on 1st Feb."
     quarters_since_visit: "Number of quarters since the patient's last visit. If a patient visited on 31st Jan, it would be 1 quarter since the visit on 1st Jan."
-    months_since_bp: "Number of months since the patient's last BP measurement. If a patient had a BP reading on 31st Jan, it would be 1 month since BP on 1st Feb."
-    quarters_since_bp: "Number of quarters since the patient's last BP measurement. If a patient had a BP reading on 31st Jan, it would be 1 quarter since BP on 1st Jan."
-    last_bp_state: "The state of the last BP measured: controlled, uncontrolled, or unkown"
+    months_since_bp: "Number of months since the patient's last BP recording. If a patient had a BP reading on 31st Jan, it would be 1 month since BP on 1st Feb."
+    quarters_since_bp: "Number of quarters since the patient's last BP recording. If a patient had a BP reading on 31st Jan, it would be 1 quarter since BP on 1st Jan."
+    last_bp_state: "The state of the last BP recorded: controlled, uncontrolled, or unknown"
```
harimohanraj89, line 68: Interesting. Should we add the threshold for control (140/90) to this description? Or leave that decoupled from these field descriptions?

config/schema_descriptions.yml
harimohanraj89: I think that sounds great 👍  PS the updated description for `month_date` is :chef_kiss:

app/models/reports/matview.rb
harimohanraj89: Re: throwing errors. I like that idea. It reminds me of the Sentry alerts we had for "regions integrity check", if any parts of our region hierarchy are out of whack. This would be a similar reminder for us to populate missing descriptions if we forget during PR review.

harimohanraj89: If we want to autogenerate this "handbook", shall we modify the YML structure to tag each field as, say, "important" and "not important", or something? Some mechanism that'll allow us to highlight the main fields that we think will be used often, and put the other verbose stuff somewhere below?

ssrihari: Oh yes, the "summary" vs "detail" distinction seems useful. I'll add that in!

config/schema_descriptions.yml
ssrihari: Yeah, the definition and comments might diverge if we add it in both places.

rsanheim:
Do the comments show up in a helpful way in metabase? Can you drop some screenshots of how they would be presented to end users once they are added?

ssrihari:
@harimohanraj89 @rsanheim 

Here's the table description:
image

Here are the field descriptions:
image

PR #2804, opened 2021-08-03
4322: add August experiments to scheduler

kpethtel:
**Story card:** [ch4322](https://app.clubhouse.io/simpledotorg/story/4322/add-both-active-and-stale-experiments-to-scheduler)

## Because

We want to start the new experiments. We want an hour between scheduling notifications and sending them to allow time to check data. 

## Testing
 
 I tried to test this locally but ran into issues of the crontab using the wrong ruby version. It didn't seem worth the effort. Aside from typos, I don't see any causes for concern.


config/schedule.rb
```diff
@@ -43,20 +43,23 @@ def local(time)
...
   if Flipper.enabled?(:weekly_telemed_report)
     rake "reports:telemedicine"
   end
 end
 
-every :day, at: local("07:00 am"), roles: [:cron] do
-  runner "AppointmentNotification::ScheduleExperimentReminders.perform_now"
+every :day, at: local("07:30 am"), roles: [:cron] do
+  if CountryConfig.current_country?("India") && SimpleServer.env.production?
+    runner "ExperimentControlService.start_current_patient_experiment(name: 'Current Patient August 2021')"
```
kpethtel, line 54: I'm not worried about them running at the same time because the current job should do all its work on Wednesday and the stale job should do its work every day after.

app/services/experimentation/runner.rb
```diff
@@ -9,6 +9,11 @@ def self.logger
     end
 
     def self.start_current_patient_experiment(name:, percentage_of_patients: 100)
+      unless Flipper.enabled?(:experiment)
+        logger.info("Experiment feature flag is off. Experiment #{name} will not be started.")
+        return
+      end
+
```
rsanheim, line 16: 👍

rsanheim (approved):

PR #2811, opened 2021-08-04
Don't allow duplicate block names in FG

prabhanshuguptagit:
**Story card:** [ch4397](https://app.clubhouse.io/simpledotorg/story/4397/regions-integrity-failure-in-production)

## Because

We found some blocks with the same name in a district. This is not a valid state and we received a Region Integrity alert.

## This addresses

Blocks are created from the `Create District` form. The form uniques blocks by name with Javascript but we don't unique them on the backend. We suspect a user managed to submit the form with the same block name twice somehow.

This uniques block names in the model callback.

The integrity failure was fixed manually by deleting the duplicate blocks (these were all created recently and had no facilities).

### DIFF
--- a/app/models/facility_group_region_sync.rb
+++ b/app/models/facility_group_region_sync.rb
@@ -30,7 +30,7 @@ class FacilityGroupRegionSync < SimpleDelegator
   def create_block_regions
     return if new_block_names.blank?
 
-    new_block_names.map { |name|
+    new_block_names.uniq.map { |name|
       Region.block_regions.create!(name: name, reparent_to: region)
     }
   end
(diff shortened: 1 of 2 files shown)


rsanheim (commented):
Looks good as a quick fix.  

How about trying this on Region?

```
  validates :name, uniqueness: conditions: -> { `some condition to get descendant path` }, if: :block_regon
```

basically assert all names are unique for all blocks with the same descendant path (ie the same district)

prabhanshuguptagit:
I tried this locally, it worked. Will need to figure out propagating the validation to the importer. Will do as a follow up.

PR #2812, opened 2021-08-04
Data migration to discard the offending BPs

harimohanraj89:
**Story card:** [ch4415](https://app.clubhouse.io/simpledotorg/story/4415/et-0-0-bps)

## Because

The ET data import process imported some rows with no BP information and created "blank" BPs (sys: 0, dia: 0)

## This addresses

This PR adds a data migration to discard the offending BPs, and their associated observations.

## Test instructions

Here's how I'll be testing this:
* [x] Pull ET data locally
* [x] Run the data migration
* [x] Ensure the dashboard is working (My Facilities, user activity log, reports)
* [x] Ensure the app can sync after the data migration

### DIFF
--- a/db/data/20210804083620_delete_zero_bps.rb
+++ b/db/data/20210804083620_delete_zero_bps.rb
@@ -1,0 +1,16 @@
+class DeleteZeroBps < ActiveRecord::Migration[5.2]
+  def up
+    return unless CountryConfig.current[:abbreviation] == "ET"
+
+    zero_bps = BloodPressure.where(systolic: 0, diastolic: 0)
+
+    zero_bps.find_each do |blood_pressure|
+      blood_pressure.observation&.discard
+      blood_pressure.discard
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/db/data_schema.rb
+++ b/db/data_schema.rb
@@ -1,2 +1,2 @@
 # encoding: UTF-8
-DataMigrate::Data.define(version: 20210730185818)
+DataMigrate::Data.define(version: 20210804083620)


db/data/20210804083620_delete_zero_bps.rb
```diff
@@ -0,0 +1,16 @@
+class DeleteZeroBps < ActiveRecord::Migration[5.2]
+  def up
+    return unless CountryConfig.current[:abbreviation] == "ET"
+
+    zero_bps = BloodPressure.where(systolic: 0, diastolic: 0)
+
+    zero_bps.find_each do |blood_pressure|
+      blood_pressure.observation&.discard
```
rsanheim, line 8: Can we just `delete` these records? We shouldn't really keep these around at all if we can help it.

harimohanraj89: Hard deletion is a classic problem we haven't solved yet 😅  Deleting records on the server has no hook that notifies clients to delete them from device, so they will linger in our distributed network of data. Furthermore, anything that touches the records on client (in this case, only a data migration or something like that) will resync these deleted BPs back to the server, and the server will have no way of recognizing that they're deleted records, and create them again.

vkrmis, line 8: Should we also discard the encounter here?

harimohanraj89: In theory, we'd check if there are any other activities associated with the encounter, and delete if nothing else is there. I chose not to make this thing that smart because all 0/0 BPs were imported from a spreadsheet where each element definitely represented a visit - ie, had a visit date, prescribed meds, etc. Hence I chose to leave the encounters alone. What do you think?

vkrmis: Okay, makes sense to leave it as is in this case 👍

vkrmis (approved):
lgtm

PR #2813, opened 2021-08-04
Initialize region service with a region

harimohanraj89:
**Story card:** [ch4413](https://app.clubhouse.io/simpledotorg/story/4413/my-facilities-bug)

Quick-fix for latest My Facilities bug. The problem surfaces when a facility and its region have different slugs (which is common). This makes everyone happy by initializing the `RegionService` with region records for now.

The follow-up robust fix would be to teach the `RegionService` to convert whatever it receives into region records as soon as possible, similar to how the `Repository` does today.

### DIFF
--- a/app/controllers/my_facilities/facility_performance_controller.rb
+++ b/app/controllers/my_facilities/facility_performance_controller.rb
@@ -28,7 +28,7 @@ class MyFacilities::FacilityPerformanceController < AdminController
 
     @facilities.each do |facility|
       slug = facility.region.slug
-      @data_for_facility[slug] = Reports::RegionService.new(region: facility,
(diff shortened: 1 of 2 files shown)


prabhanshuguptagit (approved):
Nice catch 👌🏼  There's another [usage](https://github.com/simpledotorg/simple-server/blob/962e42297ed4afe0d830986071c15d6a79d20581/app/controllers/my_facilities/facility_performance_controller.rb#L31), let's also fix that and :shipit:

harimohanraj89:
Nice catch yourself 👌 Fixed in 1ef9cf2 and I'll ship after ✅

PR #2815, opened 2021-08-04
Add monthly_registrations_by_user to cache warmer

rsanheim:
**Story card:** [ch4416](https://app.clubhouse.io/simpledotorg/story/4416/registrations-mismatch-in-hwc-jitta-kalan)

## Because

I suspect the problem w/ registration counts per user is that we just aren't busting the cache.

This adds that specific call to the cache warmer.


rsanheim:
https://dashboard-sandbox.simple.org/reports/regions/facility/ch-cascabel-village/details before the cache warmer is run on SBX:

image

kpethtel (approved):

rsanheim:
After I ran the cache warmer for the above region in question: 

image

Note we now have the 5 registrations listed...called this fixed.

PR #2817, opened 2021-08-05
Add state protocol configuration for drug stock calculations

vkrmis:
**Story card:** [ch4425](https://app.clubhouse.io/simpledotorg/story/4425/add-state-level-configuration-for-drug-stock-calculations)

## Because

We need to configure state protocol factors for use in drug stock calculations

## This addresses

Adds configuration for missing states

### DIFF
--- a/config/data/drug_stock_config.yml
+++ b/config/data/drug_stock_config.yml
@@ -1,4 +1,64 @@
-Punjab:
+Andhra Pradesh:
+  load_coefficient: 1
+  drug_categories:
+    hypertension_ccb:
+      new_patient_coefficient: 1.4
+      '329528': 1
+      '329526': 2
+      hypertension_arb:
+        new_patient_coefficient: 0.37
+        '316764': 1
+        '316765': 2
+      hypertension_diuretic:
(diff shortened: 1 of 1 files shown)


config/data/drug_stock_config.yml
```diff
@@ -1,4 +1,64 @@
...
+      '329528': 1
+      '329526': 2
+      hypertension_arb:
+        new_patient_coefficient: 0.37
+        '316764': 1
+        '316765': 2
+      hypertension_diuretic:
+        new_patient_coefficient: 0.06
+        '429503': 1
+        '316049': 2
```
prabhanshuguptagit, line 15: Does the `1` and `2` here stand for doses per day? If this isn't captured anywhere already, it might be helpful to leave a comment with the expected structure of the YAML. `State > drug_category > rxnorm_code etc. ...`

prabhanshuguptagit (approved):

config/data/drug_stock_config.yml
vkrmis: No, this is the `drug_coefficient` used for base doses calculations.

PR #2818, opened 2021-08-05
Drug stock improvements 2

vkrmis:
**Story card:** [ch4365](https://app.clubhouse.io/simpledotorg/story/4365/drug-stock-improvements)

## Because

Some additional Drug Stock improvements

## This addresses

- Order drug stock districts by name in the selector
- Simplify boolean logic to show/hide block rows
- Bugfix: the facilities subtotal base doses consumption was looking up a non-existent key
- Bugfix: the new drug stock form was not preserving region type when the month selector is used
- Custom drug category order config was incorrectly formatted, remove commas

### DIFF
--- a/app/controllers/my_facilities/drug_stocks_controller.rb
+++ b/app/controllers/my_facilities/drug_stocks_controller.rb
@@ -92,10 +92,10 @@ class MyFacilities::DrugStocksController < AdminController
   end
 
   def blocks_to_display
-    if @selected_facility_sizes != @facility_sizes || @selected_zones != @zones
-      Region.none
-    else
+    if @selected_facility_sizes == @facility_sizes && @selected_zones == @zones
       @query.blocks.order(:name)
+    else
+      Region.none
     end
   end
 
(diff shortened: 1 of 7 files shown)


app/controllers/my_facilities/drug_stocks_controller.rb
```diff
@@ -92,10 +92,10 @@ def set_region
   end
 
   def blocks_to_display
-    if @selected_facility_sizes != @facility_sizes || @selected_zones != @zones
-      Region.none
-    else
+    if @selected_facility_sizes == @facility_sizes && @selected_zones == @zones
       @query.blocks.order(:name)
+    else
+      Region.none
```
prabhanshuguptagit, line 98: Much better😅

prabhanshuguptagit (approved):
Looks good

app/controllers/my_facilities/drug_stocks_controller.rb
vkrmis: @harimohanraj89 spotted it in our original PR 🙏

PR #2825, opened 2021-08-06
Discarding patient data should discard the notifications as well

ssrihari:
**Story card:** [ch4452](https://app.clubhouse.io/simpledotorg/story/4452/notifications-are-not-deleted-when-a-patient-is-deleted)

## This addresses

- Deletion of notifications on deletion of the patient
- ~~Fault tolerance in the scheduler job~~ This is in https://github.com/simpledotorg/simple-server/pull/2827
- ~~Deduplication of notifications~~ This will be a different story, and PR

## Open questions

- How do we deal with the deletion, and deduplication of `communication`s?


app/services/patient_deduplication/deduplicator.rb
```diff
@@ -184,6 +185,10 @@ def create_encounters_and_observables(patient)
       end
     end
 
+    def create_notifications(patient)
+      create_cloned_records!(patient, Notification, Notification.where(patient_id: latest_patient.id))
```
ssrihari, line 189: We should probably also clone communications, ya?

ssrihari: And do we discard communications when we discard notifications (when we discard a patient)?

app/jobs/appointment_notification/schedule_experiment_reminders.rb
```diff
@@ -10,11 +10,18 @@ def perform
...
     next_messaging_time = Communication.next_messaging_time
-
     logger.info "scheduling #{notifications.count} notifications that are due with next_messaging_time=#{next_messaging_time}"
     notifications.each do |notification|
       notification.status_scheduled!
       AppointmentNotification::Worker.perform_at(next_messaging_time, notification.id)
+    rescue => e
+      Sentry.capture_message("Scheduling notification failed",
+        extra: {
+          reminder: notification,
```
prabhanshuguptagit, line 20: Let's send just the notification ID? Not sure how the notification will get serialised.

prabhanshuguptagit (approved):

app/jobs/appointment_notification/schedule_experiment_reminders.rb
prabhanshuguptagit, line 20:
```suggestion
          notification: notification,
```

prabhanshuguptagit:
@ssrihari On second thought, since we know this _will_ fail for deleted patients, shall we also add a guard clause to check for patient existence? People might get desensitised to this job throwing a sentry alert every other day.

harimohanraj89:
@prabhanshuguptagit dont' think it's worth it for a couple reasons:
* Makes the code more complex. It's patients today, what will it be tomorrow?
* We're only doing this rescuing because some patients may have already gotten through the door before we make the update to `discard_data`. Once we ship the `discard_data` fix here as well, we'll be stemming the influx of these kinds of notifications.

app/jobs/appointment_notification/schedule_experiment_reminders.rb
harimohanraj89: I like this idea.

app/jobs/appointment_notification/schedule_experiment_reminders.rb
```diff
@@ -10,11 +10,18 @@ def perform
 
     notifications = Notification.due_today
     next_messaging_time = Communication.next_messaging_time
-
     logger.info "scheduling #{notifications.count} notifications that are due with next_messaging_time=#{next_messaging_time}"
     notifications.each do |notification|
       notification.status_scheduled!
       AppointmentNotification::Worker.perform_at(next_messaging_time, notification.id)
+    rescue => e
+      Sentry.capture_message("Scheduling notification failed",
```
harimohanraj89, line 19: Let's include the word "experiment" in this message somehow, since this is localized to the context of an experiment.

spec/models/patient_spec.rb
```diff
@@ -693,5 +693,22 @@ def refresh_views
...
+    end
+
+    it "soft deletes the patient's membership in any experiment" do
+      patient = create(:patient)
+      experiment = create(:experiment, :with_treatment_group)
+      treatment_group = experiment.treatment_groups.first
+      treatment_group.patients << patient
+
+      patient.discard_data
+      # TODO: Fails because TreatmentGroupMembership doesn't have a 'deleted_at' column, and is not an ApplicationRecord
```
harimohanraj89, line 710: Oof, this is a doozy. Are you thinking we destroy the record instead, or add the necessary timestamp?

app/models/patient.rb
```diff
@@ -227,6 +227,8 @@ def discard_data
     phone_numbers.discard_all
     prescription_drugs.discard_all
     teleconsultations.discard_all
+    notifications.discard_all
+    treatment_group_memberships.discard_all
```
rsanheim, line 231: Feels to me like we should `delete` or `destroy` these records rather than `discard`.  There isn't much reason to soft delete a notification record.

PR #2830, opened 2021-08-08
Only use for_end_of_month if present; add sentry context

rsanheim:
It appears we are getting some cases where we have an empty end_of_month
params..this will stop trying to parse those and also add some extra
sentry context to help debugging.

**Story card:** [ch4457](https://app.clubhouse.io/simpledotorg/story/4457/argumenterror-invalid-date-for-drugstock-controller)

Sentry issue: https://sentry.io/organizations/resolve-to-save-lives/issues/2561940349/?referrer=Clubhouse


app/controllers/concerns/set_for_end_of_month.rb
```diff
@@ -3,7 +3,17 @@ def set_for_end_of_month
     # Defaults to current month if it's the last week of the month.
     @show_current_month = ((Date.current.end_of_month - Date.current).to_i <= 7)
 
-    @for_end_of_month ||= if params[:for_end_of_month]
+    Sentry.configure_scope do |scope|
```
harimohanraj89, line 6: This is set for the scope of this request cycle?

app/controllers/concerns/set_for_end_of_month.rb
```diff
@@ -3,7 +3,17 @@ def set_for_end_of_month
...
+      user = @current_admin || @current_user
```
harimohanraj89, line 7: Is it easy for us to pull this out and apply it to every controller action under `AdminController`?

harimohanraj89 (commented):
Are we adding the Sentry stuff because we're not sure if a blank param is the problem?

app/controllers/concerns/set_for_end_of_month.rb
```diff
@@ -3,7 +3,17 @@ def set_for_end_of_month
...
+      scope.set_user(id: user.id)
```
prabhanshuguptagit, line 8: It seems like we don't have the user id for any dashboard controllers. Shall we just add it at `AdminController`?

app/controllers/concerns/set_for_end_of_month.rb
```diff
@@ -3,7 +3,17 @@ def set_for_end_of_month
     # Defaults to current month if it's the last week of the month.
```
prabhanshuguptagit, line 3: The `.present?` seems to fix the issue, why are we adding the Sentry information?

rsanheim: I'm not sure that is the issue here...I want to have more context if this error (or others) show up in Sentry.

app/controllers/concerns/set_for_end_of_month.rb
rsanheim: Yup, I believe so.  The sentry API is confusing now about this, but this should be for the length of the request.

app/controllers/concerns/set_for_end_of_month.rb
rsanheim: Not easy enough where I want to do it in this PR 😁 ...this module gets mixed into Admin and API controllers, so I want to keep this change isolated.

harimohanraj89: Fair enough 😅

harimohanraj89 (approved):
Coolcool let's learn about this error 🚀

PR #2831, opened 2021-08-08
Skip patients who do not have a mobile phone number

rsanheim:
This must be some race condition where we get a patient into notifications where their phone number gets removed or perhaps isn't valid.

Also set patient_id for sentry context

I suppose this could also be the case where this returns nil due to some parsing issue:

```
  def number_with_country_code
    parsed_number = Phonelib.parse(number, Rails.application.config.country[:abbreviation]).raw_national
    default_country_code = Rails.application.config.country[:sms_country_code]
    default_country_code + parsed_number
  end
```

**Story card:** [ch4458](https://app.clubhouse.io/simpledotorg/story/4458/patients-missing-phone-numbers-cause-notification-error)
**Sentry error:** https://sentry.io/organizations/resolve-to-save-lives/issues/2529471336/?project=1217715

### DIFF
--- a/app/jobs/appointment_notification/worker.rb
+++ b/app/jobs/appointment_notification/worker.rb
@@ -24,6 +24,9 @@ class AppointmentNotification::Worker
     end
 
     notification = Notification.includes(:subject, :patient).find(notification_id)
+    patient = notification.patient
+    Sentry.set_tags(patient_id: patient.id)
+
     communication_type = notification.next_communication_type
     unless communication_type
       logger.info "skipping notification #{notification_id}, no next communication type"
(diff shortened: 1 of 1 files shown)


kpethtel (approved):
AFAICT this is only happening on sandbox, and I believe I've investigated it before and come to the conclusion that it's due to non-india phone numbers on sandbox that  (as you suspected) cause `number_with_country_code` to return nil. Regardless, this seems like a safe fix to add.

rsanheim:
> AFAICT this is only happening on sandbox, and I believe I've investigated it before and come to the conclusion that it's due to non-india phone numbers on sandbox that (as you suspected) cause `number_with_country_code` to return nil. Regardless, this seems like a safe fix to add.

The linked sentry error above has quite a few of these happening on production.  [Here is a link to a list of them](https://sentry.io/organizations/resolve-to-save-lives/issues/2529471336/events/?environment=production&project=1217715).

We may want to add some more sentry context to figure out what is happening with the skipped numbers if we see a lot of these.

PR #2832, opened 2021-08-09
Facility upload: throw duplicate facility error if facility names differ only by whitespace

vkrmis:
**Story card:** [ch4439](https://app.clubhouse.io/simpledotorg/story/4439/facility-upload-breaks-if-facilities-with-same-name-exist-with-spaces)

## Because

`duplicate_rows` does not throw an error when facility names differ by whitespace, however the transaction fails when the slugs aren't unique.

## This addresses

`duplicate_rows` throws an error for facility names that only differ by whitespace.


app/validators/csv/facilities_validator.rb
```diff
@@ -29,7 +29,9 @@ def at_least_one_facility
   end
 
   def duplicate_rows
-    fields = facilities.map { |facility| facility.slice(:organization_name, :facility_group_name, :name) }
+    fields = facilities
+      .map { |facility| facility.slice(:organization_name, :facility_group_name, :name) }
+      .map { |facility| facility[:name] = facility[:name].gsub(/\s+/, "") }
```
prabhanshuguptagit, line 34: Wouldn't this `map` return just a list of facility names? The other attributes will not be uniqued by

```
> [{a: 1}, {b: 2}].map { |a| a[:name] = "Some name" }
=> ["Some name", "Some name"]
```
```

Comment (vkrmis) [reply app/validators/csv/facilities_validator.rb:34]: d'oh

File: app/validators/csv/facilities_validator.rb [hunk 2]
```diff
@@ -29,7 +29,12 @@ def at_least_one_facility
   end
 
   def duplicate_rows
-    fields = facilities.map { |facility| facility.slice(:organization_name, :facility_group_name, :name) }
+    fields = facilities
+      .map { |facility| facility.slice(:organization_name, :facility_group_name, :name) }
+      .map do |facility|
+        facility[:name] = facility[:name].gsub(/\s+/, "")
+        facility
+      end
```
harimohanraj89, line 37: Now that we've got more logic in here, we could simplify this a little bit perhaps, using a single map.

```ruby
facilies.map { |facility|
  organization_name: facilities[:organization_name],
  facility_group_name: facilities[:facility_group_name],
  name: facilities[:name].gsub(/\s+/, "")
}
```

PR #2833, opened 2021-08-09
Move the Statsd sidekiq middleware earlier in the Sidekiq chain

rsanheim:
This will hopefully initialize the datadog / statsd message queue and
avoid this error. See the Datadog code here:

https://github.com/DataDog/dogstatsd-ruby/blob/ef41904b914346bcd6b3a2c682c8d4c7f6e6fe10/lib/datadog/statsd/sender.rb

This also adds an explicit flush to our sidekiq middleware stats to make sure every job in every thread sends metrics when job processing is complete.

**Story card:** [ch4476](https://app.clubhouse.io/simpledotorg/story/4476/datadog-missing-metrics-from-jobs)


harimohanraj89 (approved):
Ohhh, nice idea. Yup let's try it. The problem is intermittent, so I guess we'll have to keep an eye on that exception over time to confirm if this worked.

rsanheim:
Here is the current order (in dev) - we may want to move it before Sentry middleware, but at least now its before any of our application code middleware:

```
2021-08-09T17:54:09.146Z pid=53772 tid=ovuwkcs7w DEBUG: Client Middleware: Datadog::Contrib::Sidekiq::ClientTracer
2021-08-09T17:54:09.147Z pid=53772 tid=ovuwkcs7w DEBUG: Server Middleware: Sentry::Sidekiq::SentryContextServerMiddleware, RequestStore::Sidekiq::ServerMiddleware, Datadog::Contrib::Sidekiq::ServerTracer, Sidekiq::Statsd::ServerMiddleware, SidekiqMiddleware::SetLocalTimeZone, SidekiqMiddleware::FlushMetrics, Sidekiq::Throttled::Middleware
```

app/jobs/tracer_job.rb
```diff
@@ -3,6 +3,7 @@ class TracerJob
   sidekiq_options retry: false # job will be discarded if it fails
 
   def perform(submitted_at, raise_error)
+    Statsd.instance.increment("tracer_job.count")
```
harimohanraj89, line 6: This will help us confirm that our flushing works?

lib/sidekiq_middleware/flush_metrics.rb
```diff
@@ -0,0 +1,11 @@
+# Ensure that metrics are flushed to Datadog after the execution of jobs.
+# See https://github.com/DataDog/dogstatsd-ruby/blob/master/examples/sidekiq_example.rb for more details.
```
harimohanraj89, line 2: This link is very helpful 👍

harimohanraj89 (approved):
Looks great 👌

app/jobs/tracer_job.rb
rsanheim: Yup, I want to see metrics coming in every 2 minutes from the schedule tracer job.

rsanheim:
This is looking better!

image

PR #2838, opened 2021-08-10
3301 experiment data

kpethtel:
**Story card:** [ch3301](https://app.clubhouse.io/simpledotorg/story/3301/as-a-researcher-i-want-to-have-an-export-of-data-from-experiments)

## Because

We need to be able to export experiment data. 

Refer to this document for an example of what we want out this: https://docs.google.com/spreadsheets/d/1pjuNf5sqJGfKl2Bo5sKniag64kKI-73DlHuc-0jB6zo/edit#gid=0

I've had to make a number of changes, mostly to accommodate the fact that a patient can have multiple appointments within the experiment.


app/services/experimentation/data_export.rb
```diff
@@ -0,0 +1,195 @@
+module Experimentation
+  class DataExport
+    require "csv"
+
+    FOLLOWUP_CUTOFF = 10.days
```
kpethtel, line 5: I believe this was the cutoff I discussed with Dr. Gupta. We should probably get clarity. At present, only using it to determine when to stop looking for encounters after the experiment is over.

app/services/experimentation/data_export.rb
```diff
@@ -0,0 +1,195 @@
...
+    def pad_communications
+      aggregate.each do |patient_data|
+        communications_deficit = @max_communications - patient_data[:communications].count
+        communications_deficit.times do
+          patient_data[:communications] << Array.new(4, nil)
+        end
+      end
+    end
+
+    def process_appointments(patient, notifications)
```
kpethtel, line 82: This is the most complex and ambiguous portion of the export because we don't have a clear connection between appointment and encounter. For a first pass, I've just selected all encounters from the first day of notifications til ten days after experiment completion. This should probably be refined to only look for encounters within a certain date range of the appointment. Per my comment above, I think we discussed a ten day cutoff after the appointment, but I don't recall any discussion of a cutoff before the appointment.

app/services/experimentation/data_export.rb
```diff
@@ -0,0 +1,195 @@
...
+      pd = patient.prescription_drugs.find_by(device_created_at: date_range)
+      return pd if pd
+      # this shouldn't happen. consider raising error
+    end
+
+    def headers
+      keys = aggregate.first.keys
+      keys.each_with_object([]) do |key, collection|
+        case key
+        when :communications
```
kpethtel, line 158: The majority of the complexity comes from the flexible number of columns in the document and the need to pad after max length has been found.

app/services/experimentation/data_export.rb
```diff
@@ -0,0 +1,195 @@
...
+        bp_at_followup = nil
+        facility = nil
+
+        if followup_date
+          days_til_followup = (appt.scheduled_date - followup_date).to_i
+          encounter = encounter_by_date(patient, followup_date)
+          bp_at_followup = encounter.class == BloodPressure
+          facility = encounter.facility
+        end
+        [appt.device_created_at.to_date, appt.scheduled_date, followup_date, days_til_followup, bp_at_followup,
```
kpethtel, line 106: Using arrays instead of hashes hurts readability slightly but also reduces some complexity. I think the values being added are clear enough, but it's still not ideal.

app/services/experimentation/data_export.rb
```diff
@@ -0,0 +1,195 @@
...
+    require "csv"
+
+    FOLLOWUP_CUTOFF = 10.days
+
+    attr_reader :experiment, :max_communications, :max_appointments, :max_past_visits, :notification_start_date, :aggregate, :cutoff_date
+
+    def initialize(name)
+      @experiment = Experimentation::Experiment.find_by!(name: name)
+      remind_ons = experiment.reminder_templates.pluck(:remind_on_in_days)
+      @notification_start_date = experiment.start_date - remind_ons.min.days
```
kpethtel, line 12: Using this date to catch encounters that occur before the experiment start date. For example, if someone received a "day before appointment" notification for an appointment on the first day of the experiment and came in same day. Frankly, I'd like to refine the experiment's concept of start/end dates. It would be better to have `experiment.notification_start_date` and `experiment.appointment_start_date` or something like that, given the difficulty we experienced even talking about this.

spec/services/experimentation/data_export_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+      purpose: :experimental_appointment_reminder,
+      remind_on: appt.scheduled_date + template.remind_on_in_days.days,
+      reminder_template: template,
+      status: "sent",
+      subject: appt
+    )
+  end
+
+  describe "#as_csv" do
+    it "exports accurate data in the expected format" do
```
kpethtel, line 21: These specs are both tedious and incomplete, but they did help me find some bugs.

app/services/experimentation/data_export.rb
```diff
@@ -0,0 +1,214 @@
...
+          end
+        when :encounters
+          (1..@max_past_visits).step do |i|
+            collection << "Encounter #{i} Date"
+          end
+        else
+          collection << key.to_s.split("_").each(&:capitalize!).join(" ")
+        end
+      end
+    end
```
rsanheim, line 188: This is, uh, confusing 😁 ...I guess this is handling the case where there are 1 to many amounts of communications and appointments? This is really a case where we **will** want to have hashes, and then the CSV writing can be a bit smarter and write out the key/values from the hash intelligently as it goes.  This type of CSV format will be harder for researchers to work with, so we should ask some more questions about it.

One thought -- we should separate the act of returning the data from an experiment (the queries) from the steps required for writing that data out to a CSV.  This will make each piece easier to test and reason about, and will also make it easy for us to ensure correctness of each piece. We can still add some common code (constants, exceptions, whatever) to make sure things are tied together and all required values are returned...but lets definitely break apart the two steps into at least two objects.

spec/services/experimentation/data_export_spec.rb
```diff
@@ -0,0 +1,218 @@
+require "rails_helper"
+
+TIME_FORMAT = "%Y-%m-%d"
```
rsanheim, line 3: Lets move this into a module so its not top level.

rsanheim (commented):
Looks like a good start.  Left some initial thoughts below -- I'll throw some time on your calendar tomorrow for us to pair on this.

app/jobs/experimentation/data_export_worker.rb
```diff
@@ -0,0 +1,64 @@
...
+    include Sidekiq::Worker
+
+    attr_reader :experiment_name, :recipient_email_address, :results
+
+    def perform(experiment_name, recipient_email_address)
+      @experiment_name = experiment_name
+      @recipient_email_address = recipient_email_address
+      fetch_results
+      mail_csv
+    end
```
rsanheim, line 13: Just to recap what we discussed on our call...I think it would be good if this call were just the 'traffic cop' here -- maybe something like this:

```
def perform
  @experiment = Experiment.find_by!(name: name)
  csv = Experimentation::Export.call(@experiment)
  Experimentation::Mailer.deliver_csv(csv, recipient)
end
```

The export doesn't need to know about the recipient or the mail, and the mail doesn't need to know about how the data is generated or queried.

app/jobs/experimentation/data_export_worker.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    private
+
+    def mail_csv
+      mailer = ApplicationMailer.new
```
rsanheim, line 18: Should probably be its own mailer object.

spec/jobs/experimentation/data_export_worker_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      attachment_data = {
+        mime_type: "text/csv",
+        content: expected_file_contents
+      }
+      filename = @experiment.name.tr(" ", "_") + ".csv"
+
+      attachments_double = double("attachments")
+      email_double = double("email", attachments: attachments_double, deliver: nil)
+      mailer = instance_double(ApplicationMailer, mail: email_double)
+      allow(ApplicationMailer).to receive(:new).and_return(mailer)
```
rsanheim, line 46: too much mocking here, but that can be unraveled as we go if it helps to get to a first draft quicker

rsanheim (approved):
Looks like we will use Sendgrid on SBX, though I don't know off hand if that will send actual emails:

https://github.com/simpledotorg/simple-server/blob/5bb5ec86ddaec7336bb8df494388e5126bf2070e/config/environments/production.rb

app/jobs/experimentation/data_export_worker.rb
rsanheim: Oh, that would also let you write the CSV to disk on SBX pretty easily for inspection if emails weren't being sent from there.

PR #2839, opened 2021-08-10
Add visited_without_bp / missed_visits to Reporting SchemaV2

rsanheim:
* grab these numbers from schema v2 and handle the delegation
* rename the _rate method to _rates to be consistent with everything
else
* ~~add a calculated sum which we need to get visited_without_bp under
care AND LTFU~~ this was incorrect and not necessary

**Story card:** [ch4515](https://app.clubhouse.io/simpledotorg/story/4515/grab-missed-visits-and-visited-w-o-bp-from-reporting-pipeline)


app/assets/javascripts/common/reports.js
```diff
@@ -662,7 +662,7 @@ Reports = function (withLtfu) {
       uncontrolledRate: jsonData.uncontrolled_patients_rate,
       uncontrolledWithLtfuRate: jsonData.uncontrolled_patients_with_ltfu_rate,
       visitButNoBPMeasure: jsonData.visited_without_bp_taken,
-      visitButNoBPMeasureRate: jsonData.visited_without_bp_taken_rate,
+      visitButNoBPMeasureRate: jsonData.visited_without_bp_taken_rates,
```
rsanheim, line 665: Getting names more consistent here.

app/controllers/reports/regions_controller.rb
```diff
@@ -73,8 +73,8 @@ def details
     else
       [@region, @region.facility_regions].flatten
     end
-    @repository = Reports::Repository.new(regions, periods: @period_range)
-    chart_repo = Reports::Repository.new(@region, periods: chart_range)
+    @repository = Reports::Repository.new(regions, periods: @period_range, reporting_schema_v2: RequestStore[:reporting_schema_v2])
+    chart_repo = Reports::Repository.new(@region, periods: chart_range, reporting_schema_v2: RequestStore[:reporting_schema_v2])
```
rsanheim, line 77: Realized these calls weren't passing thru the v2 flag - fixing here.

app/models/reports/repository.rb
```diff
@@ -109,69 +111,5 @@ def reporting_schema_v2?
...
-        periods_with_data = periods.select { |period| period >= earliest_period }
-        results.concat(periods_with_data.to_a.map { |period| [region, period] })
-      end
-      combinations.map { |region, period| Reports::RegionPeriodEntry.new(region, period, calculation, options) }
-    end
-
-    def percentage(numerator, denominator)
-      return 0 if denominator == 0 || numerator == 0
-      ((numerator.to_f / denominator) * 100).round(PERCENTAGE_PRECISION)
-    end
```
rsanheim, line 175: none of this is needed anymore

app/models/reports/schema_v2.rb
```diff
@@ -202,6 +220,7 @@ def sum(region, field)
     end
 
     def summary_field(field)
+      raise ArgumentError, "field #{field} is not part of the FacilityState query" unless field.in?(FIELDS)
```
rsanheim, line 223: This helped catch some things in dev/test with a slightly better error.

spec/models/reports/repository_spec.rb
```diff
@@ -303,24 +303,55 @@ def refresh_views
...
+            expect(rates_without_ltfu[period]).to eq(0)
+            if v2_flag # ignore bug in v1
+              expect(counts_with_ltfu[period]).to eq(1)
+              expect(rates_with_ltfu[period]).to eq(33)
+            end
+          }
+          ("Sept 2020".to_period.."Jan 2021".to_period).each { |period|
+            expect(counts_without_ltfu[period]).to eq(0)
+            expect(counts_with_ltfu[period]).to eq(0)
+          }
```
rsanheim, line 354: This was the best I got to to make this specs readable while still covering a wide range of periods and data.

spec/models/reports/repository_spec.rb
```diff
@@ -391,61 +422,53 @@ def refresh_views
...
+            expect(result_without_ltfu[period]).to eq(0)
+            expect(result_with_ltfu[period]).to eq(2)
+          }
+          ("Apr 2019".to_period.."June 2019".to_period).each { |period|
+            expect(result_without_ltfu[period]).to eq(0)
+            expect(result_with_ltfu[period]).to eq(1)
+          }
+          ("July 2019".to_period.."Jan 2020".to_period).each { |period|
+            expect(result_without_ltfu[period]).to eq(1)
+            if v2_flag # handle bug in v1 code
```
rsanheim, line 466: Just accounting for buggy behavior in v1 code here.

rsanheim (commented):
Ready for review - leaving some notes inline for context

app/models/reports/schema_v2.rb
```diff
@@ -126,6 +126,7 @@ def initialize(regions, periods:)
       end
     end
 
+    alias_method :missed_visits_rate, :missed_visits_rates
     alias_method :missed_visits_without_ltfu, :missed_visits
     alias_method :missed_visits_without_ltfu_rates, :missed_visits_rates
```
ssrihari, line 131: `:missed_visits_rates` seems to be an alias for 2 methods?

app/models/reports/schema_v2.rb
```diff
@@ -137,6 +138,23 @@ def missed_visits_with_ltfu_rates
       missed_visits_rates(with_ltfu: true)
     end
 
+    memoize def visited_without_bp_taken(with_ltfu: false)
+      if with_ltfu
+        regions.each_with_object({}) { |region, hsh| hsh[region.slug] = sum(region, :adjusted_visited_no_bp_lost_to_follow_up) }
```
ssrihari, line 143: Hey, I think the definition here and other ltfu/non-ltfu queries may be incorrect. The `_lost_to_follow_up` numbers in the facility states table has the total for **only** lost to follow up patients, not ltfu+under_care, which is probably what we want here. Do our tests cover this well?

So we should be either summing up the under_care and ltfu numbers here, or exposing ltfu+under_care numbers from the table. In previous discussions about this, I suggested exposing ltfu, and under_care numbers separately so that we can use them as building blocks here in ruby land. Exposing ltfu+under care numbers from the DB would be quite rigid in terms of design.

app/models/reports/schema_v2.rb
```diff
@@ -137,6 +138,23 @@ def missed_visits_with_ltfu_rates
...
+      else
+        regions.each_with_object({}) { |region, hsh| hsh[region.slug] = sum(region, :adjusted_visited_no_bp_under_care) }
+      end
+    end
+
+    memoize def visited_without_bp_taken_rates(with_ltfu: false)
+      region_period_cached_query(__method__, with_ltfu: with_ltfu) do |entry|
```
ssrihari, line 150: Same question [as in previous PR](https://github.com/simpledotorg/simple-server/pull/2767/files#r684976196): Is this is for convenience, or do we need caching here?

app/models/reports/schema_v2.rb
ssrihari: Also, `missed_visits` [defined above](https://github.com/simpledotorg/simple-server/pull/2839/files#diff-e95110f3347ec2a8767b4146482eff9bf0685c7ae55302c630070cbe3411084bR116) does `field = with_ltfu ?`.. it would be good to be consistent in structuring these methods.

config/initializers/cache_stats.rb
```diff
@@ -1,6 +1,6 @@
-ActiveSupport::Notifications.subscribe(/cache_.*\.active_support/) do |name, start, finish, arg1, arg2|
-  Rails.logger.debug name: name, start: start, finish: finish, arg1: arg1, arg2: arg2
-end
+# ActiveSupport::Notifications.subscribe(/cache_.*\.active_support/) do |name, start, finish, arg1, arg2|
+#   Rails.logger.debug name: name, start: start, finish: finish, arg1: arg1, arg2: arg2
+# end
```
ssrihari, line 3: Was this a debugging change?

ssrihari (changes requested):
Only one blocking change around the definition of the ltfu numbers.

app/models/reports/schema_v2.rb
rsanheim: Yeah, I wanted to move to a plural `rates` method names, but `missed_visit_rate` is currently used in too many places to change, hence the alias here so that I can start changing it incrementally in another PR.

config/initializers/cache_stats.rb
rsanheim: It gets too noisy to have this enabled all the time in dev, but it is very useful for cache debugging and monitoring in dev.  I'll had some better comments or tooling around this to make that more clear.

app/models/reports/schema_v2.rb
rsanheim: I actually had the summing code here and then removed it after working thru tests and reconsidering if we needed it.  Now with fresh eyes I think you are right that we need it.  If we do need it, I think we should add it to the facility_states matview to keep the app side code straightforward...the additional logic to sum both fields got a little ugly and also made everything just inconsistent enough in SchemaV2 to be frustrating.

I'm going to reasses this and will open a PR to add the summed columns if need be.

rsanheim: @ssrihari Okay, I handled this in code now and updated all the specs. This also gets the numbers much more in line when viewing things in reports, so I think this is good to go after an approval.

We can add it to the matviews next week if we want, though the code to add it was not too bad after all.

app/models/reports/schema_v2.rb
rsanheim: I think we will want some caching, when I compare this to v1 code its slightly slower for very large production like data sets.  But I think we can revisit performance a bit later (see https://app.clubhouse.io/simpledotorg/story/4516/benchmark-reports-comparing-v1-to-v2).

rsanheim:
@ssrihari This is ready for another pass.

app/models/reports/schema_v2.rb
```diff
@@ -5,17 +5,25 @@ class SchemaV2
...
+      monthly_registrations
+    ].freeze
+
+    UNDER_CARE_WITH_LTFU = %i[
+      adjusted_missed_visit
+      adjusted_visited_no_bp
     ].freeze
+    UNDER_CARE_WITH_LTFU_CALCULATED_FIELDS = UNDER_CARE_WITH_LTFU.map { |field|
+      "#{field}_under_care_with_lost_to_follow_up".to_sym
+    }
```
kpethtel, line 26: Could throw on a freeze here for consistency.

spec/models/reports/repository_spec.rb
```diff
@@ -454,7 +469,7 @@ def refresh_views
 
         it "creates cache keys" do
           repo = Reports::Repository.new(facility_1, periods: Period.month("June 1 2019")..Period.month("Jan 1 2020"))
-          cache_keys = repo.send(:cache_entries, :controlled).map(&:cache_key)
+          cache_keys = repo.schema.send(:cache_entries, :controlled).map(&:cache_key)
```
kpethtel, line 472: These changes to testing the schema were the one thing that threw me here. It seems like this approach to testing would have worked before but maybe wasn't a critical distinction until now.

kpethtel (approved):
This is difficult to review as someone who doesn't regularly work in this area of code, but I don't see any issues so 🚢

app/models/reports/schema_v2.rb
```diff
@@ -202,23 +225,34 @@ def sum(region, field)
...
       "sum_#{field}"
     end
 
+    # For some fields we need to sum the under_care numbers with the LTFU numbers to support the "include LTFU"
+    # toggle in reports.
+    def under_care_with_ltfu(field)
+      Arel.sql(<<-EOL)
+        SUM(COALESCE(#{field}_under_care::int, 0) + COALESCE(#{field}_lost_to_follow_up::int, 0)) as sum_#{field}_under_care_with_lost_to_follow_up
+      EOL
+    end
```
harimohanraj89, line 238: Ah I see, this is the crux of handling the LTFU toggle. Nice.

harimohanraj89 (approved):
I'm gonna lean a little bit on you to double-check the test coverage, but the big moving parts look great. :shipit:

app/models/reports/schema_v2.rb
rsanheim: 👍

spec/models/reports/repository_spec.rb
rsanheim: yeah these cache tests will probably go away soon, so I'm not too worried about them.

PR #2840, opened 2021-08-11
Validate uniqueness of facility slugs

rsanheim:
Validation tweak to ensure facility slugs are unique.  This will at least give us better error messages for things like [ch4439](https://app.clubhouse.io/simpledotorg/story/4439/facility-upload-breaks-if-facilities-with-same-name-exist-with-spaces)


prabhanshuguptagit (approved):
💯

PR #2842, opened 2021-08-12
Send the error message in the sentry notification for invalid logins

rsanheim:
This will generate unique error messages in Sentry so that we can better track what specific things are causing problems with authentication flow.


prabhanshuguptagit (approved):

harimohanraj89:
The fact that we want to track individual user login failures, and that we're manually shooting off Sentry events, makes me think Sentry's not the right place for this. Would a log that's sent to Datadog make more sense for these failures?

PR #2844, opened 2021-08-12
Enable block and state level reports for BD

claudiovallejo:
**Story card:** [ch4535](https://app.clubhouse.io/simpledotorg/story/4535/enable-state-and-block-level-reports-for-bangladesh)

## Because

We need to enable block and state-level reports for Bangladesh as we add 1,300+ new facilities.

## This addresses

Make `extended_region_reports` equal `true`

## Test instructions

Check that the `extended_region_reports` key is `true`


harimohanraj89 (approved):
:shipit:

rsanheim:
You'll need to update the failing tests for this.

Also, it would be good to change your local env to BD and just click around to test this out - it should be fine but doesn't hurt.

spec/models/region_spec.rb
```diff
@@ -61,24 +61,17 @@
...
     end
 
-    it "excludes states and blocks for other countries" do
-      allow(CountryConfig).to receive(:current).and_return(CountryConfig.for(:BD)).at_least(:once)
+    it "is everything for Bangladesh" do
+      expect(CountryConfig).to receive(:current).and_return(CountryConfig.for(:BD)).at_least(:once)
 
       org = Seed.seed_org.region
-      state = FactoryBot.create(:region, :state, reparent_to: org)
+      state = FactoryBot.create(:region, :state, reparent_to: Seed.seed_org.region)
```
rsanheim, line 68: You can use the local org here, just a tiny bit more clear.

```suggestion
      state = FactoryBot.create(:region, :state, reparent_to: org)
```

rsanheim (approved):
LGTM, one small suggestion

PR #2848, opened 2021-08-16
Admin access form performance

prabhanshuguptagit:
**Story card:** [ch4550](https://app.clubhouse.io/simpledotorg/story/4550)

## Because

We received reports of timeout errors on the admin access page on IHCI. This was likely happening because of the large number of facilities added recently.

## This addresses

- Adds a missing `includes` that speeds up the `visible_access_tree` method. This is a regression that got introduced in https://github.com/simpledotorg/simple-server/pull/2841. On a data set half the size of prod, this cuts down the request time from 28s to 13s.
- Use a hash to calculate if a facility's checkbox should be checked. We were checking with an `include?` on every facility, which is `O(n^2)` (for 5k facilities = 25M operations). The hash lookup is `O(n)`. This brings the time down from 13s to ~3 seconds.
- (investigating) The time for a full page render is currently at ~15s of which the DOM paint takes ~10s, only ~5s is AJAX. 
From some preliminary digging, it looks we are spending the majority of this time in the function [nodeListToArray](https://github.com/simpledotorg/simple-server/blob/9f7decab010743015335312de38d4a4804455abc/app/assets/javascripts/common/admins.js#L324)
image.

This PR fixes the timeout issue. The JS can be addressed in a follow up PR.

### DIFF
--- a/app/controllers/admins_controller.rb
+++ b/app/controllers/admins_controller.rb
@@ -139,6 +139,6 @@ class AdminsController < AdminController
   end
 
   def set_facilities_pre_checks(user)
-    user.visible_facilities.map(&:id).product([true]).to_h
+    @facilities_pre_checks = user.visible_facilities.map(&:id).product([true]).to_h
   end
 end
--- a/app/controllers/admins_controller.rb
+++ b/app/controllers/admins_controller.rb
@@ -28,8 +28,12 @@ class AdminsController < AdminController
         head(:not_found) && return
       end
 
-    user_being_edited = page_for_access_tree.eql?(:edit) ? AdminAccessPresenter.new(@admin) : nil
-    @facilities_pre_checks = user_being_edited && user_being_edited.visible_facilities.map(&:id).product([true]).to_h
+    if page_for_access_tree.eql?(:edit)
+      user_being_edited = AdminAccessPresenter.new(@admin)
+      set_facilities_pre_checks(user_being_edited)
+    else
+      user_being_edited = nil
+    end
 
     render partial: access_tree[:render_partial],
            locals: {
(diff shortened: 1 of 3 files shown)


app/controllers/admins_controller.rb
```diff
@@ -28,7 +28,12 @@ def access_tree
         head(:not_found) && return
       end
 
-    user_being_edited = page_for_access_tree.eql?(:edit) ? AdminAccessPresenter.new(@admin) : nil
+    if page_for_access_tree.eql?(:edit)
+      user_being_edited = AdminAccessPresenter.new(@admin)
+      set_facilities_pre_checks(user_being_edited)
```
harimohanraj89, line 33: We have to set this to an instance var?

app/models/user_access.rb
```diff
@@ -84,7 +84,7 @@ def accessible_facility_groups(action)
   end
 
   def accessible_facilities(action)
-    return Facility.all.includes(:region) if power_user?
+    return Facility.all.includes(:region, facility_group: :organization) if power_user?
```
harimohanraj89, line 87: Ohh, we were failing to run this `includes` when it mattered the most. Ouch. Nice find.

harimohanraj89 (changes requested):

app/controllers/admins_controller.rb
prabhanshuguptagit: Bad refactor 🤦🏼  Fixed in a56a16766

harimohanraj89 (approved):
Awesome.

app/models/user_access.rb
rsanheim: ✨

rsanheim: this is something that is possible (but annoying) to add specs for to prevent regressions for this sort of perf critical includes

PR #2851, opened 2021-08-17
Update zone label in sri lanka

prabhanshuguptagit:
**Story card:** [chXXXX](URL)

## Because

The block field will be unused in Sri Lanka. We can call the dashboard label for the field `Block` for now.


harimohanraj89 (approved):

harimohanraj89:
According to Meg's latest email, we can use Town! Let's do it.

PR #2854, opened 2021-08-18
ET - BP date correction

harimohanraj89:
**Story card:** [ch4544](https://app.clubhouse.io/simpledotorg/story/4544/et-koka-hc-bad-dates)

## Because

A user in Ethiopia was using an old version of Simple and entering ET dates into a Gregorian date picker. All these BPs were being stored as being recorded in 2012 and 2013, causing their recent activity to look empty.

## This addresses

Retroactively fixes the dates on those bad BPs by doing the following:
* Select the BPs from the facility that have dates in 2012 or 2013
* Update them using the conversion formula - same formula that was used in our import spreadsheet, and I've verified the same using some online tools as well.
* Update the patient's `recorded_at` to reflect their real registration date, instead of "2013"


vkrmis (commented):
Do we need to also update other timestamps (device_created_at, device_updated_at etc)

vkrmis (approved):
left a comment ^, looks good otherwise

rsanheim:
> A user in Ethiopia was using an old version of Simple and entering ET dates into a Gregorian date picker.

Do we have a mechanism for telling a client app that it is "too out of date" or something similiar - and then prompting to force an update? That would prevent problems like this in the future /cc @msasikanth @vinaysshenoy

harimohanraj89:
@rsanheim we do have some sanity checks like this (eg. age < 120, BP can't be ridiculously high or low). The date's tricky, because 2013 isn't too far in the past. And, in this case specifically, introducing a date check won't help because folks may be on an outdated version of the app that won't have our latest checks anyway 😅 

It's a good heads up for the mobile team, though, in case there are data points we wanna start adding checks to based on patterns of misuse we see in the field.

harimohanraj89:
@vkrmis I thought about this a bit. I think we should stick to only updating `recorded_at` since that's the only clinically relevant datapoint that we care about - ie, it affects reports. The device timestamps are... well, we _could_ change them, but I feel like it's better to leave them alone if we don't need to change them. They don't have any impact on the treatment history of the patient, and since the errant dates are in the past, there should be no problems for the app to detect that the record has been updated on the server, and sync the latest version, etc.

vkrmis:
@harimohanraj89 I think @rsanheim meant telling the user that the app version is too old or unsupported, and that they should go update the app.

harimohanraj89:
Ohhh I see I see - that seems quite doable, I'd think. A server-side config and app-version check, followed by an appropriate failure response code that the client can detect.

This sounds like a cool idea, we don't have any mechanism for this yet.

harimohanraj89:
I've added this problem to our list here: https://docs.google.com/document/d/1RqJ2UmPdyP_oyAg6vMk9moN298462ddnDDQhAvLqTnQ/edit#

PR #2855, opened 2021-08-18
Admin access JS performance improvements

prabhanshuguptagit:
**Story card:** [ch4550](https://app.clubhouse.io/simpledotorg/story/4550/admin-access-page-is-timing-out)

## Because

Follow up to #2848. The admin access page is very slow to load with a large number of facilities. On IHCI it takes roughly ~110 seconds to render the page.

## This addresses

The load time for first paint (~14k facilities) is roughly:

- 8 seconds - AJAX and server side rendering of the access tree markup from `/admins/<user_id>/access_tree/edit`
- < 2 seconds - downloading the response (size of the tree's markup is ~800kb)
- < 1 second - parsing the HTML
- ~90 seconds - JS time, painting the tree

This optimises a loop in JS which updates the parent checkbox's status based on the children's checkboxes.The following was being done recursively for each checkbox at the first paint:
- Find parent of each checkbox
- Find all children of the parent, see if all, some or none are checked
- Update the parent checkbox
- Do this for parent of current checkbox

This avoids calling the loop on all checkboxes. Instead we call it only once per parent. This cuts down the JS time to 3-4 seconds.

Before (with 5k facilities):
image

After (with 5k facilities): 
image


rsanheim:
Is this a follow up to https://github.com/simpledotorg/simple-server/pull/2848 ?

Please do add cross refence PR's to your description(s) for PRs like this.  It makes it much easier to track fixes that are broken down along several PRs for a single story or bug fix.  It makes it possible to see much more of the big picture via the PR cross references.

rsanheim (approved):
> The load time for first paint (~14k facilities) is roughly:
> 
> 8 seconds - AJAX and server side rendering of the access tree markup from /admins/<user_id>/access_tree/edit

Why is this part so low?  

> < 2 seconds - downloading the response (size of the tree's markup is ~800kb)
> < 1 second - parsing the HTML
> ~90 seconds - JS time, painting the tree
> 

Looks like a good interim fix here at least 👍 

Long term it seems like we may have to just render less at a time for this page, and do less at a time in our javascript....especially with 20k+ facilities on what I'm guessing can be slower laptop cpus.

prabhanshuguptagit:
> Why is this part so low?

This takes the most amount of time _now_, actually. The JS is fairly scalable now, and it scales linearly with the number of facility groups (not facilities), so we can consider it a non-problem for the medium-term. We are mostly bottlenecked by this time.

I'll be catching up with claudio to chalk out what fetching the tree asynchronously would look like.

PR #2859, opened 2021-08-19
Bugfix: Encounter follow ups don't show up on same day

prabhanshuguptagit:
**Story card:** [ch4414](https://app.clubhouse.io/simpledotorg/story/4414/et-zeros-in-progress-tab)

## Because

Daily follow up numbers on the progress tab number don't show up on the same day.

## This addresses

When a facility has both HTN and DM enabled, we fetch daily numbers by comparing `encounters.encountered_on`. However this is a date field, not a datetime field, and the date range comparison excludes today's results. This happens because groupdate tries to convert the time zone of the date. The [recommended way](https://github.com/ankane/groupdate#date-columns)  to avoid the time zone conversion is to pass in a  `time_zone: false` to groupdate.


app/models/patient.rb
```diff
@@ -79,15 +79,15 @@ class Patient < ApplicationRecord
   scope :search_by_address, ->(term) { joins(:address).merge(Address.search_by_street_or_village(term)) }
 
   scope :follow_ups_by_period, ->(period, at_region: nil, current: true, last: nil) {
-    FollowUpsQuery.with(Encounter, period, at_region: at_region, current: current, time_column: "encountered_on", last: last)
+    FollowUpsQuery.new.with(Encounter, period, at_region: at_region, time_column: "encountered_on", current: current, last: last, time_zone: false)
```
harimohanraj89, line 82: The `time_zone: false` is the crux of the bugfix. The rest is cleanup and refactoring.

app/queries/follow_ups_query.rb
```diff
@@ -1,33 +1,27 @@
...
-      sum_groups_per_period(results)
+      query.group(group_by).count.then { |results| sum_groups_per_period(results) }
     else
       query.count
     end
   end
 
-  def self.with(model_name, period, time_column: "recorded_at", at_region: nil, current: true, last: nil)
-    raise ArgumentError, "Only day, month and quarter allowed" unless period.in?([:day, :month, :quarter])
+  def with(model, period_type, time_column: "recorded_at", at_region: nil, **groupdate_opts)
```
harimohanraj89, line 12: The key change here is accepting a blob of options to pass transparently to groupdate.

spec/queries/follow_ups_query_spec.rb
```diff
@@ -129,6 +129,19 @@
                                   second_follow_up_date => 1})
         end
 
+        it "counts encounters created today when last is supplied and Groupdate time zone is not UTC" do
+          with_reporting_time_zone do
```
rsanheim, line 133: Can you also freeze time here? Otherwise this will fail around day boundaries for certain time zones.

app/queries/follow_ups_query.rb
```diff
@@ -1,33 +1,27 @@
...
   def hypertension(region, period_type, group_by: nil)
-    query = Patient.joins(:blood_pressures)
-      .where("patients.recorded_at < #{BloodPressure.date_to_period_sql("blood_pressures.recorded_at", period_type)}")
-      .group_by_period(period_type, "blood_pressures.recorded_at", format: Period.formatter(period_type))
-      .distinct
-      .where(blood_pressures: {facility_id: region.facility_ids})
-      .with_hypertension
+    query = with(BloodPressure, period_type, at_region: region, format: Period.formatter(period_type)).with_hypertension
+
     if group_by.present?
```
rsanheim, line 5: 👍

rsanheim (approved):
Approving to unblock, but one thing about making the test consistent.

harimohanraj89:
@claudiovallejo you are tagged since you're a codeowner and this touches an indicator's query. I'm dismissing your review since this is a bugfix, and no definitions are being changed.

Good to know this thing works!

PR #2861, opened 2021-08-19
Create the CallResult model and APIs

vkrmis:
**Story card:** [ch4615](https://app.clubhouse.io/simpledotorg/story/4615/decide-api-schema-share-with-mobile-devs)

## Because

## This addresses

## Test instructions


app/schema/api/v4/models.rb
```diff
@@ -33,6 +33,10 @@ def array_of(type)
        items: {"$ref" => "#/definitions/#{type}"}}
     end
 
+    def nullable_enum(enum_values)
+      {type: [:string, "null"], enum: enum_values << nil}
+    end
```
harimohanraj89, line 38: Toit.

app/controllers/api/v4/call_results_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    params.require(:call_results).map do |call_result_params|
+      call_result_params.permit(
+        :id,
+        :user_id,
+        :appointment_id,
+        :cancel_reason,
+        :result,
+        :deleted_at,
+        :created_at,
+        :updated_at
```
harimohanraj89, line 31: We do the automagic to shove these two timestamps into `device_*` fields, right?

app/models/call_result.rb
```diff
@@ -0,0 +1,24 @@
+class CallResult < ApplicationRecord
+  include Mergeable
+  belongs_to :appointment, optional: true
+  belongs_to :user, optional: true
```
harimohanraj89, line 4: Why optional?

app/models/call_result.rb
```diff
@@ -0,0 +1,24 @@
+class CallResult < ApplicationRecord
+  include Mergeable
+  belongs_to :appointment, optional: true
```
harimohanraj89, line 3: This one's iffy. I realize we can't make it required because the appointment might reach us later. Can we validate the presence of `appointment_id` though? I'd love for our models to follow our API's contract as closely as possible.

db/schema.rb
```diff
@@ -678,6 +690,7 @@
   add_foreign_key "blood_sugars", "facilities"
   add_foreign_key "blood_sugars", "users"
   add_foreign_key "communications", "notifications"
+  add_foreign_key "drug_stocks", "facilities"
```
harimohanraj89, line 693: Hello there, what's this?

db/migrate/20210813061906_create_call_results.rb
```diff
@@ -0,0 +1,16 @@
+class CreateCallResults < ActiveRecord::Migration[5.2]
+  def change
+    create_table :call_results, id: :uuid do |t|
+      t.uuid :user_id, null: false, foreign_key: true
+      t.uuid :appointment_id, null: false, foreign_key: true
```
harimohanraj89, line 5: My migration-fu is rusty. Does this leave enough slack for our referential integrity? ie, can we create a call result before we receive the corresponding appointment?

harimohanraj89 (commented):
My early thoughts below!

db/migrate/20210813061906_create_call_results.rb
rsanheim: Yeah, if we are restricting null here we should _not_ have the appointment be optional at the model level.

db/migrate/20210813061906_create_call_results.rb
```diff
@@ -0,0 +1,16 @@
...
+
+      t.string :cancel_reason
```
rsanheim, line 7: Should this be required?

app/models/call_result.rb
rsanheim: Not sure I follow this -- wouldn't the call result and appointment JSON come in the same payload from the client?

app/models/call_result.rb
```diff
@@ -0,0 +1,24 @@
...
+  belongs_to :user, optional: true
+
+  validates :device_created_at, presence: true
+  validates :device_updated_at, presence: true
```
rsanheim, line 7: Looks like we are missing a presence validation on `result`.

db/schema.rb
rsanheim: I think I've seen this FK come in and out of diffs over the past few weeks as well...not sure if the FK really belongs in our schema or not.

spec/api/v4/call_results_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      response "200", "call_results created" do
+        let(:request_user) { FactoryBot.create(:user) }
+        let(:request_facility) { FactoryBot.create(:facility, facility_group: request_user.facility.facility_group) }
+        let(:HTTP_X_USER_ID) { request_user.id }
+        let(:HTTP_X_FACILITY_ID) { request_facility.id }
+        let(:Authorization) { "Bearer #{request_user.access_token}" }
+
+        let(:call_results) { {call_results: (1..3).map { build_call_result_payload }} }
+
+        run_test!
```
rsanheim, line 21: shared specs really obscure what is actually being tested here...where would I look to see what is actually done in the test and what assertions are made ?

vkrmis:
@harimohanraj89, @rsanheim: I reopened this PR with some changes, notably renaming `result -> result_type`, and `cancel_reason -> remove_reason` to make the relationships clearer.

app/controllers/api/v4/call_results_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+  end
+
+  def call_result_params
+    params.require(:call_results).map do |call_result_params|
+      call_result_params.permit(
+        :id,
+        :user_id,
+        :appointment_id,
+        :cancel_reason,
```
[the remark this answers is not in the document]
vkrmis: `result_type` has three possible values: 

- agreed to visit
- remind to call later
- removed from overdue list

`remove_reason` has 7 possible values right now, and they all indicate reasons why a patient was `removed from overdue list`. Eg: `died`, `invalid phone number` etc.

Its better to model them separately rather than flattening them into a single enum where 7 of 9 values need to be interpreted as `removed from overdue list`

app/controllers/api/v4/call_results_controller.rb
vkrmis: Yes, this is done by the call to `Api::V4::Transformer.from_request` in `merge_if_valid`

app/models/call_result.rb
vkrmis: Good question, I did it out of reflex since all our sync models have optional `user` and `facility` relationships.

This makes sense when the relations are to syncable resources, but I'm not sure why we also do it for users and facilities. I'm okay to make it required and see if it breaks anything.

app/models/call_result.rb
vkrmis: @harimohanraj89 Can't do it cleanly, if we validate presence of appointment_id, we'd have to stop spec-ing the optional association (`it { should belong_to(:appointment).optional }`). Can't do both since the specs are contradictory.

We actually enforce this at the db instead, the column is required. This is how we generally model all our synced resources. 

@rsanheim Appointments and CallResults are synced separately, we're setting up a new sync API end point for CallResults.

app/models/call_result.rb
harimohanraj89: Yup let's be strict from the get go, avoid persisting bad data.

db/schema.rb
vkrmis: Fixed by #2960

db/migrate/20210813061906_create_call_results.rb
vkrmis: Yup, this is how we're able to enforce presence at the db, while allowing models to sync out of order at the API level

db/migrate/20210813061906_create_call_results.rb
vkrmis: @rsanheim restricting null at the db + having it optional at the model level is how we're able to sync resources out of order and still enforce presence, this is true of all our sync resources.

`cancel_reason` is not a required field, only applies to a subset of `CallResults`

spec/api/v4/call_results_spec.rb
vkrmis: Agree about the shared specs, they really get in the way here. However, since they enforce the shared sync api contracts, I don't feel comfortable not using them for just this API. Maybe we can make a wholesale switch away from them at some point, or find a different way to test the shared contract.

You can find the specific specs included by looking them up in: [shared_spec_for_sync_controller.rb](https://github.com/simpledotorg/simple-server/blob/master/spec/controllers/shared_examples/shared_spec_for_sync_controller.rb).

I managed to split up `sync_from_user` and `sync_to_user` specs so we can just test `sync_from_user` for this endpoint.

app/models/call_result.rb
vkrmis: We validate this at the db already, so we wouldn't persist bad data. Its just inconsistent, and unnecessary since users aren't a synced from the client.

app/models/call_result.rb
harimohanraj89: Hmm. I don't want to let shoulda-matchers dictate our implementation too much. The two specs aren't really contradictory, we want to say - you can have an appt id that points to nothing, but you can't have a blank appt id, right? I'd rather handroll that spec and rely on application-level validations, rather than letting it get down to the DB and raise an exception.

app/controllers/api/v4/call_results_controller.rb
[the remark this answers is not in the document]
harimohanraj89: Ah very cool very cool. Makes sense to me. Also `remove_reason` :chef_kiss:

db/schema.rb
harimohanraj89: Nice.

harimohanraj89:
The renames are great @vkrmis . Only one outstanding point from my review regarding our model validations.

vkrmis:
@rsanheim I'm going to go ahead and merge this, please leave a comment if I've missed anything. I'll tackle any other changes in a follow on PR.

PR #2862, opened 2021-08-19
Add Sri Lanka custom medications

prabhanshuguptagit:
**Story card:** [ch4628](https://app.clubhouse.io/simpledotorg/story/4628/add-custom-medications-to-sri-lanka)

## This addresses

Adds new custom medications for Sri Lanka.

### DIFF
--- a/config/data/medications.yml
+++ b/config/data/medications.yml
@@ -1025,3 +1025,299 @@
   rxnorm_code: '757711'
   drug_category: diabetes
   deleted: false
+- id: bf3f1200-aaf6-40d7-a229-ad6ed5f7c5a9
+  name: Amlodipine
+  composition:
+  dosage: 2.5 mg
+  frequency: two_per_day
+  rxnorm_code:
+  drug_category: hypertension_ccb
+  deleted: false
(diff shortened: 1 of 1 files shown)


harimohanraj89 (approved):
Tough to review but had a thought:

The sorting doesn't work great because sorting dosage strings leads to weird things like "10" < "5". You can see it in some of the entries below.

Additionally, if we wanna make review a bit easier, wanna decouple the sorting from the addition of new entries? That is, if you still want to try to get the sorting working as expected - maybe script + some manual attention?

prabhanshuguptagit:
I tried ordering by dosages too but it got hairy quickly (split on mg, order as integers etc). With the sorting one-liner I mostly wanted to enforce alphabetic-ness for when more drugs are added in the future. It's a bit hard to manually add 20-30 drugs putting them one at a time in the right place. The dosages aren't as important.

Will pull the sorting into a different PR, this is not reviewable 👍🏼

harimohanraj89 (approved):
Fair enough. Deterministic is better than "correct" in this case 👍 

Additions look great.

PR #2863, opened 2021-08-19
App manifest v2

prabhanshuguptagit:
**Story card:** [ch4630](https://app.clubhouse.io/simpledotorg/story/4630/add-a-v2-manifest-for-listing-multiple-deployments-in-a-country)

## Because

[PRD](https://docs.google.com/document/d/1ZAOQVgMl3ALuRdLbNzXVZBGv7F-yqPJZA6Su_rzTeKY/edit#heading=h.viflbiwh0b15)

## This addresses

Adds v2 manifests for listing the Country > Deployments hierarchy

### DIFF
--- a/public/manifest/demo.json
+++ b/public/manifest/demo.json
@@ -14,7 +14,7 @@
     },
     {
       "country_code": "ET",
-      "endpoint": "https://simple-demo.moh.gov.et/api/",
+      "endpoint": "https://api-demo.et.simple.org/api/",
       "display_name": "Ethiopia",
       "isd_code": "251"
     },
--- a/public/manifest/demo.json
+++ b/public/manifest/demo.json
@@ -24,5 +24,43 @@
       "display_name": "Sri Lanka",
       "isd_code": "94"
     }
+  ],
+  "v2": [
+    {
+      "country_code": "IN",
+      "display_name": "India",
+      "isd_code": "91",
+      "deployments": {
+        "display_name": "India",
+        "endpoint": "https://api-demo.simple.org/api/"
+      }
+    },
+    {
+      "country_code": "BD",
+      "display_name": "Bangladesh",
+      "isd_code": "880",
+      "deployments": {
+        "display_name": "Bangladesh",
+        "endpoint": "https://api-demo.bd.simple.org/api/"
+      }
+    },
+    {
+      "country_code": "ET",
+      "display_name": "Ethiopia",
+      "isd_code": "251",
+      "deployments": {
+        "display_name": "Ethiopia",
+        "endpoint": "https://simple-demo.moh.gov.et/api/"
+      }
+    },
+    {
+      "country_code": "LK",
+      "display_name": "Sri Lanka",
+      "isd_code": "94",
+      "deployments": {
+        "display_name": "Sri Lanka",
+        "endpoint": "https://api-demo.lk.simple.org/api/"
+      }
+    }
   ]
 }
--- a/public/manifest/demo.json
+++ b/public/manifest/demo.json
@@ -25,50 +25,52 @@
       "isd_code": "94"
     }
   ],
-  "v2": [
-    {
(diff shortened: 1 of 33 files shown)


public/manifest/demo.json
```diff
@@ -24,5 +24,43 @@
       "display_name": "Sri Lanka",
       "isd_code": "94"
     }
+  ],
```
rsanheim, line 27: It looks like we are now duplicating a ton of data for these `json` files across each env, with just the URL being duplicated.

How about rolling them up into a `v2` and `v1` partial or something like that, and then just calling that from a `manfiest.json.erb` partial?

I think we could page cache it later if this end point gets hit a lot...

prabhanshuguptagit: Spiked out partials in cd68eac. It's hard to compress this information any further without being clever because `v1` and `v2` are fundamentally different (only the URLs are same). Did you have something different in mind?

The number of calls for this end point is very low btw, so we shouldn't have to worry about caching for a _long_ time even if we do partials etc.

public/manifest/demo.json
```diff
@@ -24,5 +24,43 @@
...
+  "v2": [
+    {
+      "country_code": "IN",
+      "display_name": "India",
+      "isd_code": "91",
+      "deployments": {
```
harimohanraj89, line 33: We would need this to be an array to be extensible, right?

**Update:** Same comment for all non-prod manifests.

public/manifest/production.json
```diff
@@ -18,5 +18,44 @@
...
+      "display_name": "India",
+      "isd_code": "91",
+      "deployments": [
+        {
+          "display_name": "IHCI",
+          "endpoint": "https://api.simple.org/api/"
+        },
+        {
+          "display_name": "Kerala",
+          "endpoint": "https://api.kerala.simple.org/api/"
```
harimohanraj89, line 34: This doesn't exist yet, shall we not include it in the prod manifest until it is?

public/manifest/production.json
```diff
@@ -18,5 +18,44 @@
       "display_name": "Ethiopia",
       "isd_code": "251"
     }
+  ],
+  "v2": [
+    {
```
harimohanraj89, line 23: While we're introducing this, reckon we should prefix this entire payload with a `countries:`?

public/manifest/production.json
```diff
@@ -18,5 +18,44 @@
...
+      "country_code": "IN",
+      "display_name": "India",
+      "isd_code": "91",
+      "deployments": [
+        {
+          "display_name": "IHCI",
```
harimohanraj89, line 29: Not necessarily useful to the client, but good to keep 👍

public/manifest/production.json
```diff
@@ -18,5 +18,44 @@
...
       "isd_code": "251"
     }
+  ],
+  "v2": [
+    {
+      "country_code": "IN",
+      "display_name": "India",
+      "isd_code": "91",
+      "deployments": [
+        {
```
harimohanraj89, line 28: I envision a world where each deployment can start overriding things from its parent, like ISD code - that's not a great example, but the flexibility is great.

harimohanraj89 (changes requested):
Big note about inconsistent data structure.

Small suggestions otherwise.

public/manifest/demo.json
prabhanshuguptagit: Oops, thanks for catching 😅  Fixed in 6f4a443e4

public/manifest/production.json
prabhanshuguptagit: Removed

public/manifest/production.json
prabhanshuguptagit: Added in b0978a006

harimohanraj89 (approved):
Noice.

PR #2864, opened 2021-08-19
Facility search on user edit page

harimohanraj89:
**Story card:** [ch3936](https://app.clubhouse.io/simpledotorg/story/3936/facility-search-bar)

## Because

It's hard for power users to scroll through a dropdown of 10000 facilities to assign a user's facility

## This addresses

Adding a search-bar-powered dropdown using standard `bootstrap-select` tools.

### Desktop gif

desktop


prabhanshuguptagit (approved):
I had ~20k facilities locally to test with, it works decently well. There's an occasional hiccup when the field is cleared but it's minor. :shipit: :shipit:

PR #2866, opened 2021-08-20
States list API

prabhanshuguptagit:
**Story card:** [ch4631](https://app.clubhouse.io/simpledotorg/story/4631/add-an-api-to-list-available-states-in-a-deployment)

## Because

The app should be able to pick a state from the available states in a deployment.

## This addresses

Adds an API endpoint (`/api/v4/states`) that lists the available state names.


app/controllers/api/v4/states_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Api::V4::StatesController < APIController
+  include Api::V3::PublicApi
+
+  def sync_to_user
```
harimohanraj89, line 4: This doesn't feel like a `sync` endpoint to me. Can we just make a simple `/states` endpoint for this?

app/controllers/api/v4/states_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+    state_names = Region.facility_regions
+      .joins("inner join regions states ON states.path @> regions.path and states.region_type = 'state'")
+      .distinct("states.name")
+      .pluck("states.name")
```
harimohanraj89, line 8: What's the difference between this and `Region.state_regions`?

harimohanraj89 (changes requested):
Thoughts below!

app/controllers/api/v4/states_controller.rb
prabhanshuguptagit: I went back and forth between a `sync_to_user` and `index`. Does it make sense to make this a sync endpoint (similar to the medications controller) for consistency's sake?

app/controllers/api/v4/states_controller.rb
prabhanshuguptagit: `Region.state_regions` can have states that don't have any facility under them. To get states that have at least one facility we could:
- Start from `Region.state_regions` and join a distinct facility on each state, filter states that have a facility
- Start from `Region.facility_regions` and get their states

The latter seemed simpler to write.

harimohanraj89: Ah I see, the state regions are _every_ state in the country from the canonical list. Neat neat.

app/controllers/api/v4/states_controller.rb
harimohanraj89: Hmm, I'm kinda torn - I'd love to keep this uber simple. Wanna poke the team and see how we feel?

prabhanshuguptagit: Makes sense 👍🏼 . I'm also leaning towards a regular GET endpoint.

swagger/v3/swagger.json
```diff
@@ -2253,7 +2253,10 @@
             "voip_call",
             "manual_call",
             "sms",
-            "whatsapp"
+            "whatsapp",
+            "missed_visit_sms_reminder",
+            "missed_visit_whatsapp_reminder",
+            "imo"
```
harimohanraj89, line 2259: Some older updates that never made it to docs yet, I guess?

harimohanraj89 (approved):
:shipit:

PR #2867, opened 2021-08-20
Facility search on Overdue page

harimohanraj89:
Follow-up to #2864 

**Story card:** [ch3936](https://app.clubhouse.io/simpledotorg/story/3936/facility-search-bar)

## Because

The overdue page also has a giant facility list dropdown

## This addresses

Making it searchable so that a list of 10k facilities can be managed.

Screen Shot 2021-08-20 at 3 37 00 PM

### DIFF
--- a/app/views/layouts/overdue.html.erb
+++ b/app/views/layouts/overdue.html.erb
@@ -5,23 +5,13 @@
   <p>
     Patients that are overdue for a follow-up visit<% if current_admin.accessible_facilities(:manage_overdue_list) %>. If you select a facility, you can download a patient list.<% end %>
   </p>
-  <% if current_admin.accessible_facilities(:manage_overdue_list) %>
-    <%= render('shared/paginate_and_filter_by_facility',
-               form_url: appointments_path,
-               current_admin: current_admin,
-               facility_id: @facility_id,
-               per_page: @per_page,
-               form_id: "search-filters",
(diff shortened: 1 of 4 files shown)


app/views/layouts/overdue.html.erb
```diff
@@ -5,23 +5,13 @@
...
   </p>
-  <% if current_admin.accessible_facilities(:manage_overdue_list) %>
-    <%= render('shared/paginate_and_filter_by_facility',
-               form_url: appointments_path,
-               current_admin: current_admin,
-               facility_id: @facility_id,
-               per_page: @per_page,
-               form_id: "search-filters",
-               scope_namespace: current_admin.accessible_facilities(:manage_overdue_list)) %>
-  <% else %>
```
vkrmis, line 16: What was this `else` branch for?

vkrmis (approved):
I'm not sure why the `else` branch was there, looks good otherwise

app/views/layouts/overdue.html.erb
harimohanraj89: I temporarily force-executed the `else` branch, it errored out. The `scope_namespace` doesn't make sense for how it's used in the partial.

This might have been an access-related change, but this page will correctly show empty results if you don't have access to any facilities, or kick you off completely if you don't have PII access in the first place.

vkrmis: Hmm, odd

PR #2868, opened 2021-08-20
Point ET Demo to simple.org subdomain

harimohanraj89:
The moh.gov.in subdomain's ssl cert has expired, and we wish
to unblock a training session immediately

**Story card:** @harimohanraj89 will fill this stuff in later.


vkrmis (approved):
Looks good

PR #2869, opened 2021-08-23
3301 followup days

kpethtel:
**Story card:** [ch3301](https://app.clubhouse.io/simpledotorg/story/3301/as-a-researcher-i-want-to-have-an-export-of-data-from-experiments)

## Because

The emails seemed to be timing out, so I removed emails. I also re-added some functionality for calculating follow up days.


app/services/experimentation/export.rb
```diff
@@ -76,6 +83,56 @@ def experimental_communications(notifications)
...
+          .pluck(:scheduled_date)
+          .uniq
+      end
+    end
+
+    def active_patient_followups(patient, notifications)
+      appointment_dates = active_patient_experimental_appointment_dates(patient, notifications)
+      appointment_dates.map.each_with_index do |appointment_date, index|
+        adjusted_index = index + 1
+        followup_date_range = ((appointment_date - FOLLOWUP_START)..(appointment_date + FOLLOWUP_CUTOFF))
```
kpethtel, line 114: These date ranges are a bit arbitrary, but there is no clear relationship between an appointment and a blood pressure/encounter. The ten day cutoff is something I believe Reena suggested. The three days before... that's for the case that a patient comes in before their appointment date.

prabhanshuguptagit (approved):

PR #2871, opened 2021-08-24
Progress tab cache bug fix

prabhanshuguptagit:
**Story card:** [chXXXX](URL)

## Because

When DM is enabled for a facility, the cached progress tab results cause an error because they don't have any diabetes related information but the view expects it. Sentry: https://sentry.io/organizations/resolve-to-save-lives/issues/2598315876/?referrer=slack

## This addresses
Adds `diabetes_enabled?` to the cache key so the cache is updated when the DM status of a facility is changed.

### DIFF
--- a/app/presenters/user_analytics_presenter.rb
+++ b/app/presenters/user_analytics_presenter.rb
@@ -4,6 +4,7 @@ class UserAnalyticsPresenter
   include PeriodHelper
   include DashboardHelper
   include ActionView::Helpers::NumberHelper
+  include BustCache
 
   def initialize(current_facility)
     @current_facility = current_facility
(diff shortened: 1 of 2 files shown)


app/presenters/user_analytics_presenter.rb
```diff
@@ -146,7 +147,7 @@ def last_updated_at
 
   def statistics
     @statistics ||=
-      Rails.cache.fetch(statistics_cache_key, expires_in: EXPIRE_STATISTICS_CACHE_IN) {
+      Rails.cache.fetch(statistics_cache_key, expires_in: EXPIRE_STATISTICS_CACHE_IN, force: bust_cache?) {
```
prabhanshuguptagit, line 150: Added this to help with debugging

harimohanraj89: We also need to update the user analytics controller to _set_ `RequestStore[:bust_cache]`, right?

harimohanraj89 (approved):
One comment, looks good. Changing DM status constitutes a cache refresh 👍

app/presenters/user_analytics_presenter.rb
prabhanshuguptagit: Hrm right, this does not inherit from `AdminController`. 👍🏼

PR #2872, opened 2021-08-24
Improve facility deletion

prabhanshuguptagit:
**Story card:** [ch3523](https://app.clubhouse.io/simpledotorg/story/3523/improve-facility-deletion-from-dashboard)

## Because

The `Delete facility` button on edit facility page does not have good messaging when it's disallowed.

## This addresses

- [X] Check if any registered users are present in the facility before allowing to delete
- [X] Check for the presence of only scheduled appointments
- [x] Display on the page why deletion is not allowed

### Display why deletion is not allowed

Currently the `Delete facility` button is hidden without any messaging. This disables the button if deletion is not allowed and adds a tooltip that shows the records that are preventing the deletion so an admin can fix them.

image

### DIFF
--- a/app/models/facility.rb
+++ b/app/models/facility.rb
@@ -210,8 +210,24 @@ class Facility < ApplicationRecord
     teleconsultation_medical_officers.map(&:full_teleconsultation_phone_number)
   end
 
+  def records_preventing_discard
+    {
+      "registered patients" => registered_patients,
+      "blood pressures" => blood_pressures,
+      "blood sugars" => blood_sugars,
+      "scheduled appointments" => appointments.status_scheduled,
+      "users" => users
+    }
+  end
+
   def discardable?
-    registered_patients.none? && blood_pressures.none? && blood_sugars.none? && appointments.status_scheduled.none? && users.none?
+    records_preventing_discard.values.all? { |records| records.none? }
+  end
+
+  def discard_prevention_reasons
+    records_preventing_discard.map do |record_name, records|
+      "#{records.count} #{record_name}" if records.any?
+    end.compact
   end
 
   def valid_block
--- a/app/models/facility.rb
+++ b/app/models/facility.rb
@@ -212,11 +212,11 @@ class Facility < ApplicationRecord
 
   def records_preventing_discard
     {
-      "registered patients" => registered_patients,
-      "blood pressures" => blood_pressures,
-      "blood sugars" => blood_sugars,
-      "scheduled appointments" => appointments.status_scheduled,
-      "users" => users
+      "registered patient" => registered_patients,
+      "blood pressure" => blood_pressures,
+      "blood sugar" => blood_sugars,
+      "scheduled appointment" => appointments.status_scheduled,
+      "user" => users
     }
   end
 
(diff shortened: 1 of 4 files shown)


app/views/admin/facilities/edit.html.erb
```diff
@@ -2,9 +2,21 @@
   <h1 class="page-title">Edit facility</h1>
   <nav class="page-nav">
 
-    <% if current_admin.accessible_facilities(:manage).find_by_id(@facility) && @facility.discardable? %>
-      <%= link_to 'Delete facility', [:admin, @facility_group, @facility], method: :delete, data: {confirm: "Are you sure you want to delete #{@facility.name}?"}, class: "btn btn-outline-danger ml-4" %>
+    <% if current_admin.accessible_facilities(:manage).find_by_id(@facility) %>
+    <span class="d-inline-block"
```
prabhanshuguptagit, line 6: The button is hidden if an admin doesn't have permission to delete.

app/models/facility.rb
```diff
@@ -210,8 +210,24 @@ def teleconsultation_phone_numbers_with_isd
...
+  end
+
   def discardable?
-    registered_patients.none? && blood_pressures.none? && blood_sugars.none? && appointments.none?
+    records_preventing_discard.values.all? { |records| records.none? }
+  end
+
+  def discard_prevention_reasons
+    records_preventing_discard.map do |record_name, records|
+      "#{records.count} #{record_name}" if records.any?
```
harimohanraj89, line 229: Let's pluralize this correctly :smile:

app/views/admin/facilities/edit.html.erb
```diff
@@ -2,9 +2,21 @@
...
+          tabindex="0"
+          data-toggle="tooltip"
+          data-placement="bottom"
+          title="<% unless @facility.discardable? %>
+                  This facility cannot be deleted because <%= @facility.discard_prevention_reasons.join(', ') %> are present.
```
harimohanraj89, line 11: Consider `because it contains` or `because it still has`?

app/views/admin/facilities/edit.html.erb
```diff
@@ -2,9 +2,21 @@
...
   <nav class="page-nav">
 
-    <% if current_admin.accessible_facilities(:manage).find_by_id(@facility) && @facility.discardable? %>
-      <%= link_to 'Delete facility', [:admin, @facility_group, @facility], method: :delete, data: {confirm: "Are you sure you want to delete #{@facility.name}?"}, class: "btn btn-outline-danger ml-4" %>
+    <% if current_admin.accessible_facilities(:manage).find_by_id(@facility) %>
+    <span class="d-inline-block"
+          tabindex="0"
+          data-toggle="tooltip"
+          data-placement="bottom"
+          title="<% unless @facility.discardable? %>
```
harimohanraj89, line 10: Does our `facilities_controller` check the facility's `discardable?` state as well in the `delete` action for good measure? It should if it doesn't already.

harimohanraj89 (approved):
Tiny suggestions, not blocking

PR #2874, opened 2021-08-26
Optimise patient summaries view

prabhanshuguptagit:
**Story card:** [ch4735](https://app.clubhouse.io/simpledotorg/story/4735/create-a-mobile-overdue-line-list-download)

## Because

Downloading the overdue list for a facility takes [~10s](https://app.datadoghq.com/apm/traces?query=service%3Asimple_server%20env%3Aproduction%20operation_name%3Arack.request%20%40http.url%3A%2Aappointments.csv%2A&streamTraces=true&start=1628774826264&end=1629984426264&paused=false). We expect a lot more download requests when we make this list available to the app and need to improve performance.

## This addresses

This optimises the `PatientSummary` view that powers the download. It converts all left outer joins for fetching a patient's associated records to lateral joins. Locally with a prod-like data set this improved downloads from 25 seconds to 3 seconds. With production infrastructure it should take 1-2s now.

Explain before: https://explain.depesz.com/s/HOZC
Explain after: https://explain.depesz.com/s/JeVcR

Diff: https://www.diffchecker.com/qIDSxKiN

TL;DR why lateral joins are faster in this case:
- The overdue list is always filtered by a facility. That means we actually need to work with a very small number of patients (1-3000).
- The left joins for finding BPs and appointments join the whole BPs/appointments table and then filter for patients
- It’s cheaper to lateral join and iterate over 1k patients rather than joining 1M BPs and filtering

This slows down the `All facilities` overdue list because it requires all patients. A follow up PR will remove the `All facilities` list and show results for only one district at a time.


harimohanraj89 (approved):
Love it. How much slower is the All Facilities page? Is it like, 20% slower, or 100% slower?

prabhanshuguptagit:
Like 1.5x slower.

harimohanraj89:
1.5 - 2x is still pretty safely within the 30 second timeouts we've got in our chain. Yup, this should be ok to go out independent of the district filtering.

prabhanshuguptagit:
Power users are already at ~1 minute, so it actually makes it ~2 minutes for them 😅 . I'll merge these both together to be safe.

PR #2877, opened 2021-08-31
Reporting pipeline - follow-ups

harimohanraj89:
**Story card:** [ch4710](https://app.clubhouse.io/simpledotorg/story/4710/add-follow-ups-tables-to-reporting-pipeline)

## Because

We want to power the follow-ups on the dashboard with the reporting pipeline as well, to facilitate upcoming iteration to the definition of "follow up"

## This addresses

Adds a `reporting_patient_follow_ups` table with the following features.
* One record per patient per facility per user per month for patient follow-up blood pressures
* No records for any visits that happened in the same month of patient registration
* Some preparation has been done to soon expand "followup BP" to "followup visit"
* This work is heavily inspired by https://metabase.simple.org/question/321, Srihari's work to run a one-time follow-up query

## Test instructions

Incoming


db/migrate/20210826080108_create_reporting_patient_follow_ups.rb
```diff
@@ -0,0 +1,7 @@
+class CreateReportingPatientFollowUps < ActiveRecord::Migration[5.2]
+  def change
+    create_view :reporting_patient_follow_ups, materialized: true
+
+    add_index :reporting_patient_follow_ups, [:patient_id, :user_id, :facility_id, :month_date], unique: true, name: "reporting_patient_follow_ups_unique_index"
```
harimohanraj89, line 5: This unique index covers all four columns of the table in recommended order of most-values to least-values, necessary for concurrent refresh.

db/views/reporting_patient_follow_ups_v01.sql
```diff
@@ -0,0 +1,75 @@
...
+    bp.facility_id,
+    bp.user_id,
+    to_char(bp.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')), 'YYYY-MM') AS month_string
+  FROM patients p
+  INNER JOIN blood_pressures bp
+    ON p.id = bp.patient_id
+    -- removing bps that were recorded the same month as registration
+    AND to_char(bp.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')), 'YYYY-MM')
+      > to_char(p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')), 'YYYY-MM')
+),
```
harimohanraj89, line 14: This is the only subquery that's being used right now, the others are in preparation for upcoming follow up adjustments.

db/views/reporting_patient_follow_ups_v01.sql
```diff
@@ -0,0 +1,75 @@
...
+
+all_follow_ups AS (
+  SELECT *
+  FROM follow_up_blood_pressures
+  -- Follow-ups only look for BPs. Consider including more criteria in v2:
+  -- -------
+  -- UNION (SELECT * FROM follow_up_blood_sugars)
+  -- UNION (SELECT * FROM follow_up_prescription_drugs)
+  -- UNION (SELECT * FROM follow_up_appointments)
+)
```
harimohanraj89, line 66: Here's how we'll add support for other kinds of follow-up activity. The `UNION` ensures that we'll still only count one record per patient per facility per user per month.

db/views/reporting_patient_follow_ups_v01.sql
```diff
@@ -0,0 +1,75 @@
...
+
+SELECT
+  all_follow_ups.patient_id,
+  all_follow_ups.facility_id,
+  all_follow_ups.user_id,
+  cal.*
+FROM all_follow_ups
+LEFT OUTER JOIN reporting_months cal
+    ON all_follow_ups.month_string = cal.month_string
```
harimohanraj89, line 75: Enriching with some utility calendar stuff

spec/models/reports/patient_follow_up_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "rails_helper"
+
+RSpec.describe Reports::PatientFollowUp, {type: :model, reporting_spec: true} do
+  describe "Associations" do
+    it { should belong_to(:patient) }
+    it { should belong_to(:facility) }
+    it { should belong_to(:user) }
+  end
```
harimohanraj89, line 8: TODO: I need to add a lot more specs here to test the matview's behavior

db/migrate/20210826080108_create_reporting_patient_follow_ups.rb
rsanheim: 👍

db/views/reporting_patient_follow_ups_v01.sql
```diff
@@ -0,0 +1,75 @@
...
+    p.id AS patient_id,
+    bp.facility_id,
+    bp.user_id,
+    to_char(bp.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')), 'YYYY-MM') AS month_string
+  FROM patients p
+  INNER JOIN blood_pressures bp
+    ON p.id = bp.patient_id
+    -- removing bps that were recorded the same month as registration
+    AND to_char(bp.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')), 'YYYY-MM')
+      > to_char(p.recorded_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')), 'YYYY-MM')
```
rsanheim, line 13: This is kind of a clunk way to compare year and month. An alternative would be using `date_part('month', recorded_at)` and `date_part('year'...)` which I believe would have the advantage of being able to use indexes.

db/views/reporting_patient_follow_ups_v01.sql
```diff
@@ -0,0 +1,75 @@
...
+  -- -------
+  -- UNION (SELECT * FROM follow_up_blood_sugars)
+  -- UNION (SELECT * FROM follow_up_prescription_drugs)
+  -- UNION (SELECT * FROM follow_up_appointments)
+)
+
+SELECT
+  all_follow_ups.patient_id,
+  all_follow_ups.facility_id,
+  all_follow_ups.user_id,
```
rsanheim, line 71: Should we go ahead and add gender to pave the way for progress tab queries - at least the non-per-day queries?

rsanheim (commented):
Looking good, some initial feedback.

db/views/reporting_patient_follow_ups_v01.sql
harimohanraj89: I'm unsure whether this table should be a bare-minimum fact table that only holds references, or whether it should be a rich denormalized table like patient states. I'd like to hold off on adding gender until we need it, it should be easy to add, given that this thing is in a completely different dependency chain, unrelated to the "states" tables.

rsanheim: 👍

harimohanraj89:
@rsanheim Silly me, my specs didn't get pushed yesterday. I've pushed up more thorough test coverage now, ready for more 👀 thanks!

db/views/reporting_patient_follow_ups_v01.sql
harimohanraj89: `date_part` is awkward because you gotta compare years, and if not you gotta compare months. I did follow the general idea here and use `date_trunc` instead of string comparisons, much cleaner.

spec/models/reports/patient_follow_up_spec.rb
```diff
@@ -0,0 +1,210 @@
...
+      :blood_pressure,
+      patient: patient,
+      user: user,
+      facility: facility,
+      recorded_at: june_2021[:end_of_month]
+    )
+
+    RefreshReportingViews.call
+
+    expect(described_class.count).to eq(1)
```
rsanheim, line 190: Is there anything to assert on here to make sure that the follow up captured is from the `end_of_month` BP - to make sure it grabs the most recent record?

rsanheim (approved):
One spec question, but otherwise LGTM.

spec/models/reports/patient_follow_up_spec.rb
harimohanraj89: There actually isn't anything that ties the record to the end-of-month BP, since we're only storing a month date on each record. However, this assertion is still pretty light. I'll add another assertion that the one record we see in here has the right attributes on it.

PR #2878, opened 2021-08-31
Extract FlipperInfo; send feature flags to DataDog as tags

rsanheim:
**Story card:** [ch4811](https://app.clubhouse.io/simpledotorg/story/4811/get-feature-flag-info-into-datadog)

We need to send current features to datadog inside the scope of a request, as it will help us better see changes in perf between feature toggles.

docs https://docs.datadoghq.com/tracing/guide/add_span_md_and_graph_it/

### DIFF
--- a/app/controllers/flipper_info.rb
+++ b/app/controllers/flipper_info.rb
@@ -11,6 +11,6 @@ module FlipperInfo
   def set_enabled_features_as_datadog_tags
     current_span = Datadog.tracer.active_span
     return if current_span.nil?
-    current_span.set_tag(:enabled_features, current_enabled_features)
+    current_span.set_tag(:enabled_features, current_enabled_features.map(&:to_s))
   end
 end
--- a/app/controllers/flipper_info.rb
+++ b/app/controllers/flipper_info.rb
@@ -1,0 +1,16 @@
+# Expose currently enabled features from Flipper for admin / dashboard requests
+module FlipperInfo
+  def self.included(base)
+    base.helper_method :current_enabled_features
+  end
+
+  def current_enabled_features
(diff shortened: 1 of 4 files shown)


app/controllers/flipper_info.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def current_enabled_features
+    @current_enabled_features ||= Flipper.features.select { |feature| feature.enabled?(current_admin) }.map(&:name)
+  end
+
+  def set_enabled_features_as_datadog_tags
+    current_span = Datadog.tracer.active_span
+    return if current_span.nil?
+    current_span.set_tag(:enabled_features, current_enabled_features)
```
harimohanraj89, line 14: Do we need to set individual tags for each flag, to make it easier to filter in datadog? If not, then ignore me.

harimohanraj89 (approved):
One minor question/suggestion. Feel free to ship after addressing 👍

app/controllers/flipper_info.rb
rsanheim: I dug around into datadog docs and some of their PRs, and its a bit confusing because:

* datadog tags are _sent_ as flat key/value pairs -- you can send other things from ruby-land, but they serialized to strings.
* datadog tags are represented in the UI as something smarter - for example, you'll notice there are tags like `http.method`, which is shown in the DD UI as a `http` hash where you can drill into the key values underneath http.  So you can setup searches or filters on `http.method` for example.

So for now I'm going to merge this and then see how it looks and filters in the Datadog UI, and if we need to build up some sort of smarter tag structure for this sort of thing we can do it.

Related docs: https://github.com/DataDog/dd-trace-rb/pull/1081 and https://docs.datadoghq.com/getting_started/tagging/using_tags?tab=assignment

PR #2879, opened 2021-08-31
Don't try to serialize if patient is missing an address

rsanheim:
**Story card:** [ch4632](https://app.clubhouse.io/simpledotorg/story/4632/undefined-method-attributes-for-nil-nilclass-in-sync-api-when-patient-does-not-have-address)

### DIFF
--- a/app/transformers/api/v3/patient_transformer.rb
+++ b/app/transformers/api/v3/patient_transformer.rb
@@ -58,19 +58,13 @@ class Api::V3::PatientTransformer
     end
 
     def to_nested_response(patient)
-      address_attrs = if patient.address.nil?
-        Statsd.instance.increment("sync.missing_address")
-        {}
-      else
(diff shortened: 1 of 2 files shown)


app/transformers/api/v3/patient_transformer.rb
```diff
@@ -73,5 +73,14 @@ def to_nested_response(patient)
...
         )
     end
+
+    private
+
+    def address_attributes(patient)
+      return Api::V3::Transformer.to_response(patient.address) if patient.address
+
+      Statsd.instance.increment("sync.missing_address")
+      {}
```
prabhanshuguptagit, line 83: We'll need to check if the app can deal with empty address objects. It's possible if we sync an empty address that the app crashes.

We haven't seen any errors on production since we undiscarded the addresses, I think we should probably just fix the data on all envs and introduce the address constraint. wdyt?

PR #2880, opened 2021-09-01
Add extend_current_patient_experiment method

vkrmis:
**Story card:** [ch4838](https://app.clubhouse.io/simpledotorg/story/4838/extend-ihci-aug-2021-experiments)

## Because

We don't want to extend these experiments as stated below, removed the data migration, but keeping the `extend_current_patient_experiment` method.   

IHCI Aug 2021 experiments were initially set up to run from Aug 5 to Sept 4. We need to extend them to run until Sept 19 (15 more days).

## This addresses

* Adds a mechanism in the `Experimentation::Runner` to extend current patient experiments


harimohanraj89 (approved):
Approving this since we tested it _on_ production data and it looked great. We can get retroactive feedback from NA next week.

PR #2887, opened 2021-09-01
Send each enabled feature as a separate key/value pair

rsanheim:
**Story card:** [ch4811](https://app.clubhouse.io/simpledotorg/story/4811/get-feature-flag-info-into-datadog)

Have to send each feature as a separate key / value pair.


harimohanraj89:
This does look pretty 👌

PR #2888, opened 2021-09-01
Bump progress tab cache time to 1 hour

rsanheim:
**Story card:** [ch4831](https://app.clubhouse.io/simpledotorg/story/4831/back-off-progress-tab-expiration-time)

We are currently spending far too much time processing and sending back user progress tab numbers to clients (see below).  Until we are able to properly optimize this end point, we should back off the caching time to 1 hour instead of every 15 minutes.

image


harimohanraj89 (commented):
Was this triggered by any specific event, or just something we've noticed from our metrics?

Additionally, can we run this by Praveen to sanity check?

rsanheim:
> Was this triggered by any specific event, or just something we've noticed from our metrics?

I was reviewing metrics after getting back from vacation, and saw the Datadog reports (above) showing that progress tab is our #1 most resource intensive request.

PR #2889, opened 2021-09-02
Move demo to medium

harimohanraj89:
**Story card:** TBD

## Because

Large datasets in Demo is interfering with, eg. Praveen's ability to effectively demo and screenshare Simple with stakeholders - [Slack ref](https://simpledotorg.slack.com/archives/C013HS4N9U3/p1630494059010900)

## This addresses

Moving Demo to medium, so we can purge and reseed

### DIFF
--- a/lib/seed/config.rb
+++ b/lib/seed/config.rb
@@ -4,13 +4,15 @@ module Seed
     def initialize
       @type = ENV["SEED_TYPE"] ||
         case SimpleServer.env
+        when "test"
+          "test"
         when "android_review"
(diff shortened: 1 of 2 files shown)


rsanheim (approved):
Makes sense.

Note that I think you will want to run `db:reset` on demo to start from a completely fresh clean slate.  The purge task keeps around facilities and facility_groups, which we really don't want we want a smaller data set size on demo.  Either task would require demo users to force a resync on their phone, so the impact should be about the same.

PR #2891, opened 2021-09-02
Show district name (in parens) when searching for facilities; also add to admin user listing

rsanheim:
for easier searching

**Story card:** [ch4865](https://app.clubhouse.io/simpledotorg/story/4865/add-district-to-the-facility-search-drop-down)

image

image

image

### DIFF
--- a/app/models/facility.rb
+++ b/app/models/facility.rb
@@ -106,7 +106,7 @@ class Facility < ApplicationRecord
   delegate :protocol, to: :facility_group, allow_nil: true
   delegate :organization, :organization_id, to: :facility_group, allow_nil: true
   delegate :follow_ups_by_period, to: :patients, prefix: :patient
-  delegate :district_region?, :block_region?, :facility_region?, :region_type, to: :region
+  delegate :district_region, :district_region?, :block_region?, :facility_region?, :region_type, to: :region
   delegate :cache_key, :cache_version, to: :region
 
   def self.parse_facilities_from_file(file_contents)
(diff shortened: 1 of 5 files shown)


timcheadle (commented):
Not sure this is needed, as in production most facilities are in 1-2 facility groups for any given CVHO. It's only confusing for users who can see every facility group, or for our sandbox.

rsanheim:
> Not sure this is needed, as in production most facilities are in 1-2 facility groups for any given CVHO. It's only confusing for users who can see every facility group, or for our sandbox.

This was done based directly via a request from Praveen, as there are enough duplicate facility names to cause confusion (see the linked clubhouse story).

app/models/facility.rb
```diff
@@ -106,7 +106,7 @@ class Facility < ApplicationRecord
   delegate :protocol, to: :facility_group, allow_nil: true
   delegate :organization, :organization_id, to: :facility_group, allow_nil: true
   delegate :follow_ups_by_period, to: :patients, prefix: :patient
-  delegate :district_region?, :block_region?, :facility_region?, :region_type, to: :region
+  delegate :district_region, :district_region?, :block_region?, :facility_region?, :region_type, to: :region
```
harimohanraj89, line 109: Why add a delegation for `district_region`? Is it used in this PR? I see we're still forming the label using facility group.

app/models/facility.rb
```diff
@@ -168,6 +168,10 @@ def child_region_type
     nil
   end
 
+  def label_with_district
+    "#{name} (#{facility_group.name})"
```
harimohanraj89, line 172: Shall we use `district_region` instead? Start relying more on regions and less on FGs?

harimohanraj89 (approved):
Non-blocking feedback, some inconsistency I think between using districts and FGs.

harimohanraj89:
This is also nice now because you can type a district name and see the facilities to choose from 😅

PR #2895, opened 2021-09-07
Return state names in an object

prabhanshuguptagit:
**Story card:** -

## Because

We return state names in the states list endpoint without a name. [Slack ref](https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1631001639130800)

## This addresses

Return state names keyed by `state` in an object.


harimohanraj89 (approved):
Looks great to me.

PR #2876, opened 2021-08-31
Filter overdue results by district

prabhanshuguptagit:
**Story card:** [ch4735](https://app.clubhouse.io/simpledotorg/story/4735/create-a-mobile-overdue-line-list-download)

## Because

Follow up to #2874. It is no longer feasible to show overdue results for `All facilities` on the dashboard.

## This addresses

This filters the overdue results by districts by default and then allows users to further filter by a facility.

https://user-images.githubusercontent.com/16774200/131478313-ef471a0d-0faa-4dbf-9159-0f6f3084083f.mov

PR #2896, opened 2021-09-07
Fix flappy overdue list spec

prabhanshuguptagit:
**Story card:** -

## Because
Made a spec flappy in #2876

## This addresses
District wasn't being selected in the spec. Selects the correct district in the dropdown.


harimohanraj89 (approved):
Ah we gotta select district everywhere. Coolbeans. Let's just run it like 50 times locally to make sure we've patched it.

PR #2903, opened 2021-09-09
Sync API - Eager load teleconsultation_medical_officers

rsanheim:
**Story card:** [ch4947](https://app.clubhouse.io/simpledotorg/story/4947/improve-performance-of-facilitymedicalofficerscontroller-sync-endpoint?vc_group_by=day&ct_workflow=all&cf_workflow=500000005)

## Because

This end point is way slower than it should be

## This addresses

adding eager loading to improve perf

## Test instructions

Ship this PR, and revisit the [datadog reporting](https://app.clubhouse.io/simpledotorg/story/4947/improve-performance-of-facilitymedicalofficerscontroller-sync-endpoint?vc_group_by=day&ct_workflow=all&cf_workflow=500000005) to see how it improves.

Locally I saw this change from 300-500 ms to well under 100 ms in testing.

### DIFF
--- a/app/controllers/api/v4/facility_medical_officers_controller.rb
+++ b/app/controllers/api/v4/facility_medical_officers_controller.rb
@@ -1,6 +1,6 @@
 class Api::V4::FacilityMedicalOfficersController < APIController
   def sync_to_user
-    medical_officers = current_facility_group.facilities.map { |facility|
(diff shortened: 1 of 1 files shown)


kpethtel (approved):
Nice. Thanks for adding performance test numbers. It's nice to know that we're seeing real improvement from this.

PR #2904, opened 2021-09-09
Sync API - Eager load teleconsultation_medical_officers phone numbers also

rsanheim:
**Story card:** [ch4947](https://app.clubhouse.io/simpledotorg/story/4947/improve-performance-of-facilitymedicalofficerscontroller-sync-endpoint?vc_group_by=day&ct_workflow=all&cf_workflow=500000005)

Follow up to https://github.com/simpledotorg/simple-server/pull/2903
## Because

This end point is way slower than it should be

## This addresses

adding eager loading to the medical officers' phone number as well to improve perf

## Test instructions

Ship this PR, and revisit the [datadog reporting](https://app.clubhouse.io/simpledotorg/story/4947/improve-performance-of-facilitymedicalofficerscontroller-sync-endpoint?vc_group_by=day&ct_workflow=all&cf_workflow=500000005) to see how it improves.

I added better test data locally to better see how this end point performs and why we are loading all sorts of PhoneNumberAuthentications, hence the added eager loading here.

### DIFF
--- a/app/controllers/api/v4/facility_medical_officers_controller.rb
+++ b/app/controllers/api/v4/facility_medical_officers_controller.rb
@@ -1,8 +1,8 @@
 class Api::V4::FacilityMedicalOfficersController < APIController
   def sync_to_user
-    medical_officers = current_facility_group.facilities.eager_load(:teleconsultation_medical_officers).map { |facility|
-      facility_medical_officers(facility)
-    }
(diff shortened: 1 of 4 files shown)


kpethtel (approved):
The actual fix looks good, but let's be sure that the random factory doesn't cause any random errors.

PR #2905, opened 2021-09-09
Enable DataDog code profiling in sandbox

rsanheim:
This is in public beta, so they recommend starting in a non prod
enviornment to start

I don't know how useful it will be, but we can get a sense of if it works and what we can see in Sandbox.  See https://docs.datadoghq.com/tracing/profiler/enabling/ruby/ for some docs.

**Story card:** [ch4951](https://app.clubhouse.io/simpledotorg/story/4951/enable-datadog-code-level-profiling?vc_group_by=day&ct_workflow=all&cf_workflow=500000005)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,7 +38,7 @@ gem "flipper-ui"
 gem "flipper"
 gem "friendly_id", "~> 5.2.4"
 gem "github-ds"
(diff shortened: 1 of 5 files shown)


kpethtel (approved):
LGTM

PR #2906, opened 2021-09-09
4921 add diabetes treatment to medical history

kpethtel:
**Story card:** [ch4921](https://app.clubhouse.io/simpledotorg/story/4921/add-receiving-treatment-for-diabetes-to-medical-history)

## Because

We would like to track whether new patients were already taking diabetes medications at the time of registration. 

## Notes

We should consider whether we need to provide function for the new field in the following locations:
app/services/patient_import/spreadsheet_transformer.rb
app/services/patient_deduplication/deduplicator.rb
db/views/reporting_patient_states_v03.sql

## Testing

I tested this locally by registering a new patient and confirming that no errors were raised and that the new patient's medical history contains `receiving_treatment_for_diabetes` of nil.


rsanheim (commented):
The model changes look good. 

I would suggest adding a couple controller specs showing how these new fields can be set or updated.  The current specs are shared examples, which aren't really useful for seeing how these particular fields work for the actual sync process.

kpethtel:
> The model changes look good.
> 
> I would suggest adding a couple controller specs showing how these new fields can be set or updated. The current specs are shared examples, which aren't really useful for seeing how these particular fields work for the actual sync process.

That's a fair point. I'm talking to Sasikanth about what the default value should be. I assumed nil and he initially agreed, but now he's considering that perhaps it should be unknown. I'm going to figure out if we want to migrate all existing users' medical histories to "unknown", then I'll add some better tests.

kpethtel:
> > The model changes look good.
> > I would suggest adding a couple controller specs showing how these new fields can be set or updated. The current specs are shared examples, which aren't really useful for seeing how these particular fields work for the actual sync process.
> 
> That's a fair point. I'm talking to Sasikanth about what the default value should be. I assumed nil and he initially agreed, but now he's considering that perhaps it should be unknown. I'm going to figure out if we want to migrate all existing users' medical histories to "unknown", then I'll add some better tests.

We ended up not setting a default value for the new field. It will default to nil.

spec/controllers/api/v3/medical_histories_controller_spec.rb
```diff
@@ -74,4 +74,121 @@ def create_record_list(n, options = {})
...
+      expect {
+        post :sync_from_user, params: params
+      }.not_to change { patient.reload.medical_history.receiving_treatment_for_diabetes }.from("no")
+      expect(response.status).to eq 200
+    end
+
+    context "without a pre-existing medical history" do
+      before :each do
+        # we don't want a medical history for this and the patient factory creates one by default
+        patient.medical_history.destroy!
```
rsanheim, line 125: We have a trait on the factory that creates patient w/o med history:

```
create(:patient, :without_medical_history)
```

spec/controllers/api/v3/medical_histories_controller_spec.rb
```diff
@@ -74,4 +74,121 @@ def create_record_list(n, options = {})
...
+        expect(history.prior_stroke).to eq(expected_values[:prior_stroke])
+        expect(history.chronic_kidney_disease).to eq(expected_values[:chronic_kidney_disease])
+        expect(history.receiving_treatment_for_hypertension).to eq(expected_values[:receiving_treatment_for_hypertension])
+        expect(history.diabetes).to eq(expected_values[:diabetes])
+        expect(history.diagnosed_with_hypertension).to eq(expected_values[:diagnosed_with_hypertension])
+        expect(history.hypertension).to eq(expected_values[:hypertension])
+        expect(history.receiving_treatment_for_diabetes).to eq(expected_values[:receiving_treatment_for_diabetes])
+      end
+
+      it "sets values to nil when not provided" do
```
rsanheim, line 154: This would maybe be better titled `leaves values at nil` I think? The sync API just doesn't pass along those attributes, so they never get set.

spec/controllers/api/v3/medical_histories_controller_spec.rb
```diff
@@ -74,4 +74,121 @@ def create_record_list(n, options = {})
...
+        expect(history.prior_stroke).to eq(expected_values[:prior_stroke])
+        expect(history.chronic_kidney_disease).to eq(expected_values[:chronic_kidney_disease])
+        expect(history.receiving_treatment_for_hypertension).to eq(expected_values[:receiving_treatment_for_hypertension])
+        expect(history.diabetes).to eq(expected_values[:diabetes])
+        expect(history.diagnosed_with_hypertension).to eq(expected_values[:diagnosed_with_hypertension])
+        expect(history.hypertension).to eq(expected_values[:hypertension])
+        expect(history.receiving_treatment_for_diabetes).to eq(nil)
+      end
+
+      it "returns 200 but returns errors and does not create a medical history when provided invalid values" do
```
rsanheim, line 181: 👍 I agree this is surprising behavior, and probably something we want to change at some point, but good to have it documented in an explicit spec like this.

rsanheim (approved):
Approving to unblock w/ some spec feedback.

PR #2907, opened 2021-09-09
Add User#can_access? for direct Facility / FacilityGroup / Organization checks

rsanheim:
Adding the ability to see if a user has access to a single facility group or facility. This is to address a significant perf issue I'm seeing in our region reports (see linked story for details), and will be generally useful in addition to our current `accessible_xxx` methods (which are better for returning the _entire_ set of accessible resources).

**Story card:** [ch4960](https://app.clubhouse.io/simpledotorg/story/4960/fix-performance-issue-in-authz-check-for-regions-show)

### Perf bottleneck (before)

image

https://app.datadoghq.com/apm/traces?query=env%3Aproduction%20service%3Asimple_server%20operation_name%3Arack.request%20resource_name%3A%22Reports%3A%3ARegionsController%23show%22&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&historicalData=true&messageDisplay=inline&sort=desc&spanID=2146490566191857791&streamTraces=true&timeHint=1631163600000&trace=AQAAAXvI9z0YCApRbQAAAABBWHZJOTBqLUFBQ2tJTXc2a2J3RG5DRTk&traceID=4404863127396276531&start=1631125942566&end=1631298742566&paused=false

### Perf bottleneck (after)

...TODO...


app/models/facility.rb
```diff
@@ -147,6 +147,10 @@ def source
     self
   end
 
+  def accessible_by(user, action: nil)
```
kpethtel, line 150: A nil value for action will raise an error in `action_to_level`. Should we make it a required param instead?

spec/models/user_spec.rb
```diff
@@ -7,6 +7,23 @@
     it { is_expected.to have_and_belong_to_many(:teleconsultation_facilities) }
   end
 
+  describe "can_access?" do
+    let(:viewer_all) { UserAccess.new(create(:admin, :viewer_all)) }
+    let(:manager) { UserAccess.new(create(:admin, :manager)) }
+    let(:power_user) { UserAccess.new(create(:admin, :power_user)) }
+
+    it "works" do
```
kpethtel, line 15: Per the comment above, it might worth testing an invalid action value.

kpethtel (approved):
Changing this to approved. My one comment isn't really blocking and the rest seems fine.

app/models/user_access.rb
```diff
@@ -201,6 +211,17 @@ def grant_access(new_user, selected_facility_ids)
...
+  def has_access?(resource)
+    case resource
+    when Facility
+      user.accesses.exists?(resource: resource) || has_access?(resource.facility_group)
+    when FacilityGroup
+      user.accesses.exists?(resource: resource) || has_access?(resource.organization)
+    when Organization
+      user.accesses.exists?(resource: resource)
+    end
+  end
```
rsanheim, line 223: check for direct access, and then works up the chain to see if a parent FG or Org grants access

app/views/reports/regions/_downloads.html.erb
```diff
@@ -27,7 +27,7 @@
     <% end %>
 
     <% if @region.region_type.in?(["district", "facility"]) %>
-      <% if @region.facilities.to_set.subset?(current_admin.accessible_facilities(:view_pii).to_set) %>
+      <% if current_admin.can_access?(@region.source, :view_pii) %>
```
rsanheim, line 30: This is the actual authz check in question from the Datadog graphs.

spec/models/user_access_spec.rb
```diff
@@ -111,11 +143,13 @@
           it "returns the facility groups an admin can perform actions on with organization access" do
             permission_matrix.each do |admin, action, current_resource, expected_resources|
               admin.accesses.create!(resource: current_resource)
+
               expect(admin.accessible_facility_groups(action)).to match_array(expected_resources),
                 error_message(admin,
                   action,
                   expected_resources,
                   admin.accessible_facility_groups(action))
+              expected_resources.each { |resource| expect(admin.can_access?(current_resource, action)).to be_truthy }
```
rsanheim, line 152: I'm buildingin onto the specs we have, because they already create a whole matrix of data that is very useful to assert on (and I don't want to create any more objects than we already are here for testing...).

rsanheim (commented):
notes

PR #2909, opened 2021-09-10
Add custom Datadog spans for Sync API (sync_to_user)

rsanheim:
Add custom spans to our sync API so we can see these parts of the sync process on their own in Datadog.

See https://docs.datadoghq.com/tracing/setup_overview/custom_instrumentation/ruby/?tab=activespan#adding-spans for an overview on how this works.

Going to only add to `sync_to_user` to verify this works before adding elsewhere.
**Story card:** [ch4964](https://app.clubhouse.io/simpledotorg/story/4964/add-spans-around-key-parts-of-sync-api)

image

image

### DIFF
--- a/app/controllers/api/manifests_controller.rb
+++ b/app/controllers/api/manifests_controller.rb
@@ -1,6 +1,6 @@
 class Api::ManifestsController < ApplicationController
   def show
-    if ENV["SIMPLE_SERVER_ENV"].in?(%w[development review])
+    if ENV["SIMPLE_SERVER_ENV"].in?(%w[development profiling review])
       @countries = %w[IN BD ET US UK]
     else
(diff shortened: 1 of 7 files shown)


kpethtel (approved):
This is a nice addition. It looks like we also have records_to_sync in the facilities and medications controllers. You can probably get rid of the `time(_mehod_)` in those locations as well.

harimohanraj89:
Could we throw this on SBX and get some screenshots from Datadog of what this will look like in action? It would be a nice thing to capture in the PR description.

rsanheim:
> Could we throw this on SBX and get some screenshots from Datadog of what this will look like in action? It would be a nice thing to capture in the PR description.

yup, I added some example screenshots I took while running Datadog locally with this doing some syncs.

PR #1214, opened 2020-08-10
Assigned facility: Overdue list

prabhanshuguptagit:
**Story card:** https://app.clubhouse.io/simpledotorg/story/801

## Because

We are making our dashboard reports assigned facility compliant.

## This addresses

This changes the definition of overdue patients at a facility in the overdue list ([PRD](https://docs.google.com/document/d/1YZqm0t5NhTDUkn3Ww8yRqAszKCtAiOqChpFED6U-QXE/edit#heading=h.7cwjocnpi9wr)). This necessitates adding the `assigned_facility_id` to `PatientSummary`.

Before:
All overdue patients whose next appointment is at the facility f AND

After:
All overdue patients whose assigned facility is f AND
Next appointment is at any facility AND

PR #2911, opened 2021-09-13
Add API for overdue list download

prabhanshuguptagit:
**Story card:** [ch4735](https://app.shortcut.com/simpledotorg/story/4735/create-a-mobile-overdue-line-list-download)

## Because

The android app should be able to download their facility's overdue list.

## This addresses

Adds a `/api/v4/analytics/overdue_list` endpoint that downloads the app user's current facility's overdue list. Authentication works the same way as the progress tab, using headers.

This also removes filtering the overdue list based on `next_appointment_facility_id`, which was supposed to be removed as part of the [assigned facility migration](https://docs.google.com/document/d/1YZqm0t5NhTDUkn3Ww8yRqAszKCtAiOqChpFED6U-QXE/edit#heading=h.7cwjocnpi9wr) in https://github.com/simpledotorg/simple-server/pull/1214. Currently patients who are assigned to an admin's jurisdiction but have an appointment outside their jurisdiction are not included. We should display these patients because:
- this ensure the app users and the dashboard users see the same list
- there's a chance these patients are falling into a hole where none of the admins can see them

All assigned patients will be included in the list after this change.


spec/controllers/api/v4/analytics/overdue_lists_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe Api::V4::Analytics::OverdueListsController, type: :controller do
+  let(:request_user) { create(:user) }
+  let(:request_facility) { create(:facility, facility_group: request_user.facility.facility_group) }
+
+  describe "#show" do
+    before :each do
+      request.env["HTTP_X_USER_ID"] = request_user.id
```
kpethtel, line 9: Given how many times we do this in our specs, would it be worth extracting this setup into a helper? We seem to have a lot of code like this:
```
        let(:request_user) { FactoryBot.create(:user) }
        let(:request_facility) { FactoryBot.create(:facility, facility_group: request_user.facility.facility_group) }
        let(:HTTP_X_USER_ID) { request_user.id }
        let(:HTTP_X_FACILITY_ID) { request_facility.id }
        let(:Authorization) { "Bearer #{request_user.access_token}" }
        let(:Accept) { "text/html" }
```

I'm not suggesting you do this as part of this PR. It's just a thought.

kpethtel (approved):
Just a couple thoughts, but it LGTM. 👍

app/controllers/api/v3/analytics_controller.rb
```diff
@@ -1,3 +1,2 @@
-class Api::V3::AnalyticsController < APIController
-  around_action :set_reporting_time_zone
+class Api::V3::AnalyticsController < Api::V4::AnalyticsController
```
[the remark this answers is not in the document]
prabhanshuguptagit: This was an oversight, thanks for catching 😅 . It should be the other way around. Fixed in 3ccfa4e.

spec/controllers/api/v4/analytics/overdue_lists_controller_spec.rb
prabhanshuguptagit: That sounds like a good idea, can defo refactor in a future PR.

prabhanshuguptagit (commented):
@kpethtel I've made some more changes since last review, requesting a re-review. There's a couple of columns that are different than the dashboard overdue list that I overlooked earlier. I've added those columns now. Also created a new ERB template since this list is sufficiently different from the dashboard list. The columns are:
- Latest BP Passport
- Latest medicines

I briefly tried doing all of this in SQL in https://github.com/simpledotorg/simple-server/pull/2911/commits/7a058b1e61318f029db87799460d6d623218281b but it's performant enough locally in its current form with an `includes`. Will run this on prod and see what the performance is like there.

rsanheim (changes requested):
Following up in slack about some concerns with this PR.

spec/controllers/api/v4/analytics/overdue_lists_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        scheduled_date: 1.month.ago,
+        facility: request_facility)
+      patient_ids = overdue_appointments.map(&:patient_id) << unvisited_patient.id
+
+      get :show, params: {format: :csv}
+
+      expect(assigns(:patient_summaries).map(&:id)).to match_array(patient_ids)
+    end
+
+    it "returns a 401 if user is not authenticated" do
```
rsanheim, line 32: Can you add some specs showing that a user can't access data from a facility they don't have access to?   For example, what happens if they change the `HTTP_X_FACILITY_ID` to a different facility that they shouldn't have PII access to?

rsanheim (changes requested):

kpethtel (approved):
Rob's feedback aside, still LGTM.

PR #2912, opened 2021-09-13
Proposed BD region localization changes

rsanheim:
**Story card:** [ch4649](https://app.shortcut.com/simpledotorg/story/4649/change-bangladesh-region-structure)

This should only be shipped after https://github.com/simpledotorg/simple-server/pull/2918 has been shipped and run and confirmed in BD production.


dburka (approved):
Changes look correct

PR #2916, opened 2021-09-14
Refactor DHIS2 rake task to its own exporter

timcheadle:
**Story card:** [ch4708](https://app.shortcut.com/simpledotorg/story/4708/refactor-dhis2-code-and-add-tests)

## Because

Rake tasks should be single commands. This moves the complexity of the DHIS2 exporter into it's own class.

## This addresses

This is just a basic refactor into `app/exporters/`, nothing else.

## Test instructions

Comment out the `Dhis2.client` export call (line 64) and replace with `pp facility_bulk_data`. Then run `rake dhis2:export`


rsanheim (approved):
LGTM

rsanheim:
looks like an intermittent failure in CI

PR #2917, opened 2021-09-14
Ethiopia DHIS2 calendar conversion

timcheadle:
**Story card:** [ch4339](https://app.shortcut.com/simpledotorg/story/4339/add-ethiopian-calendar-conversion-library)

## Because

Ethiopia's DHIS2 servers use the Ethiopian calendar for monthly periods.

## This addresses

This standardizes the DHIS2 period format strings and converts to Ethiopian periods if needed.

### DIFF
--- a/app/models/period.rb
+++ b/app/models/period.rb
@@ -185,6 +185,14 @@ class Period
     end
   end
 
+  def to_dhis2
+    if quarter?
+      "#{value.year}Q#{value.number}"
+    else
+      value.to_s(:dhis2_mon_year)
+    end
+  end
+
   # Returns a Hash with various Period related dates for eash consumption by the view
   def to_hash
     {
--- a/app/services/ethiopia_calendar_utilities.rb
+++ b/app/services/ethiopia_calendar_utilities.rb
@@ -6,4 +6,8 @@ module EthiopiaCalendarUtilities
 
     Date.new(year, 1, 1) + days_offset.days
   end
+
+  def self.gregorian_month_period_to_ethiopian(period)
+    period.advance(years: -7, months: -8)
+  end
 end
--- a/spec/services/ethiopia_calendar_utilities_spec.rb
+++ b/spec/services/ethiopia_calendar_utilities_spec.rb
@@ -23,4 +23,31 @@ RSpec.describe EthiopiaCalendarUtilities do
       end
     end
   end
+
+  describe ".gregorian_month_period_to_ethiopian" do
+    it "converts Gregorian monthly periods to Ethiopian monthly periods" do
+      conversions = {
+        [2021, 1, 1] => "2013-05",
+        [2021, 2, 1] => "2013-06",
+        [2021, 3, 1] => "2013-07",
+        [2021, 4, 1] => "2013-08",
+        [2021, 5, 1] => "2013-09",
+        [2021, 6, 1] => "2013-10",
+        [2021, 7, 1] => "2013-11",
+        [2021, 8, 1] => "2013-12",
+        [2021, 9, 1] => "2014-01",
+        [2021, 10, 1] => "2014-02",
+        [2021, 11, 1] => "2014-03",
+        [2021, 12, 1] => "2014-04",
+        [2022, 1, 1] => "2014-05"
+      }
+
+      conversions.each do |gregorian_month, ethiopian_month|
+        period = Period.month(Date.new(*gregorian_month))
+        converted_period = described_class.gregorian_month_period_to_ethiopian(period)
+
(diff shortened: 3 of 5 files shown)


spec/services/ethiopia_calendar_utilities_spec.rb
```diff
@@ -23,4 +23,31 @@
       end
     end
   end
+
+  describe ".gregorian_month_period_to_ethiopian" do
+    it "converts Gregorian monthly periods to Ethiopian monthly periods" do
+      conversions = {
```
kpethtel, line 29: Clever. I don't often see arrays as keys.

app/models/period.rb
```diff
@@ -185,6 +185,16 @@ def to_s(format = :mon_year)
     end
   end
 
+  def to_dhis2
+    # DHIS2 has its own period string formats
+    # https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/introduction.html#webapi_date_perid_format
+    if quarter?
+      "#{value.year}Q#{value.number}"
```
kpethtel, line 192: I think we could add this to time and date formats as well. I think it would just be `"%YQ%m"`.

app/services/ethiopia_calendar_utilities.rb
```diff
@@ -6,4 +6,8 @@ def self.ethiopian_to_gregorian(e_year, e_month, e_days)
 
     Date.new(year, 1, 1) + days_offset.days
   end
+
+  def self.gregorian_month_period_to_ethiopian(period)
+    period.advance(years: -7, months: -8)
```
kpethtel, line 11: Is this correct? Shouldn't it be (more or less) the inverse of `ethiopian_to_gregorian`? Is it ok that this is just more of a ballpark estimate?

app/exporters/dhis2_exporter.rb
```diff
@@ -33,8 +33,6 @@ def export
       repository = Reports::Repository.new(facility.region, periods: range)
 
       range.each do |period|
```
kpethtel, line 35: I think this code would be slightly clearer if we renamed `period` to `month` here and in `reporting_period` below. If I understand the code, it would break if we used quarters.

spec/services/ethiopia_calendar_utilities_spec.rb
```diff
@@ -23,4 +23,31 @@
...
+        [2021, 1, 1] => "2013-05",
```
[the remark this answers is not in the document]
rsanheim: Which then calls into question the entire conversion happening this late 😁 ...the numbers we are exporting were calculated by doing totals and groupings based on the Gregorian calendar. So we will be grouping counts into the wrong periods by doing the calendar conversion so late in the process.  Is the ET team okay w/ having mismatched counts for the ET calendar exported data?

As one example, I think we would be including data from Sept 1 to Sept 10th into the `2014-01` ET period with this approach, even though 2014-01 only started Sept 11th in Gregorian.

rsanheim (changes requested):
Have some concerns here...also, dates are hard. 😭 😏

app/services/ethiopia_calendar_utilities.rb
harimohanraj89: I believe the ethiopia ministry of health shoehorns ethiopia dates into their DHIS2 instance using this ballpark method, hence it's a lot simpler than a real conversion.

Should we qualify this method name to indicate that more clearly? Perhaps with some sort of `dhis2` prefix or suffix?

timcheadle:
@rsanheim I haven't heard back from Legesse or Siraj at the MoH, but in retrospect Legesse was the one who suggested this approach in the first place. Since this isn't even used in production yet, let's assume this is the right approach

rsanheim:
> @rsanheim I haven't heard back from Legesse or Siraj at the MoH, but in retrospect Legesse was the one who suggested this approach in the first place. Since this isn't even used in production yet, let's assume this is the right approach

Cool, that makes sense 👍 .

Mind incorporating the `to_s` feedback above in https://github.com/simpledotorg/simple-server/pull/2917#discussion_r708716855 ?

app/models/period.rb
```diff
@@ -177,11 +177,17 @@ def inspect
     "<Period type:#{type} value=#{value}>"
   end
 
-  def to_s(format = :mon_year)
+  def to_s(format = :default_period)
+    value.to_s(format)
```
rsanheim, line 181: Ended up cleaning this up w/ an idea of `default_period`.

rsanheim (approved):
@timcheadle I cleaned up the date/quarter formatting a bit as I saw a way to get rid of the weird conditionals.  I think this is an improvement.

Approving to unblock ya.

PR #2918, opened 2021-09-15
DataScript to update Bangladesh regions

rsanheim:
**Story card:** [ch4649](https://app.shortcut.com/simpledotorg/story/4649/change-bangladesh-region-structure)

## Context 

We want to change the BD region structure so that the lowest level above Facility is an Upazilla instead of a Union. So basically everything in the current structure will move one level down in the hierarchy. 

Below is the outline of the current region structure and the proposed region structure we want to move to in BD:

## DataScript / Generator

I'm adding a generator (`rails g data_script foo`) that we can use to change Prod data going forward.  Right now it creates a `data_script` object and a corresponding spec, and sets up some basic scaffolding for things like running safely via `dry_run`, as well as some default logging and results gathering.  We can extract more to this as we use it over time. 

This is the type of thing we will are doing more of, and standardizing how we do it and ensuring its super easy to have tests and visibility will be great to keep our data changes robust.

## Region Structure Changes

```
| canonical Name   | BD **current** | BD **proposed**
|—————————+———————+———————+———————--|--------------|
| Organization     | Organization   | Organization |
| State            | District       | Division     |
| District         | Upazila        | District     | 
| Block (aka Zone) | Union          | Upazila      | 
| Facility         | Facility       | Facility     | 
```

## Acceptance Criteria

* All Facilities need to be reparented so that their parent regions are Upazilla, District, and Division
* Create new FacilityGroups for the new districts. Associate facilities to the new facility groups.
* All Union regions need to be deleted
* Make localization changes so that new/edit screens for Facility and FacilityGroup display correct labels
* Update user access to point at the new districts instead of Upazilla
* Schedule and coordinate change w/ BD users
* Make sure all BD client apps are fully resynced and functioning properly

### Forcing client resyncs:

Clients will automatically resync after our data migration, assuming we:

* change all facilities to have a new block_region
* ensure that all facilities have their `updated_at` changed (via `touch` etc)
* see https://github.com/simpledotorg/simple-server/blob/d93f3e0eea779ae4ad129f1a2e1313b56f352a5a/app/controllers/concerns/api/v3/sync_to_user.rb#L93 and https://github.com/simpledotorg/simple-server/blob/d93f3e0eea779ae4ad129f1a2e1313b56f352a5a/app/controllers/api/v3/facilities_controller.rb#L19

### Screens that need updates

The below screen will become the *new district* page, instead of *new upazilla* -- this all comes from 
[this localization file](https://github.com/simpledotorg/simple-server/blob/master/config/locales/en_BD.yml)

FacilitiyGroup#edit / new

Facilities#edit / new


lib/generators/data_script/data_script_generator.rb
```diff
@@ -0,0 +1,9 @@
+class DataScriptGenerator < Rails::Generators::NamedBase
```
rsanheim, line 1: Adding a generator to make it easy to create a data script and a corresponding spec - this is common enough where we should automate it and keep it safe by default.

lib/generators/data_script/templates/data_script.erb
```diff
@@ -0,0 +1,13 @@
+class <%= class_name %>Script < DataScript
```
rsanheim, line 1: I'll extract more to over time...

spec/exporters/drug_stocks_report_exporter_spec.rb
```diff
@@ -2,193 +2,193 @@
 
 RSpec.describe DrugStocksReportExporter do
   around do |example|
-    I18n.with_locale(:en_IN) do
-      example.run
+    Timecop.freeze do
```
rsanheim, line 5: This was an unrelated intermittent test fix.

lib/data_script.rb
```diff
@@ -0,0 +1,9 @@
+class DataScript
```
rsanheim, line 1: This will be a base class for these sorts of data scripts going forward.

The idea here is to 
* default to a dry run mode to see what would be changed before running things
* provide a standard scaffold for things like logging and capturing results

rsanheim (commented):
Some notes for reviewiers

lib/generators/data_script/templates/data_script_spec.erb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+
+describe <%= class_name %>Script do
```
rsanheim, line 3: Specs by default for anything that touches production data.

spec/fixtures/files/bd_test_regions.csv
```diff
@@ -0,0 +1,55 @@
+Division,District,Upazila,Union,CC Name,Facility/Organization Code,Facility Name,Facility Type,,,,,,,,,,,,,,
```
rsanheim, line 1: A smaller CSV for testing purposes.

lib/data_scripts/update_bangladesh_regions_script.rb
```diff
@@ -0,0 +1,114 @@
...
+    fields = {module: :data_script, class: self.class.to_s}
+    @logger = Rails.logger.child(fields)
+    @results = {created: Hash.new(0), deleted: Hash.new(0), errors: Hash.new(0), dry_run: dry_run?}
+    @csv_path = csv_path
+    @cache = {state: {}, district: {}, block: {}, facility: {}}
+  end
+
+  def call
+    return unless CountryConfig.current_country?("Bangladesh")
+    destroy_empty_facilities
```
rsanheim, line 23: We want to clear out existing community facilities that were added, because their regions aren't correct and its easier to add them back from the CSV to ensure correctness.

lib/data_scripts/update_bangladesh_regions_script.rb
```diff
@@ -0,0 +1,114 @@
...
+    facilities = Facility.where(facility_size: ["community", nil]).where(sql)
+    facilities.each do |facility|
+      if run_safely { facility.destroy }
+        results[:deleted][:facilities] += 1
+      end
+    end
+  end
+
+  def run_safely
+    return true if dry_run?
```
rsanheim, line 97: Allow us to wrap persistence calls and just return true when in dry_run, to avoid accidentally updating data during a dry run.

lib/data_scripts/update_bangladesh_regions_script.rb
```diff
@@ -0,0 +1,114 @@
...
+        when "CC", "USC" then "community"
+        when "UHC" then "large"
+        else raise ArgumentError, "unknown facility_type #{row[:facility_type]}"
+      end
+
+      division_region = find_or_create_region(:state, division_name, org_region)
+      district_region = find_or_create_region(:district, district_name, division_region)
+      facility_group = FacilityGroup.find_by(name: district_name) || FacilityGroup.new(name: district_name, organization: org_region.source, region: district_region, protocol: protocol, generating_seed_data: true)
+      if facility_group.new_record?
+        if run_safely { facility_group.save! }
```
kpethtel, line 48: Do we want this bang here?

spec/exporters/drug_stocks_report_exporter_spec.rb
```diff
@@ -2,193 +2,193 @@
 
 RSpec.describe DrugStocksReportExporter do
   around do |example|
-    I18n.with_locale(:en_IN) do
```
kpethtel, line 5: I propose we pull this out into a separate PR since this one is pretty complex and this seems unrelated.

spec/lib/data_scripts/update_bangladesh_regions_script_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        .and change { Region.district_regions.count }.by(6)
+      Facility.all.each do |facility|
+        fg = facility.facility_group
+        expect(fg).to_not be_nil
+        expect(fg.name).to eq(fg.region.name)
+      end
+    end
+  end
+
+  it "removes Facilities without patients and users" do
```
kpethtel, line 61: I don't think this description is quite correct. Looking at the code above, it seems like it only removes facilities without patients. It might be more accurate as "removes community and unsized facilities without patients".

lib/generators/data_script/templates/data_script_spec.erb
```diff
@@ -0,0 +1,15 @@
...
+  it "runs" do
+    described_class.call
+  end
+
+  it "changes nothing in dry run mode" do
```
kpethtel, line 8: This part is what made the generator seem more valuable to me. It's an interesting way of enforcing dry_run functionality.

kpethtel (approved):
There's a lot of interesting stuff going on here. Very nice work.

lib/data_scripts/update_bangladesh_regions_script.rb
```diff
@@ -0,0 +1,114 @@
...
+    return unless CountryConfig.current_country?("Bangladesh")
+    destroy_empty_facilities
+    create_facilities
+  end
+
+  def create_facilities
+    org_region = Region.organization_regions.find_by!(slug: "nhf")
+    protocol = Protocol.find_by!(name: "Bangladesh Hypertension Management Protocol for Primary Healthcare Setting")
+    each_row do |row|
+      next if row[:division].blank? || row[:division] == "Division" || row[:facility_name].blank?
```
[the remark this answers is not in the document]
rsanheim: There are known blank lines in the CSV export due to how the multiple sheets get exported and combined, so I'm not too worried about these.

lib/data_scripts/update_bangladesh_regions_script.rb
```diff
@@ -0,0 +1,114 @@
...
+      logger.debug { "Processing #{row[:facility_name]}" }
+
+      facility_name = row[:facility_name]
+      division_name = row[:division]
+      district_name = row[:district]
+      upazila_name = row[:upazila]
+      facility_size = case row[:facility_type]
+        when "CC", "USC" then "community"
+        when "UHC" then "large"
+        else raise ArgumentError, "unknown facility_type #{row[:facility_type]}"
```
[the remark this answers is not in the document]
rsanheim: good point, I'll change to just capturing this result and moving on.

lib/data_scripts/update_bangladesh_regions_script.rb
rsanheim: Good catch, we don't.  removing

lib/data_scripts/update_bangladesh_regions_script.rb
```diff
@@ -0,0 +1,114 @@
...
+    end
+    region
+  end
+
+  def destroy_empty_facilities
+    sql = <<-SQL
+      NOT EXISTS (SELECT 1 FROM patients where patients.registration_facility_id = facilities.id) AND
+      NOT EXISTS (SELECT 1 FROM patients where patients.assigned_facility_id = facilities.id)
+    SQL
+    facilities = Facility.where(facility_size: ["community", nil]).where(sql)
```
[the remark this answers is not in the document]
rsanheim: Yeah, all the incorrect (already added) versions of the facilities are either community or don't have a size set yet.

spec/exporters/drug_stocks_report_exporter_spec.rb
rsanheim: Oh, if you view with whitespace ignored you'll see this is actually a very tiny change.  It just happened to change indentation throughtout the file.

lib/data_script.rb
vkrmis: I like the idea 👍 

Think this is worth documenting under `/doc/arch`, maybe as a follow up?
Our current data migration strategy is outlined at [007-robust-data-migrations.md](https://github.com/simpledotorg/simple-server/blob/master/doc/arch/007-robust-data-migrations.md )

lib/data_scripts/update_bangladesh_regions_script.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    fields = {module: :data_script, class: self.class.to_s}
+    @logger = Rails.logger.child(fields)
+    @results = {created: Hash.new(0), deleted: Hash.new(0), errors: Hash.new(0), dry_run: dry_run?}
+    @csv_path = csv_path
+    @cache = {state: {}, district: {}, block: {}, facility: {}}
+  end
+
+  def call
+    return unless CountryConfig.current_country?("Bangladesh")
```
vkrmis, line 22: Do we want to limit this to Bangladesh Production?

lib/data_scripts/update_bangladesh_regions_script.rb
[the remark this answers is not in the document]
vkrmis: Have actual users (eg: nurses) been registered at these facilities? 
We may want to reassign them to their new facilities or delete their accounts along with the facilities and ask them to sign up again.

spec/lib/data_scripts/update_bangladesh_regions_script_spec.rb
rsanheim: 👍

lib/data_scripts/update_bangladesh_regions_script.rb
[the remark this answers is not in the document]
rsanheim: There have been a few users who have signed up, but not too many.  I'll add some code here to delete their accounts, and we will communicate w/ BD that all users will have to sign up fresh.

lib/data_scripts/update_bangladesh_regions_script.rb
prabhanshuguptagit: Neat!

lib/data_scripts/update_bangladesh_regions_script.rb
```diff
@@ -0,0 +1,144 @@
...
+        results[:deleted][:users] += 1
+      end
+    end
+    logger.info { "Removing facility groups with no facilities" }
+    related_facility_groups.each do |facility_group|
+      if facility_group.facilities.size == 0
+        logger.info { "Removing facility group #{facility_group.name} " }
+        run_safely { facility_group.destroy }
+      end
+    end
```
prabhanshuguptagit, line 123: Adding a log here that shows `0` remaining community clinics might be helpful, just in case someone added users/patients to these facilities when we run it.

prabhanshuguptagit (commented):
The migration looks good so far, I'll do another pass when the existing facilities migration has been added.

rsanheim:
> The migration looks good so far, I'll do another pass when the existing facilities migration has been added.

@prabhanshuguptagit I'm going with a clean slate approach here as it will be easier to get things in line that way. 

So we wipe all the incorrectly added CC clinics, then reimport fresh from a full CSV of all the new CSVs.  So less data being changed in place and easier to verify things are added correctly.

rsanheim:
Some verification with data locally: 
```
{:created=>{:facility_regions=>1449, :facilities=>1449, :block_regions=>8},
 :deleted=>{:facilities=>1317, :users=>12},
 :errors=>{},
```

rsanheim:
Last remaining item for this PR: ensure that empty upazilas are deleted after facilities are removed.

rsanheim:
I need to add a couple small pieces that came out of talking w/ Daniel, and then the export. But I'm going to get this merged so future changes can be smaller PRs, this one is already larger than I'd like.

PR #2925, opened 2021-09-17
Update webpacker and webpack-dev-server

rsanheim:
Trying to resolve some of the JS vulnerabilities we have.

**Story card:** [ch4557](https://app.shortcut.com/simpledotorg/story/4557/security-vulnerabilities-in-various-js-libraries)


timcheadle:
Sorry I don't understand what's actually happening here. Webpacker needs upgrading, but there's a blank `index.html` and some new CSS? Can you add some flavor here @rsanheim ?

rsanheim:
> Sorry I don't understand what's actually happening here. Webpacker needs upgrading, but there's a blank `index.html` and some new CSS? Can you add some flavor here @rsanheim ?

@timcheadle Sure - right now webpack is _only_ used in our shim v2 home page which @claudiovallejo will implement the dashboard refresh in.  That home page is only available behind a feature flag, none of it is in prod yet.

Since I'm upgrading webpack and things involved in the CSS /JS bundling there, I wanted to have a bit more SCSS involved there to really make sure nothing breaks ...such as with imports, SCSS variables, etc.

So I added this small, nifty classless [Sakura](https://github.com/oxalorg/sakura) CSS file to have something to import and to have some default styles -- all of that can go away once @claudiovallejo starts to work on the real refresh in earnest.

app/javascript/packs/application.scss
```diff
@@ -1,21 +1,12 @@
...
+will be included in the "application" pack. Any CSS imported from application.js or as part of the
+application.js dependency graph, e.g. `import '../stylesheets/my-css.css'` will also be included
+in the "application" pack.
+
+To reference this file, add <%= stylesheet_pack_tag 'application' %> to the appropriate
+layout file, like app/views/layouts/application.html.erb
+*/
+
+@import 'bootstrap/scss/bootstrap';
+@import 'sakura'
```
timcheadle, line 12: Should this get imported _before_ bootstrap so we don't accidentally override any styles?

timcheadle (approved):
Approved, with one comment to consider

app/javascript/packs/application.scss
rsanheim: its fine to override bootstrap in the webpack bundle, because its only getting used for this one new page.  I'll remove the bootstrap import and comment on this here fact, just to make things more clear.

PR #2926, opened 2021-09-17
4062 imo notification callback 2

kpethtel:
**Story card:** [ch4062](https://app.shortcut.com/simpledotorg/story/4062/process-imo-notification-callback)

## Because

We want to know if our imo notifications were read by patients. This new endpoint receives callbacks from imo that inform us when a patient has read our notification. 

## Test instructions

I tested this locally (which is a fairly complicated process) and will do more extensive testing on sandbox.


app/models/imo_delivery_detail.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  enum result: {
+    error: "error",
+    no_imo_account: "no_imo_account",
+    not_subscribed: "not_subscribed",
+    read: "read",
+    sent: "sent"
+  }
+
+  after_create :update_authorization, if: proc {|detail| detail.result.in? ["not_subscribed", "no_imo_account"] }
```
kpethtel, line 12: Move guard to method to more tightly couple the permitted values to the functionality.

db/migrate/20210910200224_create_imo_delivery_detail.rb
```diff
@@ -0,0 +1,11 @@
+class CreateImoDeliveryDetail < ActiveRecord::Migration[5.2]
+  def change
+    create_table :imo_delivery_details do |t|
```
kpethtel, line 3: The TwilioSmsDeliveryDetail also has a `delivered_on` attribute. This is because twilio initially returns a state of `queuing` then uses callbacks to tell us when it was delivered. For Imo, the creation date is the delivery date. We could add the field if it makes querying easier. We could also consider adding a `read_on` attribute to both models, but I think the `updated_at` essentially captures that.

app/models/imo_delivery_detail.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def unsuccessful?
```
kpethtel, line 14: I added these methods to keep the model in line with the twilio version.

app/models/imo_delivery_detail.rb
```diff
@@ -0,0 +1,32 @@
...
+    read?
+  end
+
+  def in_progress?
+    sent?
+  end
+
+  private
+
+  def update_authorization
```
kpethtel, line 28: Update the patient's imo_authorization if the notification failed due to the patient unsubscribing or deleting their account.

app/controllers/api/v3/imo_callbacks_controller.rb
```diff
@@ -21,6 +24,22 @@ def subscribe
     head :ok
   end
 
+  def read_receipt
+    communication = Communication.find_by!(notification_id: params[:notification_id], communication_type: "imo", detailable_type: "ImoDeliveryDetail")
```
kpethtel, line 28: I designed this to be based on notification id because the communication/detail aren't created until after the message sends.

app/models/communication.rb
```diff
@@ -63,16 +63,20 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
...
-                communication_type: "imo", appointment_id: appointment&.id, notification_id: notification&.id
+    transaction do
+      detailable = ImoDeliveryDetail.create!(callee_phone_number: patient.latest_mobile_number, result: response)
+      communication = create!(communication_type: "imo",
+                              detailable: detailable,
+                              appointment: notification.subject,
+                              notification: notification,
+                              device_created_at: now,
+                              device_updated_at: now)
+      logger.info class: self.class.name, msg: "create_with_imo_details", communication_id: communication.id,
```
rsanheim, line 77: nitpick: mind wrapping parens around this method call?

app/models/imo_delivery_detail.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  enum result: {
+    error: "error",
+    no_imo_account: "no_imo_account",
+    not_subscribed: "not_subscribed",
+    read: "read",
+    sent: "sent"
+  }
+
+  after_create :update_authorization, if: proc { |detail| detail.result.in? ["not_subscribed", "no_imo_account"] }
```
rsanheim, line 12: Would you mind moving this responsibility out of a model callback and make it more of a directly called behavior.  Maybe something that is called in the Communication method above.

Having key control flow behavior like this hidden away in a callback can make things hard to follow and reason about down the road as more behavior and code gets added.

db/migrate/20210910200224_create_imo_delivery_detail.rb
rsanheim: Lets go with a timestamp right away, something like `read_at` .. that way its very explicit and clear when we want to calculate time between delivery and reading.

rsanheim (commented):
One question about the broader approach - it sounds like Twilio delivery details are created earlier, when the messaging is sent.  Is there a way to model IMO the same way? It seems strange to have two different state workflows for the two different communication tools.  It will create more work for us down the road for doing things like creating reports on delivery success.

kpethtel:
> One question about the broader approach - it sounds like Twilio delivery details are created earlier, when the messaging is sent. Is there a way to model IMO the same way? It seems strange to have two different state workflows for the two different communication tools. It will create more work for us down the road for doing things like creating reports on delivery success.

Is this in response to my comment about using the notification id? The flows are the same between twilio and imo. In both cases we call them api from the notifications worker, then create communications with detailables. The difference is that twilio provides us a unique SID that we use to identify the detailable when we receive the callback. There is no such SID for imo. So I needed some unique identifier and, since the communication and detailable don't exist when the call to imo is made, I'm using the notification id.

db/migrate/20210910200224_create_imo_delivery_detail.rb
rsanheim: 👍

app/controllers/api/v3/imo_callbacks_controller.rb
```diff
@@ -21,6 +24,24 @@ def subscribe
...
 
+  def read_receipt
+    communication = Communication.includes(:detailable).find_by!(
+      notification_id: params[:notification_id], communication_type: "imo", detailable_type: "ImoDeliveryDetail"
+    )
+    detail = communication.detailable
+    unless detail
+      raise ActiveRecord::RecordNotFound, "no ImoDeliveryDetail found for communication #{communication.id}"
+    end
+    if detail.result == "read"
```
rsanheim, line 35:
```suggestion
    if detail.read?
```

app/jobs/appointment_notification/worker.rb
```diff
@@ -86,7 +86,7 @@ def send_message(notification, communication_type, recipient_number:)
         context: context
       )
     when "imo"
-      notification_service.send_notification(notification.patient, notification.localized_message)
+      notification_service.send_notification(notification)
```
rsanheim, line 89: It looks like IMO will return a `post_id` for messages (see page 6 of the [docs](https://docs.google.com/document/d/1zaTouxdfGg4IqrkCk59KAP905vwynON5Up2ckUax8Mg/).  We can capture that in ImoDeliveryDetails similiar to how we grab the `sid` from Twilio and store it.

This will let us better track IMO notifications going forward, and will more closely match the Twilio flow.  I think the detail workflow will then go from `sent` directly to `read` once the callback comes back in, at least until IMO provides more callback statuses.

db/migrate/20210910200224_create_imo_delivery_detail.rb
kpethtel: Made a story for this: https://app.shortcut.com/simpledotorg/story/5171/add-read-at-to-twilio-detail

app/jobs/appointment_notification/worker.rb
kpethtel: I completely refactored this per your feedback. The `sent` to `read` behavior isn't really a change, but the way we look up the record is different.

app/controllers/api/v3/imo_callbacks_controller.rb
kpethtel: Done.

rsanheim (approved):
Thanks for adding the :post_id bits, this looks good overall so I'm approving to unblock. Please ship away once you've addressed feedback and when you are ready to get this out there. 🚢 🚢 🚢   Note that you could do a temp deploy to SBX if wanted to put this thru its paces.

Had one question about JSON error handling, and some other minor inline feedback.  
I also added some code suggestions about changing strings to symbols for Rails enums, but if you could go thru this and do that generally it would be great.  A symbol is really the better type for an enumerated thing, as it represents a specific thing of importance and isn't a user entered string or the like.

app/models/communication.rb
```diff
@@ -57,22 +57,32 @@ def self.create_with_twilio_details!(appointment:, twilio_sid:, twilio_msg_statu
...
+                              notification: notification,
+                              device_created_at: now,
+                              device_updated_at: now)
+
+      if detailable.result.in?(["not_subscribed", "no_imo_account"])
+        patient = notification.patient
+        patient.imo_authorization.update!(status: detailable.result)
+      end
+
+      logger.info(class: self.class.name, msg: "create_with_imo_details", communication_id: communication.id,
```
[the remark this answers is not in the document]
kpethtel: Good suggestion. Thanks.

app/services/imo_api_service.rb
```diff
@@ -83,18 +85,26 @@ def execute_post(url, data)
     raise Error.new("Error while calling the Imo API", path: url, exception_message: e)
   end
 
-  def process_response(response, url, action)
+  def parse_body(body)
+    JSON.parse(body)
+  rescue JSON::ParserError
```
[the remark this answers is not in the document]
kpethtel: Will do.

spec/controllers/api/v3/imo_callbacks_controller_spec.rb
```diff
@@ -41,15 +51,49 @@
...
+      end
+
+      it "returns 404 when imo_delivery_detail is not found" do
+        params = {post_id: "does_not_exist"}
+
+        post :read_receipt, params: params
+        expect(response.status).to eq(404)
+      end
+
+      it "returns 400 if the result is changed to 'read' multiple times" do
```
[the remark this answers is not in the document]
kpethtel: Good catch.

PR #2927, opened 2021-09-20
Update Punjab protocol drug stock config

prabhanshuguptagit:
**Story card:** [ch5000](https://app.shortcut.com/simpledotorg/story/5000)

## Because
Punjab protocol is missing configuration for Hydrochlorothiazide 12.5mg, causing a calculation error.

## This addresses

Updates config. Fixes a layout issue. https://simpledotorg.slack.com/archives/CFHKJ5WJY/p1631607414000800
- [x] Confirm with swagat on email

### DIFF
--- a/config/data/drug_stock_config.yml
+++ b/config/data/drug_stock_config.yml
@@ -180,7 +180,7 @@ Punjab:
       '979467': 1
     hypertension_diuretic:
       new_patient_coefficient: 0.06
-      '316047': 0.5
+      '429503': 0.5
       '316049': 1
       '331132': 1
 Rajasthan:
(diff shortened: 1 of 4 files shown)


config/data/drug_stock_config.yml
```diff
@@ -180,6 +180,7 @@ Punjab:
       '979467': 1
     hypertension_diuretic:
       new_patient_coefficient: 0.06
+      '316047': 0.5
```
vkrmis, line 183:
```suggestion
      '429503': 0.5
```

vkrmis: It looks like the drugs in our data set are using `429503` _(hydrochlorothiazide 12.5 MG Oral Tablet)_ as the rxnorm_code, rather than  `316047` _(hydrochlorothiazide 12.5 MG)_ 

We should update any other references to `316047`as well

vkrmis (approved):
Updated the rxnorm_code, looks good otherwise

PR #2932, opened 2021-09-21
Update LK patient address labels 

prabhanshuguptagit:
**Story card:** [ch4944](https://app.shortcut.com/simpledotorg/story/4944/we-ll-need-to-fix-the-mapping-on-the-server-side-not-a-big-deal-just-a-labeling-change)

## Because

The patient address labels are different from the facility address labels. `Zone` is not used by the app.

## This addresses

Fixes the patient address labels

### DIFF
--- a/config/locales/en_LK.yml
+++ b/config/locales/en_LK.yml
@@ -14,12 +14,11 @@ en_LK:
       facility_group:
         state: "Province"
       address:
-        street_address: "House and street number"
-        village_or_colony: "Village or colony"
-        zone: "Town *"
+        street_address: "House and road number"
+        village_or_colony: "Town/Area"
         district: "District"
-        state: "Province *"
-        country: "Country *"
+        state: "Province"
+        country: "Country"
         pin: "Postal Code"
       drug:
         frequency:


config/locales/en_LK.yml
```diff
@@ -14,12 +14,11 @@ en_LK:
...
-        street_address: "House and street number"
-        village_or_colony: "Village or colony"
-        zone: "Town *"
+        street_address: "House and road number"
+        village_or_colony: "Town/Area"
         district: "District"
-        state: "Province *"
-        country: "Country *"
+        state: "Province"
+        country: "Country"
```
harimohanraj89, line 21: Our Ethiopia config _has_ the asterisks. Are we sure we want to remove them?

harimohanraj89 (changes requested):

config/locales/en_LK.yml
prabhanshuguptagit: We don't display the patient address fields anywhere on the server, and I was checking if we need to localise the model at all. We don't localise for India, Bangladesh for example. Do you think we can remove this localisation?

harimohanraj89: Ah, I suspect then that this localization is used in patient line lists. Let's keep the localization, and remove the asterisks.

harimohanraj89 (approved):

PR #2935, opened 2021-09-21
Quarter perf optimization

rsanheim:
no need to parse the match twice here.

This may get called a bunch for quarter reports, so worth cleaning this up if its easy.


harimohanraj89 (approved):
Yeah :100:

PR #2936, opened 2021-09-22
Add user id to v4 login error exception reports

harimohanraj89:
**Story card:** [ch5161](https://app.shortcut.com/simpledotorg/story/5161/add-user-id-to-login-error-exception-reports)

## Because

We get this error a lot: https://sentry.io/organizations/resolve-to-save-lives/issues/2651979117/?referrer=slack

And we can't tell why this is happening. Is it bad PIN entries? Is it a deleted user account? Is it a bot crawling?

## This addresses

Adds `user_id` from params into this exception's tags.


harimohanraj89:
I'm gonna abandon this in favor of just moving all of this to DataDog.

PR #2937, opened 2021-09-22
Consolidate refresh logic: make it possible to disable CONCURRENT

rsanheim:
We need to be able to disable concurrent refreshes.  There are times
when we can't use concurrent refresh - for example, when the matviews
are completely empty after loading a fresh database.

Also, there may be times that doing a regular refresh will be more
performant and where the exclusive lock behavior doesn't matter (ie in
dev).

NOTE: This was a cherry pick from https://github.com/simpledotorg/simple-server/pull/2918, where I've been doing a lot of DB loads and matview refreshes to verify the data script there.

**Story card:** [ch5168](https://app.shortcut.com/simpledotorg/story/5168/consolidate-refresh-logic-make-it-possible-to-disable-concurrent)


app/models/reports/refreshable.rb
```diff
@@ -0,0 +1,24 @@
+module Reports
+  module Refreshable
```
rsanheim, line 2: I wanted _one_ place where we call the Scenic method to refresh matviews, so consolidating that here....

app/models/reports/refreshable.rb
```diff
@@ -0,0 +1,24 @@
...
+      ActiveRecord::Base.transaction do
+        refresh_view
+      end
+    end
+
+    private
+
+    def refresh_view
+      ActiveRecord::Base.connection.execute("SET LOCAL TIME ZONE '#{Period::REPORTING_TIME_ZONE}'")
+      Scenic.database.refresh_materialized_view(table_name, concurrently: refresh_concurrently?, cascade: false)
```
rsanheim, line 17: So that we can control `concurrently` here ...

app/models/reports/refreshable.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    def refresh_concurrently?
+      ActiveModel::Type::Boolean.new.cast(ENV["REFRESH_MATVIEWS_CONCURRENTLY"]) || true
```
rsanheim, line 21: And here we can disable it with a ENV var of false of 0 by using the Boolean cast thing from ActiveModel.

We still default to true if there is no ENV var (which is true in all deployed envs).

app/services/refresh_reporting_views.rb
```diff
@@ -74,15 +65,11 @@ def self.tz
   # LatestBloodPressuresPerPatientPerMonth should be refreshed before
   # LatestBloodPressuresPerPatientPerQuarter and LatestBloodPressuresPerPatient
   def refresh_v1
-    ActiveRecord::Base.transaction do
```
rsanheim, line 77: The tx happens for each `refresh` call now for all matviews, so no need for an outer transaction.

rsanheim (commented):
Notes for reviewers

kpethtel (approved):
My one piece of feedback is that perhaps this would be a good place to add a shared example for testing. But... this may be a challenge to test.

app/services/refresh_reporting_views.rb
```diff
@@ -15,6 +15,7 @@ def self.call
   end
 
   attr_reader :logger
+  delegate :tz, :set_last_updated_at, to: self
```
timcheadle, line 18: Never seen a delegation to self before; what does this do?

timcheadle (approved):

harimohanraj89 (changes requested):
One blocking change from me.

app/models/reports/view.rb
```diff
@@ -1,17 +1,14 @@
 module Reports
   class View < ActiveRecord::Base
+    extend Refreshable
```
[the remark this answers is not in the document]
rsanheim: Yeah, there were too many small differences amongst all our matview objects and I wanted one tiny module to _just_ change `refresh` behavior.

app/services/refresh_reporting_views.rb
rsanheim: It delegates an instance method to the class - `self` is the class when you call `delegate` ->

```ruby
[7] pry(main)> class Test
[7] pry(main)*   puts self
[7] pry(main)*   def self.hi
[7] pry(main)*     puts "hello from a class method"
[7] pry(main)*   end
[7] pry(main)*   delegate :hi, to: self
[7] pry(main)* end
Test
=> [:hi]
[8] pry(main)> Test.new.hi
hello from a class method
```

app/models/reports/refreshable.rb
rsanheim: Thanks, fixed that up.

PR #2940, opened 2021-09-23
5171: add read_at to twilio detail

kpethtel:
**Story card:** [ch5171](https://app.shortcut.com/simpledotorg/story/5171/add-read-at-to-twilio-detail)

## Because

Twilio detail already has a delivered_on attribute that's updated when the message is delivered. We added a `read_at` attribute to the new imo detail model. This is to keep parity and should make queries simpler.


timcheadle (approved):
Looks good. For Twilio or Imo, do we get two different callbacks for `delivered` and `read`, or just one?

kpethtel:
> Looks good. For Twilio or Imo, do we get two different callbacks for `delivered` and `read`, or just one?

Twilio initially just tells us the notification is queued, then gives us a callback for delivered. Then gives us a callback for read. 

Imo skips the queued state and just tells us 'sent', then gives us a callback for read. "Read" is the only callback option for imo at present.

prabhanshuguptagit (approved):

rsanheim:
✨ ✨ ✨

PR #2946, opened 2021-09-28
Improve sync other facilities performance take 2

ssrihari:
Fixes a shortcoming in https://github.com/simpledotorg/simple-server/pull/2943 where we were doing a redundant selection from the patients table.


app/models/region.rb
```diff
@@ -134,24 +134,35 @@ def dashboard_analytics(period:, prev_periods:, include_current_period: true)
...
-  def registered_patients
-    Patient.where(registration_facility: facility_regions.pluck(:source_id))
+  def syncable_patient_ids(exclude_facility: nil)
+    case region_type
+      when "block"
+        registered_patients(exclude_facility: exclude_facility)
+          .with_discarded.select(:id)
+          .or(assigned_patients(exclude_facility: exclude_facility).with_discarded.select(:id))
+          .union(appointed_patient_ids(exclude_facility: exclude_facility).with_discarded)
+          .pluck(:id)
```
prabhanshuguptagit, line 150: Having both `select` and `pluck` is a bit unintuitive, shall we leave a comment here explaining the rationale?

app/models/facility_group.rb
```diff
@@ -105,6 +105,10 @@ def syncable_patients
     registered_patients.with_discarded
   end
 
+  def syncable_patient_ids(exclude_facility: nil)
+    syncable_patients.pluck(:id)
+  end
```
prabhanshuguptagit, line 110: This is missing the `exclude_facility` exclusion. Let's add a spec for this?

app/models/region.rb
```diff
@@ -134,24 +134,35 @@ def dashboard_analytics(period:, prev_periods:, include_current_period: true)
...
 
-  def registered_patients
-    Patient.where(registration_facility: facility_regions.pluck(:source_id))
+  def syncable_patient_ids(exclude_facility: nil)
+    case region_type
+      when "block"
+        registered_patients(exclude_facility: exclude_facility)
+          .with_discarded.select(:id)
+          .or(assigned_patients(exclude_facility: exclude_facility).with_discarded.select(:id))
+          .union(appointed_patient_ids(exclude_facility: exclude_facility).with_discarded)
```
prabhanshuguptagit, line 149: This is not accurate, the `current_facility` exclusion only excludes registered patients in the current facility. Just line 146 should be it I think.

prabhanshuguptagit (changes requested):
One blocking comment around correctness of the exclusion, rest looks good.

app/models/region.rb
ssrihari: No need to pluck, fixed in a0edbc5ce3ed820302f5c2934f74688de9edc445.

app/models/facility_group.rb
ssrihari: Fixed in a0edbc5ce3ed820302f5c2934f74688de9edc445. Will add a spec though.

app/models/region.rb
ssrihari: Hrm. It's more complex than that, unfortunately. We need to exclude patients registered in the `current_facility` from all three groups :/. Thinking of a different solution now, since the appointment related changes are not viable.

prabhanshuguptagit: Oof yeah. Would using a `pluck` instead of a `select` at the top level work allow us to keep the top level `not` clause?

ssrihari:
Trying a different approach in https://github.com/simpledotorg/simple-server/pull/2949.

PR #2947, opened 2021-09-28
Sri lanka personal health number

prabhanshuguptagit:
**Story card:** [ch5212](https://app.shortcut.com/simpledotorg/story/5212/capture-personal-health-number-instead-of-national-id)

## Because

We want to capture patients' Personal Health Number as their business ID in Sri lanka, not the National Health ID.

## This addresses

Adds personal health number to accepted health number types. The Sri Lanka national ID is not removed for a possible use case that we might add in the future. [Slack ref](https://simpledotorg.slack.com/archives/C0176A5B0GN/p1632816477026300)


harimohanraj89 (approved):
I'm guessing our API and API docs automatically consume this enum's values, nice.

The name seems good to me, let's confirm it with the android team before we 🚀

msasikanth:
> I'm guessing our API and API docs automatically consume this enum's values, nice.
> 
> The name seems good to me, let's confirm it with the android team before we 🚀

Yep, looks good 👍🏾 . We are using that name in our PR as well.

prabhanshuguptagit:
@msasikanth Thanks for looking at it 😄

PR #2949, opened 2021-09-28
Perform a union on required columns only, not full rows

ssrihari:
Superseeds https://github.com/simpledotorg/simple-server/pull/2946.
Fixes a shortcoming in #2943 where we were doing a redundant selection from the patients table.

See https://explain.depesz.com/s/f5ko for query performance analysis.


app/models/region.rb
```diff
@@ -135,8 +135,9 @@ def syncable_patients
     case region_type
       when "block"
         registered_patients.with_discarded
-          .or(assigned_patients.with_discarded)
-          .union(appointed_patients.with_discarded)
+          .select(:id, :registration_facility_id, :updated_at)
+          .or(assigned_patients.with_discarded.select(:id, :registration_facility_id, :updated_at))
+          .union(appointed_patients.with_discarded.select(:id, :registration_facility_id, :updated_at))
```
prabhanshuguptagit, line 140: What's the use for selecting `registration_facility_id` and `updated_at` here? Why not just `select(:id)`?

app/models/region.rb
```diff
@@ -135,8 +135,9 @@ def syncable_patients
     case region_type
```
prabhanshuguptagit, line 135: `def syncable_patient_ids`? `syncable_patients` could just `Patient.where(id: syncable_patient_ids)` like in the previous PR.

prabhanshuguptagit (changes requested):

PR #2950, opened 2021-09-28
Remove unused indexes

ssrihari:
- Story: [ch5219](https://app.shortcut.com/simpledotorg/story/5219/removing-unnecessary-indexes)
- These indexes add to the write latency, and to the disk size.
- This is just routine maintenance, and will have some mostly minor performance benefits
- [Query used to find these indexes](https://github.com/nilenso/postgresql-monitoring#index_usage)


rsanheim (approved):
Looks like some conflicts to fix up, but otherwise LGTM!

PR #2952, opened 2021-09-28
4070: add optional phone number argument to imo invitation to enable …

kpethtel:
…easier testing

**Story card:** [ch4070](https://app.shortcut.com/simpledotorg/story/4070/test-imo-invitations)

## Because

This makes testing imo easier.

### DIFF
--- a/spec/services/imo_api_service_spec.rb
+++ b/spec/services/imo_api_service_spec.rb
@@ -40,6 +40,19 @@ describe ImoApiService, type: :model do
       before { Flipper.enable(:imo_messaging) }
 
       it "creates an ImoAuthorization on 200 success" do
+        locale = "bn-BD"
+        request_body = JSON(
+          phone: patient.latest_mobile_number,
+          msg: I18n.t("notifications.imo.invitations.message", patient_name: patient.full_name, locale: locale),
+          contents: [{
+            key: I18n.t("notifications.imo.invitations.message_key", locale: locale),
+            value: I18n.t("notifications.imo.invitations.message", patient_name: patient.full_name, locale: locale)
+          }],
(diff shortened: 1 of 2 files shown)


spec/services/imo_api_service_spec.rb
```diff
@@ -85,6 +98,29 @@
...
+          contents: [{
+            key: I18n.t("notifications.imo.invitations.message_key", locale: locale),
+            value: I18n.t("notifications.imo.invitations.message", patient_name: patient.full_name, locale: locale)
+          }],
+          title: I18n.t("notifications.imo.invitations.title", locale: locale),
+          action: I18n.t("notifications.imo.invitations.action", locale: locale),
+          callback_url: "https://localhost/api/v3/patients/#{patient.id}/imo_authorization"
+        )
+
+        stub_request(:post, request_url).with(headers: request_headers, body: request_body).to_return(status: 200, body: success_body)
```
rsanheim, line 117: I'm trying to remember the Webmock details, but I am pretty sure this will _not_ verify the phone number is passed through in the body for the request made.  It just sets a stub, not a verified mock.

It looks like you can pretty easily verify with https://github.com/bblimke/webmock#setting-expectations-in-rspec-on-the-stub.

So grab the stub, and verify after the API is called.

rsanheim (commented):
bit of feedback on the test and verifying what we have under test

rsanheim (commented):
Testing a review for our slack notifications...

spec/services/imo_api_service_spec.rb
kpethtel: That's an excellent point. These tests could easily continue passing without verifying the phone number portion if we removed the underlying web request. Thanks for pointing me in a better direction. Fixed it.

PR #2953, opened 2021-09-28
Disable net connections in regular suite:

rsanheim:
and only allow them in our features suite for Webmock.

no story, just have noticed this when reviewing PRs and it would be good to have this safety.

### DIFF
--- a/spec/features_helper.rb
+++ b/spec/features_helper.rb
@@ -5,6 +5,8 @@ require "capybara/rails"
 
 Dir[Rails.root.join("spec/pages/**/*.rb")].sort.each { |f| require f }
 
+WebMock.allow_net_connect!
+
 RSpec.configure do |config|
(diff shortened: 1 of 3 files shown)


spec/services/imo_api_service_spec.rb
```diff
@@ -40,7 +40,7 @@
       before { Flipper.enable(:imo_messaging) }
 
       it "creates an ImoAuthorization on 200 success" do
-        stub_request(:post, request_url).with(headers: request_headers).to_return(status: 200, body: success_body)
+        # stub_request(:post, request_url).with(headers: request_headers).to_return(status: 200, body: success_body)
```
harimohanraj89, line 43: I'm guessing this was to test the safety measure? We can revert this before shipping.

harimohanraj89 (approved):
Great catch, yes let's disable net connect as the default 👌

PR #2956, opened 2021-09-29
Fix patient lookup for discarded patients

prabhanshuguptagit:
**Story card:** [ch5242](https://app.shortcut.com/simpledotorg/story/5242)

## Because

When a discarded patient has a remaining business identifier that has not been discarded, the lookup fails with a 500. [Sentry](https://sentry.io/organizations/resolve-to-save-lives/issues/2557456792/?referrer=slack)

## This addresses
[Sample datadog trace](https://app.datadoghq.com/logs?query=trace_id%3A19177855750868170&cols=host%2Cservice%2Cenv&event=AQAAAXwYWSHImemQqwAAAABBWHdZV1RZUEFBQnVVSnBkQ0c5bk5RQUI&index=&messageDisplay=inline&stream_sort=desc&from_ts=1632495066121&to_ts=1632497066271&live=false)

- [ ] Investigate why some discarded patients have undiscarded business identifiers
- [ ] Fix lookup logic to account for this scenario

### DIFF
--- a/spec/controllers/api/v4/patients_controller_spec.rb
+++ b/spec/controllers/api/v4/patients_controller_spec.rb
@@ -79,6 +79,30 @@ RSpec.describe Api::V4::PatientsController, type: :controller do
       expect(response_data[:patients].count).to eq 2
     end
 
+    it "returns nothing when an identifier is not discarded but patient is discarded" do
+      patient_1 = create(:patient)
+      patient_2 = create(:patient, registration_facility: patient_1.registration_facility)
+      b = patient_1.business_identifiers.first
+      patient_1.business_identifiers.first.update(
+        identifier_type: PatientBusinessIdentifier.identifier_types.values.first
+      )
+
+      patient_2.business_identifiers.first.update(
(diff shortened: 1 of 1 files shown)


prabhanshuguptagit:
@ssrihari There's two options here:
- Make sure the transformer can deal with discarded patients
- Skip discarded patients in the lookup API? The app doesn't keep discarded records anyways. This is the simpler fix that I am leaning towards.

ssrihari:
Re the 2 options @prabhanshuguptagit, I think we should do both, ideally.

1. If there is data coming in for a discarded patient, we should mark the incoming data also discarded. This is not specific to business identifiers, and we should fix this logic for all entities.
2. Lookup shouldn't return discarded patients.

prabhanshuguptagit:
@ssrihari Created a maintenance story for 1. here https://app.shortcut.com/simpledotorg/story/5327/mark-incoming-data-for-discarded-patients-as-discarded. Since this bug is fixed by #2949 already, this PR just adds a spec.

PR #2957, opened 2021-09-30
Memoize MO transformation to avoid redundant computation

ssrihari:
- Story: [ch5328](https://app.shortcut.com/simpledotorg/story/5328/improve-performance-of-facility-medical-officer-sync)
- An MO can belong to multiple facilities in a sync region
- While responding from the controller, we end up transforming the same MO multiple times, which is expensive.


timcheadle (approved):

app/transformers/api/v4/teleconsultation_medical_officer_transformer.rb
```diff
@@ -1,6 +1,7 @@
 class Api::V4::TeleconsultationMedicalOfficerTransformer
   class << self
-    def to_response(medical_officer)
+    include Memery
+    memoize def to_response(medical_officer)
```
rsanheim, line 4: I think instead of doing this we could just retrieve distinct medical_officers, no?

ssrihari: We do in fact retrieve distinct MOs from the DB. The API wants to return a record for every MO<>Facility pair, and we end up doing this transformation multiple (100s of) times. The bulk of the time is spent in application code as per DD as well.

PR #2958, opened 2021-09-30
Reorganise Chennai hierarchy in IHCI

prabhanshuguptagit:
**Story card:** [ch4249](https://app.shortcut.com/simpledotorg/story/4249/fix-chennai-region-hierarchy)

## Because

We created an FG per block in Chennai to mitigate performance concerns before we had block level sync. Since we are able to block sync now, we can do away with these and combine them into a single FG called Chennai.

## This addresses

Adds a rake task to reorganize . This should . The mechanism is:
- Create a district called `Chennai` in `Tamil Nadu` from the dashboard
- All blocks are reparented to the new district, all old districts are deleted

### DIFF
--- a/lib/tasks/scripts/chennai_hierarchy_cleanup.rb
+++ b/lib/tasks/scripts/chennai_hierarchy_cleanup.rb
@@ -1,0 +1,22 @@
+class ChennaiHierarchyCleanup
+  def self.run
+    raise "Cannot run this outside India production" unless CountryConfig.current[:name] == "India" && SimpleServer.env.production?
+
+    state = Region.state_regions.find_by_name("Tamil Nadu")
+    new_district = Region.find_by_name("Chennai")
+    raise "Create Chennai district before running" if new_district.blank?
+
+    old_districts = state.district_regions.where.not(name: new_district.name)
+
+    ActiveRecord::Base.transaction do
+      state.block_regions.each do |block|
+        block.reparent_to = new_district
+        block.save!
+      end
+
+      state.facilities.update_all(facility_group_id: new_district.source.id)
+
+      FacilityGroup.where(id: old_districts.pluck(:source_id)).discard_all
+    end
+  end
+end
--- a/lib/tasks/scripts/chennai_hierarchy_cleanup.rb
+++ b/lib/tasks/scripts/chennai_hierarchy_cleanup.rb
@@ -16,6 +16,7 @@ class ChennaiHierarchyCleanup
 
       state.facilities.update_all(facility_group_id: new_district.source.id)
 
+      # make sure facility gropus are empty before deleting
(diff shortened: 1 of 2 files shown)


lib/tasks/scripts/chennai_hierarchy_cleanup.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    new.call
+  end
+
+  def initialize
+    @state = Region.state_regions.find_by_name("Tamil Nadu")
+    @new_district = Region.find_by_name("Chennai")
+    raise "Create Chennai district from dashboard before running" if @new_district.blank?
+
+    @old_districts = @state.district_regions.where.not(name: @new_district.name)
```
harimohanraj89, line 13: This assumes that Chennai's the only place in TN. Can we make it a bit smarter by grepping for `Chennai...`?

If that's awkward, I'd even prefer to see a flat list of Chennai zone names here, instead of slurping in all TN districts.

lib/tasks/scripts/chennai_hierarchy_cleanup.rb
```diff
@@ -0,0 +1,52 @@
...
+    @state.facilities.update_all(facility_group_id: @new_district.source.id)
+  end
+
+  def copy_accesses
+    accesses = Access.where(resource_id: @old_districts.pluck(:source_id))
+    user_ids_with_access = accesses.pluck(:user_id).uniq
+
+    user_ids_with_access.each do |user_id|
+      Access.create!(user_id: user_id, resource_type: "FacilityGroup", resource_id: @new_district.source_id)
+    end
```
harimohanraj89, line 42: This means any dashboard admin with access to _one_ Chennai zone will now gain access to _all_ Chennai zones, yes? I think that's totally fine, but worth letting Praveen know.

lib/tasks/scripts/chennai_hierarchy_cleanup.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    accesses.discard_all
+  end
+
+  def discard_old_fgs
+    raise "Old facility groups still have facilities" if FacilityGroup.where(id: @old_districts.pluck(:source_id)).map(&:facilities).any? { |facilities| facilities.count > 0 }
```
harimohanraj89, line 48: We've got a `discardable?` method on `FacilityGroup`. Should we use that?

harimohanraj89 (changes requested):
Minor updates.

lib/tasks/scripts/chennai_hierarchy_cleanup.rb
prabhanshuguptagit: Oh nice. Added in f6bf9e7

harimohanraj89 (approved):
You could execute this with a data migration if you'd like. No worries either way, the class interface is very straightforward.

PR #2959, opened 2021-09-30
5121 imo key translations

kpethtel:
**Story card:** [ch5121](https://app.shortcut.com/simpledotorg/story/5121/add-section-key-translations-for-imo-notifications)

## Because

We need to add translations for some additional portions of the imo cards. I also added English translations because it makes testing easier and allows developers to know what the Bangla strings translate to in English.

## Testing

I've tested locally and will test again on sandbox.


app/services/imo_api_service.rb
```diff
@@ -28,22 +28,29 @@ def send_invitation(patient, phone_number: nil)
...
-      msg: I18n.t("notifications.imo.invitations.message", patient_name: patient.full_name, locale: locale),
-      contents: [{
-        key: I18n.t("notifications.imo.invitations.message_key", locale: locale),
-        value: I18n.t("notifications.imo.invitations.message", patient_name: patient.full_name, locale: locale)
-      }],
-      title: I18n.t("notifications.imo.invitations.title", locale: locale),
-      action: I18n.t("notifications.imo.invitations.action", locale: locale),
+      msg: I18n.t("notifications.imo.invitation.request", patient_name: patient.full_name, locale: locale),
+      contents: [
+        {
```
kpethtel, line 40: I'm adding the name section because Daniel said it was worth keeping for notifications and I figure we should do the same for invitations.

app/services/imo_api_service.rb
```diff
@@ -28,22 +28,29 @@ def send_invitation(patient, phone_number: nil)
     phone = phone_number || patient.latest_mobile_number
     locale = patient.locale
     url = IMO_BASE_URL + "send_invite"
+
+    unless locale.in?(["bn-BD", "en"])
```
rsanheim, line 32: I don't think you need this check here - Rails will raise an error by default if this path would somehow get called in a locale w/o the IMO translations.

app/services/imo_api_service.rb
```diff
@@ -57,14 +64,29 @@ def send_notification(notification, phone_number)
     Statsd.instance.increment("imo.notifications.attempt")
 
     patient = notification.patient
+    locale = patient.locale
     message = notification.localized_message
     url = IMO_BASE_URL + "send_notification"
+
+    unless locale.in?(["bn-BD", "en"])
```
rsanheim, line 71: same here, no need for this

rsanheim (approved):
One bit about unnecessary error handling, but otherwise looks good. 👍

app/services/imo_api_service.rb
kpethtel: That wasn't my experience. In specs, it defaulted to English. I think this is the cause: 
```
    config.i18n.fallbacks = [:en]
    config.i18n.default_locale = :en
 ```

PR #2963, opened 2021-10-05
5353 already invited or subscribed

kpethtel:
**Story card:** [ch5353](https://app.shortcut.com/simpledotorg/story/5353/add-handling-for-already-invited-and-already-subscribed-patients)

## Because

## This addresses

## Test instructions

### DIFF
--- a/app/jobs/imo/invite_patient.rb
+++ b/app/jobs/imo/invite_patient.rb
@@ -7,6 +7,7 @@ class Imo::InvitePatient
     return unless Flipper.enabled?(:imo_messaging)
 
     patient = Patient.find(patient_id)
-    ImoApiService.new.send_invitation(patient)
+    status = ImoApiService.new.send_invitation(patient)
+    ImoAuthorization.create!(patient: patient, status: status, last_invited_at: Time.current)
   end
 end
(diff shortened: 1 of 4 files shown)


app/jobs/imo/invite_patient.rb
```diff
@@ -7,6 +7,12 @@ def perform(patient_id)
     return unless Flipper.enabled?(:imo_messaging)
 
     patient = Patient.find(patient_id)
-    ImoApiService.new.send_invitation(patient)
+    status = ImoApiService.new.send_invitation(patient)
+
+    if patient.imo_authorization
+      patient.imo_authorization.update!(status: status, last_invited_at: Time.current)
+    else
+      ImoAuthorization.create!(patient: patient, status: status, last_invited_at: Time.current)
```
kpethtel, line 15: I originally had the imo auth happen here but moved it out for reasons I don't recall. That was a mistake that's bothered me ever since, so I'm moving it back to where it belongs and we have better separation of concerns.

app/jobs/imo/invite_patient.rb
```diff
@@ -7,6 +7,12 @@ def perform(patient_id)
     return unless Flipper.enabled?(:imo_messaging)
 
     patient = Patient.find(patient_id)
-    ImoApiService.new.send_invitation(patient)
+    status = ImoApiService.new.send_invitation(patient)
```
rsanheim, line 10: Should we check the status here?  If Imo returns some non 200 result code (and we get back and `:error` for status), I don't think we'd want to create or update the ImoAuthorization below.

rsanheim (commented):
Looks good, just have one question on handling `:error` status.

PR #2966, opened 2021-10-06
Use SRI to improve cross domain security

ssrihari:
- Add integrity fields where possible: chart.js, chart plugin, etc
- Google fonts generates different assets based on the browser, so it's not possible to add SRI: https://github.com/google/fonts/issues/473
- The asset for redoc is generated server side, and the asset is also dynamic: https://www.jsdelivr.com/using-sri-with-dynamic-files

**Story card:** [ch5026](https://app.shortcut.com/simpledotorg/story/5026/medium-cross-domain-script-include)


ssrihari:
Thanks @rsanheim, added a howto with 0ac780b99a59a37a29ccfbfbdadde136790e4c61.

PR #2967, opened 2021-10-06
Behave the same regardless if the e-mail provided was right or wrong

ssrihari:
**Story card:** [ch5028](https://app.shortcut.com/simpledotorg/story/5028/low-user-mail-id-enumeration)


rsanheim:
Verified on the review app:

image

This can definitely result in confusion if a user fat-fingers their email or doesn't remember what email they used, but we can easily roll this back if we get too many support requests on it.

PR #3001, opened 2021-10-20
Make BD script spec a bit smarter about deleted blocks

rsanheim:
Orphaned blocks can change due to variability in our factory data, so
lets just let the spec account for it

[ch4649](https://app.shortcut.com/simpledotorg/story/4649/change-bangladesh-region-structure-and-import-new-community-facilities)

cc @prabhanshuguptagit


rsanheim:
This is blocking deploys because the spec keeps   ##flaking - going to merge away.

PR #3012, opened 2021-10-23
Move flags to flipper

harimohanraj89:
**Story card:** [ch4290](https://app.shortcut.com/simpledotorg/story/4290/move-enable-vars-to-flipper?vc_group_by=day&cf_workflow=500000031&ct_workflow=all)

[ch5358](https://github.com/simpledotorg/simple-server/pull/3012)

## Because

We want to move our ENABLE_ env-based feature flags to Flipper.

Here's the matrix of feature flags that we'l need to set in production before shipping this PR

Screen Shot 2021-10-25 at 11 47 24 PM

* [x] IN prod: exotel_whitelist_api
* [x] IN prod: skip_api_validation
* [x] QA: generate_encounter_id_endpoint
* [x] Sandbox: generate_encounter_id_endpoint

Note that we are also _removing_ the purge endpoint.


app/controllers/api/v3/encounters_controller.rb
```diff
@@ -11,6 +11,8 @@ def sync_to_user
   end
 
   def generate_id
+    raise ActionController::RoutingError.new("Not Found") unless Flipper.enabled?("generate_encounter_id_endpoint")
```
harimohanraj89, line 14: We can't use Flipper in the routes file (it's not initialized yet), so we resort to manually returning a "404 sorry" in the controller here.

rsanheim (approved):
Looks good!

One question - do we ever have encounters turned off? I think much of our system assumes and depends on encounters being created. Maybe its time to remove that feature flag altogether?

harimohanraj89:
We _do_ have "encounters turned off" in all envs except some test ones. The catch is - that feature flag controls whether encounters are syncable like any other sync resource (remnants of our initial efforts to propagate encounters across our ecosystem).

When the flag's off, our ecosystem instead relies on encounters being auto-created by the server when receiving BP or blood sugar records. These created encounters are never synced back to phone - in fact, the app doesn't know how to sync or store encounters.

You could argue that we should remove the feature flag the other way, keeping the thing off everywhere, but I'm not sure if we'll start using those codepaths soon. Additionally, I'd opt to keep that change out of the scope of this PR as well. What do you think?

rsanheim:
> We _do_ have "encounters turned off" in all envs except some test ones. The catch is - that feature flag controls whether encounters are syncable like any other sync resource (remnants of our initial efforts to propagate encounters across our ecosystem).
> 
> When the flag's off, our ecosystem instead relies on encounters being auto-created by the server when receiving BP or blood sugar records. These created encounters are never synced back to phone - in fact, the app doesn't know how to sync or store encounters.
> 
> You could argue that we should remove the feature flag the other way, keeping the thing off everywhere, but I'm not sure if we'll start using those codepaths soon. Additionally, I'd opt to keep that change out of the scope of this PR as well. What do you think?

I agree it sounds outside the scope of this PR.

Long term it sounds like we should just remove that flag, as changing how encounters work is a big project and not one we are likely to tackle right now.  So the flag is really just noise right now.

PR #3019, opened 2021-10-27
Get quarterly cohort numbers from v2 schema when enabled

rsanheim:
**Story card:** [ch5871](https://app.shortcut.com/simpledotorg/story/5871/grab-cohort-data-from-reporting-pipeline)

## Because

We have the cohort numbers in the reporting pipeline but need to grab them and display them for reports

## This addresses

showing reports from the pipeline for cohort

## Test instructions

check out the branch, view cohort reports, cycle between v1 and v2 via query string to compare


rsanheim:
There are some controller specs that need fixing, but I can handle that Tuesday.  The core services are all in order finally.

app/services/reports/region_cache_warmer.rb
```diff
@@ -62,6 +63,9 @@ def warm_patient_breakdown_caches(batch)
       end
     end
 
+    def warm_cohort_caches(regions)
+    end
```
ssrihari, line 67: Is this placeholder for upcoming code?

app/views/reports/regions/cohort.html.erb
```diff
@@ -43,7 +43,7 @@
           <div class="col-lg-9 order-lg-2">
             <table class="split-bars">
               <tr>
-                <% if cohort["registered"] > 0 %>
+                <% if cohort["registered"] && cohort["registered"] > 0 %>
```
ssrihari, line 46: We could add a `.present?` to be more explicit.

ssrihari (approved):
Excited for this to go in!

PR #3021, opened 2021-10-28
A/B enhancements: Schedule daily notifications

prabhanshuguptagit:
**Story card:** [ch5676](https://app.shortcut.com/simpledotorg/story/5676/schedule-notifications-for-the-same-day-as-part-of-daily-runner)

## Because
The existing runner schedules notifications for a patient ahead of time. We want to send messages to patients without scheduling to minimise room for erroneous notifications being sent out.

## This addresses

This implements the `Send messages` section in the experiment [flow](https://docs.google.com/document/d/1IMXu_ca9xKU8Xox_3v403ZdvNGQzczLWljy7LQ6RQ6A/edit#). 
> **Current patients**
> - For patients in control: Do nothing
> - For patients in single notification group: Send “your appt. is tomorrow” message
> - For patients in cascade group: 
>   - For patients expected tomorrow: Send “your appt. is tomorrow” message 
>   - For patients expected today: Send “your appt. is today” message
>   - For patients expected 3 days ago: Send “your appt. was 3 days ago” message
> 
> **Stale patients:**
> - For patients in control: Do nothing
> - For patients who were enrolled today: Send “you are late for your BP” notification

The "number of days" are determined from the reminder templates in the experiment at that point in time. Note that the notifications are "scheduled" for the same day, until the `AppointmentNotification::Worker` picks it up during the messaging window.

### DIFF
--- a/app/models/experimentation/current_patient_experiment.rb
+++ b/app/models/experimentation/current_patient_experiment.rb
@@ -14,10 +14,12 @@ module Experimentation
 
     # Patients whose expected return date falls on
     # one of the reminder template's remind_on days since `date`.
-    def memberships_to_notify(reminder_template, date)
-      treatment_group_memberships.status_enrolled.where(
-        expected_return_date: date - reminder_template.remind_on_in_days.days
-      )
+    def memberships_to_notify(date)
+      treatment_group_memberships
+        .status_enrolled
+        .joins(treatment_group: :reminder_templates)
+        .where("expected_return_date::timestamp + make_interval(days := reminder_templates.remind_on_in_days) = ?", date)
+        .select("treatment_group_memberships.*, reminder_templates.id template_id, reminder_templates.message message")
     end
 
     private
--- a/app/models/experimentation/current_patient_experiment.rb
+++ b/app/models/experimentation/current_patient_experiment.rb
@@ -12,14 +12,13 @@ module Experimentation
         .distinct
     end
 
-    # Patients whose expected return date falls on
+    # Memberships where the expected return date falls on
     # one of the reminder template's remind_on days since `date`.
     def memberships_to_notify(date)
       treatment_group_memberships
         .status_enrolled
         .joins(treatment_group: :reminder_templates)
         .where("expected_return_date::timestamp + make_interval(days := reminder_templates.remind_on_in_days) = ?", date)
-        .select("treatment_group_memberships.*, reminder_templates.id template_id, reminder_templates.message message")
     end
 
     private
--- a/app/models/experimentation/current_patient_experiment.rb
+++ b/app/models/experimentation/current_patient_experiment.rb
@@ -12,9 +12,12 @@ module Experimentation
         .distinct
     end
 
-    def memberships_to_notify(date)
-      # Patients where `date` equals one of their reminder template's remind_on.
-      # To be implemented in a follow up PR.
+    # Patients whose expected return date falls on
+    # one of the reminder template's remind_on days since `date`.
+    def memberships_to_notify(reminder_template, date)
+      treatment_group_memberships.status_enrolled.where(
+        expected_return_date: date - reminder_template.remind_on_in_days.days
+      )
     end
 
     private
--- a/app/models/experimentation/notifications_experiment.rb
+++ b/app/models/experimentation/notifications_experiment.rb
@@ -107,11 +107,11 @@ module Experimentation
         registration_facility_state: registration_facility&.state
       }
     end
-  end
 
-  def mark_visits
-  end
+    def mark_visits
+    end
 
-  def evict_patients
+    def evict_patients
+    end
   end
 end
--- a/app/models/experimentation/notifications_experiment.rb
+++ b/app/models/experimentation/notifications_experiment.rb
@@ -59,26 +59,7 @@ module Experimentation
     def schedule_notifications(date)
       memberships_to_notify(date)
         .in_batches(of: MEMBERSHIPS_BATCH_SIZE)
-        .each_record do |membership|
-        schedule_notification(membership.patient_id, membership.template_id, membership.message, date)
-      end
-    end
-
-    def schedule_notification(patient_id, template_id, message, date)
-      Notification.where(
-        experiment: self,
-        reminder_template_id: template_id,
-        patient_id: patient_id
-      ).exists? ||
-        Notification.create!(
-          experiment: self,
-          message: message,
-          patient_id: patient_id,
-          purpose: :experimental_appointment_reminder,
-          remind_on: date,
-          reminder_template_id: template_id,
-          status: "pending"
-        )
+        .each_record { |membership| schedule_notification(membership, date) }
     end
 
     def cancel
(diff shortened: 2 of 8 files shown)


app/models/experimentation/current_patient_experiment.rb
```diff
@@ -12,9 +12,12 @@ def eligible_patients(date)
         .distinct
     end
 
-    def memberships_to_notify(date)
-      # Patients where `date` equals one of their reminder template's remind_on.
-      # To be implemented in a follow up PR.
+    # Patients whose expected return date falls on
+    # one of the reminder template's remind_on days since `date`.
+    def memberships_to_notify(reminder_template, date)
```
ssrihari, line 17: I think I asked this in a different PR earlier: can we call this `patients_to_notify`? We don't actually notify the "Membership", which is a relationship between the patient and experiment.

app/models/experimentation/notifications_experiment.rb
```diff
@@ -39,22 +41,45 @@ def self.eligible_patients
...
-    def enroll_patients(date)
+    def enroll_patients(date, limit = MAX_PATIENTS_PER_DAY)
       eligible_patients(date)
-        .limit(MAX_PATIENTS_PER_DAY)
+        .limit([remaining_enrollments_allowed(date), limit].min)
         .includes(:assigned_facility, :registration_facility, :medical_history)
         .includes(latest_scheduled_appointments: [:facility, :creation_facility])
-        .in_batches(of: ENROLLMENT_BATCH_SIZE)
+        .in_batches(of: MEMBERSHIPS_BATCH_SIZE)
         .each_record { |patient| random_treatment_group.enroll(patient, reporting_data(patient, date)) }
```
ssrihari, line 50: Do we want to log how many patients were enrolled? Or log anything else that comes out of this?

app/models/experimentation/notifications_experiment.rb
```diff
@@ -12,10 +12,12 @@ class NotificationsExperiment < Experiment
         .distinct
     end
 
+    # The order of operations is important.
+    # See https://docs.google.com/document/d/1IMXu_ca9xKU8Xox_3v403ZdvNGQzczLWljy7LQ6RQ6A for more details.
     def self.conduct_daily(date)
       running.each { |experiment| experiment.enroll_patients(date) }
       monitoring.each { |experiment| experiment.monitor(date) }
-      notifying.each { |experiment| experiment.send_notifications(date) }
+      notifying.each { |experiment| experiment.schedule_notifications(date) }
```
ssrihari, line 20: I think I had asked this in a different PR as well: do we need to handle any errors here?

app/models/experimentation/notifications_experiment.rb
```diff
@@ -39,22 +41,45 @@ def self.eligible_patients
...
+          .each_record { |membership| schedule_notification(membership, template, date) }
+      end
+    end
+
+    def schedule_notification(membership, template, date)
+      return if Notification.find_by(
+        reminder_template: template,
+        experiment: self,
+        patient_id: membership.patient_id
+      )
```
ssrihari, line 72: `.exists?` to be explicit?

app/models/experimentation/notifications_experiment.rb
```diff
@@ -39,22 +41,45 @@ def self.eligible_patients
...
+
+      Notification.create!(
+        experiment: self,
+        message: template.message,
+        patient_id: membership.patient_id,
+        purpose: :experimental_appointment_reminder,
+        remind_on: date,
+        reminder_template: template,
+        status: "pending"
+      )
```
ssrihari, line 82: Is it not possible to use `find-or-create` here? https://guides.rubyonrails.org/active_record_querying.html#find-or-create-by

app/models/experimentation/notifications_experiment.rb
```diff
@@ -66,6 +91,14 @@ def cancel
 
     private
 
+    def remaining_enrollments_allowed(date)
+      MAX_PATIENTS_PER_DAY - enrolled_patients(date).count
+    end
+
+    def enrolled_patients(date)
+      treatment_group_memberships.where(experiment_inclusion_date: date)
```
ssrihari, line 99: This feels like an alias / delegation method, do we need it?

ssrihari: On further thought, I think this highlights that "membership" and "enrolment" might actually mean the same thing. And we might want to just stick with one word instead of two.

app/models/experimentation/reminder_template.rb
```diff
@@ -5,5 +5,13 @@ class ReminderTemplate < ActiveRecord::Base
 
     validates :message, presence: true
     validates :remind_on_in_days, presence: true, numericality: {only_integer: true}
+
+    validate :unique_message_per_group
+
+    def unique_message_per_group
+      if self.class.find_by(message: message, treatment_group: treatment_group)
+        errors.add(:message, "already exists in this treatment group")
+      end
```
ssrihari, line 14: Can we use a uniqueness validation with a scope instead? https://guides.rubyonrails.org/active_record_validations.html#uniqueness

app/models/experimentation/current_patient_experiment.rb
ssrihari: Could also consider `members_to_notify`.

app/models/experimentation/stale_patient_experiment.rb
```diff
@@ -41,9 +41,12 @@ def eligible_patients(date)
...
 
-    def memberships_to_notify(date)
-      # Patients who were enrolled on the `date`.
-      # To be implemented in a follow up PR.
+    # Patients who were enrollment date falls on
+    # one of the reminder template's remind_on days since `date`.
+    def memberships_to_notify(reminder_template, date)
+      treatment_group_memberships.status_enrolled.where(
+        experiment_inclusion_date: date - reminder_template.remind_on_in_days.days
+      )
```
ssrihari, line 49: Why are we passing in the reminder template here? An experiment method getting a reminder_template as an argument feels odd. The experiment already knows about its reminder templates.

In addition, it seems like we're getting _all_ members who are eligible to be notified the day of the given reminder template, even if they're not sent _that_ reminder template according to their treatment group?

Can we invert this to return the result with `reminder_template`s included instead?

ssrihari (changes requested):
Looks mostly good, added a couple of questions and notes.

app/models/experimentation/stale_patient_experiment.rb
prabhanshuguptagit: Addressed in 0c6a229d8. This seems much better.

app/models/experimentation/current_patient_experiment.rb
prabhanshuguptagit: I'm wary of calling it `patients_to_notify`, since this returns a `TreatmentGroupMembership` relation. The same for `members_to_notify`, we don't have the idea of a "member" anywhere. A "members" relation would ideally just be a `Patient` relation.

With the changes in 0c6a229d8, calling it `memberships` seems more appropriate, however it doesn't reflect that the result has a join with `reminders` 😅 . wdyt?

PR #3022, opened 2021-10-28
Make reporting_schema_v2 the default code path for Reports

rsanheim:
[ch5738](https://app.shortcut.com/simpledotorg/story/5738/switch-all-reports-data-access-to-v2)

This makes the v2 reporting code path in Repository the default code path for **all** repository calls.

Before this PR we had the main dashboard reports feature flagged, which is found in RegionsController. So those were already turned on everywhere in prod since the feature flag has been enabled everywhere for a bit.

We have lots of other code using the Repository, though, and those weren't using the feature flag. This consolidates the feature flag check to make sure any usage of the Repo will respect the feature flag, and also defaults it to true.  So all reports access will use the matview code paths.

This turned up a few things, most of which involved spec changes or some changes about how to check for the feature flag

* needed to update specs to make sure our test data worked correctly w/ v2 matviews
* needed to to adjust some callers of the repository so they passed in what periods they wanted
* fixed up some timezone behavior in specs
* one particular code path expected nil values for periods with 0 counts, but the repo returns the 0's to callers. For this case I'm probably going to change the expectations, because we really should return all the data and let consumers trim out 0 values if they really want to.


app/controllers/admin_controller.rb
```diff
@@ -37,16 +37,16 @@ def set_bust_cache
 
   private
 
+  # We only want to set this via query string, as it now defaults to enabled globally
   def set_reporting_schema_v2
-    RequestStore[:reporting_schema_v2] = reporting_schema_via_param_or_feature_flag
+    if params.key?(:v2)
+      param_flag = ActiveRecord::Type::Boolean.new.deserialize(params[:v2])
+      Reports.reporting_schema_v2 = param_flag
+    end
```
rsanheim, line 45: This can be simplified now because we know the value will be true everywhere, but we just want the ability to override to false for a little bit to confirm any discrepancies in prod.

app/models/reports.rb
```diff
@@ -1,10 +1,19 @@
...
+
   # The default period we report on is the current month.
   def self.default_period
     Period.month(Time.current.in_time_zone(Period::REPORTING_TIME_ZONE))
   end
 
-  REGISTRATION_BUFFER_IN_MONTHS = 3
-  MAX_MONTHS_OF_DATA = 24
-  PERCENTAGE_PRECISION = 0
+  def self.reporting_schema_v2?
```
rsanheim, line 11: Consolidating the flag here, as its global (per request) state and trying to minimize the pain from it.

app/services/monthly_district_data_service.rb
```diff
@@ -101,7 +101,7 @@ def facility_rows
   end
 
   def common_attributes(region, follow_ups_by_month)
-    complete_monthly_registrations = repo.complete_monthly_registrations.find { |k, _| k.slug == region.slug }.last
```
rsanheim, line 104: We no longer need to grab from `complete_monthly` numbers because we pass in the explicit range this service asks for.

spec/controllers/reports/regions_controller_spec.rb
```diff
@@ -14,13 +14,6 @@ def refresh_views
       RefreshReportingViews.call
     end
 
-    around do |example|
-      original = Reports::Repository.use_schema_v2?
-      Reports::Repository.use_schema_v2 = v2_flag
-      example.run
-      Reports::Repository.use_schema_v2 = original
-    end
```
rsanheim, line 22: Removing the dual path testing.

spec/models/reports/repository_spec.rb
```diff
@@ -1,623 +1,590 @@
 require "rails_helper"
 
-RSpec.describe Reports::Repository, type: :model, v2_flag: true do
```
rsanheim, line 3: This diff is only large because of whitespace - viewing with whitespace ignored will help.

spec/models/reports/repository_spec.rb
```diff
@@ -1,623 +1,590 @@
...
-      end
+  let(:organization) { create(:organization, name: "org-1") }
+  let(:user) { create(:admin, :manager, :with_access, resource: organization, organization: organization) }
+  let(:facility_group_1) { FactoryBot.create(:facility_group, name: "facility_group_1", organization: organization) }
 
-      around do |example|
-        original = Reports::Repository.use_schema_v2?
-        Reports::Repository.use_schema_v2 = v2_flag
-        with_reporting_time_zone { example.run }
-        Reports::Repository.use_schema_v2 = original
```
rsanheim, line 33: no more dual path testing

spec/models/reports/repository_spec.rb
```diff
@@ -1,623 +1,590 @@
...
+          create(:bp_with_encounter, :under_control, facility: facility_1, patient: patient, recorded_at: 4.days.from_now, user: user)
         end
+      end
+      refresh_views
+
+      RequestStore[:bust_cache] = true
+      repo = Reports::Repository.new(facility_1.region, periods: july_2020_range)
+      expect(repo.schema).to receive(:region_period_cached_query).with(:controlled_rates, with_ltfu: false).exactly(1).times
 
-        it "caches region_period entries only as far back as there is data" do
```
rsanheim, line 521: this test doesn't apply with v2 anymore

spec/queries/district_analytics_query_spec.rb
```diff
@@ -120,8 +82,11 @@
             total_registered_patients: 0
           }
         }
+        refresh_views
 
-        expect(analytics.call).to eq(expected_result)
+        with_reporting_time_zone do
```
rsanheim, line 87: We need to create the data in UTC (the default), and grab analytics in reporting tz...hence the addition of the block helper here.  Without this we will get month boundary issues.

spec/queries/drug_stocks_query_spec.rb
```diff
@@ -76,6 +79,7 @@
         for_end_of_month: for_end_of_month)
 
       other_facility_patients = create_list(:patient, 1, registration_facility: other_facility, registration_user: user)
+      refresh_views
```
rsanheim, line 82: DrugStock calls will now depend upon mat views, so adding the refresh views here.

spec/rails_helper.rb
```diff
@@ -47,6 +47,10 @@ def with_reporting_time_zone(&blk)
     end
   end
 
+  def refresh_views
```
rsanheim, line 50: making this easier to call across all specs

app/controllers/admin_controller.rb
```diff
@@ -37,16 +37,16 @@ def set_bust_cache
...
   end
 
   # A query string provided feature flag (v2=1) will always override any other settings here.
   def reporting_schema_via_param_or_feature_flag
-    param_flag = ActiveRecord::Type::Boolean.new.deserialize(params[:v2])
-    user_flag = current_admin.feature_enabled?(:reporting_schema_v2)
-    return param_flag unless param_flag.nil?
-    user_flag
```
ssrihari, line 49: Is this now an unused function?

spec/queries/drug_stocks_query_spec.rb
```diff
@@ -1,21 +1,21 @@
...
-  let!(:user) { create(:admin, :manager, :with_access, resource: facility_group) }
-  let!(:for_end_of_month) { Time.current.end_of_month }
-  let!(:drug_category) { "hypertension_ccb" }
-  let!(:stocks_by_rxnorm) {
+  let(:protocol) { create(:protocol, :with_tracked_drugs) }
+  let(:facility_group) { create(:facility_group, protocol: protocol) }
+  let(:user) { create(:admin, :manager, :with_access, resource: facility_group) }
+  let(:for_end_of_month) { Time.current.end_of_month }
+  let(:drug_category) { "hypertension_ccb" }
+  let(:stocks_by_rxnorm) {
```
ssrihari, line 11: Nice to see the `!`s go.

ssrihari (approved):
Neat set of updates! Can't wait to see the the versioning vanish altogether :)

PR #3028, opened 2021-11-02
A/B enhancements: Evict patients daily

prabhanshuguptagit:
**Story card:** [ch5389](https://app.shortcut.com/simpledotorg/story/5389/reject-ineligible-patients-from-experiments-on-an-ongoing-basis)

## Because

Evict patients from experiment (mark patient experiment result as “evicted”, record eviction date and reason) if:

- Multiple scheduled appointments exist
- New appointment was created after enrollment
- Expected return appointment is no longer status “scheduled”
- Expected return appointment remind_on is > expected return
- Previous messages failed to deliver

We should also cancel any pending notifications to these patients.

## This addresses

This adds the eviction criteria to the experiment model and runs them daily.


app/models/notification.rb
```diff
@@ -70,6 +70,10 @@ def localized_message
     end
   end
 
+  def self.cancel_pending_notifications
```
ssrihari, line 73: Is "pending" in the function name misleading since we're cancelling both pending and scheduled notifications?

app/models/experimentation/notifications_experiment.rb
```diff
@@ -79,9 +71,36 @@ def record_notification_results
...
+        .joins(patient: :latest_scheduled_appointments)
+        .where("treatment_group_memberships.appointment_id <> appointments.id")
+        .evict(reason: "new_appointment_created_after_enrollment")
+
+      treatment_group_memberships.status_enrolled
+        .joins(treatment_group: :reminder_templates)
+        .where("messages -> reminder_templates.message ->> 'result' = 'failed'")
+        .evict(reason: "notification_failed")
+
+      cancel_evicted_notifications
```
ssrihari, line 90: This looks neat!

app/models/experimentation/notifications_experiment.rb
```diff
@@ -79,9 +71,36 @@ def record_notification_results
...
+    end
+
+    def schedule_notifications(date)
+      memberships_to_notify(date)
+        .select("reminder_templates.id reminder_template_id")
+        .select("reminder_templates.message message, treatment_group_memberships.*")
+        .in_batches(of: BATCH_SIZE)
+        .each_record { |membership| schedule_notification(membership, date) }
```
ssrihari, line 98: Should this happen in a tx?

prabhanshuguptagit: Since this is idempotent, we don't need to put this in a transaction. We can re-run this in case of a failure.

app/models/notification.rb
prabhanshuguptagit: wdyt of
- `cancel_upcoming_notifications`
- `cancel_open_notifications`

ssrihari: How about just `cancel_notifications`? Assuming the `sent` ones won't be cancelled?

ssrihari (approved):

PR #3029, opened 2021-11-02
Cleanup old experiment runner

prabhanshuguptagit:
**Story card:** [ch5623](https://app.shortcut.com/simpledotorg/story/5623/refactor-runner-to-support-simplified-patient-flow)

## Because

Since `Experimentation::RunnerV2` is at parity with `Experimentation::RunnerV2`, we can cleanup the old runner now and make V2 mainline.

## This addresses

This removes the old experiment runner and makes v2 the new runner.


app/services/experimentation/runner.rb
```diff
@@ -1,101 +1,12 @@
 module Experimentation
   class Runner
-    PATIENTS_PER_DAY = 10_000
-    BATCH_SIZE = 100
-
-    def self.logger
```
rsanheim, line 6: Not sure of the context here, but not using the Notifications logger means you can't easily grep all logs w/ a module of `notifications` to follow all logs related to this sub-system. We found this pretty useful when debugging a/b testing initially.

rsanheim (approved):

app/services/experimentation/runner.rb
ssrihari: I suspect logging will go deeper in the code, and we might see this resurface elsewhere. Anyhow, I do think we should keep this in mind when we solve for observability for this round of enhancements.

ssrihari (approved):

app/services/experimentation/runner.rb
prabhanshuguptagit: This was helpful indeed 👍🏼, I will probably add a similar logger to the new runner in a follow up PR.

PR #3032, opened 2021-11-02
Ensure we get back all values in month_date order

rsanheim:
**Story card:** [ch5735](https://app.shortcut.com/simpledotorg/story/5735/registration-months-are-out-of-order-in-report-card)

I wasn't able to reproduce this in a test, which is why there are no corresponding specs to verify this behavior.  This may be some postgresql platform / version difference at work here.


rsanheim:
Verified on review apps - going to merge as this is a simple fix for a prod bug.

PR #3033, opened 2021-11-02
Track the total time for region caching 

rsanheim:
...including the patient breakdown cache as well

no story, just noticed this while browsing datadog and adding a graph to the reports dashboard.


ssrihari (approved):
lgtm

PR #3034, opened 2021-11-03
A/B enhancements: Mark patient visits daily

prabhanshuguptagit:
**Story card:** [ch5678](https://app.shortcut.com/simpledotorg/story/5678/mark-patient-visits-everyday-for-enrolled-patients)

## Because

We want to keep track of patients in the experiment visiting after being notified. For patients where new BP/sugar/meds were created after enrollment, record their visited date and mark patient experiment result as “followed up”. We also should cancel all pending notifications.

## This addresses

Adds the functionality to `NotificationExperiment`


app/models/experimentation/notifications_experiment.rb
```diff
@@ -90,6 +90,39 @@ def evict_patients
...
+          .where("recorded_at > ?", membership.experiment_inclusion_date)
+          .select(:id, :recorded_at, :device_created_at, :facility_id)
+          .min_by(&:recorded_at)
+        pd = PrescriptionDrug
+          .where(patient_id: membership.patient_id)
+          .where("device_created_at > ?", membership.experiment_inclusion_date)
+          .select(:id, :device_created_at, :facility_id)
+          .min_by(&:device_created_at)
+
+        earliest_visit = [bp, bs, pd].compact.min_by(&:device_created_at)
```
ssrihari, line 111: How about moving this to `membership#earliest_visit`?

ssrihari: Also, have we analysed performance? Is this good enough? We can consider running this `each_record` in parallel at the application level for an easy / simple performance boost.

prabhanshuguptagit: This isn't performant enough, unfortunately. It takes ~15ms per patient, which is good in isolation but since we'll be doing this for 100-200k patients everyday which is roughly 30-40 minutes. I've been trying to do this with SQL although that seems hard to get right. Will try parallelization if that doesn't work out 👍🏼 .

app/models/experimentation/notifications_experiment.rb
```diff
@@ -90,6 +90,25 @@ def evict_patients
...
+    def mark_visits
+      treatment_group_memberships.status_enrolled
+        .select("distinct on (treatment_group_memberships.patient_id) treatment_group_memberships.*, bp.id bp_id, bs.id bs_id, pd.id pd_id")
+        .joins("left outer join blood_pressures bp on bp.patient_id = treatment_group_memberships.patient_id AND bp.recorded_at > experiment_inclusion_date")
+        .joins("left outer join blood_sugars bs on bs.patient_id = treatment_group_memberships.patient_id AND bs.recorded_at > experiment_inclusion_date")
+        .joins("left outer join prescription_drugs pd on pd.patient_id = treatment_group_memberships.patient_id AND pd.device_created_at > experiment_inclusion_date")
+        .where("coalesce(bp.id, bs.id, pd.id) is not null")
+        .order("treatment_group_memberships.patient_id, bp.recorded_at, bs.recorded_at, pd.device_created_at")
+        .each do |membership|
+        membership.record_visit_details(
```
ssrihari, line 102: How about just `record_visit`?

app/models/experimentation/treatment_group_membership.rb
```diff
@@ -36,9 +36,40 @@ def record_notification_result(message, delivery_result)
...
+        status: :visited,
+        status_reason: :visit_recorded,
+        days_to_visit: (visit_date.to_date - experiment_inclusion_date.to_date).to_i
+      )
+    end
+
     private
 
+    def recorded_or_device_created_at(visit)
+      visit.recorded_at.presence || visit.device_created_at
```
ssrihari, line 67: I remember creating aliases for `recorded_at` in entities that didn't have it. Recheck if you really need to do this.

app/models/experimentation/treatment_group_membership.rb
```diff
@@ -36,9 +36,40 @@ def record_notification_result(message, delivery_result)
...
+    end
+
     def one_active_experiment_per_patient
+      return unless patient_id_changed? || treatment_group_id_changed? || experiment_id_changed?
```
ssrihari, line 71: 👏 

Although, `treatment_group_id` should _never change_.

ssrihari (approved):

app/models/experimentation/treatment_group_membership.rb
prabhanshuguptagit: Ah great https://github.com/simpledotorg/simple-server/blob/7d972f2550ad0ec9d4407d2e9d73ab4ccc273539/app/models/prescription_drug.rb#L24. Thanks 🙌🏼

PR #3039, opened 2021-11-08
A/B enhanchments: Bug fixes

prabhanshuguptagit:
**Story card:** https://app.shortcut.com/simpledotorg/epic/4929/a-b-testing-ihci-experiment-2

## Because

Found some bugs while testing.

## This addresses

- Don't considered discarded visit BP, BS, drugs for counting visits
- Fix a notification status_updated_at field
- Don't update notification result for queued notifications

### DIFF
--- a/app/models/experimentation/notifications_experiment.rb
+++ b/app/models/experimentation/notifications_experiment.rb
@@ -186,30 +186,25 @@ module Experimentation
     end
 
     def notification_result(notification_id)
-      communications = Communication.where(notification_id: notification_id)
-      successful_delivery =
-        communications.with_delivery_detail.select("delivery_detail.result, communications.*").find_by(
-          delivery_detail: {result: [:read, :delivered, :sent]}
-        )
-      queued_delivery = communications.with_delivery_detail.select("delivery_detail.result, communications.*").find_by(
-        delivery_detail: {result: [:queued]}
-      )
       notification = Notification.find(notification_id)
 
-      if successful_delivery.present?
-        {notification_status: notification.status,
-         notification_status_updated_at: notification.updated_at,
-         result: :success,
-         successful_communication_type: successful_delivery.communication_type,
-         successful_communication_created_at: successful_delivery.created_at.to_s,
-         successful_delivery_status: successful_delivery.result}
-      elsif communications.exists? && !queued_delivery.present?
-        {notification_status: notification.status,
-         notification_status_updated_at: notification.updated_at,
-         result: :failed}
-      else
-        {notification_status: notification.status,
-         notification_status_updated_at: notification.updated_at}
+      case notification.delivery_result
+        when :success
+          successful_delivery = notification.successful_deliveries.first
+
+          {notification_status: notification.status,
+           notification_status_updated_at: notification.updated_at,
+           result: :success,
+           successful_communication_type: successful_delivery.communication_type,
+           successful_communication_created_at: successful_delivery.created_at.to_s,
+           successful_delivery_status: successful_delivery.result}
+        when :failed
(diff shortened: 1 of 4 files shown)


app/models/experimentation/notifications_experiment.rb
```diff
@@ -95,10 +95,20 @@ def evict_patients
...
+          and bp.recorded_at > experiment_inclusion_date
+          and bp.deleted_at is null")
+        .joins("left outer join blood_sugars bs
+          on bs.patient_id = treatment_group_memberships.patient_id
+          and bs.recorded_at > experiment_inclusion_date
+          and bs.deleted_at is null")
+        .joins("left outer join prescription_drugs pd
+          on pd.patient_id = treatment_group_memberships.patient_id
+          and pd.device_created_at > experiment_inclusion_date
+          and pd.deleted_at is null")
```
ssrihari, line 111: Can we do this in a more railsy way? Wouldn't `has_many :blood_sugars, primary_key: :patient_id` get the discarded ones out by default?

ssrihari: Also, this needs a test.

app/models/experimentation/notifications_experiment.rb
```diff
@@ -190,13 +203,13 @@ def notification_result(notification_id)
          successful_communication_type: successful_delivery.communication_type,
          successful_communication_created_at: successful_delivery.created_at.to_s,
          successful_delivery_status: successful_delivery.result}
-      elsif communications.exists?
+      elsif communications.exists? && !queued_delivery.present?
```
ssrihari, line 206: Can we extract a `Notification#current_status` (or other name) method that returns one of `successful`, `pending`, or `failed`? We can do the experiment related things here, but the status querying seems experiment agnostic. 

Plus, we should test for this edge case.

app/models/experimentation/notifications_experiment.rb
```diff
@@ -190,13 +203,13 @@ def notification_result(notification_id)
...
         {notification_status: notification.status,
          notification_status_updated_at: notification.updated_at,
          result: :failed}
       else
         {notification_status: notification.status,
-         status_updated_at: notification.updated_at}
+         notification_status_updated_at: notification.updated_at}
```
ssrihari, line 212: Missing unit test?

ssrihari (changes requested):

app/models/experimentation/notifications_experiment.rb
prabhanshuguptagit: `has_many` doesn't help with the join conditions. There isn't a more railsy way to do this afaict :/

app/models/notification.rb
```diff
@@ -74,6 +74,28 @@ def self.cancel
...
+  def delivery_result
+    if successful_deliveries.exists?
+      :success
+    elsif queued_deliveries.exists? || !communications.exists?
+      :queued
+    else
+      :failed
+    end
+  end
+
```
ssrihari, line 98: super!

app/models/experimentation/notifications_experiment.rb
ssrihari: I tried the above snippet before posting it actually. And postgres (in this case at least) won't differentiate between the condition on a `join` and a `where`. It's not critical though, the sql is fairly readable still.

ssrihari (approved):

app/models/experimentation/notifications_experiment.rb
ssrihari: Could do an OR in the where no? 🤷

PR #3046, opened 2021-11-11
Fix stale patient selection query

prabhanshuguptagit:
**Story card:** -

## Because

The stale patient selection query has a bug where it considers any patient who visited between 35 to 365 days ago, instead of looking at the latest visit.

## This addresses

This fixes the bug. It changes the stale patient selection query to use `Reports::PatientVisit` and checks for the latest visit by looking at the current month's rows only.

It also adds an index on `visited_at` on `reporting_patient_visits`.


app/models/experimentation/stale_patient_experiment.rb
```diff
@@ -8,37 +8,18 @@ class StalePatientExperiment < NotificationsExperiment
...
+      last_visit_since = (date + PATIENT_VISITED_SINCE).beginning_of_day
+      last_visit_until = (date + PATIENT_VISITED_UNTIL).end_of_day
+      no_appointments_after = date.end_of_day
 
-      self.class.superclass.eligible_patients.where(id: sql.values)
+      self.class.superclass.eligible_patients
+        .joins("inner join reporting_patient_visits ON reporting_patient_visits.patient_id = patients.id")
+        .joins("left outer join appointments future_appointments on future_appointments.patient_id = patients.id
+                AND future_appointments.scheduled_date > '#{no_appointments_after}'")
+        .where(reporting_patient_visits: {month_date: current_month})
```
ssrihari, line 20: [nit] This would be more expressive on the join.

app/models/experimentation/stale_patient_experiment.rb
```diff
@@ -8,37 +8,18 @@ class StalePatientExperiment < NotificationsExperiment
...
+        .where("visited_at > ? AND visited_at < ?", last_visit_since, last_visit_until)
```
ssrihari, line 21: The `visited_at` in `patient_visits` also considers appointments, right? Is that okay?

app/models/experimentation/stale_patient_experiment.rb
```diff
@@ -8,37 +8,18 @@ class StalePatientExperiment < NotificationsExperiment
...
-      SQL
+      current_month = date.beginning_of_month
+      last_visit_since = (date + PATIENT_VISITED_SINCE).beginning_of_day
+      last_visit_until = (date + PATIENT_VISITED_UNTIL).end_of_day
+      no_appointments_after = date.end_of_day
 
-      self.class.superclass.eligible_patients.where(id: sql.values)
+      self.class.superclass.eligible_patients
+        .joins("inner join reporting_patient_visits ON reporting_patient_visits.patient_id = patients.id")
+        .joins("left outer join appointments future_appointments on future_appointments.patient_id = patients.id
```
ssrihari, line 18: [nit] Should we capitalise the SQL keywords for consistency/readability?

ssrihari (approved):
One question about `visited_at`, else looks good! 👌

app/models/experimentation/stale_patient_experiment.rb
prabhanshuguptagit: `.joins` doesn't have an option to bind strings like `where` can with `?`, which means you'd have to do a `#
'{current_month}'` :/ Hence the `where`.

app/models/experimentation/stale_patient_experiment.rb
prabhanshuguptagit: Yes, we want to consider appointments as visits while enrolling patients. We only skip appointments while tracking return visits.

PR #3049, opened 2021-11-12
A/B enhancements: Account for "Agreed to visit" stale patients 

prabhanshuguptagit:
**Story card:** [ch5856](https://app.shortcut.com/simpledotorg/story/5856/consider-the-agreed-to-visit-date-as-the-expected-return-date-for-stale-patients?vc_group_by=day&ct_workflow=all&cf_workflow=500000031)

## Because

We consider the `scheduled_date` of an appointment as the expected return date. For appointments which are marked as “Agreed to visit”, `scheduled_date` still stores the original date and we store the new date separately. Using `scheduled_date` in these cases is not accurate because they’ve agreed to visit on a future date. 

[Slack ref](https://simpledotorg.slack.com/archives/C01FMV58MEC/p1636519179026900)

## This addresses

Consider the “Agreed to visit” date as the expected return date if it is present
- Records the `remind_on` date as the expected return date if it is present
- Does not include any stale patients who have a `remind_on` in the future

### DIFF
--- a/app/models/experimentation/notifications_experiment.rb
+++ b/app/models/experimentation/notifications_experiment.rb
@@ -175,7 +175,7 @@ module Experimentation
         risk_level: patient.risk_priority,
         diagnosed_htn: medical_history.hypertension,
         experiment_inclusion_date: date,
-        expected_return_date: latest_scheduled_appointment&.scheduled_date,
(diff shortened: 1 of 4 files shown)


ssrihari (approved):
Looks good!

PR #3060, opened 2021-11-16
Introduce SentryHandler to report Sentry errors from `runner` tasks

rsanheim:
This is needed for areas where Sentry doesn't already install a hook via
its own middlewares...so any Rails controller or Sidekiq job is already
handled via Sentry's code.  But things that raise an error from a cron
scheduled `runner` task need to have their own rescue wired up.

This assumes that the method being called here is just `call`, which is
a good enough simplifying assumption: any tasks being kicked off from
cron should probably follow our service object conventions anyways.

Note: I'm still re-raising the exception here so that the task will fail and abort with a non-zero exit code.
Rescuing a top level error and continuing processing is often _not_ a great thing to do, and can be done manually with `rescue` if need be.

**Story card:** [ch5860](https://app.shortcut.com/simpledotorg/story/5860/it-looks-like-we-aren-t-reporting-exceptions-from-rails-runner-to-sentry)

### DIFF
--- a/app/services/runner_trace.rb
+++ b/app/services/runner_trace.rb
@@ -1,4 +1,5 @@
 class RunnerTrace
+  prepend SentryHandler
   class Error < StandardError
   end
 
@@ -15,8 +16,6 @@ class RunnerTrace
     logger.info msg: "about to raise an error",
                 sentry_debug_info: sentry_debug_info
     raise Error, "Runner trace error"
-  rescue => e
-    Sentry.capture_exception(e)
   end
 
   def sentry_debug_info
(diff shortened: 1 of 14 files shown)


app/services/runner_trace.rb
```diff
@@ -1,4 +1,5 @@
 class RunnerTrace
```
prabhanshuguptagit, line 1: Let's remove this `Runner` now?

prabhanshuguptagit (approved):

app/services/runner_trace.rb
rsanheim: Yup, I will in a follow up PR 👍 ....just want to make sure everything is working for a couple days before removing the tracer.

PR #3061, opened 2021-11-16
Update listen gem to 3.7.0

rsanheim:
Some good fixes here for guard / file watching

https://github.com/guard/listen/releases


rsanheim:
Dev tooling change only, going to merge away.

PR #3066, opened 2021-11-18
Fix Region Integrity runner

prabhanshuguptagit:
**Story card:** [chXXXX](URL)

## Because

Region Integrity check has been failing on all environments. This isn't being reported on sentry because it fails before the `SentryHelper` is loaded.

```
undefined method `call' for RegionsIntegrityCheck:Class
```

## This addresses

Fixes the Region Integrity check API.


rsanheim (approved):

rsanheim:
merging to get this running again

PR #3067, opened 2021-11-18
Handle Twilio 400s in Notification worker

prabhanshuguptagit:
**Story card:** [ch5885](https://app.shortcut.com/simpledotorg/story/5885/twilio-400s-are-not-recorded-leaves-some-pending-notifications-everyday)

## Because

Twilio returns a 400 for phone numbers that are invalid and it raises a `Twilio::REST::RestError`. There are many phone numbers in IHCI that are marked as `mobile` but are actually invalid. When a notification is sent to to such a number, the job fails and
- these notifications remain in the `pending` state
- The requests are retried by sidekiq many times before dying

[Example sentry error](
https://sentry.io/organizations/resolve-to-save-lives/issues/2586257564/events/68a8106a9d814accbe43bae27708feb0/?environment=production&project=1217715&query=is%3Aunresolved)

## This addresses

This marks the notifications as cancelled if Twilio returns a `400`. This also doesn't bubble the exception further and lets the job succeed.


app/jobs/appointment_notification/worker.rb
```diff
@@ -102,6 +61,42 @@ def send_message(notification, communication_type, recipient_number:)
...
+        raise UnknownCommunicationType, "#{self.class.name} is not configured to handle communication type #{communication_type}"
+      end
+    end
+  end
+
+  # This usually happens when the phone number is invalid.
+  # https://www.twilio.com/docs/errors/21211
+  def handle_twilio_error(notification, &block)
+    block.call
+  rescue TwilioApiService::Error => error
```
ssrihari, line 91: Would ideally like to handle this as data, not an exception, so that it's more first-class, not flow-by-exception, etc. But it's not important to address immediately. I would expect the gem to let us configure this behaviour, but I haven't looked into it deeply.

app/jobs/appointment_notification/worker.rb
```diff
@@ -102,6 +61,42 @@ def send_message(notification, communication_type, recipient_number:)
...
+    if error.reason == :invalid_phone_number
+      notification.status_cancelled!
```
ssrihari, line 93: 👌

app/jobs/appointment_notification/worker.rb
```diff
@@ -134,4 +129,35 @@ def medication_reminder_sms_sender
...
+    end
+
+    unless notification.patient.latest_mobile_number
+      logger.info "skipping notification #{notification.id}, patient #{patient.id} does not have a mobile number"
+      metrics.increment("skipped.no_mobile_number")
+      return
+    end
+
+    true
+  end
```
ssrihari, line 153: Nice method to extract 👍

app/services/twilio_api_service.rb
```diff
@@ -23,6 +23,14 @@ class TwilioApiService
   delegate :logger, to: Rails
 
   class Error < StandardError
+    STATUS_CODE_REASONS = {400 => :invalid_phone_number}
```
ssrihari, line 26: Is the sole reason we get 400s? Can we link to the documentation in a comment so that we can revisit this if it changes?

This shouldn't silently hide any client-errors we might hit in the future.

ssrihari (approved):
Some minor refactoring suggestions. Feel free to fix after merge.

app/jobs/appointment_notification/worker.rb
```diff
@@ -68,28 +42,13 @@ def send_message(notification, communication_type, recipient_number:)
...
+    response =
+      case communication_type
+        when "imo"
+          ImoApiService.new.send_notification(notification, recipient_number)
+        else
+          notify_via_twilio(notification, communication_type, recipient_number, context)
+      end
 
     metrics.increment("sent.#{communication_type}")
     return unless response
```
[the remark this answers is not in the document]
prabhanshuguptagit: This will avoid creating a communication and notification if there was an error.

app/jobs/appointment_notification/worker.rb
```diff
@@ -68,28 +42,13 @@ def send_message(notification, communication_type, recipient_number:)
...
-    end
+    response =
+      case communication_type
+        when "imo"
+          ImoApiService.new.send_notification(notification, recipient_number)
+        else
+          notify_via_twilio(notification, communication_type, recipient_number, context)
+      end
 
     metrics.increment("sent.#{communication_type}")
```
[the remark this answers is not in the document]
prabhanshuguptagit: Addressed in 89753ee9e

app/services/twilio_api_service.rb
```diff
@@ -23,6 +23,16 @@ class TwilioApiService
   delegate :logger, to: Rails
 
   class Error < StandardError
+    # https://www.twilio.com/docs/api/errors
+    ERROR_CODE_REASONS = {21211 => :invalid_phone_number,
+                          21614 => :invalid_phone_number}
+
```
prabhanshuguptagit, line 29: This handles individual twilio errors now instead of relying on the http response code.

PR #3068, opened 2021-11-18
A/B experiment: Improve notification result performance 

prabhanshuguptagit:
**Story card:** -

## This addresses

The fix shipped this morning in #3065 to skip discarded patients from recording notification results is inefficient. This makes a join (via the default discarded scope) instead of loading the full patient object per membership.


app/models/experimentation/notifications_experiment.rb
```diff
@@ -69,13 +71,11 @@ def record_notification_results
       # TODO: Look at query performance
       time(__method__) do
         treatment_group_memberships
-          .joins(treatment_group: :reminder_templates)
+          .joins(:patient, treatment_group: :reminder_templates)
```
prabhanshuguptagit, line 74: This is the main change. The rest is to support timing class methods.

rsanheim: looks reasonable -- wanna remove the TODO above?

app/models/experimentation/notifications_experiment.rb
```diff
@@ -171,6 +171,10 @@ def time(method_name, &block)
       Statsd.instance.flush # The metric is not sent to datadog until the buffer is full, hence we explicitly flush.
     end
 
+    def time(method_name, &block)
```
rsanheim, line 174: you can delegate from the instance to the class

```suggestion
delegate :time, to: self
```

rsanheim (approved):

app/models/experimentation/notifications_experiment.rb
ssrihari: What is the measured difference in performance?

ssrihari (dismissed):
Any performance improvement PR should be accompanied by the measured difference. Else, it's very difficult to evaluate the tradeoffs made in maintainability / readability.

app/models/experimentation/notifications_experiment.rb
prabhanshuguptagit: Thanks!

app/models/experimentation/notifications_experiment.rb
prabhanshuguptagit: For 200 patients, this takes 2.5s locally instead of 3.5s, because we aren't making those many extra queries. The difference will scale linearly with the number of patients.

I would like to revisit the overall performance of this query, this only improves the narrow fix I shipped yesterday morning.

PR #3069, opened 2021-11-18
Add a small dev benchmark

rsanheim:
This is a small bit of scaffold mostly cobbled together from https://gist.github.com/peterjmit/3864743 for doing some dev-related benchmarks of tests and DB seeding.  Once we have a web benchmark script that could also go in here, but for now this is mostly to compare M1 Macs vs Intel Macs and such.

(no shortcut story)


ssrihari:
> this is mostly to compare M1 Macs vs Intel Macs and such.

Curious.. why are we interested in this?

rsanheim:
> Curious.. why are we interested in this?

This is not directly related to any current project work, just something I'm curious about and have wanted to compare since the M1 chips came out.

PR #3072, opened 2021-11-21
Handle nil cohort data for user analytics

rsanheim:
**Story card:** [ch6069](https://app.shortcut.com/simpledotorg/story/6069/actionview-template-error-undefined-method-zero-for-nil-nilclass)
 
fixing an issue where we can get nil data after the v2 switch


rsanheim:
don't need this anymore

PR #3077, opened 2021-11-22
Move number_with_country_code into module

PriyangaPKini:
**Story card:** [sc-4389](https://app.shortcut.com/simpledotorg/story/4389/move-number-with-country-code-into-module)
## Because
The same process for localizing phone numbers is being used in both `PatientPhoneNumber#number_with_country_code` and `PhoneNumberAuthentication#localized_phone_number`. We should unify those into a module.
## This addresses
- Define a module PhoneNumberLocalizable with function `localized_phone_number`
- Place the module inside phone_number_localizable.rb file inside models/concerns folder
- Modify code to include the new module to use the function
- Use shared_examples to test the function for both the models
## Test instructions
- The tests corresponding to the function should be moved from model spec to helper spec. 
- All test cases should pass.


spec/helpers/patient_phone_number_helper_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+        phone_number = "1234567890"
+        expect(helper.number_with_country_code(phone_number)).to eq("+91234567890")
+      end
+
+      it "adds country code according to sms_country_code" do
+        Rails.application.config.country[:sms_country_code] = "+1"
+
+        phone_number = "+911234567890"
+        expect(helper.number_with_country_code(phone_number)).to eq("+11234567890")
```
rsanheim, line 57: nice, clear, explicit tests here 👍

app/jobs/request_otp_sms_job.rb
```diff
@@ -1,11 +1,13 @@
 class RequestOtpSmsJob < ApplicationJob
+  include PatientPhoneNumberHelper
```
rsanheim, line 2: Since we are already including this on the phone auth model, could we rely on that and not mix in a model specific behavior here into a job? I suppose you could do something like


```ruby
    TwilioApiService.new.send_sms(recipient_number: user.localized_phone_number), message: otp_message(user), context: context)
```

then

```ruby
class PatientPhoneNumber
  def localized_phone_number
    number_with_country_code(phone_number)
  end
...
```
    
Then the phone number auth is telling the job the correct number to use, and the job can concern itself just with the job handling.

PriyangaPKini: Okay. Will make these changes

app/models/patient_phone_number.rb
```diff
@@ -1,5 +1,8 @@
 class PatientPhoneNumber < ApplicationRecord
   include Mergeable
+  include PhoneNumberLocalization
+
+  alias_attribute :phone_number, :number
```
vkrmis, line 5: 👍

spec/helpers/patient_phone_number_helper_spec.rb
```diff
@@ -0,0 +1,61 @@
+require "rails_helper"
+
+RSpec.describe PatientPhoneNumberHelper, type: :helper do
```
vkrmis, line 3: Can move these specs to `spec/models/concerns/phone_number_localizable` as well, and test both `PatientPhoneNumber` and `PhoneNumberAuthentication` there.

app/controllers/concerns/phone_number_localization.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module PhoneNumberLocalization
+  extend ActiveSupport::Concern
+
+  included do
+    include PatientPhoneNumberHelper
+
+    def localized_phone_number
+      number_with_country_code(phone_number)
```
vkrmis, line 10: - Lets make this a model concern instead, maybe `PhoneNumberLocalizable` to go with the naming convention?
- Can move the body of `number_with_country_code` into `localized_phone_number`, and get rid of the `PatientPhoneNumberHelper` altogether.

vkrmis (changes requested):
Left a few comments, mostly just moving code around a bit.

app/controllers/concerns/phone_number_localization.rb
PriyangaPKini: okay

spec/helpers/patient_phone_number_helper_spec.rb
PriyangaPKini: okay

spec/models/patient_phone_number_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "rails_helper"
+require Rails.root.join "spec/models/concerns/phone_number_localizable_spec.rb"
```
ssrihari, line 2: This should not be necessary. If needed, consider including all `spec/models/concerns` in `rails_helper`.

app/models/patient.rb
```diff
@@ -3,6 +3,7 @@ class Patient < ApplicationRecord
   include Mergeable
   include Hashable
   include PatientReportable
+  include PhoneNumberLocalizable
```
ssrihari, line 6: This shouldn't be necessary, let's take another look at it.

vkrmis (approved):

PR #3080, opened 2021-11-23
Track client app version in datadog

vkrmis:
**Story card:** N/A

## Because
We want to start tracking the client app version so we can analyse traffic by app version, and push users to update.
## This addresses
Adds `X-APP-VERSION` to the list of headers we track via datadog

### DIFF
--- a/config/initializers/00_datadog.rb
+++ b/config/initializers/00_datadog.rb
@@ -13,7 +13,7 @@ Datadog.configure do |c|
   c.tracer.enabled = DATADOG_ENABLED
   c.profiling.enabled = ENABLE_DD_PROFILING
   c.version = SimpleServer.git_ref(short: true)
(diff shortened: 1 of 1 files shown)


rsanheim:
maybe link up to this story - https://app.shortcut.com/simpledotorg/story/4954/add-a-handshake-ability-between-mobile-clients-and-the-server ?

vkrmis:
> maybe link up as a first step towards this story - https://app.shortcut.com/simpledotorg/story/4954/add-a-handshake-ability-between-mobile-clients-and-the-server ?

Hmm, not sure we're doing much in that direction yet, this PR just pushes a header to datadog that the client already sends today. There's no handshake or ensuring minimum versions etc. The `push users to update` aspect involves manual intervention by CVHOs.

We're doing this because mixpanel replaces our user ids with mixpanel's own unique ids, making it hard to correlate exports from there with our users.

prabhanshuguptagit (approved):

PR #3081, opened 2021-11-23
Treatment inertia tables

ssrihari:
**Story card:** [ch3095](https://app.shortcut.com/simpledotorg/story/3095/implement-medications-dwh-tables)

## Because

- The names and dosages in the prescription drugs table are user-entered, and are not all clean. We need standardised data to report over them. The approach to clean the data, and the informational tables used for it were established in https://github.com/simpledotorg/simple-etl.
- We need monthly dosages of the list of medications we care about in a matview so that we can compute titration. The PR for this will follow.
- We need to refresh the prescriptions every day, with the rest of the matviews.

## This addresses

- Creation of the informational tables, loading them from CSVs
- Creation of the `reporting_prescriptions` matview

## Test instructions

Unit tests for this will be added after a first round of exploration and iteration on Metabase.


db/migrate/20211123060535_create_treatment_inertia_tables.rb
```diff
@@ -0,0 +1,35 @@
+class CreateTreatmentInertiaTables < ActiveRecord::Migration[5.2]
+  def up
+    create_table :clean_medicine_to_dosage, id: false do |t|
```
prabhanshuguptagit, line 3: Just `clean_medicines`/ `clean_medications`?

db/migrate/20211123060535_create_treatment_inertia_tables.rb
```diff
@@ -0,0 +1,35 @@
...
+      t.bigint :rxcui, null: false
+      t.string :medicine, null: false
+      t.float :dosage, null: false
+    end
+    execute "COPY clean_medicine_to_dosage FROM '#{Rails.root}/config/data/treatment-inertia/clean_medicine_to_dosage.csv' WITH DELIMITER ',' CSV HEADER"
+
+    create_table :raw_to_clean_medicine, id: false do |t|
```
prabhanshuguptagit, line 10: This name needs pluralisation because rails. Others also.

db/migrate/20211123060535_create_treatment_inertia_tables.rb
```diff
@@ -0,0 +1,35 @@
...
+      t.string :raw_name, null: false
+      t.string :raw_dosage, null: false
+      t.bigint :rxcui, null: false
+    end
+    execute "COPY raw_to_clean_medicine FROM '#{Rails.root}/config/data/treatment-inertia/raw_to_clean_medicine.csv' WITH DELIMITER ',' CSV HEADER"
+
+    create_table :medicine_purpose, id: false do |t|
```
prabhanshuguptagit, line 17: Should we use the rxnorm codes to map medicines with purposes?

config/data/treatment-inertia/clean_medicine_to_dosage.csv
```diff
@@ -0,0 +1,50 @@
+rxcui,medicine,dosage
```
prabhanshuguptagit, line 1: Could consider sourcing these medication names and rxnorm codes from `medications.yml`. It'll reduce duplication.

db/migrate/20211123060535_create_treatment_inertia_tables.rb
```diff
@@ -0,0 +1,35 @@
...
+      t.string :name, null: false
+      t.boolean :hypertension, null: false
+      t.boolean :diabetes, null: false
+    end
+    execute "COPY medicine_purpose FROM '#{Rails.root}/config/data/treatment-inertia/medicine_purpose.csv' WITH DELIMITER ',' CSV HEADER"
```
prabhanshuguptagit, line 22: The `COPY`s can be split into a data migration.

db/views/reporting_prescriptions_v01.sql
```diff
@@ -0,0 +1,43 @@
...
+    SELECT actual.name          as actual_name,
+           actual.dosage        as actual_dosage,
+           clean.medicine       as clean_name,
+           clean.dosage         as clean_dosage,
+           purpose.hypertension as medicine_purpose_hypertension,
+           purpose.diabetes     as medicine_purpose_diabetes
+    FROM prescription_drugs actual
+             LEFT JOIN raw_to_clean_medicine raw
+                       ON lower(regexp_replace(raw.raw_name, '\s+', '', 'g')) =
+                          lower(regexp_replace(actual.name, '\s+', '', 'g'))
```
prabhanshuguptagit, line 31: Could add a regex_replaced/slugified version in the tables while importing data to simplify the query.

prabhanshuguptagit (approved):
A few suggestions, approving to unblock.

db/migrate/20211123060535_create_treatment_inertia_tables.rb
ssrihari: I'll reconsider all names after the first pass. I've mostly gone with the names that were used in the spike.

db/migrate/20211123060535_create_treatment_inertia_tables.rb
ssrihari: As of now, all these lookup tables will only be used in the matview creation, and I don't see any need to use this from Rails. However, I'll consider this in the next pass while renaming tables. Might be a small thing to address.

db/migrate/20211123060535_create_treatment_inertia_tables.rb
ssrihari: Yeah, good idea. I also need to add the right kind of unique indexes on these. Will do this in the next pass. Thanks for this suggestion.

db/views/reporting_prescriptions_v01.sql
ssrihari: Right now the import is a plain `COPY`. And actually, since these regex/lower will remain in the query for the actual prescription drugs rows, it might be better to keep it in the same place for readability and maintainability.

db/migrate/20211123060535_create_treatment_inertia_tables.rb
ssrihari: Done in ebc10d430924d013f63a21b9e15531017cbe4627.

config/data/treatment-inertia/clean_medicine_to_dosage.csv
ssrihari: There's [a story](https://app.shortcut.com/simpledotorg/story/2886/integrate-etl-pipeline-with-existing-drug-name-dosage-cleanup-work) for this, and I'll look into it. Not all the drugs in `medications.yml` have dosages, and they aren't floats. Needs some additional thought and work. We might want to convert `medications.yml` into a table and use that for both purposes, perhaps? Not sure.

db/migrate/20211123060535_create_treatment_inertia_tables.rb
ssrihari: Had a chance to reconsider this with https://github.com/simpledotorg/simple-server/pull/3091. I went with `clean_medicine_to_dosages` since we are also cleaning up dosages here and converting them to floats, and it's easy to miss that if we name it as `clean_medicines`.

db/migrate/20211123060535_create_treatment_inertia_tables.rb
ssrihari: Pluralised with https://github.com/simpledotorg/simple-server/pull/3085.

db/migrate/20211123060535_create_treatment_inertia_tables.rb
ssrihari: This table has one row per medicine name. So, Amlo 5, 10, 20, etc are still just the one row. Which is nice, since this table is going to be maintained manually in the foreseeable future. I've added unique indexes on other tables with https://github.com/simpledotorg/simple-server/pull/3091.

db/views/reporting_prescriptions_v01.sql
ssrihari: Moved to a csv import with https://github.com/simpledotorg/simple-server/pull/3085. But as I said earlier, it makes sense to keep the `regexp_replace` in both LHS and RHS of the query for maintainability.

db/migrate/20211123060535_create_treatment_inertia_tables.rb
prabhanshuguptagit: `clean_medicine_dosages`

PR #3086, opened 2021-11-24
A/B: Record visits for evicted patients

prabhanshuguptagit:
**Story card:** [ch6119](https://app.shortcut.com/simpledotorg/story/6119/record-visits-for-evicted-patients-too)

## Because
We want to track visits for evicted patients too, in case we need the information for analysis in the future. The status should still be `evicted`.

## This addresses

This considers evicted patients in addition to enrolled patients for tracking visits.


app/models/experimentation/treatment_group_membership.rb
```diff
@@ -58,15 +58,23 @@ def record_visit(blood_pressure:, blood_sugar:, prescription_drug:)
...
-        days_to_visit: days_to_visit
+        days_to_visit: days_to_visit,
+        **visit_status_fields
       )
     end
 
     private
 
+    def visit_status_fields
+      if status_evicted?
```
ssrihari, line 69: would a guard clause read better?

spec/models/experimentation/notifications_experiment_spec.rb
```diff
@@ -478,7 +478,34 @@
       expect(membership.visit_prescription_drug_created).to eq(true)
     end
 
-    it "cancels all pending notifications for evicted patients" do
+    it "considers earliest BP created as a visit for evicted patients and does not change status" do
```
ssrihari, line 481: Not just BP, right?

ssrihari (approved):

PR #3092, opened 2021-11-25
Don't use imo in Bangladesh

prabhanshuguptagit:
**Story card:** [ch6145](https://app.shortcut.com/simpledotorg/story/6145/don-t-use-imo-or-whatsapp-in-bangladesh)

## Because

We don't want to use imo in Bangladesh for the A/B experiment given the low acceptance rate.

## This addresses

This removes `imo` as a preferred communication type. This doesn't cleanup any IMO related code, which will be done separately in [this epic](https://app.shortcut.com/simpledotorg/epic/5928/remove-imo-code-from-simple).

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -113,7 +113,6 @@ class Notification < ApplicationRecord
 
   def preferred_communication_method
     return "whatsapp" if Flipper.enabled?(:whatsapp_appointment_reminders)
-    return "imo" if Flipper.enabled?(:imo_messaging) && patient.imo_authorization&.status_subscribed?
     nil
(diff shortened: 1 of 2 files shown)


ssrihari:
What's the benefit to removing this code as opposed to just turning off `imo_messaging` on Flipper?

ssrihari (approved):
Approving to unblock, but not sure why this is needed.

prabhanshuguptagit:
> What's the benefit to removing this code as opposed to just turning off imo_messaging on Flipper?

Just to be double sure really, it's not necessary 😅

PR #3093, opened 2021-11-25
A/B: Exclude patients in Bangladesh evaluation study sites

prabhanshuguptagit:
**Story card:** [ch5604](https://app.shortcut.com/simpledotorg/story/5604/exclude-patients-in-bangladesh-evaluation-study-sites)

## Because
We don't want to include patients in the A/B experiment who are part of a different study.

## This addresses

Adds a clause to `NotificationExperiment.eligible_patients` to exclude these patients


app/models/experimentation/notifications_experiment.rb
```diff
@@ -2,6 +2,7 @@ module Experimentation
   class NotificationsExperiment < Experiment
     include ActiveSupport::Benchmarkable
     BATCH_SIZE = 1000
+    EXCLUDED_BLOCKS = YAML.load_file("config/data/blocks_excluded_from_experiments.yml")
```
rsanheim, line 5: We have country specific env var files now, see https://github.com/simpledotorg/simple-server/blob/master/.env.IN.  

You could add the ENV var there, and then load it from there to keep our configuration more consistent.

app/models/experimentation/notifications_experiment.rb
```diff
@@ -47,6 +50,7 @@ def self.eligible_patients
                                              .where(status: :scheduled)
                                              .group(:patient_id)
                                              .having("count(patient_id) > 1"))
+        .where.not(block_region: {id: EXCLUDED_BLOCKS[CountryConfig.current[:abbreviation]]})
```
rsanheim, line 53: Maybe make this conditional on having excluded blocks for the current deployment?

I'm thinking if we ever run experiments in a country without excluded blocks, I think it will look for block_regions where the id is not null.  That won't change the query results I think, but it does make the query more complex ( and slower).

rsanheim (commented):
feedback on the config and keeping it more consistent w/ our ENV vars

spec/fixtures/files/blocks_excluded_from_experiments.yml
```diff
@@ -0,0 +1,2 @@
+IN:
+  - d6877901-1f73-4e24-bebc-4c837782dea1
```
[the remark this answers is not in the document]
prabhanshuguptagit: This was just a spec fixture, removed this since we're using `.env` now

PR #3096, opened 2021-11-30
Configurable rate limit and add ability for plain text config per env

rsanheim:
A quick attempt to see if we can configure rate limit vars these in plain text .env files, because there are lots of env things that don't require encryption and shouldn't require an ansible run to change.


rsanheim:
Not going to tackle this right now, its a bigger change.

PR #3105, opened 2021-12-02
Update dashboard overdue page to work with call_results

vkrmis:
**Story card:** [ch4618](https://app.shortcut.com/simpledotorg/story/4618/update-overdue-page-on-dashboard-to-work-with-call-results), [ch3885](https://app.shortcut.com/simpledotorg/story/3885/match-language-copy-in-dashboard-result-of-follow-up-options-with-simple-app-options)

## Because

- The existing dashboard overdue page does not create call_results
- The dropdown options have drifted over time between the dashboard and app

## This addresses

- The dashboard overdue page will now create call_results when an appointment's state is updated
- The options for `Result of follow up` have been updated to be in line with the app
Screenshot 2021-12-02 at 1 15 20 PM


app/controllers/appointments_controller.rb
```diff
@@ -33,7 +33,7 @@ def update
     call_result = appointment_params[:call_result].to_sym
     search_filters = appointment_params[:search_filters]&.split || []
 
-    if set_appointment_status_from_call_result(@appointment, call_result)
+    if set_appointment_status_and_call_result(@appointment, call_result)
```
vkrmis, line 36: slight rename, because we're also creating a `call_result` record now

app/models/appointment.rb
```diff
@@ -23,7 +23,7 @@ class Appointment < ApplicationRecord
     visited: "visited"
   }, _prefix: true
 
-  enum cancel_reason: CallResult.remove_reasons
+  enum cancel_reason: CallResult.remove_reasons.except(:already_visited)
```
vkrmis, line 26: This is a new `remove_reason` that wasn't previously supported. The `cancel_reason` enum used to be defined inline here, but we moved it to `CallResult.remove_reasons` in a [previous PR](https://github.com/simpledotorg/simple-server/pull/2861/files)

app/views/appointments/index.html.erb
```diff
@@ -3,10 +3,18 @@
...
+    'Remind to call later' => [['Remind to call later', 'remind_to_call_later']],
+    'Remove from overdue list' => [['Patient has already visited', 'already_visited'],
+                                   ['Patient is not responding', 'not_responding'],
+                                   ['Phone number does not work', 'invalid_phone_number'],
+                                   ['Transferred to another public health facility', 'public_hospital_transfer'],
+                                   ['Moved to private practitioner', 'moved_to_private'],
+                                   ['Moved', 'moved'],
+                                   ['Died', 'dead'],
+                                   ['Other reason','other']]
+  }
```
vkrmis, line 16: Refactored this to match the options that nurses see in the app

PR #3110, opened 2021-12-06
Return remaining account lockout duration in seconds in `users/v3/login` API

PriyangaPKini:
**Story card:** [sc-5042](https://app.shortcut.com/simpledotorg/story/5042/return-remaining-account-lockout-duration-in-user-authenticate-api)

## Because

When a user tries to authenticate with an invalid OTP, the server locks them out for a certain period. However, the current API doesn't return that duration, requiring the Android app to handle the lockout locally. This also creates the possibility of the app saying it's open, but the server still has the lockout enabled.

Instead, let's return the remaining lockout duration in `api/v3/login`, something like this:
```json
{
  "errors": {
    "user": 
        [ "string" ]
    },
    "lockout_duration": 1234  # number of seconds remaining
}
```
## This addresses

- Define a new function `seconds_left_on_lockout` to get remaining account lockout duration in seconds
- Add `remaining_lockout_duration_in_seconds` to the error response when a user tries to authenticate with an invalid OTP for >= `USER_AUTH_MAX_FAILED_ATTEMPTS` number of times

## Test instructions
- Add test for the function`seconds_left_on_lockout`
- Add test for failed authentication when a user tries to authenticate with invalid OTP for >= `USER_AUTH_MAX_FAILED_ATTEMPTS` number of times
- All tests should pass


app/controllers/api/v3/logins_controller.rb
```diff
@@ -21,17 +21,25 @@ def login_user
...
-        }
-      }
-      render json: response, status: :unauthorized
+      render json: get_error_response(result), status: :unauthorized
     end
   end
 
   private
 
+  def get_error_response(result)
```
ssrihari, line 30: I would name this `error_response`, the `get` is superfluous.

ssrihari (approved):

app/controllers/api/v3/logins_controller.rb
PriyangaPKini: okay

PR #3112, opened 2021-12-07
Allow grouping follow ups by gender

rsanheim:
**Story card:** [ch6242](https://app.shortcut.com/simpledotorg/story/6242/support-gender-breakdown-per-facility-month-in-reporting-pipeline)

## Because

We should be able to count follow ups by gender / facility in our reporting pipeline. We will need this for the prgoress tab as well as for some DHIS2 exports.

## This addresses

Adds a gender enum to postgresql and then uses that in our follow ups matview.
Also changes our DB schema to SQL -- I know we've discussed this before, and largely agreed we should do it given how detailed our SQL schema is getting. Not only will this let use PSQL functions, check constraints, and enums, I'm also hopeful it will make it easier to keep our schema consistent between different dev machines.


lib/tasks/db.rake
```diff
@@ -28,5 +28,6 @@ namespace :db do
 end
 
 Rake::Task["db:seed"].enhance do
+  ENV["REFRESH_MATVIEWS_CONCURRENTLY"] = "false"
```
rsanheim, line 31: Need this to properly refresh matview views when seeding from an empty state.

spec/controllers/reports/regions_controller_spec.rb
```diff
@@ -9,10 +9,6 @@
   let(:facility_group_1) { FactoryBot.create(:facility_group, name: "facility_group_1", organization: organization) }
   let(:facility_1) { FactoryBot.create(:facility, name: "facility_1", facility_group: facility_group_1) }
 
-  def refresh_views
-    RefreshReportingViews.call
-  end
-
```
rsanheim, line 15: Duplicates a shared test helper.

app/models/reports/repository.rb
```diff
@@ -90,10 +90,16 @@ def warm_cache
     end
 
     def follow_ups_v2_query(group_by: nil)
+      group_field = case group_by
+        when /user_id/ then :user_id
+        when /gender/ then :patient_gender
```
rsanheim, line 95: This allows for grouping by gender now.

app/models/reports/patient_follow_up.rb
```diff
@@ -4,6 +4,11 @@ class PatientFollowUp < Reports::View
     belongs_to :patient
     belongs_to :facility, class_name: "::Facility"
     belongs_to :user
+    enum patient_gender: {
```
timcheadle, line 7: Should this use `Rails.application.config.country[:supported_genders]` like it does in https://github.com/simpledotorg/simple-server/blob/master/app/models/patient.rb?

app/models/reports/repository.rb
timcheadle: I dig this. These regex _could_ use a `^` and `$` but that's nitpicky :)

config/application.rb
```diff
@@ -39,6 +39,7 @@ class Application < Rails::Application
       g.helper false
     end
 
+    config.active_record.schema_format = :sql
```
timcheadle, line 42: Any downsides to this change? Seems like a big deal?

app/models/reports/repository.rb
rsanheim: I actually wanted it to be a loose regex, because some clients will group_by `blood_pressures.user_id` and others may ask for `user_id` for example.  I can add the end of string token tho.

app/models/reports/patient_follow_up.rb
rsanheim: I wanted to keep it simple and have the same enum across everywhere here in the reporting pipeline.  All the UI side gender pieces will still validate and enforce the `supported_genders` tho.

config/application.rb
rsanheim: So far in my testing it has seemed solid and better actually.  I ran through some db resets to verify things, and reran migrations as well.  With the schema set to SQL I get none of the schema diff thrashing we saw with it set to ruby.

timcheadle (approved):
Excited for this!

PR #3116, opened 2021-12-07
Script to duplicate metabase questions across envs

prabhanshuguptagit:
**Story card:** [ch6235](https://app.shortcut.com/simpledotorg/story/6235/copy-metabase-reports-and-slack-alerts-from-ihci)

## Because

We need to duplicate the same Metabase questions across IHCI and BD production. This is tiresome when there are many questions to be copied.

## This addresses

Adds a task to help duplicating questions across envs. It borrows from https://github.com/itmi-id/metabase-migration but re-written in ruby so we can add more things on top as and when we need them.


lib/tasks/scripts/metabase.rb
```diff
@@ -0,0 +1,106 @@
...
+
+require "net/http"
+
+class Metabase
+  def initialize(source_host = ENV["METABASE_SOURCE_HOST"],
+    source_username = ENV["METABASE_SOURCE_USERNAME"],
+    source_password = ENV["METABASE_SOURCE_PASSWORD"],
+    destination_host = ENV["METABASE_DESTINATION_HOST"],
+    destination_username = ENV["METABASE_DESTINATION_USERNAME"],
+    destination_password = ENV["METABASE_DESTINATION_PASSWORD"])
```
rsanheim, line 34: can you `fetch` on these so it will fail fast if someone misses one of these?

rsanheim (approved):
one small thing about the config, otherwise LGTM

PR #3121, opened 2021-12-08
Remove unused matviews from Reporting Pipeline (models only)

rsanheim:
**Story card:** [ch6239](https://app.shortcut.com/simpledotorg/story/6239/remove-old-reporting-code)

## Because

We want to clean up and remove old reporting code that is no longer used. This will shorten the matview refresh time and also make things easier to find going forward, as we won't have so many old queries / matviews to search through.

NOTE: this just removes the models and refresh code -- I'll remove the actual matviews in a follow up PR for safety.

## This addresses

Removing the models, queries, specs, and refresh calls for

* PatientRegistrationsPerDayPerFacility
* LatestBloodPressuresPerPatientPerQuarter

Also removes an old `RegistrationsQuery` that was using `PatientRegistrationsPerDayPerFacility`

## Test instructions

pull down the PR, run through reports and verify specs are all green.


app/services/refresh_reporting_views.rb
```diff
@@ -42,12 +42,11 @@ def self.tz
     Rails.application.config.country[:time_zone]
   end
 
+  # Keep the views below in the order as defined, as they have dependencies on earlier matviews in the lists
   V1_REPORTING_VIEWS = %w[
     LatestBloodPressuresPerPatientPerMonth
     LatestBloodPressuresPerPatient
```
rsanheim, line 48: The two views above are still used to build BloodPressuresPerFacilityPerDay, which is then used in my facilities. So we can't remove these until we have the right views for BPs per facility per day in the v2 pipeline.

claudiovallejo (approved):

PR #3123, opened 2021-12-09
Add additional drugs for lookup, pick latest dosage for a medicine

ssrihari:
**Story card:** [ch6271](https://app.shortcut.com/simpledotorg/story/6271/improve-drug-lookups-to-cover-more-htn-and-diabetes-drugs)

## Because

- We need to cover more drugs prescribed to patients, especially HTN ones.
- Duplicate drugs are created on the mobile app sometimes, we should pick the latest dosage for a medicine

## This addresses

- Deduplicates medicines by name, fetching the latest one
- Adds drugs to the lookup tables


vkrmis (approved):
LGTM

PR #3126, opened 2021-12-09
Show overdue call results

vkrmis:
**Story card:** [ch6272](https://app.shortcut.com/simpledotorg/story/6272/show-call-results)

## Because

We want to start tracking how frequently users are calling overdue patients

## This addresses

Shows call results in the region details pages


app/controllers/reports/regions_controller.rb
```diff
@@ -78,9 +78,18 @@ def details
...
     chart_repo = Reports::Repository.new(@region, periods: chart_range, follow_ups_v2: current_admin.feature_enabled?(:follow_ups_v2))
 
+    district_regions = if @region.state_region?
+      [@region, @region.district_regions].flatten
+    end
+
+    if @region.state_region?
+      @district_repository = Reports::Repository.new(district_regions, periods: @period_range, follow_ups_v2: current_admin.feature_enabled?(:follow_ups_v2))
+    end
+
```
vkrmis, line 92: This is to show call results grouped by district in the state report, this was requested explicitly.

PR #3125, opened 2021-12-09
Medication titration improvements fix

ssrihari:
https://github.com/simpledotorg/simple-server/pull/3123 had a data issue that failed deployment, this fixes it.

**Story card:** [ch6271](https://app.shortcut.com/simpledotorg/story/6271/improve-drug-lookups-to-cover-more-htn-and-diabetes-drugs)

## Because

- We need to cover more drugs prescribed to patients, especially HTN ones.
- Duplicate drugs are created on the mobile app sometimes, we should pick the latest dosage for a medicine

## This addresses

- Deduplicates medicines by name, fetching the latest one
- Adds drugs to the lookup tables

PR #3127, opened 2021-12-09
Fix review apps: Load from structure, not schema; update schema w/ latest rx matview changes

rsanheim:
Changing our heroku setup to load from the structure.sql instead of ruby schema.

Also includes a small fix to update structure.sql from https://github.com/simpledotorg/simple-server/pull/3125.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ rails db:setup
 ```
 
 We cleanup old migration files every once in a while and so running `db:migrate` would not work for the initial setup.
-While setting up a fresh DB you will need to load the schema first with `db:structure:load`. `db:setup` already takes care of this.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -103,7 +103,7 @@ rails db:setup
 ```
 
 We cleanup old migration files every once in a while and so running `db:migrate` would not work for the initial setup.
-While setting up a fresh DB you will need to load the schema first with `db:schema:load`. `db:setup` already takes care of this.
+When setting up a new database, `db:setup` wil take care of everything (it runs `db:structure:load` under the hood).
```
claudiovallejo, line 106: Typo: "... `db:setup` *will* take care..."

rsanheim: fixed!

claudiovallejo (approved):
Looks good @rsanheim! Left a comment for a typo in your `README.md` change

PR #3145, opened 2021-12-10
Fix heroku review app deployment

vkrmis:
**Story card:** [ch6279](https://app.shortcut.com/simpledotorg/story/6279/switching-to-structure-sql-broke-our-heroku-review-app-deployments)

Switching from `schema.db` to `structure.sql` is breaking heroku review app deployment.


db/structure.sql
```diff
@@ -9,48 +9,20 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
---
--- Name: plpgsql; Type: EXTENSION; Schema: -; Owner: -
---
-
-CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
```
prabhanshuguptagit, line 16: This is extra?

prabhanshuguptagit (approved):

db/structure.sql
vkrmis: I think its disallowed, I'll double check

vkrmis: It actually works fine, re-including it

Rakefile
```diff
@@ -7,3 +7,8 @@ Rails.application.load_tasks
 def is_running_migration?
   Rake.application.top_level_tasks.include?("db:migrate")
 end
+
+Rake::Task["db:schema:load"].clear
+task "db:schema:load" do
+  Rake::Task["db:structure:load"].invoke
```
rsanheim, line 13: Why do we need to redefine this task if we manually call the structure load in `heroku-setup` ?

rsanheim (commented):
thx for the fix - I guess we have to find a way to keep this in sync after the schema gets dumped ?

PR #3159, opened 2021-12-13
Remove sidekiq installation via capistrano

prabhanshuguptagit:
**Story card:** -

Accompanying PR to https://github.com/simpledotorg/deployment/pull/387. Moves sidekiq installation to the deployment repo.


rsanheim (approved):
Makes sense.

Are you going to run the sidekiq recipe via ansible for all machines? It would be good to verify those all work 100% now as we are switching this over.

prabhanshuguptagit:
> It would be good to verify those all work 100% now as we are switching this over.

Tested on india staging yesterday. It works since it's the same files/service just copied via ansible.

PR #3161, opened 2021-12-13
first commit, create my_facilities_download job and mailer and views

solracdelsol:
**Story card:** [chXXXX](https://app.shortcut.com/simpledotorg/epic/6307?vc_group_by=day&cf_workflow=500000031&ct_workflow=all)

## Because

Given potential bandwidth issues of a given region, we would prefer a my_facility_csv download to be performed asynchronously via a job scheduler, so the app user doesn't have to wait for a potentially slow download.

## This addresses

Download speeds regarding high-volume areas with low bandwidth.

## Test instructions

Enter detailed instructions for how to test this PR... (not yet)


app/jobs/my_facilities_download_job.rb
```diff
@@ -0,0 +1,8 @@
+class MyFacilitiesDownloadJob < ApplicationJob
```
rsanheim, line 1: Can you match the existing sidekiq jobs for this?

So just a normal class (no ApplicationJob)  `  include Sidekiq::Worker`

app/controllers/my_facilities_controller.rb
```diff
@@ -57,6 +57,10 @@ def csv_maker
     send_data csv_data, type: "text/csv", filename: filename
```
rsanheim, line 57: Hey I think here is where you will want to queue the download job. Something like `MyFacilitiesDownloadJob.perform_async`

PR #3176, opened 2021-12-15
Remove extension commands from structure sql

rsanheim:
**Story card:** [chXXXX](URL)

## Because

## This addresses

## Test instructions


rsanheim:
This removes too much, we need the extensions themselves, but don't want to the COMMENT ON pieces.

PR #3179, opened 2021-12-15
Remove COMMENT ON EXTENSION bits from structure.sql

rsanheim:
**Story card:** [ch6333](https://app.shortcut.com/simpledotorg/story/6333/fix-structure-sql-to-run-smoothly-on-heroku-other-envs-without-needed-superuser-privileges)

This manually strips out the `COMMENT ON EXTENSION` bits from our structure.sql that break on Heroku.

Note that we can do this in a cleaner way once we are on PostgreSQL 11, see the story for details.

Looks like it is also fixed in mainline Rails in version 7 - https://github.com/rails/rails/pull/43216

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -10,6 +10,20 @@ SET client_min_messages = warning;
 SET row_security = off;
 
 --
+-- Name: plpgsql; Type: EXTENSION; Schema: -; Owner: -
+--
+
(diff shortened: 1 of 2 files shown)


rsanheim:
Verified w/ some DB resets locally, this looks good.  Going to merge to unblock Heroku review apps so we stop hitting this.

PR #3166, opened 2021-12-14
Reporting facility follow ups

rsanheim:
**Story card:** [ch4212](https://app.shortcut.com/simpledotorg/story/4212/support-follow-ups-in-the-reporting-pipeline)

* adds total follow ups to facility states matview
* adds gender & diagnosis to follow ups matview
* adds a progress tab controller to dashboard (I'll lock this down to power users only or something similiar, this is primarily for dev)
* tests, making our fixture data better to support DM / HTN follow ups

I'll also add inline comments as I go for specifics.

PR #3188, opened 2021-12-16
Add follow ups to facility states

rsanheim:
**Story card:** [ch4212](https://app.shortcut.com/simpledotorg/story/4212/support-follow-ups-in-the-reporting-pipeline)

Follow up to https://github.com/simpledotorg/simple-server/pull/3187
Extracting more pieces from #3166 here

## Because

Adds follow ups count to the facility_states matview.

This means we can grab follow ups from the highest level matview for cases where its **not** grouped on gender.  This is much faster and quicker.

### DIFF
--- a/app/models/reports/region_summary.rb
+++ b/app/models/reports/region_summary.rb
@@ -19,6 +19,7 @@ module Reports
       lost_to_follow_up
       monthly_registrations
       monthly_overdue_calls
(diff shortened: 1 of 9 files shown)


rsanheim:
Feeling confident on this now, and it only is touching follow ups v2 code paths.

PR #3195, opened 2021-12-17
Fix width of the password field in mobile view

PriyangaPKini:
**Story card:** [sc-3791](https://app.shortcut.com/simpledotorg/story/3791/password-field-has-set-width-and-goes-beyond-the-edge-of-the-view-on-mobile-set-your-password-create-your-password-views)

## Because

Password field has set width and goes beyond the edge of the view on mobile "Set your password" / "Change your password" views

## This addresses

- Remove the unwanted styling for the password input field
- Fix the position of the `password-toggle-button`
- Style window to match others

## Test instructions
Inspect the window in different dimensions (mobile, tablet, website)

Attaching screenshots:
change-password-after
change-password-website


db/structure.sql
```diff
@@ -461,8 +461,8 @@ CREATE TABLE public.estimated_populations (
     region_id uuid NOT NULL,
```
ssrihari, line 461: Unrelated changes here.

app/assets/stylesheets/application.scss
```diff
@@ -655,11 +655,6 @@ span.inactive {
 	}
 }
 
-.password-input {
-	height: 44px;
-	width: 480px;
-}
-
```
ssrihari, line 662: This looks like it was added for a specific reason. How did you ensure that this didn't cause any regression?

ssrihari (changes requested):
One question, and one change needed to remove unrelated changes. Looks good overall, though!

db/structure.sql
PriyangaPKini: When I pull the master and ran migration these changes were present. I suppose someone forgot to commit these changes. I had discussed with @prabhanshuguptagit whether to commit these changes here. He asked me to commit them.

app/assets/stylesheets/application.scss
PriyangaPKini: The `show` button was placed on top of the password input field by setting a particular size for the field. In mobile view, because of the above set width, the password field goes beyond the screen as well as the `show` button. 
After removing these, the size of the field became relative and adjust with the change in dimension of the window.
Fixed the position of the `show` button by specifying `right-margin` instead of `left-margin`

db/structure.sql
ssrihari: Can we make this another PR, in that case? It's strange that these are getting missed, and we should deal with it separately.

app/assets/stylesheets/application.scss
ssrihari: Cool.

db/structure.sql
PriyangaPKini: Okay. I will raise a separate PR for this

claudiovallejo (commented):
Hi @PriyangaPKini — the screenshots in the PR description look great! I'd love to be able to test this in my local or a test app to test under different browser widths.

app/assets/stylesheets/application.scss
```diff
@@ -635,16 +635,16 @@ span.inactive {
...
-	line-height: 0;
-	padding: 0px;
-	position: absolute;
-	width: 47px;
-	z-index: 10;
+  color: $blue;
+  bottom: 10px;
+  font-size: 14px;
+  height: 20px;
+  right: 10px;
```
claudiovallejo, line 642: @PriyangaPKini I tested in my local and have a few minor CSS change suggestions:
1. Make `right: 9px`
2. Remove `width`
3. Set `padding-right: 4px` and `padding-left: 4px`

claudiovallejo (approved):

app/assets/stylesheets/application.scss
PriyangaPKini: Will do these chnages.

claudiovallejo (approved):

PR #3198, opened 2021-12-20
Skip phone number update if record invalid:

rsanheim:
also bump down the retry count to 5 from the default of 25 (!)

**Story card:** [ch6352](https://app.shortcut.com/simpledotorg/story/6352/handle-failed-phone-number-updates-gracefully)
This will also close out https://sentry.io/organizations/resolve-to-save-lives/issues/2861219450/?environment=production&project=1217715&query=is%3Aunresolved

## Because

We should handle this case more gracefully

## This addresses

* does not try to update a phone number if its invalid, as it will always fails
* logs the errors
* bumps retry count down to 5


vkrmis (approved):
Looks good

PR #3202, opened 2021-12-20
Fix patient risk priority calculation

prabhanshuguptagit:
**Story card:** [sc-6033](https://app.shortcut.com/simpledotorg/story/6033/fix-patient-risk-priority-calculation?vc_group_by=day&ct_workflow=all&cf_workflow=500000031&project_ids=2)

## Because

The `MedicalHistory#indicates_hypertension_risk?` method looks at `prior_heart_attack_boolean` and `prior_stroke_boolean` which are now deprecated. We should look at `prior_heart_attack` and `prior_stroke` instead.

## This addresses

Fixes `indicates_hypertension_risk`. Also adds a rachet to prevent future usages of these fields.


prabhanshuguptagit:
Reopening, needs review.

spec/rachets/medical_history_boolean_fields_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "spec_helper"
+
+RSpec.describe "Rachet down usage of boolean fields in medical history" do
+  it "prevents new usages of prior_heart_attack_boolean" do
+    expected_usages = 1
```
harimohanraj89, line 5: What's the one usage we can't remove yet? `structure.sql`?

Can we axe the field completely?

harimohanraj89 (approved):
Nice find and fix. One question, likely for a separate PR.

spec/rachets/medical_history_boolean_fields_spec.rb
prabhanshuguptagit: Yeah. Axing this field from our schema will require an API version bump, hence we've not done that here.

PR #3213, opened 2021-12-22
Days of medications dispensed graph

prabhanshuguptagit:
**Story card:** [sc-6100](https://app.shortcut.com/simpledotorg/story/6100/add-3-month-trend-card-to-region-reports)

## Because

Analyse the medication dispensation trend in all the regions

## This addresses

- Add a 3-month trend card( medications graph) to all-region reports on the details page:
  - State reports
  - District reports
  - Block reports
  - Facility reports


db/views/reporting_facility_states_v06.sql
```diff
@@ -0,0 +1,142 @@
...
+
+-- monthly follow ups
+monthly_follow_ups.follow_ups AS monthly_follow_ups,
+
+-- appointment scheduled days distribution
+reporting_appointment_scheduled_days_distributions.total_appts_scheduled AS total_appts_scheduled,
+reporting_appointment_scheduled_days_distributions.appts_scheduled_0_to_14_days AS appts_scheduled_0_to_14_days,
+reporting_appointment_scheduled_days_distributions.appts_scheduled_15_to_30_days AS appts_scheduled_15_to_30_days,
+reporting_appointment_scheduled_days_distributions.appts_scheduled_31_to_60_days AS appts_scheduled_31_to_60_days,
+reporting_appointment_scheduled_days_distributions.appts_scheduled_more_than_60_days AS appts_scheduled_more_than_60_days
```
prabhanshuguptagit, line 116: Change 1 in this view's definition.

db/views/reporting_facility_states_v06.sql
```diff
@@ -0,0 +1,142 @@
...
+    AND monthly_cohort_outcomes.region_id = rf.facility_region_id
+LEFT OUTER JOIN monthly_overdue_calls
+    ON monthly_overdue_calls.month_date = cal.month_date
+    AND monthly_overdue_calls.region_id = rf.facility_region_id
+LEFT OUTER JOIN monthly_follow_ups
+    ON monthly_follow_ups.month_date = cal.month_date
+    AND monthly_follow_ups.facility_id = rf.facility_id
+LEFT OUTER JOIN reporting_appointment_scheduled_days_distributions
+    ON reporting_appointment_scheduled_days_distributions.month_date = cal.month_date
+    AND reporting_appointment_scheduled_days_distributions.facility_id = rf.facility_id
```
prabhanshuguptagit, line 142: Change 2 in this view's definition.

app/assets/javascripts/common/medications_distribution.js
```diff
@@ -0,0 +1,191 @@
...
+    return JSON.parse(reports.getChartDataNode().textContent)["medications_dispensation"];
+  }
+
+  this.getMedicationsGraphPeriods = () => {
+    return JSON.parse(reports.getChartDataNode().textContent)["medications_dispensation_months"];
+  }
+
+  this.initializeMedicationsGraph = () => {
+    // const graphData = this.getMedicationsGraphData();
+    const graphData = {
```
PriyangaPKini, line 18: Dummy data to generate the graph.

app/models/reports/percentage.rb
```diff
@@ -0,0 +1,35 @@
...
+  def percentage(numerator, denominator, with_rounding: true)
+    return 0 if numerator.nil? || denominator.nil? || denominator == 0 || numerator == 0
+    if with_rounding
+      ((numerator.to_f / denominator) * 100).round(Reports::PERCENTAGE_PRECISION)
+    else
+      ((numerator.to_f / denominator) * 100)
+    end
+  end
+
+  def rounded_percentages(counts_hash)
```
rsanheim, line 11: I like having a function to handle this, but I'm not sure it should be baked into the repository.  My concern is that it could lead to the repo returning the rounded values that have lost precision, when I think its more the responsibility of a higher layer to handle that.

In other words, I think the repo should be trusted to return the most precise number available, and then callers could hand it off to the rounded_percentage helper if they want to make sure graphs appear correct.

app/models/reports/schema_v2.rb
```diff
@@ -191,9 +236,13 @@ def visited_without_bp_taken(with_ltfu: false)
...
-    def percentage(numerator, denominator)
-      return 0 if numerator.nil? || denominator.nil? || denominator == 0 || numerator == 0
-      ((numerator.to_f / denominator) * 100).round(PERCENTAGE_PRECISION)
+    memoize def appts_scheduled_rates(entry)
+      rounded_percentages({
+        appts_scheduled_0_to_14_days_rates: appts_scheduled_0_to_14_days[entry.region.slug][entry.period],
+        appts_scheduled_15_to_30_days_rates: appts_scheduled_15_to_30_days[entry.region.slug][entry.period],
+        appts_scheduled_31_to_60_days_rates: appts_scheduled_31_to_60_days[entry.region.slug][entry.period],
+        appts_scheduled_more_than_60_days_rates: appts_scheduled_more_than_60_days[entry.region.slug][entry.period]
+      })
```
rsanheim, line 245: I think this sort of think could probably go into a view_component that is dedicated to just the new graph, and that way the  rounded logic could live in a higher layer that the view_component uses.  Also see my other comment on this =).

app/models/reports/percentage.rb
prabhanshuguptagit: We went back and forth between a helper and keeping it in the repository. We kept it inside the repository because:
- The other methods are already exposing rounded percentages, this just fixes a rounding issue these other methods also [potentially have](https://github.com/simpledotorg/simple-server/pull/3301)
- Leaving the rounding upto the caller leaves room for them to implement their own rounding procedure which might lead to the same problem we're seeing with the proliferation of different `percentage` methods

The one reason we can think of to not do this is if in the future someone needs raw percentages from the repo. Not sure if we should build for that use case yet.

-- priyanga and prabhanshu

app/controllers/reports/regions_controller.rb
```diff
@@ -101,10 +101,21 @@ def details
...
     chart_repo = Reports::Repository.new(@region, periods: chart_range, follow_ups_v2: current_admin.feature_enabled?(:follow_ups_v2))
+    medications_dispensation_service = MedicationDispensationService.new(region: @region, period: @period)
+
+    district_regions = if @region.state_region?
+      [@region, @region.district_regions].flatten
+    end
+
+    if @region.state_region?
+      @district_repository = Reports::Repository.new(district_regions, periods: @period_range, follow_ups_v2: current_admin.feature_enabled?(:follow_ups_v2))
+    end
```
prabhanshuguptagit, line 112: Let's remove this if we didn't add this.

app/controllers/reports/regions_controller.rb
```diff
@@ -101,10 +101,21 @@ def details
 
     @repository = Reports::Repository.new(regions, periods: @period_range, follow_ups_v2: current_admin.feature_enabled?(:follow_ups_v2))
     chart_repo = Reports::Repository.new(@region, periods: chart_range, follow_ups_v2: current_admin.feature_enabled?(:follow_ups_v2))
+    medications_dispensation_service = MedicationDispensationService.new(region: @region, period: @period)
```
prabhanshuguptagit, line 104: This can be moved directly to line 117 to avoid naming.

app/controllers/reports/regions_controller.rb
```diff
@@ -101,10 +101,21 @@ def details
...
+    if @region.state_region?
+      @district_repository = Reports::Repository.new(district_regions, periods: @period_range, follow_ups_v2: current_admin.feature_enabled?(:follow_ups_v2))
+    end
 
     @chart_data = {
       patient_breakdown: PatientBreakdownService.call(region: @region, period: @period),
-      ltfu_trend: ltfu_chart_data(chart_repo, chart_range)
+      ltfu_trend: ltfu_chart_data(chart_repo, chart_range),
+      medications_dispensation: medications_dispensation_service.distribution,
+      medications_dispensation_months: medications_dispensation_service.months
```
prabhanshuguptagit, line 118: Alternative 1
```
medications_dispensation: {distribution: medications_dispensation_service.distribution,
                                 months: medications_dispensation_service.months}
```

Alternative 2
Drop the months list and build the legends by parsing the distribution data structure in JS

app/services/medication_dispensation_service.rb
```diff
@@ -0,0 +1,34 @@
+class MedicationDispensationService
+  def initialize(region:, period:)
+    @region = region
+    @period = period
+    medications_dispensation_months = -2
```
prabhanshuguptagit, line 5: Extract to a class constant: `MONTHS_OF_DATA` or `MONTHS`.

app/controllers/reports/regions_controller.rb
PriyangaPKini: Came from this PR `Show overdue call result` #3126

app/services/medication_dispensation_service.rb
```diff
@@ -0,0 +1,34 @@
...
+      "60+ days" => {color: "#007AA6",
+                     counts: medications_dispensation_repo.appts_scheduled_more_than_60_days[@region.slug],
+                     totals: medications_dispensation_repo.total_appts_scheduled[@region.slug],
+                     percentages: medications_dispensation_repo.appts_scheduled_more_than_60_days_rates[@region.slug]}
+    }
+  end
+
+  def months
+    @medications_dispensation_range.map(&:to_s)
+  end
```
prabhanshuguptagit, line 33: Is this unused?

spec/models/reports/percentage_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe Reports::Percentage, type: :helper do
+  describe "#percentage" do
```
prabhanshuguptagit, line 4: Should add specs that test `with_rounding`

spec/models/reports/region_summary_spec.rb
```diff
@@ -249,4 +254,9 @@ def refresh_views
       expect(facility_results[period]["lost_to_follow_up"]).to eq(1)
     end
   end
+
+  describe "#for_regions" do
+    it "excludes facilities where registrations, assigned patients and follow ups are all zero" do
+    end
```
prabhanshuguptagit, line 260: Needs to be filled in

spec/services/medication_dispensation_service_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe MedicationDispensationService, type: :model do
+  it "returns a hash of hashes containing key/value pairs of attributes of AppointmentsScheduledDaysDistribution for a region and a period range" do
```
prabhanshuguptagit, line 4: Alternative
```
returns bucketed days of medications data
```

This can be written more declaratively by constructing a hash with values and comparing directly.

PR #3227, opened 2022-01-03
Grab data even if a facility _just_ has follow ups

rsanheim:
...and no registered or assigned patients

**Story card:** [sc-6496](https://app.shortcut.com/simpledotorg/story/6496/follow-up-data-not-visible-in-district-details-but-visible-in-facility-details)

Fix issue where a facility has no registered or assigned patients, but does have follow up numbers.

### DIFF
--- a/app/models/reports/region_summary.rb
+++ b/app/models/reports/region_summary.rb
@@ -68,7 +68,7 @@ module Reports
     def for_regions
       FacilityState
         .where(id_field => regions.map(&:id))
-        .where("cumulative_registrations IS NOT NULL OR cumulative_assigned_patients IS NOT NULL")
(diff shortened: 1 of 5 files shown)


vkrmis (commented):
@rsanheim we also skip facilities with no patients in the view (non_facility_details)

rsanheim:
> @rsanheim we also skip facilities with no patients in the view (non_facility_details)


Fixed the query underlying that.

I'm not sure we need that skip at all in the view - maybe it was there to prevent a huge list of empty data when a Region is just getting setup? I could go either way on removing it.

vkrmis (commented):
lgtm

vkrmis:
> maybe it was there to prevent a huge list of empty data

That's probably it

PR #3231, opened 2022-01-04
Add profiling seed type with more data

rsanheim:
Larger range of facilities for profiling seed type

Need a profile type that is as close to IN production as possible.

Extracted from perf and profiling work in https://github.com/simpledotorg/simple-server/pull/3229.


rsanheim:
This is seed related only, and will get https://github.com/simpledotorg/simple-server/pull/3229 smaller and easier to iterate on.

PR #3235, opened 2022-01-05
Create January 2022 IHCI Experiment 

PriyangaPKini:
**Story card:** [sc-6538](https://app.shortcut.com/simpledotorg/story/6538/setup-sms-only-experiment)

## Because

We cancelled the previous experiment because we decided not to use WhatsApp for sending out messages further in the experimentation. One of the reasons for not using WhatsApp was it had a daily limit of 10000 messages.

## This addresses

- Created a new experiment that will be only using SMS to send out messages. 
- This also increases the number of messages send out daily because we are not using WhatsApp
   -  Current Patient Experiment can now enrol 20k patients daily 
   -  Stale Patient Experiment can now enrol 15k patients daily


harimohanraj89 (approved):
:shipit: 🚀

PR #3242, opened 2022-01-06
Add frozen_string_literal

rsanheim:
[sc6531](https://app.shortcut.com/simpledotorg/story/6531/region-details-is-slow-for-large-regions)

using https://github.com/Invoca/magic_frozen_string_literal to do this
everywhere


rsanheim:
This had a very negligible difference in performance tests i ran locally (no more than 1.08% better against details page, for example).

So not worth doing for right now.

PR #3245, opened 2022-01-07
Add facility size summaries

timcheadle:
**Story card:** [sc-6276](https://app.shortcut.com/simpledotorg/story/6276/add-facility-size-summaries-to-the-state-and-district-reports)

## Because

CVHOs requested that we add summed rows for each facility size to the monthly CSV download

## This addresses

This sums each data point by facility size and adds it to the spreadsheet. It also cleans up unneeded columns (state and district).

## Test instructions

1. Go to the test app
2. Click on "Reports", make sure you're looking at a district level report (one level below state)
3. Click the "Download" button top right
4. Click "Monthly facility data"

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -10,6 +10,9 @@ auto_cancel:
 blocks:
   - name: Cache dependencies
     task:
+      env_vars:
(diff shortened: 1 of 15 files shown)


rsanheim (approved):
LGTM

PR #3268, opened 2022-01-11
Remove bootboot from Gemfile

rsanheim:
This is breaking deploys, because the path returned for the gem is a fully qualified path from a release that has been cleaned up.

see https://github.com/rubygems/rubygems/issues/3304
and https://github.com/rubygems/rubygems/issues/3340

**Story card:** [sc-6351](https://app.shortcut.com/simpledotorg/story/6351/setup-dual-dependency-approach-to-prepare-for-rails-6-and-beyond)

PR #3270, opened 2022-01-11
Add a custom task to fix Bundler plugin path for bootboot

rsanheim:
It turns out the bundler config for plugins is cached everywhere...so rather than trying to remove it and clean it up on every server, going to try to roll forward here and fix it.

This is a follow up to https://github.com/simpledotorg/simple-server/pull/3268.

**Story card:** [sc-6351](https://app.shortcut.com/simpledotorg/story/6351/setup-dual-dependency-approach-to-prepare-for-rails-6-and-beyond)

### DIFF
--- a/lib/capistrano/tasks/deploy.rake
+++ b/lib/capistrano/tasks/deploy.rake
@@ -11,7 +11,6 @@ namespace :deploy do
   task :fix_bundler_plugin_path do
     on release_roles([:all]) do
       within shared_path do
-        # sed -i 's#/home/deploy/apps/simple-server/releases/[0-9]\+/.bundle/#/home/deploy/apps/simple-server/shared/.bundle/#g' plugin/index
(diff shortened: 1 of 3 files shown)


lib/capistrano/tasks/deploy.rake
```diff
@@ -3,6 +3,27 @@ namespace :deploy do
     invoke "sidekiq:add_default_hooks"
   end
 
+  desc "Fix Bundler plugin path so it points to the shared path instead of a release path"
+  task :fix_bundler_plugin_path do
+    on release_roles([:all]) do
+      within shared_path do
+        # sed -i 's#/home/deploy/apps/simple-server/releases/[0-9]\+/.bundle/#/home/deploy/apps/simple-server/shared/.bundle/#g' plugin/index
+        execute "sed", "-i", '"s#/home/deploy/apps/simple-server/releases/[0-9]\+/#/home/deploy/apps/simple-server/shared/#g"', ".bundle/plugin/index"
```
rsanheim, line 11: We need a carefully crafted sed command here to replace the release dir with the shared dir, but this finally works.

PR #3272, opened 2022-01-11
Fix heroku bundler plugin path

vkrmis:
Fix the bundle path for review app deployments on heroku

This is a follow up to https://github.com/simpledotorg/simple-server/pull/3270.

**Story card:** [sc-6351](https://app.shortcut.com/simpledotorg/story/6351/setup-dual-dependency-approach-to-prepare-for-rails-6-and-beyond)


rsanheim:
Thx for this @vkrmis

PR #3277, opened 2022-01-12
Require facility size

timcheadle:
**Story card:** [sc-6568](https://app.shortcut.com/simpledotorg/story/6568/not-able-to-download-monthly-facility-data-from-dashboard)

## Because

Monthly facility download doesn't work if facilities are missing facility sizes.

## This addresses

 Since missing sizes also make facilities disappear from some reports, we're making them mandatory.

### DIFF
--- a/db/data/20220112012521_add_default_facility_size.rb
+++ b/db/data/20220112012521_add_default_facility_size.rb
@@ -1,0 +1,9 @@
+class AddDefaultFacilitySize < ActiveRecord::Migration[5.2]
+  def up
+    Facility.where(facility_size: nil).update_all(facility_size: :community)
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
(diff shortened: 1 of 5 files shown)


db/data/20220112012521_add_default_facility_size.rb
```diff
@@ -0,0 +1,9 @@
+class AddDefaultFacilitySize < ActiveRecord::Migration[5.2]
+  def up
+    Facility.where(facility_size: nil).update_all(facility_size: :community)
```
vkrmis, line 3: Don't think we can do this, there are larger facilities that don't have sizes set as well. This will also run in all envs, so it'll be messy to do based on facility_type. Wdyt about just fixing this manually on prod envs via the app console?

timcheadle: We talked about this and the console is the right answer. We'll fix that before we deploy this.

vkrmis: Sizes have been set via the console, this should be safe to run now

vkrmis (approved):

PR #3279, opened 2022-01-12
Introduce in-memory Region ancestry computation

rick:
giphy-flash

**Story card:** [ch-5562](https://app.shortcut.com/simpledotorg/story/5562/improve-regions-index-performance)

We are currently still seeing slow performance of Region tree accesses in places like `/reports/regions`, though now we are hitting excessive caching overhead. Given the relatively small size of these trees, `Region.all` is very fast, and computing the needed data in memory should be much faster than caching round-trips. We expect ActiveRecord to cache `Region.all` reasonably if necessary. Relying on `path` as a `.`-separated list of path elements gives us sufficient information to compute the useful tree data structures in memory.

This is partly a spike, but should be convertible to prod quality in short order.

Screen Shot 2022-01-12 at 3 12 04 PM

**the work**

 - [x] spike: introduce a means of computing `Region#ancestor` via `Region.all` + in-memory traversal of `path` elements
 - [x] add an alternate controller action endpoint to do development in parallel
 - [x] add an alternate region tree component to allow experimenting in parallel
 - [x] use ltree's `.children` to render a new version of the regions tree
 - [x] introduce a way to get children through fast in-memory computation
 - [x] use fast children in rendering regions tree
 - [x] ensure that we prune `Region` entries that don't have any facilities
 - [x] ensure that we are obeying accessibility constraints
 - [x] get this behind a feature flag
   - [x] introduce flag, delegate #index -> #fastindex
   - [x] test flag in PR view environment
   - [x] test flag in sandbox
   - [x] test flag in bangladesh:demo
   - [x] test flag in bangladesh:production

### post merge

 - [ ] ask to enable feature flag in india:production
 - [ ] collect data on n% renderings of this code path in india:production

cc @rsanheim


rick:
Adding deployment keys to allow watching this in sandbox and Bangladesh environments at: https://github.com/simpledotorg/deployment/pull/398

PR #3289, opened 2022-01-13
Monthly district export: add indicators aggregated by facility size

vkrmis:
**Story card:** [sc-6383](https://app.shortcut.com/simpledotorg/story/6383/district-data)

## Because

We previously explored calculating facility size based indicators in the excel template itself. 
Since this is tricky to get right, we will populate these indicators server-side.


timcheadle (approved):
At some point we should unify approaches for indicators by facility size (as I did the Ruby-oriented version in the monthly downloads)

PR #3295, opened 2022-01-14
Add a patch to turn Sidekiq#fetch sampling way down

rsanheim:
**Story card:** [sc-6635](https://app.shortcut.com/simpledotorg/story/6560/excessive-datadog-spans-due-to-sidekiq-fetch-monitoring#activity-6635)

## Because

We don't want to trace millions of sidekiq operations every day.

## This addresses

Dialing down the sidekiq sample rate to 1%. It looks like this rate impacts `fetch`, `heartbeat`, and scheduled job checks. It isn't clear to me if this will _also_ impact other sidekiq processing (like job execution), but from my digging around it looks like this sample rate is _just_ for sidekiq internals.

## Test instructions

I ran DD locally for testing this, and after some hacky attempts at patching ddtrace itself, I realized the sample rates are set per service.  Here is what the sidekiq span looks like after my change -- note the massive drop in the yellow bars.

image


Gemfile
```diff
@@ -37,7 +37,7 @@ gem "bootstrap", "~> 4.5.0"
 gem "connection_pool"
 gem "data_migrate"
 gem "data-anonymization", require: false
-gem "ddtrace", "~> 0.54"
```
rsanheim, line 40: Also getting us back to plain ddtrace here, without a gem requirement.

rsanheim:
Going to merge and monitor our traces in SBX etc.

PR #3302, opened 2022-01-18
Add follow up sums for gender / diagnosis to reporting pipeline

rsanheim:
**Story card:** [sc-4212](https://app.shortcut.com/simpledotorg/story/4212/support-follow-ups-in-the-reporting-pipeline)

## Because

We need monthly counts at the facility level for follow ups by gender and diagnosis.

## This addresses

Adding counts to `reporting_facility_state_groups`


rsanheim:
@rick and I reviewed this while pairing today, and we confirmed that this doesn't yet impact production because its behind the `follow_ups_v2_progress_tab` feature flag to carve out that particular end path. Going to ship this so I can continue to iterate on the actual wiring in the controller layers.

PR #3304, opened 2022-01-20
Reports details: Fix links to add estimated hypertension population

vkrmis:
**Story card:** [sc-6800](https://app.shortcut.com/simpledotorg/story/6800/the-link-in-the-reports-details-page-to-add-estimated-htn-population-for-a-district-is-broken)

## Because

- The link to add estimated hypertension population for districts was using the `district region` instead of the `facility_group`.
- The link to add estimated hypertension population for states was checking the `view_reports` permission instead of `manage`. Users with a `view_reports` permission can't actually set the estimated hypertension population.


harimohanraj89 (approved):
Makes sense.

rsanheim (approved):
Yup, and removing the accessible check makes sense too 👍 . I'll merge after checks are green.

PR #3306, opened 2022-01-20
Update deployment docs a bit based on pairing

rsanheim:
Helping out @rick so that he can deploy, and our docs around branch
deploys could be more clear.

Sandbox does not have a country name, and the BRANCH env var could be
better explained here.


rick (approved):
giphy (9)

PR #3310, opened 2022-01-21
set :passenger_restart_with_sudo to true

vkrmis:
**Story card:** [sc-6603](https://app.shortcut.com/simpledotorg/story/6603/stale-passenger-processes-hanging-around)

## Because

We're seeing some permission related warnings in our deployments

## This addresses

Restarts passenger using sudo, this is required so stale tmp files owned by the root user can be cleaned up.

### Run the [deploy playbook](https://github.com/simpledotorg/deployment/pull/401) on all envs manually before merging this PR
- [x] Sandbox
- [x] QA
- [x] Performance
- [ ] ~Security~ env is not in use?

#### Demo
- [x] Bangladesh
- [x] Ethiopia
- [x] India
- [x] Sri Lanka 

#### Production
- [x] Bangladesh
- [x] Ethiopia
- [x] India
- [x] Sri Lanka


harimohanraj89 (approved):
Neat little config available 👍 Yup makes sense to me, let's do it.

PR #3314, opened 2022-01-24
Localize and enable monthly downloads everywhere

timcheadle:
**Story cards:**
- [sc-6598](https://app.shortcut.com/simpledotorg/story/6598/translate-state-district-facility-etc-in-report-downloads)
- [sc-6688](https://app.shortcut.com/simpledotorg/story/6688/turn-off-india-only-filter-in-navigation-views)

## Because

Monthly district and facility summary data are useful downloads, and we want them available everywhere.

## This addresses

This translates all region names in the downloads (and links), enables them outside of India, and adds a feature flag in case something goes wrong.

### DIFF
--- a/app/models/region.rb
+++ b/app/models/region.rb
@@ -196,6 +196,11 @@ class Region < ApplicationRecord
     I18n.t("region_type.#{region_type}")
   end
 
+  def localized_child_region_type
+    return nil if child_region_type.nil?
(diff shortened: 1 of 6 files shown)


rick (approved):
giphy-downsized-large (4)

claudiovallejo (approved):
Looks great! Thank you Tim :)

PR #3317, opened 2022-01-25
Fix auth checks on the estimated hypertension card

vkrmis:
**Story card:** [sc-6866](https://app.shortcut.com/simpledotorg/story/6866/state-level-reports-are-broken)

## Because

We accidentally broke state level reports by doing a `:manage` check on state regions

### DIFF
--- a/app/views/reports/regions/_hypertension_patient_coverage.html.erb
+++ b/app/views/reports/regions/_hypertension_patient_coverage.html.erb
@@ -28,7 +28,7 @@
   </div>
 <% elsif !(@region.estimated_population&.show_coverage) %>
     <% if @region.district_region? %>
-      <% if (accessible_region?(@region, :manage)) %>
+      <% if accessible_region?(@region, :manage) %>
         <div class="mb-16px p-12px br-4px bg-blue-light">
           <p class="m-0px fs-14px c-black">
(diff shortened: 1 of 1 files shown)


app/views/reports/regions/_hypertension_patient_coverage.html.erb
```diff
@@ -26,32 +26,38 @@
...
+          <p class="m-0px fs-14px c-black">
+            Ask a Dashboard admin to add the est. hypertensive population for <%= @region.name %> to see the hypertension coverage rate.
+          </p>
+        </div>
+      <% end %>
     <% elsif @region.state_region? %>
-      <p class="m-0px fs-14px c-black">
-        Ask a Dashboard admin to add the est. hypertensive population for all districts ine<%= @region.name %> to see the hypertension coverage rate.
-      </p>
+      <% if @region.district_regions.all? { |district_region| accessible_region?(district_region, :manage) } %>
```
rsanheim, line 46: This could be really slow for large states as it will do n+1 calls for every district.  Can we do this in a more efficient way?

vkrmis: Switched to a `superset?` check, so we'll fetch the `accessible_district_regions` up front.

PR #3322, opened 2022-01-27
Rate limit OTPs from the same user ID

ssrihari:
**Story card:** [sc-6911](https://app.shortcut.com/simpledotorg/story/6911/no-rate-limit-on-otp)


config/initializers/rate_limiter.rb
```diff
@@ -51,9 +44,9 @@ class Rack::Attack
       end
     end
 
-    throttle("throttle_user_activate", RateLimit.user_api_options) do |req|
+    throttle("throttle_user_activate", RateLimit.auth_api_options) do |req|
       if req.post? && req.path.start_with?("/api/v4/users/activate") && SimpleServer.env.production?
-        req.ip
+        req.ip && req.params['user']['id']
```
harimohanraj89, line 49: This is generating a boolean result, right? I believe that's the wrong thing to return here. We need to return a unique "cache key" that will be used to categorize each rate limit bucket. I think we've accidentally created only 2 buckets - true and false. We probably want to do something like 

`"#{req.ip} #{req.params['user'][id]}`

harimohanraj89 (changes requested):
Blocking change requested.

config/initializers/rate_limiter.rb
harimohanraj89: Ah, additionally, I think our specs will need to be updated so that in addition to `expect to be blocked when i > limit`, we also need the `expect _not_ to be blocked when i <= limit`. That would have surfaced this problem.

ssrihari: > This is generating a boolean result, right?

Since the IP would always exist, this would actually just return the user-id. Which is, accidentally, correct 😁 . We do have the other test as well, actually.

I've changed this in ae2174f1566369613b1b87fa9b5eb56da68aaa9a to return just the user-id. This should be alright now, ya?

harimohanraj89: Ha, that's amazing 😂 Yeah just a user ID should be great, and is simple to grok as well.

harimohanraj89 (approved):
LOL it works! :shipit:

PR #3331, opened 2022-01-28
Send validation errors to logs instead of Sentry

harimohanraj89:
**Story card:** [sc-5177](https://app.shortcut.com/simpledotorg/story/5177/sync-validation-error-move-exception-to-datadog)

## Because

We don't want to send these no-action-required incidents to Sentry, we want to send them to logs to be parsed in DD if necessary.


vkrmis (approved):
🚢

PR #3333, opened 2022-01-28
allow <a> tags to wrap

vkrmis:
**Story card:** [sc-5567](https://app.shortcut.com/simpledotorg/story/5567/fix-overflow-on-drug-consumption-screen)

## Because

Blocks and facilities with long names break styling on the drug consumption page

## This addresses

Adds a css class to allow anchor tags to wrap

### DIFF
--- a/app/assets/stylesheets/partials/_dashboard.scss
+++ b/app/assets/stylesheets/partials/_dashboard.scss
@@ -209,6 +209,10 @@ th sup {
       text-overflow: ellipsis;
       overflow: hidden;
     }
+
+    a.allow-wrap {
+      white-space: normal !important;
+    }
   }
 
   col { border-left: 1px solid $grey-lightest; }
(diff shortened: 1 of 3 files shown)


vkrmis:
The wrapping problem happens when there are long names as well as a lot of columns. We would need to ellips-ize much earlier to avoid the wrapping, since the drug consumption page has 12 columns in some places. Curiously the drug stock page introduces a horizontal scroll instead of breaking the table. Ideally we'd pick one style and go with it for both these pages.

Drug Consumption with wrapping
Screenshot 2022-01-28 at 6 51 38 PM

Drug Consumption
Screenshot 2022-01-28 at 6 49 22 PM

Drug Stock
Screenshot 2022-01-28 at 6 49 09 PM

vkrmis:
> Curiously the drug stock page introduces a horizontal scroll instead of breaking the table. 

Fixed this discrepancy. Both pages can now horizontally scroll.
Also allowing the names to wrap on both pages, so hopefully we don't end up horizontally scrolling too often.

Alternatively, we could set a much smaller column width, and ellips-ize early, but I'm not a fan of this since we have lots of facilities with very long names (they sometimes include the name of their corresponding PHC in the name). 

wdyt @dburka, @claudiovallejo

dburka:
Sounds like a fine compromise @vkrmis . Ship it.

PR #3341, opened 2022-01-31
Demo country in production

harimohanraj89:
**Story card:** [sc-6861](https://app.shortcut.com/simpledotorg/story/6861/fix-google-play-policy-issue)

## Because

We need to make the production app browsable in a "demo" mode for Google Play app reviewers to give us the 👍 Without this approval, we cannot make any releases to the production Simple app.

## This addresses

Add a "Demo" country to the production manifest that points at Sandbox. Users who select this country (eg. Google Play reviewers) can self sign-up and connect to Sandbox. Upon re-sign-in, they can use the fixed OTP of 000000, since Sandbox has fixed OTPs.

This approach avoids the following, which is good:

* Introducing conditional OTP logic into the application
* Adding test data to a production server
* Introducing any "demo mode" into the app that can be switched to, resulting in confusion for real users.

This is a follow-up PR to #3340 and #3332

### DIFF
--- a/public/manifest/demo.json
+++ b/public/manifest/demo.json
@@ -25,9 +25,9 @@
       "isd_code": "94"
     },
     {
-      "country_code": "IN",
+      "country_code": "DEMO",
       "endpoint": "https://api-sandbox.simple.org/api/",
-      "display_name": "Sandbox",
+      "display_name": "Demo Country",
       "isd_code": "91"
     }
   ],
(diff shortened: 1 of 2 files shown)


rsanheim:
Will the demo country show up for regular users in production?

Also, I'm a little concerned about pointing this at sandbox, as that server can often be less reliable then our demo server.  Could we point at india demo instead?

harimohanraj89:
You and @timcheadle were on the same page 👍 Yes, I'll point the prod manifest at India Demo instead.

And yes, this will show up for all production users. We ran it by the design team to make sure that was ok. We can get smarter with it in future iterations, but we wanted to unblock our android release pipe asap.

PR #3345, opened 2022-02-01
Show footnote for medications_dispensation feature only when it's enabled

PriyangaPKini:
**Story card:** [sc-7023](https://app.shortcut.com/simpledotorg/story/7023/show-footnote-for-medications-dispensation-only-when-it-s-enabled)

## Because

Show footnote for medications_dispensation feature only when it's enabled

## This addresses

Add a condition around the footnote to check if medications_dispensation is enabled or not

## Other Changes

- Remove extra space around the hyphens: `0-14 days` instead of `0 - 14 days`
- Change the card title to `Days of medications dispensed to follow-up patients`


prabhanshuguptagit (approved):
🚀

PR #3348, opened 2022-02-02
BD help contact - take two

rsanheim:
**Story card:** [sc-7045](https://app.shortcut.com/simpledotorg/story/7045/invalid-encoding-in-bangladesh-help-page-link)

## Because

The original PR introduced an encoding issue on BD production, so I had to revert it.

I'm just reopening the original branch for @dburka so we don't lose track of this, but the URL needs some UTF8 applied I believe.


harimohanraj89:
I've added the proper URL-encoded link, and verified it works in my browser and on a device pointed at my local server.

@rsanheim Ready for your review.

PR #3351, opened 2022-02-02
Fix bug in days of medications calculation

prabhanshuguptagit:
**Story card:** [sc-7058](https://app.shortcut.com/simpledotorg/story/7058/fix-calculation-for-days-until-next-appointment)

## Because

We noticed that the the calculation for days until next appointment is not correct and the days might not be getting bucketed correctly.

## This addresses

There are two bugs that cancel each other out:
- The thresholds for bucketing are not set correctly (should be `0, 15, 31, 61` instead of `0, 15, 30, 60`).
- The difference between `scheduled_date` and `device_created_at` is not being calculated correctly. The formula is undercounting the difference by 1.

Both these bugs together mean that the data is being bucketed correctly in the edge cases. An appointment `31 days` from now is calculated as `30 days from now` and it gets included in bucket 2 as expected.

This fixes both these issues.


spec/models/reports/facility_appointment_scheduled_days_spec.rb
```diff
@@ -24,19 +24,23 @@
...
+    create(:appointment, scheduled_date: 14.days.from_now, device_created_at: Time.current, facility: facility, patient: create(:patient, recorded_at: 1.month.ago))
+    create(:appointment, scheduled_date: 15.days.from_now, device_created_at: Time.current, facility: facility, patient: create(:patient, recorded_at: 1.month.ago))
+    create(:appointment, scheduled_date: 30.days.from_now, device_created_at: Time.current, facility: facility, patient: create(:patient, recorded_at: 1.month.ago))
+    create(:appointment, scheduled_date: 31.days.from_now, device_created_at: Time.current, facility: facility, patient: create(:patient, recorded_at: 1.month.ago))
+    create(:appointment, scheduled_date: 60.days.from_now, device_created_at: Time.current, facility: facility, patient: create(:patient, recorded_at: 1.month.ago))
+    create(:appointment, scheduled_date: 61.days.from_now, device_created_at: Time.current, facility: facility, patient: create(:patient, recorded_at: 1.month.ago))
+    create(:appointment, scheduled_date: 100.days.from_now, device_created_at: Time.current, facility: facility, patient: create(:patient, recorded_at: 1.month.ago))
```
ssrihari, line 35: Want to put the dates in an array and loop instead? Will be more readable that way.

db/structure.sql
```diff
@@ -1629,15 +1629,15 @@ CREATE MATERIALIZED VIEW public.reporting_facility_appointment_scheduled_days AS
...
             count(*) AS number_of_appointments,
             latest_appointments_per_patient_per_month.creation_facility_id AS facility_id
            FROM latest_appointments_per_patient_per_month
-          GROUP BY (width_bucket((date_part('days'::text, ((latest_appointments_per_patient_per_month.scheduled_date)::timestamp without time zone - latest_appointments_per_patient_per_month.device_created_at)))::integer, ARRAY[0, 15, 30, 60])), latest_appointments_per_patient_per_month.creation_facility_id, latest_appointments_per_patient_per_month.month_date
+          GROUP BY (width_bucket((date_part('days'::text, ((latest_appointments_per_patient_per_month.scheduled_date)::timestamp without time zone - date_trunc('day'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('UTC'::text, latest_appointments_per_patient_per_month.device_created_at))))))::integer, ARRAY[0, 15, 31, 61])), latest_appointments_per_patient_per_month.creation_facility_id, latest_appointments_per_patient_per_month.month_date
```
ssrihari, line 1640: Did the tz fixes here have a direct effect on the issue as well? Or was it an additional fix?

Also, can you explain the `date_part` vs `date_trunc` difference in words? It's a tad difficult to parse this diff and understand what changed.

db/migrate/20220202091240_update_reporting_facility_appointment_scheduled_days_to_version_2.rb
```diff
@@ -0,0 +1,9 @@
+class UpdateReportingFacilityAppointmentScheduledDaysToVersion2 < ActiveRecord::Migration[5.2]
+  def change
+    drop_view :reporting_facility_states, materialized: true, revert_to_version: 6
+    update_view :reporting_facility_appointment_scheduled_days, version: 2, revert_to_version: 1, materialized: true
+
+    create_view :reporting_facility_states, materialized: true, version: 6
+    add_index :reporting_facility_states, [:month_date, :facility_region_id], name: :facility_states_month_date_region_id, unique: true
```
ssrihari, line 7: Was this the only index?

ssrihari (approved):

db/migrate/20220202091240_update_reporting_facility_appointment_scheduled_days_to_version_2.rb
prabhanshuguptagit: Yes.

PR #3362, opened 2022-02-04
[ADR] Data API

ssrihari:


doc/arch/020-data-api.md
```diff
@@ -0,0 +1,108 @@
...
+mechanism. The idea is to provide all the data in the hands of the
+users, so that they may customize it as they see fit.
+
+Since we aren't seeing direct benefits of the API without the
+downloads interface, this will live only inside the dashboard to begin
+with.
+
+We will create the following APIs:
+- /data/v1/patients.csv:
+    - all the data in `reporting_patient_states`
```
harimohanraj89, line 62: Curious - how big would this CSV be today?

doc/arch/020-data-api.md
```diff
@@ -0,0 +1,108 @@
...
+    - Has all the data in `drug_stocks`, denormalized with `protocol_drugs`
+
+They will all be filterable using the parameters:
+- `region_id`: uuid of a state, district, block, or facility region. Should this be a list?
+- `period_type`: month/quarter
+- `month_date/period`: range in period_type
+
+All the APIs will go through the same authenticationi layer, which
+decides access to:
+- APIs
```
harimohanraj89, line 79: The only API control is whether you can see patients.csv right? We'd leverage our existing `view_reports_only` vs `view_pii` access levels to determine that?

doc/arch/020-data-api.md
```diff
@@ -0,0 +1,108 @@
...
+- Regions
+
+### Documentation
+We will have a translation layer that renames the columns to keys in
```
harimohanraj89, line 83: Could we just `humanize` the column names, since we can put the DB annotations into, say, the second row of the CSV for each column?

doc/arch/020-data-api.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Performance considerations
+Reading and downloading large swaths of data can be slow. Here are few
+considerations to keep in mind to explore while building this:
+- Email vs direct downloads: we currently email the patient line lists
+- Restricting periods and regions for performance: a drop with last
+  1/3/6 months perhaps? Disallow state level downloads?
+- Using the read replica for downloads: so that transactions aren't
+  impacted by this
+- Uplaoding CSVs to S3, or storing them locally to avoid repeated
```
harimohanraj89, line 97: *Uploading

PR #3367, opened 2022-02-07
Progress tab period report template

claudiovallejo:
**Story card:** [sc-7272](http://localhost:3000/reports/progress/uhc-mace-village)

## Because

We're creating new period subpages for the Progress Tab. This PR creates a single HTML template for the daily, monthly, and yearly report pages.

## This addresses
image

+ Creates a new `_period_report.html.erb` partial
+ Adds dummy data to the controllers (will be deleted when real data is available)
+ Creates a new `progress_tab.js` standalone script to show/hide the report pages

## Test instructions

+ Enable the `:new_progress_tab` Flipper flag
+ Go to a facility report page and click "Progress reports"
+ Click through the "Daily", "Monthly", and "Yearly" reports. Make sure that the links work, the UI looks good.
+ When you click the "back" arrow, the Home page should appear. Scroll to the very bottom of the home page and make sure just the home page is showing.
+ Open `.env.development` and set `DEFAULT_COUNTRY=ET`
+ Restart your server
+ Go to a facility report page, click "Progress reports" and open the "Daily" report. There shouldn't be a yellow banner at the end of the page (should only appear for India)


app/controllers/api/v3/analytics/user_analytics_controller.rb
```diff
@@ -13,6 +13,120 @@ def show
       @service = Reports::FacilityProgressService.new(current_facility, @period)
     end
 
+    if Flipper.enabled?(:new_progress_tab)
+      @daily_periods = ["8-Feb-2022", "7-Feb-2022", "6-Feb-2022", "5-Feb-2022", "4-Feb-2022", "3-Feb-2022", "2-Feb-2022"]
```
rsanheim, line 17: Can you break out these static values to their own object?  you could name it `Reports::FakeFacilityProgressService` (as an example).  Then we can replace it with the real service / real numbers when we get to that work.  It could return all the values you need to populate the ivars, or perhaps return a hash with keys like `daily_registered_patients`, `daily_periods`, `monthly_patients...`, etc.

app/controllers/reports/progress_controller.rb
```diff
@@ -10,7 +10,119 @@ def show
     @current_facility = @region
     @user_analytics = UserAnalyticsPresenter.new(@region)
     @service = Reports::FacilityProgressService.new(current_facility, @period)
+
     if Flipper.enabled?(:new_progress_tab)
+      @daily_periods = ["8-Feb-2022", "7-Feb-2022", "6-Feb-2022", "5-Feb-2022", "4-Feb-2022", "3-Feb-2022", "2-Feb-2022"]
```
rsanheim, line 15: This code path could use the fake service as well.

rsanheim (approved):
Approving to unblock, but please do extract all the fake numbers to their own before merging. 👍

PR #3368, opened 2022-02-08
Rack attack fix: Use dig to avoid errors if params are malformed

rsanheim:
Trying a quick fix for [sc7150](https://app.shortcut.com/simpledotorg/story/7150/nomethoderror-undefined-method-for-nil-nilclass)

We need to handle malformed params, as we are getting quite a few sentry errors where params do not have `[:user][:id]`:

https://sentry.io/organizations/resolve-to-save-lives/issues/2993018371/events/506518eaccc247d3930ff3cdba91711e/events/?project=1217715&referrer=Shortcut


vkrmis:
Closing this, these params are required and should be present. There's a fix here: https://github.com/simpledotorg/simple-server/pull/3371

PR #3375, opened 2022-02-09
Fix meds specs failing in US time

prabhanshuguptagit:
**Story card:** [sc-7058](https://app.shortcut.com/simpledotorg/story/7058/fix-calculation-for-days-until-next-appointment)

## Because

Some specs were failing due to timezone issues. 

## This addresses

The specs failed in US time afternoon because the data was created in US time but the view was being refreshed in `Asia/Kolkata` time. This freezes time to IST similar to the other specs. 

Note to reviewer: Ignore whitespace while reviewing.

### DIFF
--- a/spec/models/reports/facility_state_spec.rb
+++ b/spec/models/reports/facility_state_spec.rb
@@ -252,8 +252,8 @@ RSpec.describe Reports::FacilityState, {type: :model, reporting_spec: true} do
 
           expect(described_class.find_by(month_date: Period.current, facility: facility).appts_scheduled_0_to_14_days).to eq 2
           expect(described_class.find_by(month_date: Period.current, facility: facility).total_appts_scheduled).to eq 2
-          expect(described_class.find_by(month_date: Period.month(1.month.ago), facility: facility).appts_scheduled_32_to_62_days).to eq 1
-          expect(described_class.find_by(month_date: Period.month(2.month.ago), facility: facility).appts_scheduled_more_than_62_days).to eq 1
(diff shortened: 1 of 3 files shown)


spec/models/reports/facility_state_spec.rb
```diff
@@ -229,31 +229,32 @@
...
-          patient: patients.first,
-          scheduled_date: Date.today,
-          device_created_at: 63.days.ago)
-
-        RefreshReportingViews.new.refresh_v2
-
-        expect(described_class.find_by(month_date: Period.current, facility: facility).appts_scheduled_0_to_14_days).to eq 2
-        expect(described_class.find_by(month_date: Period.current, facility: facility).total_appts_scheduled).to eq 2
-        expect(described_class.find_by(month_date: Period.month(1.month.ago), facility: facility).appts_scheduled_32_to_62_days).to eq 1
-        expect(described_class.find_by(month_date: Period.month(2.month.ago), facility: facility).appts_scheduled_more_than_62_days).to eq 1
```
PriyangaPKini, line 256: Let's use `31.days.ago`  and  `62.days.ago`  instead of `1.month.ago` or `2.month.ago`

PriyangaPKini (changes requested):
Please check the comment.

spec/models/reports/facility_state_spec.rb
prabhanshuguptagit: Addressed in [f703355](https://github.com/simpledotorg/simple-server/pull/3375/commits/f70335578f7932b0f3311be328dc816ce5c6b198)

PriyangaPKini (approved):
🚀

PR #3377, opened 2022-02-10
Rename Config to EnvHelper; make sure it gets loaded early

rsanheim:
We have this top level module from the old days of the project called `Config` - Rails 6 really doesn't like this. We also shouldn't like this, as its a top level thing named "Config" that lives in "app/helpers" 😢 .

This moves that module to `lib` and renames it to `EnvHelper` to avoid possible name clashes.

This is just a first step towards getting our app Rails 6 ready.

**Story card:** [sc-7264](https://app.shortcut.com/simpledotorg/story/7264/get-the-app-booting-in-rails-6)


bin/rails_next
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+export RAILS_NEXT=1
+bash rails "$@"
```
rsanheim, line 3: This _seems_ to do what we would want from a `bin/rails_next` helper to pass through args to a rails 6 binary.

rick (approved):
giphy (19)

PR #3505, opened 2022-02-16
Progress diagnosis reports

claudiovallejo:
**Story card:** [sc-6922](https://app.shortcut.com/simpledotorg/story/6922/create-diagnosis-report-partial)

## Because

This PR creates the Hypertension and Diabetes report template for the new Progress Tab designs.

## This addresses

image
It creates a new `_diagnosis_report.html.erb` partial and creates the bar graph and stacked bar graph interactions in `progress_tab.js`.

## Test instructions

1. Enable the `new_progress_tab` feature flag
2. Go to any facility report and click "Progress reports"
3. Open the developer tools on your browser and select a mobile device (the graph interactions only work with touch interactions, not mouse click events)
4. Click the "Hypertension" link in the progress tab home page
5. Scroll around and test the graph interactions. Share any bugs or issues with Claudio.


timcheadle:
I enabled the `new_progress_tab` feature on the PR review app and got a 500 error:
https://simple-review-pr-3505.herokuapp.com/reports/progress/village-oregeno-village

timcheadle (changes requested):

app/helpers/progress_tab_helper.rb
```diff
@@ -29,4 +29,13 @@ def create_badge_array(current_value)
 
     badges.reverse
   end
+
+  def numbers_to_relative_percentages(array)
```
govindkrjoshi, line 33: We should use https://github.com/simpledotorg/simple-server/blob/master/app/models/reports/percentage.rb#L3 to calculate the percentages, so it's consistent across the app.

app/helpers/progress_tab_helper.rb
```diff
@@ -29,4 +29,13 @@ def create_badge_array(current_value)
...
+    max_value = array.max
+    array.map { |value| ((value.to_f / max_value.to_f) * 100.0).to_i }
+  end
+
+  def set_min_value(value, min_value)
+    value < min_value ? min_value : value
```
govindkrjoshi, line 39: [Style] 
```suggestion
    [value, min_value].max
```

Got confused here a bit, since the function returns the greater of the two values, but is called `set_min_value`

app/helpers/progress_tab_helper.rb
claudiovallejo: Great, thank you! Using the `percentage` method :)

app/helpers/progress_tab_helper.rb
claudiovallejo: Agreed! Much easier to read (and no need to create a helper function). Thanks Govind!

govindkrjoshi (approved):

PR #3507, opened 2022-02-16
Remove legacy progress tab code

rsanheim:
**Story card:** [sc-7371](https://app.shortcut.com/simpledotorg/story/7371/remove-legacy-user-analytics-presenter-cod)

## Because

This removes all the legacy UserAnalyticsPresenter code and uses a new service, FacilityProgressService, which in turn uses the new reporting pipelines for progress tab data.

## This addresses

Making sure performance is acceptable and code is clear for the new follow ups and other data on progress tab.

## Test instructions

Pull down this branch and view the progress tab locally for facilities.


app/components/reports/progress_control_component.html.erb
```diff
@@ -0,0 +1,49 @@
+<div>Add Reports::ProgressControl template here</div>
```
timcheadle, line 1: Is this supposed to be here?

timcheadle (approved):

app/components/reports/progress_control_component.html.erb
rsanheim: thx, I will remove that.

rsanheim (commented):
Notes for @simpledotorg/server-developers

PR #3512, opened 2022-02-17
Change lost to follow up definition to reflect the follow up v2 definition

rsanheim:
This is a POC for the inverse side of `follow_ups_v2` (which has shpped) - lost_to_follow_ups_v2.

Since the new version of follow ups tracks any visit as a follow up, lost to follow up should count patients without _any_ visit over the past 12 months as a lost_to_follow_up, not just a blood pressure.

**Story card:** [sc-7385](https://app.shortcut.com/simpledotorg/story/7385/change-ltfu-definition-in-reporting-pipeline)

## Because

we want lost to follow ups to match follow ups so reports are consistent

## This addresses

updating the definition in the reporting pipeline

**Update**
We've updated this PR to incorporate changes required to make the lost to follow ups definition consistent across the codebase.

This PR also address the following stories: [sc-7387](https://app.shortcut.com/simpledotorg/story/7387/update-any-other-ltfu-definitions-in-the-app-besides-those-in-reporting-pipeline), [sc-7630](https://app.shortcut.com/simpledotorg/story/7630/update-ltfu-definitions-in-the-monthly-district-data-monthly-facility-data-downloadable-spreadsheets), [sc-7386](https://app.shortcut.com/simpledotorg/story/7386/update-all-dashboard-copy-for-ltfu-v2-tooltips-and-any-definitions)


db/migrate/20220217210814_update_reporting_patient_states_to_version_5.rb
```diff
@@ -0,0 +1,21 @@
+class UpdateReportingPatientStatesToVersion5 < ActiveRecord::Migration[5.2]
+  def change
```
rsanheim, line 2: This migration probably needs some work, because we have to ensure any dependent views are dropped and then recreated after the `patient_states` change, and also have to make sure all indexes are recreated.

rsanheim (commented):
notes

db/structure.sql
```diff
@@ -4524,13 +3649,6 @@ CREATE UNIQUE INDEX index_flipper_features_on_key ON public.flipper_features USI
 CREATE UNIQUE INDEX index_flipper_gates_on_feature_key_and_key_and_value ON public.flipper_gates USING btree (feature_key, key, value);
 
 
---
--- Name: index_fs_dimensions_facility_id; Type: INDEX; Schema: public; Owner: -
---
-
-CREATE INDEX index_fs_dimensions_facility_id ON public.reporting_facility_state_dimensions USING btree (facility_id);
-
-
```
PriyangaPKini, line 4533: I am not sure why after running `rails db:migrate` deleted this index.

db/structure.sql
```diff
@@ -4839,13 +3957,6 @@ CREATE INDEX index_protocols_on_deleted_at ON public.protocols USING btree (dele
 CREATE INDEX index_protocols_on_updated_at ON public.protocols USING btree (updated_at);
 
 
---
--- Name: index_qfs_facility_id; Type: INDEX; Schema: public; Owner: -
---
-
-CREATE INDEX index_qfs_facility_id ON public.reporting_quarterly_facility_states USING btree (facility_id);
-
-
```
PriyangaPKini, line 4848: I am not sure why after running rails db:migrate deleted this index.

db/structure.sql
PriyangaPKini: Found out that the new migration overrides the migration to add these deleted indexes. Updated migration file to add those indexes back.

govindkrjoshi:
We've updated this PR to incorporate changes required to make the lost to follow ups definition consistent across the codebase. 

This PR also address the following stories: [sc-7387](https://app.shortcut.com/simpledotorg/story/7387/update-any-other-ltfu-definitions-in-the-app-besides-those-in-reporting-pipeline), [sc-7630](https://app.shortcut.com/simpledotorg/story/7630/update-ltfu-definitions-in-the-monthly-district-data-monthly-facility-data-downloadable-spreadsheets), [sc-7386](https://app.shortcut.com/simpledotorg/story/7386/update-all-dashboard-copy-for-ltfu-v2-tooltips-and-any-definitions)

claudiovallejo (commented):
image

@govindkrjoshi I caught a bug where the "Lost to follow-up" numbers in the "Hypertension coverage" and the "Lost to follow-up" cards don't match.

govindkrjoshi:
@claudiovallejo We've verified the code for both these lost to follow up numbers. The seem to be picked up from the same source. We also verified a numbers for a few states and facilities locally and on the [heroku app](https://simple-review-pr-3512.herokuapp.com/) and the numbers match up.

app/models/patient.rb
```diff
@@ -69,6 +69,8 @@ class Patient < ApplicationRecord
 
   has_many :current_prescription_drugs, -> { where(is_deleted: false).order(created_at: :desc) }, class_name: "PrescriptionDrug"
 
+  has_many :patient_states, class_name: "Reports::PatientState"
```
prabhanshuguptagit, line 72: We could add this scope to `PatientReportable` instead of `Patient`.

app/models/concerns/patient_reportable.rb
```diff
@@ -10,20 +10,15 @@ module PatientReportable
     scope :excluding_dead, -> { where.not(status: :dead) }
 
     scope :ltfu_as_of, ->(time) do
-      last_bp_after = (time - LTFU_TIME).end_of_month
-      joins("LEFT OUTER JOIN latest_blood_pressures_per_patient_per_months
-             ON patients.id = latest_blood_pressures_per_patient_per_months.patient_id
-             AND #{sanitize_sql(["bp_recorded_at > ? AND bp_recorded_at < ?", last_bp_after, time])}")
-        .where("bp_recorded_at IS NULL AND patients.recorded_at <= ?", last_bp_after)
+      left_outer_joins(:patient_states)
+        .merge(Reports::PatientState.where(month_date: time.beginning_of_month, htn_care_state: "lost_to_follow_up"))
```
prabhanshuguptagit, line 14: `PatientState` has a `by_month_date` that can be used here.

prabhanshuguptagit: [suggestion] We could also move the `htn_care_state` clause to a scope to `PatientState`, which would abstract the actual implementation. Would look something like
```
Reports::PatientState.by_month_date(time.beginning_of_month).lost_to_follow_up
````

prabhanshuguptagit (approved):
Neat :shipit: 

[non-blocking] Curious what the cause for mismatch in claudio's screenshot was. Has it been addressed?

govindkrjoshi:
@prabhanshuguptagit We got on a call and  resolved the discrepancy. It could have been something on his local machine. We have also checked this with production data for a few regions (states, districts, blocks, facilities) and these numbers match.

PR #3530, opened 2022-02-18
Shouldn't allow to delete a facility having assigned_patients

PriyangaPKini:
**Story card:** [sc-6032](https://app.shortcut.com/simpledotorg/story/6032/don-t-allow-deleting-a-facility-if-it-has-assigned-patients)

## Because

`Facility#discardable?` doesn't check for assigned patients. 

## This addresses

- Added `assigned_patients` to `Facility#records_preventing_discard`

## Test instructions

- Wrote unit test to ensure that a facility isn't discardable if it has assigned_patients.

### DIFF
--- a/app/models/facility.rb
+++ b/app/models/facility.rb
@@ -219,6 +219,7 @@ class Facility < ApplicationRecord
 
   def records_preventing_discard
     {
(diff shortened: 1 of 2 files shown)


vkrmis (approved):
Looks good

PR #3532, opened 2022-02-21
Data migration to consolidate communication types

vkrmis:
**Story card:** [sc-3585](https://app.shortcut.com/simpledotorg/story/3585/backfill-notifications-from-communications)

## Because

We have redundant communication types from old records:
- `missed_visit_sms_reminder` -> `sms`
- `missed_visit_whatsapp_reminder` -> `whatsapp`

## This addresses
Migrates old records to use the `sms`, `whatsapp` communication types instead. IHCI has ~2M sms communications and 9M whatsapp communications that will be touched.


harimohanraj89 (approved):
Sounds good to me. Could you give us a sense of the number of records we expect to touch in the PR description?

PR #3536, opened 2022-02-22
Treatment outcomes not adding up to 100%

PriyangaPKini:
**Story card:** [sc-5092](https://app.shortcut.com/simpledotorg/story/5092/treatment-outcomes-not-adding-up-to-100-in-ethiopia)

## Because

The treatment outcome graph doesn't add up to 100%

## This addresses

Use rounded_percentage algorithm to generate treatment outcome data


app/models/reports/region_summary_schema.rb
```diff
@@ -245,6 +235,15 @@ def appts_scheduled_rates(entry)
       })
     end
 
+    def visit_details_rates(entry, with_ltfu: false)
```
PriyangaPKini, line 238: I named the aggregate of missed_visit, controlled, uncontrolled, visit_no_bp as `visit_details`.
Is there a better name to it?

PR #3580, opened 2022-03-01
Increase the frequency of progress tab refresh

govindkrjoshi:
**Story card:** [sc-7469](https://app.shortcut.com/simpledotorg/story/7469/issue-with-real-time-data-of-follow-up-entries)

## Because

The progress tabs don't show the most recent data from user activity. 

## This addresses

The current intervals cause confusion in the field since the users cannot see their progress tab refreshing for some time after they have synced their local data changes. This change increases the frequency so that the reports are more recent.

## Test instructions

When you run `bundle exec whenever` you can see the crontab that would be set. The crontab should have the following line:

`0,30 3,4,5,6,7,8,9,10,11,12,13 * * * /bin/bash -l -c 'cd <PATH_TO_SIMPLE_SERVER_DIR>/simple-server && RAILS_ENV=production bundle exec rake db:refresh_daily_follow_ups --silent >> /home/deploy/apps/simple-server/shared/log/cron.log 2>&1'`

This crontab is period based on localtime. We should add 5:30 hours to the value to get the run time in India. 

`0,30 3,4,5,6,7,8,9,10,11,12,13 * * * `  translates to `At minute 0 and 30 past hour 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, and 13.` 

Co-authored-by: Priyanga P Kini <priyangapkini99@gmail.com>


prabhanshuguptagit:
Should we just say `every 30.minutes`? It's simpler to explain even if it's a little extra computation.

prabhanshuguptagit (approved):

govindkrjoshi:
We've kept the times enumerated so we can easily modify if needed.

PR #3588, opened 2022-03-03
Don't schedule a fallback when message delivery fails

vkrmis:
**Story card:** [sc-7503](https://app.shortcut.com/simpledotorg/story/7503/remove-fallback-for-failed-messages)

## Because
We don't use whatsapp for sending reminders at the moment

## This addresses
Removes code that scheduled fallback SMSes for failed whatsapp messages


harimohanraj89 (approved):
This one makes a lot of sense to me. Let's get it out of the way until the dust settles on our decoupling and stuff. We can reintroduce fallbacks later in the right context.

PR #3589, opened 2022-03-03
Refactor TwilioAPIService

prabhanshuguptagit:
**Story card:** [sc-7352](https://app.shortcut.com/simpledotorg/story/7352/make-provider-specific-api-service-classes)

## Because

In order to add BSNL as a second messaging provider for reminders (and subsequently different providers for different countries), we need to decouple the individual implementation of these providers from the selection of the provider. At the end of this exercise we should be able to supply a channel from the country config and be able to swap out providers with ease.

## This addresses

- Introduces the notion of a messaging channel with `Messaging::Channel`. This prescribes the interface that each provider's implementation must adhere to. It also contains some common functionality.
- Refactors `TwilioApiService` into the separate channels it currently abstracts over: Reminder SMSes, OTP SMSes and Whatsapp. 
- Removes logging using `context` in some places where it's redundant. The metrics we care about are already captured via Statsd and reporting it to Sentry using `context` is mostly superfluous.

Todo
- [x] Test sending a message using the new service with test credentials
- [x] Test sending a message using the new service with production credentials


app/jobs/appointment_notification/worker.rb
```diff
@@ -96,21 +91,6 @@ def create_communication(notification, communication_type, response)
...
-  end
-
-  def medication_reminder_sms_sender
-    @medication_reminder_sms_sender ||= medication_reminder_sms_senders.sample
-  end
-
-  def medication_reminder_sms_senders
-    ENV.fetch("TWILIO_APPOINTMENT_REMINDER_NUMBERS", "").split(",").map(&:strip)
-  end
-
```
prabhanshuguptagit, line 113: The Twilio channel implementation takes care of this now.

app/services/messaging/channel.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def initialize
+    @metrics = Metrics.with_object(self)
+  end
+
+  attr_reader :metrics
+
+  def send_message(recipient_number:, message:)
+    # To be implemented by the individual channel.
+  end
```
prabhanshuguptagit, line 16: This is the "common interface" for all sorts of channels.

app/services/messaging/twilio/otp_sms.rb
```diff
@@ -0,0 +1,16 @@
+class Messaging::Twilio::OtpSms < Messaging::Twilio::Api
```
harimohanraj89, line 1: Why a separate service for OTP SMS? More specifically, why are we tying purpose to these classes? I thought we would make these classes strictly _channels_ for any message.

Can we just have an SMS class? Nothing wrong in the OTP SMS's also leveraging an array of senders, right?

lib/tasks/scripts/message_patients.rb
```diff
@@ -39,22 +39,18 @@ def call
...
+        if whatsapp?
+          Messaging::Twilio::Whatsapp
         elsif sms?
-          notification_service.send_sms(recipient_number: phone_number, message: message, context: context)
+          Messaging::Twilio::ReminderSms
         end
-      rescue TwilioApiService::Error
+
+      begin
+        response = messaging_channel.new.send_message(recipient_number: phone_number, message: message)
```
harimohanraj89, line 52: Can we have the channels have a class method that delegates to an instance to make this interface a bit easier?
`messaging_channel.send_message(...)`

app/services/messaging/twilio/reminder_sms.rb
```diff
@@ -0,0 +1,21 @@
+class Messaging::Twilio::ReminderSms < Messaging::Twilio::Api
+  TWILIO_TEST_SMS_NUMBER = "+15005550006"
+
+  def communication_type
+    Communication.communication_types[:sms]
+  end
+
+  def sender_number
```
harimohanraj89, line 8: We _could_ cache this so you don't get unexpectedly changing numbers for a single instance every time you call it. Sampling per instance should be sufficient.

app/services/messaging/twilio/api.rb
```diff
@@ -0,0 +1,57 @@
+class Messaging::Twilio::Api < Messaging::Channel
```
harimohanraj89, line 1: Hmmmmm we've already gone two levels deep with the inheritance on here but there is indeed a LOT of Twilio logic. Did we explore any module-based approaches for mixing in the Twilio-specific logic? Did it look too weird?

harimohanraj89 (changes requested):
High level feedback! Round one 😅

app/services/messaging/twilio/otp_sms.rb
prabhanshuguptagit: The idea is to surface all the options available to send out a message transparently to the caller. Note that the purpose in the channel's name (Reminder, Otp) is not the purpose of the message, but the purpose of the channel itself. On our twilio dashboard we have two "services", in twilio parlance: The `Appointment Reminder Senders` pool and the one number we use for OTPs. The OTP sender exists to send OTPs. The reminder pool exists to send reminders. These channels mirror that setup.

If we wanted to send OTP SMSes via a list of senders, we would do one of the following:
- Use the same senders as the reminder senders. The OTP job should simply switch to calling `ReminderSms` in that case. This is a hack though, and dealing with it will not be the SMS class's concern. That complexity should be on the caller.
- We add more senders dedicated to OTP, in which case `OtpSms` deals with picking one of them.

Combining the OTP and reminder channels also makes the interface non-generic, where the caller would select the effective channel by passing in a flag. The SMS implementation would then try to pick a channel based on a `case` statement. This bundling doesn't buy us anything since all the SMS code is anyways shared nicely in the other implementation.

app/services/messaging/twilio/api.rb
prabhanshuguptagit: We did try a bunch of approaches, however nothing else represents the situation at hand better:
- There are many channels (potentially across different providers). `Channel` is the common interface.
- The Twilio API has some standard behaviour that is shared by all its channels and can be independently tested.
- `OtpSMS`, `Whatsapp` etc are Twilio channels.

This probably also works well because each layer is pretty thin. For ex. the bottom most layers are mostly doing just configuration.

lib/tasks/scripts/message_patients.rb
prabhanshuguptagit: Addressed in [2aa5adc](https://github.com/simpledotorg/simple-server/pull/3589/commits/2aa5adc2d864bf02037d38a6768d5680beec1505)

app/services/messaging/channel.rb
prabhanshuguptagit: Addressed in [d583af2](https://github.com/simpledotorg/simple-server/pull/3589/commits/d583af25b720a8b9dcade4b5e0c92c3474ae3189)

app/services/messaging/twilio/otp_sms.rb
harimohanraj89: I was gonna say the OTP and reminder messages don't have to choose diffferent senders in Twilio, but I remember that one advantage to separation was that any blockage or overage in our reminder messaging (rate limits etc) wouldn't affect OTPs. In that case, yeah, this looks great.

app/services/messaging/twilio/api.rb
harimohanraj89: Neat neat, yeah just wanted to make sure we did our due diligence and landed up here. Very cool.

spec/jobs/appointment_notification/worker_spec.rb
```diff
@@ -18,8 +18,8 @@ def mock_successful_delivery
       response_double = double
       allow(response_double).to receive(:status).and_return("sent")
       allow(response_double).to receive(:sid).and_return("12345")
-      allow_any_instance_of(Messaging::Twilio::Whatsapp).to receive(:send_message).and_return(response_double)
-      allow_any_instance_of(Messaging::Twilio::ReminderSms).to receive(:send_message).and_return(response_double)
+      allow(Messaging::Twilio::Whatsapp).to receive(:send_message).and_return(response_double)
```
harimohanraj89, line 21: Oof, this shortening is :chef_kiss:

app/services/messaging/channel.rb
```diff
@@ -1,5 +1,5 @@
 class Messaging::Channel
-  # On Error Handling: Any messaging channel implementation should raise
+  # Note on Error Handling: Any messaging channel implementation should raise
```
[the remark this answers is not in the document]
prabhanshuguptagit: <3

prabhanshuguptagit:
Yes 😄 The idea is to shift the responsibility of creating communications and deliverables to the deliverables. [An example here](https://github.com/simpledotorg/simple-server/pull/3582/files#diff-330831ad24d6aa07d3874901361e2b0912ddab6844867a0d199c4d42ae9897e1R27)

PR #3591, opened 2022-03-04
Exclude inactive facilities

govindkrjoshi:
**Story card:** [sc-7482](https://app.shortcut.com/simpledotorg/story/7482/exclude-inactive-facilities-from-csv-download)

## Because

Exclude inactive facilities from CSV download
- In the "district sheet" change “Facilities implementing IHCI” from total facility count to exclude inactive facilities. 
- In the "facility sheet" exclude inactive facilities

## This addresses

Add a scope `active` to the facility model which takes a `month_date` and returns only the active facilities in that `month_date`

## Test instructions

- Make sure to enable `monthly_district_report` in Flipper
- Download "Monthly district report" on the Reports page.  You should now only be able to see the active facilities in the "district sheet" and the "facility sheet"


app/models/facility.rb
```diff
@@ -64,6 +64,13 @@ class Facility < ApplicationRecord
       .select("facilities.*, reporting_facilities.*")
   }
 
+  scope :active, ->(month_date: Date.today) {
+    joins("INNER JOIN reporting_facility_states on reporting_facility_states.facility_id = facilities.id")
+      .where("reporting_facility_states.month_date" => month_date.at_beginning_of_month)
+      .where("cumulative_registrations > 0 OR cumulative_assigned_patients > 0 OR monthly_follow_ups > 0")
+      .distinct
+  }
```
harimohanraj89, line 72: The story references a definition of "inactive facilities" in the Reports Details section? I couldn't find that but what is the story referring to? Does this definition match (can you link me to it)? If so, is there room to re-use the same scope for both touchpoints? Keeping our indicators expressed in a _single_ place in our codebase is key to consistency.

harimohanraj89 (changes requested):
One big thought from me.

app/models/facility.rb
govindkrjoshi: @harimohanraj89 This definition is used in two other places in codebase: 
- https://github.com/simpledotorg/simple-server/blob/master/app/models/reports/region_summary_schema.rb#L240
- https://github.com/simpledotorg/simple-server/blob/master/app/controllers/my_facilities/drug_stocks_controller.rb#L130

I think we can refactor the drug stocks controller to use this schema, but `RegionSummarySchema` is just looking up in `reporting_facility_states`.  

Would suggest we break this scope down into smaller scopes and compose them?

govindkrjoshi: Fwiw, we added this scope with the same intention that in future, the active facilities can be looked up using a common definition :)

PriyangaPKini: We are using this "inactive facilities" definition in couple of places:
- [region_summary_schema.rb](https://github.com/simpledotorg/simple-server/blob/3d07f548c7afd67b8dd660c53f89ec4f55f4accb/app/models/reports/region_summary_schema.rb#L240). 
- [drug_stocks_controller.rb](https://github.com/simpledotorg/simple-server/blob/3d07f548c7afd67b8dd660c53f89ec4f55f4accb/app/controllers/my_facilities/drug_stocks_controller.rb#L128)

Need to think of a way to keep this in a single place.

harimohanraj89: Could we take a stab at consolidating that in this PR?

govindkrjoshi: @harimohanraj89 We've tried to consolidate this logic in scopes of `FacilityState` and `Facility`. Could you take a look?

Let us know if you have something else in mind.

app/models/reports/facility_state.rb
```diff
@@ -20,5 +20,9 @@ def self.region_id_field(region)
     def period
       @period ||= Period.month(month_date)
     end
+
+    scope :with_registered_assigned_or_follow_up_patients, -> {
```
harimohanraj89, line 24: Consider `with_patients` or `with_active_patients`? ie, let's name the "what", not the "how".

harimohanraj89 (approved):
One naming suggestion, but looks great!

PR #3593, opened 2022-03-04
Use new followup definition to check if a patient is lftu

PriyangaPKini:
**Story card:** [sc-3842](https://app.shortcut.com/simpledotorg/story/3824/change-patient-line-list-csvs-to-track-visits)

## Because

`lftu?` function  in the `materialized_patient_summary` was using an older follow-up definition

## This addresses

`ltfu` function now includes checks for if
- lastest blood pressure date < ltfu time 
- latest blood sugar date < ltfu time
- latest prescription drugs date < ltfu time
- latest appointment scheduled date < ltfu time

## Test instructions

Download Patient Line List CSV and verify that "patient status" column for each patient is marked correctly


spec/models/materialized_patient_summary_spec.rb
```diff
@@ -265,5 +265,37 @@ def refresh_view
 
       expect(described_class.find(not_ltfu_patient.id).ltfu?).to eq false
     end
+
+    it "is false if patient recorded a BP w/i the LTFU time" do
```
harimohanraj89, line 269: We didn't have a negative test for this before? Nice addition 👍

spec/models/materialized_patient_summary_spec.rb
```diff
@@ -265,5 +265,37 @@ def refresh_view
...
+      not_ltfu_patient = create(:patient, recorded_at: 2.years.ago)
+      create(:blood_pressure, patient: not_ltfu_patient)
```
harimohanraj89, line 271: Let's be very deliberate with the dates we choose to record the BPs. If the cutoff is 365 days, we should have a BP recorded at 364 days and check that it's false. We should also have a BP recorded at 365 days and assert that it's true. (Same for the other associated records)

harimohanraj89 (changes requested):

harimohanraj89:
My bad, I didn't see that the PR was still in draft. Feel free to ignore my feedback if you were already working on it.

spec/models/materialized_patient_summary_spec.rb
PriyangaPKini: Thanks, @harimohanraj89. It's an important point. Will tweak the dates accordingly.

spec/models/materialized_patient_summary_spec.rb
```diff
@@ -252,18 +252,90 @@ def refresh_view
...
+      it "is true if patient recorded a BP not w/i the LTFU time" do
+        ltfu_patient = create(:patient, recorded_at: 365.days.ago)
+        create(:blood_pressure, patient: ltfu_patient, device_created_at: 365.days.ago)
+        refresh_view
+
+        expect(described_class.find(ltfu_patient.id).ltfu?).to eq true
+      end
+    end
+
+    context "patient records a BS" do
```
harimohanraj89, line 287: BP is a really well established abbreviation, very easily understood, but can we expand the others? BS, PD, etc?

We can keep LTFU (it's well established in our codebase and would cause a _lot_ of extra text if we expanded it).

spec/models/materialized_patient_summary_spec.rb
```diff
@@ -252,18 +252,90 @@ def refresh_view
...
+      it "is false if patient recorded a BS w/i the LTFU time" do
```
harimohanraj89, line 288: Along the same vein, let's just use "within". It's not too much longer and very clear.

harimohanraj89, line 288: Along the same vein, let's just use "within". It's not too much longer and very clear.

spec/models/materialized_patient_summary_spec.rb
```diff
@@ -252,18 +252,90 @@ def refresh_view
...
+        not_ltfu_patient = create(:patient, recorded_at: 365.days.ago)
+        create(:blood_sugar, patient: not_ltfu_patient, device_created_at: 364.days.ago)
```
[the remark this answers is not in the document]
harimohanraj89: Ah I see, the blood sugar factory automatically sets the `recorded_at` to match the `device_created_at`. Let's set the `recorded_at` explicitly here anyway, since that makes it much more evident how our underlying query is implemented.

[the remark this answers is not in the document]
harimohanraj89: Ah I see, the blood sugar factory automatically sets the `recorded_at` to match the `device_created_at`. Let's set the `recorded_at` explicitly here anyway, since that makes it much more evident how our underlying query is implemented.

spec/models/materialized_patient_summary_spec.rb
```diff
@@ -252,18 +252,90 @@ def refresh_view
...
+
+        expect(described_class.find(not_ltfu_patient.id).ltfu?).to eq false
+      end
+
+      it "is true if patient recorded a BS not w/i the LTFU time" do
+        ltfu_patient = create(:patient, recorded_at: 365.days.ago)
+        create(:blood_sugar, patient: ltfu_patient, device_created_at: 365.days.ago)
+        refresh_view
+
+        expect(described_class.find(ltfu_patient.id).ltfu?).to eq true
```
harimohanraj89, line 301: Some rspec syntax sugar you could use. You could write these specs as:

```ruby
expect(described_class.find(ltfu_patient.id)).to be_ltfu

# or

expect(described_class.find(ltfu_patient.id)).not_to be_ltfu
```

harimohanraj89:
Another round of lower-level feedback provided! 👍

harimohanraj89:
Another round of lower-level feedback provided! 👍

spec/models/materialized_patient_summary_spec.rb
PriyangaPKini: Sure,  will do

spec/models/materialized_patient_summary_spec.rb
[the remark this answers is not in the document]
PriyangaPKini: Thanks, @harimohanraj89

spec/models/materialized_patient_summary_spec.rb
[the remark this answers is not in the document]
PriyangaPKini: Right, let's use `recorded_at` instead

harimohanraj89 (approved):

PR #3595, opened 2022-03-07
Help screen not properly showing/hiding subviews

claudiovallejo:
**Story card:** [sc-7526](https://app.shortcut.com/simpledotorg/story/7526/help-screen-text-cutting-off-in-common-questions)

## Because

When a user clicks one of the links to open a subpage, the content of the "Help" home screen still appears at the end (it should be hidden).

## This addresses

**Before**
image
_The "Help" section subviews show content from the home page (the "App for your patients" card shouldn't be displayed because it's part of the home page)._

**After**
image
_The "Help" section subviews don't show content from the home page._

1. Deleted the `closeWindow` and `openWindow` JS functions. Replaced with a single function, `goToPage`
2. Set the right element IDs in the HTML to open/close the home page and subviews on link click.

## Test instructions

1. Go to `localhost:3000/api/v3/help.html` on your local
2. Click the first "Help" page link, "Videos"
3. Scroll down to the end of the page and make sure only the "Videos" content is displayed. The "Help" home page content shouldn't display
4. Click the back arrow at the top of the page. Scroll down to the bottom of the home page and make sure only the home page content is displaying
5. Do this for every link

### DIFF
--- a/app/assets/javascripts/standalone/help.js
+++ b/app/assets/javascripts/standalone/help.js
@@ -1,28 +1,16 @@
 // Hides and shows each page of Help
 
-function openWindow(id, parentId) {
-   const element = document.getElementById(id);
-   element.style.display = 'block';
+function goToPage(startPageId, endPageId) {
+  const startPage = document.getElementById(startPageId);
+  startPage.style.display = "none";
+  startPage.style.height = "0";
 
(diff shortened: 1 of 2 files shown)


prabhanshuguptagit (approved):
Slick.

PR #3601, opened 2022-03-09
Group medications by name on dashboard

dburka:
**Story card:** No card

## Because

It's confusing that medications were listed over and over under the same name in the Dashboard, even though they're shown logically under a single name in the Android app.

NOTE: I also moved the time period to the bottom of the page because it's very very rarely changed.

## This addresses

Makes it easier to see how many options a user will see in the Android app when they tap on a drug name.

## Test instructions

Add or edit common drug names under a protocol in the Dashboard.


harimohanraj89 (approved):
This looks really great. We're not showing the "stock tracked?" on the list page, is that ok?

harimohanraj89:
Specs are green. Merge whenever you're ready :shipit:

PR #3603, opened 2022-03-09
Show a region's children in the Details table

timcheadle:
**Story card:** [sc-6315](https://app.shortcut.com/simpledotorg/story/6315/make-patients-and-registrations-table-in-report-details-show-district-rows-in-state-report)

## Because

The dashboard `Details` report shows a table of registrations and follow ups. Today, it lists every *facility*, so at the state level it shows hundreds of rows, making it useless.

## This addresses

This changes the table to show the region's immediate children instead of facilities. States show districts, districts show blocks, and blocks show facilities.


claudiovallejo (approved):
🔥🔥🔥

PR #3606, opened 2022-03-10
Task to refresh region names

harimohanraj89:
**Story card:** [sc-7612](https://app.shortcut.com/simpledotorg/story/7612/refresh-region-names-of-seed-data-in-india-demo)

## Because

We want to make the names of places look Indian in India demo. We had updated the "fake_names" csv, but need to effect that change.

## This PR

Introduces a one-off script to execute this rename in Sandbox first (to do a prod-like test) and then in India demo.

Of course, if we ever purged and reseeded any other demo environment, the names would become Indian, but we can deal with that later.


lib/tasks/refresh_region_names.rb
```diff
@@ -0,0 +1,56 @@
+class RefreshRegionNames
+  def self.call
+    abort "Cannot run in production!" if SimpleServer.env.production?
```
timcheadle, line 3: I believe demo still runs as `RAILS_ENV = production`. You may want to check `SIMPLE_SERVER_ENV` instead.

There's this too:
`app/helpers/simple_server_env_helper.rb`

harimohanraj89: Yup, that's why I'm checking `SimpleServer.env` instead of `Rails.env` 👍

lib/tasks/refresh_region_names.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    ActiveRecord::Base.transaction do
+      # States
+      states = Region.state_regions
+      new_state_names = Seed::FakeNames.instance.states.sample(states.count)
+
+      states.zip(new_state_names).each do |state, new_name|
+        state.slug = nil
+        state.reparent_to = state.organization_region
+        state.update!(name: new_name)
```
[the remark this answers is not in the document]
harimohanraj89: Yup! Setting the slug to `nil` above triggers the model to regenerate the slug on save. This is how FriendlyId recommends you regenerate slugs.

PR #3607, opened 2022-03-10
Fixed headers and simplified overdue call results code

timcheadle:
**Story card:** none

## Because

#3603 fixed the `Details` tables to show child regions instead of facilities, but missed fixing the table headers and the overdue call results table.

## This addresses

This fixes those headers and simplifies the overdue call results table to use the same behavior

## Test instructions

1. Go to the PR review Heroku app
2. Enable `show_call_results` in Flipper
3. Look at a `Details` report for a state; it should show districts for both tables

### DIFF
--- a/app/components/registrations_and_follow_ups_component.html.erb
+++ b/app/components/registrations_and_follow_ups_component.html.erb
@@ -39,7 +39,7 @@
         </tr>
         <tr class="sorts" data-sort-method="thead">
           <th class="row-label sort-label sort-label-small ta-center" data-sort-default>
-            Facilities
+            <%= region.child_region_type.capitalize %>
           </th>
           <th class="row-label sort-label sort-label-small ta-center" data-sort-method="number">
             Total assigned patients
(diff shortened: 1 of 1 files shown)


app/components/registrations_and_follow_ups_component.html.erb
```diff
@@ -39,7 +39,7 @@
         </tr>
         <tr class="sorts" data-sort-method="thead">
           <th class="row-label sort-label sort-label-small ta-center" data-sort-default>
-            Facilities
+            <%= region.child_region_type.capitalize %>
```
harimohanraj89, line 42: Will this show the pluralized word?

app/components/registrations_and_follow_ups_component.html.erb
```diff
@@ -159,28 +159,12 @@
...
-              <% range.each do |period| %>
-                <td class="ta-center">
-                  <%= number_or_dash_with_delimiter repository.monthly_overdue_calls[slug][period] %>
-                </td>
-              <% end %>
-            </tr>
-          <% end %>
-        <% else %>
-        <% region.facility_regions.each do |resource| %>
+        <% region.reportable_children.each do |resource| %>
```
harimohanraj89, line 162: Very nice

harimohanraj89 (approved):
Tiny question from me, but the code change looks clean.

app/components/registrations_and_follow_ups_component.html.erb
timcheadle: I left it singular to match the behavior on the `Overview` report

PR #3612, opened 2022-03-15
[P1] Daily followups bug

harimohanraj89:
**Story card:** [sc-7641](https://app.shortcut.com/simpledotorg/story/7641/progress-tab-follow-ups-shouldn-t-include-fresh-registrations)

Here’s the timeline I’m seeing:

* We crafted a daily followups reporting view. This view excluded BPs that were recorded in the same month that the patient was registered - Check out the PR.
* We then realized that this month exclusion was wrong, and reverted the exclusion in another PR . Notice that the month exclusion is gone.
* The problem is: When we removed this exclusion, we didn’t introduce a “daily” exclusion. In other words, BPs or other visit activity that happened on the same day as the patient’s registration date should NOT be counted as a follow-up visit, but our view started counting them as followups.
* Proposed solution: Add a daily exclusion (similar to the monthly exclusion that existed) to our daily followups reporting view.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -1380,7 +1380,7 @@ CREATE MATERIALIZED VIEW public.reporting_daily_follow_ups AS
             bp.recorded_at AS visited_at,
             (date_part('doy'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('UTC'::text, bp.recorded_at))))::integer AS day_of_year
            FROM (public.patients p
-             JOIN public.blood_pressures bp ON ((p.id = bp.patient_id)))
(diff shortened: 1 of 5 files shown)


db/structure.sql
```diff
@@ -1416,7 +1416,7 @@ CREATE MATERIALIZED VIEW public.reporting_daily_follow_ups AS
             app.device_created_at AS visited_at,
             (date_part('doy'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('UTC'::text, app.device_created_at))))::integer AS day_of_year
            FROM (public.patients p
-             JOIN public.appointments app ON ((p.id = app.patient_id)))
+             JOIN public.appointments app ON (((p.id = app.patient_id) AND (date_trunc('month'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('UTC'::text, app.device_created_at))) > date_trunc('month'::text, timezone(( SELECT current_setting('TIMEZONE'::text) AS current_setting), timezone('UTC'::text, p.recorded_at)))))))
```
vkrmis, line 1419: Is `date_trunc('month'...` here instead of `date_trunc('day'...` here deliberate?

govindkrjoshi: Good catch

harimohanraj89: It is not, good catch.

vkrmis (approved):
:shipit:

PR #3614, opened 2022-03-15
Add "Achievements" card to Progress tab home page

claudiovallejo:
**Story card:** [sc-7577](https://app.shortcut.com/simpledotorg/story/7577/new-achievements-card)

## Because

We're connecting updating "Achievements" section UI to show a list of the five most recent badges for follow up patients and registrations.

## This addresses
**Before**
image

**After**
image
* Added new "Achievements" card
* Deleted the old achievements and footer sections
* Reduced the font size for the gender/diagnosis `select` in the "Registrations" and "Follow-ups" cards [Praveen reported](https://simpledotorg.slack.com/archives/C013HS4N9U3/p1647318517496769) on 15-Mar-2022
* Refactored the `ProgressTabHelper` module
* Deleted the `FacilityProgressAchievementService`

## Test instructions

1. Open your local and go to the Report page of any facility
2. Click the "Progress reports" link
3. Make note of the "All-time" registration and follow-up values displayed in the respective cards
4. Scroll to the end of the page and check the "Achievements" card
5. The follow-up and registration values should match the values displayed in the cards above
6. There should be 5 badges for follow-ups and 5 badges for registrations
7. If there are `0` follow-ups or registrations, all 5 badges should be outlined in gray
8. The last badge should always be outlined in gray
9. If the follow-ups/registrations value is equal to the 4th badge (second-to-last badge) the badge should be colored in (For example, if there are 250 follow-ups, the 4th badge should read "250" and it should be colored in).


app/views/api/v3/analytics/user_analytics/show.html.erb
```diff
@@ -89,66 +90,57 @@
...
-
         <%= render(Reports::ProgressControlComponent.new(@service)) %>
-
         <div class="card">
           <a href="#" onclick="openWindow('progress-cohort', 'progress-start'); return false" class="nav-next">
             Cohort report
           </a>
         </div>
-
-        <% if @achievements.show? %>
```
vkrmis, line 101: Do we want to preserve an equivalent check to this one?

claudiovallejo: @vkrmis that's a good question. I believe we shouldn't. The achievements card should always be displayed even if the facility is brand new. The old `.show?` check tested to see if the facility had `<10` follow-ups to see if the section was displayed.

PR #3616, opened 2022-03-16
Refactor Appointment Notification Worker

prabhanshuguptagit:
**Story card:** [sc-7350](https://app.shortcut.com/simpledotorg/story/7350/refactor-appointment-notification-worker)

## Because

Sending out experiment notifications through `AppointmentNotification::Worker` is intertwined with Twilio specific API concerns at the moment. We want to break apart the worker so notifications can be dispatched to channels specified in a per-country config. [The PRD](https://docs.google.com/document/d/13zTcmigEWOqwdU_RbzmQjD5s5R8hE6-rj5dLO2str7E/edit#) has an exhaustive list of the worker's current responsiblilities that make this hard.

## This addresses
- At a high level: pushes out responsibilities from the worker.
- Adds a `NotificationDispatchService` that takes a notification and a messaging channel and dispatches it
- Picks up a messaging service from country config dynamically.
- Moves communication creation to messaging channels and deliverable classes, instead of doing `Communication#create_with_twilio_details!`.

---
- [x] Test OtpSms, ReminderSms and Whatsapp on Sandbox with real account


app/services/messaging/channel.rb
```diff
@@ -9,31 +9,35 @@ def initialize
     @metrics = Metrics.with_object(self)
   end
 
+  delegate :transaction, to: ActiveRecord::Base
+
   attr_reader :metrics
 
+  # The channel implementation is responsible for creating a Communication
+  # and delivery details. This should return the communication object that was created.
```
prabhanshuguptagit, line 17: This PR moves the responsibility of recording the communication and the detailable to the messaging channels. There's two reasons:
- Since the messaging channels can return different responses - Twilio returns an object, BSNL will return some JSON, it's hard to work with them at the `NotificationDispatch` layer. The channels should ideally return the same shape of result that also hides the implementation. A `communication` abstracts this very well.
-  Any other consumer of this API in our codebase would also probably need to record communications.

Side effect: We will start recording communications for OTPs. We _could_ override this behaviour in the OTP channel class but there isn't a strong reason to do it. We send ~3k OTPs in a month in IHCI, so it should be fine to record communications and detailables for them.

app/jobs/appointment_notification/worker.rb
```diff
@@ -1,117 +1,35 @@
 class AppointmentNotification::Worker
   include Sidekiq::Worker
+  include Memery
+  APPOINTMENT_REMINDERS_CHANNEL = CountryConfig.current[:appointment_reminders_channel].constantize
```
harimohanraj89, line 4: As we discussed, let's make the worker _even_ thinner, and push this down to the dispatch service.

app/models/notification.rb
```diff
@@ -35,6 +33,11 @@ def self.logger(extra_fields = {})
 
   scope :due_today, -> { where(remind_on: Date.current, status: [:pending]) }
 
+  def record_communication(communication)
```
harimohanraj89, line 36: It seems a bit strange that this responsibility lives on the notification model. The channel creates the detail and communication, right? Can the dispatch service orchestrate this "linkage"?

harimohanraj89, line 36: It seems a bit strange that this responsibility lives on the notification model. The channel creates the detail and communication, right? Can the dispatch service orchestrate this "linkage"?

app/services/messaging/channel.rb
```diff
@@ -9,31 +9,38 @@ def initialize
     @metrics = Metrics.with_object(self)
   end
 
+  delegate :transaction, to: ActiveRecord::Base
```
harimohanraj89, line 12: Seems a bit overkill to set up this delegation for one usage. Let's wait until there's more. And even then, should we set up a `with_transaction` method or something, instead of delegation? Delegation tends to make tracing codepaths harder through a codebase, I think.

A valid use case would be for wrapper classes, but this channel definitely doesn't "wrap" activerecord base :)

harimohanraj89, line 12: Seems a bit overkill to set up this delegation for one usage. Let's wait until there's more. And even then, should we set up a `with_transaction` method or something, instead of delegation? Delegation tends to make tracing codepaths harder through a codebase, I think.

app/services/messaging/twilio/api.rb
```diff
@@ -2,13 +2,12 @@ class Messaging::Twilio::Api < Messaging::Channel
   include Rails.application.routes.url_helpers
   include Memery
 
-  def send_message(recipient_number:, message:)
+  def send_message(recipient_number:, message:, &with_communication_do)
```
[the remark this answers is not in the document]
harimohanraj89: Consider `after_send` instead of `with_communication_do`. I know that means you have to "know" that the `communication` is exposed in the block, but I think the more conventional name is worth it.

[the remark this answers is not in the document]
harimohanraj89: Consider `after_send` instead of `with_communication_do`. I know that means you have to "know" that the `communication` is exposed in the block, but I think the more conventional name is worth it.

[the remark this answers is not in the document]
harimohanraj89: In fact, it doesn't change the caller interface at all, so I'd strongly recommend it 😅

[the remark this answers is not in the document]
harimohanraj89: In fact, it doesn't change the caller interface at all, so I'd strongly recommend it 😅

spec/services/notification_dispatch_service_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    allow(messaging_channel).to receive(:send_message) do |_, &block|
+      block.call(communication)
+    end
+
+    described_class.call(notification, messaging_channel)
+
+    expect(notification.reload.status).to eq("sent")
+    expect(notification.communications.first).to eq(communication)
+  end
```
harimohanraj89, line 62: Yeah these specs are critical 👏

harimohanraj89, line 62: Yeah these specs are critical 👏

harimohanraj89 (approved):
My review below. This refactor mostly looks solid. I've approved so you're not blocked after addressing comments.

Bigger picture, as we discussed, let's keep an eye on the moving pieces as we go. Notifications, comms, details, dispatch service, notification service. Perhaps we can sketch out this diagram to make sure it's not crazy.

app/models/notification.rb
prabhanshuguptagit: This was moved here since:

Notification is already aware of communications (and that it can have many).
Attaching a communication and changing the status to sent is an atomic operation (not in the transactional meaning). We don't want a notification to have a communication but not be marked sent. This is something the model should handle and is too much responsibility for the service.

app/services/messaging/twilio/api.rb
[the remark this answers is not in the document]
prabhanshuguptagit: The `after_send` convention coming from rails callbacks? I ended up on the `_do` since it hints at a ruby block that does something to a `communication` but `after_send` also works if it makes sense in rails' context.

prabhanshuguptagit:
@harimohanraj89 Addressed feedback. Also, inspired from our conversation to keep a check on moving pieces I've renamed a service in [99f2ce7](https://github.com/simpledotorg/simple-server/pull/3616/commits/99f2ce73303a1a1322ae778a88d9cf2327621349), hope it makes sense.

PR #3619, opened 2022-03-17
Add an abstract delivery_detail class

vkrmis:
**Story card:** [sc-7663](https://app.shortcut.com/simpledotorg/story/7663/abstract-delivery-details)

## Because

We need to handle delivery details on a per SMS service provider basis since they can differ significantly

## This addresses

Adds an abstract `DeliveryDetail` class


app/models/twilio_sms_delivery_detail.rb
```diff
@@ -1,4 +1,4 @@
-class TwilioSmsDeliveryDetail < ApplicationRecord
+class TwilioSmsDeliveryDetail < DeliveryDetail
   has_one :communication, as: :detailable
```
harimohanraj89, line 2: Does this go in the abstract class?

harimohanraj89 (approved):
One question - non blocking.

Looks like we've already set up a polymorphic `detailable` association to the Communication, eh? Makes this a nice continuation of that thought, without having to update the touchpoints where we access the detailable. Nice.

vkrmis:
> One question - non blocking.
> 
> Looks like we've already set up a polymorphic `detailable` association to the Communication, eh? Makes this a nice continuation of that thought, without having to update the touchpoints where we access the detailable. Nice.

Yup, pretty much. Also allows each delivery detail type to define their own criteria for success, failure, in_progress.

app/models/twilio_sms_delivery_detail.rb
vkrmis: moved 🙏

PR #3620, opened 2022-03-17
Add district reports template

dburka:
**Story card:** No story

## Because

Mahima's new feature needs a link to the blank Excel template for CVHOs.

## This addresses

Launch of Excel Reports

## Test instructions

Click link under Settings > Resources.
Make sure you can download the Excel.
Should only display in India.


claudiovallejo (approved):
Looks good! Verified it only displays in India 🇮🇳

PR #3625, opened 2022-03-21
Drop unused columns from Communication

prabhanshuguptagit:
**Story card:** [sc-7667](https://app.shortcut.com/simpledotorg/story/7667/drop-unused-columns-from-communication)

## Because

`Communication` has some extra attributes that have not been in use and are redundant. These cause noise while interacting with the model as well as make the responsibilities of a communication unclear. We can drop the following:
- `device_` timestamps. `Communication` was originally meant to be used as a sync resource but the modelling has evolved since then to support a `Notification` (which can have multiple communications). If we were to allow to send reminders to patients from the app, we should use the notification model.
- `user_id`. This was partially addressed this in https://github.com/simpledotorg/simple-server/pull/642. All the communications in production either have a bot user ID or a `nil` user ID.
- `appointment_id`. The original implementation allowed associating appointments to a communication. We have since moved to tying appointments to notifications and communications only represent a message delivery. They should be well separated from appointments. The `appointment_id` is also redundant since the notification has it already.

[PRD](https://docs.google.com/document/d/13zTcmigEWOqwdU_RbzmQjD5s5R8hE6-rj5dLO2str7E/edit)
## This addresses

Drops the columns and the associated code.


app/models/communication.rb
```diff
@@ -28,38 +26,25 @@ class Communication < ApplicationRecord
...
   DEFAULT_MESSAGING_END_HOUR = 16
 
-  validates :device_created_at, presence: true
-  validates :device_updated_at, presence: true
-
   scope :with_delivery_detail, -> {
     joins("inner join twilio_sms_delivery_details delivery_detail on delivery_detail.id = communications.detailable_id")
   }
 
-  def self.latest_by_type(communication_type)
```
harimohanraj89, line 44: Cool, we don't use this method anywhere?

prabhanshuguptagit:
> have all the appointment_ids been moved to the appropriate notifications

No, we decided to not backfill notifications for the older communications since there's a large number of them (100M) and it's unlikely we'll ever need to report on them: [story card](https://app.shortcut.com/simpledotorg/story/3585/backfill-notifications-from-communications#activity-7392). The last communication without a notification is from Apr 2021. 

We were planning on backing up communication IDs <> appointment IDs into s3 before merging. That would leave us a way to look this information up later if we want.

Although since we are dropping `appointment_id` from the old communications, there will be no link left for those rows with the `Patient` model. They'll just be a record of _someone_ being communicated at _such and such_ time. We can probably backup and drop those records entirely in that case. wdyt? We can punt on the `appointment_id` column drop for later in that case, after the BSNL integration is done.

harimohanraj89:
Oh I see, all we'll have is a flat list, with no way to analyze patient-level, facility-level info, etc.

Yeah let's punt on the appt_id and ship the rest 🚀

prabhanshuguptagit:
[Follow up story](https://app.shortcut.com/simpledotorg/story/7670/drop-appointment-id-from-communications?)

PR #3629, opened 2022-03-22
Add a rake task to refresh BSNL JWTs

vkrmis:
**Story card:** [sc-7669](https://app.shortcut.com/simpledotorg/story/7669/add-a-rake-task-to-refresh-bsnl-jwt-tokens)

## Because
We want to fetch fresh JWT tokens automatically using the BSNL API
## This addresses
Adds a rake task to fetch and update BSNL JWTs
Adds a `Credential` model and db table to store BSNL JWTs. This is needed since we'll want to be able to easily update the JWT in place. We considered alternative approaches like:
- Automatically updating the deployment repo when a new JWT is fetched
- Storing JWTs in an external credential service

We settled on just storing these in the DB instead since the other approaches add a lot of overhead


db/data/20220322135814_create_bsnl_credential.rb
```diff
@@ -0,0 +1,17 @@
...
+    rake.init
+    rake.add_import "lib/tasks/refresh_bsnl_sms_jwt.rake"
+    rake.load_rakefile
+
+    Credential.create(name: "BSNL_SMS_JWT", value: "")
+    rake["bsnl:refresh_sms_jwt"].invoke
+  end
+
+  def down
+    Credential.find("BSNL_SMS_JWT").delete
```
harimohanraj89, line 15: `&.delete`

app/models/credential.rb
```diff
@@ -0,0 +1,3 @@
+class Credential < ApplicationRecord
```
harimohanraj89, line 1: I'm thinking ahead since modifying database tables later is a little annoying.

Should we lean into the idea of self-serviceability and call this `Configuration`? It could contain credentials, and it could also contain nonsensitive things like:
* Twilio username
* Sync radius
* Reports refresh rate

db/data/20220322135814_create_bsnl_credential.rb
```diff
@@ -0,0 +1,17 @@
+class CreateBsnlCredential < ActiveRecord::Migration[5.2]
+  require "rake"
+
+  def up
+    rake = Rake.application
+    rake.init
+    rake.add_import "lib/tasks/refresh_bsnl_sms_jwt.rake"
+    rake.load_rakefile
+
+    Credential.create(name: "BSNL_SMS_JWT", value: "")
```
harimohanraj89, line 10: Since these aren't env vars, I don't think we need to be so loud. `bsnl_sms_jwt` reads just fine to me :)

db/structure.sql
```diff
@@ -351,6 +351,19 @@ CREATE TABLE public.communications (
 );
 
 
+--
+-- Name: credentials; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.credentials (
+    name character varying NOT NULL,
```
harimohanraj89, line 359: I don't see any artifact in the `structure.sql` that indicates that `name` is a primary key, or that it's uniqueness is enforced.

harimohanraj89 (changes requested):
Love this idea of storing config in the DB. My feedback mostly revolves around that.

app/models/credential.rb
vkrmis: Makes sense. I'll alter this to be more generically usable.

db/structure.sql
harimohanraj89: PS why make the name the unique id? Any reason to deviate from the default?

vkrmis: I'm thinking of this table like a key value store, given we'll primary access data using the name. Could leave a primary uuid key in and not use it to access data I guess.

harimohanraj89: Yeah that seems more "boring", less likely to have any gotchas. We can always introduce a `fetch` helper that'll give us a convenient interface to fetch these values.

app/models/configuration.rb
```diff
@@ -0,0 +1,8 @@
+class Configuration < ApplicationRecord
+  validates :name, presence: true
+  validates :value, presence: true
+
+  def self.fetch(name)
+    Configuration.find_by(name: name).value
```
harimohanraj89, line 6: We can drop the `Configuration.`

harimohanraj89: We can make this safer - return `nil` if the config record doesn't exist.

We can make a `fetch!` as well if we need it later.

lib/tasks/scripts/bsnl.rb
```diff
@@ -0,0 +1,28 @@
...
+    http.use_ssl = true
+    request = Net::HTTP::Post.new("/api/Create_New_API_Token", "Content-Type" => "application/json")
+    request.body = {Service_Id: service_id,
+                    Username: username,
+                    Password: password,
+                    Token_Id: token_id}.to_json
+    response = http.request(request)
+
+    if response.is_a?(Net::HTTPSuccess)
+      jwt = response.body.delete_prefix('"').delete_suffix('"')
```
harimohanraj89, line 24: UGH 🙄

lib/tasks/scripts/bsnl.rb
```diff
@@ -22,9 +22,8 @@ def refresh_sms_jwt
 
     if response.is_a?(Net::HTTPSuccess)
       jwt = response.body.delete_prefix('"').delete_suffix('"')
-      Configuration.find_or_create_by(name: "bsnl_sms_jwt") do |config|
-        config.value = jwt
-      end
+      config = Configuration.find_or_create_by(name: "bsnl_sms_jwt")
+      config.update(value: jwt)
```
harimohanraj89, line 26: Consider `update!` so we know clearly if this fails.

PR #3643, opened 2022-03-27
Added font-sizes to media queries and fixed tabbed nav on mobile

dburka:
Added font-sizes to media queries so you can choose different sizes for mobile. Edited all tabbed navigations to size down to 16px for mobile.

**Story card:** NONE

## Because

Tabbed nav in My Facilities, Drug Stock, and Reports wraps terribly on mobile. Also, it was not possible to select smaller font sizes for mobile, so I added this to the css.

## Test instructions

Check the tabbed navigations look good. They do!

### DIFF
--- a/app/assets/stylesheets/partials/_reports.scss
+++ b/app/assets/stylesheets/partials/_reports.scss
@@ -272,7 +272,8 @@
 .fs-32px {
   font-size: 32px;
 }
-.fs-xxlarge {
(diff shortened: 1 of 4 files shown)


claudiovallejo (approved):
@dburka I tested locally and the code looks good! My only comment is that I think the space between the tabbed nav should be `16px` on mobile—`12px` feels a _bit_ too tight.

PR #3646, opened 2022-03-29
BSNL template management

prabhanshuguptagit:
**Story card:** [sc-7356](https://app.shortcut.com/simpledotorg/story/7356/map-dlt-template-ids-and-the-specific-variables-they-need-to-internal-message-template-ids)

## Because

We need to maintain a mapping between our notification strings and the DLT template ID to send messages via BSNL.

## This addresses

Adds a script to assist with fetching all the templates from BSNL and storing the template ID. Also kicks off a readme to document some of this process, might break it up into separate ones as it evolves (over the next few PRs).


app/services/messaging/bsnl/api.rb
```diff
@@ -0,0 +1,35 @@
...
+    {
+      header: ENV["BSNL_IHCI_HEADER"],
+      message_type: "SI",
+      entity_id: ENV["BSNL_IHCI_ENTITY_ID"],
+      jwt: ENV["BSNL_JWT_TOKEN"]
+    }
+  end
+
+  def self.post(path, body = nil)
+    raise Messaging::Error "Missing BSNL credentials" unless credentials.values.all?
```
prabhanshuguptagit, line 19: This exception probably belongs in the `initializer`, putting here so we don't miss it later.

doc/howto/manage_bsnl_sms_reminders.md
```diff
@@ -0,0 +1,44 @@
+## Sending reminder SMS via BSNL
+
+We use BSNL to send reminder messages in India. This requires us to comply with DLT regulations, for which we need to maintain
+a mapping between IDs assigned by DLT and our messages (identified by locale keys).
+
+This is a how-to on interacting with the BSNL and DLT dashboards and maintaining the required mapping with `simple-server`.
```
prabhanshuguptagit, line 6: Would like to commit this as a WIP and expand on it later.

lib/tasks/scripts/get_bsnl_template_details.rb
```diff
@@ -0,0 +1,122 @@
+class GetBsnlTemplateDetails
+  include I18n::Backend::Flatten
+  include Memery
+
+  def initialize
+    @template_details = Messaging::Bsnl::Api.new.get_template_details
+    @template_names = @template_details.map { |detail| detail["Template_Name"] }
+  end
+
+  attr_reader :template_details, :template_names
```
govindkrjoshi, line 10: [Style][Not important] Maybe put the attr_reader above the initialize method?

app/services/messaging/bsnl/api.rb
```diff
@@ -0,0 +1,41 @@
...
+    request = Net::HTTP::Post.new(path, "Content-Type" => "application/json")
+    request.body = body.to_json if body
+    request["Authorization"] = "Bearer #{credentials[:jwt]}"
+
+    json_or_string(http.request(request).body)
+  end
+
+  def json_or_string(string)
+    JSON.parse(string)
+  rescue JSON::ParserError
```
vkrmis, line 38: Is the response not predictable here? Wondering why the rescue is needed.

config/data/bsnl_template_details.yml
```diff
@@ -0,0 +1,262 @@
+# This is an autogenerated file. Do not modify.
+# Use get_bsnl_template_details.rake to fetch a new copy of this file.
+---
+hi-IN.notifications.set01.basic:
+  Template_Id: '1407164786628639161'
+  Template_Keys: []
```
vkrmis, line 6: Is this because the vars haven't been named?

doc/howto/manage_bsnl_sms_reminders.md
```diff
@@ -0,0 +1,45 @@
...
+  The locale name needs to be hyphenated (`hi-IN` and not `hi_IN`). This format is used to associate the templates on BSNL to locale keys so it's an important step.
+- The template should show up on the BSNL dashboard in a few hours. You will need to name the variables
+in the template.
+
+<TODO flesh out these notes on dashboard usage and end-to-end flow, maybe add a video>
+
+### Maintaining template IDs
+After the new template has been added to the BSNL dashboard, we need to pull in it's `DLT Template ID` before it can be used as a reminder SMS. We have a script to help with it.
+
+**Note:** You will need production credentials to run the script.
```
vkrmis, line 27: Add a note about not using token 1 anywhere other than production. Or whichever token we want to reserve for prod use.

vkrmis (approved):
Left a few q's, but should be good to merge.

app/services/messaging/bsnl/api.rb
prabhanshuguptagit: The API returns a JSON for valid responses and an empty string when there are errors. I wanted the API to intelligently cast JSON responses to a ruby hash, hence the method. 

Renamed this to `hsh_or_string` in [8b261e6](https://github.com/simpledotorg/simple-server/pull/3646/commits/8b261e6d3a96bab449ea5ad21789336ccf6fbe18) since it returns a hash, not JSON.

config/data/bsnl_template_details.yml
prabhanshuguptagit: Yes. These should look like the `en` messages below when we fetch this list after naming.

doc/howto/manage_bsnl_sms_reminders.md
prabhanshuguptagit: I've instructed to copy the JWT token from production so they don't have to generate a new one. Should be ok?

vkrmis: Hmm, I guess that works for now. Let's come back and think about developer UX later though. Should be easy enough to set up our scripts to use a different token id for dev use.

PR #3651, opened 2022-04-02
Reports charts better mobile formatting

dburka:
**Story card:** None

## Because

Charts are hard to read on mobile.

## This addresses

* Makes the type slightly smaller for labels in chart.js
* Set's a min-height for the stacked bar charts on Overview in reports

## Test instructions

Shouldn't change anything other than what's listed above.


app/views/reports/regions/_treatment_outcomes_card.html.erb
```diff
@@ -7,7 +7,7 @@
   </h3>
 
   <div class="d-flex fd-column fd-lg-row">
-      <div class="pl-lg-12px w-lg-50 h-lg-auto h-print-12cm">
+      <div class="pl-lg-12px w-lg-50 h-lg-auto h-print-12cm" style="min-height: 300px;">
```
harimohanraj89, line 10: I know it's annoying 😅 but can we introduce this CSS through a class and a rule in the appropriate CSS file? Let's avoid inline styles.

harimohanraj89 (changes requested):
Could you also attach a couple before/after screeshots to the PR description? Very helpful for reviewers.

app/views/reports/regions/_treatment_outcomes_card.html.erb
dburka: I was wondering what's the point on this other than philosophy? Now we're writing in very specific rules into our whole CSS file for 1-off formatting on a single page.

dburka:
Before on Sandbox, new on Dev. 
image
image

app/views/reports/regions/_treatment_outcomes_card.html.erb
dburka: Anyhow, I'll move it.

dburka:
Before and after min-height on the stacked bar charts. 
image
image

app/views/reports/regions/_treatment_outcomes_card.html.erb
harimohanraj89: The philosophy is consistency. We want to minimize the number of places we need to look to find out where an element's style is coming from.

PR #3653, opened 2022-04-04
Add BSNL delivery detail

prabhanshuguptagit:
**Story card:** [sc-7624](https://app.shortcut.com/simpledotorg/story/7624/make-provider-specific-api-service-classes-bsnl)

## Because

We want to record communications and delivery results for BSNL SMSes.

## This addresses

Starts recording communications and delivery details from the BSNL channel. The detailable only contains the message ID on creation. The rest of the fields will be populated by a separate call to an API to fetch the message's status.


db/migrate/20220403123845_create_bsnl_delivery_details.rb
```diff
@@ -0,0 +1,14 @@
+class CreateBsnlDeliveryDetails < ActiveRecord::Migration[5.2]
+  def change
+    create_table :bsnl_delivery_details do |t|
+      t.string :message_id
+      t.string :message_status
+      t.string :result
```
harimohanraj89, line 6: If we wanna keep the PRs a bit scoped, wanna add just the fields we're using into this PR, and introduce the rest just-in-time? AKA the PR that populates them via API or consumes them somewhere?

harimohanraj89 (changes requested):
Tiny things

db/migrate/20220403123845_create_bsnl_delivery_details.rb
prabhanshuguptagit: That PR is coming right next up, so I'd like to keep it into a single migration for ease.

harimohanraj89: Makes sense if we have a very good picture of how this shakes out 👍

db/migrate/20220403123845_create_bsnl_delivery_details.rb
```diff
@@ -0,0 +1,14 @@
+class CreateBsnlDeliveryDetails < ActiveRecord::Migration[5.2]
+  def change
+    create_table :bsnl_delivery_details do |t|
+      t.string :message_id
```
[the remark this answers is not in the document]
harimohanraj89: What about, say, dlt_template_id? Will we ever create BSNL details without a template ID?

harimohanraj89 (approved):

harimohanraj89:
I've got one last question about more null constraints. PS we should back up the constraints with model validations. Definitely don't want to have to go all the way down to the DB to throw an (unfriendly) error.

Approving to unblock you once constraints and validations are all set!

PR #3654, opened 2022-04-05
Create page for diabetes reports

govindkrjoshi:
**Story card:** [sc-7788](https://app.shortcut.com/simpledotorg/story/7788/create-a-page-for-the-diabetes-reports)

## Because

We want to add diabetes reports to our dashboards

## This addresses
- Creates a new page and route for diabetes reports
- Adds a feature toggle to enable or disable links to the diabetes page
- Returns not found if the page is accessed without enabling the feature

## Test instructions
1. Enable `diabetes_management_reports` flag on flipper
2. There should be a link to the diabetes reports page on the Region reports page in the dashboard.
3. Clicking on this link should redirect you to the diabetes reports page. This is an empty page at the moment. 
4. If the feature flag is disabled, accessing the page should result in a 404.


app/models/region.rb
```diff
@@ -168,6 +168,11 @@ def appointed_patients
     Patient.joins(:appointments).where(appointments: {facility: facility_regions.pluck(:source_id)})
   end
 
+  def diabetes_management_enabled?
+    Flipper.enabled?(:diabetes_management_reports) &&
+      facilities.where(enable_diabetes_management: true).exists?
+  end
```
prabhanshuguptagit, line 174: [rename] Lets call this `diabetes_reports_enabled?`.

app/views/reports/regions/_header.html.erb
```diff
@@ -37,6 +37,11 @@
       <p class="mr-12px mr-lg-24px mb-0px pb-8px fs-16px fs-lg-20px fw-bold <%= params[:action] == "cohort" ? "c-red bb-2px bb-red" : "c-grey-dark" %>">
         <%= link_to_unless_current "Cohort reports", reports_region_cohort_path(@region, report_scope: params[:report_scope]), :class => "c-grey-dark td-none" %>
       </p>
```
prabhanshuguptagit, line 39: The other names also need to be prefixed with HTN, for ex `Overview` becomes `HTN Overview`. Should that be part of this PR?

spec/controllers/reports/regions_controller_spec.rb
```diff
@@ -600,4 +633,40 @@
...
     end
   end
+
+  describe "#diabetes" do
+    let(:facility) { create(:facility, facility_group: facility_group_1, enable_diabetes_management: true) }
+    let(:facility_region) { create(:region, region_type: :facility, source: facility, reparent_to: Region.root) }
+
+    before do
+      sign_in(cvho.email_authentication)
+      Flipper.enable(:diabetes_management_reports)
```
prabhanshuguptagit, line 643: Let's move this flipper enable to inside the `it`s, that way we don't have to disable it to override on line 664.

spec/models/region_spec.rb
```diff
@@ -531,6 +531,56 @@
...
+      end
+
+      it "returns false if enable_diabetes_management is set to false for the facility" do
+        facility_region = create(:region, region_type: :facility, source: facility_2, reparent_to: enabled_block_region)
+        expect(facility_region.diabetes_management_enabled?).to eq(false)
+      end
+    end
+
+    context "region contains many facilities" do
+      it "returns true if any of its facilities has enabled_diabetes_management set to  true" do
```
prabhanshuguptagit, line 564: 👾 
```suggestion
      it "returns true if any of its facilities has enabled_diabetes_management set to true" do
```

prabhanshuguptagit (approved):
Approving to unblock, some minor comments.

app/models/region.rb
govindkrjoshi: This is a method on the region, which checks if any of its facilities have diabetes management enabled. The feature flag check here is just for convenience. When we remove the feature flag, the name should make more sense.

app/views/reports/regions/_header.html.erb
govindkrjoshi: We've created a separate story for this.

app/models/region.rb
prabhanshuguptagit: I'd suggest breaking it like so in that case, since someone might want to use `diabetes_management_enabled` in a different context
```
def diabetes_management_enabled?
   facilities.where(enable_diabetes_management: true).exists?
end

def diabetes_reports_enabled?
   Flipper.enabled? && diabetes_management_enabled?
```

PR #3656, opened 2022-04-06
Show adjusted patient numbers in monthly data

prabhanshuguptagit:
**Story card:** [sc-7833](https://app.shortcut.com/simpledotorg/story/7833)

## Because

In the monthly district data, we currently show the "Total patients under care" as of the current month. The trends however use the "Adjusted patients under care number" (under care as of 3 months ago) to calculate the control/uncontrolled rates. This can be potentially misleading to CVHOs.

## This addresses

Changes the "Total patients under care" to show `adjusted_patients` instead of `under_care`.

### DIFF
--- a/app/exporters/monthly_district_report/block_data.rb
+++ b/app/exporters/monthly_district_report/block_data.rb
@@ -58,7 +58,7 @@ module MonthlyDistrictReport
         "Blocks" => block.name,
         "Total registrations" => repo.cumulative_registrations[block.slug][report_month],
         "Total assigned patients" => repo.cumulative_assigned_patients[block.slug][report_month],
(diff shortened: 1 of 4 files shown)


vkrmis (approved):

vkrmis:
nit: wdyt about renaming `adjusted_patients` to `adjusted_patients_under_care`?

prabhanshuguptagit:
This name comes from the [repository](https://github.com/simpledotorg/simple-server/blob/ced15f0f536f9f54d807fa945c0f03f1e64d7712/app/models/reports/region_summary_schema.rb#L54), going to leave it as is since it'll be a bigger change.

PR #3658, opened 2022-04-06
Fetch BSNL sms statuses

prabhanshuguptagit:
**Story card:** [sc-7838](https://app.shortcut.com/simpledotorg/story/7838/add-job-to-fetch-bsnl-sms-delivery-statuses)

## Because

BSNL does not support callbacks for status updates and instead has an API to fetch message statuses from. We will need periodically fetch those statuses from the API.

## This addresses

- Adds a job that can fetch the result for a message
- Adds the orchestrator for these jobs to `schedule.rb`. We'll fetch the statuses for all pending messages once a day.

### DIFF
--- a/app/jobs/bsnl_sms_status_job.rb
+++ b/app/jobs/bsnl_sms_status_job.rb
@@ -1,0 +1,50 @@
+class BsnlSmsStatusJob
+  include Sidekiq::Worker
+  include Sidekiq::Throttled::Worker
+  BSNL_TIME_ZONE = "Asia/Kolkata"
+
+  sidekiq_options queue: :default
+  sidekiq_options retry: 3
+
+  sidekiq_throttle(
+    threshold: { limit: 250, period: 1.minute }
+  )
+
+  def perform(message_id)
+    response = Messaging::Bsnl::Api.new.get_message_status_report(message_id)
+    handle_api_errors(message_id, response)
+
+    detailable = BsnlDeliveryDetail.find_by!(message_id: message_id)
+    detailable.message_status = response["Message_Status"] if response["Message_Status"]
+    detailable.result = response["Message_Status_Description"] if response["Message_Status_Description"]
+    detailable.delivered_on = parse_timestamp(response["Delivery_Success_Time"]) if response["Delivery_Success_Time"]
+    detailable.save!
+
+    log(response)
+  end
+
+  private
+
+  def log(response)
+    Rails.logger.info(
+      {
+        message_id: response["Message_Id"],
+        content_template_id: response["Content_Template_Id"],
+        error: response["Error"],
+        sms_count: response["SMS_Count"],
+        dlr_error_code: response["DLR_Error_Code"]
+      }
+    )
+  end
+
+  def handle_api_errors(message_id, response)
+    error = response["Error"]
+    if error.present?
+      raise Messaging::Bsnl::Error.new("Error fetching message status for #{message_id}: #{error}")
+    end
+  end
+
+  def parse_timestamp(timestamp)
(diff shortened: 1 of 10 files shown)


app/jobs/bsnl_sms_status_job.rb
```diff
@@ -0,0 +1,50 @@
+class BsnlSmsStatusJob
+  include Sidekiq::Worker
+  include Sidekiq::Throttled::Worker
+  BSNL_TIME_ZONE = "Asia/Kolkata"
+
+  sidekiq_options queue: :default
+  sidekiq_options retry: 3
+
+  sidekiq_throttle(
+    threshold: {limit: 250, period: 1.minute}
```
harimohanraj89, line 10: API limit?

app/jobs/bsnl_sms_status_job.rb
```diff
@@ -0,0 +1,50 @@
...
+  )
+
+  def perform(message_id)
+    response = Messaging::Bsnl::Api.new.get_message_status_report(message_id)
+    handle_api_errors(message_id, response)
+
+    detailable = BsnlDeliveryDetail.find_by!(message_id: message_id)
+    detailable.message_status = response["Message_Status"] if response["Message_Status"]
```
harimohanraj89, line 18: We could do `response["Message_Status"].presence`

app/jobs/bsnl_sms_status_job.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    log(response)
+  end
+
+  private
+
+  def log(response)
+    Rails.logger.info(
+      {
+        message_id: response["Message_Id"],
```
harimohanraj89, line 31: Do we need a uniform tag in here like `action: bsnl_message_send` or something, so we can isolate these logs easily?

harimohanraj89: That was a terrible name for it, but you get the idea

app/services/messaging/bsnl/api.rb
```diff
@@ -29,6 +30,12 @@ def get_template_details
     post(URL_PATHS[:get_content_template_details])["Content_Template_Ids"]
   end
 
+  def get_message_status_report(message_id)
+    post(URL_PATHS[:message_status_report], {
```
harimohanraj89, line 34: It's a POST request? Of course...

app/services/messaging/bsnl/sms.rb
```diff
@@ -3,6 +3,12 @@ def self.communication_type
     Communication.communication_types[:sms]
   end
 
+  def self.get_message_statuses
+    BsnlDeliveryDetail.in_progress.in_batches(of: 1_000).each_record do |detailable|
```
harimohanraj89, line 7: We could use `find_each` to do the same thing, less verbose.

harimohanraj89 (approved):
Tiny stuffs, looks great, approval to unblock!

app/jobs/bsnl_sms_status_job.rb
prabhanshuguptagit: Not a limitation on the API; We are limiting ourselves in order to not DOS their services, since we can have 15-20k messages on a given day.

app/jobs/bsnl_sms_status_job.rb
prabhanshuguptagit: We don't want to override a message status if it already exists, hence the `if`.

harimohanraj89: Oh I see, very cool.

PR #3661, opened 2022-04-07
Add ReportingPatientBloodSugars

govindkrjoshi:
**Story card:** [sc-7793](https://app.shortcut.com/simpledotorg/story/7793/update-the-materialized-views-to-include-diabetes-outcomes)

## Because

We need to setup a reporting pipeline for Diabetes reports. This view is a required to calculate the patient's  state per month

## This addresses

Adds a materialized view to get the latest blood sugar measurement for a patient in every calendar month. 

## Test instructions

1. Checkout to the branch locally
2. Run `bundle exec rails db:migrate`
3. You should be able to see the `reporting_patient_blood_sugars` materialized view in postgres
4. Run `bundle exec rails db:rollback` to test the rollback


db/views/reporting_patient_blood_sugars_v01.sql
```diff
@@ -0,0 +1,88 @@
+-- Only most recent BP per patient per month. BPs are ordered appropriately below.
```
vkrmis, line 1: This comment should say BS instead of BP

vkrmis, line 1: This comment should say BS instead of BP

db/views/reporting_patient_blood_sugars_v01.sql
```diff
@@ -0,0 +1,88 @@
...
+    bs.id AS blood_sugar_id,
+    bs.patient_id,
+    bs.blood_sugar_type,
+    bs.blood_sugar_value,
+    bs.facility_id AS blood_sugar_facility_id,
+
+    ------------------------------------------------------------
+    -- Risk state of the latest measurement as of given month
+    -- +----------+--------------+------------------+-------------+
+    -- |   Type   | bs-below-200 |  bs-200-to-299   | bs-over-299 |
```
vkrmis, line 22: Is it possible to translate `bs-below-200`, `bs-200-to-299`, `bs-over-299` to `low`, `medium`, `high` or something similar? It might be confusing to explain why `Hba1c: 9.5` is `bs-over-299` if these values are visible to users.

vkrmis, line 22: Is it possible to translate `bs-below-200`, `bs-200-to-299`, `bs-over-299` to `low`, `medium`, `high` or something similar? It might be confusing to explain why `Hba1c: 9.5` is `bs-over-299` if these values are visible to users.

app/models/reports/patient_blood_sugar.rb
```diff
@@ -0,0 +1,16 @@
+module Reports
+  class PatientBloodSugar < Reports::View
+    self.table_name = "reporting_patient_blood_sugars"
+    belongs_to :patient
+
+    enum blood_sugar_risk_state: {
+      bs_below_200: "bs_below_200",
+      bs_200_to_300: "bs_200_to_300",
+      bs_over_300: "bs_over_300"
+    }
```
prabhanshuguptagit, line 10: Do we foresee a case where we will need to deal with these states in the context of just blood sugars, without the patient association? This could be broken into two tables, if we feel the need to:
- A `blood_sugar_risk_states` table where these buckets are calculated for all blood sugars
- `reporting_patient_blood_sugars` can join on the latest BS from that table.

db/views/reporting_patient_blood_sugars_v01.sql
prabhanshuguptagit: Same thoughts as around naming ^ Would an internal mapping between low, medium, high help even if the users want labels to be `BS <200`?

prabhanshuguptagit (approved):
Looks good, one suggestion around breaking the table further.

app/models/reports/patient_blood_sugar.rb
govindkrjoshi: This enum state is being calculated as a part of the materialized view. Calculating it in the materialized view is the simpler, and also more consistent with the rest of the states we evaluate for reporting.

Another option I considered was to add a risk_state column to blood_sugars (column doesn't get synced), and populating this risk_state on save. This would be similar to the `blood_sugar_risk_states` table above. But I chose not to do this since we don't need the risk states outside the reporting pipeline.

db/views/reporting_patient_blood_sugars_v01.sql
govindkrjoshi: We chose not to use generic terms like `low/medium/high` or `controlled/uncontrolled` because there is no clinical consensus what the ranges should be. Using this naming convention we are trying to sidestep this confusion. 

Slack thread: https://simpledotorg.slack.com/archives/CLS8YD22H/p1649156053372179

app/models/reports/patient_blood_sugar.rb
prabhanshuguptagit: Sorry, I meant a `blood_sugar_risk_states` materialised view (not a table). Basically break the bucket definition and the "latest per patient" into two views instead of one.

govindkrjoshi: We can pull this view out in the future if it's necessary. Since this is only needed in a reporting view now, adding an extra materialized view will also mean adding an extra when fetching reports and it seems like unnecessary complexity at the moment.

PR #3672, opened 2022-04-11
Use `Date.today` as the mock time

PriyangaPKini:
**Story card:** [sc-7915](https://app.shortcut.com/simpledotorg/story/7915/fix-facility-progress-spec)

## Because

Mat views use the system time which is different from the initial mock time, resulting the test to fail after a specific date.

## This addresses

It fixes the mock time so that the test is independent of the current date


prabhanshuguptagit (approved):
[minor] Please change the PRs title to something that is more meaningful in the context of the master branch while merging (ex: Fix timecop usage in reporting spec).

PR #3678, opened 2022-04-13
Add a rake task to check BSNL account balance

vkrmis:
**Story card:** [sc-7910](https://app.shortcut.com/simpledotorg/story/7910/create-a-task-that-alerts-when-bsnl-balance-goes-below-a-threshold)

## Because

We want to be alerted if we're in danger of running out of BSNL SMS credits

## This addresses

Adds a rake task to check BSNL SMS balance, and run it daily in India production


lib/tasks/scripts/get_bsnl_account_balance.rb
```diff
@@ -0,0 +1,27 @@
+class GetBsnlAccountBalance
+  # (20_000 * 3 patients from current experiments + 15_000 patients from stale experiments) * 4 segments per message
+  # We're unlikely to actually hit this number on any given day, but we'd like to alert early if we're close to the limit
+  MAX_DAILY_MESSAGE_SEGMENT_COUNT = 75_000 * 4
+  BALANCE_EXPIRY_ALERT_DAYS = 3
```
vkrmis, line 5: @prabhanshuguptagit can you verify this math looks good?

Also setting `BALANCE_EXPIRY_ALERT_DAYS` to be low deliberately, can bump it up to a week once the larger recharge goes through

lib/tasks/scripts/get_bsnl_account_balance.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def call
+    if expiry_date < BALANCE_EXPIRY_ALERT_DAYS.days.from_now
+      raise BalanceError.new("Account balance is going to expire in less than #{BALANCE_EXPIRY_ALERT_DAYS} days. Please extend validity before #{expiry_date.strftime("%d-%b-%y")}")
+    end
+
+    if total_balance_remaining < (MAX_DAILY_MESSAGE_SEGMENT_COUNT * BALANCE_EXPIRY_ALERT_DAYS)
+      raise BalanceError.new("Account balance remaining is #{total_balance_remaining} segments, may run out in less than #{BALANCE_EXPIRY_ALERT_DAYS} days")
+    end
```
vkrmis, line 25: There is a potential edge case that I didn't handle here, when there are multiple recharges: 
- The one with most of the credits is expiring first
- The others have minimal credits but expire later

This alert won't pick up on the problem till the first set of credits expires. Could handle it, but this logic will become much more complex. Think it's okay to ignore right now.

lib/tasks/scripts/get_bsnl_account_balance.rb
```diff
@@ -0,0 +1,27 @@
+class GetBsnlAccountBalance
+  # (20_000 * 3 patients from current experiments + 15_000 patients from stale experiments) * 4 segments per message
+  # We're unlikely to actually hit this number on any given day, but we'd like to alert early if we're close to the limit
+  MAX_DAILY_MESSAGE_SEGMENT_COUNT = 75_000 * 4
```
prabhanshuguptagit, line 4: This is a bit on the higher side, with these many SMS we probably have only 5 days remaining. 75_000 * 3 is a bit more realistic (still high).

lib/tasks/scripts/get_bsnl_account_balance.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  class BalanceError < StandardError
+  end
+
+  attr_reader :recharge_details, :expiry_date, :total_balance_remaining
+
+  def initialize
+    @recharge_details = Messaging::Bsnl::Api.new.get_account_balance
+    @expiry_date = Date.parse(recharge_details.map { |recharge| recharge["Balance_Expiry_Time"] }.max).beginning_of_day
+    @total_balance_remaining = recharge_details.reduce(0) { |total, recharge| total + recharge["SMS_Balance_Count"].to_i }
```
prabhanshuguptagit, line 15:
```suggestion
    @total_balance_remaining = recharge_details.map { |recharge_detail| recharge_detail["SMS_Balance_Count"].to_i }.sum
```

lib/tasks/scripts/get_bsnl_account_balance.rb
```diff
@@ -0,0 +1,27 @@
+class GetBsnlAccountBalance
+  # (20_000 * 3 patients from current experiments + 15_000 patients from stale experiments) * 4 segments per message
+  # We're unlikely to actually hit this number on any given day, but we'd like to alert early if we're close to the limit
+  MAX_DAILY_MESSAGE_SEGMENT_COUNT = 75_000 * 4
+  BALANCE_EXPIRY_ALERT_DAYS = 3
+
+  class BalanceError < StandardError
```
prabhanshuguptagit, line 7: Let's namespace this under `Messaging::Bsnl::`?

prabhanshuguptagit (approved):

lib/tasks/scripts/get_bsnl_account_balance.rb
vkrmis: Alright, I'll change it to `* 3`

lib/tasks/scripts/get_bsnl_account_balance.rb
vkrmis: Cool, wasn't sure if this is a `Messaging` error necessarily, but it should be okay.

PR #3680, opened 2022-04-14
Versioning DLT templates

prabhanshuguptagit:
**Story card:** [sc-7901](https://app.shortcut.com/simpledotorg/story/7901/make-dlt-templates-versionable)

## Because

Since DLT templates can't be edited or deleted, we need a system to map updates to notification strings in simple-server to template names in DLT.

## This addresses

Makes `DltTemplate` version aware and adds methods to support picking up the latest version of a template.


harimohanraj89:
Why can't we just remove the old template from our mapping? In other words, update our mapping with a _new_ template ID etc? Is there a reason we need to keep the full template history in our mapping config?

prabhanshuguptagit:
We want to keep all the templates since it also serves as a source of truth for what templates we have on the platform. We found it simpler to keep the file itself quite dumb and do the versioning afterwards, decoupling versioning from keeping track of templates in general.

harimohanraj89:
The BSNL dashboard doesn't expose this source of truth for us? I'd love to keep this file simple if we can. If we need to maintain this source of truth ourselves, then could we stick a guide into `doc/howto` to describe how to update a template? Link to the doc at the top of the YML file?

prabhanshuguptagit:
BSNL has this on the dashboard but you have to click around a lot since we have a many templates, it's a bit unmanageable.

> Link to the doc at the top of the YML file?
We'll add this as a part of [this story](https://app.shortcut.com/simpledotorg/story/7859/polish-bsnl-related-documentation) (picking this up after delivering a report to Dr Reena first). There's a WIP readme [here](https://github.com/simpledotorg/simple-server/blob/master/doc/howto/manage_bsnl_sms_reminders.md).

app/models/messaging/bsnl/dlt_template.rb
```diff
@@ -108,4 +131,12 @@ def resize_variables_equally(permitted_length, content)
       [k, v[0, lengths[k]]]
     end
   end
+
+  class << self
+    def numeric?(string)
```
harimohanraj89, line 136: Wanna just declare `def self.numeric?`, consistent with the other class methods in here?

app/models/messaging/bsnl/dlt_template.rb
```diff
@@ -23,11 +26,31 @@ def initialize(dlt_template_name)
...
+    dlt_template_name.chomp(".#{version_number(dlt_template_name)}")
+  end
+
+  def self.version_number(dlt_template_name)
+    version = dlt_template_name.split(".").last
+
+    if numeric?(version)
+      version.to_i
+    else
+      DEFAULT_VERSION_NUMBER
```
harimohanraj89, line 50: If we return `1` here, won't the `latest_name_of` method break for all templates that have only a single version and don't have a numeric suffix in the template name?

app/models/messaging/bsnl/dlt_template.rb
```diff
@@ -23,11 +26,31 @@ def initialize(dlt_template_name)
...
+    end
+  end
```
harimohanraj89, line 52: These mechanics are a bit awkward because we've got the version number in the template name, yes, but also as a key in the template. Looks like we _also_ have a reference to the latest version in each blob, which will have to be updated for every existing version of a template if we update a template. Could I suggest simplifying the data model a little bit and making these mechanics methods a bit smarter?

Eg.

```yml
template_name.thing.1: # we can keep the suffix optional of course, it's just a way to give each template a unique name in this YML file
  version: 1
template_name.thing.2:
  version: 2
```

```ruby
def versions(template_name)
  templates.select { |k, v| k.start_with?(template_name) }
end

def latest_version(template_name)
  name, template = versions(template_name).max_by {|k, v| v["version"] || DEFAULT_VERSION_NUMBER}
  # return whatever you want. the name or the template
end
```

This way:
* We don't have to add and maintain a latest version ref to each template
* We don't have to add a version key to templates with only a single version
* We can rely on the version key, and not worry about the template names so much

harimohanraj89 (changes requested):
I understand the rationale but have some ideas on simplifying the execution.

app/models/messaging/bsnl/dlt_template.rb
prabhanshuguptagit: standardrb enforces this:
```
Lint/IneffectiveAccessModifier: `private` (on line 102) does not make singleton methods private. Use `private_class_method` or `private` inside a `class << self` block instead.
```

app/models/messaging/bsnl/dlt_template.rb
prabhanshuguptagit: I've chosen to add the references to the YAML since doing the `max_by` requires us to iterate over all the templates to find the latest at "runtime". We'll make 40-50k calls to `DltTemplate.new` every day, at which point the performance gain by storing the version in the YAML once would be. The tradeoff

> which will have to be updated for every existing version of a template if we update a template

This should be ok since it's automated. One shouldn't have to touch the YAML directly ever.

prabhanshuguptagit: I've chosen to add the references to the YAML since doing `max_by` requires us to iterate over all the templates to find the latest at "runtime". We'll make 40-50k calls to `DltTemplate.new` every day, so it's more performant to gain by storing the version in the YAML once would be. I haven't quantified what the gains would be (we have ~200 templates) but can check that out if the tradeoff seems too bad.

> which will have to be updated for every existing version of a template if we update a template

This should be ok since it's automated. One shouldn't have to touch the YAML directly ever.

app/models/messaging/bsnl/dlt_template.rb
harimohanraj89: But we don't need to make this private, yeah? We could just do it the other way and move the method above the `private`?

app/models/messaging/bsnl/dlt_template.rb
harimohanraj89: Ohhh I see. Yeah if that thing is auto-generated, then sounds good!

app/models/messaging/bsnl/dlt_template.rb
prabhanshuguptagit: Wanted to make it private since no one should need to call `DltTemplate.numeric?` ever.

app/models/messaging/bsnl/dlt_template.rb
prabhanshuguptagit: The `latest_name_of` looks at templates after removing the version from the template name, so it will find out the first version without the name as the "default".

harimohanraj89 (approved):
Let's go 👍

PR #3682, opened 2022-04-15
Exclude English patients temporarily from current A/B experiment

prabhanshuguptagit:
**Story card:** -

## Because

We had to resubmit english SMS templates for DLT approval because they were originally submitted as "Non-regional". This prevented us from using unicode characters in the messages. Since the new templates have been submitted, we don't have any approved english DLT messages. Any english messages sent will be rejected.

## This addresses

This adds a temporary exclusion to the experiment's selection criteria for patients in english speaking regions (only Nagaland). This PR should be reverted once the english SMSes have been approved on DLT.

### DIFF
--- a/app/models/experimentation/notifications_experiment.rb
+++ b/app/models/experimentation/notifications_experiment.rb
@@ -47,7 +47,14 @@ module Experimentation
                                              .where(status: :scheduled)
                                              .group(:patient_id)
                                              .having("count(patient_id) > 1"))
+        .then { |patients| exclude_nagaland_patients(patients) }
         .then { |patients| exclude_bangladesh_blocks(patients) }
+    end
+
+    def self.exclude_nagaland_patients(patients)
+      return patients unless CountryConfig.current_country?("India")
+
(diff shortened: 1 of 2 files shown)


app/models/experimentation/notifications_experiment.rb
```diff
@@ -47,9 +47,16 @@ def self.eligible_patients
                                              .where(status: :scheduled)
                                              .group(:patient_id)
                                              .having("count(patient_id) > 1"))
+        .then { |patients| exclude_nagaland_patients(patients) }
```
kitallis, line 50: I love `then`.

prabhanshuguptagit: More 
https://github.com/simpledotorg/simple-server/blob/2292ffb3753345aa8432cc8394b263ec136a8b4c/app/models/messaging/bsnl/dlt_template.rb#L31

vkrmis (approved):
Looks good, lets merge this today

PR #3686, opened 2022-04-20
Fixed print formatting for Praveen's bug.

dburka:
**Story card:** [sc-7799](https://app.shortcut.com/simpledotorg/story/7799/graph-alignment-is-broken-in-pdf-download)

## Because

Praveen pointed out that Reports graphs were overflowing the cards.

## This addresses

Someone coded these to be very fragile with specific "cm" widths for the cards. This is not a good approach. I took out the specific widths and will later come back and make them relative %s. I confirmed with Praveen that a single row of stacked cards is acceptable for now.

## Test instructions

Load any report. Try PRINT and look at print preview or PDF.

### DIFF
--- a/app/views/shared/_nav_bar.html.erb
+++ b/app/views/shared/_nav_bar.html.erb
@@ -164,7 +164,7 @@
 
 </nav>
 
-<div class="mobile-header">
+<div class="mobile-header d-print-none">
     <a href="#" onclick="openMenu()" class="navigation-icon"><i class="fa-regular fa-bars pr-1"></i> MENU</a>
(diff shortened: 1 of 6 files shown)


app/views/shared/_nav_bar.html.erb
```diff
@@ -164,7 +164,7 @@
 
 </nav>
 
-<div class="mobile-header">
+<div class="mobile-header d-print-none">
```
harimohanraj89, line 167: Does this class simply hide this section in the printable version?

dburka: Yes, exactly.

PR #3695, opened 2022-04-26
Update Inactive facilities definition

prabhanshuguptagit:
**Story card:** [sc-8061](https://app.shortcut.com/simpledotorg/story/8061/update-inactive-facilities-definition-on-dashboard-to-match-metabase-s-active-facilities-definition)

## Because

We want to update the definition of an Inactive facility on the dashboard to match what is shown on metabase's `Active Facilities` query.

## This addresses

This modifies the thresholds to "at least 1 BP in the last 30 days" for an inactive facility. It
- Does not count non-hypertensive patient's BPs
- Only counts 1 BP per patient per day

Metabase currently counts simply all blood pressures for counting `Active facilities`. We will update it to match the dashboard definition as well.


dburka (approved):
This all looks good. Nice to see so much code removed.

PR #3706, opened 2022-05-03
Process one line list at a time in sidekiq

prabhanshuguptagit:
**Story card:** [sc-7933](https://app.shortcut.com/simpledotorg/story/7933/don-t-queue-multiple-line-list-jobs-for-a-single-region)

## Because

Our sidekiq server tends to go down if multiple line lists for big regions are queued up, since line lists are CPU intensive. This happens when the user requests the line lists many times while waiting for previous jobs to finish.

## This addresses

Makes sure that only one patient line list per region is allowed. Adds the `sidekiq-unique-jobs` gem, it looks like the go-to gem for doing locks in sidekiq. Although it's a bit overkill for what we want, it's simpler than implementing locks by ourself.

- [] Only one line list per region should be allowed. Any new requests should be directly moved to `dead`
- [] Multiple line lists across regions should be allowed


app/jobs/patient_list_download_job.rb
```diff
@@ -1,11 +1,18 @@
-class PatientListDownloadJob < ApplicationJob
+class PatientListDownloadJob
+  include Sidekiq::Worker
+
+  sidekiq_options lock: :until_executed,
+                  lock_timeout: 15.minutes,
+                  lock_ttl: 15.minutes,
+                  on_conflict: :reject
```
prabhanshuguptagit, line 7: From the docs:
```
A lock is created when perform_async is called. Then it is either unlocked when lock_ttl is hit or when Sidekiq has called the perform method on your worker.
```

app/jobs/patient_list_download_job.rb
```diff
@@ -1,14 +1,21 @@
-class PatientListDownloadJob < ApplicationJob
+class PatientListDownloadJob
+  include Sidekiq::Worker
+
+  sidekiq_options lock: :until_executed,
```
harimohanraj89, line 4: So this lock is for subsequent jobs that are queued with the exact same parameters? If two different people request a facility's linelist, they'd both get it right?

app/jobs/patient_list_download_job.rb
```diff
@@ -1,14 +1,21 @@
...
+    on_server_conflict: :reject
+
   def perform(recipient_email, model_type, params)
     case model_type
     when "facility"
-      model = Facility.find(params[:facility_id])
-      model_name = model.name
-    when "facility_group"
-      model = FacilityGroup.find(params[:id])
+      model = Facility.find(params["facility_id"])
```
harimohanraj89, line 12: Is this hash not indifferent-access?

harimohanraj89 (approved):
Seems straightforward, and seems like a maintained gem as well. Just a question from me.

app/jobs/patient_list_download_job.rb
prabhanshuguptagit: Yes. Tested this.

app/jobs/patient_list_download_job.rb
prabhanshuguptagit: No, not sure why but `Sidekiq::Worker`'s `perform_later` doesn't pass indifferent hashes.

PR #3717, opened 2022-05-09
Allow enabling diabetes management reports at an actor level

govindkrjoshi:
**Story card:** [sc-8057](https://app.shortcut.com/simpledotorg/story/8057/unable-to-turn-on-flipper-for-individual-users)

## Because

We want to be able to enable the DM reports for individual users in order to test and audit the reports in production. There is a bug currently that prevents this. 

## This addresses

Fixes the bug so flipper checks the flag for individual actors

## Test instructions

Enable the flipper flag for a specific user and login as that user. You should be able to see the Diabetes link on the reports page. This link should not be visible for any other user.

### DIFF
--- a/spec/controllers/reports/regions_controller_spec.rb
+++ b/spec/controllers/reports/regions_controller_spec.rb
@@ -214,7 +214,7 @@ RSpec.describe Reports::RegionsController, type: :controller do
     end
 
     context "when region has diabetes management enabled" do
-      before :each do
+      before do
         Flipper.disable(:diabetes_management_reports)
       end
 
@@ -227,7 +227,7 @@ RSpec.describe Reports::RegionsController, type: :controller do
       end
 
       it "contains a link to the diabetes management reports if the feature flag is enabled for an individual user" do
-        enabled_user = create(:admin, full_name: "FOO")
+        enabled_user = create(:admin)
         @facility.update(enable_diabetes_management: true)
         Flipper.enable_actor(:diabetes_management_reports, enabled_user)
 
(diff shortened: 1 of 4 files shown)


spec/controllers/reports/regions_controller_spec.rb
```diff
@@ -214,6 +214,10 @@
     end
 
     context "when region has diabetes management enabled" do
+      before :each do
```
harimohanraj89, line 217: You can drop the `:each`

harimohanraj89 (approved):
Tiny comments, :shipit:

spec/controllers/reports/regions_controller_spec.rb
```diff
@@ -222,13 +226,33 @@
         assert_select "a[href*='diabetes']", count: 1
       end
 
+      it "contains a link to the diabetes management reports if the feature flag is enabled for an individual user" do
+        enabled_user = create(:admin, full_name: "FOO")
```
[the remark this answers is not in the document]
PriyangaPKini: We used `full_name: "FOO"`  for debugging purpose. We shall remove it

spec/controllers/reports/regions_controller_spec.rb
PriyangaPKini: I think we need to keep it because we have to disable the feature flag - `:diabetes_management-reports` before every `it` block.

harimohanraj89: Yup, but the default behavior of `before` is to do a `before :each`. We generally don't explicitly specify the `:each`.

PR #3719, opened 2022-05-09
Add graph to show the diabetes visits breakdown

govindkrjoshi:
**Story card:** [sc-7783](https://app.shortcut.com/simpledotorg/story/7783/add-diabetes-treatment-outcomes-graph)

## Because

We need to add a graph showing the breakdown of diabetic visits

## This addresses

Adds HTML and javascript changes necessary to show the breakdown graph

## Test instructions

1. Turn on diabetes_management_reports on the flipper
2. Visit the reports page of a facility which has enable_diabetes_management set to true
3. Click on the Diabetes link
4. You should see the Treatment outcomes of diabetes patients under care graph

Screenshot 2022-05-10 at 8 37 26 PM

### DIFF
--- a/app/assets/javascripts/common/reports.js
+++ b/app/assets/javascripts/common/reports.js
@@ -1156,7 +1156,7 @@ Reports = function (withLtfu) {
             drawBorder: false,
           },
           ticks: {
-            display: true,
+            display: false,
             autoSkip: false,
             fontColor: this.darkGreyColor,
             fontSize: 10,
@@ -1181,7 +1181,7 @@ Reports = function (withLtfu) {
             drawBorder: false,
           },
           ticks: {
-            display: true,
+            display: false,
             autoSkip: false,
             fontColor: this.darkGreyColor,
             fontSize: 10,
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/common/reports.js
```diff
@@ -1413,6 +1414,208 @@ Reports = function (withLtfu) {
...
+    const barsToDisplay = Math.min(
+        Object.keys(data.bsBelow200Rate).length,
+        maxBarsToDisplay
+    );
+
+    diabetesVisitDetailsGraphConfig.data = {
+      labels: Object.keys(data.bsBelow200Rate).slice(-barsToDisplay),
+      datasets: [
+        {
+          label: "BP <200",
```
prabhanshuguptagit, line 1431: Typo: BS <200, BS 200-299 etc. Not sure where this shows on the page though.

app/assets/javascripts/common/reports.js
```diff
@@ -1413,6 +1414,208 @@ Reports = function (withLtfu) {
...
+          data: Object.values(data.bsOver300Rate).slice(-barsToDisplay),
+          type: "bar",
+        },
+        {
+          label: "Visit but no BP measure",
+          backgroundColor: this.mediumGreyColor,
+          hoverBackgroundColor: this.darkGreyColor,
+          data: Object.values(data.visitButNoBSMeasureRate).slice(
+              -barsToDisplay
+          ),
```
prabhanshuguptagit, line 1457: [nit] This indentation seems off. Could just single line it.

app/models/reports/region_summary_schema.rb
```diff
@@ -226,10 +226,16 @@ def visited_without_bp_taken(with_ltfu: false)
       values_at(field)
     end
 
-    def visited_without_bs_taken
+    memoize def visited_without_bs_taken
```
prabhanshuguptagit, line 229: Why is this `memoize` required?

spec/models/reports/region_summary_schema_spec.rb
```diff
@@ -322,6 +322,45 @@ def refresh_views
       end
     end
 
+    describe "#visited_without_bs_taken_rates" do
```
prabhanshuguptagit, line 325: Slightly unrelated to this PR: We could add a test for the cases where we saw a `0%` in the visits graphs for really small numbers.

config/locales/en.yml
```diff
@@ -66,10 +66,12 @@ en:
     numerator: "Patients with no visit in the last 3 months."
   diabetes_missed_visits_copy:
     reports_card_subtitle: "Diabetes patients in %{region_name} with no visit in the last 3 months"
-    numerator: "Diabetes patients with no visit in the last 3 months."
+    numerator: "Patients with no visit in the last 3 months."
```
prabhanshuguptagit, line 69: Intentional?

prabhanshuguptagit (approved):
Looks great ✨

config/locales/en.yml
PriyangaPKini: yes. In the Figma design for DM reports and the htn cards also use this pattern

spec/models/reports/region_summary_schema_spec.rb
PriyangaPKini: right, we could add it.

app/models/reports/region_summary_schema.rb
PriyangaPKini: CMIIW I thought we were supposed to cache all the functions pulling data from tables.

prabhanshuguptagit: This `memoize` caches return values of methods at the object level (unrelated to caching to redis). We usually add this when the same method is being called multiple times in an object's lifetime, to avoid repeating computation. It comes from this gem https://github.com/tycooon/memery

spec/models/reports/region_summary_schema_spec.rb
PriyangaPKini: Adding a test to check this behaviour in this function will make us create a lot of records. That will make the test suite really slow. So we have added this test to the rounded percentage function to test this behaviour.

app/models/reports/region_summary_schema.rb
govindkrjoshi: This method can get called multiple times in a  given request, which is why we've added the memoize keyword

PR #3721, opened 2022-05-10
Setup May 2022 IHCI experiments

prabhanshuguptagit:
**Story card:** [sc-8220](https://app.shortcut.com/simpledotorg/story/8220)

## Because

We want to start an experiment soon as the current experiment ends.

## This addresses

Sets up the experiments in IHCI.


vkrmis (approved):
Looks good 👍

PR #3722, opened 2022-05-10
Fix data queries for region whatsapp graphics download

vkrmis:
**Story card:** [sc-8035](https://app.shortcut.com/simpledotorg/story/8035/registration-number-for-all-quarters-and-follow-numbers-for-one-quarter-is-missing)

## Because

The downloadable summary (aka "whatsapp graphics") is broken

## This addresses
- Use the correct period (`quarter` instead of `month`)
- Exclude current quarter for the `DashboardAnalytics` query)

### DIFF
--- a/app/controllers/reports/regions_controller.rb
+++ b/app/controllers/reports/regions_controller.rb
@@ -216,7 +216,7 @@ class Reports::RegionsController < AdminController
     @year = params[:year].to_i if params[:year].present?
 
     @cohort_analytics = @region.cohort_analytics(period: :quarter, prev_periods: 3)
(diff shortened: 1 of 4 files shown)


harimohanraj89 (approved):
Looks like a good set of fixes. While we're here, I wanted to ask, how close are we to... removing the analytics queries altogether? Do we still need them?

vkrmis:
> Looks like a good set of fixes. While we're here, I wanted to ask, how close are we to... removing the analytics queries altogether? Do we still need them?

I think we could do it now, everything we need should be available via the reporting matviews. Didn't want to pick it up as part of this bugfix though. Let's schedule it separately?

PR #3726, opened 2022-05-11
Consolidate hypertension reports

harimohanraj89:
**Story card:** [sc-8281](https://app.shortcut.com/simpledotorg/story/8281/consolidate-hypertension-reports)

## Because

Diabetes reports are going to be launched in a new "Diabetes" tab on the reports page. This makes the information architecture of the reports a little weird, with several HTN tabs and one DM tab.

## This addresses

This PR consolidates the Overview, Details, and Cohort Reports pages into a single "Hypertension" page in the reports section. Here are some additional notes about the change.

* The restructuring is only effective when the `diabetes_management_reports` feature flag is turned on. Several controller actions and views and stuff can be deleted when we are removing the feature flag.
* We have chosen to simply display both monthly and quarterly cohort reports one after the other, instead of using a dropdown to toggle between them
* There is some substantial overlap of controller code between the different sections of the hypertension report. We will schedule a cleanup for later. We want to make the restructure right away so we can get ahead of the launch of the DM reports.

## Test instructions

* Turn off the DM flag, all tabs of the reports tab should look the same
* Turn on the DM flag, there should now be only "Hypertension" and "Diabetes" tabs on the reports page (also "Progress" for facility reports)
* The "Hypertension" report should contain the overview stuff ( 4 cards, treatment outcome bar graph, etc), details stuff (registrations and followups, LTFU, etc), and monthly and quarterly cohort reports.

## Screenshots
Screen Shot 2022-05-12 at 10 40 47 PM
Screen Shot 2022-05-12 at 10 40 57 PM
Screen Shot 2022-05-12 at 10 41 06 PM


app/components/reports/cohort_component.html.erb
```diff
@@ -0,0 +1,116 @@
+<div class="card w-100pt mt-lg-0">
+  <div class="mb-24px mb-lg-24px">
+    <h3>
+      <%= cohort_report_type(period) %> cohort
+    </h3>
+    <% if period.type == :month %>
+      <p class="mb-0px c-grey-dark c-print-black">
+        The result for all assigned hypertensive patients registered in a month at their follow-up visit in the following two months
```
govindkrjoshi, line 8: [Suggestion] Should we move these strings into `en.yml`?

govindkrjoshi (approved):

app/components/reports/cohort_component.html.erb
harimohanraj89: We don't have plans to translate the dashboard (verrrrry unlikely) so I'd leave it in here like the rest of our dashboard, avoid the indirection.

PR #3729, opened 2022-05-12
Add BS measurement details

PriyangaPKini:
**Story card:** [sc-7784](https://app.shortcut.com/simpledotorg/story/7784/add-bs-measurement-details-card)

## Because

We need to add a graph showing the BS measurement details

## This addresses

Adds HTML and javascript changes necessary to show the breakdown graph

## Test instructions

1. Turn on diabetes_management_reports on the flipper
2. Visit the reports page of a facility which has enable_diabetes_management set to true
3. Click on the Diabetes link
4. You should see the BS measure details graph

See the screenshots for reference:
1.
Screenshot 2022-05-12 at 8 24 53 PM
2.
Screenshot 2022-05-12 at 8 24 41 PM
3.
Screenshot 2022-05-12 at 8 25 15 PM


PriyangaPKini:
Have some doubts related to this graph @dburka 
-  The order of `percentages` and `counts` of this graph in Figma design for DM reports is different from the performance graph on the overview page. Should we follow the order in Figma design or the one in the performance graph on the overview page?
- We were using a range of colours to show the percentage in the performance graph on the overview page and we are only using 2 colours for DM Reports in the Figma design. Is this intentional?
- In this tooltip, numerator is `number of RBS/PPBS 200-299 patients` and  denominator is `total number of BS 200-299 patients`. In Figma design, denominator was suppose to be `BS <200 + BS 200-299 + BS >=300`.  Since that didn't make sense I changed. Is that ok?
Screenshot 2022-05-12 at 8 24 41 PM

app/models/reports/region_summary_schema.rb
```diff
@@ -302,6 +302,12 @@ def visited_without_bp_taken(with_ltfu: false)
       end
     end
 
+    memoize def diabetes_treatment_outcome_breakdown_counts(blood_sugar_risk_state)
```
harimohanraj89, line 305: Is it kinda weird that this method is called `_counts` and the internal method is called `_by_counts`? I see that the three pieces involved here are:

1. The method that gets the counts
1. The method that calculates rates based on (1)
1. The method that caches (1) 

What naming convention do we follow for HTN stuff? Maybe:

1. The method that gets the counts - `diabetes_treatment_outcome_breakdown_counts`
1. The method that calculates rates based on (1) - `diabetes_treatment_outcome_breakdown_rates`
1. The method that caches (1) - `diabetes_treatment_outcome_breakdown`

WDYT?

harimohanraj89 (changes requested):
One question about the naming we're putting in place

app/models/reports/region_summary_schema.rb
PriyangaPKini: @harimohanraj89, I have made these changes 
- The method that gets the counts - `diabetes_treatment_outcome_breakdown_counts`
   - Private method that gets the counts of a region - `diabetes_treatment_outcome_breakdown_counts_of_region`
- The method that calculates rates based on (1) - `diabetes_treatment_outcome_breakdown_rates`
  - Private method that gets the counts of a region `diabetes_treatment_outcome_breakdown_rates_of_region`

harimohanraj89: Very cool very cool :shipit:

harimohanraj89 (approved):
I like the naming a lot better now 👌 I'm still not sure if it's consistent with the other data elements we do this for, but we can massage that as we go in future PRs. Nice stuff! :100:

dburka (approved):
This looks good. Let's push it live and test in production!

PR #3730, opened 2022-05-12
Changed how large percentage numbers are displayed

dburka:
No card.

Numbers used to align terribly between 2-digit and 3-digit numbers. I made them all inline-block with a min-width. I also reduced the font-size of the registration numbers in DM Reports because when 100,000 patients are displayed in 3 columns, things get funky.

See alignment of the big colorful numbers in the screenshot below.
 
image


dburka:
Note that the smaller numbers in this Registrations card is intentional.
image

PR #3735, opened 2022-05-13
Changed BS to Blood sugar as per request from WHO-India.

dburka:
Changed BS to Blood sugar as per request from WHO-India. Also changed some > < ≥ symbols to HTML entities.

Note: RBS/FBS/PPBS can stay as acronyms!

**To test:**
Make sure the changes don't bust any logic, but I think this should just be cosmetic.

image


harimohanraj89 (approved):
:shipit: Let me know if you want help testing bits of this.

This is also a good clarification to share with the team in Slack, so we don't repeat.

PR #3736, opened 2022-05-15
World Hypertension Day on Progress Tab

dburka:
Praveen asked us to add a special notice for World Hypertension Day on Tuesday, May 17. I added this _entirely inline on purpose_ so we could easily tear it out in 1-2 weeks when we don't want it anymore.

- English-only is OK (Praveen confirmed)
- I thought making this a Flipper flag was overkill
- I also thought adding a date range around it was also overkill. We will manually remove.
- CSS and SVG are inlined on purpose so this is easy to tear out later.
- The size and margin of the "Updated" date at the top has been goofy for months. I fixed this too.

**TESTING**

1. Ensure this is country-specific. I tested it and it should only show up in India.
2. After deploy to Sandbox we should test that the SVG renders correctly on Android... this is a simple SVG with no masks, but sometimes Android chokes on SVGs.

image


prabhanshuguptagit (approved):
:shipit:

PR #3742, opened 2022-05-19
Added more months, formatted better on mobile, removed ticks

dburka:
Showing 6 months is tight on mobile, but works. This is much more useful for seeing the trend.
image
image


harimohanraj89 (approved):
Nice JS linting 😉 

This looks great :shipit:

PR #3743, opened 2022-05-19
Add Blood sugar measurement details table

govindkrjoshi:
**Story card:** [sc-7784](https://app.shortcut.com/simpledotorg/story/7784/add-bs-measurement-details-card)

## Because

We reverted the previous PR that added the BS measurement details table. 

The previous PR had a bug when a region did not have diabetic data, which would throw an uncaught exception. 

This PR fixes this bug, along with some enhancements to the table, so it matches the tables in hypertension reports.

## This addresses

- Adds Blood sugar measurement table
- Fixes bug to handle regions where diabetic data might be missing
- Updates the table to match tables in hypertension reports

## Test instructions

- Turn on diabetes_management_reports on the flipper
- Visit the reports page of a facility which has enable_diabetes_management set to true
- Click on the Diabetes link
- You should see the BS measure details graph

Screenshot from 2022-05-19 15-27-08


app/assets/javascripts/common/reports.js
```diff
@@ -875,14 +875,16 @@ Reports = function (withLtfu) {
       );
       registrationsPeriodEndNode.innerHTML =
         periodInfo.bp_control_registration_date;
-      const breakdown = data.bsBelow200Breakdown[period];
+      const breakdown = data.bsBelow200BreakdownRates[period];
       rbsPPBSPercentNode.innerHTML = this.formatPercentage(
-        breakdown["random"] + breakdown["post_prandial"]
+        breakdown ? breakdown["random"] + breakdown["post_prandial"] : 0
```
govindkrjoshi, line 880: In case breakdown is undefined (there is no data for the period in the current region), there would be an uncaught exception preventing the rest of the JS from rendering. 

We've added this change so that there are no errors when loading the page, and we can default empty graphs to zeros.

app/models/reports/region_summary_schema.rb
```diff
@@ -296,9 +304,41 @@ def visited_without_bp_taken(with_ltfu: false)
...
+    memoize def diabetes_patients_with_bs_taken_breakdown_rates
+      region_period_cached_query(__method__) do |entry|
+        bs_taken_breakdown_hash = {}
+        Reports::PatientBloodSugar.blood_sugar_risk_states.keys.each do |blood_sugar_risk_state|
+          BloodSugar.blood_sugar_types.keys.each do |blood_sugar_type|
+            bs_taken_breakdown_hash[[blood_sugar_risk_state.to_sym, blood_sugar_type.to_sym]] =
+              diabetes_under_care(blood_sugar_risk_state, blood_sugar_type)[entry.region.slug][entry.period]
+          end
+        end
+        rounded_percentages(bs_taken_breakdown_hash)
```
govindkrjoshi, line 341: We weren't able to reuse `diabetes_patients_with_bs_taken_breakdown_counts` here trivially, because `region_period_cached_query` does some grouping and nesting of data for regions and periods.

app/presenters/reports/repository_presenter.rb
```diff
@@ -44,7 +44,9 @@ def to_hash(region)
         bs_below_200_patients: bs_below_200_patients[slug],
         bs_below_200_rates: bs_below_200_rates[slug],
         bs_below_200_with_ltfu_rates: bs_below_200_rates(with_ltfu: true)[slug],
-        bs_below_200_breakdown: diabetes_treatment_outcome_breakdown(:bs_below_200)[slug],
+        bs_below_200_breakdown_rates: diabetes_treatment_outcome_breakdown_rates(:bs_below_200)[slug],
+        bs_200_to_300_breakdown_rates: diabetes_treatment_outcome_breakdown_rates(:bs_200_to_300)[slug],
```
govindkrjoshi, line 48: We added this earlier, but did some refactoring so that 
- bs_200_to_300_breakdown_rates
- bs_200_to_300_breakdown_rates 
are not required in the current PR. 

We are leaving them here, although. Since we'll  need them in a clean up task to add the breakdowns for BS >200 graph.

PR #3745, opened 2022-05-19
Formatting some of the Reports tables to look more consistent

dburka:
Formatting of the tables to look more consistent and be more clear about what is HTN vs DM.

See how the first columns line up in the screenshot. Also, made it more clear that some data is HTN and Overdue represents HTN+DM.

image


dburka:
Note, I intentionally removed a Region name from the translations document. It was unnecessary.

harimohanraj89 (approved):
:shipit:

PR #3752, opened 2022-05-23
24 > 18 months, 10 px font > 12px

dburka:
Praveen asked for this. Now, Reports only have 18 months of data, which lets us use 12 px fonts in the "months" ticks even on mobile.

image


app/models/reports.rb
```diff
@@ -1,6 +1,6 @@
 module Reports
   REGISTRATION_BUFFER_IN_MONTHS = 3
-  MAX_MONTHS_OF_DATA = 24
+  MAX_MONTHS_OF_DATA = 18
```
harimohanraj89, line 3: So happy we made this one number.

harimohanraj89 (approved):
Let's take all region types for a spin and then :shipit:

PR #3754, opened 2022-05-23
New "Recent Updates" and link to get BP Passport Creator app

dburka:
We update the Recent Updates once per month. Also added a link to the new BP Passport creator software for all countries.


dburka:
See new section and new menu option for BP Passport Creator
image

dburka:
See latest "Recent Update." The note about Drug Stock is only being shown in India.

image

prabhanshuguptagit (approved):

PR #3755, opened 2022-05-23
Hotfix failing to render child tables w hen data is missing

govindkrjoshi:
**Story card:** N/A

## Because

There is an exception being thrown when there is missing data for a region / blood sugar type

## This addresses

Defaults the values to 0 if the data is missing

### DIFF
--- a/app/views/reports/regions/_bs_measurement_child_table_row.html.erb
+++ b/app/views/reports/regions/_bs_measurement_child_table_row.html.erb
@@ -14,13 +14,13 @@
     <% if blood_sugar_type == :rbs_ppbs
          patients_percentage_in_category =
            row_data.dig(:diabetes_patients_with_bs_taken_breakdown_rates, period) == 0 ? 0 :
-             (row_data.dig(:diabetes_patients_with_bs_taken_breakdown_rates, period, [blood_sugar_risk_state, :random]) +
-               row_data.dig(:diabetes_patients_with_bs_taken_breakdown_rates, period, [blood_sugar_risk_state, :post_prandial]))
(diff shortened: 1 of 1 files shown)


app/views/reports/regions/_bs_measurement_child_table_row.html.erb
```diff
@@ -14,13 +14,13 @@
     <% if blood_sugar_type == :rbs_ppbs
          patients_percentage_in_category =
            row_data.dig(:diabetes_patients_with_bs_taken_breakdown_rates, period) == 0 ? 0 :
-             (row_data.dig(:diabetes_patients_with_bs_taken_breakdown_rates, period, [blood_sugar_risk_state, :random]) +
-               row_data.dig(:diabetes_patients_with_bs_taken_breakdown_rates, period, [blood_sugar_risk_state, :post_prandial]))
+             ((row_data.dig(:diabetes_patients_with_bs_taken_breakdown_rates, period, [blood_sugar_risk_state, :random]) || 0)+
```
prabhanshuguptagit, line 17: Should the repository be defaulting this to 0 already? Asking since the view should be dumb if possible.

vkrmis (approved):
Could also default to zero when constructing the map, but I like this better. Looks good 👍

app/views/reports/regions/_bs_measurement_child_table_row.html.erb
govindkrjoshi: I need to investigate why this is returning nil. This is a quick hotfix to unblock the production issues. If it's genuinely possible to return nil from here, I'll move the defaulting to the repository.

PR #3760, opened 2022-05-24
Added a disabled DOWNLOADS button to DM reports. 

dburka:
No story. It's just confusing that we have a Downloads dropdown with HTN downloads on the DM Reports. See screenshot below for the disabled button.

I also removed a completely unnecessary print-only menu that is now extra confusing with the DM Reports. I removed everything after "Apr-2022 Report" in this code (it's totally not needed).
```
 <p class="d-none fw-bold d-print-block">
    <%= @period %> Report — <% if params[:action] == "show" %>Overview<% elsif params[:action] == "details" %>Details<% elsif params[:action] == "cohort" %>Cohort Reports<% end %>
  </p>
```

image

Test that this doesn't screw up anything when `diabetes_management_reports` are enabled/disabled in Flipper. I tested, but good to have more eyes on it.


harimohanraj89 (approved):
Looks great. I took the liberty to rename the partial to "hypertension_downloads", as that's more appropriate now. I also cleaned up the decision tree a tiny bit.

:shipit:

PR #3766, opened 2022-05-27
Reorganized HTN and DM Reports cards

dburka:
**Story card:** 
Praveen requested that we re-order the cards in HTN reports here on Slack: https://simpledotorg.slack.com/archives/CK95QFJA2/p1653633373452729

## Important notes

- [x] Praveen and WHO asked us to remove Monthly Cohort Reports! This is good because it reduces complexity. We should create an Epic to remove them from the code.
- [x] I removed link to download Monthly Cohort Report data as well
- [x] The DM reports will need the same re-ordering. I can do this, but will wait until @govindkrjoshi and @PriyangaPKini are done with the tables so I don't cause merge conflicts.
- [x] I tested these changes with DM reports enabled/disabled and I think we're good, but please triple check.


harimohanraj89 (approved):
Looks great and consistent 👍 I took the liberty of ripping out some more remnants of the monthly cohort report from the controller.

PR #3772, opened 2022-05-30
Added coming soon to cards that don't exist yet

dburka:
Added quite a few missing "Coming soon" cards to show what's coming. 
image

### DIFF
--- a/app/assets/stylesheets/partials/_dashboard.scss
+++ b/app/assets/stylesheets/partials/_dashboard.scss
@@ -23,6 +23,24 @@ Dashboard UI
   background: #fff;
 }
 
+.empty {
+  background: $green-light;
+  color: $green-dark;
+  text-align: center;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/partials/_dashboard.scss
```diff
@@ -23,6 +23,24 @@ Dashboard UI
   background: #fff;
 }
 
+.empty {
```
harimohanraj89, line 26: Want to use a name like "coming-soon" or "preview"? I can see us using "empty" to signify a card that doesn't have any data to show yet (like at a facility that hasn't registered patients yet), and I don't think we'd want the green highlighting for those.

dburka: Good idea. Fixing

harimohanraj89 (approved):
:shipit: Looks great.

PR #3782, opened 2022-06-02
Add graph for diabetes days of medication

PriyangaPKini:

## Because

## This addresses

## Test instructions


app/views/reports/regions/diabetes.html.erb
```diff
@@ -1,3 +1,4 @@
+<script src="https://cdn.jsdelivr.net/npm/chart.js@2.8.0/dist/Chart.min.js" integrity="sha256-Uv9BNBucvCPipKQ2NS9wYpJmi8DTOEfTA/nH2aoJALw=" crossorigin="anonymous"></script>
```
PriyangaPKini, line 1: We are using a plugin called `ChartDataLabel`. For this plugin to work we have to load the JS script first.

spec/services/medication_dispensation_service_spec.rb
```diff
@@ -12,7 +12,45 @@
```
vkrmis, line 12: Let's explicitly make this patient hypertensive

spec/models/reports/region_summary_schema_spec.rb
```diff
@@ -76,37 +76,51 @@ def refresh_views
...
       expect(schema.appts_scheduled_0_to_14_days_rates[facility.slug][range.last]).to eq(100)
+
+      expect(schema.diabetes_appts_scheduled_0_to_14_days_rates[facility.slug][range.first]).to eq(0)
+      expect(schema.diabetes_appts_scheduled_0_to_14_days_rates[facility.slug][range.to_a.second]).to eq(0)
+      expect(schema.diabetes_appts_scheduled_0_to_14_days_rates[facility.slug][range.last]).to eq(100)
     end
 
     it "returns percentages of appointments scheduled in a month in the given range" do
       facility = create(:facility)
-      patient = create(:patient, assigned_facility: facility, recorded_at: 4.month.ago)
```
vkrmis, line 93: We're currently only testing the case where appointments are being scheduled across different months. Let's add a spec where we test multiple appointments created in the same month, but across buckets. This would test that we're passing in the right data to `rouded_percentages`

spec/services/medication_dispensation_service_spec.rb
PriyangaPKini: ok

PR #4109, opened 2022-06-27
A/B: Replace running with enrolling

prabhanshuguptagit:
**Story card:** -

## Because

An experiment is currently described as `running` while it is enrolling patients. This is slightly misleading since an experiment is actually "running" up until 15 days after enrollment while patients are being monitored.

## This addresses

This does away with the `running` phase and replaces it with a more clear `enrolling` phase.


vkrmis (approved):
Looks good

PR #4110, opened 2022-06-27
A/B: Support repeating messages

prabhanshuguptagit:
**Story card:** [sc-8551](https://app.shortcut.com/simpledotorg/story/8551/setup-jul-1-basic-variations-experiment)

## Because

A/B experiments are currently unable to support repeating a message with the same content because the message names are intertwined with how we report message statuses in treatment group memberships. 

## This addresses

The JSON blob that reports message status in the `messages` column in `treatment_group_memberships` was earlier keyed by message locale names. This meant that we could only send one message once to a patient in a treatment group. This fixes the reporting to work with `reminder_template` IDs as the key. Experiments which start from 1st Jul onwards will report in this new schema. This is in order to not break any reporting for currently running experiments (which is done via metabase dashboards). We will use a new dashboard to report experiments after Jul 2022.


app/models/communication.rb
```diff
@@ -30,10 +30,6 @@ class Communication < ApplicationRecord
   DEFAULT_MESSAGING_START_HOUR = 14
   DEFAULT_MESSAGING_END_HOUR = 16
 
-  scope :with_delivery_detail, -> {
-    joins("inner join twilio_sms_delivery_details delivery_detail on delivery_detail.id = communications.detailable_id")
-  }
-
```
prabhanshuguptagit, line 36: This is an unrelated unused scope.

app/models/experimentation/notifications_experiment.rb
```diff
@@ -81,17 +84,25 @@ def monitor
       end
     end
 
+    def messages_report_key
+      if start_time > NEW_MESSAGE_REPORTING_DATE
+        NEW_MESSAGE_REPORTING_DATE
```
vkrmis, line 89: Should this be `NEW_MESSAGE_REPORTING_KEY`?

app/models/experimentation/notifications_experiment.rb
```diff
@@ -81,17 +84,25 @@ def monitor
...
+      else
+        OLD_MESSAGE_REPORTING_KEY
+      end
+    end
+
     def record_notification_results
       # TODO: Look at query performance
```
vkrmis, line 96: Is this still pending?

prabhanshuguptagit: Haven't looked at it closely but it hasn't caused any problems even with 50k messages/day. It's decently quick to finish. Will remove it.

app/models/experimentation/notifications_experiment.rb
prabhanshuguptagit: Fixed in [0729c29](https://github.com/simpledotorg/simple-server/pull/4110/commits/0729c29ae51489e55337bb33a724c575ea462807)

app/models/experimentation/reminder_template.rb
```diff
@@ -5,6 +5,5 @@ class ReminderTemplate < ActiveRecord::Base
 
     validates :message, presence: true
     validates :remind_on_in_days, presence: true, numericality: {only_integer: true}
-    validates :message, uniqueness: {scope: :treatment_group, message: "already exists in this treatment group"}
```
vkrmis, line 8: 👌

app/models/experimentation/notifications_experiment.rb
```diff
@@ -81,17 +84,24 @@ def monitor
...
     end
 
+    def messages_report_key
+      if start_time > NEW_MESSAGE_REPORTING_DATE
+        NEW_MESSAGE_REPORTING_KEY
+      else
+        OLD_MESSAGE_REPORTING_KEY
+      end
+    end
+
```
vkrmis, line 94: Can we do a data migration to get rid of this? Doesn't have to be a part of this PR, can come later. We'd have to rewrite the JSON blobs and key them by template ids instead.

vkrmis (approved):
Looks good, recommend a future PR so we can ditch the variable message reporting key

app/models/experimentation/notifications_experiment.rb
prabhanshuguptagit: We could but there's ~3M membership records so we'd have to think about a way to do it that's sufficiently performant, hence punting it.

PR #4111, opened 2022-06-27
Add pending facility graphs to diabetes dashboard

govindkrjoshi:
**Story card:** 
- [sc-8295](https://app.shortcut.com/simpledotorg/story/8295/in-dm-reports-for-a-facility-add-patient-registration-and-follow-up-by-users-table)
-  [sc-8296](https://app.shortcut.com/simpledotorg/story/8296/in-diabetes-reports-for-a-facility-add-health-care-worker-activity-section)
- [sc-8297](https://app.shortcut.com/simpledotorg/story/8297/in-diabetes-reports-for-a-facility-add-the-recent-blood-sugar-measurement-tables)

## Because

These are pending graphs on the diabetes dashboard which were not part of the original scope

## This addresses

Adds cards to facility diabetes dashboard for
- Registrations and follow ups by user
- Healthcare worker activity
- Recent blood sugar measurements at a facility

It also adds a card to the admin/user page for recent blood sugar measurements by a user. 

## Test instructions

- Checkout this branch
- Turn on `:diabetes_management_enabled` flipper flag
- Look at the diabetes reports for a facility. The facility should have `enable_diabetes_management` set to true
- You should see the following graphs
  - Registrations and follow ups by user
  - Healthcare worker activity
  - Recent blood sugar measurements at a facility
 - Click on a username. It should take you to an admin/user page. Here you should be able to the recent blood sugar measurements by a user.


app/components/application_component.rb
```diff
@@ -0,0 +1,2 @@
+class ApplicationComponent < ViewComponent::Base
```
prabhanshuguptagit, line 1: Seems like a good idea. Should we modify the other views that already inherit from `ViewComponent::Base`?

prabhanshuguptagit, line 1: Seems like a good idea. Should we modify the other views that already inherit from `Base`?

app/components/dashboard/card/title_component.rb
```diff
@@ -0,0 +1,11 @@
+class Dashboard::Card::TitleComponent < ApplicationComponent
```
prabhanshuguptagit, line 1: This is quite neat.

app/components/dashboard/card/tooltip_component.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    data-toggle="tooltip"
+    data-placement="right"
+    data-trigger="hover click"
+    data-html="true"
+    <% if definitions.is_a? String %>
+      title="<%= definitions %>"
+    <% else %>
+      title="<% definitions.each do |name, description| %>
+               <p class='mb-4px'><strong><%= name %>:</strong> <%= description %></p>
+             <% end %>"
```
prabhanshuguptagit, line 13: This logic can belong to a `render_definitions` method in the component. 

What's the case where we want to pass in a string? Wasn't able to find one in this PR.

app/components/dashboard/diabetes/recent_measurements_component.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  attr_reader :region, :user, :recent_blood_sugars, :display_model, :page
+
+  def initialize(recent_blood_sugars:, display_model:, page:, region: nil, user: nil)
+    @region = region
+    @user = user
+    @recent_blood_sugars = recent_blood_sugars
+    @display_model = display_model
+    @page = page
+    @recent_blood_sugars.first.risk_state
```
prabhanshuguptagit, line 12: This failed with an `undefined method 'risk_state' for nil:NilClass` if there's no blood sugars recorded at a facility.

app/helpers/reports/regions_helper.rb
```diff
@@ -1,7 +1,7 @@
 module Reports::RegionsHelper
   def sum_registration_counts(repository, *keys)
-    slug, user_id = keys
-    repository.monthly_registrations_by_user.dig(slug)
+    slug, user_id, diagnosis = keys
```
prabhanshuguptagit, line 3: I didn't quite follow the reason to accept arguments as a splatted array here (vs just regular keyword params). `diagnosis` can be defaulted to much more nicely if we could use arguments.

app/models/blood_sugar.rb
```diff
@@ -31,19 +31,40 @@ class BloodSugar < ApplicationRecord
...
+    bs_below_200:
+      {random: 0..200, post_prandial: 0..200, fasting: 0..126, hba1c: 0..7},
+    bs_200_to_300:
+      {random: 200..300, post_prandial: 200..300, fasting: 126..200, hba1c: 7..9},
+    bs_over_300: {
+      random: 300..Float::INFINITY,
+      post_prandial: 300..Float::INFINITY,
+      fasting: 200..Float::INFINITY,
+      hba1c: 9..Float::INFINITY
+    }
```
prabhanshuguptagit, line 49: If I understood correctly
- This adds the notions of `bs_below_200` and `bs_200_to_300`.
- It does away with the `high` terminology and calls it `bs_over_300` now.
- We're adding it to support showing the states of individual blood sugars in the activity log.

Is that correct?

app/queries/registered_patients_query.rb
```diff
@@ -30,6 +25,16 @@ def count(region, period_type, group_by: nil)
 
   private
 
+  # TODO: Why not materialized views?
```
prabhanshuguptagit, line 28: I'm guessing this is because we want to show these in realtime?

prabhanshuguptagit (commented):
Looks fantastic. I've done a first pass, will come back again for a more thorough review.

app/components/dashboard/diabetes/recent_measurements_component.html.erb
```diff
@@ -0,0 +1,72 @@
...
+            <td class="nowrap" style="width: 10em; <%= "border: none;" if !print_date %>">
+              <% if print_date %>
+                <strong>
+                  <%= l blood_sugar.recorded_at.to_date %>
+                </strong>
+              <% end %>
+            </td>
+            <td class="nowrap text-muted"><%= blood_sugar.recorded_at.strftime("%I:%M %p") %></td>
+            <td class="nowrap <%= risk_state_color_class(blood_sugar) %>">
+              <i class="fas fa-heart mr-2"></i> <%= blood_sugar.to_s %> <%= blood_sugar.blood_sugar_type.to_s.capitalize %></td>
```
prabhanshuguptagit, line 44: `Post prandial` shows up with an underscore
image

app/views/reports/regions/diabetes.html.erb
```diff
@@ -286,31 +286,26 @@
...
+    period_range: @details_period_range,
+    current_admin: current_admin
+  ) %>
+
+  <%= render Dashboard::Diabetes::RecentMeasurementsComponent.new(
+    region: @region,
+    recent_blood_sugars: @recent_blood_sugars,
+    display_model: :facility,
+    page: @page
+  ) %>
```
prabhanshuguptagit, line 308: ✨

app/queries/registered_patients_query.rb
```diff
@@ -6,18 +6,13 @@ class RegisteredPatientsQuery
...
   # Returns a count of registered patients over the past last days
   def count_daily(region, last:, diagnosis: :hypertension)
-    scope = case diagnosis
-    when :all then region.registered_patients
-    when :hypertension then region.registered_hypertension_patients
-    when :diabetes then region.registered_diabetes_patients
-    else raise ArgumentError, "unknown diagnosis #{diagnosis}"
-    end
-    scope.group_by_period(:day, :recorded_at, last: last).count
+    registered_patients_with_diagnosis(region, diagnosis)
```
prabhanshuguptagit, line 9: 👌🏼

prabhanshuguptagit (approved):
Looks great. :shipit: when addressed.

app/components/application_component.rb
PriyangaPKini: yeah, we should

app/components/dashboard/card/tooltip_component.html.erb
PriyangaPKini: > What's the case where we want to pass in a string? Wasn't able to find one in this PR.
When we need to render just a summary instead of a key-value pair type definition.
Like the tooltips in the `hypertension/diabetes patients` card under the details section

app/components/dashboard/diabetes/recent_measurements_component.rb
govindkrjoshi: This is some code we were using to debug, which has sneaked in to the the PR. Removing this.

app/helpers/reports/regions_helper.rb
govindkrjoshi: We will make it keyword params, it would be both easier to read and use.

app/models/blood_sugar.rb
govindkrjoshi: Yes.

PR #4112, opened 2022-06-27
Add translations for missed visits in the progress tab

govindkrjoshi:
**Story card:** [sc-8241](https://app.shortcut.com/simpledotorg/story/8241/include-missed-visits-in-progress-tab-cohort-reports)

## Because

We added a new string to the progress tab which was missing translations. 

## This addresses

This adds translations for the new string in other languages, except for `sid_ET`


PriyangaPKini (approved):
lgtm👍🏻

config/locales/api/am_ET.yml
```diff
@@ -1,4 +1,4 @@
-am-ET:
+am_ET:
```
harimohanraj89, line 1: IIRC, this key should not change. It should remain hyphenated. Let's revert this diff. 

Transifex annoyingly makes this diff every time ([documented here](https://github.com/simpledotorg/simple-server/blob/master/doc/wiki/transifex-localization.md#3-check-and-fix-all-top-level-language-keys))

harimohanraj89 (changes requested):
Blocking review.

harimohanraj89 (approved):
:shipit:

PR #4125, opened 2022-07-12
Use view components for dashboard cards

govindkrjoshi:
**Story card:** [sc-8179](https://app.shortcut.com/simpledotorg/story/8179/refactor-report-js)

## Because

We had duplicated view logic for a lot of the dashboard components

## This addresses

- Extract dashboard views into view components
- Adds a common graph component
- Adds a common table component


app/assets/javascripts/common/reports.js
```diff
@@ -1,77 +1,174 @@
...
-  this.lightGreyColor = "rgba(240, 242, 245, 0.9)";
-  this.whiteColor = "rgba(255, 255, 255, 1)";
-  this.amberColor = "rgba(250, 190, 70, 1)";
-  this.darkAmberColor = "rgba(223, 165, 50, 1)";
-  this.transparent = "rgba(0, 0, 0, 0)";
-  this.tealColor = "rgba(48, 184, 166, 1)";
-  this.darkTealColor = "rgba(34,140,125,1)";
-  this.maroonColor = "rgba(71, 0, 0, 1)";
-  this.darkMaroonColor = "rgba(60,0,0,1)";
+const COLORS = {
```
prabhanshuguptagit, line 1: The `Reports` "class" (closure) acted as a namespace for all the methods in this class. These methods (and variables like `COLORS`) will now be available globally since all our scripts are bundled together. Can we keep the `.listen()` interface around with this refactor to prevent name collisions?

spec/components/dashboard/card/graph_component_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+RSpec.describe Dashboard::Card::GraphComponent, type: :component do
+  pending "add some examples to (or delete) #{__FILE__}"
+
+  # it "renders something useful" do
```
prabhanshuguptagit, line 6: What should these tests be checking?

prabhanshuguptagit (approved):
Looks fantastic. One comment around keeping the javascript scoped. 

I think a small readme on how to add a new graph/table using this setup might be useful.

app/assets/javascripts/common/reports.js
PriyangaPKini: yeah.  Done

PriyangaPKini:
> I think a small readme on how to add a new graph/table using this setup might be useful.

@govindkrjoshi will be adding a readme for this soon

PR #4134, opened 2022-07-18
Add blood sugar type breakdown for BS Over 200 graph

govindkrjoshi:
**Story card:** [sc-8229](https://app.shortcut.com/simpledotorg/story/8229/add-breakdown-by-blood-sugar-type-for-bs-200-graph)

## Because

We want to show the breakdown of a BS over 200 trends by blood sugar type

## This addresses

Updates the view and report js to add these breakdowns to bs over 200 graph

## Test instructions

You should see the breakdown count at the bottom of the BS over 200 graph. The percentages should change as you hover over the graph

Screenshot from 2022-07-18 13-51-42

### DIFF
--- a/app/assets/javascripts/common/reports.js
+++ b/app/assets/javascripts/common/reports.js
@@ -1017,12 +1017,9 @@ Reports = function (withLtfu) {
         "[data-registrations-period-end]"
       );
 
-      const bs200T0300RbsPPBSPercentNode = cardNode.querySelector("[data-over-300-rbs-ppbs]");
-      const bsOver300RbsPPBSPercentNode = cardNode.querySelector("[data-bs-200-to-300-rbs-ppbs]");
-      const bs200T0300FastingPercentNode = cardNode.querySelector("[data-over-300-fasting]");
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/common/reports.js
```diff
@@ -1048,6 +1055,28 @@ Reports = function (withLtfu) {
       registrationsPeriodEndNodes.forEach(
         (node) => (node.innerHTML = periodInfo.bp_control_registration_date)
       );
+
+      const bs200To300breakdown = data.bs200To300BreakdownRates[period];
+      const bsOver300breakdown = data.bsOver300BreakdownRates[period];
+
+      bs200T0300RbsPPBSPercentNode.innerHTML = this.formatPercentage(
+          bs200To300breakdown ? bs200To300breakdown["random"] + bs200To300breakdown["post_prandial"] : 0
```
prabhanshuguptagit, line 1063: The calculation can be separated into a separate method from the `innerHTML` update bit since doing this sort of calculation seems unusual for reports.js.

prabhanshuguptagit (approved):

PR #4137, opened 2022-07-20
Diabetes dashboard fix:  BP > Blood Sugar

jamiecarter7:
Change 'BP' to 'Blood Sugar' and the associated ascii code for the symbol


app/views/reports/regions/_diabetes_footnotes.html.erb
```diff
@@ -45,7 +45,7 @@
       </div>
       <div class="mb-24px">
         <p class="mb-8px fs-16px fw-bold c-grey-dark">
-          BP <200
+          Blood Sugar &lt;200
```
dburka, line 48: Tiny thing @jamiecarter7 but we generally use "Sentence case" capitalization... so the "Sugar" in each of these should be "sugar"

jamiecarter7: I will correct to sentence case. @dburka

jamiecarter7:
I will update the cards with the associated ampersand icon.
I will check the BP page and update those as well if need be.

jamiecarter7:
Ok, after having a dig around it looks like the rest of the icons are written as strings stringName = "<200" so they are correctly rendered in the app. They also wont break anything as they don't break any open tags.
Only the HTML based icons needed to be changed.
Ampersand icons do not actually render correctly when defined in the string. The just render directly as the text (eg. '&lt;' rather than '<')

PR #4138, opened 2022-07-20
Attempted fix for .DS_Store files, as they are still appearing

jamiecarter7:
## Because

DS_Store files still creeping into my work flow

## This addresses

Added additional rule to git-ignore '**/DS_Store'

## Test instructions

Will test going forward


harimohanraj89:
Hmm, `.DS_Store` is supposed to be the same as `**/.DS_Store`. Did it fix your problem locally when you updated the .gitignore file?

govindkrjoshi:
This is github's recommended gitignore file for macos: https://github.com/github/gitignore/blob/ce6f84024931408ce801808fe9f4587f7588b283/Global/macOS.gitignore

FYI, You can also add a global gitignore for your computer incase you don't want to add a gitignore to project file https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

PR #4159, opened 2022-07-25
Readme on running A/B experiments

prabhanshuguptagit:
**Story card:** [sc-7859](https://app.shortcut.com/simpledotorg/story/7859/polish-bsnl-related-documentation)

## This addresses

Adds a manual for running A/B experiments. Deletes some outdated documentation.


doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,78 @@
+# Running SMS reminder experiments
+
+## Prerequisites
+To start an SMS experiment you'll need to gather the following information:
+
+- A `name` for the experiment.
+- The `type` of experiment - it can target either "Current" patients or "Stale" patients.
```
harimohanraj89, line 7: Let's include what "current" and "stale" mean here itself.

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,78 @@
...
+- The number of `max_patients_per_day` that will be enrolled per day.
```
harimohanraj89, line 8: We could make all of these requirements in the format:
```
- `the_thing`: What the thing means
```

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,78 @@
...
+- `start_time` - enrollments will begin at this time.
+- `end_time` - enrollments will stop at this time .
+- `treatment_groups` - Each treatment group will have a different content or frequency of messages sent to it.
+    - You will probably also need a `control` group as a baseline in addition to all the groups that will receive messages.
+- The content and frequency of messages in each treatment group. This will be captured in the
+  `reminder_templates` of each group.
+  For current patients, the `remind_on` of reminder templates is set relative to the date
+  they are expected to return. For stale patients, it's the date of enrollment.
```
harimohanraj89, line 16: This bullet is still a bit opaque to new folks. Can we elaborate in great detail with some examples? It would be very helpful to show how to set up a 3-message cascade for current patients for, say, -1, 0, and +3 days from appt.

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,78 @@
...
+**Note about consecutive experiments**
+
+An experiment that sends notifications in advance enrolls patients who have appointments in the future.
+For an experiment starting right after, this can mean that all the patients might've already been enrolled
+in the first one. Make sure there's at least a gap of these many days (only if this value is positive) to avoid this
+```ruby
+ experiment_1 = Experimentation::Experiment.find("old experiment's id")
+ experiment_2 = Experimentation::Experiment.find("new experiment's id")
+ experiment_1.earliest_remind_on - experiment_2.earliest_remind_on
+```
```
harimohanraj89, line 35: I think we can be clearer about this problem, and may not need the "these many days" math.

Could we say something like:
When scheduling experiments one after the other, ensure that there is enough time between experiments for...

I actually don't exactly understand the problem either, from this description 🤔 Could you explain this to me?

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Running the experiment
+
+A cron in `schedule.rb` orchestrates the experiment. It runs a set of tasks everyday.
+When an experiment starts, you should
+- Check in on the metabase dashboards every once in a while.
```
harimohanraj89, line 41: Let's add links to the dashboards for the countries we have (India only I guess?)

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,78 @@
...
+- We get a daily summary on slack on #ab-testing-stats. Check on the "Pending notifications" report everyday. 
+  This number should always be 0. If not, this points to an issue in the system.
```
harimohanraj89, line 43: Consider:

If not, some reminder messages are not being sent to patients and this problem requires investigation. Inspect error reports in Sentry (add link) to find and fix the problem.

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,78 @@
...
+Notes:
+- Depending on the cadence, notifications may go out for a few days after the experiment has "ended".
+- Visits are monitored and patients are evicted until 15 days (`MONITORING_BUFFER`) from the last enrollment date.
+
+### Important links
+
+- [IHCI metabase dashboard](https://metabase.simple.org/dashboard/54-notifications-experiment-generic-dashboard)
+- [Bangladesh metabase dashboard](https://metabase.bd.simple.org/dashboard/10-notifications-experiment-generic-dashboard)
+- [IHCI CSV export](https://metabase.simple.org/question/496-a-b-experiments-statistical-analysis-report)
+- [Bangladesh CSV export](https://metabase.bd.simple.org/question/132-a-b-experiments-nhf-statistical-analysis-report)
```
harimohanraj89, line 57: Let's keep these links here too, but also add them above inline where we mention "go look at the metabase dashboards"

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,78 @@
...
+
+If you're adding any new reports to the IHCI dashboard save them in this [collection](https://metabase.simple.org/collection/43-a-b-testing-ihci-shared)
+so it's viewable by other people.
+
+### Cancelling an experiment
+ 
+```ruby
+Experimentation::NotificationsExperiment.find("experiment id").cancel
+```
```
harimohanraj89, line 66: Let's add some words about when you'd want to cancel and experiment, is it safe to do so at any time, what this does to the experiment (eg. "all remaining reminder messages will not be sent" etc). Any data artifcats that this creates, any cleanup required, can you schedule another experiment right away

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,78 @@
...
+
+## In the long run
+
+- When we decide on a message format and start sending a single message eventually, the current plan
+  is to run an "experiment" with a single bucket. Although, a patient can be enrolled in an experiment only once.
+  A single long running experiment will hence not work for patients who need to follow up every month.
+  A new experiment will need to be started every month.
```
harimohanraj89, line 73: Is this actually the plan? Wouldn't a default behavior be easier to maintain? Setting up an experiment every month is definitely not tenable.

harimohanraj89 (approved):
Preemptive approval to unblock you, but several feedbacks below!

doc/howto/sms_reminder_experiments.md
prabhanshuguptagit: This is the default plan since it'll be the least amount of work (as opposed to modifying the reminder service, setting up monitoring around it etc). Once we arrive at a long term messaging strategy we can setup the experiment automatically every month, it won't need any recurring effort.

doc/howto/sms_reminder_experiments.md
prabhanshuguptagit: 🙏🏼

doc/howto/sms_reminder_experiments.md
prabhanshuguptagit: Technically, it's allowed to have experiments start right next to each other. Although, if the first experiment sends reminders _before_ their appointment it'll enroll people who have appointments "in the future". If the second experiment wants to send reminders _after_ appointments is missed, there won't be any eligible patients left to enroll because they all got enrolled by the previous experiment. This causes the second experiment to end up with a tiny sample size.

Could maybe explain this over a call and you could help me write this, a little out of words 😛

doc/howto/sms_reminder_experiments.md
harimohanraj89: Hmmm easier to set up but requires more ongoing maintenance I guess. Let's keep the docs and the plan like this, we can adjust later if needed.

doc/howto/sms_reminder_experiments.md
harimohanraj89: Yes I need explanation haha. We can update offline. Ship this thing before it stale-ifies any more.

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,108 @@
...
+  Each treatment group will have a different content or frequency of messages sent to it.
+- You will probably also need a `control` group that doesn't receive any messages as a baseline.
+
+#### Reminder templates
+
+- The content and frequency of messages in each `treatment_group` is determined by it's `reminder_templates`.
+  Each reminder template has
+  - a `message` - The locale key of the message (for ex. `notifications.set01.basic`).
+  - a `remind_on` -  The day on which this message will be sent out relative to the expected date of return.
+- If the experiment has new messages and translations you'll need to get them [approved by DLT](doc/howto/bsnl/sms_reminders.md) and make sure they're present in
```
olttwa, line 28: Add more info:
> DLT is only for India.

> The text must be added to `config/locales/notifications`. Run `get_bsnl_templates.rake` to generate the `config/data/bsnl_templates.yml`

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,108 @@
...
+
+A cron in `schedule.rb` orchestrates the experiment. It runs a set of tasks everyday.
+When an experiment starts, you should
+- Check in on the metabase dashboards ([IHCI](https://metabase.simple.org/dashboard/54-notifications-experiment-generic-dashboard),
+  [Bangladesh](https://metabase.bd.simple.org/dashboard/10-notifications-experiment-generic-dashboard))
+  every once in a while.
+- We get a daily summary on slack on #ab-testing-stats. Check on the "Pending notifications" report everyday. 
+  This number should always be 0. If not, some reminder messages are not being sent to patients and this problem requires investigation.
+  Inspect error reports in [Sentry](https://sentry.io/organizations/resolve-to-save-lives/issues/?project=1217715) to find and fix the problem.
+- Run `rake:get_account_balance` every once in a while, check the balance and recharge if needed.
```
olttwa, line 75: s/get_account_balance/bsnl_ get_account_balance

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,108 @@
...
+
+**Notes**
+
+- Depending on the cadence, notifications may go out for a few days after the experiment has "ended".
+- Visits are monitored and patients are evicted until 15 days (`MONITORING_BUFFER`) from the last enrollment date.
```
olttwa, line 80: monitoring SOP & reason for eviction should be documented.

doc/howto/sms_reminder_experiments.md
```diff
@@ -0,0 +1,108 @@
...
+#### Experiment
+
+- `name` - A name for the experiment.
+- `type` - There are two kinds of experiments depending on the type of the patients that will be enrolled.
+  - `current_patients` - Patients who have an upcoming appointment.
+  - `stale_patients` - Patients who have not visited in the last year and do not have a scheduled appointment in the future.
+- `max_patients_per_day` - The maximum number of patients to be enrolled per day.
+- `start_time` - Enrollments will begin at this point in time.
+- `end_time` - Enrollments will stop at this point in time.
+
```
olttwa, line 15: Add a specimen command to create an experiment.

PR #4170, opened 2022-07-29
[Quick fix]Fix org level diabetes link

PriyangaPKini:
## Because

The organization-level diabetes link is broken. Organization-level slug and region-level slug for `Sri Lanka is different.

## This addresses

This is a workaround for fix.


dburka (approved):
Looks good. Let's try and revert if we screw anything up.

PR #4173, opened 2022-08-01
Fix failing test

PriyangaPKini:
## Because

Some tests were failing because of the month change

## This addresses

Freeze the time at a particular date


prabhanshuguptagit (approved):
:shipit:

PR #4207, opened 2022-08-05
Fix drugstock color classes

prabhanshuguptagit:
**Story card:** [sc-8138](https://app.shortcut.com/simpledotorg/story/8138/drug-stock-color-highlights-are-broken)

## Because

The label colors for drug stock counts don't match.
image

## This addresses

Fixes the highlight colors.


harimohanraj89 (approved):
Looks like a classic bad merge.

PR #4221, opened 2022-08-23
Add check for if `new_progress_tab` is enabled for a single actor

PriyangaPKini:
## Because

The check for if `new_progress_tab` is enabled for a single actor was missing

## This addresses

Adds check for if `new_progress_tab` is enabled for a single actor

## Test instructions

> Settings> Flipper> Enable feature flag `new_progress_tab` for a single actor
> Sign in as that actor
> You should be able to see the new progress tab

## Note
These changes are for development purposes and will be reverted once the feature is released


govindkrjoshi (approved):
Since these are all view components you could also use conditional rendering instead of having the `if` clause in the erb file. 

https://viewcomponent.org/guide/conditional_rendering.html

PriyangaPKini:
> Since these are all view components you could also use conditional rendering instead of having the `if` clause in the erb file.
> 
> https://viewcomponent.org/guide/conditional_rendering.html

I think we cannot use `#render?` because we have `if else` condition.

PR #4222, opened 2022-08-25
Don't sync appointment patients to blocks

prabhanshuguptagit:
**Story card:** [sc-8970](https://app.shortcut.com/simpledotorg/story/8970/remove-appointment-patients-from-list-of-syncable-patients)

## Because

The appointments table is expensive to query in `syncable_patients` because IHCI has a lot of records. https://app.shortcut.com/simpledotorg/write/IkRvYyI6I3V1aWQgIjYzMDcxZTA2LTcwZjEtNDI5Yy04NmUyLTdlMmNjNTNiZmViMCI=

## This addresses

Removes appointed patients from syncable patients.


app/controllers/api/v3/patients_controller.rb
```diff
@@ -27,12 +27,10 @@ def other_facility_records
     time(__method__) do
       other_facilities_limit = limit - current_facility_records.size
       @other_facility_records ||=
-        Patient
-          .where(id: current_sync_region.syncable_patients)
+        current_sync_region.syncable_patients
           .where.not(registration_facility: current_facility)
           .for_sync
           .updated_on_server_since(other_facilities_processed_since, other_facilities_limit)
-          .order(:id) # needed for performance, see https://stackoverflow.com/questions/21385555/postgresql-query-very-slow-with-limit-1/27237698
```
prabhanshuguptagit, line 35: Don't need this anymore since we're on PG 14.3. Verified the query plans.

harimohanraj89: I was about to ask. Nice.

app/models/region.rb
```diff
@@ -156,9 +156,7 @@ def syncable_patients
     case region_type
       when "block"
         registered_patients.with_discarded
-          .select(:id, :registration_facility_id)
-          .or(assigned_patients.with_discarded.select(:id, :registration_facility_id))
-          .union(appointed_patients.with_discarded.select(:id, :registration_facility_id))
+          .union(assigned_patients.with_discarded)
```
harimohanraj89, line 159: There a few notable changes here.
* We were selecting only a handful of columns before, and we no longer are
* `.or(assigned_patients...` is now `.union(assigned_patients...`

Can you talk me through why these changes are made, or aren't important?

harimohanraj89 (changes requested):
Just one confusion area for me.

app/models/region.rb
prabhanshuguptagit: We were doing a `union` on a restricted set of columns earlier because it's expensive to union on all columns. We were then doing a subquery like so to get back all the columns
```
Patient.where(id: current_sync_region.syncable_patients)
```
It turns out it's faster to do a `union` on all columns (new) than doing the `union` + subquery lookup (old). The query is now doing 
```
current_sync_region.syncable_patients.updated_on_server_since ...
```

It's also faster than doing an `or`. (plan attached)

Subquery with union: https://explain.depesz.com/s/ebl0 (slow)
Or: https://explain.depesz.com/s/VLwB (slower)
Union: https://explain.depesz.com/s/BUYM (fastest)

harimohanraj89: Thank you for this explanation, very helpful! Let's do this 😎

harimohanraj89 (approved):

PR #4226, opened 2022-08-26
Create PG extensions in separate schema for review apps

prabhanshuguptagit:
**Story card:** [sc-8965](https://app.shortcut.com/simpledotorg/story/8965/fix-breaking-heroku-deployments-because-of-change-in-pg-extensions-policy)

## Because

Heroku has mandated creating PG extensions in a `heroku_ext` schema and doesn't let us create them in the `public` schema anymore. https://devcenter.heroku.com/changelog-items/2446

## This addresses

Adds a hook to `db:structure:load` that modifies the file on review apps accordingly.

## Test instructions

Spin up an app with this PR.

### DIFF
--- a/lib/tasks/support_pg_extensions_in_heroku.rake
+++ b/lib/tasks/support_pg_extensions_in_heroku.rake
@@ -1,0 +1,16 @@
+task :support_pg_extensions_in_heroku do
+  # https://devcenter.heroku.com/changelog-items/2446
+  next unless (ENV["SIMPLE_SERVER_ENV"] == "android_review" || ENV["SIMPLE_SERVER_ENV"] == "review")
+
+  path = "db/structure.sql"
+  f = File.open(path)
+  contents = f.read.to_s
+  f.close
+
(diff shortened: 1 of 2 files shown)


lib/tasks/support_pg_extensions_in_heroku.rake
```diff
@@ -0,0 +1,19 @@
+task :support_pg_extensions_in_heroku do
+  # https://devcenter.heroku.com/changelog-items/2446
+  next unless ENV["SIMPLE_SERVER_ENV"] == "android_review" || ENV["SIMPLE_SERVER_ENV"] == "review"
```
harimohanraj89, line 3: `ENV["SIMPLE_SERVER_ENV"].in?(%w[android_review review])`

harimohanraj89 (approved):
I mean, it works! This should be just fine for our review app setup. How is this not a bigger problem yet? Is this something we can expect a longer-term solution to from Heroku or Rails?

prabhanshuguptagit:
The problem _is_ big enough actually that there's a lot of people ranting and sharing fixes 😛 https://blog.testdouble.com/posts/2022-08-15-migrating-postgres-extensions-to-the-heroku_ext_schema/

People are expecting heroku to follow up but it's been a while.

PR #4227, opened 2022-08-26
Add  `patient initiated on treatment` card at the end of monthly reports

PriyangaPKini:
**Story card:** [sc-8978](https://app.shortcut.com/simpledotorg/story/8978)

## Because

The yellow card saying "Where are the patients initiated on treatment data?" was missing on the monthly reports page

## This addresses

Adds a yellow card saying "Where are the patients initiated on treatment data?" on the monthly reports page

## Test instructions

- Checkout to branch
- Enable `new_progress_tab_v1` feature flag
- Sidebar>Reports>Choose any facility>Progress Tab> Monthly 
- Scroll down to the end. You should be able to see the card


app/views/api/v3/analytics/user_analytics/_progress_monthly_report.html.erb
```diff
@@ -46,4 +46,18 @@
       <%= render(Reports::MonthlyProgressComponentV1.new(dimension, service: @service, current_user: @current_user)) %>
     <% end %>
   </div>
+
+  <% if CountryConfig.current_country?("India") %>
```
govindkrjoshi, line 50: Instead of country config directly here, you can make it a helper in the controller.

govindkrjoshi (approved):

config/locales/en.yml
```diff
@@ -180,8 +180,8 @@ en:
         title: "Follow-up patients"
         subtitle: "Patients that had a BP taken, a blood sugar taken, an appointment scheduled, or their medications updated at %{facility_name}."
       patients_initiated_on_treatment_card:
-        title: 'Where is the "Patients initiated on treatment" data?'
-        main_text: 'Use "Registered patients", it will always equal "Patients initiated on treatment" for patients registered on Simple.'
+        title: 'Where are the "Patients initiated on treatment" data?'
```
harimohanraj89, line 183: This should be "is", I think - since we're asking "Where is the data?"

config/locales/en.yml
```diff
@@ -180,8 +180,8 @@ en:
...
+        main_text: 'Use "Registered patients", it will always be equal to the "Patients initiated on treatment" for patients registered on Simple.'
```
harimohanraj89, line 184: The original string was grammatically ok, and more concise. Reckon we can revert?

harimohanraj89 (changes requested):
Couple points about the grammatical updates

config/locales/en.yml
PriyangaPKini: My bad :(

config/locales/en.yml
PriyangaPKini: yeah I am reverting that commit

govindkrjoshi (approved):

PR #4233, opened 2022-09-01
Add missing diabetes related keys to warm cache

PriyangaPKini:
**Story card:** [sc-9053](https://app.shortcut.com/simpledotorg/story/9053/discrepancy-between-dashboard-and-app-counts-for-registered-patients)

## Because

Inconsistency between registration numbers in the dashboard

## This addresses

Adds missing diabetes related keys to `warm_cache`

## Test instructions

- Registration numbers should be consistent across dashboard


harimohanraj89 (changes requested):
Do tests need to be updated? If not, do tests need to be written for this? 😅

PriyangaPKini:
> Do tests need to be updated? If not, do tests need to be written for this? 😅

There are no tests associated with this function. We will have to write new tests for this.

PriyangaPKini:
It looks like there is some test associated with the warm_cahe but I am not sure how to use it for non-delegate functions
https://github.com/simpledotorg/simple-server/blob/f46f76ccef252d64a66c685adec34f25fe39d467/spec/services/reports/region_cache_warmer_spec.rb#L74

spec/services/reports/region_cache_warmer_spec.rb
```diff
@@ -86,5 +86,38 @@
...
+        /blood_sugar_measures_by_user/,
+        /monthly_registrations_by_user\/group_by\/registration_user_id\/period_type\/month\/diagnosis\/hypertension/,
+        /monthly_registrations_by_user\/group_by\/registration_user_id\/period_type\/month\/diagnosis\/diabetes/,
+        /monthly_registrations_by_gender/,
+        /controlled_by_gender/,
+        /overdue_calls_by_user/
+      ]
+
+      expected_keys_in_cache.each do |expected_key|
+        expect(cache_keys.any? { |key| key.match(expected_key) }).to be_truthy
```
harimohanraj89, line 119: Can we simplify this with `cache_keys.include?(expected_key)`?

harimohanraj89 (approved):
Looks great! Nice test addition 👌 One last comment.

spec/services/reports/region_cache_warmer_spec.rb
prabhanshuguptagit: Can't do an `include?` because this is matching on regexes.

spec/services/reports/region_cache_warmer_spec.rb
```diff
@@ -86,5 +86,38 @@
       repo.controlled_rates
       repo.controlled_rates(with_ltfu: true)
     end
+
+    it "caches all the result of repository methods" do
```
prabhanshuguptagit, line 90: Nice test indeed👌🏼

PR #4237, opened 2022-09-02
Use device_created_at to pick latest overdue appointment

prabhanshuguptagit:
**Story card:** [sc-9056](https://app.shortcut.com/simpledotorg/story/9056/use-the-latest-appointment-by-device-created-at-for-calculating-patient-overdueness)

## Because

We use the latest scheduled_date to find a patient's latest overdue appointment. If a user creates many appointments for a patient, we should use the latest created appointment and not the one that's farthest into the future (latest scheduled appointment) to evaluate overdueness.

## This addresses

Updates the "overdue" calculation for a patient in overdue list and line list SQL to:
- Find the latest appointment for a patient by `device_created_at` (instead of `scheduled_date`)
- Call a patient overdue if the status of _this_ appointment is `scheduled` or `cancelled` and the `scheduled_date` has passed

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -1145,7 +1145,7 @@ CREATE MATERIALIZED VIEW public.patient_registrations_per_day_per_facilities AS
 
 CREATE VIEW public.patient_summaries AS
  SELECT p.recorded_at,
-    concat(date_part('year'::text, p.recorded_at), ' Q', date_part('quarter'::text, p.recorded_at)) AS registration_quarter,
+    concat(date_part('year'::text, p.recorded_at), ' Q', EXTRACT(quarter FROM p.recorded_at)) AS registration_quarter,
     p.full_name,
         CASE
             WHEN (p.date_of_birth IS NOT NULL) THEN date_part('year'::text, age((p.date_of_birth)::timestamp with time zone))
@@ -1166,7 +1166,7 @@ CREATE VIEW public.patient_summaries AS
     latest_blood_pressure.systolic AS latest_blood_pressure_systolic,
     latest_blood_pressure.diastolic AS latest_blood_pressure_diastolic,
     latest_blood_pressure.recorded_at AS latest_blood_pressure_recorded_at,
-    concat(date_part('year'::text, latest_blood_pressure.recorded_at), ' Q', date_part('quarter'::text, latest_blood_pressure.recorded_at)) AS latest_blood_pressure_quarter,
+    concat(date_part('year'::text, latest_blood_pressure.recorded_at), ' Q', EXTRACT(quarter FROM latest_blood_pressure.recorded_at)) AS latest_blood_pressure_quarter,
     latest_blood_pressure_facility.name AS latest_blood_pressure_facility_name,
     latest_blood_pressure_facility.facility_type AS latest_blood_pressure_facility_type,
     latest_blood_pressure_facility.district AS latest_blood_pressure_district,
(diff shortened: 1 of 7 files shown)


db/structure.sql
```diff
@@ -916,7 +916,8 @@ CREATE MATERIALIZED VIEW public.materialized_patient_summaries AS
             medical_histories.diagnosed_with_hypertension,
             medical_histories.deleted_at,
             medical_histories.user_id,
-            medical_histories.hypertension
+            medical_histories.hypertension,
+            medical_histories.receiving_treatment_for_diabetes
```
prabhanshuguptagit, line 920: Outdated schema that wasn't committed.

db/views/materialized_patient_summaries_v02.sql
```diff
@@ -0,0 +1,138 @@
...
+      WHERE identifier_type = 'simple_bp_passport'
+        AND deleted_at IS NULL
+      ORDER BY patient_id,
+               device_created_at DESC) AS latest_bp_passport ON latest_bp_passport.patient_id = p.id
+         LEFT OUTER JOIN
+     (SELECT DISTINCT ON (patient_id) *
+      FROM appointments
+      WHERE deleted_at IS NULL
+      ORDER BY patient_id, device_created_at DESC) AS next_scheduled_appointment ON next_scheduled_appointment.patient_id = p.id
+                                                                     AND next_scheduled_appointment.status = 'scheduled'
```
prabhanshuguptagit, line 134: This is the only change in this file (Change 2).

db/views/patient_summaries_v09.sql
```diff
@@ -0,0 +1,125 @@
...
+    AND bp_passport.patient_id = p.id
+    ORDER by bp_passport.device_created_at desc
+    LIMIT 1
+) latest_bp_passport ON TRUE
+
+LEFT JOIN LATERAL (
+    SELECT *
+    FROM appointments a
+    WHERE a.patient_id = p.id
+    ORDER by a.device_created_at DESC
```
prabhanshuguptagit, line 120: This is the only change in this file (Change 1).

db/views/materialized_patient_summaries_v02.sql
harimohanraj89: We should NOT filter by scheduled appointments here, right? If the most-recent-device-created appointment has been marked "visited", then the patient is not overdue. We wouldn't be looking at that appointment if we keep this clause in here, right?

harimohanraj89 (changes requested):
One big question from me.

db/views/materialized_patient_summaries_v02.sql
prabhanshuguptagit: This clause applies after the `DISTINCT ON` has been applied. If there is no `scheduled` appointment (only a `visited` appointment) the `next_scheduled_appointment` would be NULL and the line list will not show this patients as overdue.

prabhanshuguptagit: Added a spec [acf7d4c](https://github.com/simpledotorg/simple-server/pull/4237/commits/acf7d4cd4a53b56eb3e4e1b8f8e6fa85d8ebc6d9) for this

harimohanraj89: Thanks for the spec. What about the situation we inspected in real prod data last week? What if a patient today has a "scheduled"  appointment for Sept 7th, and a more recently created "visited" appointment for Sept 1st?

prabhanshuguptagit: This patient wouldn't be counted as overdue. 
- `DISTINCT ON` finds the `visited` appointment as the latest appointment
- The appointment doesn't get included in the left join because it's not `scheduled`
- Patient is not counted as overdue

There's a [unit test](https://github.com/simpledotorg/simple-server/pull/4237/files#diff-b7f8fb63d18f5bca6e2e35e61d918bd6fed3ff9f5fd4e9a2b5d8b9b60ab8318eR171) for this scenario.

harimohanraj89: Perfect 👌

harimohanraj89 (approved):

PR #4238, opened 2022-09-02
Add sync to user endpoint for call results 

prabhanshuguptagit:
**Story card:** [sc-9057](https://app.shortcut.com/simpledotorg/story/9057/add-sync-to-user-endpoint-for-call-results)

## Because

We should sync call results to devices since the client has started relying on these records to categorize patients in the overdue tab (under the collapsibles).

## This addresses

Adds a GET `/api/v4/call_results/sync` endpoint to sync call results to devices

- [x] Adds an optional `patient_id` and `facility_id` to call results
- [x] Backfills the two columns for existing call results using these two columns
- [x] Adds to fill patient_id for new incoming call results

TODO:
- [x] Test how long the data migration takes on prod sized data


app/controllers/api/v4/call_results_controller.rb
```diff
@@ -1,28 +1,44 @@
...
+  end
+
+  def sync_to_user
+    __sync_to_user__("call_results")
   end
 
   private
 
+  def metadata(call_result_params)
+    {facility_id: call_result_params["facility_id"] || current_facility.id}
```
prabhanshuguptagit, line 13: Use the `facility_id` supplied in the body or default to the facility ID header.

app/transformers/api/v4/call_result_transformer.rb
```diff
@@ -0,0 +1,9 @@
+class Api::V4::CallResultTransformer < Api::V4::Transformer
+  def from_request(call_result_payload)
+    attributes = super(call_result_payload)
+    return attributes if attributes["patient_id"]
+
+    appointment = Appointment.find_by(id: call_result_payload["appointment_id"])
```
prabhanshuguptagit, line 6: To support older app versions that don't supply the `patient_id`

PR #4270, opened 2022-09-12
Fix diabetes treatment outcome description

PriyangaPKini:
**Story card:** [sc-9093](https://app.shortcut.com/simpledotorg/story/9093/error-with-dm-descriptions-on-the-dashboard-in-treatment-status-of-patients-under-care)

## Because

The description of all the treatment outcomes was the same.

## This addresses

Add a discrete description to each of the treatment outcomes.

## Other Changes
 
Fix padding of the diabetes treatment outcome card

## Test instructions

- Checkout to this branch
- Reports > <Any Region> > Diabetes > Treatment Outcome Card
- You should see discrete definitions for each outcome

Before
Screenshot 2022-09-12 at 2 50 16 PM

After
Screenshot 2022-09-12 at 2 48 01 PM

### DIFF
--- a/app/components/dashboard/diabetes/treatment_outcomes_card_component.html.erb
+++ b/app/components/dashboard/diabetes/treatment_outcomes_card_component.html.erb
@@ -25,7 +25,8 @@
 
               <p class="m-0px c-black">
                 <span data-key="<%= treatment_outcome[:count] %>" data-format="numberWithCommas"></span>
-                patients with no visit from
+                <%= treatment_outcome[:description] %>
+                from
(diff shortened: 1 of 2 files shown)


jamiecarter7:
I think you are able to remove the pt-20px class as its is covered by the p-20px.
Otherwise this looks good to push

jamiecarter7 (approved):

PriyangaPKini:
> I think you are able to remove the pt-20px class as its is covered by the p-20px. Otherwise this looks good to push

oh okay. I will remove pt-20px then

PR #4272, opened 2022-09-13
Add local SMS channel for bangladesh

prabhanshuguptagit:
**Story card:** [sc-8985](https://app.shortcut.com/simpledotorg/story/8985/integrate-the-apis-of-new-bd-sms-provider)

## Because

We want to use https://sms.net.bd for sending SMSes in bangladesh.

API docs: https://sms.net.bd/api

## This addresses

Adds a messaging channel to send SMSes through this provider. This does not fetch statuses for SMS deliveries yet, this will be added in a nightly rake task in a follow-up PR.


app/services/messaging/alpha_sms/sms.rb
```diff
@@ -0,0 +1,44 @@
...
+        send_sms(recipient_number, message)["data"]["request_id"],
+        &with_communication_do
+      )
+    end
+  end
+
+  private
+
+  def send_sms(recipient_number, message)
+    Messaging::AlphaSms::Api.new.send_sms(
```
harimohanraj89, line 20: We can make class methods that delegate to these instance methods. That's more consistent with our other SMS channel interfaces I think.

config/initializers/country_config.rb
```diff
@@ -45,7 +45,7 @@ class CountryConfig
         monthly_registrations: "tI9g0mtCzOv"
       },
       enabled_diabetes_population_coverage: true,
-      appointment_reminders_channel: "Messaging::Twilio::ReminderSms"
+      appointment_reminders_channel: "Messaging::AlphaSms::Sms"
```
harimohanraj89, line 48: Why no distinction between OTP and Reminder SMS in this channel? Is it because there aren't any "multiple senders" to leverage like we had in Twilio?

spec/services/messaging/alpha_sms/api_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      message = "Test message"
+      recipient_number = "+8801885374409"
+
+      request = stub_request(:post, "https://api.sms.net.bd/sendsms")
+      described_class.new.send_sms(recipient_number: recipient_number, message: message)
+      expect(request.with(body: {
+        api_key: "ABCDEF",
+        msg: message,
+        to: recipient_number
+      })).to have_been_made
```
harimohanraj89, line 27: The API seems simple enough, which is nice.

app/services/messaging/alpha_sms/api.rb
```diff
@@ -0,0 +1,47 @@
+class Messaging::AlphaSms::Api
+  HOST = "api.sms.net.bd"
+  URL_PATHS = {
+    send_sms: "/sendsms",
+    request_report: "/report/request",
+    user_balance: "/user/balance"
```
harimohanraj89, line 6: These two aren't used yet. Wanna add them to this hash now or later?

harimohanraj89 (changes requested):
Minor things

app/services/messaging/alpha_sms/sms.rb
prabhanshuguptagit: I don't find the class methods very helpful tbh. In this case for example, the `.new` explicitly tells that an instance of the API client has been instantiated. I was a bit iffy about the channel's interface too so thinking I'd leave this be😅

harimohanraj89: Coolbeans, I don't feel strongly about this, just thought of saving a few keystrokes, nothing major.

harimohanraj89 (approved):
Looks great. I like how easy it was to add a new channel, and how our abstractions fit nicely into this piece of the puzzle.

PR #4274, opened 2022-09-14
Add new diagnosis to daily followup

PriyangaPKini:
**Story card:** [sc-7364](https://app.shortcut.com/simpledotorg/story/7364/add-new-diagnoses-to-daily-follow-up-rb)

## Because

Scopes like `with_hypertension_only`,`with_diabetes_only and `with_hypertension_and_diabetes' were missing

## This addresses

Add new scopes to DailyFollowUp model - `with_hypertension_only`,`with_diabetes_only` and `with_hypertension_and_diabetes`


spec/models/reports/daily_follow_up_spec.rb
```diff
@@ -51,6 +51,9 @@
     expect(follow_ups.count).to eq(3)
     expect(follow_ups.with_hypertension.count).to eq(2)
     expect(follow_ups.with_diabetes.count).to eq(2)
+    expect(follow_ups.with_diabetes_only.count).to eq(1)
```
prabhanshuguptagit, line 54: Could match the blood pressure IDs here instead of matching counts to check if the right follow up is included in the right scope.

prabhanshuguptagit (approved):
:shipit:

spec/models/reports/daily_follow_up_spec.rb
PriyangaPKini: Right. Will make the change before merging the PR

prabhanshuguptagit:
Can get rid of the `count` assertions now since they're redundant.

PR #4275, opened 2022-09-14
Hide drug stock subtitle when patient_days is error

PriyangaPKini:
**Story card:** [sc-9155](https://app.shortcut.com/simpledotorg/story/9155/drug-stock-reports-show-error-days-of-drug-stock-need-a-fix)

## Because

The subtitle under drug stock was confusing and not user-friendly. 

## This addresses

Hide subtitle when ever a drug is not in formula or other configuration error. 

## Test instructions

1. Check out this branch
2. Make sure `new_progress_tab_v2` is enabled
3. goto `http://localhost:3000/api/v3/analytics/user_analytics.html`. Use ModHeader or Postman to add necessary headers
4. See more details [here](https://github.com/simpledotorg/simple-server#testing-web-views)
5. You should be able to see the progress tab reports and verify if the subtitle is shown only when patient_days is not error

## Screenshots

**Before:**
Screenshot 2022-09-14 at 7 56 36 PM

**After:**
Screenshot 2022-09-14 at 7 55 54 PM

### DIFF
--- a/app/components/progress_tab/drug_stock_component.html.erb
+++ b/app/components/progress_tab/drug_stock_component.html.erb
@@ -10,7 +10,9 @@
       <%= protocol_drug_labels[category][:full] %>
     </h3>
     <p class="m-0px mb-16px p-0px ta-left fw-regular fs-16px lh-150 <%= patient_days_css_class(patient_days, prefix: "c") %>">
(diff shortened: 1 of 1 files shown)


dburka (approved):
Looks great

PR #4277, opened 2022-09-15
Changed text, added color and bottom margin.

dburka:
**Story card:** No card

## Because

The Ethiopia team was very confused by the copy of the Overdue Calls card on Reports. The current copy suggests that we're counting unique patients contacted:

```
Overdue patients contacted

All overdue patients with hypertension or diabetes called in a month by healthcare workers.
```

I changed the text to make it clear that we're actually counting the # of calls made (i.e. number of "Result of call..." actions)

```
Calling overdue patients

The number of calls made to overdue diabetes and hypertension patients in a month.
``` 

## This addresses

User confusion and imprecision in the text.

image

### DIFF
--- a/app/components/registrations_and_follow_ups_component.html.erb
+++ b/app/components/registrations_and_follow_ups_component.html.erb
@@ -114,12 +114,14 @@
 
 <% if current_admin.feature_enabled?(:show_call_results) %>
   <div class="card w-100pt mt-0 pr-0 pr-md-3 pb-inside-avoid">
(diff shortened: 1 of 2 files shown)


jamiecarter7 (approved):
Changes look good on all screen widths

I wonder if we should use the title format from the block below (quarterly report) for consistency?

Screenshot 2022-09-16 at 09 41 33

Screenshot of the template in quarterly report, the div of the wrapper is highlighted/selected.

PR #4281, opened 2022-09-19
Create prescription drugs while seeding patients

govindkrjoshi:
**Story card:** [sc-8986](https://app.shortcut.com/simpledotorg/story/8986/validate-cphc-migration-script-in-staging)

## Because

We are currently not creating prescription drugs when seeding patients

## This addresses

Adds a prescription drugs seeder


lib/seed/runner.rb
```diff
@@ -42,6 +42,9 @@ def call
       UserSeeder.call(config: config)
       seed_drug_stocks
 
+      ProtocolSeeder.call(config: config)
```
prabhanshuguptagit, line 45: This is already called via `FacilitySeeder`, is it needed?

lib/seed/prescription_drug_seeder.rb
```diff
@@ -0,0 +1,80 @@
...
+        return {}
+      end
+
+      prescription_drugs = []
+      protocol_drugs = if facility.protocol.present?
+        facility.protocol.protocol_drugs
+      else
+        []
+      end
+      encounters.each_with_index do |encounter, index|
```
prabhanshuguptagit, line 45: We don't associate drugs to encounters anywhere currently. Shall we create drugs without encounters as the default, and let the user enable encounters for drugs through config if they want?

lib/seed/prescription_drug_seeder.rb
```diff
@@ -0,0 +1,80 @@
...
+      @raw_to_clean_medicines = DrugLookup::RawToCleanMedicine.all
+      @logger.info "Starting #{self.class} with #{config.type} configuration"
+    end
+
+    def encounters
+      @facility.encounters
+    end
+
+    def prescription_drugs_to_create
+      (0...5).to_a.sample
```
prabhanshuguptagit, line 30: Let's pick this number from config akin to `MAX_BPS_TO_CREATE`.

prabhanshuguptagit (approved):
Looks great.

lib/seed/prescription_drug_seeder.rb
govindkrjoshi: We are not creating encounters for prescription drugs, but creating prescription drugs on the day of the encounter.

PR #4283, opened 2022-09-20
Updated copy of "Calls made to overdue patients" tooltip

dburka:
Tiny copy fix to make tooltip consistent with the UI copy change from last week.

**Story card:** No card

## Because

We made the text more clear in the "Calls made to overdue patients" card that the count is how many 'result of calls' were marked, not how many unique patients were called. This had caused confusion for Ethiopia team and others.

## This addresses

Text consistency.

## Test instructions

Shouldn't break the Facility Report page.


jamiecarter7 (approved):

dburka:
> The updated title is much clearer wording.
> 
> I wonder if we should update the heading for the same card at the district/region level. The wording is much clearer.
> 
> Screenshot 2022-09-20 at 11 05 23

Good catch. Changing this text too.

PR #4290, opened 2022-09-22
Dashboard: Add missing table key value (>90 patient days of stock)

jamiecarter7:
**Story card:** none

## Because

There is a missing key value '>90 patient days of stock'

Screenshot 2022-09-22 at 14 23 03

## This addresses

Green colour showing on the table data but no key to explain why.

Screenshot 2022-09-22 at 14 20 02

## Test instructions

Screenshot 2022-09-22 at 14 22 29

Found at: http://localhost:3000/my_facilities/drug_stocks?facility_group=aligarh&for_end_of_month=Aug-2022&size=small

Scroll to page bottom.

To check colour match insert data into CCB Tablets (this works) using the '+' button to add drug data.


dburka (approved):
Looks good. Good catch.

PR #4295, opened 2022-09-23
Dashboard: Quarterly report 'visit but no bp' and 'missed visits' order reversed

jamiecarter7:
**Story card:** None

## Because

The order of the quarterly report is blocks is shown differently between the dashboard (reports) and the app (progress-tab).

## This addresses

Dashboard shows 'visit but no BP recorded' last while the 'progress-tab' shows 'missed visits' last.
Screenshot 2022-09-23 at 12 53 22

It makes more sense to follow the app layout where "patient has visited" blocks sit next to one another and the missed visits sit at the end.

## Test instructions

Go to any facility or region report page: http://localhost:3000/reports/regions - then select any region or facility

Scroll the bottom of the page its the last block under 'cohort reports'

Screenshot 2022-09-23 at 13 01 11

### DIFF
--- a/app/components/reports/cohort_component.html.erb
+++ b/app/components/reports/cohort_component.html.erb
@@ -65,16 +65,6 @@
                   <%= (cohort["uncontrolled"] > 1 && uncontrolled_percent == 0) ? "< 1" : uncontrolled_percent %>%
                 </td>
                 <td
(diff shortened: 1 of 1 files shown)


dburka (approved):
Looks good. Nice work.

PR #4310, opened 2022-10-07
Use prosopite in development to detect and avoid n+1 queries

govindkrjoshi:
## Because

It's easy to add n+1 queries which can cost at lot in terms of view load times. 
 
## This addresses

Prosopite has methods to detect and report n+1 queries. This change will log instances of n+1 queries in the request path, which we can try to fix.


harimohanraj89:
Is there anything more intrusive that we can introduce into the dev environment? I see myself missing the logs.

prabhanshuguptagit:
We can raise these as an exception or send them to a separate log file. The exceptions can be quite intrusive though since some pages have intentional N+1s. I found rails logs to be a decent middle ground. The logline looks big and red, bit hard to miss :P
image

PR #4284, opened 2022-09-20
Add reporting_daily_follow_ups_and_registrations

PriyangaPKini:
**Story card:** [sc-7017](https://app.shortcut.com/simpledotorg/story/7017/change-progress-counts-of-registrations-and-follow-ups-to-be-htn-only-dm-only-and-co-morbid-patients-i-e-htn-and-dm)

## Because

We need to get the live registration and follow-up data for the progress tab. Currently, we have `reporting_daily_follow_ups` which only has the daily follow-ups. 

## This addresses

- Adds live registrations data 
- Rename the `reporting_daily_follow_ups` to `reporting_facility_state_daily_dimensions`. 
  - One row for a facility per day

## Test instructions
 - I have added the query as a [question](https://metabase.simple.org/question/528-facility-wise-daily-breakdown-by-gender-and-diagnosis) in mat-view for testing with the live data (copied the [question](https://metabase.bd.simple.org/question/151-facility-wise-daily-breakdown-by-gender-and-diagnosis) in Bangladesh Metabase as well)
 - It takes about 38-40s to get the result in Metabase with India production data
 - I am thinking of refreshing this view every 30 minutes

PR #4315, opened 2022-10-10
Progress tab: Daily follow ups and registration fixes

prabhanshuguptagit:
**Story card:** [sc-7017](https://app.shortcut.com/simpledotorg/story/7017/change-progress-counts-of-registrations-and-follow-ups-to-be-htn-only-dm-only-and-co-morbid-patients-i-e-htn-and-dm)

## Because

We found some bugs in the new matview added for daily follow ups and registrations in  https://github.com/simpledotorg/simple-server/pull/4284

## This addresses

- Fixes the registration facility ID calculation
- Fixes date grouping calculation for visit_date
- Fixes the `last_updated_at` (which was set to a fake data)


harimohanraj89 (approved):
One question, but looks good 👍

PR #4316, opened 2022-10-10
Reports: Fix treatment outcomes card

prabhanshuguptagit:
**Story card:** [sc-9091](https://app.shortcut.com/simpledotorg/story/9091/treatment-outcomes-card-is-blank-for-facilities-with-dm-disabled)

## Because
Treatment outcomes card is blank for regions with DM disabled. This happens due to an unrelated JS bug.
image

## This addresses
In DM-enabled regions we show a breakdown of registrations by HTN and DM. On DM-disabled facilities we don't show this data element and the JS breaks because it can't find the `data-` attribute on the page.

image

Fixes `reports.js` to handle this case.


harimohanraj89 (approved):
Great find and great fix.

PR #4330, opened 2022-10-25
Cache warmer performance improvements #2

prabhanshuguptagit:
**Story card:** [sc-9236](https://app.shortcut.com/simpledotorg/story/9236/parallelize-cache-warming)

## Because

The cache warmer cron job takes a long time to complete which runs into peak traffic hours. This leads to requests going to the cron server returning [5xxes](https://simpledotorg.slack.com/archives/C012BDG130D/p1664603549089529).

## This addresses

Adds some improvements suggested in https://github.com/simpledotorg/simple-server/pull/4312

- Adds missing indexes for reporting queries
- Remove more metrics that don't need caching
- Parallelises cache refresh using sidekiq. We will run throttle this to run 2 jobs at a time and increase the concurrency if the box isn't affected.

### DIFF
--- a/app/jobs/reports/region_cache_warmer_job.rb
+++ b/app/jobs/reports/region_cache_warmer_job.rb
@@ -4,7 +4,7 @@ class Reports::RegionCacheWarmerJob
 
   sidekiq_options queue: :default
   sidekiq_throttle(
-    concurrency: { limit: 2 }
+    concurrency: {limit: 1}
   )
 
   def perform(region_type, limit, offset)
--- a/app/jobs/reports/region_cache_warmer_job.rb
+++ b/app/jobs/reports/region_cache_warmer_job.rb
@@ -1,0 +1,33 @@
+class Reports::RegionCacheWarmerJob
+  include Sidekiq::Worker
+  include Sidekiq::Throttled::Worker
+
+  sidekiq_options queue: :default
+  sidekiq_throttle(
+    concurrency: { limit: 2 }
+  )
+
+  def perform(region_type, limit, offset)
+    if Flipper.enabled?(:disable_region_cache_warmer)
+      Rails.logger.info "Cache warmer disabled via flipper - exiting"
+      return
+    end
+
+    Rails.logger.info "Warming repository cache for #{region_type}, limit: #{limit}, offset: #{offset}"
+    warm_cache(region_type, limit, offset)
+    Rails.logger.info "Finished warming repository cache for #{region_type}, limit: #{limit}, offset: #{offset}"
+  end
+
+  def warm_cache(region_type, limit, offset)
+    RequestStore.store[:bust_cache] = true
+    Time.use_zone(Period::REPORTING_TIME_ZONE) do
+      period = Reports.default_period
+      regions = Region.where(region_type: region_type).limit(limit).offset(offset)
+      range = (period.advance(months: -23)..period)
+
+      Reports::Repository.new(regions, periods: range).warm_cache
+    end
+  ensure
+    RequestStore.store[:bust_cache] = false
+  end
+end
--- a/app/jobs/reports/region_cache_warmer_job.rb
+++ b/app/jobs/reports/region_cache_warmer_job.rb
@@ -4,7 +4,7 @@ class Reports::RegionCacheWarmerJob
 
   sidekiq_options queue: :default
   sidekiq_throttle(
-    concurrency: {limit: 1}
+    concurrency: {limit: 2}
   )
 
   def perform(region_type, limit, offset)
--- a/app/models/reports/repository.rb
+++ b/app/models/reports/repository.rb
@@ -142,7 +142,6 @@ module Reports
         blood_sugar_measures_by_user
         monthly_registrations_by_user(diagnosis: :hypertension)
         monthly_registrations_by_user(diagnosis: :diabetes)
-        monthly_registrations_by_gender
         overdue_calls_by_user
       end
     end
--- a/app/services/refresh_reporting_views.rb
+++ b/app/services/refresh_reporting_views.rb
@@ -107,6 +107,7 @@ class RefreshReportingViews
         klass = name.constantize
         klass.refresh
       end
+      Statsd.instance.flush
     end
   end
 
@@ -114,7 +115,9 @@ class RefreshReportingViews
     name = "refresh_reporting_views.#{operation}"
     benchmark(name) do
       Datadog::Tracing.trace("refresh_matview", resource: operation) do |span|
-        yield
+        Statsd.instance.time(name) do
+          yield
+        end
       end
     end
   end
--- a/app/services/reports/region_cache_warmer.rb
+++ b/app/services/reports/region_cache_warmer.rb
@@ -14,6 +14,11 @@ module Reports
     end
 
     def call
+      if Flipper.enabled?(:disable_region_cache_warmer)
+        Rails.logger.info "Cache warmer disabled via flipper - exiting"
+        return
+      end
+
       Rails.logger.info "Starting queuing cache warming jobs"
 
       Region::REGION_TYPES.excluding("root").each do |region_type|
--- a/app/services/reports/region_cache_warmer.rb
+++ b/app/services/reports/region_cache_warmer.rb
@@ -1,69 +1,36 @@
 module Reports
   class RegionCacheWarmer
     prepend SentryHandler
+    BATCH_SIZE = 250
 
-    def self.call
-      new.call
+    def self.call(*args)
+      new(*args).call
     end
 
-    attr_reader :name
-    attr_reader :period
+    attr_reader :batch_size
 
-    def initialize(period: Reports.default_period)
-      @period = period
-      @name = self.class.name.to_s
-      notify "Starting #{name} warming"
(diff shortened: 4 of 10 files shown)


app/models/reports/repository.rb
```diff
@@ -142,7 +142,6 @@ def warm_cache
         blood_sugar_measures_by_user
         monthly_registrations_by_user(diagnosis: :hypertension)
         monthly_registrations_by_user(diagnosis: :diabetes)
-        monthly_registrations_by_gender
```
prabhanshuguptagit, line 145: This metric is also used by DHIS2 reports only, no need to cache.

app/services/refresh_reporting_views.rb
```diff
@@ -107,14 +107,17 @@ def refresh
...
+      Statsd.instance.flush
     end
   end
 
   def benchmark_and_statsd(operation)
     name = "refresh_reporting_views.#{operation}"
     benchmark(name) do
       Datadog::Tracing.trace("refresh_matview", resource: operation) do |span|
-        yield
+        Statsd.instance.time(name) do
```
prabhanshuguptagit, line 118: We used to have statsd timing earlier which was removed in favour of tracing. However, the trace's time - [sample trace](https://app.datadoghq.com/apm/traces?query=%40_top_level%3A1%20env%3Aindia-production%20service%3Asimple_server%20operation_name%3Arefresh_matview&cols=service%2Cresource_name%2C%40duration%2C%40http.method%2C%40http.status_code%2C%40_duration.by_service&graphType=flamegraph&historicalData=true&saved-view-id=1277623&spanID=2209728970880325150&spanType=service-entry&spanViewType=metrics&timeHint=1666649615069&trace=AgAAAYQMDurdGcSDDQAAAAAAAAAYAAAAAEFZUU1Edm9BQUFDNGFicUVQVWlvM0ZRcgAAACQAAAAAMDE4NDBjNmEtYWJiNS00NjIxLWFlMzYtNDU3MmRkMzI1ZTQz&traceID=3122197871034418358&viz=stream&start=1666521747668&end=1666694547668&paused=false) can't be plotted on a graph over time.

config/schedule.rb
```diff
@@ -96,7 +96,7 @@ def local(time)
   runner "RecordCounterJob.perform_async"
 end
 
-every :day, at: local("02:30 am"), roles: [:cron] do
+every :day, at: local("04:30 am"), roles: [:cron] do
```
prabhanshuguptagit, line 99: The matviews refresh in ~3 hours. We start at 12.30, which doesn't leave any margin.

db/migrate/20221024071410_add_reporting_facility_state_region_indexes.rb
```diff
@@ -0,0 +1,8 @@
+class AddReportingFacilityStateRegionIndexes < ActiveRecord::Migration[6.1]
+  def change
+    add_index :reporting_facility_states, [:organization_region_id, :month_date], name: :index_fs_organization_month_date
+    add_index :reporting_facility_states, [:state_region_id, :month_date], name: :index_fs_state_month_date
+    add_index :reporting_facility_states, [:district_region_id, :month_date], name: :index_fs_district_month_date
+    add_index :reporting_facility_states, [:block_region_id, :month_date], name: :index_fs_block_month_date
```
prabhanshuguptagit, line 6: These lookups happen while refreshing org/state/district/block reports. Adding these on a BD prod-sized dataset reduced time by half, however the matview refresh time is expected to go up slightly.

db/migrate/20221024071710_add_reporting_overdue_calls_indexes.rb
```diff
@@ -0,0 +1,6 @@
+class AddReportingOverdueCallsIndexes < ActiveRecord::Migration[6.1]
+  def change
+    add_index :reporting_overdue_calls, [:call_result_created_at], name: :index_overdue_calls_call_result_created_at
+    add_index :reporting_overdue_calls, [:appointment_facility_region_id], name: :index_overdue_calls_appointment_facility
```
prabhanshuguptagit, line 4: `OverdueCallsQuery` uses these.

app/services/reports/region_cache_warmer.rb
```diff
@@ -1,69 +1,36 @@
...
-          name: name
-        },
-        class: name,
-        module: "reports"
-      }.merge(extra).merge(msg: msg)
-      Rails.logger.info data
+    def queue_job(region_type, batch_index)
+      limit = batch_size
+      offset =  batch_index * batch_size
+      RegionCacheWarmerJob.perform_async(region_type, limit, offset)
```
prabhanshuguptagit, line 33: If a region is added/removed after submitting the jobs, the limit-offset boundaries might shift. The batches are hence not guaranteed to be precise, tracking them precisely will need more code. 

It should be acceptable to miss a few regions as long as the bulk of the regions are refreshed.

app/models/reports/repository.rb
harimohanraj89, line 145: We're omitting this because it's only for DHIS2 once a month?

app/services/reports/region_cache_warmer.rb
```diff
@@ -1,69 +1,36 @@
...
-      notify "Starting #{name} warming"
+    def initialize(batch_size: BATCH_SIZE)
+      @batch_size = batch_size
     end
 
     def call
-      if Flipper.enabled?(:disable_region_cache_warmer)
-        notify "disabled via flipper - exiting"
-        return
-      end
```
harimohanraj89, line 22: Let's keep the flipper check here too. Having it here as well as inside the job ensure that
* We're not wasteful with queuing and no-opping a bunch of jobs
* We can kill queued jobs in progress if needed halfway through a warm sesh

db/migrate/20221024071410_add_reporting_facility_state_region_indexes.rb
harimohanraj89: Do we need combination indices, or can we just add indices on org, state, district, and block id? I'm assuming month_date is already indexed? (If not, 😱 )

harimohanraj89 (changes requested):
My feedback

db/migrate/20221024071410_add_reporting_facility_state_region_indexes.rb
prabhanshuguptagit: The combination indexes make queries that query by both columns faster than two separate indexes. We usually always have a `month_date` filter in our queries, hence the combination indexes. 

We have a composite index for `month_date` and `facility_region_id`, so if someone makes a query with `month_date` alone, that one can be used.

app/services/reports/region_cache_warmer.rb
prabhanshuguptagit: > We can kill queued jobs in progress if needed halfway through a warm sesh

Unsure about this, how would this work?

app/models/reports/repository.rb
prabhanshuguptagit: Yes.

app/services/reports/region_cache_warmer.rb
harimohanraj89: I mean - if we keep the feature flag inside the job also, then we can disable the flag, say, midway through a warming session, and the rest of the jobs in the queue would no-op. I don't know, in case DB CPU was spiking or something.

db/migrate/20221024071410_add_reporting_facility_state_region_indexes.rb
harimohanraj89: Nice, makes sense to me. Composite! That's the word I was looking for 😎

harimohanraj89 (approved):
This looks great, I'm a big fan of pushing each region into a separate job 👌

app/services/reports/region_cache_warmer.rb
prabhanshuguptagit: Oh right. +1 for keeping in both places.

app/jobs/reports/region_cache_warmer_job.rb
```diff
@@ -0,0 +1,33 @@
+class Reports::RegionCacheWarmerJob
+  include Sidekiq::Worker
+  include Sidekiq::Throttled::Worker
+
+  sidekiq_options queue: :default
+  sidekiq_throttle(
+    concurrency: {limit: 1}
```
harimohanraj89, line 7: Holupaminute. Was this revert intentional? https://github.com/simpledotorg/simple-server/pull/4330/commits/3f899f82a99a575e2164e0264154a2fcb107c4a6

I think it's fine though. Keep the concurrency at 1 first, make sure the job works well, then crank it up.

prabhanshuguptagit: Accidental change, just pushed ba0ed099d. We'll start at 2 and then crank up. Keeping it at 1 won't give us any benefits, so.

prabhanshuguptagit:
> pushing each region into a separate job

We're actually pushing a batch of regions into separate jobs. 1 job per region would cause a lot of SQL queries. Keeping it to a smaller batch lets us keep this jobs small and yet not make a lot of queries.

harimohanraj89:
Ah yup I see it now 🆗

PR #4340, opened 2022-11-02
Progress Tab: Phase 2 Monthly Reports, HTN Reports, Overdue Call Result

PriyangaPKini:
**Story cards:** 
[sc-7633](https://app.shortcut.com/simpledotorg/story/7633/phase-2-add-total-overdue-patients-contacted-to-achievements-card-in-progress-tab-home-page)
[9294](https://app.shortcut.com/simpledotorg/story/9294/phase-2-populate-monthly-reports)
[7365](https://app.shortcut.com/simpledotorg/story/7365/phase-2-populate-htn-reports)

## Because

- Monthly Reports 
  - Current monthly reports don't show monthly data breakdown by diagnosis and gender on one page
- The achievement section does not have over due call results section

## This addresses

- Monthly Reports 
  - Show monthly data breakdown by diagnosis and gender on one page
- Adds the number of total overdue calls made under the Achievement section
- [refactor]Hypertension Reports
  - Extract the partial into a view component

## Test instructions

- Ensure to turn on the feature flag -`new_progress_tab_v2`
- Goto Reports -> <region> -> Progress
- Check 

1. There should be an overdue calls section under Achievements 
Screenshot 2022-11-02 at 6 41 00 PM

2. Monthly reports should have a breakdown by diagnosis and gender on the same page
Screenshot 2022-11-02 at 6 40 43 PM
Screenshot 2022-11-02 at 6 40 48 PM

3. Hypertension reports should have all the data as in the previous phase
Screenshot 2022-11-02 at 6 41 07 PM
Screenshot 2022-11-02 at 6 41 14 PM
Screenshot 2022-11-02 at 6 41 19 PM
Screenshot 2022-11-02 at 6 41 29 PM


app/assets/javascripts/standalone/progress_tab.js
```diff
@@ -187,7 +187,7 @@ function removeCSSClasses(element, cssClasses) {
 // Daily report
 //
 function updateDailyReport() {
-    const dropdown = document.getElementById("period-dropdown");
+    const dropdown = document.getElementById("daily-period-dropdown");
```
prabhanshuguptagit, line 190: This ID also needs to change in the HTML. The dropdowns are still referenced as `period-dropdown` so the daily dropdown is not updating numbers in the report.

app/views/api/v3/analytics/user_analytics/show_v2.html.erb
```diff
@@ -212,16 +209,16 @@
...
+      current_user: @current_user,
+      period: @period,
+      drug_stock_query: @drug_stock_query,
+      drugs_by_category: @drugs_by_category,
+      last_updated_at: @user_analytics.last_updated_at,
+      cohort_data: @user_analytics.statistics.dig(:cohorts)
+    ))
     %>
 
     <%= render partial: "api/v3/analytics/user_analytics/diagnosis_report",
```
prabhanshuguptagit, line 224: Let's leave a todo here to refactor this into a view component and remove the partial?

app/services/reports/facility_progress_service.rb
```diff
@@ -122,51 +131,87 @@ def dimension_combinations_for_v1(indicator)
...
+
+    memoize def monthly_registrations_breakdown
+      registrations_breakdown(period_type: "monthly", facility_data: @monthly_facility_data)
+    end
+
+    memoize def monthly_follow_ups_breakdown
+      follow_ups_breakdown(period_type: "monthly", facility_data: @monthly_facility_data)
+    end
+
+    memoize def registrations_breakdown(period_type:, facility_data:)
```
prabhanshuguptagit, line 170: The method's signature fits very well with the `daily_` and `monthly_` column names in the matviews. V nice 👌🏼

app/components/progress_tab/hypertension/diagnosis_report_component.html.erb
```diff
@@ -0,0 +1,41 @@
...
+        <%= inline_file("chevron-left.svg") %>
+      </div>
+    </a>
+    <div class="pr-16px pl-16px">
+      <h1 class="m-0px mb-8px fw-bold fs-24px">
+        <%= t("progress_tab.hypertension_report_title") %> report
+      </h1>
+      <p class="m-0px p-0px ta-left fw-normal fs-16px c-grey-dark">
+        <%= t("progress_tab.last_updated_at",
+              date: display_date(@last_updated_at),
```
prabhanshuguptagit, line 18: [non-blocking] I ran into this locally: If `last_updated_at` is not present in the redis cache, this breaks the view since `display_date` can't handle nil values. We can put this in a story and fix later: ideally missing redis keys shouldn't break anything on the view and `display_date` should show an empty string.

config/locales/en.yml
```diff
@@ -182,6 +182,9 @@ en:
       follow_up_patients_card:
```
prabhanshuguptagit, line 182: What's the difference in usage between `follow_up_patients_card` and `follow_up_visits_card`?

app/services/reports/facility_progress_service.rb
prabhanshuguptagit: Let's make these two methods private, since the other methods will act as the public interface.

prabhanshuguptagit (changes requested):
Looks stellar 💯 
- Let's check time taken to render the progress tab before/after this change. The progress tab endpoint receives a lot of traffic so we might have to pre-populate some of the new queries (the overdue calls one specifically) if they're expensive.
- There's no overdue calls in our seed data so I couldn't test the achievements section. Is there a quick way to get some overdue calls populated in the reports?

app/services/reports/facility_progress_service.rb
```diff
@@ -50,6 +55,10 @@ def total_follow_ups
       total_counts[:monthly_follow_ups_htn_or_dm]
     end
 
+    memoize def total_overdue_calls
+      Reports::FacilityState.where(facility: facility).pluck(:monthly_overdue_calls).compact.sum
```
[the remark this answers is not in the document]
PriyangaPKini: Sure!

app/assets/javascripts/standalone/progress_tab.js
PriyangaPKini: good catch.

app/views/api/v3/analytics/user_analytics/show_v2.html.erb
PriyangaPKini: Sure!

app/components/progress_tab/hypertension/diagnosis_report_component.html.erb
PriyangaPKini: Done

config/locales/en.yml
PriyangaPKini: In yearly reports, we add up the number of monthly reports. The aggregate number now indicates follow visits made at that facility and not the unique patients visited at that facility.
The same id true for monthly reports as well.
More details here
https://simpledotorg.slack.com/archives/C0190BDFRK2/p1664967435685769?thread_ts=1664887896.482949&cid=C0190BDFRK2

app/components/progress_tab/monthly_report_component_v2.rb
```diff
@@ -0,0 +1,48 @@
...
+
+class ProgressTab::MonthlyReportComponentV2 < ApplicationComponent
+  include AssetsHelper
+  include ProgressTabHelper
+
+  MONTH_DATE_FORMAT = Date::DATE_FORMATS[:month_year]
+
+  attr_reader :service, :current_user, :title, :subtitle, :region
+
+  def initialize(service, current_user, title:, subtitle:)
```
olttwa, line 11: Some args are named, and other positional. Is this intentional?

app/helpers/application_helper.rb
```diff
@@ -24,6 +24,10 @@ def bootstrap_class_for_flash(flash_type)
   end
 
   def display_date(date_time)
+    unless date_time.present?
+      date_time
+
+    end
     if date_time.is_a? String
       DateTime.parse(date_time).strftime(STANDARD_DATE_DISPLAY_FORMAT)
     else
```
olttwa, line 33: Can you explain the use of & in `date_time&.strftime()`? I don't understand it & Google didn't help

app/helpers/application_helper.rb
```diff
@@ -24,6 +24,10 @@ def bootstrap_class_for_flash(flash_type)
   end
 
   def display_date(date_time)
+    unless date_time.present?
+      date_time
+
+    end
     if date_time.is_a? String
```
olttwa, line 31: IMO, A switch case of instance checks for string, datetime? & defaulting to return date_time would be more readable than doing a `.present?` check followed by if/else.

app/services/reports/facility_progress_service.rb
```diff
@@ -122,51 +131,89 @@ def dimension_combinations_for_v1(indicator)
       end
     end
 
+    memoize def monthly_total_follow_ups
```
olttwa, line 134: Do we use memoize frequently? Since past data is allowed to be modified & syncs happen after a user is online, the cache data might be stale. Can we switch from memoize to a cache which can be invalidated?

olttwa (changes requested):

app/helpers/application_helper.rb
PriyangaPKini: `&.` will only call the function `strftime()` if `date_time` is not `nil`

app/helpers/application_helper.rb
PriyangaPKini: Will make it more readable

app/services/reports/facility_progress_service.rb
PriyangaPKini: We use it frequently. memoize caches result for a req thread. 

We have a mechanism to invalidate the Redis cache. We invalidate the cache by passing a `force_bust`(boolean) param with the request.

app/components/progress_tab/monthly_report_component_v2.rb
PriyangaPKini: Will make it all named arguments for consistency

app/services/reports/facility_progress_service.rb
olttwa: Aha, so memoize caches results inside redis? I was under the assumption it's in-memory caching like Clojure

app/components/progress_tab/hypertension/diagnosis_report_component.html.erb
prabhanshuguptagit: Can you drop a link to the story here for posterity?

config/locales/en.yml
prabhanshuguptagit: Aren't monthly reports are counting unique patient visits on a monthly visit? The query does a distinct on `patient_id` for each month https://github.com/simpledotorg/simple-server/blob/35b42b06bc6ca53ecc3bd8873a240b8c92076fa0/db/views/reporting_facility_monthly_follow_ups_and_registrations_v01.sql#L57

app/services/reports/facility_progress_service.rb
prabhanshuguptagit: This comes from https://github.com/tycooon/memery. It's just caching with instance variables with some sugar. Since each request to the controller creates a new instance of `FacilityProgressService` it ends acting as a per request in-memory cache.

PR #4347, opened 2022-11-08
Add yearly reports

PriyangaPKini:

## Because

## This addresses

## Test instructions


app/services/reports/facility_progress_service.rb
```diff
@@ -29,6 +30,7 @@ def initialize(facility, period)
           .for_region(region)
           .where("month_date > ?", 6.months.ago.beginning_of_month.to_date)
           .load
+      @yearly_facility_data = FacilityYearlyFollowUpsAndRegistrationsQuery.new(facility, @current_user).call.values
```
harimohanraj89, line 33: Looks like we're jumping through some extra hoops with `.values` here to make the data structure consistent with monthly and daily data. Perhaps we can update the query class to return an array/collection directly instead.

spec/queries/facility_yearly_follow_ups_and_registrations_query_spec.rb
```diff
@@ -0,0 +1,113 @@
+require "rails_helper"
+
+RSpec.describe FacilityYearlyFollowUpsAndRegistrationsQuery do
+  it "yearly reports" do
```
harimohanraj89, line 4:
```ruby
RSpec.describe FacilityYearlyFollowUpsAndRegistrationsQuery do
  describe "#call" do
    it "aggregates yearly reports from monthly data for HTN and DM" do
      # Make sure the flag is diabled in this big spec
      # The big spec
    end

    it "aggregates by financial year when progress_financial_year is enabled" do
      # The tiny spec
    end
  end
end
```

spec/queries/facility_yearly_follow_ups_and_registrations_query_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+    expect(yearly_reports[6.months.ago.year]["yearly_follow_ups_htn_and_dm_female"]).to eq(2)
+
+    expect(yearly_reports[6.months.ago.year]["yearly_follow_ups_htn_male"]).to eq(0)
+    expect(yearly_reports[6.months.ago.year]["yearly_follow_ups_htn_female"]).to eq(2)
+    expect(yearly_reports[6.months.ago.year]["yearly_follow_ups_htn_transgender"]).to eq(1)
+    expect(yearly_reports[6.months.ago.year]["yearly_follow_ups_dm_male"]).to eq(1)
+    expect(yearly_reports[6.months.ago.year]["yearly_follow_ups_dm_female"]).to eq(2)
+    expect(yearly_reports[6.months.ago.year]["yearly_follow_ups_dm_transgender"]).to eq(0)
+
+    expect(yearly_reports[3.years.ago.year].values).to eq([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3.years.ago.year])
```
harimohanraj89, line 69:
```ruby
expect(data_thing.except("year").values).to all?(eq(0)) # or whatever this syntax is
```

app/queries/facility_yearly_follow_ups_and_registrations_query.rb
```diff
@@ -0,0 +1,58 @@
...
+    month_date
+    state_region_id
+  ]
+  FINANCIAL_YEAR_START_MONTH = 4
+  YEAR_START_MONTH = 1
+
+  def initialize(region, current_user)
+    @region = region
+    @current_user = current_user
+    @year_start_month = Flipper.enabled?(:progress_financial_year, @current_user) ? FINANCIAL_YEAR_START_MONTH : YEAR_START_MONTH
```
harimohanraj89, line 17: Consider a name that's agnostic of touchpoint in the app (remove "progress"), and very specific about april and march and stuff.

✨ `yearly_reports_start_from_april` ✨ 
`yearly_reports_start_in_april`
`yearly_reports_from_april_to_march`
`yearly_reports_from_april`
`...`

harimohanraj89 (changes requested):
Feedback below!

app/queries/facility_yearly_follow_ups_and_registrations_query.rb
```diff
@@ -22,11 +22,11 @@ def call
     records_per_period = records.group_by { |facility|
       year(facility.month_date)
     }
-    records_per_period.each_with_object(Hash.new(0)) do |record_per_period, result|
+    records_per_period.each_with_object([]) do |record_per_period, result|
```
harimohanraj89, line 25: Looks like this can just be a `records_per_period.map` now!

app/queries/facility_yearly_follow_ups_and_registrations_query.rb
```diff
@@ -47,11 +47,9 @@ def year(month_date)
   def totals(yearly_data)
     count_columns = Reports::FacilityMonthlyFollowUpAndRegistration.column_names - NON_COUNT_FIELDS.map(&:to_s)
     yearly_data.each_with_object(Hash.new(0)) do |monthly_data, totals|
-      monthly_data.attributes.map do |key, value|
-        if count_columns.include?(key)
-          key = "yearly" + key[7..]
-          totals[key] += value
-        end
+      count_columns.map do |key|
+        new_key = key.gsub("monthly", "yearly")
```
harimohanraj89, line 51: This looks great. We could ensure that the substitution happens only in the beginning of the key with a regex

`/^monthly/`, what do you think?

harimohanraj89 (approved):
Final thoughts, approving to unblock you. Have fun! 🚀 ✨

app/queries/facility_yearly_follow_ups_and_registrations_query.rb
PriyangaPKini: yeah, that's right

app/queries/facility_yearly_follow_ups_and_registrations_query.rb
PriyangaPKini: yeah, that makes sense.

spec/queries/facility_yearly_follow_ups_and_registrations_query_spec.rb
```diff
@@ -83,17 +15,69 @@
...
+      expect(yearly_reports_of_6_months_ago["yearly_follow_ups_htn_and_dm_female"]).to eq(2)
+
+      expect(yearly_reports_of_6_months_ago["yearly_follow_ups_htn_male"]).to eq(0)
+      expect(yearly_reports_of_6_months_ago["yearly_follow_ups_htn_female"]).to eq(2)
+      expect(yearly_reports_of_6_months_ago["yearly_follow_ups_htn_transgender"]).to eq(1)
+      expect(yearly_reports_of_6_months_ago["yearly_follow_ups_dm_male"]).to eq(1)
+      expect(yearly_reports_of_6_months_ago["yearly_follow_ups_dm_female"]).to eq(2)
+      expect(yearly_reports_of_6_months_ago["yearly_follow_ups_dm_transgender"]).to eq(0)
+
+      expect(yearly_reports_of_3_years_ago.except("year").values.all? { |x| x.zero? }).to be_truthy
```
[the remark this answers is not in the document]
PriyangaPKini: Great! This is more clean and readable

PR #4360, opened 2022-11-17
Default to zero when registered patients is nil

PriyangaPKini:
**Story card:** [sc-9496](https://app.shortcut.com/simpledotorg/story/9496/progress-tab-default-to-zero-when-registered-patients-are-nil)

## Because

The cohort reports in the old progress tab break when the number of registered patients is zero

## This addresses

Defaults to zero when the number of registered patients is zero

## Test instructions

- Goto Reports> `Choose a facility with zero registered patients at some quarter`> Progress tab > Cohort Reports
- Check if the Cohort reports are rendered as expected

*Before:*
Screenshot 2022-11-17 at 2 31 17 PM

*After:*
Screenshot 2022-11-17 at 2 30 57 PM

### DIFF
--- a/app/views/api/v3/analytics/user_analytics/_progress_cohort_report.html.erb
+++ b/app/views/api/v3/analytics/user_analytics/_progress_cohort_report.html.erb
@@ -30,7 +30,7 @@
     <% @user_analytics.statistics.dig(:cohorts).each do |cohort| %>
       <div class="card cohort">
         <h6>
-          <%= cohort[:registered] %> <%= raw t("analytics.patient", count: cohort[:registered]) %>
+          <%= registered_patients = cohort[:registered] || 0 %> <%= raw t("analytics.patient", count: registered_patients) %>
           <%= raw t("analytics.registered_in") %> <%= cohort[:registration_period] %>
         </h6>
         <div style="font-size: 16px; margin: 8px 0 12px 0;">
@@ -38,7 +38,7 @@
         </div>
         <table class="cohort-bars">
           <tr>
-            <% if cohort[:registered].blank? %>
+            <% if registered_patients.zero? %>
               <td class="cohort-none">
                   <%= raw t("analytics.no_patients") %>
               </td>


app/views/api/v3/analytics/user_analytics/_progress_cohort_report.html.erb
```diff
@@ -30,15 +30,15 @@
     <% @user_analytics.statistics.dig(:cohorts).each do |cohort| %>
       <div class="card cohort">
         <h6>
-          <%= cohort[:registered] %> <%= raw t("analytics.patient", count: cohort[:registered]) %>
+          <%= registered_patients = cohort[:registered] || 0 %> <%= raw t("analytics.patient", count: registered_patients) %>
```
harimohanraj89, line 33: The first ERB tags should be `<% %>`. We don't want/need to print the result the assignment to the page.

harimohanraj89: Do we want to use a helper method like this one? https://github.com/simpledotorg/simple-server/blob/master/app/helpers/dashboard_helper.rb#L7

harimohanraj89: Or one step further, do we want to do this defaulting straight at the source, inside `UserAnalytics`?

harimohanraj89 (changes requested):
Main feedback is around where to do the defaulting.

app/views/api/v3/analytics/user_analytics/_progress_cohort_report.html.erb
PriyangaPKini: >The first ERB tags should be <% %>. We don't want/need to print the result the assignment to the page.

We need this `<%= %>` because we are showing this number on the page. 

Also, this view will be deleted when we release the new progress tab. Shall we just keep it this way?

PR #4364, opened 2022-11-21
Progress tab:Add template for no data

PriyangaPKini:

## Because

## This addresses

## Test instructions


app/components/progress_tab/daily_progress_component_v2.rb
```diff
@@ -28,6 +29,10 @@ def last_30_days
     (@start..@now).to_a.reverse.map { |date| display_date(date) }
   end
 
+  def is_data_available_for(date)
+    data_available_for_dates.include?(date.to_date)
```
prabhanshuguptagit, line 33: Should do `daily_registrations_breakdown[date.to_date].present?` instead. `includes?` is an array lookup in `O(n)`, a hash lookup would be `O(1)`.

prabhanshuguptagit (approved):

app/components/progress_tab/daily_progress_component_v2.rb
PriyangaPKini: yeah, will make the change

PR #4365, opened 2022-11-22
Nov update

jamiecarter7:
**Story card:** None

## Because

Adding new updates for nov

## This addresses

New features released

## Test instructions

Appears on the dashboard home screen
-**Limited to India and Bangladesh currently**

Removing the India flipper tag should make the section disappear on localhost


dburka (approved):
Looks good. Nice fixes to capitalization too.

PR #4367, opened 2022-11-23
Notify patients in select states in IHCI

prabhanshuguptagit:
**Story card:** [sc-9556](https://app.shortcut.com/simpledotorg/story/9556/send-reminders-only-in-ap-tn-and-wb-in-ihci)

## Because

Restrict notifications to a select list of states in IHCI.

## This addresses

Adds a whitelist of states in India that will be included in criteria for eligibility in experiments.

### DIFF
--- a/app/models/experimentation/notifications_experiment.rb
+++ b/app/models/experimentation/notifications_experiment.rb
@@ -43,6 +43,7 @@ module Experimentation
                                              Time.current, MONITORING_BUFFER.ago
                                            ).select(:patient_id))
         .then { |patients| exclude_bangladesh_blocks(patients) }
+        .then { |patients| restrict_indian_states(patients) }
     end
 
     def self.exclude_bangladesh_blocks(patients)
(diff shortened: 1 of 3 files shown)


app/models/experimentation/notifications_experiment.rb
```diff
@@ -54,6 +55,14 @@ def self.exclude_bangladesh_blocks(patients)
         .where.not(block_region: {id: excluded_block_ids.presence})
     end
 
+    def self.restrict_indian_states(patients)
```
olttwa, line 58: Rename the method to `exclude_non_experiment_indian_states`?
`exclude` seems better than `restrict`

olttwa (changes requested):

app/models/experimentation/notifications_experiment.rb
prabhanshuguptagit: wdyt of `include_selected_indian_states`? It applies a whitelist so it's not actually `exclude`ing.

olttwa: What this function is doing is removing non_experiment states from the list.
I'll prefer `filter_non_experiment_indian_states` to `include_experiment_indian_states`. Your call. No more discussion on this.

harimohanraj89: I like `exclude_india_states`. Consistent with the exclusion above (exclude_bangladesh_blocks). Note that it's `India`, not `Indian`, to be consistent 😅

prabhanshuguptagit: >What this function is doing is removing non_experiment states from the list.
There is no notion of `non_experiment` states, only `included` states.

Going with `exclude_india_states`, can't think of a name that captures the whitelist well.

prabhanshuguptagit: Actually, `include_only_allowed_india_states` fits the bill. Picking this.

PR #4368, opened 2022-11-23
Dashboard: Chartjs - move all colors to a single occurrence

jamiecarter7:
**Story card:** [sc-9552](https://app.shortcut.com/simpledotorg/story/9552/phase-1-dry-colors-data-objects)

## Because

Duplicate colors object in report.js

There are 2 different sections with different structures for creating the graphs inside Reports.js
Each section has its own colors object with the same key values.

## This addresses

DRY out the reports.js code
Prep towards updating to v3.9.1

## Test instructions

Ensure that `medications_dispensation` flipper flag is enabled on localhost

Check that all charts render correctly on 'reports'

Charts to check:
- BP controlled
- Registrations
- BP not controlled
- Missed visits
- Treatment status of patients under care
- Lost to follow-up
- Medications - Days of medications dispensed to follow-up patients
- Cohort report - Quarterly report

Check charts on both 
- Hypertension
- Diabetes

Check that small charts render correctly inside 'trends'
- Its the tiny little line graph with no axis.

### DIFF
--- a/app/assets/javascripts/common/reports.js
+++ b/app/assets/javascripts/common/reports.js
@@ -1,30 +1,32 @@
-const color = {
-  darkGreen: "rgba(0, 122, 49, 1)",
-  mediumGreen: "rgba(0, 184, 73, 1)",
-  lightGreen: "rgba(242, 248, 245, 0.5)",
-  darkRed: "rgba(184, 22, 49, 1)",
-  mediumRed: "rgba(255, 51, 85, 1)",
-  lightRed: "rgba(255, 235, 238, 0.5)",
-  darkPurple: "rgba(83, 0, 224, 1)",
-  lightPurple: "rgba(169, 128, 239, 0.5)",
-  darkBlue: "rgba(12, 57, 102, 1)",
-  mediumBlue: "rgba(0, 117, 235, 1)",
-  lightBlue: "rgba(233, 243, 255, 0.75)",
-  darkGrey: "rgba(108, 115, 122, 1)",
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/common/reports.js
```diff
@@ -1,28 +1,29 @@
+const color = {
```
harimohanraj89, line 1: I like `colors` a bit better, since this thing's got a bunch inside.

harimohanraj89 (approved):
Very cool. Let's test this in a production-like environment since it involves assets. You can ask a server dev to plop this on Sandbox for a little bit whenever you're ready to test.

app/assets/javascripts/common/reports.js
jamiecarter7: Priyanga mentioned that this might pollute the global space as it is no longer in a local scope.

What are your thoughts?

I will do some reading on imports and global and local scope also

jamiecarter7: Ok, so after doing some digging into the server it looks like all javascript files that are present inside `app/assets/javascripts/common` are imported on every page of the server. 

Meaning all these files become global.

Following Priyanga's thinking there is a risk of variable clashes/overwrites, especially using 'colors' as a name.

I will convert this to a function that return and object of colors for the time being. Ill give it a longer name in an attempt to keep the name relatively unique and hence less likely to suffer clashes/overwrites.

govindkrjoshi (approved):

harimohanraj89:
Great balance of DRYness and scope cleanliness. :shipit:

PR #4369, opened 2022-11-23
Fixes jump to id in CPHC mapping script 

govindkrjoshi:


prabhanshuguptagit (approved):

lib/tasks/dell_demo.rake
```diff
@@ -253,7 +253,7 @@ namespace :dell_demo do
           break
         end
 
-        if input.to_i != 0
+        if input.size == 36
```
harimohanraj89, line 256: Could we get a little bit of context on what's going on here? A one-sentence PR description would suffice. Good for future-us looking back.

lib/tasks/dell_demo.rake
```diff
@@ -309,7 +309,7 @@ namespace :dell_demo do
           break
         end
 
-        if input.to_i != 0
+        if input.size != 36
```
harimohanraj89, line 312: The condition above says `==` and this one says `!=`. However, both original conditions were `!= 0`. Is this correct/intentional?

harimohanraj89 (commented):
Comments 🙃

lib/tasks/dell_demo.rake
govindkrjoshi: Earlier we were using integers as ids and now we moved to using UUIDs. So instead of checking if the input is an integer, we are checking if the input length is 36.

lib/tasks/dell_demo.rake
govindkrjoshi, line 312:
```suggestion
        if input.size == 36
```

govindkrjoshi: Good catch. This was a typo. I've fixed the condition to use `==`

PR #4370, opened 2022-11-24
Chartjs: Create baseConfig - line charts and combo bar&line charts baseConfig implementation [DM reports only]

jamiecarter7:
**Story card:** [sc-9576](https://app.shortcut.com/simpledotorg/story/9576/phase-1-dry-chartjs-config-dm-only-line-graph-all-charts-with-axes)

## Because

Large amounts of chartjs config duplication

## This addresses

This is part of a multiphase update to the chartjs 'reports.js' file.

### **Overall aim**: 
- to reduce code duplication and update to v4.0.1 at end of epic.

### **Aim of this PR**: 
1. Create a base config file for all line charts and combo line&bar charts on the dashboard
2. Base config will follow the styling currently used for said charts
3. Implement base config fully (or partially) on said charts shown on the 'Diabetes' reports page (only diabetes for now)

Charts of focus in this PR: 
These are screenshots from before this PR (master)
Screenshot 2022-11-30 at 18 52 43
Screenshot 2022-11-30 at 18 52 49
Screenshot 2022-11-30 at 18 52 54
Screenshot 2022-11-30 at 18 53 14
Screenshot 2022-11-30 at 18 53 29

### Why partial implementation on some chart?
There is a difficulty in fully implementing the base config on charts showing more than a one dataset:
- Registrations and follow-ups
- Blood sugar 200-299 or ≥300

This is due to the structure of the 'scales' object [its an array] and the way withMerge works [specifically around merging arrays].
withMerge combines arrays making it hard to overwrite a specific key value.
In this case the original scales object applied after the additional and base config are combined to maintain original config.
This is a temporary solution as in v3 of chartjs 'scales' structure is changed into an object [rather than an Array] allowing me to clean this code up after the upgrade.

## Test instructions

Please go to Dashboard > 'Reports' > 'Diabetes'

Check localhost against production for visual discrepancies between all of the follow charts:
- Blood sugar <200 (line graph)
- Registrations and follow-ups (line graph and bar combo)
- Blood sugar 200-299 or ≥300 (stack bar chat with axes)
- Missed visits (line graph)
- Lost to follow-up (line graph) - See detail below for this graph

Lost to follow-up (line graph) is a single implementation for both HTN and DM.
Please go to Dashboard > 'Reports' > 'Hypertension' and check that graph renders as expected


app/assets/javascripts/common/reports.js
```diff
@@ -1699,3 +1678,21 @@ const intersectDataVerticalLine = {
   },
 };
 
+// ---
+// Object merging functions
```
harimohanraj89, line 1682: We can axe this comment, I don't think it's required.

app/assets/javascripts/common/reports.js
```diff
@@ -1699,3 +1678,21 @@ const intersectDataVerticalLine = {
...
+function combineConfigWithBaseConfig(additionalConfig) {
+  return _.mergeWith(
+    baseLineGraphConfig(),
+    additionalConfig,
+    mergeWithCustomizerArrayMerging
```
harimohanraj89, line 1687: Could we call this something that represents the mode of merging? Maybe `mergeArraysWithConcatenation`?

app/assets/javascripts/common/reports.js
```diff
@@ -645,69 +582,32 @@ DashboardReports = () => {
...
+              pointBackgroundColor: colors.white,
+              hoverBackgroundColor: colors.white,
+              hoverBorderWidth: 2,
+              data: Object.values(data.ltfuPatientsRate),
+              type: "line",
+            },
           ],
+        },
       };
+      const config = combineConfigWithBaseConfig(
```
harimohanraj89, line 604: The juggling could be simplified a bit. I'd say you can call the custom config in each of these methods simply `config` (it should be evident that it's the config related to the method in question).

Then, you can simply do a `return combineConfigWithBaseConfig(config)`, instead of calling the method here and doing the return below.

app/assets/javascripts/common/reports.js
```diff
@@ -1669,6 +1569,85 @@ Reports = function (withLtfu) {
   };
 };
 
+// Base config object which can be built on using combineConfigWithBaseConfig(AdditionalConfigObject)
+// AdditionalConfigObject is used to add or overwrite specfic key: value - [will take priority]
+
+function baseLineGraphConfig() {
```
harimohanraj89, line 1575: Consider `baseGraphConfig`, since we "combine with base" in our methods, not "combine with baseline"

app/assets/javascripts/common/reports.js
```diff
@@ -1699,3 +1678,21 @@ const intersectDataVerticalLine = {
...
+// Object merging functions
+function combineConfigWithBaseConfig(additionalConfig) {
+  return _.mergeWith(
+    baseLineGraphConfig(),
+    additionalConfig,
+    mergeWithCustomizerArrayMerging
+  );
+}
+
+//  Array merging function within object
```
harimohanraj89, line 1691: If we rename the method as suggested above, this comment can be axed.

app/assets/javascripts/common/reports.js
```diff
@@ -1699,3 +1678,21 @@ const intersectDataVerticalLine = {
   },
 };
 
+// ---
+// Object merging functions
+function combineConfigWithBaseConfig(additionalConfig) {
```
harimohanraj89, line 1683: I'd say we can even name this method `withBaseConfig`.

harimohanraj89 (changes requested):
Very cool PR! I've left a bunch of feedback around the naming and juggling of stuff to hopefully make the code more intentional without as many comments, etc.

app/assets/javascripts/common/reports.js
jamiecarter7: Removed comment

app/assets/javascripts/common/reports.js
jamiecarter7: Great suggestion, I will do this with the configs

Note: there will be 2 graphs where this is currently not possible until after the upgrade due to how the 'scales' object is handled in this version - I will include these in a tidy up for a future PR

app/assets/javascripts/common/reports.js
jamiecarter7: The idea for this name is that this is the "base - line graph - config" rather than "baseline".
My intention is to explore splitting the config for line graphs and bar charts for simplified modifications in future.

app/assets/javascripts/common/reports.js
jamiecarter7: Done

app/assets/javascripts/common/reports.js
harimohanraj89: Ohhhh that makes perfect sense. Neato.

harimohanraj89: If that's the case, let's perhaps rename the `withBaseConfig()` method to `withBaseLineConfig()`, since that's the one that this method adds on. We can then add `withBaseGraphConfig()` later as well. Does that make sense?

harimohanraj89 (approved):
I have one last suggestion, but it's not a blocker. This looks great! I will let you test this out thoroughly before shipping 👍

app/assets/javascripts/common/reports.js
jamiecarter7: Changes done 👍

PR #4372, opened 2022-11-29
Design fixes

jamiecarter7:
**Story card:** None
Jamie bug list - https://rtsl.monday.com/boards/2648270782/pulses/3575560575 

## Because

Number alignements
Margin fix
Overflow bug when using smaller screen

## This addresses

Numbers listed in tables should be aligned right
There is a margin bug between this element and the element below
Before:
Screenshot 2022-11-29 at 11 58 35
After fix:
Screenshot 2022-11-29 at 12 02 04

Overflow bug before:
Screenshot 2022-11-29 at 12 29 09
Overflow bug after:
Screenshot 2022-11-29 at 12 30 29

Also fixed overflow bug with patient reg and follow-ups block (DM only)
Screenshot 2022-11-29 at 13 43 17
Screenshot 2022-11-29 at 13 29 19

## Test instructions

Go to simple dashboard > reports > 'diabetes' (for any region/district/facility)

Check spacing at full screen

Check horizontal scrolling at small widths

### DIFF
--- a/app/components/dashboard/diabetes/measurement_child_comparison_table_component.html.erb
+++ b/app/components/dashboard/diabetes/measurement_child_comparison_table_component.html.erb
@@ -9,7 +9,7 @@
   ) do |c| %>
     <%= c.tooltip({ "Types of blood sugars measured" => t("bs_measurement_details_copy.numerator", region_name: @region.name) }) %>
   <% end %>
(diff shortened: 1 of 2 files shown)


dburka (approved):
Great work. Nice to see a little clean-up. Good use of classes.

PR #4376, opened 2022-11-29
Dashboard facility level 'facility activity' design fixes

jamiecarter7:
**Story card:** 

None (monday - https://rtsl.monday.com/boards/2648270782/pulses/3494472243)

## Because

Alignment of numbers in the tables are centered 
'users' column feels off with different widths

## This addresses

Numbers in tables should be right aligned
'Users' columns now has the same width for consistency

Hypertension
Before:
Screenshot 2022-11-29 at 15 35 00

After:
Screenshot 2022-11-29 at 15 35 58

Diabetes
Before:
Screenshot 2022-11-29 at 15 37 14

After:
Screenshot 2022-11-29 at 15 39 57

## Test instructions

Dashboard > reports > facility ('facility level' - lowest level) > 'Compare performance' section.

Check BOTH hypertension and diabetes pages.

All numbers in both tables should be right aligned
'Users' columns should have the same width

### DIFF
--- a/app/components/dashboard/diabetes/healthcare_worker_activity_component.html.erb
+++ b/app/components/dashboard/diabetes/healthcare_worker_activity_component.html.erb
@@ -9,7 +9,7 @@
     <%= render Dashboard::Card::TableComponent.new(id: "measurementChildComparisonTable") do |table| %>
       <% table.column_group do %>
         <colgroup>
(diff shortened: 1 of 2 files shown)


dburka (approved):
Nice simple fixes. Well done.

PR #4602, opened 2022-12-05
Progress Tab:Show no data card on Monthly and Yearly Reports

PriyangaPKini:
**Story card:** [sc-9720](https://app.shortcut.com/simpledotorg/story/9720/handle-missing-data-in-progress-tab-before-matviews-are-refreshed)

## Because

Progress Tab reports didn't handle no-data cases .i.e accessing report before a matview refresh

## This addresses

Shows a 'data coming soon' card when mat-view is not refreshed

## Test instructions
Screenshot 2022-12-05 at 7 14 29 PM

Screenshot 2022-12-06 at 3 04 57 PM


app/components/progress_tab/yearly_report_component_v2.rb
```diff
@@ -30,7 +30,11 @@ def render?
     Flipper.enabled?(:yearly_reports_start_from_april, @current_user)
   end
 
-  def last_n_years
+  def is_data_available_for(year)
```
harimohanraj89, line 33: Consider avoiding `is`, and using something like `def data_available?(year:)`

harimohanraj89 (changes requested):
Tiny feedbacks

app/components/progress_tab/yearly_report_component_v2.rb
PriyangaPKini: yeah, will do

PR #4629, opened 2022-12-06
Update qa server deployment ip

roypeter:
## Because

Update QA server deployment IP

## This addresses
IP got updated due to an instance restart

## Test instructions

NA


PriyangaPKini (approved):
:shipit:

PR #4656, opened 2022-12-12
Add Swagger for Screening Reports MVP

olttwa:
Co-authored-by: Prabhanshu Gupta <guptaprabhanshu1@gmail.com>

**Story card:** [sc-9758](https://app.shortcut.com/simpledotorg/story/9758/create-swagger-for-screening-reports-mvp)

## Because

Mobile devs can begin implementing the Sync APIs considering the Swagger as finalized contract.

## Test instructions

Checkout `sc-9758` branch locally, run rails server & go to the APIs
1. Questionnaires Sync to user - WIP
2. [Questionnaire-Responses Sync to user](http://127.0.0.1:3000/api-docs/index.html#tag/Questionnaire-Responses/paths/~1api~1v4~1questionnaire_responses~1sync/get)
3. [Questionnaire-Responses Sync from user](http://127.0.0.1:3000/api-docs/index.html#tag/Questionnaire-Responses/paths/~1api~1v4~1questionnaire_responses~1sync/post)

### DIFF
--- a/app/schema/api/v4/models.rb
+++ b/app/schema/api/v4/models.rb
@@ -389,6 +389,31 @@ class Api::V4::Models
        required: %w[protocol_drug_id in_stock received]}
     end
 
+    def questionnaire_response
+      {type: :object,
+       properties: {
+         id: {"$ref" => "#/definitions/uuid"},
+         deleted_at: {"$ref" => "#/definitions/nullable_timestamp"},
+         created_at: {"$ref" => "#/definitions/timestamp"},
+         updated_at: {"$ref" => "#/definitions/timestamp"},
+         # TODO: Change non_empty_string to questionnaire_type enum.
+         questionnaire_type: {"$ref" => "#/definitions/non_empty_string"},
+         questionnaire_id: {"$ref" => "#/definitions/uuid"},
+         facility_id: {"$ref" => "#/definitions/uuid"},
+         user_id: {"$ref" => "#/definitions/nullable_uuid"},
+         content: {type: :object}
+       },
+       required: %w[id
+        created_at
+        updated_at
+        questionnaire_type
+        questionnaire_id
+        facility_id
+        user_id
+        content]
+      }
+    end
+
     def definitions
       {
         activate_user: activate_user,
@@ -425,6 +450,8 @@ class Api::V4::Models
         phone_numbers: array_of("phone_number"),
         prescription_drug: Api::V3::Models.prescription_drug,
         prescription_drugs: array_of("prescription_drug"),
+        questionnaire_response: Api::V4::Models.questionnaire_response,
+        questionnaire_responses: array_of("questionnaire_response"),
         teleconsultation: teleconsultation,
         teleconsultations: array_of("teleconsultation"),
         timestamp: timestamp,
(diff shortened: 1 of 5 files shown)


app/schema/api/v4/models/questionnaire.rb
```diff
@@ -0,0 +1,65 @@
+class Api::V4::Models::Questionnaire
```
harimohanraj89, line 1: This is the part that I don't understand - are we baking in the _screening_ form schema into this? How does that work with us updating the schema later? Will we update API versions if we make breaking changes? Additionally, this endpoint is the questionnaire endpoint, not the screening endpoint, right? Shouldn't it be more generic?

I imagined that this model would describe the general form DSL, and would allow screening and other forms to be exchanged - also allowing us to update the _screening_ form version, without having to update the _API_ version. Could you help me understand this?

harimohanraj89 (changes requested):
One primary concern from me

app/schema/api/v4/models/questionnaire.rb
olttwa: Hey Hari, addressing this 1 by 1.

> Additionally, this endpoint is the questionnaire endpoint, not the screening endpoint
Modified to `Models::Questionnaire::MonthlyScreeningReports`

> I imagined that this model would describe the general form DSL
As we add more forms, we'll update API-docs to contain a specimen layout for each DSL. We cannot generalise this using Swagger as random example strings won't make sense without the context it has currently.

> Will we update API versions if we make breaking changes?
We can go about this in 3 ways. Prabhanshu & I are still not decided on this. We'll loop Govind & you tomorrow (Thu, 15 Dec) if we don't have clarity. The 3 options:

1. We send all form-layout versions to the app. App chooses which one to render
2. We bump up api-version from v4 -> v5
3. We store minimum supported app-version on server-side

olttwa: > Will we update API versions if we make breaking changes?

@harimohanraj89 we've decided to go ahead with a questionnaire_api_version to solve this problem. Refer to Screening PRD header: [API Version changes](https://docs.google.com/document/d/1oRL3RyE6BO4kMhDLxjmWSHMwtKVnknxBGuYu9yyVwn4/edit#heading=h.fqghw6d73kj8) for more details

harimohanraj89: Regarding Swagger, I understand. However, it seems a bit weird to put a specific screening schema into the Swagger spec for this generic API. I did a 10 second google for how to express a generic recursive schema in Swagger, and found this: https://stackoverflow.com/questions/36866035/how-to-refer-to-enclosing-type-definition-recursively-in-openapi-swagger

Could we use something like this so that our API spec and docs accurately represent the generic questionnaire schema, and we can enrich the API docs with an example?

harimohanraj89: Rename: ✅ 

Questionnaire API version: Looks good. I've left some clarifying questions (for my own understanding) in the PRD itself. This looks like a good approach.

harimohanraj89 (commented):
I've got one more avenue for us to pursue regarding our use of Swagger 😅 Feel free to timebox that effort, and abandon it if it's complicated or outdated. Thanks!

PR #4657, opened 2022-12-13
Add support for migrating CHC and DH patients

govindkrjoshi:
**Story card:** [sc-9728](https://app.shortcut.com/simpledotorg/story/9728/modify-script-to-transfer-chc-and-dh-data)

## Because

Importing and creating users for CHC and DH is different from that of PHC and Subcenters

## This addresses

- Separate script to import CHC and DH
- Use CHC and DH dummy locations
- Create users for CHC and DH


app/models/cphc_facility.rb
```diff
@@ -47,6 +47,52 @@ def self.create_subcenter_from_mapping(mapping)
...
+      facility.cphc_facility_id = row[:cphc_facility_id]
+      facility.cphc_facility_name = row[:cphc_facility_name]
+      facility.cphc_district_id = row[:cphc_district_id]
+      facility.cphc_district_name = row[:cphc_district_name]
+      facility.cphc_taluka_id = row[:cphc_taluka_id]
+      facility.cphc_taluka_name = row[:cphc_taluka_name]
+      facility.cphc_user_details = nil
+      facility.cphc_facility_type = "CHC"
+      facility.cphc_facility_type_id = OneOff::CphcEnrollment::FACILITY_TYPE_ID["CHC"]
+      facility.cphc_location_details = {
```
harimohanraj89, line 62: We don't need to populate location details for PHCs and HWCs?

harimohanraj89 (approved):
Just a question, but overall looks just fine. Of course, we'll rely on testing this out thoroughly to merge confidently.

app/models/cphc_facility.rb
govindkrjoshi: For PHCs and HWCs we look up the location from CPHC mappings. For CHCs and DHs we had to change this to accommodate for dummy locations.

harimohanraj89: Ah makes sense. 👍

PR #4659, opened 2022-12-13
Chartjs dry config htn line charts

jamiecarter7:
**Story card:** [sc-9592](https://app.shortcut.com/simpledotorg/story/9592/dry-chartjs-config-implement-new-config-on-line-charts-htn-only)

## Because

Part of the DRY process for the reports.js

## This addresses

Focus of this implementation: HTN only

Take the newly created baseLineGraphConfig() and use its implementation for all the line charts on the HTN page.

Charts affected:

- BP controlled
- BP uncontrolled
- HTN cumulative registrations
- HTN missed visits

Note: LTFU chart is not affected as this was updated with the DM charts (as its the same implementation for both pages)

## Test instructions

Check that hypertension graphs on this branch `chartjs-dry-config-htn-line-charts` look like they do in the picture below (taken from 
master):
Screenshot 2022-12-13 at 20 11 40


harimohanraj89 (approved):
The scales stuff comes after the upgrade if I understand correctly. Nice stuff! I will let you test thoroughly before merging.

PR #4671, opened 2022-12-16
Add Questionnaires Sync API

olttwa:
**Story card:** [sc-9733](https://app.shortcut.com/simpledotorg/story/9733/create-questionnaires-sync-api)

## Because

We are adding an API endpoint to sync offline-forms (questionnaires) to android apps. [PRD](https://docs.google.com/document/d/1oRL3RyE6BO4kMhDLxjmWSHMwtKVnknxBGuYu9yyVwn4/edit#)

## This addresses

- Adds a `Questionnaire` and `QuestionnaireVersion` model
- Adds a sync API

### On modelling Questionnaire and QuestionnaireVersion separately

The two models serve different purposes: `Questionnaire` keeps the latest version of a questionnaire type and `QuestionnaireVersion` maintains a history of the layout (for reporting).

We've made this split since `QuestionnaireVersion` cannot directly act as a sync resource:
- The version history of a questionnaire should be immutable. It shouldn't have a `deleted_at` attribute. Although we do want to propagate the deletion of a questionnaire to apps.
- We only want to keep one layout per `questionnaire_type` on the app for simplicity. This leads to edge cases around deleting questionnaires and propagating them to the app, which are tricky to handle on the server.

`QuestionnaireVersion` is supposed to act as an append-only log. Records in this table will be created only through the `Questionnaire` model. To ensure this we will add model level constraints (and a lightweight interface) as part of [this story](https://app.shortcut.com/simpledotorg/story/9786/crd-form-for-questionnaire-layouts).

### Questionnaire DSL spec version 1
(to be moved to somewhere more permanent)

`questionnaire.layout`: is a hash. It can have the following keys
- `item` (optional): is a list of sub-layouts (each of which have the same structure as a `questionnaire.layout`)
- `link_id` (optional): The ID that needs to be populated in the `content` in a questionnaire response.
- `text` (required if `link_id` is present): a string that needs to be displayed
- `type` (required if `link_id` is present): can be `integer`, `group`
- `display` (required if `link_id` is present): is a hash that can contain
    - `orientation` (required if `link_id` is present): can be vertical, horizontal
    - `view_type` (required): can be `header`, `sub_header`, `input_field`
- `validations` (required when `display.view_type` is `input_field`): is a hash
    - for type `integer`, it will contain
        - `min`: value inputted by user in this field should be >= `min`
        - `max`: value inputted by user in this field should be <= `max`


app/schema/api/v4/models.rb
```diff
@@ -394,10 +394,10 @@ def questionnaire
        properties: {
          id: {"$ref" => "#/definitions/uuid"},
          deleted_at: {"$ref" => "#/definitions/nullable_timestamp"},
-         created_at: {"$ref" => "#/definitions/timestamp"},
```
prabhanshuguptagit, line 397: We are not syncing this to the client anymore, since the server will take care of syncing just 1 questionnaire per type.

spec/models/questionnaire_version_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "rails_helper"
+
+RSpec.describe QuestionnaireVersion, type: :model do
+  describe "#localized_layout" do
```
prabhanshuguptagit, line 4: This assumes that the layout is in the correct shape. We'll add validity checks before traversing through swagger (which are being added in [this PR](https://github.com/simpledotorg/simple-server/pull/4674))

app/models/questionnaire_version.rb
```diff
@@ -0,0 +1,29 @@
+class QuestionnaireVersion < ActiveRecord::Base
+  has_one :questionnaire, required: false
+
+  def localized_layout
```
govindkrjoshi, line 4: What do you think about moving the localisation logic to the transformers?

app/models/questionnaire.rb
```diff
@@ -0,0 +1,17 @@
+class Questionnaire < ApplicationRecord
+  self.primary_key = "version_id"
```
govindkrjoshi, line 2: Why are we calling the `id` as `version_id` here?

app/models/questionnaire.rb
```diff
@@ -0,0 +1,17 @@
...
+  belongs_to :questionnaire_version, foreign_key: "version_id"
+
+  delegate :localized_layout, :id, :layout, to: :questionnaire_version
```
govindkrjoshi, line 5: This should not be necessary if we move the localisation logic to the transformers

app/models/questionnaire.rb
```diff
@@ -0,0 +1,17 @@
...
+  enum questionnaire_type: {
+    monthly_screening_reports: "monthly_screening_reports"
+  }
+
+  validates :dsl_version, uniqueness: {
+    scope: :questionnaire_type,
+    message: "has already been taken for given questionnaire_type"
+  }
+
+  scope :for_sync, -> { with_discarded.includes(:questionnaire_version) }
```
govindkrjoshi, line 16: Why are we doing an `includes` here?

govindkrjoshi:
Questionnaires and QuestionnaireVersions have a 1-1 relationship, should we denormalise the two into a single table for simplicity?

spec/transformers/api/v4/questionnaire_transformer_spec.rb
```diff
@@ -0,0 +1,6 @@
+require "rails_helper"
+
+RSpec.describe Api::V4::QuestionnaireTransformer do
+  describe "to_response" do
```
govindkrjoshi, line 4: Missing tests?

spec/models/questionnaire_version_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+         ]},
+        {"text" => "Test"}
+      ]}
+
+      questionnaire = build(:questionnaire_version, layout: layout)
+      expect(questionnaire.localized_layout).to eq(localized_layout)
+    end
+  end
+
+  it ""
```
govindkrjoshi, line 42: Extra `it`

spec/requests/api/v4/questionnaire_sync_spec.rb
```diff
@@ -0,0 +1,106 @@
+require "rails_helper"
+
+RSpec.describe "Questionnaires sync", type: :request do
```
govindkrjoshi, line 3: Let's add a test the check only one version of the questionnaire is synced?

app/controllers/api/v4/questionnaires_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def transform_to_response(questionnaire)
+    Api::V4::QuestionnaireTransformer.to_response(questionnaire)
+  end
+
+  def current_facility_records
+    []
+  end
+
+  def other_facility_records
```
govindkrjoshi, line 14: Why not just override `records_to_sync`?

app/models/questionnaire_version.rb
```diff
@@ -0,0 +1,29 @@
...
+    sub_layout
+      .then { |l| localize_text(l) }
+      .then { |l| localize_items_recursively(l) }
+  end
+
+  def localize_text(sub_layout)
+    text = sub_layout["text"]
+    return sub_layout unless text
+
+    sub_layout.merge({"text" => I18n.t(text)})
```
govindkrjoshi, line 20: What would happen if we don't have translations for a locale? 

Is it an uncaught exception? What will the API response code be in this case?

olttwa:
> Questionnaires and QuestionnaireVersions have a 1-1 relationship, should we denormalise the two into a single table for simplicity

@govindkrjoshi Having multiple versions of Questionnaires was making Sync logic quite complicated. Separating them into 2 tables helps us achieve these things:
1. `QuestionnaireVersions` are immutable now
2. `Questionnaires` can be synced to mobile like any other resource. It has `updated_at`, `deleted_at` columns like other resources

Think about the data model like this: Questionnaire is the source of truth for forms on mobile, QuestionnaireVersions is for audit purposes that existing QuestionnaireResponses can point to.

app/controllers/api/v4/questionnaires_controller.rb
olttwa: In `response_process_token`, we're following the convention followed in other Sync controllers.
```ruby
      other_facilities_processed_since: processed_until(other_facility_records) || other_facilities_processed_since,
```
So, overriding `other_facility_records` makes more sense.

app/models/questionnaire_version.rb
olttwa: Currently, missing translations simply read `translation missing: text`.

Adding a `!` to `I18n.t!` will respond with 500 when translations are missing.

olttwa: Another approach that we thought of and discarded: When translations are missing, Server sends a Sentry error.
In the API response, the form with missing translation will be omitted. Response will contain other forms that do have translations.
Issue with this is updated-at cursor would have moved, and there's no way to sync missing translation form unless it's touched in DB.

govindkrjoshi: We should not expose missing translations through the api, so the response should not contain `translation missing: text`.

> Issue with this is updated-at cursor would have moved, and there's no way to sync missing translation form unless it's touched in DB.

This is an important point, it's possible that the translations get updated, how do we handle those situations?

app/controllers/api/v4/questionnaires_controller.rb
govindkrjoshi: Since you're overriding `response_process_token` you could define it as `processed_until(records_to_sync)`.

I guess my gripe with this naming is that the questionnaires are for the current facility, not other facilities. I see that we have done similar naming in facilities sync as well, but there I think it works because we are syncing other facilities anyways. 

This is not a blocker, either ways.

app/models/questionnaire.rb
olttwa: Because `version_id` is a foreign key that points to `questionnaire_version`'s `id`.

app/models/questionnaire_version.rb
olttwa: Localization is a model-level concern. Keeping it in model enables reuse in future on dashboards, reports, etc.

app/models/questionnaire.rb
olttwa: Because `Questionnaire` gets layout from `QuestionnaireVersion`,
`includes` will do just 1 query for n records.

If we were to remove `includes`, call to `.localized_layout` will result in n queries for n records.

spec/requests/api/v4/questionnaire_sync_spec.rb
olttwa: Questionnaire controller queries questionnaires table, which has uniqueness constraint on type+dsl_version. There's a test for dsl_version.

Since there can be only 1 type in the DB, this test becomes unnecessary I think?

app/models/questionnaire_version.rb
olttwa: > the response should not contain `translation missing: text.`
Agreed. That's why we'll respond with 500.

> it's possible that the translations get updated, how do we handle those situations?
In current design, the Questionnaire table's `updated_at` row must be touched to handle this situation.

spec/transformers/api/v4/questionnaire_transformer_spec.rb
prabhanshuguptagit: Added in 50e678f6d

PR #4672, opened 2022-12-20
Show activerecord logs only when rails server is run

prabhanshuguptagit:
**Story card:** -

## Because

Follow up to https://github.com/simpledotorg/simple-server/pull/4324. We added activerecord logs in development explicitly, however that's made rake tasks noisy.

## This addresses

Add [a rails tie hook](https://api.rubyonrails.org/classes/Rails/Railtie.html) to only initialise activerecord logs when rails server is run.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -62,8 +62,10 @@ Rails.application.configure do
 
   config.active_job.queue_adapter = :inline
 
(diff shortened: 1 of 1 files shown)


harimohanraj89 (approved):
I was thinking about this right now. Thank you.

PR #4683, opened 2023-01-05
Add Questionnaire Responses Sync API

olttwa:
**Story card:** [sc-9734](https://app.shortcut.com/simpledotorg/story/9734/create-questionnaire-response-sync-api)

## Because

We want to sync responses to questionnaires between the server and the client. This needs a new sync resource.

## This addresses

Adds the questionnaire responses sync resource. 
- The `content` property holds responses to a questionnaire. It's a JSON blob where the keys correspond to the question's `link_id` in the questionnaire. Questions that were not filled in will not be present in the `content` body.
- Responses to a questionnaire are stored such that we accumulate responses by different users
  - If a different set of questions are being answered by two users, keep them both
  - If a user submits the same question as another user, keep the latest response based on the `device_created_at`


app/controllers/api/v4/questionnaires_controller.rb
```diff
@@ -30,9 +30,14 @@ def force_resync?
...
   end
 
+  def current_facility_processed_since
+    return Time.new(0) if force_resync?
+    process_token[:current_facility_processed_since].try(:to_time) || Time.new(0)
+  end
+
   def response_process_token
     {
-      other_facilities_processed_since: processed_until(other_facility_records) || other_facilities_processed_since,
```
olttwa, line 35: We moved to current_facilities_processed_since from other_facilities_processed_since because the layouts get synced for current facility.

other_facilities_processed_since would have confused future devs reading this code. 

Reason we were doing other facilities before was default `current_facility_processed_since` method is doing too many things. We've overridden that method.

app/models/questionnaire_response.rb
```diff
@@ -0,0 +1,50 @@
...
+      errors.add(:facility_id, "cannot be changed for a questionnaire response")
+    end
+  end
+
+  class << self
+    def merge(attributes)
+      new_record = new(attributes)
+
+      QuestionnaireResponse.transaction do
+        existing_record = lock.with_discarded.find_by(id: attributes["id"])
```
prabhanshuguptagit, line 32: Since `content` is "shared mutable state" being read and modified across requests, we take a lock on the row to prevent race conditions. 
For example:
- Request 1 sends `content: {htn_male: 1}`
- At the same time Request 2 sends `content: {htn_female: 2}`
- In the absence of a lock, depending on the order of the requests, `content` will finally be either `{htn_male: 1}` or `{htn_female: 2}`

We want from all requests to be merged, and the end result should be
```
{htn_male: 1, htn_female: 2}
```

app/schema/api/v4/models.rb
```diff
@@ -408,23 +408,23 @@ def questionnaire_response
...
-         questionnaire_type: {type: :string, enum: Questionnaire.questionnaire_types.keys},
          questionnaire_id: {"$ref" => "#/definitions/uuid"},
+         questionnaire_type: {type: :string, enum: Questionnaire.questionnaire_types.keys},
          facility_id: {"$ref" => "#/definitions/uuid"},
          user_id: {"$ref" => "#/definitions/nullable_uuid"},
-         content: {type: :object}
+         content: {type: :object},
+         created_at: {"$ref" => "#/definitions/timestamp"},
+         updated_at: {"$ref" => "#/definitions/timestamp"},
+         deleted_at: {"$ref" => "#/definitions/nullable_timestamp"}
```
prabhanshuguptagit, line 418: This is an unrelated ordering change.

spec/controllers/api/v4/questionnaire_responses_controller_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    facility = options[:facility] || request_facility
+    questionnaire = options[:questionnaire] || create(:questionnaire)
+
+    create_list(:questionnaire_response, n, questionnaire: questionnaire, facility: facility)
+  end
+
+  it_behaves_like "a sync controller that authenticates user requests"
+  it_behaves_like "a sync controller that audits the data access: sync_to_user"
+
+  describe "a sync controller that audits the data access: sync_from_user" do
```
prabhanshuguptagit, line 31: We aren't using the `a sync controller that audits the data access: sync_from_user` shared spec because of these subtle differences:
- Since `content` can be merged from `old` requests, it is technically an `update`. This is different from other resources where `old` requests are disregarded.
- The facility ID of a questionnaire cannot be changed once set, and the spec needs to accommodate for that

swagger/v4/swagger.json
```diff
@@ -2735,9 +2735,6 @@
         "id": {
```
prabhanshuguptagit, line 2735: This is an unrelated ordering change.

spec/models/questionnaire_response_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+      existing_questionnaire_response = FactoryBot.create(:questionnaire_response, device_updated_at: timestamp)
+      updated_questionnaire_response = QuestionnaireResponse.find(existing_questionnaire_response.id)
+
+      updated_questionnaire_response.device_updated_at = existing_questionnaire_response.device_updated_at
+
+      expect(Statsd.instance).to receive(:increment).with("merge.QuestionnaireResponse.updated")
+      QuestionnaireResponse.merge(updated_questionnaire_response.attributes)
+    end
+
+    it "acquires a lock before merging to avoid race conditions" do
```
prabhanshuguptagit, line 115: We tested the lock implementation [in this commit](https://github.com/simpledotorg/simple-server/pull/4683/commits/145948f1df69cd3636dabe38c372eacbbd89af44#diff-1df17d78ab6b0319fcc921b6c0bb36359d32d5d81c267ef4cae7cf22d07e64b5R126). Rspec tests run in a transaction and testing this required us to disable them. This left the DB in a broken state which was difficult to cleanup. Added this mock test instead.

app/models/questionnaire.rb
```diff
@@ -1,4 +1,6 @@
 class Questionnaire < ApplicationRecord
+  has_many :questionnaire_responses
+
```
govindkrjoshi, line 3:
```suggestion
belongs_to :user
belongs_to :facility
```

app/validators/api/v4/questionnaire_response_payload_validator.rb
```diff
@@ -0,0 +1,19 @@
+class Api::V4::QuestionnaireResponsePayloadValidator < Api::V3::PayloadValidator
+  attr_accessor(
+    :id,
+    :questionnaire_id,
+    :questionnaire_type,
+    :facility_id,
+    :user_id,
+    :content,
```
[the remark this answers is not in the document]
olttwa: @govindkrjoshi there are 2 kinds of validations & we feel we cannot do them as part of the sync_from_user API request
1. Invalidate presence of extra keys 
  - We cannot invalidate any extra keys that don't belong to a questionnaire layout because they might belong to older questionnaire's layout
2. Validate data type of all keys, or at-least those keys that are found inside given questionnaire layout
  - This is quite expensive performance-wise. For each record, we'll have to first fetch the layout from DB, iterate over it recursively to fetch key-type pairs and do validation for each matching key in the content.

Rather than doing this, we feel it'll be easier to just create a [Metabase Dashboard subscription](https://www.metabase.com/docs/latest/users-guide/dashboard-subscriptions) to flag anomalies in content periodically.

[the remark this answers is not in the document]
olttwa: > It might be worth putting the content schema in the schema.json as well as documentation.

This cannot be done because of 2 reasons:
1. QuestionnaireResponse Sync API is for all Questionnaires, not just monthly screening report. The type of questionnaire isn't present in the Sync API payload, rather it's derived based on `questionnaire_id`
2. We want to be able to add more layouts independently of code changes. Doing schema.json based validation would mean modifying code every time we're creating newer layouts.

PR #4685, opened 2023-01-05
Show previous financial year as default year for the first 3 months

PriyangaPKini:
**Story card:** [sc-9832](https://app.shortcut.com/simpledotorg/story/9832/remove-the-upcoming-financial-year-say-april-2023-march-2024-from-the-options-for-the-first-3-months-of-every-year)

## Because

Progress Yearly report has `April 2023 - March 2024` as an option(default option). As a result, the report shows " This report is coming soon"

## This addresses

Remove the upcoming financial year (say `April 2023 - March 2024`) from the options for the first 3 months of every year.

## Test instructions

- Make sure to enable flipper feature `yearly_reports_start_from_april`  
- Goto Reports > <facility region> > Progress > Yearly
- You should be able to see `April 2023 - March 2024` as the default option and corresponding report.

### DIFF
--- a/app/components/progress_tab/yearly_report_component_v2.rb
+++ b/app/components/progress_tab/yearly_report_component_v2.rb
@@ -33,6 +33,7 @@ class ProgressTab::YearlyReportComponentV2 < ApplicationComponent
   memoize def last_n_years
     years = (SIMPLE_START_YEAR..Date.current.year).to_a.reverse
     if report_in_financial_year?
+      years.delete(Date.current.year) if Date.current.month < FINANCIAL_YEAR_START_MONTH
       years.push(SIMPLE_START_YEAR - 1)
     end
 
--- a/app/components/progress_tab/yearly_report_component_v2.rb
+++ b/app/components/progress_tab/yearly_report_component_v2.rb
@@ -31,11 +31,19 @@ class ProgressTab::YearlyReportComponentV2 < ApplicationComponent
   end
 
   memoize def last_n_years
-    years = (SIMPLE_START_YEAR..Date.current.year).to_a.reverse
     if report_in_financial_year?
-      years.delete(Date.current.year) if Date.current.month < FINANCIAL_YEAR_START_MONTH
-      years.push(SIMPLE_START_YEAR - 1)
+      start_year = SIMPLE_START_YEAR - 1
+      end_year = if Date.current.month < FINANCIAL_YEAR_START_MONTH
+        Date.current.year - 1
+      else
+        Date.current.year
+      end
+    else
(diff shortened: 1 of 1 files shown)


app/components/progress_tab/yearly_report_component_v2.rb
```diff
@@ -33,6 +33,7 @@ def data_available?(year:)
   memoize def last_n_years
     years = (SIMPLE_START_YEAR..Date.current.year).to_a.reverse
     if report_in_financial_year?
+      years.delete(Date.current.year) if Date.current.month < FINANCIAL_YEAR_START_MONTH
```
prabhanshuguptagit, line 36: Would just changing `display_year` work?
```ruby
  def last_n_years
    years = (SIMPLE_START_YEAR..Date.current.year).to_a.reverse
    years.each_with_object({}) do |year, hsh|
      hsh[year] = display_year(year)
    end
  end

  def display_year(year)
    start_date = if report_in_financial_year?
      Date.new(year - 1, FINANCIAL_YEAR_START_MONTH)
    else
      Date.new(year, YEAR_START_MONTH)
    end
    end_date = start_date + ONE_YEAR
    "#{display_date(start_date)} to #{display_date(end_date)}"
  end
```

PriyangaPKini: @prabhanshuguptagit Your suggestion only formats the display year. What we need to do is remove the year 2023 from the options and treat it as if Jan 2023, Feb 2023, and March 2023 belong to the year 2022. 
Data structure for yearly data for reference:
```
{ 
2022 => {:hypertension => {} ..}
2021 => {:hypertension => {} ..}
..
}
```
i.e Jan 2023, Feb 2023 and March 2023 data will be grouped into `2022` when `:yearly_reports_start_from_april` is enabled

prabhanshuguptagit: Ah I see. In that case `last_n_years` could do it the same way as `display_year`. The list mutation makes it hard to figure out the limits of the list. Something like
```ruby
def last_n_years
  if report_in_financial_year?
    start_year = SIMPLE_START_YEAR - 1
    end_year = if Date.current.month < FINANCIAL_YEAR_START_MONTH
      Date.current.year - 1
    else
      Date.current.year
    end
  else
    start_year = SIMPLE_START_YEAR
    end_year = Date.current.year
  end

  years = (start_year..end_year).to_a.reverse
  ...
```

PriyangaPKini: yeah, that make sense.

prabhanshuguptagit (approved):

PR #4740, opened 2023-01-20
Sri Lanka production environment migration to data center

roypeter:
**Story card:** [sc-9896](https://app.shortcut.com/simpledotorg/story/9896)

## Because

Sri Lanka production infrastructure is migrated to the new k8s environment

### DIFF
--- a/.semaphore/sri_lanka_demo_deployment.yml
+++ b/.semaphore/sri_lanka_demo_deployment.yml
@@ -24,3 +24,7 @@ promotions:
     pipeline_file: sri_lanka_production_deployment.yml
     auto_promote:
       when: (branch = 'master' AND result = 'passed') OR (tag =~ '^release-.*' AND result = 'passed')
+  - name: Sri Lanka Production Deployment k8s
+    pipeline_file: sri_lanka_production_deployment_k8s.yml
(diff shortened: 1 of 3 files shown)


harimohanraj89:
We'll probably have a period of time where we want to deploy to both servers. Can we _add_ argo instead of replacing the old pipeline with it?

roypeter:
> We'll probably have a period of time where we want to deploy to both servers. Can we _add_ argo instead of replacing the old pipeline with it?

if we run both systems in parallel, will there be any issues with the cron jobs?

.semaphore/sri_lanka_demo_deployment.yml
```diff
@@ -24,3 +24,7 @@ promotions:
     pipeline_file: sri_lanka_production_deployment.yml
     auto_promote:
       when: (branch = 'master' AND result = 'passed') OR (tag =~ '^release-.*' AND result = 'passed')
+  - name: Sri Lanka Production Deployment k8s
```
harimohanraj89, line 27: Maybe we call this "Sri Lanka Production Deployment", and the other one we update to "Sri Lanka Production Deployment AWS" or "Old" or something

PR #4788, opened 2023-01-24
Add facility-level toggle for Monthly Screening Reports

olttwa:
**Story card:** [sc-9782](https://app.shortcut.com/simpledotorg/story/9782)

## Because

Admins can enable/disable monthly screening reports for 1 facility or multiple facilities in bulk.

## This addresses

Enabling/disabling monthly screening reports for 1 facility or multiple facilities in bulk.

## Test instructions

1. Toggle `Monthly Screening Reports enabled?` facility from admin page
2. Verify the field is present with correct value in the Facilities Sync API `http://127.0.0.1:3000/api/v3/facilities/sync?limit=1000`

### DIFF
--- a/app/views/admin/facilities/_form.html.erb
+++ b/app/views/admin/facilities/_form.html.erb
@@ -43,6 +43,8 @@
         <%= form.check_box :enable_diabetes_management, id: :facility_enable_diabetes_management %>
         <h3 class="mt-5">Teleconsultation enabled?</h3>
         <%= form.check_box :enable_teleconsultation, id: :facility_enable_teleconsultation, onclick: "new FacilityTeleconsultFields().toggleTeleconsultationFields()" %>
+        <h3 class="mt-5">Monthly Screening Reports enabled?</h3>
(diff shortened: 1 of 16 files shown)


app/views/admin/facilities/_form.html.erb
```diff
@@ -43,6 +43,8 @@
         <%= form.check_box :enable_diabetes_management, id: :facility_enable_diabetes_management %>
         <h3 class="mt-5">Teleconsultation enabled?</h3>
         <%= form.check_box :enable_teleconsultation, id: :facility_enable_teleconsultation, onclick: "new FacilityTeleconsultFields().toggleTeleconsultationFields()" %>
+        <h3 class="mt-5">Monthly Screening Reports enabled?</h3>
```
harimohanraj89, line 46: Let's do sentence case. Only capital M.

harimohanraj89 (commented):
Some light feedback from me. I'll let you get official review from someone else.

Should there be more model and/or controller specs too?

PR #4792, opened 2023-01-27
Chartsjs upgrade to v3.9.1

jamiecarter7:
**Story card:** [sc-9834](https://app.shortcut.com/simpledotorg/story/9834/update-chartjs-to-v3-9-1)

## Because

Updating ChartJS to the latest v3 (v3.9.1)
- Improved config (reduced bloat)
- Additional features
- Faster rendering

## This addresses

Unable to use newer features due to older version

## Test instructions

Graphs to check:
All HTN [bar and line]
All DM [bar and line]
Trending (small graph) [line]

Checking all graphs load 
Check hovering on all graphs works are expected - no errors thrown

I have done extensive checks on the functions that were updated to ensure the data point gathered is identical.
I have checked all graphs (and all data) on my localhost for a single region to ensure the data matches between this branch and the master
I have made checks to ensure the design is unchanged - this is the case in 98% of the design. There is some padding changes when upgrading.


app/assets/javascripts/common/my_facilities.js
```diff
@@ -48,6 +48,7 @@ function createBaseTrendChartConfig() {
       elements: {
         line: {
           borderJoinStyle: "round",
+          tension: 0.4,
```
harimohanraj89, line 51: We're building up the tension I see... what does `tension` do?

app/assets/javascripts/common/reports.js
```diff
@@ -140,99 +140,111 @@ DashboardReports = () => {
...
+          beginAtZero: true,
+          min: 0,
+        },
+
+        // temporary fix for hiding y axis
+        y: {
+          display: false,
+        },
+
+        yCumulativeDiabetesRegistrations: {
```
harimohanraj89, line 170: I know we used to do this too, but can you explain to me why we choose to set a custom ID for the y-axis and use that instead of the default?

app/assets/javascripts/common/reports.js
```diff
@@ -140,99 +140,111 @@ DashboardReports = () => {
...
+          // id: "cumulativeDiabetesRegistrations",
```
harimohanraj89, line 171: We can remove

app/assets/javascripts/common/reports.js
```diff
@@ -140,99 +140,111 @@ DashboardReports = () => {
...
-          {
-            id: "monthlyDiabetesFollowups",
-            position: "right",
-            stacked: true,
+          beginAtZero: true,
+          min: 0,
+          max: cumulativeDiabetesRegistrationsYAxis.max,
+        },
+
+        yMonthlyDiabetesRegistrations: {
```
harimohanraj89, line 196: I see the object structure instead of the array structure now, nice.

app/assets/javascripts/common/reports.js
```diff
@@ -421,37 +435,34 @@ DashboardReports = () => {
...
-                return (
-                    formatNumberWithCommas(numerators[tooltipItem.index]) +
-                    " of " +
-                    formatNumberWithCommas(denominators[tooltipItem.index]) +
-                    " follow-up patients"
-                );
+            tooltip: {
+              mode: "x", // not required
+              intersect: false, // not required
+              enabled: true, // not required?
```
harimohanraj89, line 441: Let's finish this cleanup before shipping

app/assets/javascripts/common/reports.js
```diff
@@ -1297,6 +1337,7 @@ function baseLineGraphConfig() {
     type: "line",
     options: {
       animation: false,
+      clip: false,
```
harimohanraj89, line 1340: Neat, what does this do?

app/assets/javascripts/common/reports.js
```diff
@@ -1312,55 +1353,61 @@ function baseLineGraphConfig() {
...
+          },
+          ticks: {
+            autoSkip: false,
+            color: colors.darkGrey,
+            font: {
+              size: 10,
+              family: "Roboto",
             },
+            padding: 8,
+            stepSize: 25, // unitStepSize??
```
harimohanraj89, line 1405: Flagging another potential cleanup - or are we addressing this later?

app/views/reports/regions/show.html.erb
```diff
@@ -1,5 +1,4 @@
-<script src="https://cdn.jsdelivr.net/npm/chart.js@2.8.0/dist/Chart.min.js" integrity="sha256-Uv9BNBucvCPipKQ2NS9wYpJmi8DTOEfTA/nH2aoJALw=" crossorigin="anonymous"></script>
-
+<script src="https://cdn.jsdelivr.net/npm/chart.js@3.9.1/dist/chart.min.js" integrity="sha256-+8RZJua0aEWg+QVVKg4LEzEEm/8RFez5Tb4JBNiV5xA=" crossorigin="anonymous"></script>
```
harimohanraj89, line 1: This is a hell of a lot of touchpoints to include this script, and is a bit unconventional that we're loading it from a CDN, instead of installing it into the codebase via npm/yarn. Can you create a story to "revisit loading of chartJS to be more Railsy" for a future maintenance task? No need to change the way we do that in this PR.

harimohanraj89 (changes requested):

app/assets/javascripts/common/my_facilities.js
jamiecarter7: Tension controls how curvy a line is on a line graph when it crosses through a point.
0 means that there is no curve and 1 means its very curvy.
In v2.8 the default tension is 0.4, but in v3.9.1 the default tension has been changed to 0.
This is added to keep design consistency.
v3.9 default - [Line configuration defaults](https://www.chartjs.org/docs/3.9.1/configuration/elements.html#line-configuration)
You can 'increase' the tension here: https://www.chartjs.org/docs/3.9.1/charts/line.html - Click the 'setup' tab to find the tension option

app/assets/javascripts/common/reports.js
jamiecarter7: Clip is a new option that crops elements rendered inside the graph/chart at the edge of the graph/chart.
Default `clip: 0` crops 0px from the edges of the graph/chart - this is basically the border of the area.
This results in a cropping of the points that sit along the bottom of the chart.
Setting this to default stops this cropping.
Screenshot 2023-02-07 at 12 07 41
Screenshot 2023-02-07 at 12 07 26

app/views/reports/regions/show.html.erb
jamiecarter7: I agree, I already have a [story](https://app.shortcut.com/simpledotorg/story/9325/explore-cdn-import-statements-do-we-need-all-of-these) as part of the epic to address this so that its not missed 👍

app/assets/javascripts/common/reports.js
jamiecarter7: I have left this here for the time being because:
- I am unsure if this is required in the current place, it needs some research to check its not needed - here I am flagging for a check
- I am half way through the next PR where I am removing unused config and want to not mix the update with a new DRY-ing.
I separated these steps to update, then DRY.
This PR was purely to get v3.9.1 working and to keep design consistency.

app/assets/javascripts/common/reports.js
jamiecarter7: This currently is not affecting the graph but is flagged for review - in the upcoming PR

app/assets/javascripts/common/reports.js
jamiecarter7: I did miss this comment, but this will be removed in the upcoming PR drying the config.

app/assets/javascripts/common/reports.js
jamiecarter7: Each dataset added in the current setup adds a new Y axis.
Each Y axis name must match exactly the yAxisID stated in the dataset config:
```js
datasets: [
            {
              yAxisID: "yCumulativeDiabetesRegistrations",
              ...
            },
            {
              yAxisID: "yMonthlyDiabetesFollowups",
              ...
            },
            {
              yAxisID: "yMonthlyDiabetesRegistrations",
              ...
            },
```
(Scroll up to line 104 above)
This means I need change the y-axis object names to match yAxisID's.
Under the current setup the first dataset added should always be y then any other datasets will be named - this only occurs on the registration line graphs.
I have to hide the Y axis on this occasion as the default config means that it shows (even with no data present).
Its a plaster over the issue - but in the next DRY PR I have addressed this by merging axes.
Again this is purely a PR to push v3.9.1 over the line and I have comeback to remove config under the new structure in the next PR 😅 

Does this answer your question?

PR #4796, opened 2023-02-03
Check if region is managable only for districts

govindkrjoshi:


prabhanshuguptagit (approved):

prabhanshuguptagit:
[non-blocking] Wonder why didn't this break locally 🤔

govindkrjoshi:
This didn't break locally since my user had power user credentials.

prabhanshuguptagit:
Oh right. Same.

PR #4801, opened 2023-02-10
Chartjs: dry-ing config post v3.9.1 upgrade

jamiecarter7:
**Story card:** [sc-9944](https://app.shortcut.com/simpledotorg/story/9944/dry-chartjs-functions-after-v3-9-1-update)

## Because

This is a next pass DRY-ing refactor:
- Remove duplicate config related to scales
- Remove any old config previously left because the focus was just to get it working
- Some minor styling alterations (nit picking)
- A few changes to the base configs
- MedicationDispensation chart switched from bar to line base config as it follows the design more closely

For both instances of the registrations graph the following has been changed:
- The line segment of the chart has been placed onto the 'y' axis, this follows standard styling similar to the % graphs. For diabetes both lines are merged into the y axis as they use the same scaling method (follow-up is a proportion of the total).
- The bar in the graphs remains on its own y-axis as its numbers are not connected to the line.
- The bar y axis has 0 config as the grid and numbers (ticks) are not displayed at all (the grid you see is part of the line axis). This means we can have almost 0 config for this additional axis.

## This addresses

Any remaining duplicate config and the removal of any remaining config thats no longer used.

## Test instructions

Best way is to load the page on the master branch, then open a new page on this branch to make comparisons.


harimohanraj89 (approved):
This scales DRYing is beautiful. Test it and ship it!

PR #4802, opened 2023-02-10
Show est population coverage totals column

jamiecarter7:
**Story card:** [sc-10082](https://app.shortcut.com/simpledotorg/story/10082/update-to-coverage-section-to-include-a-totals-column-showing-the-total-estimate-population) Story to be created

This adds the additional total column for the coverage section.

## This addresses

This puts the design in line with the other sections in this card showing a percentage and a total.

## Test instructions

Check top level districts - this section is present in the 'Compare Performance' > `Hypertension indicators`

Please test with:
- empty estimated populations
- semi complete populations
- fully complete populations
- all populations entered as 0 for all regions

Please hover all cells during all stages to ensure correct wording on tooltip hovers

Please check the Coverage (?) tooltip is working as expected.


app/views/reports/regions/_child_data_tables.html.erb
```diff
@@ -194,7 +212,7 @@
             <%= number_with_delimiter(result.dig(:cumulative_registrations, @period)) %>
           </td>
           <% if @region.state_region? %>
-            <% if child.estimated_population.present? %>
+              <% if child.estimated_population.present? && child.estimated_population.population > 0 %>
```
prabhanshuguptagit, line 215: All these checks can be simplified to 
```
if child.estimated_population&.population > 0
```

prabhanshuguptagit (approved):
Looks good. This column is being added to `state` reports yeah (not districts)?

PR #4805, opened 2023-02-14
Add 'refused to come back' to remove-reasons of call result

PriyangaPKini:
**Story card:** [sc-10023](https://app.shortcut.com/simpledotorg/story/10023/include-new-result-of-call-option-refused-to-come-back)

## Because

 There is no option to mark the overdue patients who 'refuse to come back'. These people are incorrectly marked as either 'other' or 'not responding'

## This addresses

- Adds a new option `refused to come back` to remove reasons of a call result


PriyangaPKini:
Unit test for this particular change is already covered in [here](https://github.com/simpledotorg/simple-server/blob/7d42ea855ffb5a8ef4073a938994664294b63e89/spec/models/appointment_spec.rb#L244)

app/models/call_result.rb
```diff
@@ -22,6 +22,7 @@ class CallResult < ApplicationRecord
   enum remove_reason: {
     already_visited: "already_visited",
     not_responding: "not_responding",
+    refused_to_come_back: "refused_to_come_back",
     invalid_phone_number: "invalid_phone_number",
```
prabhanshuguptagit, line 26: We'll need to update a couple more places:
- In `Appointment#update_patient_status`, I think if the patient has refused to come back, they should be marked as `migrated`.
- Add it to `Patient#could_not_contact_reasons`
- In the dashboard overdue list `views/appointment/index.html`, we should add this reason in the dropdown to mark overdue results

I looked for `public_hospital_transfer` across the codebase, should check if there's any other touchpoints.

prabhanshuguptagit: Lets also regenerate the swagger docs (by running `rake docs`) to update the enum there.

prabhanshuguptagit (changes requested):
⠀

app/models/call_result.rb
PriyangaPKini: Thanks, @prabhanshuguptagit. I have to check with someone if we should be marking these patients as `migrated`.

danySam: @prabhanshuguptagit `refused_to_come_back` doesn't necessarily mean that they have `migrated` right?

danySam: Also, `Patient#could_not_contact_reason` seems independent from `remove_reason`, in that case, do we need to add it here?

prabhanshuguptagit: @danySam Yeah actually, think it makes sense to not add it to`Patient#could_not_contact_reason`. Seems like it's also dead code so unsure what context it was supposed to be used in.

PR #4811, opened 2023-02-17
Return errors when QuestionnaireResponse is invalid during sync_from_user

olttwa:
**Story card:** -

## Because

If a sync-from-user QuestionnaireResponse payload passes JSON schema validations but fails ActiveRecord validations, simple-server does not send errors in response. This lets the client presume that sync was successful, when it wasn't.

## This addresses

If there are any model-validation failures on a QuestionnaireResponse Sync resource, the API response would include that in the error. For example, if a POST Sync API request is made with correct JSON-schema, but non-existent `questionnaire_id`:
- the payload is not saved to the DB
- the server does not return any errors
- the client presumes the payload has been synced since there were no errors

Now, the API returns an error hash:
```
{"errors":[{"questionnaire":["must exist"],"id":"1637b6b4-7ee5-46de-bcb6-077103b948fa"}]}%
```


app/models/application_record.rb
```diff
@@ -15,4 +15,8 @@ def self.updated_on_server_since(timestamp, number_of_records = nil)
   def readonly?
     RequestStore[:readonly] || @readonly
   end
+
+  def errors_hash
+    errors.to_hash.merge(id: id)
+  end
```
prabhanshuguptagit, line 21: Unnecessary method, remove.

govindkrjoshi (approved):

govindkrjoshi:
Let's add a spec to test the condition in which the server will return the model errors.

olttwa:
> Let's add a spec to test the condition in which the server will return the model errors.

Added a spec

PR #4813, opened 2023-02-21
Cleanup migrations

harimohanraj89:
**Story card:** [sc-9948](https://app.shortcut.com/simpledotorg/story/9948/cleanup-old-migrations-and-unused-views-from-rails-app)

## Because

Spring cleaning in the style of #2513

## This addresses

Removal of
* All DB migrations
* All data migrations
* All older DB views (keep only the current version of each view)

## Test instructions

You should be able to set up Simple from scratch still.


prabhanshuguptagit (approved):
Looks good :shipit:

PR #4817, opened 2023-02-23
Performant materialized patient linelist download

govindkrjoshi:
**Story card:** [sc-8205](https://app.shortcut.com/simpledotorg/story/8205/generate-line-lists-using-a-materialized-view)

[[OLD PR]] https://github.com/simpledotorg/simple-server/pull/4691

## Because

Generating line lists for large regions is very slow and resource intensive. We have received complaints about the line list download taking upwards of 30 minutes, and have also observed our sidekiq boxes run out of resources when multiple line list download jobs are triggered (often by the same user waiting for their earlier downloads to come through). 

## This addresses
Improves the querying performance so that the mat view refreshes can terminate in a reasonable amount of time. 

Summary of time taken to run the query:

```
+--------------------------+-----------------+----------------------------+
|           Desc           | No. of Patients | Appoximate time to execute |
+--------------------------+-----------------+----------------------------+
| Query locally            | 4M              | 5.5 minutes                |
| Refresh mat view locally | 4M              | 6.6. minutes               |
| Query metabase prod BD   | 200k            | 40 seconds                 |
| Query metabase prod IN   | 3.5M            | 8 minutes                  |
+--------------------------+-----------------+----------------------------+
```

[[Details from OLD PR]]
This PR converts our line list into a materialised Postgres view that gets updates on a daily basis. This precomputes and materialises the patient summary for all the patients, so we don't have to run additional queries and load unnecessary records in memory. 

## Test instructions

Download the patient line list from the district reports page while observing the active queries in postgres. It should not make additional queries to table other than materialized_patient_summaries.

```
\watch 1 "SELECT procpid, age(clock_timestamp(), query_start), usename, current_query FROM pg_stat_activity WHERE current_query != '<IDLE>' AND current_query NOT ILIKE '%pg_stat_activity%'  ORDER BY query_start desc";
```


db/views/materialized_patient_summaries_v03.sql
```diff
@@ -0,0 +1,394 @@
...
+    select distinct on (patient_id) *
+    from patient_phone_numbers
+    where deleted_at is null
+    order by patient_id, device_created_at desc
+), latest_medical_history as (
+    SELECT DISTINCT ON (patient_id) * FROM medical_histories WHERE deleted_at IS NULL
+),ranked_prescription_drugs as (
+    select
+        bp.id as bp_id,
+        array_agg(array[name, dosage] order by prescription_drugs.is_protocol_drug desc, prescription_drugs.name, prescription_drugs.device_created_at desc) as blood_pressure_drugs,
```
govindkrjoshi, line 16: Using `group by` here instead of using rank. 

The earlier version had to join onto itself 5 times, a ranked_prescription_drugs result set of cardinality `O(no. of bps * no. of prescription drugs)`. The current version simply returns a result set of cardinality of `O(no. of bps)` with all the prescription drugs in the correct order in the array_agg.

db/views/materialized_patient_summaries_v03.sql
```diff
@@ -0,0 +1,394 @@
...
+        array_agg(name || '-' || dosage order by prescription_drugs.is_protocol_drug desc, prescription_drugs.name, prescription_drugs.device_created_at desc) as drug_strings
+    from blood_pressures bp
+        join prescription_drugs
+            on prescription_drugs.patient_id = bp.patient_id
+                and date(prescription_drugs.device_created_at) <= date(bp.recorded_at)
+                and (prescription_drugs.is_deleted is false or (prescription_drugs.is_deleted is true and date(prescription_drugs.device_updated_at) > date(bp.recorded_at)))
+    where bp.deleted_at is null and prescription_drugs.deleted_at is null
+    group by bp.id
+), blood_pressure_follow_up as (
+    select distinct on (bp.patient_id, date(bp.recorded_at)) bp.id bp_id, appointments.*
```
govindkrjoshi, line 26: Instead of using a lateral join to get the follow bp follow up, which would loop over the appointments table `O(no. of bps)` times, we can create a view of the latest follow up on the day of the BP and join using this table. The result set of the two queries is the same, and can be verified using 

```
with query1 as (
    select bp.id, a.id
    from blood_pressures bp
             left outer join (
        select distinct on (bp.patient_id, date(bp.recorded_at)) bp.id bp_id, appointments.*
        from blood_pressures bp
                 join appointments on appointments.patient_id = bp.patient_id and date(appointments.device_created_at) = date(bp.recorded_at)
        order by bp.patient_id, date(bp.recorded_at), appointments.device_created_at desc) a on a.bp_id  = bp.id
    where bp.deleted_at is null
), query2 as (
    select bp.id, a.id aid
    from blood_pressures bp
             left outer join lateral (
        select distinct on(patient_id) *
        from appointments
        where patient_id = bp.patient_id and date_trunc('day', device_created_at) = date_trunc('day', bp.recorded_at)
        order by patient_id, device_created_at desc
        ) a on a.patient_id = bp.patient_id
    where bp.deleted_at is null and a.deleted_at is null
)

(select * from query1 except  select * from query2) union all (select * from query2 except  select * from query1)
```

db/views/materialized_patient_summaries_v03.sql
```diff
@@ -0,0 +1,394 @@
...
+        bp_drugs.blood_pressure_drugs[4][1] as prescription_drug_4_name,
+        bp_drugs.blood_pressure_drugs[4][2] as prescription_drug_4_dosage,
+
+        bp_drugs.blood_pressure_drugs[5][1] as prescription_drug_5_name,
+        bp_drugs.blood_pressure_drugs[5][2] as prescription_drug_5_dosage,
+
+        (select string_agg(value, ', ') from unnest(bp_drugs.drug_strings[6:]) as value) as other_prescription_drugs,
+        (select string_agg(value, ', ') from unnest(bp_drugs.drug_strings) as value) as all_prescription_drugs,
+
+        rank() over (partition by bp.patient_id order by bp.recorded_at desc, bp.id) rank
```
govindkrjoshi, line 68: In very rare cases (outside of tests) it's possible that the patient has multiple blood pressures with exactly the same recorded at value. In these cases the two BPs would be assigned the same rank, resulting in multiple
 rows for the patient in generated view. 

To avoid having multiple rows with the same rank value earlier, we were doing `distinct on (patient_id, rank)` which is slow and time consuming. It is also incorrect, in that it drops the blood_pressure instead of giving it a successive rank value. 

Using the id as an additional sort parameter makes sure that rank is unique across alll BPs, and there are no duplicate patient summary rows in the result of this view.

db/views/materialized_patient_summaries_v03.sql
```diff
@@ -0,0 +1,394 @@
...
+        bs.blood_sugar_type,
+        bs.blood_sugar_value,
+        f.name as facility_name,
+        f.facility_type,
+        f.district,
+        f.state,
+        follow_up_facility.name as follow_up_facility_name,
+        a.scheduled_date as follow_up_date,
+        greatest(0, date_part('day', a.scheduled_date - date_trunc('day',a.device_created_at)))::int as follow_up_days,
+        rank() over (partition by bs.patient_id order by bs.recorded_at desc, bs.id) rank
```
govindkrjoshi, line 170: Same logic as using `bp.id` to rank blood pressures.

govindkrjoshi (commented):
This is still some scope to improve the performance of the query (by 10-15%, if we use group by for latest blood pressures and blood sugars, similar to prescription drugs), that comes at a significant cost to the readability of the query, and having to deal with null json values in the end result of the view. 

Since this query is already difficult to follow, I've opted to not make that change for now.

harimohanraj89 (approved):
This new view is bumpin', incredible optimizations and nice work ✨

harimohanraj89:
I'll let you get the specs verified by someone else if you need to, I'll leave that up to you.

PR #4818, opened 2023-02-23
UX fix drug stock scrolling

jamiecarter7:
No story 

## Because

Quick and dirty fix to implement browser level horizontal scroll on the drug stock and drug consumption pages

## This addresses

Praveen mentioned that it is hard to scroll horizontally on the drug stock pages because the 

## Test instructions

Add some extra medicines and ensure the whole screen becomes the scrollable object.

Things broken:
- on mobile the nav bar does not extend over the 100% screen size: most likely not visible when scrolling down and across the drug tables.
- the card wont extend to 100% width when only a few medicines present.


dburka (approved):
Approved.

PR #4819, opened 2023-02-27
Disaggregated dhis2 data: Add queries for 4 HTN indicators on patient states

qptr:
**Story card:**
[sc-10061](https://app.shortcut.com/simpledotorg/story/10061/)
[sc-10062](https://app.shortcut.com/simpledotorg/story/10062/)
[sc-10063](https://app.shortcut.com/simpledotorg/story/10063/)
[sc-10059](https://app.shortcut.com/simpledotorg/story/10059/)

## Because

We need to send disaggregated HTN indicator data to Bnagladesh DHIS2. This can also be used when generating reports that are not necessarily needed by the dashboard.

## This addresses

This adds the following queries:
- Cumulative assigned patients query
- Controlled patients query
- Uncontrolled patients query
- Disaggregated patient count query
- Missed visits query

## Test instructions

The queries can be run and the results can be tested against reporting facility states columns. We've added specs to do the same.


app/queries/patient_states/missed_visits_query.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize(region, period)
+    @region = region
+    @period = period
+  end
+
+  def call
+    PatientStates::CumulativeAssignedPatientsQuery.new(region, period)
+      .call
+      .where(hypertension: "yes")
+      .where("months_since_registration >= ?", 3)
```
harimohanraj89, line 13: Can we use the `excluding_recent_registrations` method instead?

app/exporters/disaggregated_dhis2_exporter.rb
```diff
@@ -1,11 +1,33 @@
...
-      data_elements_map: CountryConfig.current.fetch(:dhis2_data_elements)
+      data_elements_map: CountryConfig.current.fetch(:disaggregated_dhis2_data_elements)
     )
     exporter.export do |facility_identifier, period|
+      region = facility_identifier.facility.region
+      {
+        cumulative_assigned: disaggregated_counts(PatientStates::CumulativeAssignedPatientsQuery.new(region, period)),
+        controlled: disaggregated_counts(PatientStates::ControlledPatientsQuery.new(region, period)),
+        uncontrolled: disaggregated_counts(PatientStates::UncontrolledPatientsQuery.new(region, period)),
+        missed_visits: disaggregated_counts(PatientStates::MissedVisitsQuery.new(region, period))
```
harimohanraj89, line 16: If these new query classes are only used for this disaggregation, it seems like a lot of moving pieces for this one purpose. Have we considered keeping everything in one place for now, and extracting separate query classes if we start re-using the queries in other places? (eg. other DHIS2 pipes, the "repository" itself, etc).

I'm thinking something like:

```ruby

# In the DHIS2 Exporter
def self.query
  @query ||= PatientStates::DisaggregatedPatientCountQuery.new(
    region: region,
    period: period,
    age_buckets: BUCKETS
  )
end

def self.export
  # ...
  cumulative_assigned: query.cumulative_assigned,
  controlled: query.controlled,
  uncontrolled: query.uncontrolled,
  # ...
end

# And the query class would look like
class PatientStates::DisaggregatedPatientCountQuery
  def initialize(region:, period:, age_buckets:)
     # initialize stuff
  end

  def cumulative_assigned
    # the cumulative assigned query
  end

  def controlled
    cumulative_assigned.where(criteria_for_controlled_patients)
  end
end
```

Sure, it's weird for a "query" to have multiple queries inside, but I'd rather change the name of the class and still keep everything in one place, than split it up just yet. What do you think?

harimohanraj89 (changes requested):
One big thought about the class structure. The rest looks stellar.

PR #4833, opened 2023-03-06
Disaggregated dhis2 data: Remaining queries

qptr:
**Story card:**
[sc-10064](https://app.shortcut.com/simpledotorg/story/10064)
[sc-10065](https://app.shortcut.com/simpledotorg/story/10065)
[sc-10066](https://app.shortcut.com/simpledotorg/story/10066)
[sc-10067](https://app.shortcut.com/simpledotorg/story/10067)

## Because

We need to send disaggregated HTN indicator data to Bangladesh DHIS2. This can also be used when generating reports that are not necessarily needed by the dashboard. This adds on top of this [PR](https://github.com/simpledotorg/simple-server/pull/4819)

## This addresses

This adds the following queries:
- Cumulative registrations
- Monthly registrations
- Dead patients
- Lost to follow up patients

## Test instructions

The queries can be run and the results can be tested against reporting facility states columns. We've added specs to do the same.


.tool-versions
```diff
@@ -0,0 +1 @@
+postgres 14.3
```
qptr, line 1: oops. i'll delete later.

PR #4834, opened 2023-03-06
Seed questionnaire & questionnaire_responses

olttwa:
**Story card:** [sc-10142](https://app.shortcut.com/simpledotorg/story/10142/seed-monthly-screening-reports-data-in-simple-server-repository)

## Because

Mobile can run integration tests against data in review server & developers can seed Questionnaire data locally.

## This addresses

Seeding of monthly screening forms

## Test instructions

Run rake task `db:seed` & check Database for Questionnaire & QuestionnaireResponse entries.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -45,8 +45,3 @@ end
 Rake::Task["db:structure:dump"].enhance do
   Rake::Task["db:structure:clean"].invoke
 end
-
-Rake::Task["db:seed"].enhance do
-  ENV["REFRESH_MATVIEWS_CONCURRENTLY"] = "false"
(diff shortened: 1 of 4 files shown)


lib/tasks/db.rake
```diff
@@ -45,8 +45,3 @@ end
 Rake::Task["db:structure:dump"].enhance do
   Rake::Task["db:structure:clean"].invoke
 end
-
-Rake::Task["db:seed"].enhance do
-  ENV["REFRESH_MATVIEWS_CONCURRENTLY"] = "false"
```
olttwa, line 50: `PreFillMonthlyScreeningReports` depends on Mat-views to be created before-hand. Hence, we had to move this to a direct function call inside `seed/runner.rb`

PR #4857, opened 2023-03-13
Add new header required by CPHC user creation API

govindkrjoshi:
The CPHC user creation API has changed recently and requires us to send an additional header value `facilityType` set to `STATE`.


harimohanraj89 (commented):
Alrighty then 👍

PR #4872, opened 2023-03-30
Improve user enumeration prevention through auth APIs

prabhanshuguptagit:
**Story card:** [sc-9976](https://app.shortcut.com/simpledotorg/story/9976/user-enumeration-through-api-endpoints)

## Because

These endpoints are called in sequence for authenticating users
1. `/api/v4/users/find` - Takes a phone number and returns the associated user's ID if it exists.
2. `/api/v4/users/activate` - Takes a user ID, PIN and requests an OTP. Also returns the user's details (name, phone number). 

It's possible to brute force combinations of phone numbers and PINs across these APIs. This is a potential source to harvest user data. (Note that a successful brute force doesn't give access to patient data since there's an OTP step after this.)

From the [audit report](https://drive.google.com/drive/folders/1mE8MOxt-DFFL3PO1w-IbFHfMTsD0-gA-):
image

## This addresses

To prevent harvesting of user data,
- This enables rate limiting on `/activate` by user ID (in addition to IP address). 
- We rate limit `/activate` by IP only in production (to 5 requests/30 minutes, which is quite aggressive). This enables it on security envs also so it's not flagged again in audits. 
- Sets up logging on requests that get throttled. We can optionally setup alerting to flag such users via Datadog.
- Also adds a github action that can be manually triggered for running android QA tests against a branch.

Earlier work: https://github.com/simpledotorg/simple-server/pull/3371

- [x] Run a QA integration test agains this branch


.github/workflows/android-qa-tests.yml
```diff
@@ -0,0 +1,80 @@
+# This workflow needs to be kept at par with
+# https://github.com/simpledotorg/simple-android/blob/master/.github/workflows/ci_checks.yml
```
prabhanshuguptagit, line 2: This workflow is to run a QA integration test against a branch to catch regressions before releasing to master.

We had a version of this which was removed in https://github.com/simpledotorg/simple-server/pull/3287

lib/seed/runner.rb
```diff
@@ -81,8 +81,8 @@ def feature_flags_enabled_by_default
         :monthly_screening_reports,
         :follow_ups_v2_progress_tab,
         :notifications,
-        (:auto_approve_users if SimpleServer.env.android_review?),
-        (:fixed_otp if SimpleServer.env.android_review?)
+        (:auto_approve_users if SimpleServer.env.android_review? || SimpleServer.env.review?),
+        (:fixed_otp if SimpleServer.env.android_review? || SimpleServer.env.review?)
```
prabhanshuguptagit, line 85: This is to support running the qa tests.

config/initializers/rate_limiter.rb
```diff
@@ -1,67 +1,71 @@
-unless SimpleServer.env.sandbox? || SimpleServer.env.qa? || SimpleServer.env.android_review? || SimpleServer.env.review?
```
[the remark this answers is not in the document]
prabhanshuguptagit: There was another clause [here](https://github.com/simpledotorg/simple-server/pull/4872/files#diff-de9a703378551d3e4bcf7fc506f350a660177bc46459f4a19daee17f7fc1b665L55) that only allowed `SimpleServer.env` to be `production`.

harimohanraj89 (approved):
Looks great to me. I'll let you decide whether to rip out the Android QA workflow or not

.github/workflows/android-qa-tests.yml
prabhanshuguptagit: Ripping this out in favor of https://github.com/simpledotorg/simple-android/pull/4645/files

PR #4877, opened 2023-04-05
Ensure order is maintained when patient business identifier is missing

govindkrjoshi:
**Story card:** -

## Because

If a patient doesn't have a patient business identifier, the order of line lists gets messed up

## This addresses

Adds an empty cell if the business identifier is missing. This prevents the order from being messed up.

## Test instructions

Create a patient without a patient business identifier. There should be an empty cell in the corresponding column for that patient in the line list download.

### DIFF
--- a/app/exporters/patients_with_history_exporter.rb
+++ b/app/exporters/patients_with_history_exporter.rb
@@ -112,7 +112,10 @@ class PatientsWithHistoryExporter
       patient_summary.current_age.to_i,
       patient_summary.gender.capitalize,
       status(patient_summary),
-      patient_summary.latest_phone_number,
+      patient_summary.latest_phone_number
+    ]
+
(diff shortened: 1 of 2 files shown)


harimohanraj89 (approved):
Neat. The other `compact` on the address is also incorrect, right? Our intention is to either include or exclude the zone column consistently, but it might inadvertently strip out blank address info. Even if we have validations on that data, this seems fragile.

Wanna fix that too?

govindkrjoshi:
@harimohanraj89 I've made a fix to avoid the compact in address details as well. 

https://github.com/simpledotorg/simple-server/pull/4877/files#diff-824496dfb587ff8ff14d47f7f514138b41899bc05a923cdf69b78988b81cbaf7R124

PR #4879, opened 2023-04-06
Add seeder for CallResults

danySam:

## Because

## This addresses

## Test instructions

### DIFF
--- a/lib/seed/runner.rb
+++ b/lib/seed/runner.rb
@@ -9,7 +9,7 @@ module Seed
     include ActionView::Helpers::NumberHelper
     include ConsoleLogger
     SIZES = Facility.facility_sizes
-    SUMMARY_COUNTS = [:patient, :blood_pressure, :blood_sugar, :appointment, :facility, :facility_group, :prescription_drug]
+    SUMMARY_COUNTS = [:patient, :blood_pressure, :blood_sugar, :appointment, :facility, :facility_group, :prescription_drug, :call_result]
 
     attr_reader :config
     attr_reader :logger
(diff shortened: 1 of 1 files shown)


lib/seed/runner.rb
```diff
@@ -157,25 +158,43 @@ def create_progress_bar
...
-          user_id: user_id
-        }
-        attrs << FactoryBot.attributes_for(:appointment, hsh)
-      }
-      Appointment.import(attrs)
+          user_id: user_id)
+      end.compact
+
+      call_result_attributes = appointment_attributes.map do |appointment|
+        number_of_calls = config.rand_or_max(0..1) # you can get more than 1 call if you are overdue
```
olttwa, line 182: Should this be `rand_or_max(0..3)` going by the comment?

lib/seed/runner.rb
```diff
@@ -157,25 +158,43 @@ def create_progress_bar
...
           updated_at: created_at,
-          user_id: user_id
-        }
-        attrs << FactoryBot.attributes_for(:appointment, hsh)
-      }
-      Appointment.import(attrs)
+          user_id: user_id)
+      end.compact
+
+      call_result_attributes = appointment_attributes.map do |appointment|
```
olttwa, line 181: [suggestion] Moving call_results generation to a separate method will be neater. This method can provide appointment attributes to the new method.

lib/seed/runner.rb
danySam: Technically it would make sense. But looking at the data 50% call rate for appointments is satisfactory. I'll remove the comment to avoid confusion.

lib/seed/runner.rb
danySam: Accpeted

lib/seed/runner.rb
olttwa: Make sense.

PR #4920, opened 2023-04-11
Finishing touches on Questionnaire Server

olttwa:
**Story card:** [sc-10370](https://app.shortcut.com/simpledotorg/story/10370/finishing-touches-on-questionnaire-server)

## Because

Check Shortcut story

## This addresses

Check Shortcut story

## Test instructions

1. Missing translations should show-up as English


prabhanshuguptagit (approved):
Looks good. We should add a spec for `Questionnaire#localize_layout` to capture the `I18n.t!` behavior where it doesn't raise an exception if a default translation is available.

olttwa:
> where it doesn't raise an exception if a default translation is available.

@prabhanshuguptagit added a test to check undefined key raises exception, and to assert that the default/fallback locale is english.

PR #4922, opened 2023-04-12
[Spike] Download API

harimohanraj89:
**Story card:** TBD

This PR is a spike to try out a "download API"

**Use case:** Chennai corporation would like to automatically extract Chennai patient data on a regular (nightly or weekly) basis to shuttle to a TN patient registry system. We can't provide a standard API token because it has read/write privileges. We can't provide a Metabase user because we can't enforce regional constraints on the data access.

**What does this do:** 

A new API endpoint that auths requests based on a new API key called "Download API Key". This is a readonly key that auths you to your district. It doesn't touch existing sync API tokens (which are heavily coupled to app users).

### DIFF
--- a/app/exporters/patients_with_history_exporter.rb
+++ b/app/exporters/patients_with_history_exporter.rb
@@ -29,6 +29,17 @@ class PatientsWithHistoryExporter
 
   attr_reader :display_blood_sugars
 
+  def self.json(*args)
+    new.json(*args)
+  end
+
+  def json(patients, display_blood_sugars: true)
+    @display_blood_sugars = display_blood_sugars
+    summaries = MaterializedPatientSummary.where(patient: patients)
+
+    summaries.to_json
+  end
+
   def self.csv(*args)
     new.csv(*args)
   end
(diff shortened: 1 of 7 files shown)


app/exporters/patients_with_history_exporter.rb
```diff
@@ -29,6 +29,17 @@ class PatientsWithHistoryExporter
...
 
+  def self.json(*args)
+    new.json(*args)
+  end
+
+  def json(patients, display_blood_sugars: true)
+    @display_blood_sugars = display_blood_sugars
+    summaries = MaterializedPatientSummary.where(patient: patients)
+
+    summaries.to_json
```
harimohanraj89, line 40: We will need to get fine-grained with the JSON presentation of a patient summary. Matching the CSV list would be a good first step.

harimohanraj89:
Feedback from Govind:
* We should paginate and limit the responses.
* We could use devise_token_auth instead of hand rolling our own auth. https://github.com/lynndylanhurley/devise_token_auth

PR #5073, opened 2023-04-17
Add a materialised view for Overdue Patients

PriyangaPKini:
**Story card:** [sc-10149](https://app.shortcut.com/simpledotorg/story/10149/add-a-mat-view-for-creating-a-line-list-of-overdue-patients)

## Because

We want to show a line list of overdue patients and report the aggregate data on the dashboard. The current reporting pipeline doesn't allow us to easily fetch and display this data.

## This addresses

The new materialised view enables Simple to report Overdue Patient's information at the start of each reporting month. 

Since this a fairly large data, it may impact the performance of our Reporting Refreshes. A couple of profiling work was done to ensure that this doesn't impact the refresh negatively.

### Profiling

We were unable to use the production replica dataset to test the new materialised view. Instead, we used the seed data with the profiling flag to test how long it would take to complete the refresh with and without the new materialised view.

The conclusion was that the new view will take 10% longer to complete when introduced.

Before:
Before

After:
After

Other Metrics (First spike is without introducing the new materialised view and the second spike is with the new)

Read Throughput

CPU Utilization

Read IOPS

Write IOPS


db/views/reporting_overdue_patients_v01.sql
```diff
@@ -0,0 +1,89 @@
+SELECT DISTINCT ON (reporting_patient_states.month_date, reporting_patient_states.patient_id)
+  reporting_patient_states.month_date AS month_date,
+  reporting_patient_states.patient_id AS patient_id,
+  reporting_patient_states.assigned_facility_id AS assigned_facility_id,
```
govindkrjoshi, line 4: We should add the other region ids and slugs for block, district and state so that filtering a querying on the table is easier in metabase and other ad-hoc reports

db/views/reporting_overdue_patients_v01.sql
```diff
@@ -0,0 +1,89 @@
+SELECT DISTINCT ON (reporting_patient_states.month_date, reporting_patient_states.patient_id)
+  reporting_patient_states.month_date AS month_date,
```
govindkrjoshi, line 2: What do you think about adding month, year, quarter, month-string and quarter-string, to make it more convenient to use with metabase, and other ad-hoc reports and queries.

danySam: These columns will take up a lot of space on the view, right? They are primarily redundant and can be realised by using a join on `reporting_months`.

danySam: But I guess we want to prioritize convenience for the Metabase users.

db/views/reporting_overdue_patients_v01.sql
```diff
@@ -0,0 +1,89 @@
...
+    WHEN patient_phone_numbers.number IS NULL THEN 'no'
+    ELSE 'yes'
+  END AS has_phone,
+  CASE
+    WHEN previous_call_results.result_type = 'removed_from_overdue_list' THEN 'yes'
+    ELSE 'no'
+  END AS removed_from_overdue_list
+FROM reporting_patient_states
+  LEFT JOIN appointments ON reporting_patient_states.patient_id = appointments.patient_id
+  AND appointments.device_created_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')) < reporting_patient_states.month_date
```
prabhanshuguptagit, line 51: [non-blocking] A different way to write this could be
- get latest appointment per patient per month in a WITH clause
- left outer join on appointments.month = reporting_patient_states.month - 1 

Suspecting this might be more performant since the join happens on a smaller table (instead of all of appointments which is pretty large).

spec/models/reports/overdue_patient_spec.rb
```diff
@@ -0,0 +1,303 @@
...
+        create(:blood_pressure, patient: patient, recorded_at: month_date + 1.day)
+
+        RefreshReportingViews.refresh_v2
+
+        with_reporting_time_zone do
+          expect(described_class.find_by(patient_id: patient.id, month_date: month_date).is_overdue).to eq("yes")
+        end
+      end
+
+      it "should be no when the previous appointments scheduled date and visited date is in the previous month and the visit took place after the scheduled date" do
```
prabhanshuguptagit, line 149: Nice spec 👌🏼

db/views/reporting_overdue_patients_v01.sql
```diff
@@ -0,0 +1,89 @@
...
+    ON reporting_patient_states.patient_id = next_call_results.patient_id
+      AND next_call_results.device_created_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')) >= reporting_patient_states.month_date
+      AND next_call_results.device_created_at AT TIME ZONE 'UTC' AT TIME ZONE (SELECT current_setting('TIMEZONE')) < reporting_patient_states.month_date + interval '1 month'
+WHERE
+  reporting_patient_states.status <> 'dead'
+ORDER BY
+  reporting_patient_states.month_date,
+  reporting_patient_states.patient_id,
+  appointments.device_created_at DESC,
+  visits.visited_at_after_appointment,
```
prabhanshuguptagit, line 87: Just confirming, we want the earliest `visited_at_after_appointment`?

db/views/reporting_overdue_patients_v01.sql
```diff
@@ -0,0 +1,89 @@
...
+  next_call_results.device_created_at,
+  previous_call_results.device_created_at DESC
```
prabhanshuguptagit, line 89: How many months of data do we want to show on the dashboard? Do you think we can restrict this query to only pick, say the last 24 months?

prabhanshuguptagit (approved):
View looks sick. Nice specs and benchmarks also. :shipit:

db/views/reporting_overdue_patients_v01.sql
danySam: On the dashboard, we need only 6 months of data. But we need it to be available for as long as possible for folks to download the line list. I'll check with @jamiecarter7 if that's the case.

db/views/reporting_overdue_patients_v01.sql
danySam: Yes, that's correct. The immediate visit after an appointment

db/views/reporting_overdue_patients_v01.sql
```diff
@@ -0,0 +1,89 @@
...
+  reporting_patient_states.assigned_facility_region_id AS assigned_facility_region_id,
+  appointments.id AS previous_appointment_id,
+  appointments.device_created_at AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' AS previous_appointment_date,
+  appointments.scheduled_date AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' AS previous_appointment_scheduled_date,
+  visits.visited_at_after_appointment AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' AS visited_at_after_appointment,
+  next_call_results.device_created_at AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' AS next_called_at,
+  previous_call_results.device_created_at AT TIME ZONE 'UTC' AT TIME ZONE 'UTC' AS previous_called_at,
+  reporting_patient_states.hypertension AS hypertension,
+  reporting_patient_states.diabetes AS diabetes,
+  next_call_results.result_type AS next_call_result_type,
```
danySam, line 14: @PriyangaPKini Reminder to add user id of the calling user

db/views/reporting_overdue_patients_v01.sql
danySam: Interesting, we'll check this out separately. But I suspect the query planner will do this optimisation.

PR #5079, opened 2023-04-21
Fix some papercuts in the setup script

tfidfwastaken:
**Story card:** 🤷 

## Because

I tried to check this codebase out today and run the local setup but faced a few issues while trying to do so. These should hopefully fix them :)

## This addresses

* The bash setup script had several logical errors stemming from incorrect usage of the `[` command, as well as an incorrect invocation of `nvm`. These have been fixed.

* The script also exited with a failure status because the database wasn't cleaned up before the parallel tests ran.

* A dependency of rails, `nio4r` failed to compile on installations having clang 16+. This has been patched upstream, so we bump up the version to make our setup script work.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -34,9 +34,6 @@ blocks:
           value: "test"
         - name: TB_RSPEC_OPTIONS
           value: "--profile --format RspecJunitFormatter --out junit.xml"
-      secrets:
-        - name: sentry-release-auth-token
-        - name: semaphore-deploy-key
       jobs:
         - name: StandardRB
           commands:
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -377,7 +377,7 @@ GEM
       net-ssh (>= 2.6.5, < 8.0.0)
     net-ssh (7.0.1)
     netrc (0.11.0)
-    nio4r (2.5.8)
+    nio4r (2.5.9)
     nokogiri (1.13.8)
       mini_portile2 (~> 2.8.0)
       racc (~> 1.4)
--- a/bin/setup
+++ b/bin/setup
@@ -11,7 +11,9 @@ init (){
   verify_ruby_version && verify_node_version
   verify_bundled_gems && verify_yarn_packages
   verify_foreman
-  verify_postgres && verify_database_migrations
+  verify_postgres && verify_database_migrations &&
+  cleanup_database
+
   configure_parallel_testing
 }
 
@@ -23,18 +25,15 @@ is_darwin (){
 
 has_homebrew (){
   echo 'Checking for homebrew'
-  [ is_darwin ] && [ hash brew >/dev/null 2>&1 ]
+  is_darwin && hash brew >/dev/null 2>&1
 }
 
 verify_homebrew (){
   echo 'Verifying Homebrew installed'
-  if [ ! has_homebrew ]
+  if ! has_homebrew
   then
     echo 'Homebrew is not installed'
-    if [ is_darwin ]
-    then
-      echo 'Please install homebrew. https://brew.sh'
-    fi
+    is_darwin && echo 'Please install homebrew. https://brew.sh'
     exit 1
   fi
 }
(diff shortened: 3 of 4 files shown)


bin/setup
```diff
@@ -23,18 +25,15 @@ is_darwin (){
 
 has_homebrew (){
   echo 'Checking for homebrew'
-  [ is_darwin ] && [ hash brew >/dev/null 2>&1 ]
+  is_darwin && hash brew >/dev/null 2>&1
```
tfidfwastaken, line 28: * `[ darwin ]` is equivalent to `test "darwin"` (`[` is an alias to `test`) which always evaluates to true, because "darwin" is a non-empty string.
* `[ hash ... ]` is passing multiple arguments to `[` which it does not take.

Fixed these here and elsewhere in the file to match the intention of what this script was trying to do.

bin/setup
```diff
@@ -144,13 +148,14 @@ migrate_existing_database (){
 }
 
 setup_database (){
-  bundle exec rails db:setup
+  bundle exec rails db:setup ||
+  echo 'Unable to create the databases for you.
+Please ensure your database.yml is configured for your system and try again.'
+}
```
tfidfwastaken, line 154: Just a small refactor, a successful `echo` will always return zero.

bin/setup
```diff
@@ -11,7 +11,9 @@ init (){
   verify_ruby_version && verify_node_version
   verify_bundled_gems && verify_yarn_packages
   verify_foreman
-  verify_postgres && verify_database_migrations
+  verify_postgres && verify_database_migrations &&
+  cleanup_database
```
tfidfwastaken, line 15: Not sure if this is the right thing to do here, but the script before this change was causing an error that looked like:
```
psql:/Users/atharva/simple-server/db/structure.sql:48: ERROR:  type "gender_enum" already exists
rake aborted!
failed to execute:
psql --set ON_ERROR_STOP=1 --quiet --no-psqlrc --file /Users/atharva/simple-server/db/structure.sql simple-server_test6

Please check the output above for any errors and make sure that `psql` is installed in your PATH and has proper permissions.
```

This was because the `structure.sql` file was being run twice. I assume a cleanup was the intended step here, before running the parallel tasks.

bin/setup
```diff
@@ -144,13 +148,14 @@ migrate_existing_database (){
...
+  bundle exec rails db:setup ||
+  echo 'Unable to create the databases for you.
+Please ensure your database.yml is configured for your system and try again.'
+}
 
-  if [ $? -ne 0 ]
-  then
-    echo 'Unable to create the databases for you. Please ensure your database.yml is configured for your system and try again.'
-    exit 0
-  fi
```
tfidfwastaken, line 153: Just a small refactor for readability. Also, a successful `echo` will always return zero.

bin/setup
```diff
@@ -117,8 +113,16 @@ has_node_version (){
 }
 
 install_nodejs (){
+  is_darwin || {
+    echo "Please install node for your platform."
+    exit 1
+  }
+
```
tfidfwastaken, line 120: Added this bit, so that it does not attempt to `brew install nvm` on Linux machines.

Gemfile.lock
```diff
@@ -377,7 +377,7 @@ GEM
       net-ssh (>= 2.6.5, < 8.0.0)
     net-ssh (7.0.1)
     netrc (0.11.0)
-    nio4r (2.5.8)
+    nio4r (2.5.9)
```
govindkrjoshi, line 380: Why is this version bump required?

tfidfwastaken: I answered this in the PR description:
> A dependency of rails, nio4r failed to compile on installations having clang 16+. This has been patched upstream, so we bump up the version to make our setup script work.

To elaborate a bit more, the bundle step failed, because it resolves this dependency and tries to compile it, but fails. This was the issue: https://github.com/socketry/nio4r/issues/290
...and this was the patch: https://github.com/socketry/nio4r/commit/dd69a1c053bf13a580ec6a4e392c15c269239fd6

govindkrjoshi (approved):

.semaphore/semaphore.yml
```diff
@@ -34,9 +34,6 @@ blocks:
           value: "test"
         - name: TB_RSPEC_OPTIONS
           value: "--profile --format RspecJunitFormatter --out junit.xml"
-      secrets:
```
PriyangaPKini, line 37: Could please explain the reason why you deleted these 'secrets' here?

tfidfwastaken: @prabhanshuguptagit and I were pairing to resolve CI issues blocking this merge, and he noticed that these secrets were not used in this job, so we removed it. Forgot to update the description to mention this. 

In an ideal world, this would go in a separate PR, but this also gave us an excuse to retrigger the job as well (:

tfidfwastaken: Learnt that this broke something. I'll raise a new PR without this change. Sorry folks.

PR #5094, opened 2023-05-05
Deploy manifest files to github pages

roypeter:
**Story card:** [10334](https://app.shortcut.com/simpledotorg/story/10334)

## Because

We have a Rails application deployed for multiple countries, each with its own instance. The application relies on a manifest.json file to display a list of countries. Currently, this file is served from the India deployment, creating a dependency on that specific instance. In order to improve reliability and maintainability, we need to decouple the manifest.json file from the India deployment and host it on a separate deployment.

## This addresses

Deploy manifest to GitHub pages

### DIFF
--- a/.github/workflows/deploy-manifest.yml
+++ b/.github/workflows/deploy-manifest.yml
@@ -1,0 +1,41 @@
+# Workflow for deploying public/manifest content to GitHub Pages
+name: Deploy public/manifest content to Pages
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["master"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-manifest.yml
```diff
@@ -0,0 +1,41 @@
...
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
```
prabhanshuguptagit, line 28: Where does this URL come from?

prabhanshuguptagit (approved):

.github/workflows/deploy-manifest.yml
roypeter: page_url is from https://github.com/simpledotorg/simple-server/settings/pages

PR #5099, opened 2023-05-11
Goal-lines - Minimum 6 months period data [bug correction]

jamiecarter7:
I was checking the length of an object instead of an array.
- Convert object to array of key names
- Check length to ensure minimum of 6 months before calculating goal

## This addresses

Dashboard throws and error on facilities with less than 6 months data present

## Test instructions

Requires 'goal_lines' flipper flag
View facility with less than 6 months of data.

### DIFF
--- a/app/assets/javascripts/common/reports.js
+++ b/app/assets/javascripts/common/reports.js
@@ -452,7 +452,8 @@ Reports = function ({ withLtfu, showGoalLines }) {
   }
 
   function lessThan6MonthsSinceFirstPatientRegistered(periodValues) {
-    if (periodValues.length < 6) {
+    const periodKeysArray = Object.keys(periodValues)
+    if (periodKeysArray.length < 6) {
       return true;
     }
     return false;


app/assets/javascripts/common/reports.js
```diff
@@ -471,22 +473,21 @@ Reports = function ({ withLtfu, showGoalLines }) {
   function goalPeriodValue(periodValues) {
     const dateKeysArray = Object.keys(periodValues);
     const decemberKeys = dateKeysArray.filter((item) => item.includes("Dec"));
-    
-    const indexOfLatestDecember = dateKeysArray.indexOf(
-      decemberKeys[decemberKeys.length - 1]
-    );
+    const mostRecentDecemberKey = decemberKeys[decemberKeys.length - 1];
+    const indexOfMostRecentDecemberInPeriodValues = dateKeysArray.indexOf(mostRecentDecemberKey);
```
prabhanshuguptagit, line 477: The `InPeriodValues` bit is implied, no? `indexOfLatestDecember` seems like a good name by itself :D

app/assets/javascripts/common/reports.js
```diff
@@ -451,8 +452,9 @@ Reports = function ({ withLtfu, showGoalLines }) {
     return mergeConfig(config, goalLineConfig);
   }
 
-  function lessThan6MonthsSinceFirstPatientRegistered(periodValues) {
-    if (periodValues.length < 6) {
+  function isLessThanDefaultMonthsSinceFirstPatientRegistered(periodValues) {
```
prabhanshuguptagit, line 455: Could make a function like so and use that, to save the super long name
```suggestion
  function monthsSinceFirstRegistration(periodValues) {    
      const periodKeysArray = Object.keys(periodValues);
      periodKeysArray.length
  }  
```

prabhanshuguptagit (approved):

app/assets/javascripts/common/reports.js
```diff
@@ -451,8 +452,9 @@ Reports = function ({ withLtfu, showGoalLines }) {
...
+    const periodKeysArray = Object.keys(periodValues);
+    if (periodKeysArray.length < defaultMonthsRequired) {
       return true;
     }
     return false;
```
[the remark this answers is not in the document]
jamiecarter7: Good point!

app/assets/javascripts/common/reports.js
jamiecarter7: Reverted to 'indexOfLatestDecember'

app/assets/javascripts/common/reports.js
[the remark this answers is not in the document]
jamiecarter7: Done

app/assets/javascripts/common/reports.js
jamiecarter7: I like this suggestion as its easy to read.

I did the following

```
if (monthsSinceFirstRegistration(periodValues) < defaultMonthsRequired) {
...
}
```
and I used the new function you suggested to return monthsSinceFirstRegistration

PR #5100, opened 2023-05-12
Use semantic-versioning like for DSL Version upgrades

olttwa:
**Story card:** [sc-10324](https://app.shortcut.com/simpledotorg/story/10324/extend-questionnaires-to-support-supplies-reports)

## Because

- When a new DSL version is introduced, server needs to modify sync logic to send all forms until the DSL version supported by Mobile.
- When there are multiple questionnaires of the same type for different DSL versions, Server will de-duplicate them by choosing the latest one.

## Test instructions

Check specs in `questionnaires_controller_spec.rb`


app/controllers/api/v4/questionnaires_controller.rb
```diff
@@ -10,9 +10,22 @@ def transform_to_response(questionnaire)
   def current_facility_records
     # TODO: Current implementation always responds with 1 JSON minimum. Reason:
     # process_token.last_updated_at has precision upto 3 milliseconds & is always lesser than updated_at.
```
olttwa, line 12: Logic of DSL Version semvering:
- DSLs will upgrade in following manner: 1.0, 1.1, 1.2, 1.3
- Backwards incompatible DSLs will be upgraded as 2.0
- Mobile supporting DSL version 1.0 cannot be sent 1.1 as it cannot compile them while parsing response & will crash. _(This has been confirmed with Sid)_
- When Mobile requests Questionnaires for 1.2, server sends all between 1.0 and 1.2
- When there are more than 1 Questionnaires of _**same type**_ between 1.0 and 1.2, server picks the one with higher DSL Version.
- The INNER JOIN query below seems to be the most efficient & code-clean way to make such a query. More details in [this SO answer](https://stackoverflow.com/questions/48269660/select-distinct-values-from-a-column-based-on-highest-value-in-another-column).
- If any better solution, do suggest.

db/migrate/20230512070357_update_questionnaires.rb
```diff
@@ -0,0 +1,11 @@
+class UpdateQuestionnaires < ActiveRecord::Migration[6.1]
+  def up
+    add_column :questionnaires, :description, :string
+    change_column :questionnaires, :dsl_version, :decimal, precision: 4, scale: 2
```
olttwa, line 4: We can change column from integer to decimal without losing data or affecting existing code. Won't lead to any crashes, we've tested locally.

app/controllers/api/v4/questionnaires_controller.rb
olttwa: > Most importantly, we are doing `X.Y` instead of `X.Y.Z` because we can leverage the decimal data type to do a greater-than/less-than comparison, and questionnaires won't really have patch upgrades. A new questionnaire of same DSL version will be picked because that will be `active` and the rest will be inactive.

PR #5104, opened 2023-05-15
Set up DSL versioning

qptr:
**Story card:** [sc-10324](https://app.shortcut.com/simpledotorg/story/10324/extend-questionnaires-to-support-supplies-reports)

## Because

- When a new DSL version is introduced, server needs to modify sync logic to send all forms until the DSL version supported by Mobile.
- When there are multiple questionnaires of the same type for different DSL versions, Server will de-duplicate them by choosing the latest one.

## Test instructions

Check specs in `questionnaires_controller_spec.rb`


app/controllers/api/v4/questionnaires_controller.rb
```diff
@@ -10,9 +10,23 @@ def transform_to_response(questionnaire)
   def current_facility_records
     # TODO: Current implementation always responds with 1 JSON minimum. Reason:
     # process_token.last_updated_at has precision upto 3 milliseconds & is always lesser than updated_at.
```
olttwa, line 12: Logic of DSL Version semvering:
- DSLs will upgrade in following manner: 1.0, 1.1, 1.2, 1.3
- Backwards incompatible DSLs will be upgraded as 2.0
- We are doing `X.Y` instead of `X.Y.Z` because we can leverage the decimal data type to do a greater-than/less-than comparison, and questionnaires won't really have patch upgrades. A new questionnaire of same DSL version will be picked because that will be `active` and the rest will be inactive.
- Mobile supporting DSL version 1.0 cannot be sent 1.1 as it cannot compile them while parsing response & will crash. _(This has been confirmed with Sid)_
- When Mobile requests Questionnaires for 1.2, server sends all between 1.0 and 1.2
- When there are more than 1 Questionnaires of _**same type**_ between 1.0 and 1.2, server picks the one with higher DSL Version.
- The INNER JOIN query below seems to be the most efficient & code-clean way to make such a query. More details in [this SO answer](https://stackoverflow.com/questions/48269660/select-distinct-values-from-a-column-based-on-highest-value-in-another-column).
- If any better solution, do suggest.

db/migrate/20230512070357_update_questionnaires.rb
```diff
@@ -0,0 +1,11 @@
+class UpdateQuestionnaires < ActiveRecord::Migration[6.1]
+  def up
+    add_column :questionnaires, :description, :string
+    change_column :questionnaires, :dsl_version, :decimal, precision: 4, scale: 2
```
olttwa, line 4: We can change column from integer to decimal without losing data or affecting existing code. Won't lead to any crashes, we've tested locally.

app/controllers/api/v4/questionnaires_controller.rb
```diff
@@ -10,9 +10,23 @@ def transform_to_response(questionnaire)
...
+    dsl_version = params.require("dsl_version").to_f.truncate(2)
+
+    # For dsl_version 1.5, return all questionnaires from 1 to 1.5.
+    # De-duplicate multiple questionnaires of same type by choosing latest dsl_version.
     Questionnaire
       .for_sync
-      .where(dsl_version: params.require("dsl_version").to_i)
+      .select("*")
```
prabhanshuguptagit, line 19: I think this select might be redundant.

app/controllers/api/v4/questionnaires_controller.rb
```diff
@@ -10,9 +10,23 @@ def transform_to_response(questionnaire)
...
+      .joins("
+        INNER JOIN (
```
prabhanshuguptagit, line 21: Can this be written with a `DISTINCT ON` in postgres? Something like
```
select distinct on (questionnaire_type)
from questionnaires
where dsl_version between #{dsl_version.floor} AND #{dsl_version}
order by dsl_version

```

lib/seed/questionnaire_seeder.rb
```diff
@@ -7,6 +7,7 @@ def self.call
       FactoryBot.create(:questionnaire,
         questionnaire_type: "monthly_screening_reports",
         is_active: true,
+        description: "screening_reports, specimen-seed",
```
prabhanshuguptagit, line 10: Should we just use a human readable description?

spec/controllers/api/v4/questionnaires_controller_spec.rb
```diff
@@ -134,15 +134,30 @@ def create_record_list(n, options = {})
       end
     end
 
-    it "returns questionnaires only for given DSL Version" do
-      version_1_questionnaire = create(:questionnaire, :active, questionnaire_type: "monthly_screening_reports", dsl_version: 1)
-      version_2_questionnaire = create(:questionnaire, :active, questionnaire_type: "monthly_screening_reports", dsl_version: 2)
+    it "returns questionnaires from floor to current dsl_version" do
```
prabhanshuguptagit, line 137: with the `same major version`?

prabhanshuguptagit (changes requested):

spec/controllers/api/v4/questionnaires_controller_spec.rb
olttwa: done.

app/controllers/api/v4/questionnaires_controller.rb
olttwa: The flexibility provided by using a string isn't required by us. `X.Y` will serve the purpose

olttwa: Switching from decimal to numeric, because we discovered the SQL query stays the same, and no need for trim/split_part.

app/controllers/api/v4/questionnaires_controller.rb
olttwa: Modified. Wow, distinct on really removed the complexity.

PR #5113, opened 2023-05-18
Ensure a `deleted_reason` for deduplicated patients

tfidfwastaken:
**Story card:** [sc-8392](https://app.shortcut.com/simpledotorg/story/8392/set-deleted-reason-for-deduplicated-patients)

## Because

The `deleted_reason` is currently left empty for patients who are de-duplicated and merged automatically. It should be set to `duplicate`. We also ensure that the `deleted_reason` is preserved across patient sync merges.

## This addresses

When we delete a patient during a merge, we must provide a valid deletion reason. This change ensures that the reason is:
- "duplicate" whenever we de-duplicate a patient 
    - occurrences: `SoftDeleteDuplicatePatients` and `PatientDeduplicator` module
- the already-present `deleted_reason` when a sync merge happens
    - occurrences: `MergePatientService`

## Test instructions

* Unit tests, etc.
* Mark a patient as duplicate from the deduplication page in the dashboard and verify if the deletion reason is set to "duplicate" in the database.

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -16,6 +16,8 @@ class Patient < ApplicationRecord
 
   enum status: STATUSES.zip(STATUSES).to_h, _prefix: true
 
+  enum deleted_reason: DELETED_REASONS.zip(DELETED_REASONS).to_h, _prefix: true
+
   enum reminder_consent: {
     granted: "granted",
     denied: "denied"
@@ -224,9 +226,9 @@ class Patient < ApplicationRecord
      gender: gender}
   end
 
-  def discard_data(deleted_reason: "unknown")
+  def discard_data(reason:)
     ActiveRecord::Base.transaction do
-      update(deleted_reason: deleted_reason)
+      update(deleted_reason: reason)
       address&.discard
       appointments.discard_all
       blood_pressures.discard_all
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -224,8 +224,9 @@ class Patient < ApplicationRecord
      gender: gender}
   end
 
-  def discard_data
+  def discard_data(deleted_reason: "unknown")
     ActiveRecord::Base.transaction do
+      update(deleted_reason: deleted_reason)
       address&.discard
       appointments.discard_all
       blood_pressures.discard_all
@@ -239,5 +240,7 @@ class Patient < ApplicationRecord
       teleconsultations.discard_all
       discard
     end
+
+    self
   end
 end
--- a/app/services/merge_patient_service.rb
+++ b/app/services/merge_patient_service.rb
@@ -5,7 +5,6 @@ class MergePatientService
   end
 
   def merge
-    patient_before_merge = existing_patient
     merged_address = Address.merge(payload[:address]) if payload[:address].present?
 
     patient_attributes =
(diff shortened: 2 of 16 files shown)


app/models/patient.rb
```diff
@@ -224,8 +224,9 @@ def anonymized_data
      gender: gender}
   end
 
-  def discard_data
+  def discard_data(deleted_reason: "unknown")
```
tfidfwastaken, line 227: I'm adding a default `deleted_reason` to an explicit `unknown`, after seeing the possible values in `DELETED_REASONS`. Before this, the default was unintentionally `nil`. Grepping the code, it shouldn't affect any reporting query, but let me know if you prefer default should remain `nil`.

I have also ignored migrating the existing `nil` value columns to `unknown`s for now.

spec/models/patient_spec.rb
```diff
@@ -739,13 +751,21 @@
...
       expect(Teleconsultation.where(patient: patient)).to be_empty
     end
+
+    it "soft deletes the patient itself" do
+      patient = create(:patient)
+
+      patient.discard_data
+
+      expect(patient.deleted_at).not_to be_nil
+    end
```
tfidfwastaken, line 769: Not directly related to the change I made, but I felt we should test this too.

app/models/patient.rb
prabhanshuguptagit, line 227: Could call `deleted_reason` just `reason`.

app/services/merge_patient_service.rb
```diff
@@ -23,7 +23,9 @@ def merge
     touch_patient(merged_patient) if associations_updated?(merged_address, merged_phone_numbers, merged_business_ids)
 
     # Patient has been soft-deleted by the client, server should soft-delete the patient and their associated data.
-    discard_patient_data(merged_patient) if discarded_in_this_merge?(merged_patient, patient_before_merge)
+    if discarded_in_this_merge?(merged_patient, patient_before_merge)
+      discard_patient_data(merged_patient, deleted_reason: merged_patient.deleted_reason)
```
prabhanshuguptagit, line 27: Can move the `merged_patient.deleted_reason` inside `discard_patient_data`.

spec/services/merge_patient_service_spec.rb
```diff
@@ -51,15 +51,43 @@
...
+
+        expect_any_instance_of(Patient).not_to receive(:discard_data)
+        described_class.new(payload, request_metadata: metadata).merge
+      end
+
+      it "preserves the existing deleted_reason after merge" do
+        patient = create(:patient, deleted_reason: "duplicate")
+        now = Time.current
+        # set the updated_at so that it is treated as an update
+        patient_attributes = build_patient_payload(patient).merge(updated_at: now, deleted_at: now)
```
prabhanshuguptagit, line 83: Should set `deleted_reason` to something different from `duplicate`.

Can also check if the mock is passed the right thing.

app/models/patient.rb
prabhanshuguptagit: Let's make this a required parameter so a user of this class will need to supply a reason and we don't have to overload `unknown`.

app/services/patient_deduplication/deduplicator.rb
```diff
@@ -54,7 +54,9 @@ def mark_as_merged(new_patient)
       @patients.each do |patient|
         track(Patient, patient, new_patient)
       end
-      @patients.map(&:discard_data)
+      @patients.map do |p|
+        p.discard_data(deleted_reason: "duplicate")
```
prabhanshuguptagit, line 58: Prefer using `:duplicate` over string quotes.

app/services/merge_patient_service.rb
prabhanshuguptagit: Let's make `deleted_reason` an [enum](https://api.rubyonrails.org/v5.1.7/classes/ActiveRecord/Enum.html) on `Patient`.

prabhanshuguptagit (changes requested):
👌🏼 Feel free to ship after addressing the comments.

tfidfwastaken:
@prabhanshuguptagit I've noticed that the `merge` method of the `Mergeable` concern already queries the existing record to set the merge status (ie, doing the same thing as the `existing_patient` private method in `MergePatientService`). I've done a light refactor to take advantage of this and reduce duplication.

tfidfwastaken:
@prabhanshuguptagit I've noticed that the `merge` method of the `Mergeable` concern already queries the existing record to set the merge status (ie, doing the same thing as the `existing_patient` private method in `MergePatientService`). I've done a light refactor to take advantage of this and reduce duplication.

(Thanks to @qptr for providing inspiration that led to this refactor)

prabhanshuguptagit (approved):
Good refactor.

PR #5130, opened 2023-05-22
fix foreman Procfile `clock` for non-GNU developers

tfidfwastaken:
`sleep infinity` is a GNU extension to `sleep`; so this command breaks for MacOS users, which in turn causes all the processes started by foreman to exit. We change this to a POSIX compliant version that loops once every 2^24 seconds, ie, 5ish years (this is much below the POSIX-required maximum of 2147483647 seconds [~68 years], but I found this number less intimidating).


harimohanraj89 (approved):
Yeah 5 years is juuust fine 👌

PR #5137, opened 2023-05-26
Goal lines update may 2023

jamiecarter7:
**Story card:** 
[sc-10614](https://app.shortcut.com/simpledotorg/story/10614/add-control-for-visibility-in-region-type)
[sc-10615](https://app.shortcut.com/simpledotorg/story/10615/add-below-and-above-text-to-goal-string)
[sc-10613](https://app.shortcut.com/simpledotorg/story/10613/corrections-to-date-selected-if-current-month-is-dec-then-choose-previous-month)
[sc-10618](https://app.shortcut.com/simpledotorg/story/10618/change-design-to-include-goal-zone-background-colour)

## Because

'Region' level specific visibility
Design improvements
Algorithm fixes

### 'Region' level specific visibility

Adds the ability to control whether the goal line is shown based on the region level [see docs](Regions).
The function checks the name of the current regionType against a list of enabled region names. Where name is duplicated across countries the country initials have been added for enabling in individual countries.

### Design improvements

1. Added additional text 'Above' or 'Below' inside the text string (part of the bubble stating the goal). For improved clarity.
2. 'Goal zones' - changing the fill from the chart to the area below/above the goal line to highlight where you should aim to be.

### Algorithm fixes

Fixing a missed check where if the last month in the array of dates is a December [...., 'Nov-2023', 'Dec-2023'] then ignore this value. 
If the last value is a December the date picker should select a previous December (2022) or the 6th month to calculate the goal. Only in Jan-2024 we should use Dec-2023.

## Test instructions

Goal lines wont show at all with less than 6 months of data

### Test 1: 'Region' level specific visibility.

To see the goal lines you will have to add 'state' to **enabledRegions** object inside reports.js file.
To check at facility level you will need to add 'facilityIN' as localhost pretends to be IN (India) then navigate to a facility.
To enable at each level can be done in dev env using the following:
- 'state'
- 'districtIN'
- 'block'
- 'facilityIN'
I've followed this table to build specificity - [simple docs](https://docs.simple.org/reports/what-we-report#region-reports)

The 'organization' is defined in the config but actually falls behind the flipper flag, ensuring it only shows where the flag is enabled inside a country. Also not all organizations have the ability to view data at this level, BD does show org level dashboard, but dev env does not.

Note: goal lines will only show if at least 6 months of data are present.

### Test 2: Check 'Above' or 'Below' text present.

Above: 
- BP Controlled
Below:
- BP Not Controlled
- Missed Visits

Screenshot 2023-05-30 at 14 09 20

### Test 3: Goal zones
This can be tested by adding and removing 'state' or at facility level using 'facilityIN' in the enabledRegions object.

When a goal line is active and calculated then the following changes occur:
- Area below/above goal line becomes filled
- Area below the chart line is no longer filled

Screenshot 2023-05-30 at 13 57 27

When a goal line is no active (not shown) then the chart should remain unchanged:
- The chart line should be filled as expected.

Screenshot 2023-05-30 at 13 57 42

### Test 4: Algorithm fixes

**The images look slightly different here but the focus is the goal line calculation**

When viewing the dashboard at any level with a goal change the view to dec-2022 and the goal lines calculations should update to be calculated from the dec-2021 value (or month 6 if dec is part of the first 6 months).

Current calculation is 10% of the remaining improvement, note the calculation is slightly different for upwards and downwards goals.
Eg upwards: date value (40%)
100 - 40 = 60
60 * 0.1 = 6
Goal = 46% 
Upwards goals are rounded up, even 45.1%

Eg downwards: date value 40%
40 * 0.1 = 4
Goal = 36%
Downwards goals are rounded down, even 36.9%

Chose Dec-2022 and check the goal is calculated from Dec-2021 value.
Note: if the there is only 1 December value in the data, and that Decemeber value is one of the first 5 months the goal is calculated from the 6th month. 
Screenshot 2023-05-26 at 14 29 19

For example this facility shows a goal calculated using Apr-2023
Screenshot 2023-05-26 at 14 28 59


govindkrjoshi (approved):

app/assets/javascripts/common/reports.js
```diff
@@ -543,18 +577,61 @@ Reports = function ({ withLtfu, showGoalLines }) {
...
+    ctx.closePath();
+
+    const rgbaChartColor = chart.config.data.datasets[0].borderColor;
+    const fillColor = changeRGBAColorOpacity(rgbaChartColor, 0.065);
+    const textColor = changeRGBAColorOpacity(rgbaChartColor, 0.4);
+
+    ctx.fillStyle = fillColor || "rgba(0, 0, 0, 0.1)";
+    ctx.fill();
+
+    // canvasDrawGoalZoneText(chart)
```
govindkrjoshi, line 615: We can remove this commented code.

jamiecarter7: Good spot!

Removed

PR #5153, opened 2023-06-02
Fix translation locale selection

olttwa:
**Story card:** Bug-fix

This PR does 3 things:
1. Prunes `config.i18n.available_locales` to only contain keys which are in the translation backends.
2. Move from `compatible_language_from` to `language_region_compatible_from` method of http_accept_language gem, which is more correct.
    1. Simple server was ignoring region part of translation in favour of first occurring primary language, which is incorrect. (For ex, `bn-IN` got inferred as `bn-BD` because that was defined first in config)
3. Change all underscores (en_BD) to hyphens (en-BD), because hyphen is the convention.

## Previous PR cause a production issue because:

1. `config.i18n.available_locales` contained a lot of locales that were not present in translation files as backend.
3. `en-IN` was defaulting to `en-IN` by `language_region_compatible_from` method, because `config.i18n.available_locales` contained `en_IN` key, and a backend for this wasn't present.
    1. This was causing an error: `I18n::InvalidLocale: "en-IN" is not a valid locale`
4. This wasn't an issue before this PR, because previously we were using this method: `compatible_language_from`, which inferred `en-IN` as `en`.


harimohanraj89:
Hmm, the en_*** locales are definitely used on the _dashboard_ to determine country-specific field names like districts, blocks, states, etc. Can we double-check/confirm this before removing them from the supported locale list?

config/application.rb
```diff
@@ -43,7 +43,7 @@ class Application < Rails::Application
 
     # Locale configuration
     config.i18n.load_path += Dir[Rails.root.join("config", "locales", "**", "*.{rb,yml}")]
-    config.i18n.available_locales = %w[en mr-IN pa-Guru-IN bn-BD kn-IN en_BD en_IN en_ET en_LK en-IND bn-IN hi-IN ta-IN te-IN am-ET om-ET si-LK sid-ET so-ET ta-LK ti-ET es]
+    config.i18n.available_locales = %w[am-ET bn-BD bn-IN en es hi-IN kn-IN mr-IN om-ET pa-Guru-IN si-LK sid-ET so-ET ta-IN ta-LK te-IN ti-ET]
```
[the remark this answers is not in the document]
govindkrjoshi: We should not remove `en_IN` from this list. `en` and `en_IN` are two different locales and have different files in config/locales as well.

config/initializers/country_config.rb
```diff
@@ -23,7 +23,7 @@ class CountryConfig
       states: COUNTRYWISE_STATES["Bangladesh"],
       dashboard_locale: ENV["DEFAULT_PREFERRED_DASHBOARD_LOCALE"] || "en_BD",
       time_zone: ENV["DEFAULT_TIME_ZONE"] || "Asia/Dhaka",
-      faker_locale: "en-IND",
+      faker_locale: "en",
```
govindkrjoshi, line 26: Shouldn't the faker_locale be the same as dashboard_locale?

spec/requests/shared_examples/api/v4/shared_spec_for_sync_requests.rb
```diff
@@ -158,4 +158,32 @@ def assert_sync_success(response, request_process_token)
...
+    it "sets locale correctly" do
+      expect(I18n).to receive(:with_locale).with("bn-IN")
+
+      get sync_route, params: {}, headers: headers.merge({"Accept-Language" => "bn-IN"})
+    end
+
+    it "defaults to first primary language when it cannot find region" do
+      expect(I18n).to receive(:with_locale).with("en")
+
+      get sync_route, params: {}, headers: headers.merge({"Accept-Language" => "en-IN"})
```
govindkrjoshi, line 172: `en-IN` should match to itself, and not `en`

spec/requests/shared_examples/api/v4/shared_spec_for_sync_requests.rb
```diff
@@ -158,4 +158,32 @@ def assert_sync_success(response, request_process_token)
       expect(JSON(response.body)[response_key].count).to eq(1)
     end
   end
+
+  context "translations are correctly inferred" do
```
govindkrjoshi, line 162: These tests are coupled to our application config currently, it would be nice to decouple the unit test from the application config. One option could be to mock the application config for these tests.

config/locales/faker/en-IN.yml
```diff
@@ -1,4 +1,4 @@
-en-IND:
```
olttwa, line 1: changed `en-IND` to `en-IN`, because  we don't have `-IND` region.

spec/requests/shared_examples/api/v4/shared_spec_for_sync_requests.rb
olttwa: Mocking application config would defeat the purpose of these tests, which IMO, can only be done well using integration tests.

olttwa:
> Hmm, the en_*** locales are definitely used on the _dashboard_ to determine country-specific field names like districts, blocks, states, etc. Can we double-check/confirm this before removing them from the supported locale list?

@harimohanraj89 , my bad. I found usage of the `en_***` locales & have kept them in the application config.

spec/requests/shared_examples/api/v4/shared_spec_for_sync_requests.rb
olttwa: updated test.

config/initializers/country_config.rb
olttwa: Tests use faker_locale to populate sync data. If we were to keep it same as dashboard locale, tests will fail as they cannot find this data.

govindkrjoshi (approved):

config/locales/api/en_IN.yml
```diff
@@ -0,0 +1,5 @@
+# A place-holder file for en-IN locale.
+# All translations for en-IN default to :en
+# Value of `en-IN` needs to be non-empty for rails to pick it.
+en-IN:
+  place: "holder"
```
govindkrjoshi, line 5: Is the `place` key required? If not, let's remove it. 

Do we not need this empty locale file for `en-BD` etc, since the mobile app doesn't send these locales?

olttwa: We need at-least 1 key-value pair for rails to ACK the locale. Else, it doesn't include it in the I18n.backends.available_locales, and the API will begin failing again.

PR #5158, opened 2023-06-08
Debug rspec test time

roypeter:

## Because

## This addresses

## Test instructions

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
   around_action :switch_locale
 
   def switch_locale(&action)
-    locale = http_accept_language.language_region_compatible_from(I18n.available_locales) || I18n.default_locale
+    locale = http_accept_language.compatible_language_from(I18n.available_locales) || I18n.default_locale
     I18n.with_locale(locale, &action)
   end
 
--- a/spec/api/v4/questionnaires_spec.rb
+++ b/spec/api/v4/questionnaires_spec.rb
@@ -16,7 +16,7 @@ describe "Questionnaires v4 API", swagger_doc: "v4/swagger.json" do
         parameter param
       end
 
-      let("Accept-Language") { "en-IND" }
+      let("Accept-Language") { "en-IN" }
       let(:dsl_version) { "1.1" }
 
       response "200", "Questionnaires Synced to user device" do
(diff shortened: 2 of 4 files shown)


config/schema_descriptions.yml
```diff
@@ -92,24 +92,6 @@ reporting_facility_states:
     diabetes_appts_scheduled_15_to_31_days: "The total number of appointments scheduled by healthcare workers for diabetic patients at the facility in the reporting month between 15 and 31 days from the visit date"
     diabetes_appts_scheduled_32_to_62_days: "The total number of appointments scheduled by healthcare workers for diabetic patients at the facility in the reporting month between 32 and 62 days from the visit date"
     diabetes_appts_scheduled_more_than_62_days: "The total number of appointments scheduled by healthcare workers for diabetic patients at the facility in the reporting month more than 62 days from the visit date"
-    overdue_patients: "The total of overdue patients at the facility at the beginning of the reporting month. Dead and lost to follow-up patients are excluded."
```
olttwa, line 95: Why are we deleting these descriptions now?

db/migrate/20230523084623_update_reporting_facility_states_to_version11.rb
```diff
@@ -1,8 +0,0 @@
-class UpdateReportingFacilityStatesToVersion11 < ActiveRecord::Migration[6.1]
```
olttwa, line 1: We shouldn't delete a migration. If the intention is to revert to version 10, I'll suggest creating version 12.

Deleting a migration requires a manual rollback not only on all production environments, but individual developer machines as well.

spec/factories/appointments.rb
```diff
@@ -15,8 +15,7 @@
     user
 
     trait :overdue do
-      device_created_at { rand(61..90).days.ago }
-      scheduled_date { rand(35..60).days.ago }
+      scheduled_date { rand(30..90).days.ago }
```
olttwa, line 18: Switch from 35 to 30 was intentional, right?

spec/models/reports/facility_state_spec.rb
```diff
@@ -616,639 +612,4 @@
       expect(rows.pluck(:facility_id)).not_to include(facility_2.id)
     end
   end
-
-  context "monthly hypertension overdue patients", run_with_custom_date: true do
```
olttwa, line 620: All these tests got deleted because we reverted from v11 to v10?

olttwa (changes requested):

PR #5161, opened 2023-06-09
Dashboard overdue UI alterations

jamiecarter7:
**Story card:** [sc-10735](https://app.shortcut.com/simpledotorg/story/10735/fix-z-index-layering-in-overdue-patients-call-chart)

## Because

UI fixes for overdue dashboard:
- Add missing divider to table
- Fix missing toggle styling
- Some changes to sentences (title/subtitles)
- Fix z-index issues on chart line graph

### Fix missing toggle styling
Before:
image
After:
image

### Add missing divider to table

There is a missing column divider in the table.

Before:
image

After:
image

### Some changes to sentences (title/subtitles)
Minor changes to title and subtitle wording

### Fix z-index issues on chart line graph
There is an issue with some charts having the layers that make the background below the line being stacked over the top of the line nodes when they are close.
Screenshot 2023-06-09 at 12 54 33
The fix requires a change in how the chart draws the elements.
Previously the line segment of the chart was attached to the 'remove from overdue list' but the way stacking is done in chartjs means I need to split this into 2 elements a line and the red band.
Screenshot 2023-06-09 at 16 06 32
'order' sets the draw order (z-index) and 'stack' groups the stacks together. We want to stack the call results and then have the line in a separate stack otherwise they are all stacked on top of one another.


app/components/dashboard/hypertension/overdue_patients_called_table_component.html.erb
```diff
@@ -19,6 +19,7 @@
           <col class="table-divider">
           <col>
           <col class="table-divider">
+        <col>
```
danySam, line 22:
```suggestion
          <col>
```

danySam (approved):
There's a minor indentation error, otherwise looks good

jamiecarter7:
Good spot, committed suggestion

PR #5163, opened 2023-06-12
Refresh only required views to reduce the refresh time

PriyangaPKini:
**Story card:** [sc-10688](https://app.shortcut.com/simpledotorg/story/10688/fix-ci-pipeline-slowness-after-introducing-new-tests)

## Because

Some tests on overdue patient fields in the `facility_state_spec` were slow.  I found that most of the time is taken to refresh the mat views.

## This addresses

Refreshing only the required mat views to reduce the time

### DIFF
--- a/spec/models/reports/facility_state_spec.rb
+++ b/spec/models/reports/facility_state_spec.rb
@@ -617,7 +617,7 @@ RSpec.describe Reports::FacilityState, {type: :model, reporting_spec: true} do
     end
   end
 
-  xcontext "monthly hypertension overdue patients", run_with_custom_date: true do
+  context "monthly hypertension overdue patients", run_with_custom_date: true do
     describe "overdue_patients" do
       it "should return number of overdue patients assigned to the facility at beginning of a month" do
         month_date = reporting_dates[:beginning_of_month]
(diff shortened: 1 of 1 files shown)


spec/models/reports/facility_state_spec.rb
```diff
@@ -617,15 +617,23 @@
...
+  context "monthly hypertension overdue patients", run_with_custom_date: true do
     describe "overdue_patients" do
       it "should return number of overdue patients assigned to the facility at beginning of a month" do
         month_date = reporting_dates[:beginning_of_month]
         facility = create(:facility)
         _overdue_patients = create_list(:patient, 2, :with_overdue_appointments, assigned_facility: facility, registration_facility: facility)
         _patient = create(:patient, :with_appointments, appointment_creation_date: month_date - 1, scheduled_date: month_date + 15.days, assigned_facility: facility, registration_facility: facility)
 
-        RefreshReportingViews.refresh_v2
+        views = %w[
```
olttwa, line 628: Can we move views to a let block to avoid duplication?

PriyangaPKini: yeah, make sense

PR #5166, opened 2023-06-13
fix is_overdue calculation

PriyangaPKini:
**Story card:** [sc-10684](https://app.shortcut.com/simpledotorg/story/10684/fix-is-overdue-column-calculation-in-overdue-patient-s-matview)

## Because

`is_overdue` returns 'yes' when a patient has no `previous_appointment`

## This addresses

`is_overdue` returns `no` when a patient has no `previous_appointment`


db/views/reporting_overdue_patients_v02.sql
```diff
@@ -0,0 +1,230 @@
...
+    visited_at_after_appointment,
+    called_by_user_id,
+    next_called_at,
+    previous_called_at,
+    next_call_result_type,
+    next_call_removed_from_overdue_list_reason,
+    previous_call_result_type,
+    previous_call_removed_from_overdue_list_reason,
+    CASE
+        WHEN previous_appointment_id IS NULL THEN 'no'
```
PriyangaPKini, line 191: > WHEN previous_appointment_id IS NULL THEN 'no'

This is the only change in the file

db/structure.sql
```diff
@@ -3914,6 +3914,7 @@ CREATE MATERIALIZED VIEW public.reporting_overdue_patients AS
     patient_with_call_results_and_phone.previous_call_result_type,
     patient_with_call_results_and_phone.previous_call_removed_from_overdue_list_reason,
         CASE
+            WHEN (patient_with_call_results_and_phone.previous_appointment_id IS NULL) THEN 'no'::text
```
olttwa, line 3917: migration file is adding a bunch of indexes, but they aren't showing in structure.sql. Is this correct behaviour?

danySam: Those migrations already exist on the `reporting_overdue_patients` table. Hence don't need to add them again.

PriyangaPKini: > Those indices already exist on the `reporting_overdue_patients` table. Hence don't need to add them again.

`reporting_facility_states` *

db/structure.sql
```diff
@@ -7430,6 +7431,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20230512070357'),
 ('20230522081701'),
 ('20230523084623'),
+('20230613081116'),
```
danySam, line 7434: This needs to go after the latest migration. Update the filename also

PR #5167, opened 2023-06-13
Add filters for notification experiments

olttwa:
**Story card:** [sc-10617](https://app.shortcut.com/simpledotorg/story/10617/setup-sms-experimentation-in-bd-june-1-2023)

## Because

TODO


olttwa:
Sample query looks like the following, only last 3 lines have changed
```sql
SELECT DISTINCT block_region.id AS block_region_id, facilities.*, patients.*
FROM "patients"
         INNER JOIN "medical_histories"
                    ON "medical_histories"."deleted_at" IS NULL AND "medical_histories"."patient_id" = "patients"."id"
         INNER JOIN "patient_phone_numbers" ON "patient_phone_numbers"."deleted_at" IS NULL AND
                                               "patient_phone_numbers"."patient_id" = "patients"."id"
         INNER JOIN "facilities"
                    ON "facilities"."deleted_at" IS NULL AND "facilities"."id" = "patients"."assigned_facility_id"
         INNER JOIN regions facility_regions ON facility_regions.source_id = facilities.id
         INNER JOIN regions block_region
                    ON block_region.path @ > facility_regions.path AND block_region.region_type = 'block'
WHERE "patients"."deleted_at" IS NULL
  AND "medical_histories"."deleted_at" IS NULL
  AND "medical_histories"."hypertension" = 'yes'
  AND "patients"."reminder_consent" = 'granted'
  AND "patients"."status" != 'dead'
  AND "patient_phone_numbers"."phone_type" = 'mobile'
  AND (EXTRACT(YEAR FROM COALESCE(age('2023-06-14', date_of_birth),make_interval(years => age) + age('2023-06-14', age_updated_at))) >= 18)
  AND (NOT EXISTS (SELECT "treatment_group_memberships"."patient_id" FROM "treatment_group_memberships" INNER JOIN "treatment_groups" ON "treatment_groups"."id" = "treatment_group_memberships"."treatment_group_id" INNER JOIN "experiments" ON "experiments"."deleted_at" IS NULL AND "experiments"."id" = "treatment_groups"."experiment_id" WHERE "treatment_group_memberships"."deleted_at" IS NULL AND (treatment_group_memberships.patient_id = patients.id) AND (end_time > '2023-06-13 18:59:25.410055' OR treatment_group_memberships.created_at > '2023-05-29 18:59:25.410095')))
  AND "facilities"."deleted_at" IS NULL
  AND "facilities"."state" IN ('state-1', 'state-2')
  AND "block_region"."id" NOT IN ('block-1', 'block-2')
  AND ("patients"."assigned_facility_id" NOT IN ('facility-1','facility-2'))
```

app/models/experimentation/notifications_experiment.rb
```diff
@@ -42,34 +42,59 @@ def self.eligible_patients
                                              "end_time > ? OR treatment_group_memberships.created_at > ?",
                                              Time.current, MONITORING_BUFFER.ago
                                            ).select(:patient_id))
-        .then { |patients| exclude_bangladesh_blocks(patients) }
```
olttwa, line 45: I've deleted config-based filtering of India & BD environments. Post deployment, I'll update the active experiments in India's database with filters.
Bangladesh does not have any active experiment at the moment.

app/models/experimentation/notifications_experiment.rb
```diff
@@ -42,34 +42,59 @@ def self.eligible_patients
...
+        patients
+      end
+    end
+
+    def self.filter_facilities(patients, facilities)
+      return patients unless facilities
+
+      # require 'pry'; binding.pry
+      if facilities["include"]
+        patients.where('"patients"."assigned_facility_id" IN (?)', facilities["include"])
```
olttwa, line 87: I had to write a manual `where` query due to a bug that was putting assigned facilities as null.
If someone can tell how to make this query work, or why this works for states & blocks but not for assigned_facility_id, that'd help:
```ruby
patients.where(assigned_facility_id: facilities["include"])
```

db/structure.sql
```diff
@@ -9,13 +9,6 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
---
--- Name: public; Type: SCHEMA; Schema: -; Owner: -
---
-
```
olttwa, line 15: @PriyangaPKini this deletion has your git comment. Is it safe to delete?

app/models/experimentation/experiment.rb
```diff
@@ -12,6 +12,7 @@ class Experiment < ApplicationRecord
     validates :experiment_type, presence: true
     validate :validate_date_range
     validate :one_active_experiment_per_type
+    validate :validate_filters
```
olttwa, line 15: I'll add unit tests for this before merging. Have tested a bit manually though.

app/models/experimentation/notifications_experiment.rb
```diff
@@ -21,15 +21,15 @@ def notifying?
...
+        enrolling.each { |experiment| experiment.enroll_patients(date, experiment.filters) }
         monitoring.each { |experiment| experiment.monitor }
         notifying.each { |experiment| experiment.schedule_notifications(date) }
       end
     end
 
     # Returns patients who are eligible for enrollment. These should be
     # filtered further by individual notification experiments based on their criteria.
-    def self.eligible_patients
+    def self.eligible_patients(filters = {})
```
olttwa, line 32: Unit tests are failing. Will remove older cases, and test newer cases before merging.

app/models/experimentation/notifications_experiment.rb
```diff
@@ -42,34 +42,57 @@ def self.eligible_patients
...
-    def self.include_only_allowed_india_states(patients)
-      return patients unless CountryConfig.current_country?("India") && SimpleServer.env.production?
-      return patients unless ENV["EXPERIMENT_ALLOWED_STATES"].present?
+    def self.filter_states(patients, states)
+      return patients unless states
 
-      states = ENV.fetch("EXPERIMENT_ALLOWED_STATES", "").split(",").map(&:strip)
-      patients.where(facilities: {state: states})
+      if states["include"]
+        patients.where(facilities: {state: states["include"]})
```
olttwa, line 61: I'm sticking to names of state instead of UUID as mentioned in regions/reporting_patient_states table because extracting state_id becomes too complicated with the existing`with_block_region_id` scope.

olttwa: State name is used in current design, so I am reusing that logic in new design.

db/structure.sql
danySam: It's just a comment. It's fine

db/structure.sql
```diff
@@ -4219,748 +4213,6 @@ CREATE MATERIALIZED VIEW public.reporting_facility_states AS
...
 
---
--- Name: MATERIALIZED VIEW reporting_facility_states; Type: COMMENT; Schema: public; Owner: -
---
-
-COMMENT ON MATERIALIZED VIEW public.reporting_facility_states IS 'Monthly summary of a facility''s indicators. This table has one row per facility, per month, from the month of the facility''s first registration.';
-
-
---
--- Name: COLUMN reporting_facility_states.month_date; Type: COMMENT; Schema: public; Owner: -
```
danySam, line 4230: All these comments removed in this file are necessary yeah? Maybe out of date with the migrations on master?

app/models/experimentation/notifications_experiment.rb
danySam: `patients.where(assigned_facility_id: facilities["include"])` should work on its own. I'm guessing this has to do with some lazy eval and filtering each other out

app/models/experimentation/notifications_experiment.rb
danySam: Sounds good. Between deployment and updating the filter, there's no chance of an experiment running no?

app/models/experimentation/notifications_experiment.rb
danySam: Actually, I think the `.merge` is changing the behaviour of `patients`

danySam (commented):
Looks good mostly. Please ensure `structure.sql` is up to date with master. 

Not a blocker, but do you think we can convert the input filter hash to a hash that `ActiveRecord.where` accepts? That way we need to apply the `.where` condition only once and the code would look much more cleaner.

app/models/experimentation/notifications_experiment.rb
olttwa: The enrollment happens at 5.45 AM, hence no.

app/models/experimentation/notifications_experiment.rb
```diff
@@ -42,34 +42,57 @@ def self.eligible_patients
...
+        patients
+      end
+    end
+
+    def self.filter_facilities(patients, facilities)
+      return patients unless facilities
+
+      if facilities["include"]
+        patients.where('"patients"."assigned_facility_id" IN (?)', facilities["include"])
+      elsif facilities["exclude"]
```
danySam, line 86: qq: Can you not have facilities that you want to keep but some facilities that you want to exclude?

danySam:
I think we may have to rename `includes` to `keep` and `excludes` to 'remove`. That may reflect the underlying meaning better

db/structure.sql
PriyangaPKini: We should run `db:refresh_matviews` after running `db:migrate` to add comments to the table columns. 
Can you please run `db:refresh_matviews and add back those comments?

app/models/experimentation/experiment.rb
PriyangaPKini: Can you please add the unit tests as well? It would help to understand the logic better

olttwa:
> I think we may have to rename `includes` to `keep` and `excludes` to `remove`. That may reflect the underlying meaning better

`keep` won't convey the meaning I think. `include` is meant as a selection only & `exclude` as a blocklist/ban/complement.

I'll prefer a pair of words. Like include/exclude, allow/disallow instead of keep/remove.

app/models/experimentation/experiment.rb
olttwa: Just pushed unit tests for validation filter. Will push tests for eligible_patients in a while.

app/models/experimentation/notifications_experiment.rb
olttwa: Have added unit tests for new design.

db/structure.sql
olttwa: I've added back those comments by running db:migrate.

olttwa:
> Please ensure structure.sql is up to date with master

@danySam this is done.

PR #5175, opened 2023-06-15
add additional note to table tooltip

jamiecarter7:

## Because

Add the 'note' text from the patients called chart tooltip to the patients called table tooltip

## This addresses

A mismatch in the explanation
The figures shown are identical in calculation so this should read the same in the tooltip

## Test instructions

Hover the tooltip ? in the overdue table element
You should see this:
Screenshot 2023-06-15 at 13 23 23


jamiecarter7:
Please merge this once its reviewed - so its live ASAP

PR #5178, opened 2023-06-19
Fix for scrolling to top when using toggle

jamiecarter7:
**Story card:** [sc-10766](https://app.shortcut.com/simpledotorg/story/10766/preserve-scroll-position-after-clicking-the-non-contactable-toggle)

## Because

Clicking the toggle causes a page reload and the user is navigated to the top of the page losing their current scroll position.

The user then has to scroll back down to the overdue section.

## This addresses

This corrects the problem and keeps the user in the position they were previously viewing by scrolling to the same position after the pageload

## Test instructions

Scroll down to the overdue section and click the 'non-contactable patients' toggle.
Your page should refresh and the scroll position preserved.
Check both activating and deactivating the toggle.


jamiecarter7:
> You might be able to achieve this without using JS by using an id tag on the overdue section and adding an action to the form to go "#overdue" whenever it's submitted.
> 
> ```
> +  <div id="overdue" class="d-flex jc-space-between">
>      <h4 class="mt-5 mb-32px">Overdue</h4>
> 
>      <div class="d-flex" style="margin-top: 43px;">
> -      <form method="get" name="overdue_form">
> +      <form method="get" action="#overdue" name="overdue_form">
>          <div class="custom-control custom-switch checkbox-outside-card">
> ```

The only downside is that this lingers in the url - if the user later refreshes when browsing other data then they will be taken back to the overdue section.

danySam:
> The only downside is that this lingers in the url - if the user later refreshes when browsing other data then they will be taken back to the overdue section and might be confused why?

That's true. I guess if we go down that path, we'll need to have an `id` tag for each section which needs to be updated as scrolled. I guess the approach you have works.

app/views/reports/regions/show.html.erb
```diff
@@ -324,4 +324,16 @@
       countryAbbreviation: '<%= CountryConfig.current[:abbreviation] %>'
     }).initialize();
   });
+  // Overdue toggle plage refresh scroll
```
PriyangaPKini, line 327: [nit-pick] There is a typo in the comment

app/views/reports/regions/show.html.erb
```diff
@@ -324,4 +324,16 @@
...
+  window.addEventListener('DOMContentLoaded', (event) => {
+    const overdueToggleElement = document.getElementById('overdue-section');
```
PriyangaPKini, line 329: [non-blocking] Can we add `addEventListeners` to 'ltfu toggle' as well to preserve the scroll position?

PriyangaPKini (approved):

app/views/reports/regions/show.html.erb
jamiecarter7: See more recent comment

app/assets/javascripts/common/toggleScrollToPosition.js
```diff
@@ -0,0 +1,31 @@
+// Check for scroll position and scroll
+window.addEventListener("DOMContentLoaded", (event) => {
+  if (doesYPositionExistInLocalStorage()) {
+    scrollToYPosition();
+    removeYPositionLocalStorage();
+  }
+});
```
jamiecarter7, line 7: There is a check to see if a scroll position exists and if so scroll to that position and remove the position from localStorage

app/assets/javascripts/common/toggleScrollToPosition.js
```diff
@@ -0,0 +1,31 @@
...
+
+// Create toggle change listener and add y position on 'change' event
+function createToggleListener(id) {
+  const toggleElement = document.getElementById(id);
+  toggleElement.addEventListener("change", (event) => {
+    setYPositionLocalStorage();
+  });
+}
```
jamiecarter7, line 15: I've re-written the whole thing to be a generic function that creates a listener for each toggle. All that is required is the toggle ID value

app/components/dashboard/diabetes/bs_over200_graph_component.html.erb
```diff
@@ -6,7 +6,7 @@
       <%= title.tooltip({ "Blood sugar 200-299 numerator" => t("bs_over_200_copy.bs_200_to_299.numerator"),
                           "Blood sugar ≥300 numerator" => t("bs_over_200_copy.bs_over_300.numerator"),
                           "Denominator" => t(denominator_copy, region_name: @region.name) }) %>
-      <%= title.ltfu_toggle(id: 'controlledGraphLtfuToggle', enabled: with_ltfu) %>
+      <%= title.ltfu_toggle(id: 'notControlledGraphLtfuToggle', enabled: with_ltfu) %>
```
jamiecarter7, line 9: I've had to rename the diabetes toggles, when multiple things have the same ID only the first has the effect applied.

I cant see any other references to the ID in the code - hopefully this will not effect any of the data?

app/views/reports/regions/diabetes.html.erb
```diff
@@ -67,5 +67,12 @@
...
-
 <%= render "reports/regions/diabetes_footnotes" %>
+<script>
+  window.addEventListener("DOMContentLoaded", () => {
+    const checkboxes = document.querySelectorAll('input[type="checkbox"]');
+    checkboxes.forEach(element => {
+      createToggleListener(element.id)
+    });
+  })
+</script>
```
jamiecarter7, line 78: Get all checkboxes and create a listener using the function

app/assets/javascripts/common/toggleScrollToPosition.js
```diff
@@ -0,0 +1,31 @@
+// Check for scroll position and scroll
+window.addEventListener("DOMContentLoaded", (event) => {
+  if (doesYPositionExistInLocalStorage()) {
+    scrollToYPosition();
+    removeYPositionLocalStorage();
+  }
```
tfidfwastaken, line 6: Consider inlining these functions.

Reasoning:
* These functions aren't reused anywhere else yet.
* They aren't abstracting away implementation details in a helpful way, IMO. Especially if they are just called for side effects, I'd like to see what the side effect is without having to look for its definition.
* They are short one-liners, so easy to scan even when inlined.

app/assets/javascripts/common/toggleScrollToPosition.js
tfidfwastaken: [suggestion, non-blocking] Perhaps this could be renamed to something like `createScrollTrackingListener`. Essentially, something that would convey what event we are interested in listening to.

app/assets/javascripts/common/toggleScrollToPosition.js
```diff
@@ -0,0 +1,31 @@
...
+// Create toggle change listener and add y position on 'change' event
+function createToggleListener(id) {
+  const toggleElement = document.getElementById(id);
+  toggleElement.addEventListener("change", (event) => {
+    setYPositionLocalStorage();
+  });
+}
+
+function setYPositionLocalStorage() {
+  localStorage.setItem("toggleScrollToPosition", window.pageYOffset);
```
tfidfwastaken, line 18: Can the key of the `localStorage` item be renamed to something more precise, like `scrollPosition`? That clearly conveys what we are storing. `toggleScrollToPosition` is a verb, so it makes it seem like we are storing an action of some kind.

app/components/dashboard/diabetes/bs_over200_graph_component.html.erb
tfidfwastaken: This looks like the right thing to do. The ID should definitely be different. It doesn't look like this is referenced elsewhere. Paging @PriyangaPKini to check again, since I'm not that familiar with this part of the code.

[nit] The BP not controlled graph has the ID `uncontrolledGraphLtfuToggle`, perhaps this could have the same name for symmetry.

app/components/dashboard/diabetes/missed_visits_graph_component.html.erb
```diff
@@ -5,7 +5,7 @@
       subtitle: t("diabetes_missed_visits_copy.reports_card_subtitle", region_name: @region.name)) do |title| %>
       <%= title.tooltip({ "Numerator" => t("diabetes_missed_visits_copy.numerator"),
                           "Denominator" => t("diabetes_missed_visits_copy.denominator", region_name: @region.name) }) %>
-      <%= title.ltfu_toggle(id: 'controlledGraphLtfuToggle', enabled: with_ltfu) %>
+      <%= title.ltfu_toggle(id: 'missedVisitsLtfuToggle', enabled: with_ltfu) %>
```
tfidfwastaken, line 8: Nice.

tfidfwastaken (commented):
I have suggested a few naming changes and inlining some functions that I think were extracted out a bit early.

Other than that, solid work! Using the toggles feels much more satisfying now :D

app/assets/javascripts/common/toggleScrollToPosition.js
jamiecarter7: Done

app/assets/javascripts/common/toggleScrollToPosition.js
jamiecarter7: changed to `yScrollPosition`

tfidfwastaken:
Looks good to me @jamiecarter7 👍

PR #5183, opened 2023-06-20
Fix drug category and coefficient of BD states

PriyangaPKini:
**Story card:** [sc-10673](https://app.shortcut.com/simpledotorg/story/10673/drug-stock-in-bd-showing-error-in-arb-and-diuretic-drugs)

## Because

Drug stock in BD shows errors in ARB and Diuretic drugs. The drugs were under the wrong category, which caused the error.

## This addresses

Fixes the drug category and adds missing coefficient


dburka (approved):
LGTM

PR #5184, opened 2023-06-21
Introduce Overdue Patients Returned Card on Dashboard

PriyangaPKini:
**Story card:** [sc-10784](https://app.shortcut.com/simpledotorg/story/10784/integrate-data-with-return-to-care-graph)

## Because

We want to see the impact of the calls made to overdue patients by measuring the number of patients who return to care after receiving the call.

## This addresses

This card introduces a new graph, which shows the percentage of overdue patients who received a call and returned to care. 

Additionally, it also shows a breakdown of the returned to care number by the call result (ie: 'agreed to visit', 'remind to call later' or 'removed from overdue list').

## Test instructions

Enable the feature flag (`overdue_patients_returned`) for your user or a user group to see this graph on this HTN dashboard.


app/assets/stylesheets/application.scss
```diff
@@ -748,3 +748,27 @@ span.inactive {
 }
 
 .o-65 { opacity: 0.65; }
+
+.card-chip {
```
danySam, line 752: @jamiecarter7 the CSS for these are the same for overdue patients called cards. Are these duplicates?

app/components/dashboard/hypertension/overdue_patients_return_to_care_component.rb
```diff
@@ -0,0 +1,52 @@
...
+      patientsReturnedRemindToCallLaterRates: data[:patients_returned_with_result_remind_to_call_later_rates],
+      patientsReturnedRemovedFromOverdueListRates: data[:patients_returned_with_result_removed_from_list_rates],
+      **period_data
+    }
+  end
+
+  def periods
+    start_period = @period.advance(months: -17)
+    Range.new(start_period, @period)
+  end
```
danySam, line 38: This method is not used. Let's delete

app/views/reports/regions/show.html.erb
```diff
@@ -274,19 +274,28 @@
...
-
-  <% if current_admin.feature_enabled?(:medications_dispensation) %>
-    <h4 class="mt-5 mb-32px">Medications</h4>
-    <%= render Dashboard::MedicationDispensationComponent.new(data: @data, region: @region, period: @period) %>
-  <% end %>
-
+<% if current_admin.feature_enabled?(:medications_dispensation) %>
+  <h4 class="mt-5 mb-32px">Medications</h4>
+  <%= render Dashboard::MedicationDispensationComponent.new(data: @data, region: @region, period: @period) %>
+<% end %>
```
danySam, line 294: Let's revert this change

app/assets/stylesheets/application.scss
jamiecarter7: They aren't - I realise that I left the css as styles in the 'patients called' card. 

This is my attempt to convert to a class. I can retro fit this into the other card.

danySam: Oh, in that case, could you use these classes to replace the inline CSS on the called cards as well?

spec/factories/appointments.rb
```diff
@@ -15,8 +15,8 @@
     user
 
     trait :overdue do
-      device_created_at { rand(61..90).days.ago }
-      scheduled_date { rand(35..60).days.ago }
+      device_created_at { rand(81..90).days.ago }
```
olttwa, line 18: Did the time range of an overdue appointment change?

db/structure.sql
```diff
@@ -9,6 +9,13 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
+--
```
olttwa, line 12: this seems like an incorrect addition. Please re-concile structure.sql with master

app/components/dashboard/hypertension/overdue_patients_return_to_care_component.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def initialize(region:, data:, period:, with_removed_from_overdue_list:)
+    @region = region
+    @data = data
+    @period = period
+    @contactable = !with_removed_from_overdue_list
+  end
+
+  def graph_data
+    if @contactable
```
olttwa, line 12: For readability, can be changed to 
```ruby
if @contactable
  {}
else
  {}
end
```

olttwa: also, we're using/not-using `@` interchangeably. Either refer to contactable & data with @ or without.

app/views/reports/regions/show.html.erb
danySam: Not reverting this. Since this fixes the indendation

app/views/reports/regions/show.html.erb
```diff
@@ -274,19 +274,30 @@
...
       current_admin: @current_admin,
       with_non_contactable: @with_non_contactable)) %>
+    <% if current_admin.feature_enabled?(:overdue_patients_returned) %>
+      <%= render(Dashboard::Hypertension::OverduePatientsReturnToCareComponent.new(
+        region: @region,
+        data: @data,
+        period: @period,
+        with_removed_from_overdue_list: @with_removed_from_overdue_list)) %>
+      <div class="d-lg-flex w-lg-50 pl-lg-2 hidden" style="margin-bottom: 16px;">
+        <div class="" style="display: flex; justify-content: center; align-items: center; border: dashed 2px #CECECE; width: 50%; border-radius: 4px; width: 100%;">
```
danySam, line 284: Let's move these inline styles to a class

spec/factories/appointments.rb
danySam: No, this was something we used to temporarily fix a test. Reverting this.

qptr (commented):
[Non-blocker] Just repeating what I said in the channel, that `remind to call` and `removed from list` should be under `unanswered` while `agreed to visit` is a stand-in for `answered`.

If remind to call doesn't tell us if a patient answered or not, maybe the app should be capturing this information.

app/models/reports/region_summary.rb
```diff
@@ -70,6 +70,14 @@ def self.call(regions, range: nil)
       contactable_patients_called_with_result_agreed_to_visit
       contactable_patients_called_with_result_remind_to_call_later
       contactable_patients_called_with_result_removed_from_list
+      patients_returned_after_call
```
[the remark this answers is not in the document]
qptr: are there patients_returned_after_call that don't have 1 phone number 🤔

config/locales/en.yml
```diff
@@ -175,6 +175,10 @@ en:
       ‘Remove from overdue list‘ reason must be one of following:
     list: "Patient not responding, Phone number does not work, Transferred to another public health facility, Moved to a private practitioner, Other reason"
 
+  overdue_patients_return_to_care:
+    reports_card_subtitle: "Overdue hypertension patients that returned to care within 15 days of receiving their first call"
+    numerator: "Patients that returned to care within 15 days of their first call."
+    denominator: "Hypertension patients assigned to %{region_name} with at least one ‘Result of call...’ marked during the month. Deleted, dead and lost to follow-up are excluded."
```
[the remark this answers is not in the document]
PriyangaPKini: Yeah, it had to be a little verbose because there are several key points that we wanna mention here like `hypertension`, `assigned`, `at least one ‘Result of call...’` , `during the month`

app/assets/stylesheets/application.scss
jamiecarter7: yes, once this is merged to prod I can make those css class changes to the called card 👍

PR #5191, opened 2023-06-28
Overdue patients maintenance tasks

danySam:
**Story card:** [sc-10245](https://app.shortcut.com/simpledotorg/story/10245/remove-old-call-data-from-report-tables)

## Because

This includes maintenance tasks for the Overdue Patients Epic.

## This addresses

- [x] Remove the old call data tables from the dashboard
- [x] Refactor `cap_call_result_rates` method in OverduePatientsCalledComponent
- [x] Refactor `proportional_call_rate` method in OverduePatientsCalledComponent


app/models/reports/region_summary_schema.rb
```diff
@@ -325,45 +325,57 @@ def initialize(regions, periods:)
       end
     end
 
+    memoize def patients_called_breakdown_rates(entry)
```
danySam, line 328:
```suggestion
    memoize def patients_called_breakdown_rates(slug, period)
```

The `entry` var's meaning is lost once it's outside the loop it's called in.

app/models/reports/region_summary_schema.rb
```diff
@@ -325,45 +325,57 @@ def initialize(regions, periods:)
...
+      slug, period = entry.slug, entry.period
+      rounded_percentages({
+        agreed_to_visit: patients_called_with_result_agreed_to_visit[slug][period],
+        remind_to_call_later: patients_called_with_result_remind_to_call_later[slug][period],
+        removed_from_overdue_list: patients_called_with_result_removed_from_list[slug][period]
+      })
+    end
+
+    memoize def contactable_patients_called_breakdown_rates(entry)
```
danySam, line 337: You need to add tests for these methods

app/models/reports/region_summary_schema.rb
PriyangaPKini: Right, will make the change

PriyangaPKini: it's done ✅

PR #5193, opened 2023-06-28
Fix alpha sms detail class

olttwa:
**Story card:** bug

## Because
Notifications aren't being sent out in Bangladesh environment because of a `NoMethodError` error in the "monitoring" step. Details:
1. Everyday at 5.45 AM BD time,  `Experimentation::Runner.call` rake task runs
1. `conduct_daily` performs enrolling, monitoring & notifying steps sequentially
1. Enrolling step is working fine as there are treatment group memberships created until today
1. In the monitoring step, `successful_delivery.detailable.result` is failing with NoMethodError
1. Hence, the rake task never gets to Notifying step.

**Solution**: `BsnlDeliveryDetail` does have a `.result` attribute, but `AlphaSmsDeliveryDetail` has `.request_status` instead. We have added a `result` method to AlphaSms class which returns value of `request_status`

### DIFF
--- a/app/models/alpha_sms_delivery_detail.rb
+++ b/app/models/alpha_sms_delivery_detail.rb
@@ -17,6 +17,10 @@ class AlphaSmsDeliveryDetail < DeliveryDetail
     request_status.blank?
   end
 
+  def result
+    request_status
+  end
+
   def self.create_with_communication!(request_id:, recipient_number:, message:)
     ActiveRecord::Base.transaction do
       delivery_detail = create!(
--- a/db/data/20230628132154_setup_july_v2_experiments_bangladesh.rb
+++ b/db/data/20230628132154_setup_july_v2_experiments_bangladesh.rb
@@ -1,0 +1,85 @@
+class SetupJulyV2ExperimentsBangladesh < ActiveRecord::Migration[6.1]
+  START_TIME = DateTime.new(2023, 7, 1).beginning_of_day
+  END_TIME = DateTime.new(2023, 8, 1).beginning_of_day
+  CURRENT_PATIENTS_EXPERIMENT = "Current Patient July V2 2023"
+  STALE_PATIENTS_EXPERIMENT = "Stale Patient July V2 2023"
+  PATIENTS_PER_DAY = 5000
+  FILTERS = {
+    "states" => {"include" => ["Sylhet"]},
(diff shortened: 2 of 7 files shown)


danySam (approved):

app/models/alpha_sms_delivery_detail.rb
```diff
@@ -17,6 +17,10 @@ def in_progress?
     request_status.blank?
   end
 
+  def result
+    request_status
+  end
```
govindkrjoshi, line 22:
```suggestion
alias result request_status
```

db/data/20230628132154_setup_july_v2_experiments_bangladesh.rb
```diff
@@ -0,0 +1,85 @@
...
+        cascade3 = experiment.treatment_groups.create!(description: "cascade3")
+        cascade3.reminder_templates.create!(message: "notifications.set03.basic", remind_on_in_days: 0)
+        cascade3.reminder_templates.create!(message: "notifications.set03.basic", remind_on_in_days: 3)
+        cascade3.reminder_templates.create!(message: "notifications.set03.basic", remind_on_in_days: 7)
+      end
+    end
+  end
+
+  def down
+    Experimentation::Experiment.current_patients.find_by_name(CURRENT_PATIENTS_EXPERIMENT).cancel
```
govindkrjoshi, line 82: Down migration is not the inverse of the up migration.

PriyangaPKini:
> 

Looks like the PR description is incomplete

db/data/20230628132154_setup_july_v2_experiments_bangladesh.rb
olttwa: I'm assuming that by saying * Down migration is not the inverse of the up migration*, you meant we aren't undoing cancel of 2 experiments, right? Did you mean to say something else?

Down migration will cancel (soft-delete) the created experiments. We cannot (shouldn't?) undo the soft-deletion of previous experiments because that will lead to unintended consequences.

app/models/alpha_sms_delivery_detail.rb
olttwa: Adding an alias won't work because `request_status` is not a static attribute or a method, it's added at runtime. The code fails with error: `NameError: undefined method `request_status' for class `#<Class:0x0000000148e7eef8>'`

govindkrjoshi (approved):

PR #5197, opened 2023-07-03
Overdue dashboard: CSS class corrections

jamiecarter7:
**Story card:** [sc-10836](https://app.shortcut.com/simpledotorg/story/10836/refactor-move-all-the-inline-css-to-classes)

## Because

Some styles left in the HTML code rather than CSS classes

## This addresses

Fixes to change to CSS classes

## Test instructions

1. Intentionally blank card (next to 'returned to care' chart card)
This should be hidden when looking at the site on a mobile device - when the number of cards in a row drops to 1. (View with a skinny screen width).
2. Check the chips (the colored bubbles with the % call outcomes and return to care).
Ensure that they work at all screen widths, they should not have more than a single line of text and where the bubble is too small to display the text then ... appears.
The 'Remove from overdue list' is cropped initially to improve reading - hovering the chip shows the full text.
Screenshot 2023-06-28 at 14 22 06


app/assets/stylesheets/application.scss
```diff
@@ -749,6 +749,13 @@ span.inactive {
 
 .o-65 { opacity: 0.65; }
 
+.card-chip-container {
```
danySam, line 752: You can move this into the '.card-chip' class and use a [parent-selector](https://sass-lang.com/documentation/style-rules/parent-selector/) to organize this

It will look something like this
```
.card-chip {
  ...
  &.container {
    ....
  }
}
```

app/assets/stylesheets/application.scss
```diff
@@ -759,16 +766,39 @@ span.inactive {
...
   text-align: center;
   text-overflow: ellipsis;
+
+  &.card-chip-green {
+    background: rgba(78, 206, 0, 0.15);
+    color: rgba(44, 115, 0, 1);
+  }
+
+  &.card-chip-amber {
+    background: rgba(252, 247, 193, 0.8);
```
danySam, line 776: It'd be nice if we can use variables for the colours. It'll become easier to refactor later if the definitions are in one place.

app/components/dashboard/hypertension/overdue_patients_called_component.html.erb
```diff
@@ -1,48 +1,48 @@
...
-          <div>
-            <p class="m-0px c-black">
-              <span data-key="overduePatientsCalled" data-format="numberWithCommas"></span>
-              patients called in
-              <span data-key="endDate"></span>
-            </p>
+    <%end%>
+    <%= c.summary do %>
+      <div class="mb-12px d-lg-flex align-lg-center">
+        <p class="c-print-black graph-percent fs-28px" data-key="overduePatientsCalledRate" data-format="percentage"  style="color: #e0b000;"></p>
```
danySam, line 11: There's a style element here, we can move that to the class too

danySam (approved):

PR #5205, opened 2023-07-10
Goal lines: 3 month average

jamiecarter7:
**Story card:** [sc-10869](https://app.shortcut.com/simpledotorg/story/10869/change-goal-lines-to-3-month-average)

## Because

There can be large differences in performance from month to month.
This can mean higher or lower than expected targets if there is a large swing during the goal month.

## This addresses

Using a 3 month average taking the 'goal month' and using the previous 2 months to make an average helps to smooth this swinging difference.

For example: If 'dec-2022' is the goal month, then we use the average of oct, nov and dec (2022) to create an average value.

## Test instructions

On your dashboard use the following formula to calculate your goal:
For BP controlled:
[average of oct, nov and dec (2022)] - (100 - [avg oct nov dec]) * 0.1
0.1 because its 10% increase per year (it is different if a goal is calculated from a month different to dec)

For BP uncontrolled and Missed visit:
[average of oct, nov and dec (2022)] - [avg oct nov dec] * 0.1


tfidfwastaken (approved):
Looks good to me 🚢

PR #5208, opened 2023-07-11
Add API docs and mock server for import API

tfidfwastaken:
We add an import API in Simple behind a feature flag. It shall currently be a mock server that only runs the validation on the request and return a success or failure. We also generate swagger documentation for this API.

**Story card:** [sc-10849](https://app.shortcut.com/simpledotorg/story/10849) and [sc-10850](https://app.shortcut.com/simpledotorg/story/10850)

## Because

Partner organizations in Bangladesh need a reference to build the clients that integrate with us. They have requested detailed API specifications as well as a mock server.

## This addresses

* Adds API with payload validations to the `/import` endpoint.
* Adds a not-currently-exhaustive integration test suite which mostly behaves as a smoke test for now. We are deferring making this exhaustive, covering all resource types, etc., to get the mock server out and unblock the partner orgs.
* Adds detailed and descriptive Swagger specs.

## Not addressed

* OAuth endpoints are not covered. Partner organizations can test against something like [this](https://oauth.tools/collection/1599045253169-GHF) while we build this into Simple.
* The mock import endpoint currently works without an `Authorization` header, but this should not be a problem for those needing this as a development reference.

## Test instructions

* Launch the Simple Server.
* Ensure http://localhost:3000/api-docs has a page called `Import` with the API spec.
* Add flipper flag called `imports_api`.
* Run this request using Postman/cURL whatever, adjust the payload to be anything in the spec:
```
PUT http://localhost:3000/api/v4/import
Content-Type: application/json; charset=UTF-8
HTTP_X_ORGANIZATION_ID: 220d5e23-dc13-47c6-9ef7-30c51b1d4456

{"resources":[{"resourceType":"Patient","meta":{"lastUpdated":"2019-08-24T14:15:22Z","createdAt":"2019-08-24T14:15:22Z"},"identifier":[{"value":"string"}],"active":true,"name":{"text":"string"},"telecom":[{"value":"string","use":"home"}],"gender":"male","birthDate":"2019-08-24","deceasedBoolean":true,"managingOrganization":[{"value":"string"}],"registrationOrganization":[{"value":"string"}],"address":[{"line":["string"],"city":"string","district":"string","state":"string","postalCode":"string"}]}]}
```

The above should give `202 Accepted` with no errors. Any other payload would give a `400 Bad Request`.


app/controllers/api/v4/imports_controller.rb
```diff
@@ -0,0 +1,19 @@
+class Api::V4::ImportsController < ApplicationController
+  skip_before_action :verify_authenticity_token
+
+  def import
+    return head :not_found unless Flipper.enabled?(:imports_api)
```
danySam, line 5: Is the flipper flag temporary? Else we should add it to `doc/wiki/feature-flags.md`

spec/factories/import_resources/patients.rb
```diff
@@ -0,0 +1,40 @@
+def build_patient_import_resource
```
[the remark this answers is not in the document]
tfidfwastaken: Do you have some doc/reference to see what `OpenStruct` is? I could not find any documentation for it for whatever reason.

At the moment, I'm not too keen on generating these hashes any other way. We have built hashes this way in some other tests before and I think this is the least magical and simplest way to do it for our needs.

app/controllers/api/v4/imports_controller.rb
tfidfwastaken: Thanks for catching this; I think this will stick around for a while, I'll document it.

app/schema/api/v4/imports.rb
```diff
@@ -0,0 +1,520 @@
...
+            observation_blood_pressure
+            observation_blood_sugar
+            condition
+            medication_request
+          ].map { |resource| {"$ref" => "#/definitions/#{resource}"} }
+        }
+      }
+    end
+
+    def all_definitions
```
danySam, line 508: Why not call this simple `definitions`?

app/controllers/api/v4/imports_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    definition = Api::V4::Imports.import_resource_list.merge(definitions: Api::V4::Imports.all_definitions)
```
danySam, line 7: Can we not do this merge in `Api::V4:Imports`? `import_resource_list` + `all_definitions` = imports right?

app/schema/api/v3/models.rb
```diff
@@ -1,42 +1,5 @@
 class Api::V3::Models
   class << self
-    def timestamp
```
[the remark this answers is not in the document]
danySam: Oh, we moved these to common. Nvm

danySam (approved):
Left some minor comments. Otherwise, looks good to merge.

app/schema/api/v4/imports.rb
tfidfwastaken: > Why not call this simply `definitions`?

Mostly for consistency with the [methods returning definitions in other schemas](https://github.com/simpledotorg/simple-server/blob/03c414a330e3520367ba534942eded482a8869b8/app/schema/api/schema.rb#L88-L89). I don't feel strongly about either name for this particular case, so leaving it unchanged for now.

> Also, we can move this def to the top? And make other defs private?

I will move it to the top 👍 

I'd prefer not to make the other methods private because the pattern that I am seeing with these schema definitions is that we are trying to emulate it being Plain Old Data, rather than provide an OOP-class providing some kind of interface + behaviour. If I didn't care about being consistent with the other schemas, I'd have made this a module with module functions rather than a Class with class methods.

app/controllers/api/v4/imports_controller.rb
tfidfwastaken: Makes sense. This operation is done in the Validation layer for other schemas, but I hadn't made a validation class given how small this currently is. I'll create a `schema_with_definitions` in `Api::V4::Imports` for now.

spec/factories/import_resources/patients.rb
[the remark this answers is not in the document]
govindkrjoshi: https://thoughtbot.com/blog/mind-bending-factories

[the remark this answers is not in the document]
tfidfwastaken: Nice, thanks!

I'll consider this down the line if we need something more sophisticated like a trait. Still split on if we should do it this way, though.

PR #5210, opened 2023-07-13
Implement authentication for stakeholder resource access

tfidfwastaken:
**Story card:** [sc-10740](https://app.shortcut.com/simpledotorg/story/10740)

## Context

[Link to relevant section of the PRD](https://github.com/doorkeeper-gem/doorkeeper/wiki)

Currently, we need to create OAuth Applications for users that want to send us data through the `/import` API.

Since our partner organizations are trusted machine clients in this case, we create and send them a client ID and secret that is used for the [client credential flow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/client-credentials-flow) needed for the import API. We also create a machine user associated with these organizations.

We introduce a new dependency, [Doorkeeper](https://github.com/doorkeeper-gem/doorkeeper/wiki) to implement our OAuth provider.

The expiry token is set to 8 hours by default.

## This addresses

* Adds an OAuth provider, which involves migrations to hold OAuth-related details, all mostly generated by doorkeeper.
* Adds a machine user model, which currently only holds a name and organization.
* Secure the `/import` endpoint to only those who provide an OAuth token with the `write` scope.
* Adds a rake task to register and set up an OAuth application, whose details we will hand over to the partner orgs so they can integrate with us.

## Test instructions

Run the rake task (refer to the howto doc added in the PR).

Run the following HTTP request to obtain a token:
```
POST http://localhost:3000/oauth/token
Authorization: Basic <base64(client_id:client_secret)>
Content-Type: application/x-www-form-urlencoded;charset=UTF-8

grant_type=client_credentials
```

Copy the token ID, and call the API with it:
```
PUT http://localhost:3000/api/v4/import
Content-Type: application/json; charset=UTF-8
X-Organization-ID: <use the org ID passed into the rake task>
Authorization: Bearer <token>

{"resources": []} # use a valid resource if you want a 202 response instead of a 400
```

Any status code that's not 401 or 403 means the authorization worked. Conversely, an invalid token should return a 403 and no token at all should return a 401.

### DIFF
--- a/app/controllers/api/v4/imports_controller.rb
+++ b/app/controllers/api/v4/imports_controller.rb
@@ -1,5 +1,7 @@
 class Api::V4::ImportsController < ApplicationController
   skip_before_action :verify_authenticity_token
+  before_action :doorkeeper_authorize!
+  before_action :validate_token_organization
 
   def import
     return head :not_found unless Flipper.enabled?(:imports_api)
@@ -17,5 +19,12 @@ class Api::V4::ImportsController < ApplicationController
 
   def import_params
     params.require(:resources)
+  end
+
+  def validate_token_organization
+    token_organization = MachineUser.find_by(id: doorkeeper_token.application&.owner_id)&.organization_id
+    unless token_organization.present? && token_organization == request.headers["HTTP_X_ORGANIZATION_ID"]
+      head :forbidden
+    end
   end
 end
--- a/config/initializers/doorkeeper.rb
+++ b/config/initializers/doorkeeper.rb
@@ -1,0 +1,507 @@
+# frozen_string_literal: true
+
+Doorkeeper.configure do
+  # Change the ORM that doorkeeper will use (requires ORM extensions installed).
+  # Check the list of supported ORMs here: https://github.com/doorkeeper-gem/doorkeeper#orms
+  orm :active_record
+
+  # This block will be called to check whether the resource owner is authenticated or not.
+  resource_owner_authenticator do
+    # Since Authorization Code flow is currently disabled, we will redirect to the admin homepage.
+    redirect_to admins_url
+  end
+
+  # If you didn't skip applications controller from Doorkeeper routes in your application routes.rb
+  # file then you need to declare this block in order to restrict access to the web interface for
+  # adding oauth authorized applications. In other case it will return 403 Forbidden response
+  # every time somebody will try to access the admin web interface.
+  #
+  # admin_authenticator do
+  #   # Put your admin authentication logic here.
+  #   # Example implementation:
+  #
+  #   if current_user
+  #     head :forbidden unless current_user.admin?
+  #   else
+  #     redirect_to sign_in_url
+  #   end
+  # end
+
+  # You can use your own model classes if you need to extend (or even override) default
+  # Doorkeeper models such as `Application`, `AccessToken` and `AccessGrant.
+  #
+  # Be default Doorkeeper ActiveRecord ORM uses it's own classes:
+  #
+  # access_token_class "Doorkeeper::AccessToken"
+  # access_grant_class "Doorkeeper::AccessGrant"
+  # application_class "Doorkeeper::Application"
+  #
+  # Don't forget to include Doorkeeper ORM mixins into your custom models:
+  #
+  #   *  ::Doorkeeper::Orm::ActiveRecord::Mixins::AccessToken - for access token
+  #   *  ::Doorkeeper::Orm::ActiveRecord::Mixins::AccessGrant - for access grant
+  #   *  ::Doorkeeper::Orm::ActiveRecord::Mixins::Application - for application (OAuth2 clients)
+  #
+  # For example:
+  #
+  # access_token_class "MyAccessToken"
+  #
+  # class MyAccessToken < ApplicationRecord
+  #   include ::Doorkeeper::Orm::ActiveRecord::Mixins::AccessToken
+  #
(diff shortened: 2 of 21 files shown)


app/controllers/api/v4/imports_controller.rb
```diff
@@ -1,5 +1,7 @@
 class Api::V4::ImportsController < ApplicationController
   skip_before_action :verify_authenticity_token
+  before_action :doorkeeper_authorize!
```
tfidfwastaken, line 3: This is a helper that `doorkeeper` provides us. It carries out the token verification based on our configs.

app/controllers/api/v4/imports_controller.rb
```diff
@@ -18,4 +20,11 @@ def import
   def import_params
     params.require(:resources)
   end
+
+  def validate_token_organization
+    token_organization = MachineUser.find_by(id: doorkeeper_token.application&.owner_id)&.organization_id
+    unless token_organization.present? && token_organization == request.headers["HTTP_X_ORGANIZATION_ID"]
+      head :forbidden
+    end
+  end
```
tfidfwastaken, line 29: This validates if the Auth token is associated with the organization that the requester intended. Mostly included for defense-in-depth.

config/locales/doorkeeper.en.yml
tfidfwastaken, line 1: This locale file was generated by doorkeeper. Translations are not in the scope of this PR.

config/routes.rb
```diff
@@ -11,6 +11,8 @@
     end
   end
 
+  use_doorkeeper
```
tfidfwastaken, line 14: This creates all the OAuth endpoints for us.

db/migrate/20230713065237_create_machine_users.rb
tfidfwastaken, line 1: Every OAuth application is associated with a `MachineUser`.

db/migrate/20230713065420_create_doorkeeper_tables.rb
tfidfwastaken, line 1: Most of this file is generated by doorkeeper.

db/migrate/20230713065420_create_doorkeeper_tables.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    add_foreign_key(
+      :oauth_access_tokens,
+      :oauth_applications,
+      column: :application_id
+    )
+
+    # Uncomment below to ensure a valid reference to the resource owner's table
+    add_foreign_key :oauth_access_grants, :machine_users, column: :resource_owner_id
+    add_foreign_key :oauth_access_tokens, :machine_users, column: :resource_owner_id
```
tfidfwastaken, line 96: Every token and grant is associated with a `MachineUser` as well.

db/migrate/20230713135154_add_owner_to_application.rb
tfidfwastaken, line 1: This is another generated doorkeeper migration that creates an association between an OAuth application and an "owner" of this application. This is useful for validating that the application is able to only do whatever the owner is allowed to do.

lib/tasks/setup_oauth_application.rake
tfidfwastaken, line 1: We'll use these to create the MachineUser and OAuth application for our partner orgs. Since there's going to only be 4 orgs for now, it didn't make sense to implement a self-service dashboard.

spec/api/v4/imports/imports_spec.rb
```diff
@@ -3,25 +3,34 @@
...
+      FactoryBot.create(:oauth_access_token,
+        application: application,
+        scopes: "write",
+        resource_owner_id: machine_user.id)
+    }
     put "Send bulk resources to Simple" do
       tags "import"
       security [access_token: [], import_auth: ["write"]]
-      parameter name: :organization_header,
+      parameter name: :HTTP_X_ORGANIZATION_ID,
```
tfidfwastaken, line 19: Making this more consistent with the other API header texts in the generated swagger. Unfortunately, due to how rswag works, there's no way to give the correct name for this API header, which is `X-Organization-ID`. The workaround for now is to explicitly call it out in the description below.

config/locales/doorkeeper.en.yml
tfidfwastaken, line 1: This locale file was generated by doorkeeper. Translations are not in the scope of this PR.

lib/tasks/setup_oauth_application.rake
tfidfwastaken, line 1: We'll use this rake task to create the MachineUser and OAuth application for our partner orgs. Since there's going to only be 4 orgs for now, it didn't make sense to implement a self-service dashboard.

db/migrate/20230713065237_create_machine_users.rb
tfidfwastaken, line 1: Every OAuth application will be associated with this `MachineUser`.

db/migrate/20230713135154_add_owner_to_application.rb
tfidfwastaken, line 1: This is another generated doorkeeper migration that creates an association between an OAuth application and an "owner" of this application. This is useful for validating that the application is able to only do whatever the owner is allowed to do.

config/initializers/doorkeeper.rb
tfidfwastaken, line 1: This file is generated by doorkeeper. I found it useful to keep the comments around that explain all the config knobs.

olttwa (approved):
1st pass review. Everything looks good.

PR #5212, opened 2023-07-17
Support Drug Stock Reports via questionnaires (aka dynamic forms)

olttwa:
**Story card:** [sc-10883](https://app.shortcut.com/simpledotorg/story/10883/move-et-drug-stocks-to-questionnaires)

## Because

We are moving drug-stock form from a web-view to an offline-first dynamic questionnaire in Ethiopia. We need date input for recording expiration of a batch of drugs.

## This addresses
1. Support for date input field in questionnaires
2. Support for drug-stock report type in questionnaires
3. Initialize questionnaire responses every month


app/helpers/questionnaire_responses_helper.rb
```diff
@@ -0,0 +1,16 @@
+module QuestionnaireResponsesHelper
+  def latest_active_questionnaire_id(type)
+    Questionnaire.active.where(questionnaire_type: type).order(dsl_version: :desc).first.id
+  end
+
+  def month_date_str
+    @month_date.strftime("%Y-%m-%d")
```
olttwa, line 7: A QuestionnaireResponse service has an instance variable `@month_date` based on which the string is generated. We're using that in the helper module.
It is not considered best practice to use instance variables in a module, but is [acceptable in the rails world](https://stackoverflow.com/questions/70355343/ruby-on-rails-helper-method-instance-variable), and we're using it in other modules in Simple too.

app/services/questionnaire_responses/monthly_screening_reports.rb
```diff
@@ -21,10 +23,6 @@ def pre_fill
 
   private
 
-  def month_date_str
```
olttwa, line 24: Moved 3 methods, which were reused by form initializing services to a common helper module.

app/services/questionnaire_responses/monthly_screening_reports.rb
```diff
@@ -55,32 +49,26 @@ def monthly_screening_report_exists?(facility_id)
...
-          "monthly_screening_report.diagnosed_cases_on_follow_up_htn.male" => facility_report.monthly_follow_ups_htn_male,
-          "monthly_screening_report.diagnosed_cases_on_follow_up_htn.female" => facility_report.monthly_follow_ups_htn_female,
-          "monthly_screening_report.diagnosed_cases_on_follow_up_dm.male" => facility_report.monthly_follow_ups_dm_male,
-          "monthly_screening_report.diagnosed_cases_on_follow_up_dm.female" => facility_report.monthly_follow_ups_dm_female,
-          "monthly_screening_report.diagnosed_cases_on_follow_up_htn_and_dm.male" => facility_report.monthly_follow_ups_htn_and_dm_male,
-          "monthly_screening_report.diagnosed_cases_on_follow_up_htn_and_dm.female" => facility_report.monthly_follow_ups_htn_and_dm_female
-        }
-      )
+      {
+        **monthly_reports_base_content,
```
olttwa, line 55: Moved from merge function to splatting for merging 2 hashes.

config/schedule.rb
```diff
@@ -127,6 +127,9 @@ def local(time)
   if Flipper.enabled?(:monthly_supplies_reports)
     runner "QuestionnaireResponses::MonthlySuppliesReports.new.seed"
   end
+  if Flipper.enabled?(:drug_stock_questionnaires)
```
olttwa, line 130: We are not using `:drug_stock` flipper flag, because that is already enabled in all the deployments. Instead, we've introduced a new flag to mark drug stocks' migration to dynamic forms.

app/helpers/questionnaire_responses_helper.rb
```diff
@@ -0,0 +1,16 @@
+module QuestionnaireResponsesHelper
+  def latest_active_questionnaire_id(type)
+    Questionnaire.active.where(questionnaire_type: type).order(dsl_version: :desc).first.id
```
tfidfwastaken, line 3: [suggestion, non-blocking]
If there are no Questionnaires in the database, this line would throw a NoMethodError exception. Can this be a safe navigation, so that response creation fails in the `seed` method instead?
```suggestion
    Questionnaire.active.where(questionnaire_type: type).order(dsl_version: :desc).first?.id
```

If I were a downstream user of this mixin, I wouldn't conventionally expect a method called `latest_active_questionnaire_id` to throw an exception when there's no active questionnaire.

spec/services/questionnaire_responses/drug_stock_reports_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      expect(QuestionnaireResponse.where(facility: facility).count).to eq(1)
+      expect(QuestionnaireResponse.find_by_facility_id(facility)).to eq(existing_drug_stock_report)
+    end
+
+    it "overrides date when provided as an argument" do
+      three_months_ago = 3.months.ago
+
+      QuestionnaireResponses::DrugStockReports.new(three_months_ago).seed
+
+      expect(QuestionnaireResponse.find_by_facility_id(facility).content).to eq({"month_date" => three_months_ago.beginning_of_month.strftime("%Y-%m-%d"), "submitted" => false})
```
tfidfwastaken, line 41: [nit] This line is very looooooooooooooonnnnnggggg
perhaps the hashmap could be spread over multiple lines

tfidfwastaken (approved):
Looks solid. Left a suggestion on handling failure when there's no active questionnaire. Approved to unblock.

spec/services/questionnaire_responses/drug_stock_reports_spec.rb
olttwa: Shortened it.

app/helpers/questionnaire_responses_helper.rb
olttwa: Good suggestion. `first?` method won't work on arrays, have changed it to `first.try(:id)`

spec/factories/questionnaires.rb
```diff
@@ -2,7 +2,7 @@
   factory :questionnaire do
     id { SecureRandom.uuid }
     questionnaire_type { "monthly_screening_reports" }
```
PriyangaPKini, line 4: Let's randomly pick the type

swagger/v4/swagger.json
```diff
@@ -3524,6 +3529,289 @@
...
+      "example": {
+        "id": "2e8ce537-616c-4c4c-a651-ad065d05f220",
+        "type": "group",
+        "view_type": "view_group",
+        "item": [
+          {
+            "id": "7c660864-b50a-4b3c-b019-c37a3fba3b8e",
+            "type": "display",
+            "view_type": "paragraph",
+            "text": "Enter the supplies left in stock at the end of every month"
```
PriyangaPKini, line 3543: Let's use translated strings

spec/factories/questionnaires.rb
```diff
@@ -2,7 +2,7 @@
...
-    dsl_version { "1.1" }
+    dsl_version { "1.2" }
```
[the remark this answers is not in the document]
olttwa: Not doing this now, since a latest dsl_version doesn't mean anything in context of a factory data generation.

PR #5214, opened 2023-07-21
Bugfix: Experiment notifications not going out as expected

qptr:
**Story card:** [sc-10920](https://app.shortcut.com/simpledotorg/story/10920)

## Because

Notifications haven't been getting created or sent out in Bnagladesh and India since July 3. This is blocking the Bangladesh SMS experiment, so this is a quick fix. A more long-term fix can be picked up as part of [this card](https://app.shortcut.com/simpledotorg/story/10921/).

## This addresses

When checking for patients to be notified for the day, a timestamp and date are compared. Since the date gets implicitly converted to 00:00 in UTC, this comparison has been failing ever since we wrapped the experiment runner so as to run within the local timezone. This fix saves these dates with the correct date in the UTC format while creating a treatment group membership.

The long term fix is to add a schema migration to convert the `expected_return_date` and `experiment_inclusion_date` fields into dates.

## Test instructions

- Create stale and current experiments, their treatment groups and reminder templates.
- Run conduct_daily on them with any date
- Check that treatment group memberships are being created as expected, and more importantly that notifications are being created.


spec/models/experimentation/current_patient_experiment_spec.rb
```diff
@@ -30,6 +30,26 @@
...
+        treatment_group.enroll(patient_1, expected_return_date: 2.days.from_now.beginning_of_day)
+      end
+      Time.use_zone(CountryConfig.current[:time_zone]) do
+        treatment_group.enroll(patient_2, expected_return_date: 2.days.from_now.middle_of_day)
+      end
+      membership_1 = experiment.treatment_group_memberships.find_by(patient_id: patient_1.id)
+      membership_2 = experiment.treatment_group_memberships.find_by(patient_id: patient_2.id)
+
+      expect(membership_1.expected_return_date.to_date).to eq(1.day.from_now.to_date)
+      expect(membership_2.expected_return_date.to_date).to eq(2.day.from_now.to_date)
```
qptr, line 49: thinking if i should remove these:
- they'll fail in case the configured tz for tests is not IST
- but i added them to make it easy to understand why the fix is necessary
- the configured tz might not change (check this)
- this is a temporary fix, though unsure when we'll pick up the longterm fix of changing the field types from timestamp to date

qptr: Set this to Asia/Kolkata so it's not dependednt on the country config.

app/models/experimentation/current_patient_experiment.rb
```diff
@@ -15,11 +15,15 @@ def eligible_patients(date, filters = {})
...
+    #
+    # Hotfix: expected_return_date is currently saved after being converted
+    # from a date, into a timestamp with local timezone, and then to utc.
+    # So, to get back the correct expected_return_date, we're reversing that process here.
     def memberships_to_notify(date)
       treatment_group_memberships
         .status_enrolled
         .joins(treatment_group: :reminder_templates)
-        .where("expected_return_date::timestamp + make_interval(days := reminder_templates.remind_on_in_days) = ?", date)
+        .where("date_trunc('day', expected_return_date AT TIME ZONE 'UTC' AT TIME ZONE '#{TIMEZONE}') + make_interval(days := reminder_templates.remind_on_in_days) = ?", date)
```
govindkrjoshi, line 26:
```suggestion
        .where("date_trunc('day', expected_return_date AT TIME ZONE 'UTC' AT TIME ZONE ?) + make_interval(days := reminder_templates.remind_on_in_days) = ?", TIMEZONE, date)
```

app/models/experimentation/stale_patient_experiment.rb
```diff
@@ -32,11 +32,15 @@ def eligible_patients(date, filters = {})
...
+    #
+    # Hotfix: experiment_inclusion_date is currently saved after being converted
+    # from a date, into a timestamp with local timezone, and then to utc.
+    # So, to get back the correct experiment_inclusion_date, we're reversing that process here.
     def memberships_to_notify(date)
       treatment_group_memberships
         .status_enrolled
         .joins(treatment_group: :reminder_templates)
-        .where("experiment_inclusion_date::timestamp + make_interval(days := reminder_templates.remind_on_in_days) = ?", date)
+        .where("date_trunc('day', experiment_inclusion_date AT TIME ZONE 'UTC' AT TIME ZONE '#{TIMEZONE}') + make_interval(days := reminder_templates.remind_on_in_days) = ?", date)
```
govindkrjoshi, line 43: Same here.

govindkrjoshi (changes requested):
As per our earlier discussion, also need to make sure that the membership is created for the right day.

app/models/experimentation/notifications_experiment.rb
```diff
@@ -259,14 +260,16 @@ def reporting_data(patient, date)
...
+      expected_return_date_utc_timestamp = expected_return_date&.to_time(:utc)
 
       {
         gender: patient.gender,
         age: patient.current_age,
         risk_level: patient.risk_priority,
         diagnosed_htn: medical_history.hypertension,
-        experiment_inclusion_date: date,
-        expected_return_date: latest_scheduled_appointment&.remind_on || latest_scheduled_appointment&.scheduled_date,
+        experiment_inclusion_date: date.to_time(:utc),
```
olttwa, line 271:
```suggestion
        experiment_inclusion_date: date.utc,
```

app/models/experimentation/notifications_experiment.rb
```diff
@@ -259,14 +260,16 @@ def reporting_data(patient, date)
       latest_scheduled_appointment = patient.latest_scheduled_appointment
       assigned_facility = patient.assigned_facility
       registration_facility = patient.registration_facility
+      expected_return_date = latest_scheduled_appointment&.remind_on || latest_scheduled_appointment&.scheduled_date
+      expected_return_date_utc_timestamp = expected_return_date&.to_time(:utc)
```
olttwa, line 264: `to_time` is for converting string instances to date. If the intention is to convert to utc timezone, it might be better to use `.utc` instead?
```suggestion
      expected_return_date_utc_timestamp = expected_return_date&.utc
```

olttwa: Do we want to convert a `Date` object into a `DateTime` object, and then persist to DB?
If yes, then this implementation is correct.

olttwa (commented):
Just left a few questions.

app/models/experimentation/stale_patient_experiment.rb
```diff
@@ -32,11 +32,15 @@ def eligible_patients(date, filters = {})
 
     # Memberships where enrollment date falls on
     # one of the reminder template's remind_on days since `date`.
+    #
```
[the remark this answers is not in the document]
qptr: Yeah, before we did a couple of tz conversions here. Now, the timestamp field is getting truncated to a date before comparing so there's no implicit, hidden conversion.

olttwa (approved):

govindkrjoshi (approved):

PR #5217, opened 2023-07-26
Set up SMS experiments in Bangladesh for August 2023

qptr:
Story id: [10990](https://app.shortcut.com/simpledotorg/story/10990)

We're running sms experiments in Bangladesh. We scrapped the last one due to a [bug](https://app.shortcut.com/simpledotorg/story/10913). Now we're creating a fresh set to be in place when [the fix](https://github.com/simpledotorg/simple-server/pull/5214) gets merged.

This experiment ends on 8 Sep instead of on 30 Sep because we want a full 30-day run of the experiment for a sufficient sample size. We want to eventually revert back to starting and ending experiments on the first and last days of the month for ease of tracking.


govindkrjoshi (approved):

qptr:
Keeping this PR around till the NHF gets back with when we can restart sending sms reminders in Bangladesh.

PR #5224, opened 2023-08-03
Recent updates: Goal line 3 month average

jamiecarter7:
## Because

Recent updates: Aug-2023

## This addresses

Inform users that the goal line calculations have changed from Dec of the previous year, to an average of Q4 months from the previous year.

## Test instructions

Goal lines wont show on your localhost as they are hidden from India (not live in India).
You will need to change `!= 'India'` to `== 'India'`


dburka (approved):
Fixed your compound sentence. Otherwise LGTM.

PR #5229, opened 2023-08-09
Allow import of patients via the bulk import API

tfidfwastaken:
**Story card:** https://app.shortcut.com/simpledotorg/story/10962/add-ingestion-for-patients
[sc-10962]

## Because

We are currently mocking bulk API imports. We now actually accept real resources if they are patient resources.

## This addresses

* Adds an import job and import type dispatcher.
* Adds another feature flag to have everything be a mock. This flag will be removed when the implementation of all resources is complete. We should enable this flag in sandbox, while partner organisations are still in the process of sending their facility data.
* Moves the `ImportUser` class out of the `PatientImport` namespace, since we are going to use this user for our FHIR imports.
* Updates some swagger schema errors.

## Test instructions

Follow the instructions in this document to insert a resource in Sandbox:
https://docs.google.com/document/d/1f-H474ozfNBjiISMd7IOLQtZREVJh86u6JBnw4235qw/edit?usp=sharing


app/controllers/api/v4/imports_controller.rb
```diff
@@ -11,14 +11,48 @@ def import
       import_params.to_json
     )
 
+    unless Flipper.enabled?(:mock_imports_api)
```
tfidfwastaken, line 14: I'll enable this flag as soon as we ship this feature.

app/schema/api/v4/imports.rb
tfidfwastaken, line 1: All changes in this file are enhancements and fixes for our API docs, not directly related to the import implementation.

app/services/bulk_api_import/fhir_importable.rb
```diff
@@ -0,0 +1,24 @@
+module BulkApiImport::FhirImportable
+  def import_user
+    ImportUser.find_or_create
+  end
+
+  def import
+    raise NotImplementedError
+  end
```
tfidfwastaken, line 8: leaving this here to signal that any class that includes this mixin is expected to have an `import` method.

app/services/bulk_api_import/importer.rb
```diff
@@ -0,0 +1,30 @@
...
+    %w[Patient].include?(resource[:resourceType])
+  end
+
+  def resource_importer(resource)
+    case resource[:resourceType]
+    when "Patient"
+      BulkApiImport::FhirPatientImporter.new(resource)
+    # when "Appointment"
+    #   BulkApiImport::FhirAppointmentImporter.new(resource)
+    # ...
```
tfidfwastaken, line 25: Temporary comment to show how the rest of this dispatching works. This code will be real in the next PR.

spec/services/bulk_api_import/importer_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  describe "#resource_importer" do
+    it "fetches the correct importer" do
+      importer = described_class.new(resource_list: [])
+
+      [
+        {input: {resourceType: "Patient"}, expected_importer: BulkApiImport::FhirPatientImporter}
+        # {input: {resourceType: "Appointment"}, expected_importer: BulkApiImport::FhirAppointmentImporter},
+        # {input: {resourceType: "Observation"}, expected_importer: BulkApiImport::FhirObservationImporter}
```
tfidfwastaken, line 39: These will also become real test cases in the next PR, left it here to show how it would look like for the rest of the cases

app/services/bulk_api_import/fhir_importable.rb
```diff
@@ -0,0 +1,24 @@
...
+      updated_at: @resource.dig(:meta, :lastUpdated)
+    }
+  end
+
+  def translate_id(id)
+    Digest::UUID.uuid_v5(Digest::UUID::DNS_NAMESPACE, id)
+  end
+
+  def translate_facility_id(id)
+    FacilityBusinessIdentifier.where(identifier_type: :external_org_facility_id, identifier: id).take.facility.id
```
govindkrjoshi, line 22:
```suggestion
    FacilityBusinessIdentifier.find_by(identifier_type: :external_org_facility_id, identifier: id).facility.id
```

govindkrjoshi: This will also throw an uncaught exception here if the business identifier is not found, and the user will not get any feedback for this error. Maybe we should check in the controller that the facility business identifier exists for all the facility ids sent in the payload. Alternatively, maybe we can create dummy facilities for the organisation if the id they've sent doesn't exist in our system. Maybe we should discuss this a bit. 

Note although that this is not a blocker for the PR.

app/services/bulk_api_import/fhir_patient_importer.rb
```diff
@@ -0,0 +1,91 @@
+class BulkApiImport::FhirPatientImporter
+  include BulkApiImport::FhirImportable
```
govindkrjoshi, line 2: Why not use inheritance?

app/services/bulk_api_import/importer.rb
```diff
@@ -0,0 +1,30 @@
+class BulkApiImport::Importer
+  def initialize(resource_list:)
+    @resources = resource_list
+  end
+
```
govindkrjoshi, line 5: IMPORTERS = { "Patient" =>  BulkApiImport::FhirPatientImporter }

app/services/bulk_api_import/importer.rb
```diff
@@ -0,0 +1,30 @@
...
+      resource_importer(resource).import if import_enabled?(resource)
+    end
+  end
+
+  # TODO:
+  # This method is a stopgap to ensure that our mock imports and real imports can coexist.
+  # Once we have implemented all resource types, we should remove this method entirely.
+  def import_enabled?(resource)
+    %w[Patient].include?(resource[:resourceType])
+  end
```
govindkrjoshi, line 17:
```suggestion
    IMPORTERS[resource[:resourceType]].present?
  end
```

app/services/bulk_api_import/fhir_importable.rb
tfidfwastaken: Noted. There's another PR of the same stuff coming through, let's solidify an error handling strategy

app/services/bulk_api_import/importer.rb
```diff
@@ -0,0 +1,30 @@
...
+  def resource_importer(resource)
+    case resource[:resourceType]
+    when "Patient"
+      BulkApiImport::FhirPatientImporter.new(resource)
+    # when "Appointment"
+    #   BulkApiImport::FhirAppointmentImporter.new(resource)
+    # ...
+    else
+      throw NotImplementedError
+    end
```
govindkrjoshi, line 28:
```suggestion
    importer = IMPORTERS[resource[:resourceType]]
    throw NotImplementedError unless importer.present?
    importer.new(resource)
```

app/services/import_user.rb
```diff
@@ -1,4 +1,4 @@
-class PatientImport::ImportUser
+class ImportUser
```
govindkrjoshi, line 1: Why has this service moved?

app/services/bulk_api_import/fhir_patient_importer.rb
tfidfwastaken: A few reasons behind this:
* I intended to model `FhirImportable` as behaviour, and it does not encapsulate any state.
* I don't see us instantiating and using this as an independent object ever.
* I don't see much benefit to having an inheritance hierarchy here—it's as reusable as a mixin, but comes with some more baggage like the lack of composability, IMO

Open to changing my mind for this if you favour inheritance here

app/services/bulk_api_import/importer.rb
tfidfwastaken: This is nice, thanks.

app/services/bulk_api_import/importer.rb
tfidfwastaken: I'll skip this one, because this stopgap method will go away in the next PR.

app/services/import_user.rb
tfidfwastaken: This import module was created and used for Patient import via CSV. I think the scope of this import user has now expanded beyond being a utility for only importing patients—it's now also used by all the FHIR importers for all kinds of resources. So I thought of moving this out.

It also fixes an awkward dependency graph where FHIR import services depend on an internal module of the `PatientImport` service.

PR #5233, opened 2023-08-10
Overdue: Date picker to change number of dashed lines

jamiecarter7:
- MonthsDashedBased is the base number of dashes for a chart
- adding the current date string (todays date Mmm-yyyy) to the function arguments will affect the number of dashed segments depending on how many months between today and the latest month on the chart.

`monthIndexFromChartLabelDateString()` has been moved to a global function as its used in both parts of the reports.js file so I can no longer place it in just one part.

**Story card:** [sc-11007](https://app.shortcut.com/simpledotorg/story/11007/overdue-dashed-lines-showing-incorrectly-on-previous-month-reports)

## Because

Its been highlighted that when selecting an old date on the date picker in the dashboard (eg Jul-2023), the number of dashed line segments show the same even for older dates where the figures are now fixed.
Aug-2023
Screenshot 2023-08-10 at 11 35 18
Jul-2023
Screenshot 2023-08-10 at 11 38 52

## This addresses

It makes the charts confusing to read as the dashboard is stating the there are still changes occurring on older date ranges, which is not the case.

## Test instructions

For Aug-2023
Overdue called - should show 1 dashed segment
Overdue returned: 
- 1st to 15th - 2 dashed lines
- 16th to end of month - 1 dashed line

Now select Jul-2023 from the date picker
Overdue called - should have no dashed lines
Overdue returned: 
- 1st to 15th - 1 dashed line
- 16th to end of month - 0 dashed lines

Now select Jun-2023 from the date picker
Overdue called - should have no dashed lines
Overdue returned - 0 dashed lines

Please check that no errors are thrown with the goal lines activated on all of the date ranges. They share a common function which has change name - `monthIndexFromChartLabelDateString`
Flipper flag: `goal-lines`

You will need to add the following to the config to show goal lines in localhost
in `const enabledRegions`
change the IN line to the following 
`IN: ["organization", "state", "district"],`


app/assets/javascripts/common/reports.js
```diff
@@ -696,33 +708,14 @@ Reports = function ({
...
-      "jun",
-      "jul",
-      "aug",
-      "sep",
-      "oct",
-      "nov",
-      "dec",
-    ];
-
-    return months.indexOf(month.toLowerCase());
```
PriyangaPKini, line 723: Let's use Javascript built-in function to get the month number from month string.
```
new Date("jan-2023").getMonth()
```

app/assets/javascripts/common/reports.js
```diff
@@ -1782,3 +1796,23 @@ function mergeArraysWithConcatenation(objValue, srcValue) {
...
+    "jun",
+    "jul",
+    "aug",
+    "sep",
+    "oct",
+    "nov",
+    "dec",
+  ];
+
+  return months.indexOf(month.toLowerCase());
```
PriyangaPKini, line 1817: Let's use Javascript built-in function to get the month number from the month string.
```
new Date("jan-2023").getMonth()
```

PriyangaPKini (changes requested):
I have requested few changes.

app/assets/javascripts/common/reports.js
jamiecarter7: I did not know that was a thing. This is way better 👍

jamiecarter7:
Great suggestions - the pop() is much simpler!

Changes pushed

app/assets/javascripts/common/reports.js
```diff
@@ -676,7 +682,9 @@ Reports = function ({
 
     const decemberKeys = dateKeysArray.filter((item) => item.includes("Dec"));
```
PriyangaPKini, line 683: Nitpick:  While naming a variable, we use the plural to indicate it's a collection. Instead of `dateKeysArray`, use  `dateKeys`.

app/assets/javascripts/common/reports.js
```diff
@@ -676,7 +682,9 @@ Reports = function ({
...
     const isLastDateKeysArrayMonthDec =
-      monthIndexFromDateString(dateKeysArray[dateKeysArray.length - 1]) === 11;
+      monthIndexFromChartLabelDateString(
+        dateKeysArray[dateKeysArray.length - 1]
+      ) === 11;
     if (isLastDateKeysArrayMonthDec) {
```
PriyangaPKini, line 688: Nitpick: We could rename  'isLastDateKeysArrayMonthDec' to 'isEndMonthOfYear' or 'isLastMonthOfYear'. This makes the function a little more clear.

PR #5236, opened 2023-08-16
Import API: allow importing remaining FHIR resources

tfidfwastaken:
**Story card:** [sc-10963], [sc-10964], [sc-10965], [sc-10966], [sc-10967]

## This addresses

This is similar to #5229, except we now allow the rest of the resource types that we are accepting.
See: https://api.simple.org/api-docs/index.html#tag/import/paths/~1import/put for the import resource payloads.

We also introduce an extra validation that ensures that all facilities in the payload have been mapped.

PRD: https://docs.google.com/document/d/1K5ve1i3_BI7jI92n-Dl1v-NQOU3bT-emVsJcgXMAml0/edit?usp=sharing

## Test instructions

Same as #5229, but with other resource types.

### DIFF
--- a/app/controllers/api/v4/imports_controller.rb
+++ b/app/controllers/api/v4/imports_controller.rb
@@ -6,10 +6,7 @@ class Api::V4::ImportsController < ApplicationController
   def import
     return head :not_found unless Flipper.enabled?(:imports_api)
 
-    errors = JSON::Validator.fully_validate(
-      Api::V4::Imports.schema_with_definitions,
-      import_params.to_json
-    )
+    errors = BulkApiImport::Validator.new(organization: header_organization_id, resources: import_params).validate
 
     unless Flipper.enabled?(:mock_imports_api)
       BulkApiImportJob.perform_later(resources: import_params) unless errors.present?
@@ -130,8 +127,12 @@ class Api::V4::ImportsController < ApplicationController
 
   def validate_token_organization
     token_organization = MachineUser.find_by(id: doorkeeper_token.application&.owner_id)&.organization_id
-    unless token_organization.present? && token_organization == request.headers["HTTP_X_ORGANIZATION_ID"]
+    unless token_organization.present? && token_organization == header_organization_id
       head :forbidden
     end
+  end
+
+  def header_organization_id
+    request.headers["HTTP_X_ORGANIZATION_ID"]
   end
 end
--- a/app/services/bulk_api_import/validator.rb
+++ b/app/services/bulk_api_import/validator.rb
@@ -1,0 +1,71 @@
+class BulkApiImport::Validator
+  def initialize(organization:, resources:)
+    @organization = organization
+    @params = resources
+  end
+
+  def validate
+    error = validate_schema
+    unless error.present?
+      error = validate_facilities
+    end
+
+    error
+  end
+
+  def validate_schema
+    schema_errors = JSON::Validator.fully_validate(
+      Api::V4::Imports.schema_with_definitions,
+      @params.to_json
+    )
+    {schema_errors: schema_errors} if schema_errors.present?
+  end
+
+  def validate_facilities
+    facility_ids = [
+      *patient_resource_facilities,
+      *appointment_resource_facilities,
+      *observation_resource_facilities,
+      *medication_request_resource_facilities
+    ]
+
+    found_facilities = FacilityBusinessIdentifier
+      .joins(facility: :facility_group)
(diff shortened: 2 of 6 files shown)


app/schema/api/v4/imports.rb
tfidfwastaken, line 1: These are a bunch of mistakes that crept up in our API schemas. Making them more FHIR-compliant.

app/services/bulk_api_import/fhir_patient_importer.rb
```diff
@@ -8,10 +8,12 @@ def initialize(patient_resource)
...
-    transformed_params = Api::V3::PatientTransformer.from_nested_request(build_attributes)
-    patient = MergePatientService.new(transformed_params, request_metadata: request_metadata).merge
-    AuditLog.merge_log(import_user, patient) if patient.present?
-    patient
+    merge_result = build_attributes
+      .then { Api::V3::PatientTransformer.from_nested_request(_1) }
+      .then { MergePatientService.new(_1, request_metadata: request_metadata).merge }
+
+    AuditLog.merge_log(import_user, merge_result) if merge_result.present?
+    merge_result
```
tfidfwastaken, line 16: Refactored this to match how I wrote it for the other resources. The chaining version expresses the intent better to my eyes.

swagger/v4/swagger.json
```diff
@@ -3541,7 +3541,7 @@
             "id": "7c660864-b50a-4b3c-b019-c37a3fba3b8e",
             "type": "display",
             "view_type": "paragraph",
-            "text": "Enter the supplies left in stock at the end of every month"
+            "text": "If you have more than 3 batches then add any additional batch stocks to batch 3 using the same date"
           },
```
tfidfwastaken, line 3545: @olttwa Can you check if this is the intended value? I haven't tweaked anything related to questionnaires, and I suspect in #5212 the PR hadn't included the latest generated swagger JSON. This showed up when I regenerated it for the Bulk Import Docs.

olttwa: @tfidfwastaken acc to master branch, the expected text here is `"Enter the supplies left in stock at the end of every month"`. Can rollback your merged changes.

tfidfwastaken: Yeah, I saw that in the master branch too, but is the master branch correct? This `swagger.json` is generated from `specimen_layout.rb`, and currently the value of that is `If you have more than 3 batches...` as seen here: https://github.com/simpledotorg/simple-server/blob/f5bb61a2266eabbaaec9cfe4a8dcc33ba84f7e87/app/schema/api/v4/models/questionnaires/specimen_layout.rb#L220

So the source and generated file went out of sync at some point—I'd have assumed that the source file should be the one to go with, not the generated file. Can you confirm again?

tfidfwastaken: Okay, I went digging a bit more and this Slack thread suggests that it should be `"Enter the ..."`: https://simpledotorg.slack.com/archives/C01BYR9M933/p1690968264030099?thread_ts=1690904295.849239&cid=C01BYR9M933

The change #5223 made in response to that changed only the `en.yml` locale file and missed changing the `specimen_layout.rb`, which is why any subsequent PR that calls `rswag:generate` will create this incorrect diff (Example: [this one](https://github.com/simpledotorg/simple-server/pull/5242/files#diff-87e0f7763f4c24150621d28d7df5966c0e574b07507b039c7d8cdc9ab8e872afL3544-R3549)).

I'll make a separate PR that fixes this minor bug, and rebase this PR on that.

govindkrjoshi (approved):

app/services/bulk_api_import/fhir_observation_importer.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  def import_blood_pressure
+    build_blood_pressure_attributes
+      .then { Api::V3::BloodPressureTransformer.from_request(_1) }
+      .then { merge_encounter_observation(:blood_pressures, _1) }
+  end
+
+  def import_blood_sugar
+    build_blood_sugar_attributes
+      .then { Api::V3::Transformer.from_request(_1) }
```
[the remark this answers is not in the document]
tfidfwastaken: Blood Sugar uses the base transformer only before merging: https://github.com/simpledotorg/simple-server/blob/d1516b69a59c3ebff54685d64c0db97190b065fe/app/controllers/api/v4/blood_sugars_controller.rb#L26

`BloodSugarTransformer` doesn’t have a `from_request` method

PR #5260, opened 2023-09-02
Create dhis2 indentifiers for ethiopia facilities

govindkrjoshi:
**Story card:** [sc-10341](https://app.shortcut.com/simpledotorg/story/10341/configure-all-facilities-in-production)

## Because

We need to configure the dhis org unit ids as facility business identifiers

Source: https://docs.google.com/spreadsheets/d/1FHbXkbXd8EMe0qqoSOzdMLic374vlWE3uFKRkn_cOew/edit#gid=0
## This addresses

- Adds a data migration that runs only in **Ethiopia production** and creates the necessary facility business identifiers
- It also fixes as typo in the rake task

## Test instructions
- Create facilities locally with the same ids present in Ethiopia production
```ruby
# run the following commands in a rails console
require 'factory_bot_rails'
def common_org; @common_org ||= Organization.find_or_create_by!(name: "Common Test Organization"); end
require './db/data/20230831064255_add_facility_business_identifiers_in_ethiopia.rb'
AddFacilityBusinessIdentifiersInEthiopia::FACILITY_IDENTIFIERS.keys.each { |id| FactoryBot.create(facility: id: id) }
```
- Run the data migration 
```bash
DEFAULT_COUNTRY=ET SIMPLE_SERVER_ENV=production rake data:migrate
``` 
- Verify the facility business identifiers are create correctly according to the source.

### DIFF
--- a/db/data/20230831064255_add_facility_business_identifiers_in_ethiopia.rb
+++ b/db/data/20230831064255_add_facility_business_identifiers_in_ethiopia.rb
@@ -1,0 +1,103 @@
+# frozen_string_literal: true
+
+class AddFacilityBusinessIdentifiersInEthiopia < ActiveRecord::Migration[6.1]
+  FACILITY_IDENTIFIERS = {
+    "20b9225f-c1be-4585-882f-45f6d8160174" => "HCIjxskLtAy",
+    "8e9098a3-ba71-438c-96b8-17a9b85a8feb" => "k0NDQTXTSkD",
+    "0ee397aa-66c4-42ea-879c-19378d66cbfa" => "hBM37JUG2mJ",
+    "2e722691-b75c-46e1-a960-43eb795a2408" => "iBzxf4Kbomn",
+    "3993765b-1ec6-48f4-aa49-bf3203c9cfc7" => "SZoLvBrQ4rJ",
+    "7c4eccd9-6b5d-4c54-ad59-e9d72e935462" => "dSLlwKmF6hP",
+    "97b674a3-7889-492e-903a-4e913628bebc" => "sBLijW98tkD",
+    "27288d0c-73ff-42d9-984c-f85148a0a44c" => "GIrO4Vy5AU8",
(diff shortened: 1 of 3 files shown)


db/data/20230831064255_add_facility_business_identifiers_in_ethiopia.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+    FACILITY_IDENTIFIERS.each do |simple_id, dhis2_org_id|
+      facility = Facility.find_by(id: simple_id)
+      unless facility.present?
+        print "Facility not found: id #{simple_id}"
+        next
+      end
+
+      facility.business_identifiers.find_by(
```
olttwa, line 96: This down command will delete records which were not created by this migration because of the `find_or_create_by` command. We can include a `created_at` > 4 Sep to avoid that, maybe.

db/data/20230831064255_add_facility_business_identifiers_in_ethiopia.rb
```diff
@@ -0,0 +1,102 @@
...
+        identifier_type: :dhis2_org_unit_id,
```
olttwa, line 97: Not a review comment, just a doubt. 

> What is difference between identifier types `dhis2_org_unit_id` & `external_org_facility_id`?

olttwa (commented):
No major blockers, just 1 comment & 1 doubt.

db/data/20230831064255_add_facility_business_identifiers_in_ethiopia.rb
govindkrjoshi: dhis2_org_unit_id are org_unit_ids of the corresponding facility. 

We added `external_org_facility_id` to track the facility ids from organizations using the import API.

db/data/20230831064255_add_facility_business_identifiers_in_ethiopia.rb
govindkrjoshi: I think it's fine since we are specifically deleting the records we are creating using both simple and dhis2 ids.

PR #5263, opened 2023-09-04
Overdue: Correct number of dashed line segments for past months

jamiecarter7:
**Story card:** [sc-11007](https://app.shortcut.com/simpledotorg/story/11007/overdue-dashed-lines-showing-incorrectly-on-previous-month-reports)

## Because

When looking back at previous months the dashboard, the dashed part of the overdue graphs does not reflect correctly.

## This addresses

When viewing previous dates the dashed segment should be shorter because we are viewing data that is complete. The dashed lines indicate data that is still changing.

An example: In Sept 2023 I can see that the overdue called chart has sept (the last segment) as dashed, but when viewing Aug-2023 data on the dashboard (using the dropdown) there should be no dashed segments on the chart because aug data is complete. 

This also occurs on the overdue returned to care chart.

This PR corrects both charts. The code now takes into account if a user is looking back through the history as shows the dashed segments correctly.

I've also done alot of refactoring to simplify the functions and improve the variable names.

## Test instructions

Load the dashboard, check how many dashed segments are present in the charts:
- 'Overdue called' = 1 dashed segment
- 'Overdue return to care' = 2 dashed segments

Change the view of the dashboard to august, this should reduce the number of dashed segments on each chart by 1:
- 'Overdue called' = 0 dashed segment
- 'Overdue return to care' = 1 dashed segments

Change the view of the dashboard to july,:
- 'Overdue called' = 0 dashed segment
- 'Overdue return to care' = 0 dashed segments

Test all remaining months:
- 'Overdue called' = 0 dashed segment
- 'Overdue return to care' = 0 dashed segments


app/assets/javascripts/common/reports.js
```diff
@@ -1693,15 +1695,25 @@ function baseBarChartConfig() {
...
-  const segmentStartIndex = ctx.p0DataIndex
-  return isSegmentDashed(segmentStartIndex, numberOfXAxisTicks, numberOfDashedSegments) ? dashStyle : undefined;
+
+  const endDateKeyIndex = monthIndexFromDateKey(endDateKey);
+  const dateNowMonth = new Date().getMonth()
+
+  const monthsInPast = dateNowMonth - endDateKeyIndex;
+  const dashedSegments = Math.max(0, monthsDashedBase - monthsInPast);
+
+  const segmentStartNodeIndex = ctx.p0DataIndex
```
PriyangaPKini, line 1708: What does `ctx.p0DataIndex` return?

app/assets/javascripts/common/reports.js
```diff
@@ -1693,15 +1695,25 @@ function baseBarChartConfig() {
 const dynamicChartSegementDashed = (
   ctx,
   numberOfXAxisTicks,
```
PriyangaPKini, line 1697: `numberOfXAxisTicks` does this return the number of months on the x-axis?

app/assets/javascripts/common/reports.js
```diff
@@ -1693,15 +1695,25 @@ function baseBarChartConfig() {
...
+  return isSegmentDashed(segmentStartNodeIndex, numberOfXAxisTicks, dashedSegments) 
+    ? [4, 3] // dashStyle
+    : undefined;
 };
 
-function isSegmentDashed(segmentStartIndex, numberOfXAxisTicks, segmentsToDashFromEnd) {
-  return segmentStartIndex >= numberOfXAxisTicks - (segmentsToDashFromEnd + 1)
+function isSegmentDashed(segmentStartNodeIndex, numberOfXAxisTicks, dashedSegments) {
+  const dataNodeToStartDashedLine = numberOfXAxisTicks - 1 - dashedSegments
+  return segmentStartNodeIndex >= dataNodeToStartDashedLine
```
PriyangaPKini, line 1716: Could you please explain what is happening in this line? I am feeling little difficult to understand the logic here.

PriyangaPKini (commented):
Left few comments

app/assets/javascripts/common/reports.js
jamiecarter7: Chartjs looks at each segment one by one starting on the left and moving to the right 
Each segment has a p0 and a p1 index which are the nodes at each end of the segment (left and right) 
Ctx.p0DataIndex Is the index 0 node the node on the left of the segment

app/assets/javascripts/common/reports.js
jamiecarter7: they are the same figure, but I wanted to make it generic incase the x axis is not month strings.
I used the data point with the labels (for the x axis) but I could have also used the actual data array for the line. I just need to get the number of data points in the chart for the x-axis

app/assets/javascripts/common/reports.js
jamiecarter7:
```function isSegmentDashed(segmentStartNodeIndex, numberOfXAxisTicks, dashedSegments) {
  const dataNodeToStartDashedLine = numberOfXAxisTicks - 1 - dashedSegments
  return segmentStartNodeIndex >= dataNodeToStartDashedLine
};```
This function is run on every segment of the line - a segment is between 2 data points.
`dataNodeToStartDashedLine` is the index of the point (data node) where the dashed line begins. For example the dashed line is for 2 segments this number will be 15 (0 index).
`segmentStartNodeIndex` is actually `p0` for the segment but I wanted to be explicit in name - maybe this can be changed

when the `dataNodeToStartDashedLine` is equal to or greater than the current 'p0' the line is dashed

PriyangaPKini (approved):
Calculation seems bit complex but I can't think of any other good solution to improve the readability right now

PR #5275, opened 2023-09-12
Disable users in migrated facilities

govindkrjoshi:
## Because

We have received instructions to disable users in all regions other than: 

- Chennai district
- West Bengal (all districts)
- Nagaland districts
- Pimpri Chinchwad Municipal Corporation
- Pune municipal corporation in Maharashtra

## This addresses

Adds a data migration that disables these users

## Test instructions

1. Copy the following tables locally: 
- Organizations
- Facility Groups
- Facilities
- Regions
- Users
- User authentications
- Phone number authentications

2. Very the number of users to be present before and after the migration. In a rails console, run: 
```
require './db/data/20230907045337_disable_users_in_migrated_facilities.rb'
regions = DisableUsersInMigratedFacilities::REGIONS_WITH_ONGOING_ACCESS.map { |detail| Region.find_by(detail) }
all_users_with_access = User.all.sync_approval_status_allowed
users_on_going_access = regions.map { |region| region.state_region? ? region.children.flat_map { |d| d.source.users.sync_approval_status_allowed } : region.source.users.sync_approval_status_allowed }.flatten

all_users_with_access.count
users_on_going_access.count
```

3. Run the migration using `rake data:migrate`

4. Verify that `all_users_with_access = users_on_going_access` and that users with on going access only belong to the above mentioned regions.

### DIFF
--- a/db/data/20230907045337_disable_users_in_migrated_facilities.rb
+++ b/db/data/20230907045337_disable_users_in_migrated_facilities.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class DisableUsersInMigratedFacilities < ActiveRecord::Migration[6.1]
+  REGIONS_WITH_ONGOING_ACCESS = [
+    {region_type: :state, slug: "west-bengal"},
+    {region_type: :state, slug: "nagaland"},
+    {region_type: :district, slug: "chennai"},
+    {region_type: :district, slug: "pimpri-chinchwad-municipal-corporation"},
+    {region_type: :district, slug: "pune-municipal-corporation"}
+  ]
+  def up
+    users_with_remaining_access = REGIONS_WITH_ONGOING_ACCESS.flat_map do |region_details|
+      region = Region.find_by(region_details)
+      next unless region
+
+      if region.district_region?
+        region.source.users.sync_approval_status_allowed
+      elsif region.state_region?
+        region.children.flat_map { |district| district.source.users.sync_approval_status_allowed }
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


db/data/20230907045337_disable_users_in_migrated_facilities.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      if region.district_region?
+        region.source.users.sync_approval_status_allowed
+      elsif region.state_region?
+        region.children.flat_map { |district| district.source.users.sync_approval_status_allowed }
+      end
+    end
+
+    all_allowed_users = User.all.sync_approval_status_allowed.pluck(:id)
+    disabled_users = all_allowed_users - users_with_remaining_access.map(&:id)
```
tfidfwastaken, line 28: [nit] could pluck again? not that important here though.
```suggestion
    disabled_users = all_allowed_users - users_with_remaining_access.pluck(:id)
```

govindkrjoshi: Can't use pluck here since `users_with_remaining_access` is an array and not an active record relation.

tfidfwastaken (approved):

db/data/20230907045337_disable_users_in_migrated_facilities.rb
tfidfwastaken: ActiveSupport defines `pluck` on all Enumerables these days :)
https://api.rubyonrails.org/classes/Enumerable.html#method-i-pluck

PR #5276, opened 2023-09-12
Raise error when bsnl jwt referesh fails

qptr:
Related story card: [sc-11112](https://app.shortcut.com/simpledotorg/story/11112/p1-bsnl-smses-failing-with-401)

The task to refresh bsnl token would fail silently when we were arbitrarily blacklisted. We want to an add alert to catch these issues with the api earlier on.


tfidfwastaken (approved):
LGT Me

PR #5281, opened 2023-09-22
Re-enable experiments in India

qptr:
**Story card:** [sc-11112](https://app.shortcut.com/simpledotorg/story/11112/p1-bsnl-smses-failing-with-401)

We temporarily disabled sms experiments in India due to a bug where smses were failing with HTTP 401. Now that the BSNL blacklist is lifted, we're enabling them again. Details on long term fixes are captred in the card.

### DIFF
--- a/lib/tasks/experiments.rake
+++ b/lib/tasks/experiments.rake
@@ -1,11 +1,9 @@
 namespace :experiments do
   desc "Conduct an experiment for the day and send out notifications"
   task conduct_daily: :environment do
-    unless CountryConfig.current_country?("India")
-      Time.use_zone(CountryConfig.current[:time_zone]) do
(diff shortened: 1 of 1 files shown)


qptr:
> Let's add a flipper flag so we don't have to do a deploy to disable the SMSes.
@govindkrjoshi Looks like there's already an :experiments flipper flag that is checked in both the [experimentation runner](https://github.com/simpledotorg/simple-server/blob/master/app/services/experimentation/runner.rb#L8) and [reminder scheduler](https://github.com/simpledotorg/simple-server/blob/master/app/jobs/appointment_notification/schedule_experiment_reminders.rb#L16). Missed this last time :/

olttwa (approved):

PR #5289, opened 2023-10-04
Initialize ET dynamic forms 1 month in advance

PriyangaPKini:
**Story card:** [sc-11191](https://app.shortcut.com/simpledotorg/story/11191/initialize-all-dynamic-form-months-for-et)

## Because

All the dynamic forms are initialised on first of every month. In Ethiopia, due to the difference in calendar, these forms appear around 20th of a month. 

## This addresses

Disclaimer: We are going for a hack to fix this because accommodating Ethiopian calendar in Simple will take up more than 2-3 months of work.

- We have decided to initialise the `current month` form in Ethiopia instead of `previous month` form from next month onwards. That is, On 1st Nov, we will be initialising the form for `Nov-2023` in Ethiopia and `Oct-2023` in other countries. 
- We will have to manually initialise `Oct-2023` form for Ethiopia to catch up with other countries.
---
- We have moved the initialisation code to a rake task
  - We had to pass in a different `date` argument to initialise forms in different countries. 
  - With `runner` taking a `string` was not practical in our case
## Test instructions

- Run the rake task locally and verify if the forms are being created
    - `bundle exec rake questionnaires:initiliaze`

### DIFF
--- a/app/services/questionnaire_responses/monthly_screening_reports.rb
+++ b/app/services/questionnaire_responses/monthly_screening_reports.rb
@@ -3,17 +3,17 @@ class QuestionnaireResponses::MonthlyScreeningReports
 
   def initialize(date = 1.month.ago)
     @month_date = date.beginning_of_month
-    @facility_reports = monthly_followup_and_registration
+    @facility_ids = Facility.select("id").pluck(:id)
     @questionnaire_id = latest_active_questionnaire_id(Questionnaire.questionnaire_types[:monthly_screening_reports])
   end
 
-  def pre_fill
-    @facility_reports.each do |facility_report|
-      unless monthly_screening_report_exists?(facility_report.facility_id)
+  def seed
+    @facility_ids.each do |facility_id|
+      unless monthly_screening_report_exists?(facility_id)
         QuestionnaireResponse.create!(
           questionnaire_id: @questionnaire_id,
-          facility_id: facility_report.facility_id,
-          content: prefilled_responses(month_date_str, facility_report),
+          facility_id: facility_id,
+          content: monthly_reports_base_content,
           device_created_at: @month_date,
           device_updated_at: @month_date
         )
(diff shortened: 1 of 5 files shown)


lib/tasks/questionnaires.rake
```diff
@@ -0,0 +1,16 @@
+namespace :questionnaires do
+  desc "Initialize questionnaire responses"
+  task initialize: :environment do
+    date = CountryConfig.current_country?("Ethiopia") ? Date.current : 1.month.ago
```
govindkrjoshi, line 4: Shouldn't this be one month later? 

In the current month, I want to initialize the questionnaire for the next month?

govindkrjoshi: For future, wondering if the initialization config should be a part of the questionnaire itself.

olttwa: > In the current month, I want to initialize the questionnaire for the next month?

@govindkrjoshi For all countries, we initialize questionnaire for previous month in current month. That leads to a delay of 20-26 days in Ethiopia. Initializing for current month negates that and users will start seeing the forms 5-10 days in advance depending on the month.

olttwa: > wondering if the initialization config should be a part of the questionnaire itself.

Good suggestion. This logic can/must be moved from code to data. We need 2 pieces of information: `when to initialize` and `for which month-date to initialze`. This can be represented as such
```ruby
# When: Every 1 month at 6 AM
# month-date = -1/0/1.month.from_now
questionnaire.init_config = {when: "0 6 1 * *", relative_month_date: 0}```

However, I'm not sure how to generate a config during runtime that `whenever` gem understands and puts in crontab. Will need to spike that unknown part.

File: app/services/questionnaire_responses/monthly_screening_reports.rb [hunk 2]
```diff
@@ -3,17 +3,17 @@ class QuestionnaireResponses::MonthlyScreeningReports
 
   def initialize(date = 1.month.ago)
     @month_date = date.beginning_of_month
-    @facility_reports = monthly_followup_and_registration
+    @facility_ids = Facility.select("id").pluck(:id)
```
olttwa, line 6: We moved from iterating over `facility_reports` to `facilities` because it's possible mat-view refreshes wouldn't have completed and generated reports for the advanced month for which we're initializing dynamic forms.

app/services/questionnaire_responses/monthly_screening_reports.rb
```diff
@@ -3,17 +3,17 @@ class QuestionnaireResponses::MonthlyScreeningReports
...
     @questionnaire_id = latest_active_questionnaire_id(Questionnaire.questionnaire_types[:monthly_screening_reports])
   end
 
-  def pre_fill
-    @facility_reports.each do |facility_report|
-      unless monthly_screening_report_exists?(facility_report.facility_id)
+  def seed
```
olttwa, line 10: We're just seeding(initializing) forms instead of pre-filling data, because India doesn't have screening forms and it isn't feasible to pre-fill data in Ethiopia's forms due to new logic of advanced initialization.

PR #5290, opened 2023-10-05
Change overdue dashboard order

jamiecarter7:
**Story card:** [sc-11217](https://app.shortcut.com/simpledotorg/story/11217/change-placement-of-overdue-section-in-dashboard)

## Because

A request from the India team to move the overdue section further down on the dashboard, below other more important cards.

## This addresses

This actions the request. The overdue section is place under the 'COMPARE PERFORMANCE' section

## Test instructions

Load the dashboard and check that the overdue section is now positioned below the compare performance section.


PriyangaPKini (approved):
LGTM

PR #5291, opened 2023-10-05
Fix dates in ltfu graph

PriyangaPKini:
**Story card:** [sc-11279](https://app.shortcut.com/simpledotorg/story/11279/change-dates-in-ltfu-chart)

## Because

The dates in the LTFU graph is incorrect.

## This addresses

Use the correct LTFU start date and LTFU end date in the graph.
For eg. If current month is OCT-2023, then start date should be "30-SEP-2023" and end date should be "31_OCt-2023"

*After*:
image

*Before*
image


jamiecarter7 (approved):
LGTM

PR #5298, opened 2023-10-11
Remove india prod capistrano deployment config

roypeter:
**Story card:** [sc-11227](https://app.shortcut.com/simpledotorg/story/11227)

## Because

India production environment is migrated to the new AWS EKS environment


qptr:
Do we want to update the README too? It can siimply point to the container-deployment instructions.

qptr (commented):
Can we update the README too? It can simply point to the container-deployment instructions.

PR #5303, opened 2023-10-16
import api: validate incorrect observation codes

tfidfwastaken:
# Why

Our OpenAPI schema validator is not flexible enough to capture some edge cases for observation resources–namely, it is possible to pass in a blood pressure observation, with the systolic or diastolic code specified twice. This will cause a failure at the database constraint level. We instead validate this at the time of a request so that the user is immediately informed of this error, and the request is not propagated any further.

# The fix

We add new validations in our Imports API validator class. Any observation resource must satisfy either one of these conditions:
* It contains a single component that is a valid Blood Sugar measurement LOINC code.
* It contains two components, one of which is a valid systolic BP measurement and the other is a valid diastolic BP measurement.

### DIFF
--- a/app/services/bulk_api_import/validator.rb
+++ b/app/services/bulk_api_import/validator.rb
@@ -6,8 +6,13 @@ class BulkApiImport::Validator
 
   def validate
     error = validate_schema
+
     unless error.present?
       error = validate_facilities
+    end
+
+    unless error.present?
+      error = validate_observation_codes
     end
 
     error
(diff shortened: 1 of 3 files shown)


app/services/bulk_api_import/validator.rb
```diff
@@ -21,6 +26,36 @@ def validate_schema
...
+            For blood pressure observation: Ensure that both systolic and diastolic measurements are present with the correct codes.
+            For blood sugar observations: Ensure that only one code is present and matching the list of accepted codes."
+          ERRSTRING
+        }
+      }
+    end
+  end
+
+  def valid_blood_pressure_codes?(codes)
+    codes.to_set == Api::V4::Imports::ALLOWED_BP_CODES.to_set
```
govindkrjoshi, line 52: Why is `codes.count == 1` not necessary here?

tfidfwastaken: A blood pressure observation has two codes associated with it: systolic and diastolic ie `Set{8480-6, 8462-4}`.
A blood sugar observation on the other hand only takes one measurement at a time, ie, either RBS or FBS/PPBS or Hb1ac.

I could have put `codes.count == 2` on this line, but that is already implicitly taken care of by the set.

tfidfwastaken:
I have managed to narrow down the cause of the test failure. It's because of a bad interaction between `bangladesh_disaggregated_dhis2_exporter_spec.rb` and `ethiopia_exporter_spec.rb` (could reproduce this on `origin/master`). It's only failed in this PR because I have once again become the victim of bad luck/parallel test roulette.

@govindkrjoshi If I don't fix that spec failure by tomorrow, can we consider merging this anyway to unblock the Bangladesh work?

govindkrjoshi (approved):

PR #5307, opened 2023-10-18
Cancel outdated experiments and create new experiments for India

olttwa:
**Story card:** [sc-11186](https://app.shortcut.com/simpledotorg/story/11186/bsnl-bug-re-enable-sms-experiments-in-india)

## Because

Migrating to new deployment environment in India had caused pending notifications to cause a thundering herd, leading to blacklisting from BSNL.

## This addresses

Cancels existing experiments and their notifications, evicts patients in those experiments and creates new experiments for month of Nov & Dec.

## Test instructions

Uncomment country check and run `rake data:migrate`. Expect 4 experiments to be created (2 current & 2 stale for months of Nov/Dec)


db/data/20231018062055_re_enable_sms_india.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+class ReEnableSmsIndia < ActiveRecord::Migration[6.1]
```
qptr, line 3:
```suggestion
class ReduceSmsBatchSizeIndia < ActiveRecord::Migration[6.1]
```

db/data/20231018062055_re_enable_sms_india.rb
```diff
@@ -0,0 +1,74 @@
...
+  PATIENTS_PER_DAY = 5000
+  FILTERS = {
+    "states" => {"include" => ["Maharashtra", "Andhra Pradesh", "West Bengal"]}
+  }.freeze
+  CANCELLED_EXPERIMENTS_MONTHS = (9..12).map do |month_number|
+    Date::ABBR_MONTHNAMES[month_number]
+  end
+
+  NEW_EXPERIMENTS_DATA = (11..12).map do |month_number|
```
qptr, line 12: Shall we add another experiment for the last 10 days of October?

db/data/20231018062055_re_enable_sms_india.rb
```diff
@@ -0,0 +1,74 @@
...
+      stale_patients_experiment_name: "Stale Patient V2 #{month} 2023"
+    }
+  end
+
+  def up
+    return unless CountryConfig.current_country?("India") && SimpleServer.env.production?
+
+    CANCELLED_EXPERIMENTS_MONTHS.each do |month|
+      outdated_current_experiment = Experimentation::CurrentPatientExperiment.find_by_name("Current Patient #{month} 2023")
+      outdated_current_experiment.cancel
```
qptr, line 27: Have we tested if `cancel` works as expected on an experiment that has already ended? The September experiment has ended, but has pending notifications.

db/data/20231018062055_re_enable_sms_india.rb
```diff
@@ -0,0 +1,74 @@
...
+      outdated_current_experiment.evict_patients
```
qptr, line 28: This isn't necessary since `evict_patients` only conditionally evicts treatment group memberships. The patients in these experiments will also be available to the next experiment considering the amount of time between them.

db/data/20231018062055_re_enable_sms_india.rb
```diff
@@ -0,0 +1,74 @@
...
+      outdated_stale_experiment.evict_patients
+    end
+
+    NEW_EXPERIMENTS_DATA.map do |experiment_data|
+      ActiveRecord::Base.transaction do
+        Experimentation::Experiment.current_patients.create!(
+          name: experiment_data[:current_patients_experiment_name],
+          start_time: experiment_data[:start_time],
+          end_time: experiment_data[:end_time],
+          max_patients_per_day: PATIENTS_PER_DAY,
```
qptr, line 41: Do we want to keep the patient count at 5000 for both experiments? What is the plan for further experiments?

qptr (changes requested):

db/data/20231018062055_re_enable_sms_india.rb
olttwa: I'm rewording it to SetupSMSExperimentsInIndia, since this data migration now has new filters to update.

db/data/20231018062055_re_enable_sms_india.rb
olttwa: Cancelling an experiment will mark all pending notifications as cancelled. We need not worry about pending notifications. Details:

> When cancelling an experiment, `notifications.cancel` gets called. 
Before calling `NotificationDispatchService.call` to send SMS, `AppointmentNotification::Worker` checks if status of Notification is scheduled. Since the status would be `cancelled`, no SMS will be sent.

db/data/20231018062055_re_enable_sms_india.rb
olttwa: > In India production, for a given day, eligible patient count is in the range of 10k for Current patients
Query for Stale patients did not complete in more than 20 mins, maybe due to load, should run fine in morning hours.
A max patients config of 20k will do for now.

PR #5310, opened 2023-10-20
Split DHIS2 exports into manageable units and retry failures

qptr:
**Story card:** [sc-11298](https://app.shortcut.com/simpledotorg/story/11298)

This PR aims to make the monthly exports of indicator data to country-specific Dhis2 instances more robust by breaking them up into retryable jobs that export data per facility identifier instead of all at once.

It deletes the old exporter files and creates job files and their corresponding specs. The jobs for each identifier are created in the rake task now.


app/jobs/dhis2/bangladesh_exporter_job.rb
```diff
@@ -0,0 +1,23 @@
+module Dhis2
+  class BangladeshExporterJob < Dhis2ExporterJob
+    private
+
+    def facility_data_for_period(region, period)
+      {
+        htn_cumulative_assigned: PatientStates::Hypertension::CumulativeAssignedPatientsQuery.new(region, period).call.count,
```
qptr, line 7: @govindkrjoshi Thinking of changing all of these query methods into class methods because they're a pain to test and also don't seem necessary. What does it mean to have a CumulativeAssignedPatientsQuery instance

PriyangaPKini (approved):
LGTM

app/jobs/dhis2/bangladesh_disaggregated_exporter_job.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Dhis2
+  class BangladeshDisaggregatedExporterJob < Dhis2ExporterJob
+    STEP = 5
+    BUCKETS = (15..75).step(STEP).to_a
```
danySam, line 6: What's the relevance of 15 and 75?

qptr: They're the age range across which the age buckets are split. I'll put them in constants as MIN_AGE and MAX_AGE for readability

PR #5313, opened 2023-10-27
Allow duplication of business identifiers for every org

tfidfwastaken:
**Story card:** sc-11347

An organization that is using the import API is forced to ensure that
their identifiers (for facilities and patients) do not clash with any
identifier chosen by other organizations. This is a bug we did not
intend to introduce, and was caused by us failing to include the
organization ID in our UUID hash when deriving our own resource IDs.

Thus our translation from partner org ID to Simple ID previously looked
like:

    UUIDv5(their_resource_ID) => Simple Resource ID

The issue is that if some Org1 has a Patient with ID of 1, and Org2 has
a Patient with ID of 2, they will be translated to the same UUID value
in Simple.

To fix this, we now have:

    UUIDv5(their_org_ID + their_resource_ID) => Simple Resource ID

which was the intended design.

### DIFF
--- a/app/services/bulk_api_import/fhir_importable.rb
+++ b/app/services/bulk_api_import/fhir_importable.rb
@@ -14,11 +14,24 @@ module BulkApiImport::FhirImportable
     }
   end
 
-  def translate_id(id, ns_prefix: "")
-    Digest::UUID.uuid_v5(Digest::UUID::DNS_NAMESPACE + ns_prefix, id)
+  def translate_id(id, org_id:, ns_prefix: "")
+    Digest::UUID.uuid_v5(Digest::UUID::DNS_NAMESPACE + org_id + ns_prefix, id)
   end
 
-  def translate_facility_id(id)
-    FacilityBusinessIdentifier.find_by(identifier: id, identifier_type: :external_org_facility_id).facility.id
+  def translate_facility_id(id, org_id:)
+    FacilityBusinessIdentifier
+      .joins(facility: :facility_group)
+      .where(identifier_type: :external_org_facility_id,
+        facility_business_identifiers: {identifier: id},
+        facility_groups: {organization_id: org_id})
+      .take.facility.id
+  end
+
+  def translate_patient_id(id, org_id:)
+    translate_id(
+      id,
+      org_id: org_id,
+      ns_prefix: "patient_business_identifier"
+    )
   end
 end
--- a/app/services/bulk_api_import/fhir_importable.rb
+++ b/app/services/bulk_api_import/fhir_importable.rb
@@ -21,10 +21,10 @@ module BulkApiImport::FhirImportable
   def translate_facility_id(id, org_id:)
     FacilityBusinessIdentifier
       .joins(facility: :facility_group)
-      .where(identifier_type: :external_org_facility_id,
+      .find_by(identifier_type: :external_org_facility_id,
         facility_business_identifiers: {identifier: id},
         facility_groups: {organization_id: org_id})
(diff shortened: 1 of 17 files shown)


app/services/bulk_api_import/fhir_importable.rb
```diff
@@ -14,11 +14,24 @@ def timestamps
...
+      .take.facility.id
+  end
+
+  def translate_patient_id(id, org_id:)
+    translate_id(
+      id,
+      org_id: org_id,
+      ns_prefix: "patient_business_identifier"
+    )
   end
```
tfidfwastaken, line 36: Added a new helper, since this particular `ns_prefix` found a lot of reuse across the FHIR resources.

spec/jobs/bulk_api_import_job_spec.rb
```diff
@@ -1,15 +1,15 @@
 require "rails_helper"
 
-RSpec.describe ImportFacilitiesJob do
+RSpec.describe BulkApiImportJob do
```
tfidfwastaken, line 3: Fixed this misnomer caused by copypasta in some old commit.

spec/services/bulk_api_import/fhir_condition_importer_spec.rb
```diff
@@ -3,8 +3,13 @@
...
   before { create(:facility) }
   let(:import_user) { ImportUser.find_or_create }
+  let(:org_id) { import_user.organization_id }
   let(:identifier) { SecureRandom.uuid }
-  let(:patient) { build_stubbed(:patient, id: Digest::UUID.uuid_v5(Digest::UUID::DNS_NAMESPACE, identifier)) }
+  let(:patient) do
+    build_stubbed(:patient, id: Digest::UUID.uuid_v5(
+      Digest::UUID::DNS_NAMESPACE + org_id + "patient_business_identifier", identifier
+    ))
+  end
```
tfidfwastaken, line 12: One weak point in our design is that we must vigilantly ensure that resource creation and resource lookup are in sync. If someone were to change the way we map an ID on resource creation in the future, the subsequent lookups from various importers are going to fail, and this unit test cannot catch this.

The only way to ensure that we are in sync is to write an integration test suite that creates a resource using the API and then subsequently updates it and asserts on the new state. Right now I have left such a test out because it would require us to spin up sidekiq or stub it with a functional equivalent that actually persists data—I felt that this is not yet worth the effort for such a nascent API. Moreover, this issue is slightly mitigated by the fact that ID creation and ID lookup are reusing the same method, so we are protected against the modification of the mapping implementation itself.

app/services/bulk_api_import/fhir_importable.rb
```diff
@@ -14,11 +14,24 @@ def timestamps
...
+  def translate_id(id, org_id:, ns_prefix: "")
+    Digest::UUID.uuid_v5(Digest::UUID::DNS_NAMESPACE + org_id + ns_prefix, id)
   end
 
-  def translate_facility_id(id)
-    FacilityBusinessIdentifier.find_by(identifier: id, identifier_type: :external_org_facility_id).facility.id
+  def translate_facility_id(id, org_id:)
+    FacilityBusinessIdentifier
+      .joins(facility: :facility_group)
+      .where(identifier_type: :external_org_facility_id,
```
govindkrjoshi, line 24:
```suggestion
      .where(identifier_type: :external_org_facility_id,
      .find_by(identifier_type: :external_org_facility_id, 
```

govindkrjoshi (approved):

PR #5314, opened 2023-10-30
Add options to configure Sidekiq log level

roypeter:
**Story card:** [sc-11295](https://app.shortcut.com/simpledotorg/story/11295)

## Because

The Sidekiq log volume is huge due to the info log level

## This addresses

We've set the Sidekiq log level to 'error' in order to reduce the log volume

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -48,6 +48,8 @@ Sidekiq.configure_server do |config|
   SidekiqUniqueJobs::Server.configure(config)
 end
 
+Sidekiq.logger.level = Rails.logger.level
+
 Sidekiq::Throttled.setup!
 SidekiqUniqueJobs.configure do |config|
   config.enabled = !Rails.env.test?


config/initializers/sidekiq.rb
```diff
@@ -48,6 +48,8 @@ def self.connection_pool
   SidekiqUniqueJobs::Server.configure(config)
 end
 
+Sidekiq.logger.level = Rails.logger.level
```
PriyangaPKini, line 51: Currently, [Rails.logger.level](https://github.com/simpledotorg/simple-server/blob/724b90920c2b58d556ee5b6260455be02af51cb3/config/environments/production.rb#L55) is `info` level right? How will this set `error` log in Sidekiq?

PriyangaPKini (changes requested):

config/initializers/sidekiq.rb
olttwa: @PriyangaPKini the syntax you pointed to defaults to `info` in case env variable is missing. In production environment, it is set to `error`. @roypeter PR is good to for merging!

olttwa (approved):

config/initializers/sidekiq.rb
PriyangaPKini: Thanks for clarifying 👍

PR #5317, opened 2023-10-31
Import API: Validate existence of patients

tfidfwastaken:
**Story card:** [sc-11348]

We previously allowed dangling references to patients from its associated resources like appointments to allow integrators of the API to send resources in any order they please. But experience with testing the API on a sandbox has made it clear that the problem of dangling patient references is more pressing than we initially assumed. This change ensures that any dependent resource references a patient that already exists in the Simple database. It returns an error similar to the one about unmapped facility if this patient is not found.

### DIFF
--- a/spec/requests/oauth_spec.rb
+++ b/spec/requests/oauth_spec.rb
@@ -3,6 +3,7 @@ require "rails_helper"
 RSpec.describe "OAuth Credentials", type: :request do
   context "import API" do
     before { Flipper.enable(:imports_api) }
+
     context "when unauthorized" do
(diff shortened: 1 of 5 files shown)


spec/requests/oauth_spec.rb
```diff
@@ -40,6 +45,9 @@
                     HTTP_X_ORGANIZATION_ID: organization.id},
           params: {resources: [resource]}.to_json
 
+        puts Patient.first.to_json
```
govindkrjoshi, line 48: Unnecessary puts

PR #5320, opened 2023-11-06
Import API: Fix incorrect patient IDs in patient-associated resources

tfidfwastaken:
The patient identifiers were created incorrectly in the dependent resources. We accidentally added a namespace prefix that's used to identify the patient business identifier, rather than the patient itself (which does not use any kind of prefix).

While refactoring the `translate_patient_id` method, it became clear that this utility had no usage across multiple FHIR resource importers, so we removed it and made the mixin interface smaller.

Our specs also need rectification, they used the wrong patient creation logic.

Since the business identifiers were created correctly, but the resources associated with those patients were not, we add a data migration to rectify the situation.

It turns out that we have been creating these dependent resources with the `patient_id` set to the `id` of the business identifier, instead of the `patient_id` of the business identifier. This migration swaps out the `patient_id` of the dependent resource.

### DIFF
--- a/app/services/bulk_api_import/fhir_condition_importer.rb
+++ b/app/services/bulk_api_import/fhir_condition_importer.rb
@@ -27,7 +27,7 @@ class BulkApiImport::FhirConditionImporter
   def build_attributes
     {
       id: translate_id(@resource.dig(:identifier, 0, :value), org_id: @organization_id),
-      patient_id: translate_patient_id(@resource[:subject][:identifier], org_id: @organization_id),
+      patient_id: translate_id(@resource[:subject][:identifier], org_id: @organization_id),
       prior_heart_attack: "unknown",
       prior_stroke: "unknown",
       chronic_kidney_disease: "unknown",
--- a/app/services/bulk_api_import/fhir_importable.rb
+++ b/app/services/bulk_api_import/fhir_importable.rb
@@ -26,12 +26,4 @@ module BulkApiImport::FhirImportable
         facility_groups: {organization_id: org_id})
       .facility.id
   end
-
-  def translate_patient_id(id, org_id:)
-    translate_id(
-      id,
-      org_id: org_id,
-      ns_prefix: "patient_business_identifier"
-    )
-  end
 end
--- a/app/services/bulk_api_import/fhir_patient_importer.rb
+++ b/app/services/bulk_api_import/fhir_patient_importer.rb
@@ -102,7 +102,7 @@ class BulkApiImport::FhirPatientImporter
   def business_identifiers
     [
       {
-        id: translate_patient_id(identifier, org_id: @organization_id),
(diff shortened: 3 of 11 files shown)


app/services/bulk_api_import/fhir_importable.rb
```diff
@@ -26,12 +26,4 @@ def translate_facility_id(id, org_id:)
...
       .facility.id
   end
-
-  def translate_patient_id(id, org_id:)
-    translate_id(
-      id,
-      org_id: org_id,
-      ns_prefix: "patient_business_identifier"
-    )
-  end
```
tfidfwastaken, line 36: This method ended up having only one usage, during patient creation, so I inlined it there and removed it from this mixin.

app/services/bulk_api_import/fhir_patient_importer.rb
```diff
@@ -102,7 +102,7 @@ def address
   def business_identifiers
     [
       {
-        id: translate_patient_id(identifier, org_id: @organization_id),
+        id: translate_id(identifier, org_id: @organization_id, ns_prefix: "patient_business_identifier"),
```
tfidfwastaken, line 105: Inlined the method here. No change in functionality.

app/services/bulk_api_import/fhir_condition_importer.rb
```diff
@@ -27,7 +27,7 @@ def metadata
   def build_attributes
     {
       id: translate_id(@resource.dig(:identifier, 0, :value), org_id: @organization_id),
-      patient_id: translate_patient_id(@resource[:subject][:identifier], org_id: @organization_id),
+      patient_id: translate_id(@resource[:subject][:identifier], org_id: @organization_id),
```
tfidfwastaken, line 30: Here and elsewhere: We want to translate the patient_id to the ID of the Patient and not the ID of its PatientBusinessIdentifier. This was the source of the bug.

db/data/20231106123819_fix_incorrect_fhir_translations.rb
tfidfwastaken, line 1: The error in production was that for every patient-associated resource, we set the `patient_id` to the primary key of `PatientBusinessIdentifier` when it should have been set to the `patient_id` of `PatientBusinessIdentifier`.

In this data migration, we edit the `patient_id` of each patient-associated resource to point to the `patient_id` attribute of the business identifier instead.

spec/services/bulk_api_import/fhir_importable_spec.rb
```diff
@@ -53,8 +51,7 @@
     end
 
     it "translates the patient ID correctly" do
-      expect(Object.new.extend(described_class)
-                   .translate_patient_id(patient_business_identifier.identifier, org_id: org_id))
+      expect(Object.new.extend(described_class).translate_id(patient_business_identifier.identifier, org_id: org_id))
```
tfidfwastaken, line 54: All the assertion changes in this PR are just formatting changes.

PR #5324, opened 2023-11-10
Set up appointment reminders experiment in Bangladesh for December

qptr:
**Story card:** [sc-11379](https://app.shortcut.com/simpledotorg/story/11379)

## Because

We want to conduct another experiment with appointment reminder templates in Bangladesh

## This addresses

This PR adds a data migration file to set up the experiment for the month of December with the requirements given in the story card.

It also modifies the region filter in NotificationExperiment to filter patients by the assigned facility's slugs instead of ids.

## Test instructions

To be added

### DIFF
--- a/app/models/experimentation/notifications_experiment.rb
+++ b/app/models/experimentation/notifications_experiment.rb
@@ -84,13 +84,13 @@ module Experimentation
       end
     end
 
-    def self.filter_facilities(patients, facilities)
-      return patients unless facilities
+    def self.filter_facilities(patients, facility_slugs)
+      return patients unless facility_slugs
 
-      if facilities["include"]
-        patients.where('"patients"."assigned_facility_id" IN (?)', facilities["include"])
-      elsif facilities["exclude"]
-        patients.where('"patients"."assigned_facility_id" NOT IN (?)', facilities["exclude"])
+      if facility_slugs["include"]
+        patients.where('"facilities"."slug" IN (?)', facility_slugs["include"])
+      elsif facility_slugs["exclude"]
+        patients.where('"facilities"."slug" NOT IN (?)', facility_slugs["exclude"])
       else
         patients
       end
--- a/db/data/20231109143922_set_up_bangladesh_december_experiment.rb
+++ b/db/data/20231109143922_set_up_bangladesh_december_experiment.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class SetUpBangladeshDecemberExperiment < ActiveRecord::Migration[6.1]
+  def up
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/db/data/20231109143922_set_up_bangladesh_december_experiment.rb
+++ b/db/data/20231109143922_set_up_bangladesh_december_experiment.rb
@@ -1,10 +1,102 @@
 # frozen_string_literal: true
 
 class SetUpBangladeshDecemberExperiment < ActiveRecord::Migration[6.1]
+  START_TIME = DateTime.parse("1 Dec 2023").beginning_of_day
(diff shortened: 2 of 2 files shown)


app/models/experimentation/notifications_experiment.rb
```diff
@@ -84,13 +84,13 @@ def self.filter_blocks(patients, blocks)
       end
     end
```
olttwa, line 85: You will have to update validate_filters in experiment.rb code to use facility_slugs instead of facilities?

olttwa (changes requested):

app/models/experimentation/notifications_experiment.rb
qptr: Oops, thanks missed checking that. It doesn't seem necessary with the current naming convention (since they all use different fields). But it will be clear what is expected. Shall I also change states and blocks to `state_names` and `block_ids`?

olttwa:
@qptr I just realized, if we are making this change, we'll have to run a data migration to update filters in India/LK environment to use `facility_slugs` instead of `facilities`. If you choose to rename states and blocks as well, the data migration will have to update those fields too.

olttwa:
@qptr I just realized, if we are making this change, we'll have to run a data migration to update filters in India/LK environment to use `facility_slugs` instead of `facilities`. If you choose to rename states and blocks as well, the data migration will have to update those fields too.

qptr:
> @qptr I just realized, if we are making this change, we'll have to run a data migration to update filters in India/LK environment to use `facility_slugs` instead of `facilities`. If you choose to rename states and blocks as well, the data migration will have to update those fields too.

Right. I was also thinking of leaving it as is until we refactor all region filters to use regions and slugs. Leaving this as is then 👍

db/data/20231109143922_set_up_bangladesh_december_experiment.rb
```diff
@@ -0,0 +1,101 @@
...
+    chattogram_facilities: ["uhc-hathazari", "uhc-fatikchori", "uhc-anwara", "uhc-satkania", "uhc-lohagara", "uhc-patiya", "uhc-rangunia", "uhc-sitakunda", "uhc-chandanaish", "uhc-raozan", "uhc-banshkhali", "uhc-boalkhali", "uhc-karnaphuli", "uhc-mirsharai", "uhc-sandwip"],
+    feni_facilities: ["uhc-fulgazi", "uhc-parsuram", "uhc-daganbhuiya", "uhc-sonagazi", "uhc-chhagalniya"],
+    netrokona_facilities: ["uhc-atpara", "uhc-barhatta", "uhc-kalmakanda", "uhc-kendua", "uhc-khaliajuri", "uhc-madan", "uhc-mohanganj", "uhc-purbadhala", "uhc-durgapur-dbd6efea-8950-410e-abc3-ed9a329fb9d3"],
+    sirajganj_facilities: ["uhc-belkuchi", "uhc-chowhali", "uhc-kamarkhanda", "uhc-kazipur", "uhc-shahzadpur", "uhc-tarash", "uhc-ullapara"],
+    pabna_facilities: ["uhc-bera", "uhc-bhangura", "uhc-faridpur", "uhc-iswardi", "uhc-santhia", "uhc-chatmohar", "uhc-sujanagar-4b973cb1-d584-4100-91ca-778f218b502e", "atgharia-upazila-health-complex"],
+    jamalpur_facilities: ["uhc-islampur", "uhc-melandah", "uhc-sarishabari", "uhc-dewanganj", "uhc-bakshiganj", "uhc-madarganj"],
+    sherpur_facilities: ["uhc-jhenaigati", "uhc-nalitabari", "uhc-nakhla", "uhc-sribordi"]
+  }.values.reduce(:+)
```
olttwa, line 22: This is good way of representing data in the expected format, and having readability. Better than using inline comments. Good one!

PR #5330, opened 2023-11-22
Hotfix: Handle exception and log debug info while recording a visit on a treatment group membership

qptr:
**Story card:** [sc-11504](https://app.shortcut.com/simpledotorg/story/11504)

This is a hotfix for the bug where, when recording a visit on a treatment
group membership, a NoMethodError is raised when the facility
association on a visit doesn't exist for whatever reason. Details of the preliminary investigation are captured
in the card.

This fix handles the exception and logs some debugging information.

### DIFF
--- a/app/models/experimentation/treatment_group_membership.rb
+++ b/app/models/experimentation/treatment_group_membership.rb
@@ -63,7 +63,7 @@ module Experimentation
         **visit_status_fields
       )
     rescue NoMethodError
-      Rails.logger.error("Error recording visit: Membership id: #{self.id} Blood pressure: #{blood_pressure&.id} Blood sugar: #{blood_sugar&.id} Prescription drug: #{prescription_drug&.id}")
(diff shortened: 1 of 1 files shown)


app/models/experimentation/treatment_group_membership.rb
```diff
@@ -62,6 +62,8 @@ def record_visit(blood_pressure:, blood_sugar:, prescription_drug:)
         days_to_visit: days_to_visit,
         **visit_status_fields
       )
+    rescue NoMethodError
```
olttwa, line 65: Begin/Rescue work when within a blog. You'll need to add a begin block at line 43 in order for the rescue to work.

qptr: Can't find the documentation for some reason but this works as well. Other places in the code with this style: 
https://github.com/simpledotorg/simple-server/blob/master/app/controllers/admin/facility_groups_controller.rb#L69
https://github.com/simpledotorg/simple-server/blob/master/app/models/reports/drug_stock_calculation.rb#L25

PR #5332, opened 2023-11-27
Add logging to the Imports Controller

tfidfwastaken:
**Story card:** [sc-11514]

What we are logging:
* Failure modes from the Import Controller layer. Validation errors are logged.

Why we aren't logging the request payload:
* Not sure if we should, given it's PII. Can make a debug log if needed.
* Successful merges can be verified with the audit log.

### DIFF
--- a/spec/api/v4/imports/imports_spec.rb
+++ b/spec/api/v4/imports/imports_spec.rb
@@ -46,6 +46,13 @@ describe "Import v4 API", swagger_doc: "v4/import.json" do
         let(:import_request) { {"resources" => [{"invalid" => "invalid"}]} }
         run_test!
       end
+
+      response "403", "Forbidden" do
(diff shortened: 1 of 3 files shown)


spec/api/v4/imports/imports_spec.rb
```diff
@@ -46,6 +46,13 @@
         let(:import_request) { {"resources" => [{"invalid" => "invalid"}]} }
         run_test!
       end
+
+      response "403", "Forbidden" do
+        let(:HTTP_X_ORGANIZATION_ID) { "wrong_org" }
+        let(:Authorization) { "Bearer #{token.token}" }
+        let(:import_request) { {"resources" => [{"doesn't" => "matter"}]} }
+        run_test!
+      end
```
tfidfwastaken, line 55: Also not related to the PR, but noticed this edge case was missing in the tests, so I added that.

PR #5335, opened 2023-11-29
FHIR export

govindkrjoshi:
This PR adds exporters to convert the following Simple resources into FHIR-compliant resources:
Appointment, blood pressure, blood sugar, patient, prescription drug and medical history

It also adds a rake task to export patients and all their related resources with PII removed to a json file


app/services/one_off/fhir/patient_exporter.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def export
+    FHIR::Patient.new(
+      identifier: patient_identifiers,
+      name: [
+        FHIR::HumanName.new(
+          text: patient.full_name
+        )
+      ],
+      active: patient.status == "active",
```
tfidfwastaken, line 34:
```suggestion
      active: patient.status_active?,
```

app/services/one_off/fhir/patient_exporter.rb
```diff
@@ -0,0 +1,82 @@
...
+      gender: gender,
+      birthDate: birth_date.iso8601,
+      deceasedBoolean: patient.status == "dead",
```
tfidfwastaken, line 37:
```suggestion
      deceasedBoolean: patient.status_dead?,
```

app/services/one_off/fhir/appointment_exporter.rb
```diff
@@ -0,0 +1,57 @@
...
+          cancellationReason: cancellation_code,
+          participant: [
+            FHIR::Appointment::Participant.new(
+              actor: FHIR::Reference.new(
+                id: FHIR::Patient.new(
+                  identifier: FHIR::Identifier.new(
+                    value: appointment.patient_id
+                  )
+                )
+              )
```
tfidfwastaken, line 25: [question] Can you verify if this is the correct way to build a reference?

The library is unfortunately sparsely documented, but the source code suggests that a hash is expected here, not a reference to another model?
https://github.com/fhir-crucible/fhir_models/blob/be9087e8fc0e49dc93f504d18ceb76cbb32b3811/lib/fhir_models/bootstrap/model.rb#L9-L18

tfidfwastaken: This comment applies to every reference in this patch.

app/services/one_off/fhir/prescription_drug_exporter.rb
```diff
@@ -0,0 +1,118 @@
...
+                value: prescription_drug.id
+              ),
+              code: FHIR::CodeableConcept.new(
+                coding: FHIR::Coding.new(
+                  system: "http://www.nlm.nih.gov/research/umls/rxnorm",
+                  code: prescription_drug.rxnorm_code,
+                  display: prescription_drug.name
+                )
+              )
+            )
```
tfidfwastaken, line 42: We shouldn't have to repeat everything here (not sure if this would lead to valid FHIR). We can directly reference the contained resource like so:

```ruby
FHIR::Reference.new(reference: "##{contained_thing_id}")
```

app/services/one_off/fhir/prescription_drug_exporter.rb
```diff
@@ -0,0 +1,118 @@
...
+            lastUpdated: prescription_drug.device_updated_at.iso8601,
+            createdAt: prescription_drug.device_created_at.iso8601
+          )
+        )
+      end
+
+      def dosage_instruction
+        if prescription_drug.frequency.nil?
+          return nil
+        end
```
tfidfwastaken, line 70: [non-blocking]

```suggestion
        return if prescription_drug.frequency.nil?
```

app/services/one_off/fhir/patient_exporter.rb
```diff
@@ -0,0 +1,82 @@
...
+        id: FHIR::Organization.new(
+          identifier: [
+            FHIR::Identifier.new(
+              value: patient.assigned_facility_id
+            )
+          ]
+        )
+      ),
+      meta: FHIR::Meta.new(
+        lastUpdated: patient.updated_at.iso8601,
```
tfidfwastaken, line 48: [q] should this be `device_updated_at`?

app/services/one_off/fhir/prescription_drug_exporter.rb
```diff
@@ -0,0 +1,118 @@
...
+        [
+          FHIR::Dosage.new(
+            doseAndRate: [
+              FHIR::Dosage::DoseAndRate.new(
+                doseQuantity: FHIR::Quantity.new(
+                  value: prescription_drug.dosage.split("MG").first,
```
tfidfwastaken, line 76: [nonblocking, style]
```suggestion
                  value: prescription_drug.dosage.delete_suffix("MG"),
```

lib/tasks/fhir.rake
```diff
@@ -0,0 +1,50 @@
...
+    patients = Patient.order(:created_at)
+    patients = remove_pii(patients)
+    file_path = args[:file_path]
+    resources = []
+
+    patients.each do |patient|
+      resources << OneOff::Fhir::PatientExporter.new(patient).export
+      resources << patient.blood_pressures.map { |bp|
+        OneOff::Fhir::BloodPressureExporter.new(bp).export
+      }
```
tfidfwastaken, line 15: [nit, nonblocking] Could use `do` syntax here and elsewhere.

lib/tasks/fhir.rake
```diff
@@ -0,0 +1,50 @@
...
+      resources << OneOff::Fhir::MedicalHistoryExporter.new(patient.medical_history).export
+      resources << patient.appointments.map { |appointment|
+        OneOff::Fhir::AppointmentExporter.new(appointment).export
+      }
+    end
+
+    File.open(file_path, "w+") do |f|
+      resources.flatten.each do |resource|
+        f.puts(resource.as_json.to_json)
```
tfidfwastaken, line 30: [q] is `as_json` needed? I'd have thought `to_json` would do the necessary conversion.

tfidfwastaken (changes requested):
Requesting some changes and clarification:
* Errors in some LOINC URLs and LOINC code values
* Questions about fallback values, optional fields
* Question about how references are used in this library

I have left some comments on code style and random details too (none of these are a blocker).

Also: Should we be adding specs for these? It seems like a chunky script and the kind of thing we might have to run again someday.

app/services/one_off/fhir/prescription_drug_exporter.rb
tfidfwastaken: I realised this suggestion wouldn't universally work. But the `split` won't work universally either.

Here are all the variations of values I have found looking at metabase.

There are over 100 occurrences of each of these values in Bangladesh and India:
```
500 mg TDS
12.5 mg
30mg5
0+0+1mg
50+mg
.5mg
mg

mg20
2000 mg
6.5mg
```

And these are not even getting into corner cases with a few occurrences like:
```
24iu
50plus
20+0+16unit
5tab
,3mg
2omg
```

Notably `50MGOD` is a format that only exists in sandbox data.

Newer data seems to be a lot more consistent since the app doesn't make dosage input very freeform anymore, but it's still one of `50 mg` or `50 mg OD`. But I wonder if we should be trying to parse this at all. One compromise I can think of is conditionally parsing dosage only if it matches the aforementioned format.

app/services/one_off/fhir/prescription_drug_exporter.rb
tfidfwastaken: [suggestion] Instead of returning nil if there's no frequency, can we create a `doseAndRate` without the `timing` field? That should be allowed according to the standard.

app/services/one_off/fhir/medical_history_exporter.rb
```diff
@@ -0,0 +1,52 @@
...
+      DM_CONDITION_CODE = "73211009"
+      HTN_CONDITION_CODE = "38341003"
+
+      def initialize(medical_history)
+        @medical_history = medical_history
+      end
+
+      def export
+        conditions = []
+        if medical_history.diabetes == "yes"
```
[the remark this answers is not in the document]
qptr: Did you mean `medical_history.diabetes_yes?` Otherwise it returns true for any non-nil value

lib/tasks/fhir.rake
qptr: We wanted to print each json on a single line. `to_json` adds newlines and indentation to prettify the json string

app/services/one_off/fhir/prescription_drug_exporter.rb
qptr: Going to go with your example, since it's closer to the examples

app/services/one_off/fhir/medical_history_exporter.rb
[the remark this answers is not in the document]
tfidfwastaken: woops, yeah

app/services/one_off/fhir/prescription_drug_exporter.rb
qptr: I guess if the reference is to a contained resource, it's all the same

app/services/one_off/fhir/appointment_exporter.rb
qptr: Thinking I'll go ahead with [this](https://github.com/simpledotorg/simple-server/pull/5335#discussion_r1432810127) approach and just pass whatever resource id on the `identifier` key of a Reference. No validation errors.

qptr: Don't know if I followed the question, you're saying the actor field expects a hash?

But re: building references correctly:

Using id was a hack to get the validation checks to pass a while ago. From [this](https://github.com/simpledotorg/simple-server/pull/5335#discussion_r1432810127), I'm going to rewrite all the references to directly use the `identifier` key and pass to it the respective id. In this example, rewrite this as:
```
 actor: FHIR::Reference.new(
    identifier: FHIR::Identifier.new(
      value: appointment.patient_id
   )
 )
```
No validation errors, and returns the expected json:
```
{... {"actor"=>{"identifier"=>{"value"=>"1dae7c00-7e55-4879-be50-184b2f535e0f"}}} ... }
```

app/services/one_off/fhir/patient_exporter.rb
```diff
@@ -0,0 +1,82 @@
...
+    identifiers = [
+      FHIR::Identifier.new(
+        value: patient.id.to_s,
+        use: "official"
+      )
+    ]
+    patient.business_identifiers.simple_bp_passport.each do |identifier|
+      identifiers << FHIR::Identifier.new(
+        value: identifier.identifier.to_s,
+        use: "secondary"
```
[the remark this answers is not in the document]
qptr: Think we wanted to export bp passport ids and have them be differentiable from patient ids. I think we can keep them around for now.

app/services/one_off/fhir/blood_pressure_exporter.rb
```diff
@@ -0,0 +1,65 @@
...
+      ],
+      code: FHIR::CodeableConcept.new(
+        coding: FHIR::Coding.new(
+          system: "http://lonic.com/",
+          code: "85354-9"
+        )
+      ),
+      component: [
+        observation_component("8460-6", blood_pressure.systolic),
+        observation_component("8460-8", blood_pressure.diastolic)
```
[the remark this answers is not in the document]
qptr: Shit. Nice catch

app/services/one_off/fhir/appointment_exporter.rb
```diff
@@ -0,0 +1,57 @@
...
+            lastUpdated: appointment.device_updated_at.iso8601,
+            createdAt: appointment.device_created_at.iso8601
+          )
+        )
+      end
+
+      def cancellation_code
+        case appointment.cancel_reason
+        when "dead" then "pat-dec"
+        else "pat"
```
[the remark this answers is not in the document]
qptr: Removing these entirely. There's not much overlap with our reasons and the accepted codes, and this is a not very significant optional field.

app/services/one_off/fhir/appointment_exporter.rb
tfidfwastaken: Lovely. This explanation unconfused me.

app/services/one_off/fhir/prescription_drug_exporter.rb
qptr: Uh, read about contained resources. `contained` + `medicationReference` makes a lot of sense now and seems better to do than exporting prescription drugs separately. There can be a lot of duplicates, and it's not like just exporting protocol drugs. It's also not how we store things in Simple.

qptr:
> Also: Should we be adding specs for these? It seems like a chunky script and the kind of thing we might have to run again someday.

Right now this is under the `one_off` folder, so I think we can treat it that way. The only card tracking this rn is a spike card to do a one-off fhir export. I think let this be for now. If this becomes something we'll use more seriously and often, we can add specs then.

app/services/one_off/fhir/prescription_drug_exporter.rb
qptr: :/ Yeah. This also was a hack to get things working quickly.
Conditionally pasring sounds fine. Will log a warning

qptr: Also, just came across `2 puff`, `BD` on metabse

qptr: This is done

PR #5336, opened 2023-11-30
Evict patient from experiment if their visit facility is deleted

qptr:
An error occurs while recording a visit for a patient enrolled in an experiment (also called a membership), if their visit facility has been soft deleted.

Since this is invalid data that can affect the experiment's outcomes, and also disrupts the daily experiment runner, these memberships will be evicted from the experiment.

Another change this PR makes is to filter out evicted memberships that have visited but haven't been marked as visited, from those whose visits need to be marked. This is for 2 reasons:
1. Memberships who were evicted due to their visit facilities being soft deleted won't get picked up again to record the invalid visit
2. Evicted patients aren't expected to continue to be part of the experiment flow

### DIFF
--- a/app/models/experimentation/treatment_group_membership.rb
+++ b/app/models/experimentation/treatment_group_membership.rb
@@ -48,7 +48,7 @@ module Experimentation
       visit_facility = earliest_visit.facility
 
       if visit_facility.nil?
-        self.class.where(id: id).evict(reason: "visit facility soft deleted")
+        Experimentation::TreatmentGroupMembership.where(id: id).evict(reason: "visit facility soft deleted")
         return
       end
 
--- a/app/models/experimentation/treatment_group_membership.rb
+++ b/app/models/experimentation/treatment_group_membership.rb
@@ -22,18 +22,24 @@ module Experimentation
     def record_notification(reminder_template_id, notification)
       reload # this is to reload the `messages` field to avoid staleness while updating.
       self.messages ||= {}
+      localized_message = notifiaction_localized_message(notification)
       self.messages[reminder_template_id] = {
         message_name: notification.message,
         remind_on: notification.remind_on,
         notification_status: notification.status,
         notification_id: notification.id,
-        localized_message: notification.localized_message,
+        localized_message: localized_message,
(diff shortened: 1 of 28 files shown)


spec/models/experimentation/treatment_group_membership_spec.rb
```diff
@@ -83,6 +83,22 @@
...
+      visit_facility = create(:facility)
+      visit = create(:prescription_drug, device_created_at: 2.days.ago, patient: membership.patient, facility: visit_facility)
+
+      visit_facility.discard!
+      visit.reload
+
+      membership.record_visit(blood_pressure: nil, blood_sugar: nil, prescription_drug: visit)
+      membership.reload
+
+      expect(membership.status).to eq("evicted")
```
danySam, line 98: Do you want to add a negative test for non "evicted" scenarios?

danySam (approved):

PriyangaPKini (approved):

app/models/experimentation/treatment_group_membership.rb
```diff
@@ -22,18 +22,24 @@ def self.evict(reason:)
     def record_notification(reminder_template_id, notification)
       reload # this is to reload the `messages` field to avoid staleness while updating.
       self.messages ||= {}
+      localized_message = notifiaction_localized_message(notification)
```
govindkrjoshi, line 25:
```ruby
localized_message = notifiaction_localized_message(notification) || ""
```

app/models/experimentation/treatment_group_membership.rb
```diff
@@ -22,18 +22,24 @@ def self.evict(reason:)
...
         created_at: notification.created_at.to_s
       }
       save!
+    end
+
+    def notifiaction_localized_message(notification)
+      notification.localized_message
     rescue Messaging::MissingReferenceError => error
       Rails.logger.error("Failed fetching localised message on notification #{notification.id}, reason: #{error.message}")
+      ""
```
govindkrjoshi, line 42: With the above change this is not necessary

qptr: Rails.logger.error returns true @govindkrjoshi

PR #5338, opened 2023-12-04
spec reporting helpers: fix out of range month

tfidfwastaken:
This is a December special bug 🎁 . The month added to 13, which breaks the time construction. We instead try to match the original intent of the code by using ActiveSupport time arithmetic.

Example failure: https://simple.semaphoreci.com/workflows/3d2e4248-1cb5-4032-b862-4951ac4485b2/summary?pipeline_id=5d9b7866-fb0d-45fb-8552-9fbda1fb854f


PriyangaPKini (approved):
LGTM

PR #5352, opened 2023-12-12
Recent updates dec 2023

jamiecarter7:
**Story card:** None

## Because

Recent updates for Dec-2023

## This addresses

Additions:
- Patient data protection agreement
- App data encryption
- Improved titration reports


dburka:
Added a comment that should be addressed. Otherwise looks good to merge.

PR #5354, opened 2023-12-14
Add data migration to cleanup a Bangladesh facility

tfidfwastaken:
## Because

We need to remove all this data with one of our partners integrating with the import API.


db/data/20231213121605_cleanup_bd_org_data.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class CleanupBdOrgData < ActiveRecord::Migration[6.1]
```
govindkrjoshi, line 3: Can name this specific to ICDDRB org

tfidfwastaken: Was not sure how much to reveal since this is a public repo. Will change.

PR #5356, opened 2023-12-15
Add DHIS2 aggregate data element config to sandbox

qptr:
Adding ids of aggregate data elements from DHIS2 sandbox to be able to export aggregate hypertension data from sandbox.


PriyangaPKini:
Did you try exporting the disaggregated data element values?

qptr:
> Did you try exporting the disaggregated data element values?
These are aggregate data elements. Yes, they show up on sandbox.

PriyangaPKini (approved):

PR #5362, opened 2023-12-20
Fix validation errors for import user numbers

tfidfwastaken:
Switching to an import user per organization introduced a bug where creating a new Import User causes a bug because the phone number fails the uniqueness validations. We no longer have a globally constant phone number—we instead generate a random string as our fake phone number that is unique to each organization and guaranteed to not collide again.

Explanations for some choices:
* We convert the UUID of the org into a url-safe base64 representation because:
    * It is simple and reversible
    * It is only 22 bytes, which is shorter than the 36-byte UUID representation (this helps with displaying and copying)
    * There's guaranteed to be no collisions since organizations are universally unique
* We do not include a data migration to migrate the existing "0000000001" numbers because the phone number of a bot user is a filler value with no intrinsic purpose.

### DIFF
--- a/app/services/import_user.rb
+++ b/app/services/import_user.rb
@@ -1,13 +1,11 @@
 class ImportUser
-  IMPORT_USER_PHONE_NUMBER = "0000000001"
-
   def self.find_or_create(org_id:)
     find(org_id) || create(org_id)
   end
 
   def self.find(org_id)
     PhoneNumberAuthentication.joins(:user)
-      .find_by(phone_number: IMPORT_USER_PHONE_NUMBER, user: {organization_id: org_id})&.user
+      .find_by(phone_number: bot_phone_number(org_id), user: {organization_id: org_id})&.user
   end
 
   def self.create(org_id)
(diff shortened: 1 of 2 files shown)


app/services/import_user.rb
```diff
@@ -43,4 +41,12 @@ def self.create(org_id)
...
     "#{rand(10)}#{rand(10)}#{rand(10)}#{rand(10)}"
   end
+
+  def self.bot_phone_number(org_id)
+    # Losslessly compresses an organisation UUID value into a url-safe base64 representation.
+    # This is a bit more compact than stuffing a UUID string as a phone number directly. Other
+    # forms of hashing may lead to information loss, doing it this way keeps the collision
+    # probability the same as the organization UUID itself.
+    [[org_id.remove("-")].pack("H*")].pack("m0").tr("+/", "-_")[..21]
+  end
```
tfidfwastaken, line 51: I know this bit of code is ugly, but given that it's short and used in a contained manner, it felt okay.

govindkrjoshi (approved):

PR #5365, opened 2023-12-22
Automate AlphaSms balance check

qptr:
**Story card:** [sc-11640](https://app.shortcut.com/simpledotorg/story/11640/automate-alphasms-balance-check)

We want to automate balance check in Bangladesh and alert on low balance.

The reason the error message is being logged and not raised as an exception is so we can add a Datadog alert on the log. The corresponding BSNL task relies on Sentry to alert on an uncaught exception for low balance, but we have to manually resolve the alert each time for it to show up again next time. This is not an issue on Datadog.

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -50,6 +50,12 @@ every :day, at: local("02:00 pm"), roles: [:cron] do
   end
 end
 
+every :day, at: local("02:00 pm"), roles: [:cron] do
+  if CountryConfig.current_country?("Bangladesh") && SimpleServer.env.production?
+    rake "alpha_sms:check_balance"
+  end
+end
+
 every :day, at: local("05:30 pm"), roles: [:cron] do
   runner "Messaging::Bsnl::Sms.get_message_statuses"
 end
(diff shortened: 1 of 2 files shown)


config/schedule.rb
```diff
@@ -50,6 +50,12 @@ def local(time)
   end
 end
 
+every :day, at: local("02:00 pm"), roles: [:cron] do
+  if CountryConfig.current_country?("Bangladesh") && SimpleServer.env.production?
```
olttwa, line 54: Conventionally, we're keeping checks inside the every block. I'm wondering if we should put every block inside the condition checks? That way, each country's crontab files will only have rake tasks relevant for that environment. Something like this, maybe?

```ruby
if CountryConfig.current_country?("Bangladesh") && SimpleServer.env.production?
  every :day, at: local("02:00 pm"), roles: [:cron] do
    rake "alpha_sms:check_balance"
  end
end
```

qptr: Looks like this is happening regardless. The task only gets converted to cron syntax if the condition matches.

olttwa: Hmm, that's nice then. PR looks good to be merged.

PR #5375, opened 2024-01-08
Asynchronous patient line list generation

govindkrjoshi:
**Story card:** [sc-10665](https://app.shortcut.com/simpledotorg/story/10665/fix-to-enable-automating-chennai-patient-line-list)

## Because

Line lists are currently generated in memory. This can consume a lot of memory for large regions, causing the sidekiq processes to crash. 

## This addresses

This adds an alternative code path to test downloading the linelist through an asynchronous rack process instead of a sidekiq job. Since each line of the generated csv will be flushed immediately, downloading a linelist should not cause a spike in resource consumption. 

Additionally, the use can see the downloading file in their browsers downloads tab. 

This PR also adds instrumentation to test this thesis in production before replacing the existing codepath.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,6 +71,7 @@ gem "rswag-api"
 gem "rswag-ui"
 gem "ruby-progressbar", require: false
 gem "rubyzip"
+gem "zip_tricks"
 gem "sassc-rails"
 gem "scenic"
 gem "scientist"
--- a/app/controllers/reports/patient_lists_controller.rb
+++ b/app/controllers/reports/patient_lists_controller.rb
@@ -9,25 +9,27 @@ class Reports::PatientListsController < AdminController
 
   def show
     if Flipper.enabled?(:patient_list_direct_download)
-      patients = @region.assigned_patients.excluding_dead
-      enumerator = PatientsWithHistoryExporter.csv_enumerator(patients)
+      Datadog::Tracing.trace("Direct patient line list download") do
+        patients = @region.assigned_patients.excluding_dead
+        enumerator = PatientsWithHistoryExporter.csv_enumerator(patients)
 
-      file_name = "patient-list_#{@region.region_type}_#{@region.name}_#{I18n.l(Date.current)}"
-      csv_file_name = "#{file_name}.csv"
-      zip_archive_name = "#{file_name}.zip"
-      response.headers["Content-Disposition"] = ActionDispatch::Http::ContentDisposition.format(
-        disposition: "attachment",
-        filename: zip_archive_name
-      )
+        file_name = "patient-list_#{@region.region_type}_#{@region.name}_#{I18n.l(Date.current)}"
+        csv_file_name = "#{file_name}.csv"
+        zip_archive_name = "#{file_name}.zip"
+        response.headers["Content-Disposition"] = ActionDispatch::Http::ContentDisposition.format(
+          disposition: "attachment",
+          filename: zip_archive_name
+        )
 
-      first_row = enumerator.peek
-      response.headers["Last-Modified"] = first_row.second
+        first_row = enumerator.peek
+        response.headers["Last-Modified"] = first_row.second
 
-      return zip_tricks_stream do |zip|
-        zip.write_deflated_file(csv_file_name) do |sink|
-          CSV(sink) do |csv_write|
-            enumerator.each do |row|
-              csv_write << row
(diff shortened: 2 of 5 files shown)


Gemfile
```diff
@@ -71,6 +71,7 @@ gem "rswag-api"
 gem "rswag-ui"
 gem "ruby-progressbar", require: false
 gem "rubyzip"
+gem "zip_tricks"
```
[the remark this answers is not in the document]
tfidfwastaken: Moreover, the `zip_tricks` gem is written in the [Hippocratic License](https://firstdonoharm.dev/version/1/2/license/) which is not a valid Open Source license because it has a non-enforceable and nebulous "do no harm" clause.

app/controllers/reports/patient_lists_controller.rb
```diff
@@ -1,11 +1,41 @@
...
+
+        return zip_tricks_stream do |zip|
+          zip.write_deflated_file(csv_file_name) do |sink|
+            CSV(sink) do |csv_write|
+              enumerator.each do |row|
+                csv_write << row
+              end
+            end
+          end
+        end
```
tfidfwastaken, line 35: [minor, nonblocking] This part leaks CSV generation logic into the controller.

For an enumerator created with the `#csv_enumerator` method, I'd have expected an object that yields actual CSV strings, but instead it yields plain arrays—this is why we need to now call the `CSV` module in our controller to actually perform the conversion. Moreover, it leads to 3 layers of block indentation which is a little tricky to parse.

As an alternate proposal, I have annotated a suggestion in the `patients_with_history_exporter.rb` that would make this enumerator return strings instead of arrays. Once that happens, we could rewrite this section as:

```suggestion
        return zip_tricks_stream do |zip|
          zip.write_deflated_file(csv_file_name) do |sink|
              enumerator.each { |row| sink << row }
          end
        end
```

FYI, I'm okay with the current code also and I haven't fully tested if my above proposal works. Consider this suggestion nonblocking.

app/exporters/patients_with_history_exporter.rb
```diff
@@ -33,6 +33,27 @@ def self.csv(*args)
...
+      yeilder << measurement_headers
+      yeilder << csv_headers
+
+      summary.in_batches(of: BATCH_SIZE).each do |batch|
+        batch.each do |patient_summary|
+          yeilder << csv_fields(patient_summary)
+        end
+      end
+    end
+  end
```
tfidfwastaken, line 55: [nonblocking]
...continuing from the previous comment, the change needed to make this yield CSV strings rather than arrays:
```suggestion
  def csv_enumerator(patients, display_blood_sugars: true)
    @display_blood_sugars = display_blood_sugars
    summary = MaterializedPatientSummary.where(patient: patients)

    Enumerator.new do |yeilder|
      yeilder << timestamp.to_csv
      yeilder << measurement_headers.to_csv
      yeilder << csv_headers.to_csv

      summary.in_batches(of: BATCH_SIZE).each do |batch|
        batch.each do |patient_summary|
          yeilder << csv_fields(patient_summary).to_csv
        end
      end
    end
  end
```

There's probably a better way, but can't think of one right now.

tfidfwastaken: Also typo: `yielder` not `yeilder`

tfidfwastaken (approved):
PR looks good to me.

Concerns:
* `zip_tricks` is under the Hippocratic license which is not OSI/FSF approved, and has a nebulous "do no harm" clause. While Simple obviously does no harm, I'm slightly unsure about the license's rigour and enforceability.

The clause:

> No Harm: The software may not be used by anyone for systems or activities that actively and knowingly endanger, harm, or otherwise threaten the physical, mental, economic, or general well-being of other individuals or groups, in violation of the United Nations Universal Declaration of Human Rights ([https://www.un.org/en/universal-declaration-human-rights/)](https://www.un.org/en/universal-declaration-human-rights/).
> 
> Services: If the Software is used to provide a service to others, the licensee shall, as a condition of use, require those others not to use the service in any way that violates the No Harm clause above.

This raises many questions with no clear answers like: What happens if someone else uses a Simple fork to violate Universal Declaration of Human Rights? What does that even mean? Is this license viral like the GPL?

Non-blocking:
* Suggested a refactor for the CSV enumerator to make the CSV logic not leak into the controller. Not a strong opinion.

PR #5378, opened 2024-01-17
import api: ensure no duplicates for Conditions

tfidfwastaken:
**Story card:** [sc-11839]

## Because

Conditions were wrongly assumed to be longitudinal, resulting in 1:n mapping of patients to medical histories. Ideally every patient must have only one medical history to ensure that the reporting pipeline is accurate.

## Changes

* Change in the import API to ensure that all Conditions will always map to a previous medical history (it will only create a new one if it didn't exist at all for a patient)
* A data migration that merges all the previous medical histories for ICDDRB (no one else has sent multiple medical_histories yet).

### DIFF
--- a/app/services/bulk_api_import/fhir_condition_importer.rb
+++ b/app/services/bulk_api_import/fhir_condition_importer.rb
@@ -27,7 +27,7 @@ class BulkApiImport::FhirConditionImporter
 
   def build_attributes
     {
-      id: translate_id(@resource.dig(:identifier, 0, :value), org_id: @organization_id),
+      id: condition_id,
       patient_id: translate_id(@resource[:subject][:identifier], org_id: @organization_id),
       prior_heart_attack: "unknown",
       prior_stroke: "unknown",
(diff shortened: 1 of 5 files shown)


app/services/bulk_api_import/fhir_condition_importer.rb
```diff
@@ -41,10 +41,19 @@ def build_attributes
...
+  def condition_id
+    @condition_id ||= translate_id(@resource[:subject][:identifier], org_id: @organization_id, ns_prefix: "condition")
+  end
+
   def diagnoses
-    @resource[:code][:coding].each_with_object({hypertension: "no", diabetes: "no"}) do |coding, diagnoses|
-      condition = CONDITION_TRANSLATION[coding[:code]]
-      diagnoses[condition] = "yes"
+    @diagnoses ||= begin
+      prev_htn, prev_dm = MedicalHistory.where(id: condition_id).pluck(:hypertension, :diabetes).first
```
govindkrjoshi, line 50: [Nitpick] Since we only care for 2 cases, hypertension and diabetes, Wondering if this would be simpler to read if we enumerated the conditions directly?

db/data/20240117114102_dedupe_icddrb_medical_histories.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class DedupeIcddrbMedicalHistories < ActiveRecord::Migration[6.1]
+  FACILITY_ID = "f472c5db-188f-4563-9bc7-9f86a6ed6403"
```
govindkrjoshi, line 4: Do we need to do this for other facilities/orgs also?

tfidfwastaken: nope, nobody else has sent multiple conditions per patient, I checked this

PR #5380, opened 2024-01-24
Make visit window for stale patients configurable

qptr:
**Story card:** [sc-11826](https://app.shortcut.com/simpledotorg/story/11826/set-up-sms-reminders-in-bangladesh)

## Because

In stale patient experiments, patients need to have visited in a particular time window in order to be eligible to be notified. This is currently hard-coded and we want to make this configurable per country. For example, we only want to notify patients who are between 35-90 days overdue in Bangladesh, vs 35-365 days overdue in India.

## This addresses

This PR moves the visited since and visited until days to env files
Related PR updating env files per country: https://github.com/simpledotorg/container-deployment/pull/92


.env.profiling
```diff
@@ -53,6 +53,8 @@ SIMPLE_SERVER_DATABASE_HOST=localhost
 SIMPLE_SERVER_ENV=development
 SIMPLE_SERVER_HOST='localhost:3000'
 SIMPLE_SERVER_HOST_PROTOCOL=https
+STALE_EXPERIMENT_VISITED_SINCE_DAYS=365
+STALE_EXPERIMENT_VISITED_UNTIL_DAYS=35
```
tfidfwastaken, line 57: I am not sure how to interpret these envvar names.

The way I'm reading it right now is: "Stale patients visited since 365 days, until 35 days". Which does not make sense.

Riff on an alternatives:

```suggestion
STALE_EXPERIMENT_MIN_DAYS_SINCE_LAST_VISIT=35
STALE_EXPERIMENT_MAX_DAYS_SINCE_LAST_VISIT=365
```

It's quite possible that the current envvar name hasn't clicked for me for whatever reason and the name is fine. An explanation would help.

tfidfwastaken (approved):
left a question about how to interpret the envvar name. Other than that PR looks good, hence approving anyway.

.env.profiling
qptr: This name is much clearer. Updated

PR #5381, opened 2024-01-24
ADR: Use OAuth for external Simple clients

tfidfwastaken:
**Story card:** [sc-11959](https://app.shortcut.com/simpledotorg/story/11959)


danySam (approved):
Looks good.

Minor suggestions:
1. Can remove all the mentions of operational information (ex: Bangladesh, FHIR, etc)
2. The ## Decision section can start with the description of the decision and then go on to explain why

tfidfwastaken:
> Minor suggestions:
> 
> 1. Can remove all the mentions of operational information (ex: Bangladesh, FHIR, etc)

I am not sure if it's a good idea to omit this information since it's a relevant force that influenced this technical decision.

From [Documenting Architecture Decisions](https://www.cognitect.com/blog/2011/11/15/documenting-architecture-decisions):
> Each record describes a set of forces and a single decision in response to those forces. Note that the decision is the central piece here, so specific forces may appear in multiple ADRs.

We have also included [similar kind of context](https://github.com/simpledotorg/simple-server/blob/2f5a3d616393988a55a8d29535f177c244d4428c/doc/arch/014-region-level-sync.md#L19) [in past ADRs](https://github.com/simpledotorg/simple-server/blob/9da5c76ddc3497c8af9e0eb4aad35dbf86da57ee/doc/arch/018-imo-messaging.md#L5-L4).

PR #5383, opened 2024-01-29
Fix flaky test in overdue patient specs

PriyangaPKini:
**Story card:** [sc-10942](https://app.shortcut.com/simpledotorg/story/10942/investigate-flaky-test-in-overdue-patients-in-region-summary-spec)

## Because

The test for `return the count of contactable patients returned to care who were removed from overdue list` is flaky. 
 
Current test arrangement:
 - [One of the patients](https://github.com/simpledotorg/simple-server/blob/0e548dcafba79c72357a5367b3182496725c4ff0/spec/models/reports/region_summary_spec.rb#L1469) has 2 call results, one in the previous month and one in the current month. 
 - The patient returns to care on `the 2nd of the current month`.  
 - The test will pass if the visit date of the patient fall within 15 days of the call .i.e (visit date - call date <=15).
 - First call was made on `1.month.ago`. 
 - `1.month.ago` is relative to the current day and the date changes daily.
 
For eg, assume today is 31-Jan-2024
 - date of first call: `1.month.ago` = 31-Dec-2023 (changes daily)
 - visit date: `one_month_ago + 2.days` = 2-Jan-2023 (fixed date)
 - The test pass only when `(visit date - date of first call) <= 15`.  Since this difference (visit date - date of first call) changes daily, the test turned to be flaky.

## This addresses

- Use a fixed date for `date of first call`
  - date of first call: `one_month_ago + 20.days` = 20-Dec-2023 (A random date which falls in the 15 day window)
  - [`one_month_ago + 20.days`](https://github.com/simpledotorg/simple-server/blob/0e548dcafba79c72357a5367b3182496725c4ff0/spec/models/reports/region_summary_spec.rb#L734)  is always the 20th of the previous month.


tfidfwastaken (approved):
LGTM, approving the fix 🚢 

As discussed on a call today: before merging, let's add a short example of where the spec fails to make the bug easier to understand.

PR #5384, opened 2024-01-29
Set up sms reminders in India for Feb to Apr

qptr:
**Story card:** [sc-11962]

This PR adds a data migration to set up sms reminders in IHCI states for the next 3 months

### DIFF
--- a/db/data/20240129115821_set_up_sms_reminders_india_feb_apr_2024.rb
+++ b/db/data/20240129115821_set_up_sms_reminders_india_feb_apr_2024.rb
@@ -20,7 +20,7 @@ class SetUpSmsRemindersIndiaFebApr2024 < ActiveRecord::Migration[6.1]
   end
 
   def up
-    # return unless CountryConfig.current_country?("India") && SimpleServer.env.production?
+    return unless CountryConfig.current_country?("India") && SimpleServer.env.production?
 
     EXPERIMENTS_DATA.each do |experiment_data|
       ActiveRecord::Base.transaction do
--- a/db/data/20240129115821_set_up_sms_reminders_india_feb_apr_2024.rb
+++ b/db/data/20240129115821_set_up_sms_reminders_india_feb_apr_2024.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+class SetUpSmsRemindersIndiaFebApr2024 < ActiveRecord::Migration[6.1]
+  # Regions where Simple is active and we send sms reminders currently:
+  # States - All of West Bengal, all of Nagaland, some districts in Tamil Nadu and Maharashtra
+  # Districts - Chennai, pimpri-chinchwad-municipal-corporation, pune-municipal-corporation
+  REGION_FILTERS = {
+    "states" => {"include" => ["West Bengal", "Nagaland", "Tamil Nadu", "Maharashtra"]},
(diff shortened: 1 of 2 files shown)


db/data/20240129115821_set_up_sms_reminders_india_feb_apr_2024.rb
```diff
@@ -0,0 +1,62 @@
...
...a-0ed693aad928 cf455f76-0bc0-4f59-b2be-c628de2b85c0 39731fe9-f13a-4370-a148-0968e73ccd16 42ca44b1-2fe8-49d3-af66-33bc2c36763f 8ba7d5df-5e36-4db9-a02c-ad410a6e6e31 90e846bb-d2f7-489f-b758-c5b5fdae3f70 f6395338-4ecb-468c-9a72-d1b5b8e1169e 12077f18-82ff-4d24-abb6-f968f4ae31d2 7b3e5865-dfdd-42c6-ac79-bf1570669591 70ecb380-93b1-4dc0-9aa3-bbebf3df30be 5d652b3f-29db-4e80-b82e-8d690aa16d9f e56e32a6-822a-44c8-aba5-2d066ef9c84b c953f71f-fca9-4456-ac16-a363cafef035 b3c2e413-d9df-49a6-afa0-2bbc4610e884 91be7a95-4dec-4f0d-b2e9-5ddb87f05a6a bf97c3a0-fa91-4b66-857a-374083dcc271 dc0d57ce-239c-4ea0-b3a2-c5aadf683cb1 86bbbcd2-7142-4378-8078-8d101be8932c 0d7e42ac-b4ed-4c24-bfc9-5211ded7a9f0 795048aa-4d6c-49b4-abef-4e73c1a2aecf 9c1f3013-1347-4adb-b027-cfd642d2e20a 5309ff56-7c53-4e7d-9939-d25e2cd2018d dadd4cb6-dc42-4f12-8c3c-c4fa1c8fd46d 7435afbf-28bf-432f-9624-fda905b837c8 4da7d239-b439-4003-9a3f-a242dd4db856 ef6fbb8d-e4b5-4844-a7e3-8627e9328df5 f1dc8d43-4da9-4378-a1f4-a52dbd5294d3 43e60477-65bf-4407-a92d-81a6c727c0dd b6c07887-b600-4840-89c9-90d65386c9d8 ff16e816-893c-4908-890b-bc64bce22c10 179e7cef-c812-4ce0-af6e-766c8c2e15f0 683472ef-f787-4da9-b2d0-0fa34003688e b36a729c-0fea-4406-a63b-edaf26e8a662]}
```
govindkrjoshi, line 9: If the intent is to include a few districts, why are we excluding a long list of blocks instead? 
This long list of block ids in the code, is a code smell. I think this should be refactored. 

Also you should be very careful when committing ids to the codebase.

db/data/20240129115821_set_up_sms_reminders_india_feb_apr_2024.rb
```diff
@@ -0,0 +1,62 @@
...
+    {
+      start_time: DateTime.new(2024, month_number).beginning_of_month,
+      end_time: DateTime.new(2024, month_number).end_of_month,
+      current_patients_experiment_name: "Current Patient #{month} 2024",
+      stale_patients_experiment_name: "Stale Patient #{month} 2024"
+    }
+  end
+
+  def up
+    # return unless CountryConfig.current_country?("India") && SimpleServer.env.production?
```
govindkrjoshi, line 23: Should this be uncommented?

db/data/20240129115821_set_up_sms_reminders_india_feb_apr_2024.rb
govindkrjoshi: Using slugs instead of ids or names is better since they are human read able and don't have issues related names either.

govindkrjoshi (changes requested):

db/data/20240129115821_set_up_sms_reminders_india_feb_apr_2024.rb
qptr: There's a card for this refactor [here](https://app.shortcut.com/simpledotorg/story/11699/refactor-regions-filter-in-notification-experiments) but it's a bit involved. Would be good to pick it up soon.

This file is copying over values from the first migration that added these values that we know works, in case we want to set this up until the refactor is done.

qptr: > If the intent is to include a few districts, why are we excluding a long list of blocks instead?
This is how filters are coded right now. Think this requirement with districts wasn't anticipated when it was written.

govindkrjoshi (approved):

PR #5387, opened 2024-02-01
ADR: Prefer FHIR-compatible payloads for data transfer

tfidfwastaken:
**Story card:** [sc-11974]

### DIFF
--- a/doc/arch/022-use-fhir-for-data-transfer.md
+++ b/doc/arch/022-use-fhir-for-data-transfer.md
@@ -1,0 +1,127 @@
+# Prefer FHIR-compatible payloads for data transfer
+## Context
+
+There's been a need for Simple Server to integrate a variety of other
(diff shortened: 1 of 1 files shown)


govindkrjoshi:
I think we should also talk about the consequences of it not being fully fhir compliant. 

This ADR also seems like a good place to mention the places where we differ from FHIR and why.

tfidfwastaken:
@govindkrjoshi This section was meant to cover that:

> * We may require case-by-case customisation to ensure compatibility with other 
  systems with varying FHIR profiles (given the flexibility built into the
  standard).
> * There will be some friction for systems that are not FHIR-compatible to use
  our FHIR based APIs for import and export.

But I shall take your comment to mean that it perhaps should be demarcated more clearly and elaborated with a little more detail.

> This ADR also seems like a good place to mention the places where we differ from FHIR and why.

👍

PR #5389, opened 2024-02-15
data migration: clear ICDDRB prescription_drugs

tfidfwastaken:
**Story card:** [sc-12002]

## How do I know this is correct?

I converted the activerecord query on rails console like this:

```ruby
    destroyed_drugs = PrescriptionDrug
      .joins(facility: {facility_group: :organization})
      .where(organization: {id: ORG_ID})
      .to_sql
```

I pasted the resulting SQL output on Metabase BD and inspected the result to ensure that I would be deleting the right thing.

### DIFF
--- a/db/data/20240215112358_destroy_icddrb_prescription_drugs.rb
+++ b/db/data/20240215112358_destroy_icddrb_prescription_drugs.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class DestroyIcddrbPrescriptionDrugs < ActiveRecord::Migration[6.1]
+  ORG_ID = "1bb836eb-9a0b-4646-96ad-7342b53a9155"
+
+  def up
+    unless CountryConfig.current_country?("Bangladesh") && SimpleServer.env.production?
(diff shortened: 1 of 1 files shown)


govindkrjoshi (approved):

db/data/20240215112358_destroy_icddrb_prescription_drugs.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class DestroyIcddrbPrescriptionDrugs < ActiveRecord::Migration[6.1]
+  ORG_ID = "1bb836eb-9a0b-4646-96ad-7342b53a9155"
+
+  def up
+    unless CountryConfig.current_country?("Bangladesh") && SimpleServer.env.production?
+      print "DestroyIcddrbPrescriptionDrugs is only for production Bangladesh"
```
danySam, line 8: It should print and return, right? Otherwise, it'll continue execution.

tfidfwastaken: good catch 🐟 . will fix.

PR #5401, opened 2024-03-01
Update region filter for march-april experiment in BD

PriyangaPKini:
**Story card:** [sc-12132](https://app.shortcut.com/simpledotorg/story/12132/send-sms-reminders-only-to-uhc-type-facilities-in-bd)

## This addresses

Sends SMS reminders only to facilities with facility type "UHC"


db/data/20240301082336_set_up_sms_reminders_bd_mar_apr2024.rb
```diff
@@ -0,0 +1,77 @@
...
+        end
+      end
+    end
+  end
+
+  def down
+    return unless CountryConfig.current_country?("Bangladesh") && SimpleServer.env.production?
+
+    EXPERIMENTS_DATA.each do |experiment_data|
+      Experimentation::Experiment.current_patients.find_by_name(experiment_data[:current_experiment_name])&.cancel
```
PriyangaPKini, line 73: I am only cancelling the updated experiments here.

db/data/20240301082336_set_up_sms_reminders_bd_mar_apr2024.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+class SetUpSmsRemindersBdMarApr2024 < ActiveRecord::Migration[6.1]
+
+  EXPERIMENTS_TO_DISCARD = (3..4).map do |month_number|
```
govindkrjoshi, line 5: Since we are cancelling the experiments and not discarding (calling .discard), maybe let's call it `EXPERIMENTS_TO_CANCEL`

db/data/20240301082336_set_up_sms_reminders_bd_mar_apr2024.rb
govindkrjoshi: Ideally the down migration should be the reverse of the up migration. Can you clarify in code comments why we are choosing to not do this.

PriyangaPKini: Sure 👍

db/data/20240301082336_set_up_sms_reminders_bd_mar_apr2024.rb
PriyangaPKini: > xperiments and not discarding (calling .discard), maybe let's call it `EXPERIMENTS_TO_CANCEL`

We are calling `.discard` inside the `cancel` method. That's why I named it as `EXPERIMENTS_TO_DISCARD`.

PR #5406, opened 2024-03-18
data migration: remove duplicate medical histories for ICDDRB

tfidfwastaken:
ICDDRB had pushed some duplicate medical history data in production due to an error in their import API integration. We have had to fix this before under different circumstances in #5378, and we are mostly repeating that same data migration from the aforementioned PR, but for all the facilities that have the duplicate as of this day.

**Story card:** [sc-12139]

### DIFF
--- a/db/data/20240318132105_dedupe_icddrb_medical_histories_2.rb
+++ b/db/data/20240318132105_dedupe_icddrb_medical_histories_2.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+class DedupeIcddrbMedicalHistories2 < ActiveRecord::Migration[6.1]
+  FACILITY_IDS = %w[
+    4280917c-a10e-4125-bb79-5cf6ebe4bd2b
+    75bdedd2-ca3a-483a-a81c-9f923d723489
+    8207fda1-a37b-4c19-be25-6056c47e374b
+    8a34af23-f0a4-46e1-8932-4b390b0bee64
+    8c6840c6-9ce8-4e01-82ee-1b8af00675a2
+    f472c5db-188f-4563-9bc7-9f86a6ed6403
+    fdb34f45-923c-4612-aa15-3f8b6cc3175f
+  ]
+
+  def up
+    unless CountryConfig.current_country?("Bangladesh") && SimpleServer.env.production?
+      return print "DedupeIcddrbMedicalHistories2 is only for production Bangladesh"
+    end
+
+    # NOTE: these merge resolution rules do not account for the case where the
+    # attribute to merge is blank. Since we have verified that this blank edge
+    # case does not occur in ICDDRB's data, we can get away with fewer rules.
+    merge_resolution_rules = {
+      Set["no"] => "no",
+      Set["yes"] => "yes",
+      Set["unknown"] => "unknown",
+      Set["no", "yes"] => "yes",
+      Set["no", "unknown"] => "no",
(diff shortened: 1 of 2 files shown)


db/data/20240318132105_dedupe_icddrb_medical_histories_2.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    fields_to_resolve = MedicalHistory.defined_enums.keys
+
+    transaction do
+      medical_histories_by_patient = MedicalHistory.joins(:patient)
+        .where(patient: {assigned_facility_id: FACILITY_IDS})
+        .group_by(&:patient_id)
+
+      medical_histories_by_patient.each_value do |merge_candidates|
+        next if merge_candidates.size < 2
```
danySam, line 37: Might be better to have this happen outside the block with a filter. Makes it slightly easier to understand and more intentional. Change not required for this migration though.

db/data/20240318132105_dedupe_icddrb_medical_histories_2.rb
```diff
@@ -0,0 +1,52 @@
...
+        merged_med_history_attrs = merge_candidates.first&.attributes&.except("id")
+        fields_to_resolve.each do |field_name|
+          conflicting_values = merge_candidates.map(&field_name.to_sym).to_set
+          merged_med_history_attrs[field_name] = merge_resolution_rules.fetch(conflicting_values)
```
danySam, line 41: `fetch` can potentially raise an exception. The value can be "" also?

danySam: Try this on the DB: 
```select distinct(receiving_treatment_for_diabetes) from medical_histories```

Comment (tfidfwastaken) [reply db/data/20240318132105_dedupe_icddrb_medical_histories_2.rb:41]: Nice catch. I had intentionally used `fetch` so that the data migration fails in case I missed an edge case like this.

I actually checked the value for every enum, there are blank occurrences for only `receiving_treatment_for_diabetes` in Bangladesh.

I'm not changing anything, because this case does not occur in any of ICDDRB's facilities:

```
SELECT
  *
FROM medical_histories AS mh
JOIN patients AS p
  ON p.id = mh.patient_id
WHERE
  (mh.receiving_treatment_for_diabetes IS NULL OR mh.receiving_treatment_for_diabetes = '') AND
  (
    p.assigned_facility_id = '4280917c-a10e-4125-bb79-5cf6ebe4bd2b'
    OR p.assigned_facility_id = '75bdedd2-ca3a-483a-a81c-9f923d723489'
    OR p.assigned_facility_id = '8207fda1-a37b-4c19-be25-6056c47e374b'
    OR p.assigned_facility_id = '8a34af23-f0a4-46e1-8932-4b390b0bee64'
    OR p.assigned_facility_id = '8c6840c6-9ce8-4e01-82ee-1b8af00675a2'
    OR p.assigned_facility_id = 'f472c5db-188f-4563-9bc7-9f86a6ed6403'
    OR p.assigned_facility_id = 'fdb34f45-923c-4612-aa15-3f8b6cc3175f'
  );
```

Accounting for the NULL case for that one enum adds a bit more complexity to the migration for no real gain. But I think I'll add a comment to avert a potential future copypasta disaster.

danySam: Sounds good 👍

danySam (approved):

PR #5409, opened 2024-03-28
Bangladesh dhis2 diabetes exporter

govindkrjoshi:
Epic: https://app.shortcut.com/simpledotorg/epic/8589/sending-dm-data-to-dhis2-bangladesh

## Because

We have received a request to export Diabetes data to DHIS2 in Bangladesh

## This addresses

Adds the necessary metric and queries to do the export


spec/queries/patient_states/diabetes/cumulative_assigned_patients_query_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  let(:regions) { setup_district_with_facilities }
+  let(:period) { Period.current }
+
+  context "cumulative assigned patients" do
+    it "returns all the assigned patients in a facility as of the given period" do
+      facility_1_patients = create_list(:patient, 2, :diabetes, assigned_facility: regions[:facility_1])
+      facility_2_patients = create_list(:patient, 2, :diabetes, assigned_facility: regions[:facility_2])
+      refresh_views
```
tfidfwastaken, line 15: [flag, nonblocking?] The spec runs in the CI are slow (the last one took 1 hour), and adding a whole bunch of tests, that refreshes the views $spec_count number of times.

I am not sure how much this will help, but could try to only refresh the minimum required views to shave off time, like
```
RefreshReportingViews.call(views: [...])
```

The reason this review comment can be ignored for now IMO is that not all test runs seem that slow: https://simple.semaphoreci.com/branches/b6a0d51a-d06b-48d0-87f5-a8e252163358

But the older ones might have had lesser refreshes.

spec/queries/patient_states/hypertension/cumulative_registrations_query_spec.rb
tfidfwastaken, line 1: Why are we modifying this file to change all references of HTN to diabetes?
Shouldn't the diabetes cumulative registration spec go in a new file?

spec/queries/patient_states/diabetes/cumulative_registrations_query_spec.rb
tfidfwastaken, line 1: The spec in this file is for hypertension, but it's in the `/diabetes` folder. This is related to the other comment about the diabetes cumulative registration spec, which is in the hypertension folder.

tfidfwastaken (changes requested):
The cumulative registration specs for diabetes and hypertension are in the wrong folders and should be swapped.

Other than that, the rest of my comments can be considered nonblocking—can answer those questions and leave it at that.

Reviewing it commit-by-commit was quite helpful, thanks for keeping that in mind.

app/jobs/dhis2/bangladesh_disaggregated_diabetes_exporter_job.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module Dhis2
+  class BangladeshDisaggregatedDiabetesExporterJob < Dhis2ExporterJob
```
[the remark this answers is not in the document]
govindkrjoshi: This is because the application job inherits from active job, and does not include sidekiq job. 

Active job is compatible with sidekiq but it does not include all the features of sidekiq jobs.

app/queries/patient_states/diabetes/cumulative_assigned_patients_query.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def call
+    Reports::PatientState
+      .where(
+        assigned_facility_id: region.facility_ids,
+        month_date: period
+      )
+      .where(diabetes: "yes")
+      .where.not(htn_care_state: "dead")
```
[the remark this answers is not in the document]
govindkrjoshi: Yup, this is some tech debt. `htn_care_state` should have been called `care_state` since it's not dependent on the patients condition. But rename it would mean having to rename it in all the metabase reports also. 

Maybe we can create an duplicate column with the name `care_state` in future.

tfidfwastaken (approved):

PR #5412, opened 2024-04-01
data migration: remove old invalid BUHS drugs

tfidfwastaken:
BUHS' integration with the import API had a bug which caused duplicate drug data to be sent. We delete all the data sent before March 8 (which is around the time the error was fixed) as per their request. Thankfully, we don't have to be too exact with the date boundary, as there was no drug data sent between 17th January to 11th March.

You can confirm this yourself by running this query, which is generated from our migration's activerecord query, with only the date value changed to 1st April:
```sql
SELECT prescription_drugs.*
FROM prescription_drugs
INNER JOIN facilities ON facilities.deleted_at IS NULL
AND facilities.id = prescription_drugs.facility_id
INNER JOIN facility_groups ON facility_groups.deleted_at IS NULL
AND facility_groups.deleted_at IS NULL
AND facility_groups.id = facilities.facility_group_id
INNER JOIN organizations organization ON organization.deleted_at IS NULL
AND organization.id = facility_groups.organization_id
WHERE prescription_drugs.deleted_at IS NULL
  AND organization.id = 'afb485a8-094e-4d86-812f-0bac90934d89' -- org ID of BUHS
  AND (prescription_drugs.updated_at < '2024-04-1')
ORDER BY prescription_drugs.updated_at DESC; -- I added this bit
```

Since BUHS will resend all the drug data again, these should be safe to delete.

**Story card:** [sc-12178]

### DIFF
--- a/db/data/20240401110047_remove_invalid_buhs_drugs.rb
+++ b/db/data/20240401110047_remove_invalid_buhs_drugs.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+class RemoveInvalidBuhsDrugs < ActiveRecord::Migration[6.1]
+  ORG_ID = "afb485a8-094e-4d86-812f-0bac90934d89"
+
+  def up
+    unless CountryConfig.current_country?("Bangladesh") && SimpleServer.env.production?
+      return print "RemoveInvalidBuhsDrugs is only for production Bangladesh"
+    end
+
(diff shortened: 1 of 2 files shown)


PriyangaPKini (changes requested):
Data migration looks good 👍. 
Make sure run the data migration locally and commit the file `db/data_schema.rb`.  This saves the latest version of the data migration.

PR #5420, opened 2024-05-06
Initialise questionnaire forms one month in advance for ET

PriyangaPKini:
**Story card:** [sc-12507](https://app.shortcut.com/simpledotorg/story/12507/as-a-user-i-am-unable-to-see-the-screening-forms-in-ethiopia-for-the-current-month-at-the-beginning-of-the-month)

## Because

The screening forms appear only after a week in Ethiopia. 
The ET reporting period for the 8th month is `7/21/2016 - 8/20/2016` in the ET calendar (in the Gregorian calendar, it's `30/03/2024 - 28/04/2024`). They expect to see the 8th-month form by `8/20/2016`. Since we initialise the current month form on the first of every month in the Gregorian Calendar, the form for the 8th month appears only on `8-23-2016` (`01-05-2024` in the Gregorian Calendar). That's about a week after the start of the 8th month.

## This addresses

So we have decided to initialise forms two months in advance once. After that, we would initialise the form every month in ET for the next month. Whereas in other countries, we will continue to initialise the previous month form every month.

### DIFF
--- a/lib/tasks/questionnaires.rake
+++ b/lib/tasks/questionnaires.rake
@@ -4,7 +4,7 @@ namespace :questionnaires do
     # Due to mismatch in Gregorian<>Ethiopian calendars,
     # Dynamic forms initialization had a delay of 20+ days in Ethiopia.
     # To avoid this delay, we initialize dynamic forms 1 month in advance in Ethiopia.
-    date = CountryConfig.current_country?("Ethiopia") ? Date.current : 1.month.ago
+    date = CountryConfig.current_country?("Ethiopia") ? 1.month.from_now : 1.month.ago
 
     if Flipper.enabled?(:monthly_screening_reports)
       QuestionnaireResponses::MonthlyScreeningReports.new(date).seed
--- a/lib/tasks/questionnaires.rake
+++ b/lib/tasks/questionnaires.rake
@@ -2,8 +2,8 @@ namespace :questionnaires do
   desc "Initialize questionnaire responses"
   task initialize: :environment do
     # Due to mismatch in Gregorian<>Ethiopian calendars,
-    # Dynamic forms initialization had a delay of 20+ days in Ethiopia.
-    # To avoid this delay, we initialize dynamic forms 1 month in advance in Ethiopia.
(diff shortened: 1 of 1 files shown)


lib/tasks/questionnaires.rake
```diff
@@ -4,7 +4,7 @@ namespace :questionnaires do
     # Due to mismatch in Gregorian<>Ethiopian calendars,
     # Dynamic forms initialization had a delay of 20+ days in Ethiopia.
     # To avoid this delay, we initialize dynamic forms 1 month in advance in Ethiopia.
```
tfidfwastaken, line 6: ^ This comment needs to be updated to reflect the current situation.

From what I understand the main issue is:
* The Ethiopian calendar is broken up differently than the Gregorian calendar and thus out of sync, thus we need to initialise responses in advance.
* One month in advance is not sufficient, because Ethiopian months (ignoring Pagume) are 30 days, and Gregorian Months are usually 30 or 31 days, which creates a few days' delay between when the users expect the new form to initialise and when they actually initialise.
* Thus, 2 months in advance safely covers up for all potential delays.

Is this right?

PriyangaPKini: > One month in advance is not sufficient, because Ethiopian months (ignoring Pagume) are 30 days, and Gregorian Months are usually 30 or 31 days, which creates a few days' delay between when the users expect the new form to initialise and when they actually initialise.

Not exactly. The Ethiopian Calendar has 12 months of 30 days each, plus five or six additional days (sometimes known as the 13th month). More on ET time [here](https://ethiopianembassy.org/ethiopian-time/). 
Their reporting period for a month starts in a month and ends in another month. Whereas in Gregorian calendar, reporting period is starts and end in the same month. 
The ET reporting period for the 8th month is `7/21/2016 - 8/20/2016`.  In ET we initialise current month form at the start of Gregorian calendar which is  `8-23-2016`.  We will be initialising 8th month form on `8-23-2016` which is delayed by week. We thought we are be pre-initialising the forms in ET but in reality forms are created after the month with a weeks delay.

tfidfwastaken (approved):

PR #5425, opened 2024-05-27
Setup SMS reminders in India for June 2024 and July 2024

danySam:
**Story card:** [sc-12546](https://app.shortcut.com/simpledotorg/story/12546/setup-sms-reminders-in-ihci-states-in-india-for-the-month-of-may-and-june)

## Because

We want to set up SMS reminders for India for June 2024 and July 2024 in the state of West Bengal and the district of Chennai.

### DIFF
--- a/db/data/20240527121747_set_up_sms_reminder_india_2024_june_july.rb
+++ b/db/data/20240527121747_set_up_sms_reminder_india_2024_june_july.rb
@@ -7,7 +7,7 @@ class SetUpSmsRemindersIndiaJuneJuly2024 < ActiveRecord::Migration[6.1]
   ].flatten
 
   # Regions where Simple is active and we send sms reminders currently:
-  # States - All of West Bengal, selcted districts in Tamil Nadu
+  # States - All of West Bengal and select districts in Tamil Nadu
   # Districts - Chennai
   REGION_FILTERS = {"facilities" => {"include" => INCLUDED_FACILITY_SLUG}}
   PATIENTS_PER_DAY = 20_000
--- a/db/data/20240527121747_set_up_sms_reminder_india_2024_june_july.rb
+++ b/db/data/20240527121747_set_up_sms_reminder_india_2024_june_july.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+class SetUpSmsRemindersIndiaJuneJuly2024 < ActiveRecord::Migration[6.1]
+  INCLUDED_FACILITY_SLUG = [
+    Facility.where(district: "Chennai").pluck(:slug),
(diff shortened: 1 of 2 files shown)


db/data/20240527121747_set_up_sms_reminder_india_2024_june_july.rb
```diff
@@ -0,0 +1,64 @@
...
+  PATIENTS_PER_DAY = 20_000
+  EXPERIMENTS_DATA = (6..7).map do |month_number|
+    month = Date::ABBR_MONTHNAMES[month_number]
+    {
+      start_time: DateTime.new(2024, month_number).beginning_of_month,
+      end_time: DateTime.new(2024, month_number).end_of_month,
+      current_patients_experiment_name: "Current Patient #{month} 2024",
+      stale_patients_experiment_name: "Stale Patient #{month} 2024"
+    }
+  end
```
tfidfwastaken, line 22: [nitpick] This is correct, but the number bounds and ranges make me a bit nervous (have to look up always to remember which sides are inclusive and if months start from zero or one).

This is how we have done it in the last couple of these data migrations:
https://github.com/simpledotorg/simple-server/blob/9797949b44290be14a8a278500ef2652f8ab8faf/db/data/20240425072432_set_up_sms_reminders_bd_may_oct_2024.rb#L4-L10

db/data/20240527121747_set_up_sms_reminder_india_2024_june_july.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+class SetUpSmsRemindersIndiaJuneJuly2024 < ActiveRecord::Migration[6.1]
+  INCLUDED_FACILITY_SLUG = [
+    Facility.where(district: "Chennai").pluck(:slug),
+    Facility.where(state: "West Bengal").pluck(:slug)
+  ].flatten
+
+  # Regions where Simple is active and we send sms reminders currently:
+  # States - All of West Bengal, selcted districts in Tamil Nadu
```
tfidfwastaken, line 10: [typo]
```suggestion
  # States - All of West Bengal and select districts in Tamil Nadu
```

tfidfwastaken (approved):
LGTM, put a nitpick but ok to ship this as is also.

I haven't checked if the cascade and reminder templates used here are correct, but I am assuming they are.

db/data/20240527121747_set_up_sms_reminder_india_2024_june_july.rb
danySam: This makes sense. Applying the change 👍

PR #5427, opened 2024-05-29
Introduce Patients Protected and Hypertension Cascade Cards on the Dashboard

danySam:
**Story card:** [sc-12629](https://app.shortcut.com/simpledotorg/story/12629/implement-patients-protected-and-hypertension-treatment-cascade-numbers-on-the-simple-dashboard-data-layer)

## Because

We want to introduce the patients-protected and HTN cascade cards to the HTN dashboard part of the [big positive number epic](https://app.shortcut.com/simpledotorg/epic/11405).

## This addresses

We introduce 2 new cards to the HTN dashboard. 
1. The patients protected card, shows the 36-month trend of controlled HTN patients numbers
2. The HTN cascade card, shows the estimated HTN population in a region and the cascaded information on what percentage of that population are:
    - % Registered with us
    - % Patients under care
    - % Patients with controlled BP

### HTN Cascade
- We'll only show this at a district level and above
- We'll roll up the district numbers to regions at higher hierarchy
- We'll show an information box to nudge users to add the estimated population with HTN numbers if there's no information for that region or partial information (eg: a state can have all but one district with HTN estimates)
- HTN Cascade is a sub-feature and will be rolled out individually
- When HTN Cascade need not be shown, we will expand the patients-protected screen to fit the screen width

## Test instructions

- Enable the following [feature flags](./docs/wiki/feature-flags.md)
  - `patients_protected`
  - `patients_protected_htn_cascade`

## Screenshots
Note: Seed data was used to create these screenshots. They're not real-world data.

### District Level
District Level

### State Level
State Level

### Block Level
Block Level

### State Level with no HTN estimate
State Level with no HTN estimate


app/components/dashboard/hypertension/hypertension_cascade_component.html.erb
```diff
@@ -0,0 +1,97 @@
...
+                            <div class="coverage-bar">
+                                <div class="overage-bar-fill bg-grey-light" style="height: 100%;">
+                                    <p class="coverage-estimated c-grey-dark">100%</p>
+                                </div>
+                            </div>
+                            <p class="mb-0px">
+                            <%= show_estimate? ? @estimated_population.population : "?" %>
+                            </p>
+                            <p class="text-grey fs-lg-14px mb-0px">
+                                Estimated people with hypertension
```
PriyangaPKini, line 43: q: These copies need to be localised. Are we not doing it here because of the any time constraints to release this feature?

app/components/dashboard/hypertension/hypertension_cascade_component.rb
```diff
@@ -0,0 +1,38 @@
+class Dashboard::Hypertension::HypertensionCascadeComponent < ApplicationComponent
+  attr_reader :data, :contactable, :period
+
+  FIXED_RATE_WHEN_NO_ESTIMATE = "2%" # We should a fixed rate when there's no esimate for the region
```
PriyangaPKini, line 4: [nit-pick]: Need to fix the typo in the comment.

PriyangaPKini: [question]: Does this mean  we need to show a 2% rate when there is no estimated population for a district? Looks like it's not adding that fixed rate to the card.
image

app/components/dashboard/hypertension/patients_protected_component.html.erb
```diff
@@ -0,0 +1,11 @@
...
+    <%= c.title(title: "Patients protected from heart attacks & strokes in " + @region.name, 
+    subtitle: t("bp_controlled_copy.reports_card_subtitle")) do |title| %>
+      <%= title.tooltip({ "Definition" => t("bp_controlled_copy.numerator") }) %>
+    <%end%>
+    <%= c.summary do %>
+      <div class="mb-12px d-lg-flex align-lg-center">
+        <p class="c-print-black c-green-dark graph-percent fs-28px" data-key="patientsProtectedWithSuffix" ></p>
+      </div>
+    <%end%>
+  <%end%>
```
PriyangaPKini, line 11: [nit-pick]: Add an extra line at the end of the file

app/components/dashboard/patient_breakdown_component.html.erb
```diff
@@ -26,21 +30,12 @@
...
-    <p class="mb-8px">
-      Dead patients
-    </p>
-    <p class="mb-8px" id="dead-patients">
-      <%= number_with_delimiter(data[:dead_patients]) %>
-    </p>
-  </div>
 </div>
 <div class="d-flex ai-center jc-between">
-  <p class="mb-8px fw-bold">
```
PriyangaPKini, line 39: [question]: What the reason behind removing the cumulative registrations from the patient coverage graph?

PriyangaPKini (changes requested):
Left some questions and comments.

app/components/dashboard/hypertension/hypertension_cascade_component.rb
danySam: We don't need to show the actual percentage, since there's no denominator to get the %. We just set the graphs to 2% for a visual cue that nothing compares.

app/components/dashboard/hypertension/hypertension_cascade_component.html.erb
danySam: Cmiiw, my understanding is that we only localize tooltips. I can see a lot of non-localized content on the dashboard

app/components/dashboard/hypertension/hypertension_cascade_component.rb
PriyangaPKini: Gotcha!

app/components/dashboard/hypertension/hypertension_cascade_component.html.erb
PriyangaPKini: Just check this. Looks like we only localise tooltip copies and subtitles for certain graphs

app/components/dashboard/patient_breakdown_component.html.erb
danySam: Because we're going to show this at the to with the new graph

app/components/dashboard/hypertension/hypertension_cascade_component.html.erb
danySam: Checked this. Only the webviews need to be localized.

PR #5431, opened 2024-06-10
Fix the offset in ltfu_since_date calculation

PriyangaPKini:
**Story card:** [sc-12779](https://app.shortcut.com/simpledotorg/story/12779/fix-the-ltfu-since-date-in-the-simple-dashboard)

## Because

The `ltfu_since_date` in the dashboard was off by a month

## This addresses

Fix the offset in the `ltfu_since_date` calculation

### DIFF
--- a/app/models/period.rb
+++ b/app/models/period.rb
@@ -94,7 +94,7 @@ class Period
   end
 
   def ltfu_since_date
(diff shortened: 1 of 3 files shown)


tfidfwastaken (approved):
LGTM!

PR #5436, opened 2024-07-03
Fix: Patients protected does not show last 3Y data

danySam:
**Story card:** [sc-12887](https://app.shortcut.com/simpledotorg/story/12887/bug-patients-protected-not-showing-last-36-months-of-data)

## Because

On the dashboard, we only show 18 months of data. Reports::MAX_MONTHS_OF_DATA configures this. Changing this value, however, will make all the charts on the dashboard display 36 months of data.

## This addresses

We're creating a new repository object in the patients-protected component to override the max number of months value.

## Test instructions

Enable the `patients_protected` feature flag to see this graph

### DIFF
--- a/app/components/dashboard/hypertension/patients_protected_component.rb
+++ b/app/components/dashboard/hypertension/patients_protected_component.rb
@@ -1,16 +1,15 @@
 class Dashboard::Hypertension::PatientsProtectedComponent < ApplicationComponent
-  attr_reader :data, :contactable, :period, :period_start
+  attr_reader :period, :period_start
 
-  def initialize(region:, data:, period:)
+  def initialize(region:, period:)
     @region = region
-    @data = data
     # Show only data up to last month
     @period = period.advance(months: -1)
     @period_start = period.advance(months: -36)
+    @repo = Reports::Repository.new(@region, periods: Range.new(@period_start, @period))
   end
 
   def graph_data
-    patients_protected = fill_empty_period_data(:controlled_patients)
     {
       patientsProtected: patients_protected,
       patientsProtectedWithSuffix: patients_protected.map { |k, v| [k, "#{v} patients"] }.to_h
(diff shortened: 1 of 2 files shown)


app/components/dashboard/hypertension/patients_protected_component.rb
```diff
@@ -1,16 +1,15 @@
...
-  attr_reader :data, :contactable, :period, :period_start
+  attr_reader :period, :period_start
 
-  def initialize(region:, data:, period:)
+  def initialize(region:, period:)
     @region = region
-    @data = data
     # Show only data up to last month
     @period = period.advance(months: -1)
     @period_start = period.advance(months: -36)
```
danySam, line 8: I couldn't think of a cleaner way to make this work. The repository is already initialized in the region controller, however, the `MAX_MONTHS_OF_DATA` is configured for all the charts. You can't get a single chart to have a different number of months to display without affecting other charts. Or you'll have to make the repository much more composable (to accept different reporting periods for different columns), which is not something we can dedicate our time to at the moment.

app/components/dashboard/hypertension/patients_protected_component.rb
```diff
@@ -1,16 +1,15 @@
...
+    @repo = Reports::Repository.new(@region, periods: Range.new(@period_start, @period))
   end
 
   def graph_data
-    patients_protected = fill_empty_period_data(:controlled_patients)
     {
       patientsProtected: patients_protected,
       patientsProtectedWithSuffix: patients_protected.map { |k, v| [k, "#{v} patients"] }.to_h
```
PriyangaPKini, line 15: [nit-pick]: you could #transform_values instead of #map
```suggestion
      patientsProtectedWithSuffix: patients_protected.transform_values { |v| "#{v} patients" }
```

PriyangaPKini (approved):
Looks good to me

app/components/dashboard/hypertension/patients_protected_component.rb
PriyangaPKini: Acknowledge

PR #5438, opened 2024-07-03
Address failing specs due to timezones

tfidfwastaken:
## Because

Running our specs between midnight and 5:30 AM IST causes a lot of breakage.

## This addresses

Our suite has many specs that fail when run at odd hours due to inconsistent application of timezone reporting helpers. One common anti-pattern is to set up the test in UTC but make the assertions within a block that sets the timezone to the configured reporting time zone. We move the setup within this block where necessary.

For `RegionSummarySchema`, the helpers that align the spec timezone with the database timezone are not used at all, we fix these as well.

For `MaterializedPatientSummaries`, between 00:00 to 05:30 AM IST, the follow-up days have one extra day added to them (due to `date_trunc()` in the matview definition messing up the time boundaries). This edge case is a very rare one, and until we prioritise fixing it, we mark it as a known bug in our test suite so that it doesn't fail when run at these times.

## Test instructions

Set your system clock to a time between midnight and 5:30 AM IST, run all the specs and check if they all pass.

### DIFF
--- a/spec/exporters/patients_with_history_exporter_spec.rb
+++ b/spec/exporters/patients_with_history_exporter_spec.rb
@@ -340,7 +340,12 @@ RSpec.describe PatientsWithHistoryExporter, type: :model do
   describe "#csv" do
     it "generates a CSV of patient records" do
       Timecop.freeze do
-        timestamp = ["Report generated at:", Time.current]
+        now = Time.current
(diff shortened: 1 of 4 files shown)


PriyangaPKini (approved):
LGTM

PR #5439, opened 2024-07-10
Fix incorrect exporter mapping in Bangladesh

tfidfwastaken:
## Because

The DHIS2 disaggregated diabetes exporter relies on the `bangladesh-production.yml` file to determine which data elements the values should be sent to. The lookup would break in production because the keys on the backend do not exactly match the keys in the YAML file.

**Story card:** Part of [sc-12888](https://app.shortcut.com/simpledotorg/story/12888/turn-on-flag-to-schedule-monthly-transfer-of-dm-data-bangladesh)

## This addresses

 We unify the names and make them consistent with the key names convention established in the hypertension exporter.


aagrawalrtsl (approved):
Looks good to me. We can just fix the failing specs and merge it.

tfidfwastaken:
The spec failures are caused by unrelated issues that will be addressed by #5438 

I'll merge this after the spec failure fixes are merged.

PR #5440, opened 2024-07-11
Fixed the padding-right on patients protected card

dburka:
Fixes a tiny padding bug on the Patients Protected card when the Treatment Cascade is not visible.

**Story card:** NONE

## Because

The Patients Protected card had a small extra padding on the right when the Cascade isn't displayed.

## This addresses

Uses Flipper to only have PR-2 padding when the cascade is visible.

## Test instructions

None


danySam (approved):
Added a fix to remove the padding at facility level as well

PR #5443, opened 2024-07-16
Integrate Mobitel as an SMS provider for SL

danySam:
**Story card:** [sc-12889](https://app.shortcut.com/simpledotorg/story/12889/integrate-mobitel-api-for-lk)

## Because

Mobitel is a new SMS vendor that we'll be using in SL.

### Note
- Mobitel does not have a delivery status, once the API is sent, it can only be perceived as delivered.
- There are no callback APIs, or status check APIs to check if the SMS is delivered. So we'll consider all SMSs that passed the API validation checks successful.

## Test instructions

The following ENV variables should be added to test out this vendor
- `MOBITEL_API_USERNAME`
- `MOBITEL_API_PASSWORD`
- `MOBITEL_SMS_ALIAS`

You can find the credentials in our vault

### DIFF
--- a/app/models/mobitel_delivery_detail.rb
+++ b/app/models/mobitel_delivery_detail.rb
@@ -1,0 +1,29 @@
+class MobitelDeliveryDetail < DeliveryDetail
+  # Mobitel APIs are not capable of providing delivery details.
+  # Hence we consider all the requests that passed the API validation as successful
+  def successful?
+    true
+  end
+
+  def unsuccessful?
+    false
+  end
+
+  def in_progress?
+    false
+  end
+
+  def self.create_with_communication!(message:, recipient_number:)
+    ActiveRecord::Base.transaction do
+      delivery_detail = create!(
+        message: message,
+        recipient_number: recipient_number
+      )
+
+      Communication.create!(
+        communication_type: Messaging::Mobitel::Sms.communication_type,
+        detailable: delivery_detail
+      )
+    end
+  end
+end
--- a/spec/services/messaging/mobitel/sms_spec.rb
+++ b/spec/services/messaging/mobitel/sms_spec.rb
@@ -2,23 +2,6 @@ require "rails_helper"
 
 RSpec.describe Messaging::Mobitel::Sms do
   describe "#send_message" do
-    it "passes sanitised variable content to the API" do
-      mock_api = double("MobitelApiDouble")
-      allow(Messaging::Mobitel::Api).to receive(:new).and_return(mock_api)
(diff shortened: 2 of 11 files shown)


app/models/mobitel_delivery_detail.rb
```diff
@@ -0,0 +1,29 @@
+class MobitelDeliveryDetail < DeliveryDetail
+  # Mobitel APIs are not capable of providing delivery details.
+  # Hence we consider all the requests that passed the API validation as successful
+  def successful?
+    true
+  end
```
tfidfwastaken, line 6: [suggestion] Can we return `false` here? That is also equally misleading as a contract (the return status is neither `true` or `false`, it's `i don't know`.

But the benefit of returning false is that we seem to have code paths that are able to handle `successful`, `unsuccessful` and `in_progress` being all false and assuming an `unknown` communication result:
https://github.com/simpledotorg/simple-server/blob/1cee413645a3dc8baa90cfebc1e86e5214e015b7/app/models/communication.rb#L54-L64

But alas, this won't be enough—this kind of logic is not handled properly in the Experiment Notification model:
https://github.com/simpledotorg/simple-server/blob/f51d1510bafe2aa668e3b1e4ac31d7a282747d11/app/models/notification.rb#L68-L90

I wonder if we can fix the above logic. The result I'm trying to avoid is to have records in our DB that show a message is successfully sent when it might not be.

danySam: I get your point. We can't say that a message is successful in binary anymore. However, with most SMS vendors, we rely on the information they provide as the truth.

spec/services/messaging/mobitel/sms_spec.rb
```diff
@@ -0,0 +1,69 @@
+require "rails_helper"
+
+RSpec.describe Messaging::Mobitel::Sms do
+  describe "#send_message" do
+    it "passes sanitised variable content to the API" do
```
tfidfwastaken, line 5: question: What is "sanitised variable content"? What does unsanitised content look like, and what's the expected behaviour?

tfidfwastaken (approved):
Left a question, but approving because everything else looks solid 🪨

spec/services/messaging/mobitel/sms_spec.rb
danySam: Whoops. This test is not necessary. I had removed it, must have added it back when trying to fix the bad merge.

We sanitise the SMS variables for BSNL. We don't do that for other vendors.

PR #5445, opened 2024-07-19
Add index to patient follow-ups

tfidfwastaken:
**Story card:** [sc-12988](https://app.shortcut.com/simpledotorg/story/12988/add-indexes-to-reporting-facility-follow-ups)

## Because

Our facility reports are slow due to there being only a suboptimal compound index on the patient follow ups matview.

## This addresses

We add an index on the facility ID which when analysed on Sandbox offered better query plans and performance characteristics—it used a Bitmap Index Scan along with a Bitmap Heap Scan, instead of a slow Index Scan that read too many blocks.

### Before the migration
image
image

### After the migration that adds index on `facility_id`

image
image

### Other approaches evaluated

I tried the following compound indexes as well, but the query plans, cost and times were not all that different:

* (facility_id, user_id)
* (facility_id, hypertension, user_id)
* (facility_id, user_id, month_date)

For the first two, the `user_id` didn't appear to be used at all, and for the latter, the index was utilised properly, but it didn't make a huge difference, it was also in the <100ms ballpark.

### DIFF
--- a/db/migrate/20240719202605_add_index_to_reporting_patient_follow_ups.rb
+++ b/db/migrate/20240719202605_add_index_to_reporting_patient_follow_ups.rb
@@ -1,0 +1,5 @@
+class AddIndexToReportingPatientFollowUps < ActiveRecord::Migration[6.1]
+  def change
+    add_index :reporting_patient_follow_ups, :facility_id
+  end
+end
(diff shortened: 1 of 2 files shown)


PriyangaPKini (approved):
LGTM ⚡

PR #5434, opened 2024-06-20
Setup SMS reminders for Sri Lanka for June July 2024

danySam:
**Story card:** [sc-12668](https://app.shortcut.com/simpledotorg/story/12668/set-up-sms-reminders-in-sri-lanka)

## Because

We want to setup SMS reminders with the following configurations for the months of June and July 2024

- Cascade on day -1 and 3 after scheduled appointment
- Templates
  - one_day_before_appointment
  - three_days_missed_appointment

## Test instructions

Run `bin/rails data:migrate` to run this migration.

PR #5447, opened 2024-07-25
Setup test SMS remidners in SL and switch vendor to Mobitel

danySam:
**Story card:** [sc-12899](https://app.shortcut.com/simpledotorg/story/12899/test-sms-in-dh-gornaduwa)

## Because

We want to test the new SMS vendor: Mobitel in Sri Lanka with a facility (DH Gonaduwa) where we can monitor the experiment closely.

## Notes
- We are also switching the vendor from Twilio to Mobitel in this PR

### DIFF
--- a/config/initializers/country_config.rb
+++ b/config/initializers/country_config.rb
@@ -75,7 +75,7 @@ class CountryConfig
       sms_country_code: ENV["SMS_COUNTRY_CODE"] || "+94",
       supported_genders: %w[male female],
       patient_line_list_show_zone: false,
-      appointment_reminders_channel: "Messaging::Twilio::ReminderSms",
+      appointment_reminders_channel: "Messaging::Mobitel::Sms",
       sorted_facility_sizes: %w[large medium small community teaching_hospital national_hospital],
       facility_sizes: {
         community: "community",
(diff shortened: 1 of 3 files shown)


tfidfwastaken:
Question:
* Would this change create a clash with the change in #5434 for Gonaduwa and cause two experiments to run there at the same time? Because if it does, it might affect enrollments and lead to high experiment eviction rates.

Suggestions:
* Should we approach this with a Flipper flag to use Mobitel (partially enabled only for Gonaduwa), and reuse the ongoing experiment? That could spare us from the unwieldiness of data migrations.

danySam:
> Question:
> 
>     * Would this change create a clash with the change in [Setup SMS reminders for Sri Lanka for June July 2024 #5434](https://github.com/simpledotorg/simple-server/pull/5434) for Gonaduwa and cause two experiments to run there at the same time? Because if it does, it might affect enrollments and lead to high experiment eviction rates.
> 
> 
> Suggestions:
> 
>     * Should we approach this with a Flipper flag to use Mobitel (partially enabled only for Gonaduwa), and reuse the ongoing experiment? That could spare us from the unwieldiness of data migrations.

PR #5454, opened 2024-08-05
Fix HTN Cascade formatting error when estimated is present

danySam:
**Story card:** [sc-12856](https://app.shortcut.com/simpledotorg/story/12856/changes-to-patients-protected-and-htn-cascade-components)

## Because

The HTN Cascade card not loading when a population estimate is present

## This addresses

Fixed by moving the string formatting to the views


tfidfwastaken (commented):
LGTM

PR #5457, opened 2024-08-09
Fix: Mobitel unable to send multi-line SMS

danySam:
**Story card:** [sc-13084](https://app.shortcut.com/simpledotorg/story/13084/bug-multiline-smss-not-working-with-mobitel)

## Because

The old Mobitel API does not support multi-line SMS body.

## This addresses

We're replacing that with a new API provided by the Mobitel team.


PriyangaPKini (approved):
LGTM ⚡

PR #5459, opened 2024-08-14
changing numerator in bs < 200 and over 200 charts to include only fa…

aagrawalrtsl:
…sting and hba1c measurements

**Story card:** https://app.shortcut.com/simpledotorg/story/13103/rename-blood-sugar-200-299-or-300-to-blood-sugar-uncontrolled
https://app.shortcut.com/simpledotorg/story/13102/rename-blood-sugar-200-card
https://app.shortcut.com/simpledotorg/story/13104/update-blood-sugar-controlled-card

## Because

Sri Lanka wanted to change the numerator in diabetes controlled and uncontrolled chart to include only fasting and hba1c measurements, unlike the current implementation where it includes random measurements as well.

## This addresses

Changes are put behind a feature flag "use_who_standard_for_diabetes_indicator". If this is disabled the indicators would behave as before. 
All the text changes for the dashboard, mentioned in the stories, are locale based and would be effective when DEFAULT_PREFERRED_DASHBOARD_LOCALE is set to "en-LK".
Separate methods have been created to be used based to on the feature flag to only include fasting and hba1c readings for the numerator.

## Test instructions

Enable use_who_standard_for_diabetes_indicator
Set DEFAULT_PREFERRED_DASHBOARD_LOCALE to "en-LK"
Check the diabetes controlled and uncontrolled charts in Simple dashboard, it should now include only fasting and hba1c readings.


app/components/dashboard/diabetes/bs_below200_graph_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <div id="bs-below-200" class="d-lg-flex w-lg-50 pr-lg-2">
   <%= render Dashboard::Card::GraphComponent.new(id: "bsBelow200PatientsTrend", data: graph_data, period: period) do |c| %>
-    <%= c.title(title: "Blood sugar &lt;200".html_safe, subtitle: t("bs_below_200_copy.reports_card_subtitle", region_name: region.name)) do |title| %>
+    <%= c.title(title: t("bs_below_200_copy.reports_card_title").html_safe, subtitle: t("bs_below_200_copy.reports_card_subtitle", region_name: region.name)) do |title| %>
```
danySam, line 3: Is the copy specific for SL? Else, we can tie this in with the same Flipper flag

aagrawalrtsl: the key "bs_below_200_copy" is already present in the default en.yml file. This text was translated though, here I have just added translation for it and added the updated needed for SL in en-LK.yml
This locale will be handled through DEFAULT_PREFERRED_DASHBOARD_LOCALE variable

tfidfwastaken:
image

question: Does this change cover cohort reports? The tooltip says that the numerator only counts FBS ≥ 200 but the label says Blood Sugar ≥ 300. Same for the other labels.

tfidfwastaken:
image

issue: The definitions at the bottom of the page for SL should have FBS-appropriate labels (≤126, 126-199, ≥200)

aagrawalrtsl:
> image
> **question**: Does this change cover cohort reports? The tooltip says that the numerator only counts FBS ≥ 200 but the label says Blood Sugar ≥ 300. Same for the other labels.

I did not make changes to the cohort reports, but I think the same translations keys are being used for other labels as well, should we change those also.

spec/models/reports/repository_spec.rb
```diff
@@ -802,4 +802,125 @@
       expect(repo.cumulative_registrations[facility_1.slug]).to eq({})
     end
   end
+
+  context "constant values" do
```
tfidfwastaken, line 806: nice addition to the specs 👍

spec/presenters/reports/repository_presenter_spec.rb
```diff
@@ -1,9 +1,125 @@
...
+        :patients_returned_with_result_agreed_to_visit_rates,
+        :patients_returned_with_result_remind_to_call_later_rates,
+        :patients_returned_with_result_removed_from_list_rates,
+        :contactable_patients_returned_after_call,
+        :contactable_patients_returned_after_call_rates,
+        :contactable_patients_returned_with_result_agreed_to_visit_rates,
+        :contactable_patients_returned_with_result_remind_to_call_later_rates,
+        :contactable_patients_returned_with_result_removed_from_list_rates
+      ]
+      expect(subject.keys).to match_array(expected_keys)
```
tfidfwastaken, line 106: **quibble (non-blocking)**: I might just inline `subject` so that I don't have to scroll up to see what it expands to.

spec/presenters/reports/repository_presenter_spec.rb
tfidfwastaken, line 1: **note (non-blocking)**: We could additionally have a spec for `Dashboard::Card::GraphComponent`. We don't seem to have specs in place for most components, except this one:
https://github.com/simpledotorg/simple-server/blob/09e6b15747db9b2885b49e6d1e42af4f2ad2875a/spec/components/dashboard/patient_breakdown_component_spec.rb

For the scope of this PR, the minimum viable spec would be to test if the correct labels are being used when the flag is enabled. Once the spec file is in place, anyone else who touches the component can add more exhaustive specs later.

But this is a nice-to-have, not a blocker for merging to me.

app/components/dashboard/diabetes/bs_below200_graph_component.html.erb
```diff
@@ -1,7 +1,8 @@
 <div id="bs-below-200" class="d-lg-flex w-lg-50 pr-lg-2">
   <%= render Dashboard::Card::GraphComponent.new(id: "bsBelow200PatientsTrend", data: graph_data, period: period) do |c| %>
-    <%= c.title(title: "Blood sugar &lt;200".html_safe, subtitle: t("bs_below_200_copy.reports_card_subtitle", region_name: region.name)) do |title| %>
-      <%= title.tooltip({ "Numerator" => t("bs_below_200_copy.numerator"),
+    <%= c.title(title: (Flipper.enabled?(:use_who_standard_for_diabetes_indicator) ? t("bs_below_200_copy.reports_card_title_fbs") : t("bs_below_200_copy.reports_card_title")).html_safe,
```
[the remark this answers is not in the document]
tfidfwastaken: **chore**: Document the feature flag in this file: https://github.com/simpledotorg/simple-server/blob/09e6b15747db9b2885b49e6d1e42af4f2ad2875a/doc/wiki/feature-flags.md#L1-L0

**quibble (non-blocking)**: The feature flag could have a more concise name, like `diabetes_who_standard_indicator` (in my head, flags are nouns, not verbs).

tfidfwastaken (changes requested):
Nice work figuring out a good approach for getting this to work, these parts of the code tend to be quite a maze.

I agree with the approach of using feature flags over localisation to change the labels since we are changing the behaviour of the indicator, not just a label for a single country.

I have left some refactoring suggestions, which I think should be addressed.

tfidfwastaken:
> image
> 
> **question**: Does this change cover cohort reports? The tooltip says that the numerator only counts FBS ≥ 200 but the label says Blood Sugar ≥ 300. Same for the other labels.

I took the latest pull of your code, and I'm not able to reproduce this anymore. So you can ignore this, seems like it got fixed at some point?

app/models/reports/region_summary_schema.rb
```diff
@@ -721,6 +769,22 @@ def diabetes_appts_scheduled_rates(entry)
...
+        visited_without_bs_taken_rates: visited_without_bs_taken[entry.region.slug][entry.period],
+        bs_below_200_rates_random: adjusted_diabetes_under_care(:bs_below_200, :random)[entry.region.slug][entry.period],
+        bs_below_200_rates_post_prandial: adjusted_diabetes_under_care(:bs_below_200, :post_prandial)[entry.region.slug][entry.period],
+        bs_200_to_300_rates_random: adjusted_diabetes_under_care(:bs_200_to_300, :random)[entry.region.slug][entry.period],
+        bs_200_to_300_rates_post_prandial: adjusted_diabetes_under_care(:bs_200_to_300, :post_prandial)[entry.region.slug][entry.period],
+        bs_over_300_rates_random: adjusted_diabetes_under_care(:bs_over_300, :random)[entry.region.slug][entry.period],
+        bs_over_300_rates_post_prandial: adjusted_diabetes_under_care(:bs_over_300, :post_prandial)[entry.region.slug][entry.period]
+      })
+    end
+
```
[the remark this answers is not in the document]
aagrawalrtsl: This method is needed because of how rounded_percentages is implemented. In order for it to calculate percentages correctly, it needs to have a balanced numerator and denominator part. Since in this we are only including fasting and hba1c patient counts in the numerator, but the denominator contains all the measurements, adding the rest of the measurements in the hash makes up for it. 
Marking this as resolved.

spec/presenters/reports/repository_presenter_spec.rb
aagrawalrtsl: added specs for individual components

spec/presenters/reports/repository_presenter_spec.rb
aagrawalrtsl: done

danySam:
> Set DEFAULT_PREFERRED_DASHBOARD_LOCALE to "en-LK"
AFAIK locales are not used on the dashboard except for web pages for mobile views. We should be using alternative keys in the main language file instead of using "en-LK" because this is not a LK specific change but rather using "WHO-recommended standard"

config/locales/en_LK.yml
```diff
@@ -49,3 +49,31 @@ en-LK:
...
+      numerator_fbs: "Patients with FBS 126-199, or HbA1c 7.0-8.9 at their last visit in the last 3 months"
+      summary: "patients with a blood sugar 200-299 from"
+      summary_fbs: "patients with a blood sugar 126-199 from"
+    bs_over_300:
+      title: "Blood sugar ≥300"
+      title_fbs: "Blood sugar ≥200"
+      numerator: "Patients with RBS/PPBS ≥300, FBS ≥200, or HbA1c ≥9.0 at their last visit in the last 3 months"
+      numerator_fbs: "Patients with FBS ≥200, or HbA1c ≥9.0 at their last visit in the last 3 months"
+      summary: "patients with a blood sugar ≥300 from"
+      summary_fbs: "patients with a blood sugar ≥200 from"
```
danySam, line 79: What I said here: https://github.com/simpledotorg/simple-server/pull/5459#issuecomment-2314444759

tfidfwastaken (approved):

PR #5462, opened 2024-08-21
HTN cascade: Aggregate estimate population count at org level

danySam:
**Story card:** [sc-13213](https://app.shortcut.com/simpledotorg/story/13213/fix-htn-estimate-not-working-summing-up-at-the-org-level)

## Because

Reports-view at the org level with estimate set for all sub-regions was throwing an error because `EstimatedPopulation` doesn't roll up at the org level

## This addresses

We'll roll up the estimated population for Org-level regions on demand.


PriyangaPKini (approved):
lgtm ⚡

PR #5467, opened 2024-08-30
Added metrics to prometheus

Gyan-Gupta-Rtsl:
**Story card:** [sc-12274]

## Because

Added metrics to prometheus

## This addresses

It provides metrics to prometheus such as sendgrid_http_response_time, sendgrid_total_emails,sendgrid_emails_remaining_count,sendgrid_email_used_count,sendgrid_plan_limit_expire,sendgrid_email_limit_exceeded_by_threshold_limit

## Test instructions
Need to configure prometheus, alertmanager and check that metrics are exposed to prometheus.

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -10,13 +10,13 @@ class MetricsController < ApplicationController
       render plain: "Error fetching metrics: #{result[:error]}", status: :unprocessable_entity
     else
       metrics = {
-        sendgrid_total_emails: result[:total],
+        sendgrid_email_limit_count: result[:total],
         sendgrid_emails_remaining_count: result[:remain],
         sendgrid_email_used_count: result[:used],
         sendgrid_plan_limit_expire: result[:plan_completion_status],
         sendgrid_email_limit_exceeded_by_threshold_limit: result[:exceeded_limit_status],
-        sendgrid_http_return_code: result[:http_return_code],
-        sendgrid_http_response_time: result[:http_response_time].round(2)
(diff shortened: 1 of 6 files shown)


app/controllers/metrics_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# MetricsController is responsible for handling requests related to metrics,
+# including exposing SendGrid-related metrics in a format compatible with Prometheus.
+class MetricsController < ApplicationController
+  def index # rubocop:disable Metrics/MethodLength
+    result = SendgridService.new.check_credits
```
tfidfwastaken, line 7: **issue**: The Prometheus instance is scraping the `/metrics` endpoint every 30s
(I verified here: https://prometheus.simple.org/config)

As such, Prometheus scrapes are always happening at very fast intervals—we don't need to shoot off API calls to collect balance for every scrape, and we may risk hitting rate limits as well.

**suggestion**: Instead, we can setup a scheduled job that will observe the gauge every hour (see how we check SMS balances in `scheduler.rb`)

tfidfwastaken (changes requested):
Requesting changes:

* Let's use the abstraction setup with `PrometheusExporter` that already exists in our codebase
* Let's update the balance and statuses at a fixed longer interval, rather than at every scrape of the metrics endpoint

Let me know if you have any questions!

app/controllers/metrics_controller.rb
Gyan-Gupta-Rtsl: @tfidfwastaken Arnaud asked me to do this and not to use schdeuled job for service.

PR #5468, opened 2024-09-02
updating feature flag to work conditionally with the user

aagrawalrtsl:
**Story card:** https://app.shortcut.com/simpledotorg/story/13321/make-feature-flag-take-logged-in-user-into-consideration

## Because

We wanted the feature flag for global diabetes indicator to work with logged in users, for making it easier to test on prod before making it live for everyone

## This addresses

updating diabetes_who_standard_indicator, wherever it's used, to work with logged in user

## Test instructions

Login to Simple dashboard -> Go to Flipper UI and enable diabetes_who_standard_indicator flag only for an actor. 
Log in with the same actor.
Feature flag should work only for the actors added and not for other users.

### DIFF
--- a/app/presenters/reports/repository_presenter.rb
+++ b/app/presenters/reports/repository_presenter.rb
@@ -1,19 +1,18 @@
 module Reports
   class RepositoryPresenter < SimpleDelegator
-    def self.create(regions, period:, months: Reports::MAX_MONTHS_OF_DATA, current_admin: nil)
+    def self.create(regions, period:, months: Reports::MAX_MONTHS_OF_DATA, use_who_standard: nil)
       start_period = period.advance(months: -(months - 1))
       range = Range.new(start_period, period)
-      repo = Reports::Repository.new(regions, periods: range, current_admin: current_admin)
+      repo = Reports::Repository.new(regions, periods: range, use_who_standard: use_who_standard)
       new(repo)
     end
 
-    def call(region)
-      to_hash(region)
+    def call(region, use_who_standard = nil)
+      to_hash(region, use_who_standard)
     end
 
-    def to_hash(region)
+    def to_hash(region, use_who_standard = nil)
       slug = region.slug
-      use_who_standard = Flipper.enabled?(:diabetes_who_standard_indicator, current_admin)
       {
         adjusted_patient_counts_with_ltfu: adjusted_patients_with_ltfu[slug],
         adjusted_patient_counts: adjusted_patients_without_ltfu[slug],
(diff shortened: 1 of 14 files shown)


tfidfwastaken (changes requested):
**issue**: The following classes all now have a conceptual knowledge of what an _admin user_ is:

* All the diabetes component classes (like `BsBelow200GraphComponent` etc) and footnote views
* The reports repository and repository presenter

This is a violation of single responsibility principle,
* Graph components should only have knowledge on how to render graphs, it should not know about any details about the user models
* Repository/Repository Presenter should only know about the details of data calculations and matviews, it should not know anything about user models/admins

**suggestion**:

Knowledge of the admin user should be left to the control plane of the web application, in this case, our regions controller only. Instead of delegating the current admin from the control plane all the way down to our components, we can instead pass in an option switch that the graph looks at so that it knows what to render (without needing to know about the admin user).

It might look something like this:

regions_controller.rb:
```ruby
  def diabetes
    # flipper check goes here
    use_who_standard = Flipper.enabled?(:diabetes_who_standard_indicator, current_admin)

    start_period = @period.advance(months: -(Reports::MAX_MONTHS_OF_DATA - 1))
    range = Range.new(start_period, @period)
    @repository = Reports::Repository.new(@region, periods: range)

    # we pass in that flag as an option switch
    @repository = Reports::Repository.new(@region, periods: range, use_who_standard) # takes an additional `opts` argument

    @presenter = Reports::RepositoryPresenter.new(@repository)
    @data = @presenter.call(@region)
```

diabetes.html.erb:

```ruby
# here, we only check the instance var that we have in our controller
<% args_with_ltfu = { data: @data, region: @region, period: @period, with_ltfu: @with_ltfu, use_who_standard: @use_who_standard } %>
```

And now the rest of the code will only check the `use_who_standard` flag to figure out how to calculate/render the WHO indicators.

The other benefit of doing this is that we end up centralising the Flipper check in only the control layer, ie, regions_controller.

app/presenters/reports/repository_presenter.rb
```diff
@@ -1,19 +1,18 @@
...
-      to_hash(region)
+    def call(region, use_who_standard = nil)
+      to_hash(region, use_who_standard)
     end
 
-    def to_hash(region)
+    def to_hash(region, use_who_standard = nil)
       slug = region.slug
-      use_who_standard = Flipper.enabled?(:diabetes_who_standard_indicator, current_admin)
       {
```
tfidfwastaken, line 16: I think because this is a `SimpleDelegator`, we should be able to access `use_who_standard` (which would be delegated to the `attr_reader` of `Repository`) directly without needing to pass it in as an argument.

I am not a 100% sure but if that works, we can eliminate needing to take this arg. If that doesn't this is fine.

tfidfwastaken (approved):

PR #5479, opened 2024-10-24
Add national dashboard link in the top navigation [sc-13777]

Gyan-Gupta-Rtsl:
**Story card:** [sc-13777] https://app.shortcut.com/simpledotorg/story/13777/add-national-dashboard-link-in-the-top-navigation-bar-in-the-dashboard

## Because

 Breadcrumbs on top of the reports page does not show the organization name.

## This addresses

Includes the organization name in the breadcrumbs.

## Test instructions

Need to enable the organization_reports flipper

### DIFF
--- a/app/views/reports/regions/_header.html.erb
+++ b/app/views/reports/regions/_header.html.erb
@@ -4,8 +4,8 @@
 
   <nav class="breadcrumb mb-0px pb-16px">
     <%= link_to "All reports", dashboard_districts_path %>
-
-    <% @region.ancestors.where(region_type: %w[state district block facility]).order(:path).each do |region| %>
+    
+    <% @region.ancestors.where(region_type: %w[organization state district block facility]).order(:path).each do |region| %>
       <i class="fas fa-chevron-right"></i>
       <%= link_to_if(accessible_region?(region, :view_reports), region.name, reports_region_path(region.slug, report_scope: region.region_type)) %>
     <% end %>


Gyan-Gupta-Rtsl:
Evidence attached from local
Screenshot 2024-10-24 at 7 20 18 PM

app/views/reports/regions/_header.html.erb
```diff
@@ -4,10 +4,17 @@
...
+    <% if current_admin.feature_enabled?(:organization_reports) %>
+      <% @region.ancestors.where(region_type: %w[organization state district block facility]).order(:path).each do |region| %>
+        <i class="fas fa-chevron-right"></i>
+        <%= link_to_if(accessible_region?(region, :view_reports), region.name, reports_region_path(region.slug, report_scope: region.region_type)) %>
+      <% end %>
+    <% else %>
+      <% @region.ancestors.where(region_type: %w[state district block facility]).order(:path).each do |region| %>
+        <i class="fas fa-chevron-right"></i>
+        <%= link_to_if(accessible_region?(region, :view_reports), region.name, reports_region_path(region.slug, report_scope: region.region_type)) %>
+      <% end %>
```
tfidfwastaken, line 17: **suggestion**: this can be simplified;

```suggestion
    <% region_types = current_admin.feature_enabled?(:organization_reports) ? %w[organization state district block facility] :  %w[state district block facility] %>
      <% @region.ancestors.where(region_type: region_types).order(:path).each do |region| %>
        <i class="fas fa-chevron-right"></i>
        <%= link_to_if(accessible_region?(region, :view_reports), region.name, reports_region_path(region.slug, report_scope: region.region_type)) %>
    <% end %>
```

spec/controllers/reports/regions_controller_spec.rb
```diff
@@ -59,6 +58,17 @@
...
+      admin = create(:admin, :viewer_reports_only, :with_access, resource: organization)
+      allow(admin).to receive(:feature_enabled?).with(:organization_reports).and_return(true)
+
+      sign_in(admin.email_authentication)
+      get :show, params: {id: organization.slug, report_scope: "organization"}
+
+      expect(response).to be_successful
+      expect(response.body).to include(organization.name)
+    end
+
```
tfidfwastaken, line 71: Some issues with the specs for this:

* This spec always passes. I changed the feature flag mock to return false, and the spec still passed. I deleted the `allow` line and it still passed.
* The reason it always passes is because we `show` the organization reports page in this, and the organization name shows up in the title of the page. And there is only an assertion on the presence of the organization name anywhere on the page, which is too broad and most likely to lead to false positives.
* There's no negative spec, ie, the link should not show up if the flag is disabled.

**suggestion**: These are the assertions we need at a minimum. And they should be a spec on the view template, not on the request, ie, like [this](https://github.com/simpledotorg/simple-server/blob/22eb0a44c9fbe9f7b5aeeb14c542be167c796260/spec/views/reports/regions/_diabetes_footnotes.html.erb_spec.rb) spec in the `/views` folder.

Make sure we are asserting on all of the following:
* When the flag is enabled and we are on the org-level report, the name should show up in the breadcrumb, and should not be a hyperlink.
* When the flag is enabled and we are on any report that's below the org-level (eg, state or facility) the org report should be in the breadcrumb and also should be a hyperlink that correctly navigates us to the org-level report.
* When the flag is disabled, the org name should NOT appear as a breadcrumb item. (test the negative case)

spec/controllers/reports/regions_controller_spec.rb
```diff
@@ -33,7 +33,6 @@
       expect(flash[:alert]).to eq("You are not authorized to perform this action.")
       expect(response).to be_redirect
     end
-
```
tfidfwastaken, line 36: **style**: remove this change from the patch

tfidfwastaken (changes requested):
Requesting some changes to fix the specs and make the code a bit more terse.

spec/controllers/reports/regions_controller_spec.rb
Gyan-Gupta-Rtsl: @tfidfwastaken there is no specs for app/views/reports/regions/_header.html.erb ,
So i need to write the specs from initial stage,this will increase the scope of the ticket.

tfidfwastaken: Good, exhaustive specs need to be accompanied with every change we make to the code.

It should always be considered as part of the scope, and we need to consider the effort to add them in all our estimates from hereon.

spec/utils.rb
```diff
@@ -30,3 +30,28 @@ def reset_controller
...
+# making them accessible within the view spec.
+def inject_controller_helper_methods(controller_class, context)
+  helper_module = controller_class.new._helpers
+  helper_methods = helper_module.instance_methods(false).sort
+  helper_method = ->(method) { helper_module.instance_method(method) }
+
+  helper_methods.each do |method|
+    context.class.define_method(method, helper_method[method])
+  end
+end
```
tfidfwastaken, line 57: **issue**: I don't think it's a good idea to perform all this monkey patching to make our view spec behave like a controller spec. If we relied on these many helper methods within the view, we'd be better off testing this from the controller. I was under the impression that a view spec would be minimal and sufficient.

We can still do a view spec though; see below.

spec/views/reports/regions/_header.html.erb_spec.rb
```diff
@@ -0,0 +1,66 @@
+require "rails_helper"
+
+RSpec.describe "reports/regions/_header.html.erb", type: :view do
+  let(:organization) { create(:organization) }
+  let(:current_admin) { create(:admin, :manager) }
+  let(:organization_region) { create(:region, name: "Organization", region_type: "organization", path: "organization") }
+  let(:organization_region_1) { create(:region, name: "facility", region_type: "facility", path: "organization.state.district") }
```
tfidfwastaken, line 7: **suggestion**: it's better to call this `facility_region`, since this represents a facility, and not an organization.

spec/views/reports/regions/_header.html.erb_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  let(:current_period) { Period.current }
+
+  before do
+    inject_controller_helper_methods(Reports::RegionsController, view)
```
tfidfwastaken, line 11: **suggestion**: If we choose to keep this a view spec, we can include just the helpers we need, like this:

```ruby
  helper do
    def accessible_region?(*args)
      true
    end

    def active_action?(*args)
      false
    end
  end
```

**thought** (no action needed): I think going with the view spec for now is okay, but down the line, we should perhaps break down this view into more components so that each individual part is easier to test and we won't need to stub helpers like this.

tfidfwastaken: Also when I tried to do this I realised we might have to stub out the other components like so:

```ruby
    stub_const("Dashboard::Hypertension::DownloadsDropdownComponent", Class.new {
      def initialize(*); end
      def render_in(*); ""; end
    })
    stub_const("Dashboard::Diabetes::DownloadsDropdownComponent", Class.new {
      def initialize(*); end
      def render_in(*); ""; end
    })
```

But adding these stubs explicitly seems better than the magic involved in injecting controller methods via monkey patching.

spec/views/reports/regions/_header.html.erb_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    allow(current_admin).to receive(:feature_enabled?).with(:organization_reports).and_return(true)
+  end
+
+  describe "when the feature flag for organization_reports is enabled" do
+    context "and we are on the organization-level report" do
+      it "renders the organization name in the breadcrumb without a hyperlink" do
+        render partial: "reports/regions/header", locals: {current_admin: current_admin}
+        expect(rendered).to include('<a href="/dashboard/districts">All reports</a>')
+        expect(rendered).to match(/<i class="fas fa-chevron-right"><\/i>\s*Organization/)
+        expect(rendered).not_to include('<a href="/reports/regions/organization/organization">Organization</a>')
```
tfidfwastaken, line 29: **suggestion** (non-blocking): these assertions are okay for now, but I think it could be made more precise by using some Capybara helpers, like [here](https://github.com/simpledotorg/simple-server/blob/a3aa67b375195d5b8994c91edfeca7d3361a4a45/spec/views/reports/regions/_diabetes_footnotes.html.erb_spec.rb#L12). So we are matching on HTML nodes more directly as opposed to plain string matching.

spec/views/reports/regions/_header.html.erb_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      it "renders the organization name as a hyperlink in the breadcrumb" do
+        render partial: "reports/regions/header", locals: {current_admin: current_admin}
+        expect(rendered).to include('<a href="/dashboard/districts">All reports</a>')
+        expect(rendered).to include('<a href="/reports/regions/organization/organization">Organization</a>')
+        expect(rendered).to match(%r{<i class="fas fa-chevron-right"></i>\s*State Subregion})
+      end
+    end
+
+    context "when the feature flag for organization_reports is disabled" do
```
tfidfwastaken, line 51: **suggestion**: This context does not belong inside the "when the feature flag for organization_reports is enabled" context (since it's doing the opposite of what that context is saying). Let's move it out to a `describe` one level above.

spec/views/reports/regions/_header.html.erb_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      before do
+        allow(current_admin).to receive(:feature_enabled?).with(:organization_reports).and_return(false)
+        allow(current_admin).to receive(:feature_enabled?).with(:dashboard_progress_reports).and_return(false)
+        allow_any_instance_of(UserAccess).to receive(:can_access?).and_return(true)
```
tfidfwastaken, line 55: **question**: This stub seems unnecessary, our admin created above already has `:manager` access?

tfidfwastaken (changes requested):
Suggesting some improvements to the specs.

I strongly believe we should not be dynamically opening up our view class and injecting helper methods—it adds a lot of hidden complexity to the specs and makes it difficult to understand what is happening—if we need to do this, then we might as well just write the spec at the controller level.

I have included some suggestions to make it work at the view spec layer with some stubbing instead, but I will leave the decision for whether to do it at a controller or view level up to you. I'm still somewhat favouring doing it at the view layer.

spec/utils.rb
Gyan-Gupta-Rtsl: @tfidfwastaken This patching can help us test better on view level, for example in my case I need to test my views  as per the change which relies on controllers helper methods, therefore I believe a context in the views specs which can behave as per the controllers helper method would let us test better way how the view is going to behave.
If this view is being used for other controllers, we can inject those controllers helper method for our specific context and test it in isolation too.
I agree we should test it from our controller through but specs need to be accompanied with every change we make to the code. This way we can test our views better with all possible scenarios. Let me know your thoughts on this.

spec/views/reports/regions/_header.html.erb_spec.rb
Gyan-Gupta-Rtsl: @tfidfwastaken I think this is not related to injecting part as components were accessible in views, but the helper methods weren't, even if we don't stub this components or inject helpers it will work eventually as components are accessible in views.

spec/utils.rb
tfidfwastaken: How is dynamically injecting controller helpers (ie a controller responsibility) into a view template better than stubbing the helper methods?

I need to be convinced about why introducing this hard-to-reason utility method is worth the gains in reusability (in this case, _potential_ reuse, since we aren't reusing this anywhere yet).

Monkey-patching/metaprogramming is a very last-resort tool and it needs extremely strong justification if we have to use it. I am okay with this way of doing it, but only if the justification against all other alternatives is clear.

> I agree we should test it from our controller through but specs need to be accompanied with every change we make to the code.

What do you mean by this?

> This way we can test our views better with all possible scenarios

What are all these scenarios? Why is it better than stubbing in those scenarios?

Gyan-Gupta-Rtsl: The stubbing here in this case wasn't working 
`allow(view).to receive(:accessible_region?).and_return(true)` 
since our view does not implement: `accessible_region?` as its a helper method which is not available in view.

This injecting approach uses the actual helper methods as they would be used in the controller or view during a real request, providing more accurate and integrated tests.

Instead of manually defining stubs for each helper method, we can use real helper methods, reducing the need for maintaining and updating stubs when changes occur.

```
helper do
  def accessible_region?(*args)
    true
  end

  def active_action?(*args)
    false
  end
end
```

If we use it like this then for helper methods we are bypassing the actual logic. it may have some underlining logic instead of just returning true/false which could have affected our test case.


**What do you mean by this?**
regarding this it was a reply regarding the point you mentioned above - " If we relied on these many helper methods within the view, we'd be better off testing this from the controller."


**What are all these scenarios? Why is it better than stubbing in those scenarios?**
Any case where other helper methods are required or when using the actual helper method we can test the case by running those actual helpers methods logic based on our parameters passed via view file.

Gyan-Gupta-Rtsl:
@tfidfwastaken Could you please review the PR.

tfidfwastaken (approved):
Approved and LGTM :shipit:

PR #5484, opened 2024-11-13
Add smoking and BMI to sync models

igbanam:
**Story card:** [sc-14097](https://app.shortcut.com/simpledotorg/story/14097/add-bmi-and-smoking-to-medical-history)

## Because

We need to calculate the CVH risk score, and that depends on smoking and BMI

## This addresses

- **add columns to medical_histories**
- **sync both fields**

## Test instructions

- existing test for syncing `medical_histories`
- curl request to `medical_histories` sync endpoint

### DIFF
--- a/app/controllers/api/v3/medical_histories_controller.rb
+++ b/app/controllers/api/v3/medical_histories_controller.rb
@@ -45,6 +45,8 @@ class Api::V3::MedicalHistoriesController < Api::V3::SyncController
         :diabetes,
         :hypertension,
         :diagnosed_with_hypertension,
+        :bmi,
+        :is_smoking,
         :created_at,
         :updated_at
       )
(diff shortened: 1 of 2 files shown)


app/controllers/api/v3/medical_histories_controller.rb
```diff
@@ -45,6 +45,8 @@ def medical_histories_params
         :diabetes,
         :hypertension,
         :diagnosed_with_hypertension,
+        :bmi,
+        :is_smoking,
```
danySam, line 49: Should this API be versioned? We want to keep the API backwards compatible with the client

danySam: Generally, we want to be safer when making API changes because we can't control how app updates are rolled out to users (Android).

danySam: It boils down to the client supporting it. We should capture this in the tech spec and have the Android devs approve this. We can also test this change locally with an un-updated Android version to verify that there aren't any issues. You can read more on the setup here: https://github.com/simpledotorg/simple-server?tab=readme-ov-file#developing-with-the-android-app

PR #5492, opened 2024-11-21
Add Smoking and Cholesterol to Medical Histories

igbanam:
**Story card:** [sc-14155](https://app.shortcut.com/simpledotorg/story/14155/modify-medical-histories)

## Because

We need these to calculate the CVD risk

## This addresses

Adding the smoking and cholesterol hazards to medical histories

## Test instructions

To check there are no regressions…
- Existing tests suite should pass
- Deploy the PR to sandbox and check the app syncs


aagrawalrtsl (approved):
Looks ok to me

PR #5506, opened 2024-12-05
Add Statins sync-resource docs to Swagger UI

igbanam:
**Story card:** [sc-14275](https://app.shortcut.com/simpledotorg/story/14275/generate-documentation-for-statins-routes)

## Because

Running `bundle exec rake docs` by itself does not update the API documentation. There has to be Rswag tests somewhere. It's these tests that update the API documentation. Having tests in the sync-controller alone — with all the resources attached to it — does not give API documentation

## This addresses

Adding API documentation to the Swagger UI

## Test instructions

Visit <deployed-url>/api-docs

### DIFF
--- a/swagger/v4/swagger.json
+++ b/swagger/v4/swagger.json
@@ -381,7 +381,15 @@
         ],
         "responses": {
           "200": {
-            "description": "cvd_risks created"
+            "description": "some, or no errors were found",
+            "schema": {
+              "type": "object",
+              "properties": {
+                "errors": {
+                  "$ref": "#/definitions/errors"
+                }
+              }
+            }
           },
           "403": {
             "description": "user is not allowed to sync"
@@ -389,7 +397,7 @@
         }
       },
       "get": {
-        "summary": "Sync patient attribute data from server to device",
+        "summary": "Sync CVD risk data from server to device",
         "tags": [
           "CVD Risk"
         ],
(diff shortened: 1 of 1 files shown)


spec/api/v4/cvd_risk_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  path "/cvd_risks/sync" do
+    post "Syncs cvd_risks data from device to server." do
+      tags "CVD Risk"
+      security [access_token: [], user_id: [], facility_id: []]
+      parameter name: "HTTP_X_USER_ID", in: :header, type: :uuid
+      parameter name: "HTTP_X_FACILITY_ID", in: :header, type: :uuid
+      parameter name: :cvd_risks, in: :body, schema: Api::V4::Schema.cvd_risk_sync_from_user_request
+
+      response "200", "cvd_risks created" do
```
tfidfwastaken, line 18:
```suggestion
      response "200", "cvd risks received" do
```

This would be more consistent with the rest of the documentation and more technically correct, because a sync merge may or may not create a new resource on the server.

spec/api/v4/cvd_risk_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      response "200", "some, or no errors were found" do
+        schema Api::V4::Schema.sync_from_user_errors
+        let(:cvd_risks) { {cvd_risks: (1..3).map { build(:cvd_risk, :invalid).attributes.with_payload_keys }} }
+        run_test!
+      end
+
+      include_examples "returns 403 for post requests for forbidden users", :cvd_risks
+    end
+
+    get "Sync patient attribute data from server to device" do
```
tfidfwastaken, line 32: This should say "Sync CVD risk data from server to device".

image

tfidfwastaken (changes requested):
requesting two small changes; looks good otherwise.

tfidfwastaken (approved):
Approved, but let's rerun the doc generation command, the swagger.json isn't updated with the new changes!

PR #5508, opened 2024-12-10
Modify CVD Risk for range-score and stale-delete

igbanam:
**Story card:** [sc-14311](https://app.shortcut.com/simpledotorg/story/14311/modify-cvd-risk-for-range-score-and-stale-delete)

## Because

- We would need to give the risk scores as a range sometimes
- Risk scores may need to be deleted when they've gone _stale_

## This addresses

- The blocker building out the CVD sync resource on the Android side

## Test instructions

- suite tests

### DIFF
--- a/swagger/v4/swagger.json
+++ b/swagger/v4/swagger.json
@@ -1898,10 +1898,10 @@
           "$ref": "#/definitions/uuid"
         },
         "risk_score": {
-          "type": "number"
+          "$ref": "#/definitions/non_empty_string"
         },
         "deleted_at": {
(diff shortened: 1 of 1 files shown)


swagger/v4/swagger.json
```diff
@@ -1900,6 +1900,9 @@
         "risk_score": {
           "type": "number"
```
aagrawalrtsl, line 1901: this should be changed to string as well ?

aagrawalrtsl (changes requested):

aagrawalrtsl (approved):

PR #5515, opened 2025-01-08
Optimize Patient Summaries MatView

igbanam:
**Story card:** [sc-13796](https://app.shortcut.com/simpledotorg/story/13796/patient-line-list-returning-no-data-in-simple-dashboard-ethiopia)

## Because

We need a more efficient matview query which works across countries and scales as Simple is used. Currently, the v03 fails in Ethiopia, causing the line list to return an empty zip. V04 uses some advanced SQL to optimize the query, and ensure it scales regardless of how much the underlying data grows

## This addresses

- do away with lateral joins (joining the same set over and over)
- filter early
- do away with the non-performant `select *`
- leverage `lead()` and `lag()` to reduce joins

## Test instructions

See ticket for the run link

### DIFF
--- a/db/views/materialized_patient_summaries_v04.sql
+++ b/db/views/materialized_patient_summaries_v04.sql
@@ -1,0 +1,474 @@
+with latest_bp_passport as (
+    select
+        distinct on (patient_id)
+        id,
+        identifier,
+        patient_id
+    from patient_business_identifiers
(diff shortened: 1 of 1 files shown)


igbanam:
Following this would be a PR to migrate Simple to the more performant V04 patient summaries.

Gyan-Gupta-Rtsl (approved):

PR #5528, opened 2025-02-04
Cancel upcoming BD SMSs

igbanam:
**Story card:** [sc-14668](https://app.shortcut.com/simpledotorg/story/14668/delete-all-upcoming-bd-notifications-for-stale-patients)

## Because

Stale patient experiments use up credit meant for current patients

## This addresses

Cancelling the experiments for stale patients

## Test instructions

1. Ran the underlying SQL in metabase to confirm the set
1. Ran the query (without the `destroy`) in prod to match (1)

### DIFF
--- a/db/data/20250204101724_cancel_bangladesh_stale_experiments.rb
+++ b/db/data/20250204101724_cancel_bangladesh_stale_experiments.rb
@@ -2,8 +2,9 @@
 
 class CancelBangladeshStaleExperiments < ActiveRecord::Migration[6.1]
   def up
-    Experimentation::Experiment.upcoming.where(experiment_type: 'stale_patients').each do |experiment|
-      experiment.destroy  # ...because we need the associated notifications gone too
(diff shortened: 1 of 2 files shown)


db/data/20250204101724_cancel_bangladesh_stale_experiments.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CancelBangladeshStaleExperiments < ActiveRecord::Migration[6.1]
+  def up
+    Experimentation::Experiment.upcoming.where(experiment_type: 'stale_patients').each do |experiment|
+      experiment.destroy  # ...because we need the associated notifications gone too
```
aagrawalrtsl, line 6: I am not sure, but would this remove the patients from the treatment group, and destroy that group as well ?

igbanam: thank you. that's a good catch. patch would be up in a bit

PR #5532, opened 2025-02-18
sc-14776 added target at html

Gyan-Gupta-Rtsl:
**Story card:** [sc-14776](https://app.shortcut.com/simpledotorg/story/14776/open-links-in-a-new-tab-on-the-simple-dashboard)

Because
To enable quick links to open in a new tab for a seamless user experience.

This Addresses
This change allows users to compare the Metabase reports with the Simple dashboard simultaneously, improving usability and workflow efficiency.

Test Instructions
Ensuring that when a user clicks a link from the Simple dashboard, it opens in a new tab, allowing them to view both the dashboard and external reports (e.g., Metabase) side by side.

### DIFF
--- a/app/views/reports/regions/show.html.erb
+++ b/app/views/reports/regions/show.html.erb
@@ -9,19 +9,18 @@
     <% size = @region.source.facility_size.present? ? @region.source.facility_size : '' %>
     <% zone = @region.source.zone %>
     <% last_month = Date.today.last_month.strftime("%b-%Y") %>
-    <div>⚡ <a href="<%= ENV.fetch('DRUG_STOCK_REPORT_URL', '') %><%= district %>&zone=<%= zone %>&for_end_of_month=<%= last_month %>&size=<%= size %> ">
+    <div>⚡ <a href="<%= ENV.fetch('DRUG_STOCK_REPORT_URL', '') %><%= district %>&zone=<%= zone %>&for_end_of_month=<%= last_month %>&size=<%= size %> " target="_blank">
       Drug stock report
     </a></div>
-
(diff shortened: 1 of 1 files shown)


igbanam (commented):
Two questions:

- How should the app behave when the environment variable is not set?
- Are we only concerned about these two reports? What do we do when a country wants other reports from their metabase on their dashboard?

Gyan-Gupta-Rtsl:
> * How should the app behave when the environment variable is not set?
**Please refer this file: simple-server/.env.BD**(This feature is only enabled for BD)
> * Are we only concerned about these two reports? What do we do when a country wants other reports from their metabase on their dashboard?
**As per current scenario we are only concerned with these 2 reports only for the region**

igbanam (changes requested):
> This feature is only enabled for BD
>
> ~ https://github.com/simpledotorg/simple-server/pull/5532#issuecomment-2665002339

If this is just for BD, but this change would affect all deployed regions, then we have to ensure we design for regions where this feature is not available; ET for example.

Gyan-Gupta-Rtsl:
> > This feature is only enabled for BD
> > ~ [#5532 (comment)](https://github.com/simpledotorg/simple-server/pull/5532#issuecomment-2665002339)
> 
> If this is just for BD, but this change would affect all deployed regions, then we have to ensure we design for regions where this feature is not available; ET for example.

I believe this condition might have been overlooked:
<% if @is_desktop && @region.facility_region? && Flipper.enabled?(:quick_link_for_metabase, current_admin) %>

Could you kindly review the reasoning behind the PR I raised? It seems there may be a misunderstanding or lack of context regarding this matter.

igbanam (approved):
Right. LGTM.

The context I needed was outside the scope of this changeset.

PR #5534, opened 2025-03-05
feature(OpenSRP Exporter): Time Boundaries

igbanam:
**Story card:** [sc-14814](https://app.shortcut.com/simpledotorg/story/14814/opensrp-exports-export-a-specific-time-window)

## Because

Our stakeholders have asked for reports which are deltas from when they received the last report.

## This addresses

Adding some time boundaries to the report's config so operators can define for how long they want this report.

## Test instructions

- Did an export
- Inspected the JSON

### DIFF
--- a/config/opensrp-export.sample.yml
+++ b/config/opensrp-export.sample.yml
@@ -1,3 +1,6 @@
+time_boundaries:
+  report_start: 2001-01-01
+  report_end: 2002-02-02
 facilities:
   d1dbd3c6-26bb-48e7-aa89-bc8a0b2bf75b:
     name: Health Facility 1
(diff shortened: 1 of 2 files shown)


aagrawalrtsl (commented):
Should we also include updated_at time to get the records for different models, so that if the record is updated within the requested timeframe, that also gets included ?

aagrawalrtsl:
> This is a good suggestion @aagrawalrtsl. I've added the changes for `appointments` and `prescription_drugs`. The readings — blood pressure, and blood sugar — technically should be immutable. Let me know if that assumption is wrong.

I think they can be updated. Just checked in Sri Lanka and there are many instances for blood_sugars where created_at and updated_at are different by months even.

igbanam:
@aagrawalrtsl I've updated those too

aagrawalrtsl (approved):

PR #5544, opened 2025-03-20
Minor changes in text [sc-15050]

Gyan-Gupta-Rtsl:
**Story card:** [sc-15050](https://app.shortcut.com/simpledotorg/story/15050/minor-changes-in-text)

## Because

Text needs to be changed

## This addresses

Names to be changed

Facility trends
Drug stock

## Test instructions

NA


igbanam (commented):
Should the environmennt variable names change as well?

Gyan-Gupta-Rtsl:
> Should the environmennt variable names change as well?

The environment variable names do not need to be changed. The update was only to improve the display text for clarity, but the underlying functionality and purpose of the reports remain the same.

PR #5569, opened 2025-04-23
chore: update to ruby-2.7.8

igbanam:
igbanam/sc-15306/upgrade-ruby

PR #5570, opened 2025-04-24
chore: Remove unused wkhtmltoimage gem

igbanam:
igbanam/sc-15228/bump-ruby-to-2.7-latest

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -91,7 +91,6 @@ gem "uuidtools", require: false
 gem "view_component"
 gem "webpacker", "6.0.0.rc.6"
 gem "whenever", require: false
-gem "wkhtmltoimage-binary"
 gem "table_print"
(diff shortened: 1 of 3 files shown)


igbanam:
The `wkhtmltoimage` is not used anywhere. The gem exists to convert html sites to images. It's meant to be a binary and has nothing to do with the simple-server.

#5569 broke Semaphore's setup because the PPA for libpng needed for this gem's binary has been deprecated. The only way to get libpng was to build it from source. This friction created the need to remove the gem totally from the app.

igbanam:
Turns out this binary was actually instrumented in the codebase, but for a feature which doesn't work. Currently deliberating removing the feature altogether

igbanam (commented):
still ongoing

PR #5600, opened 2025-05-20
chore: Make tests faster

igbanam:
**Story card:** [sc-15268](https://app.shortcut.com/simpledotorg/story/15268/versioning-adopt-semantic-versioning)

## Because

CI takes 1 hour to build

## This addresses

Optimizing matview refreshes during unit tests

Ensure `refresh_views` can refresh specific views.

Originally, this refreshed all the mat views which — for some tests
— may be a recipe for wasting time. It's this query that's in focus for
this change. There are other places where `refresh_views` is used, but
those do not seem to run slow.

## Test instructions

suite tests


igbanam:
I can't believe it 🤯 

Screenshot_2025-05-20_at_21_25_07

aagrawalrtsl (approved):

PR #5601, opened 2025-05-20
Add CI

igbanam:


igbanam:
This is technically meant for [sc-15269](https://app.shortcut.com/simpledotorg/story/15269/versioning-use-release-drafter-to-create-releases)

igbanam:
Right now, we're choosing to go the Semaphore route. We would revisit Github Actions later.

PR #5602, opened 2025-05-20
Whitelabeling around resources tab

aagrawalrtsl:
**Story card:** [sc-15243](https://app.shortcut.com/simpledotorg/story/15243/the-name-of-the-application-simple-should-come-from-the-configuration)

## Because

We need to remove some links in under Resources tab in dashboard, and Help section in Android app.

## This addresses

Introducing a new "WHITELABEL_APP" flag in the environment variable, which will control this.
When set to TRUE, it entails that we are whitelabeling the server app, and hence will remove the simple branded video links from the resources section.

## Test instructions

Set up following environment vars
WHITELABEL_APP
DEPLOYMENT_CHECKLIST_LINK
PRIVACY_LINK
LICENSE_LINK
Check the links under Resources & Training section. All "Simple" branded videos should not be available.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -84,3 +84,7 @@ HELP_EMAIL_ID="help_email@test.org"
 CVHO_EMAIL_ID="cvho_email@test.org"
 ENG_EMAIL_ID="eng-backend@test.org"
 FAVICON_URL="https://simple.org/images/favicon.png"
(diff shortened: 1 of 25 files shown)


igbanam (approved):
LGTM

PR #5614, opened 2025-06-04
Adding a partitioned reporting_patient_states table

aagrawalrtsl:
**Story card:** [sc-15583](https://app.shortcut.com/simpledotorg/story/15583/new-reporting-pipeline-for-reporting-patient-states)

## Because

Materialized view refresh is very slow.
We want to introduce partitioned table approach, where with each month it will drop that partition, regenerate data for that partition and attach it to the main table.
With this new approach we will not be doing a full refresh of the table every day.

## This addresses

This is the first step towards redesigning our reporting pipeline.
We are adding a partitioned reporting_patient_states table, under a different schema. The Mat view version of same would continue to behave as it is, until we decide to finally make the partitioned table live.
Dashboards are still going to use the mat view for getting the data. 

When the daily rake task for runs, then for reporting_patient_states will refresh only the current month and previous month.
We still want to refresh last 15 months data to handle delayed sync scenario. So last 15 months data would be refreshed once at some point during the current month.

There is a rake task for getting this table fully refreshed. This would delete everything from the table and regenerate all the data from the beginning. Ideally, this task would run only once, in the start when the feature is deployed.

## Test instructions

Run the migrations.
rake "db:refresh_reporting_views" -  for monthly/daily refresh
rake "reporting:full_partitioned_refresh" -  for full refresh of reporting_patient_states.


app/models/reports/view.rb
```diff
@@ -33,5 +33,12 @@ def self.add_column_description(column_description, column_name)
     def self.materialized?
       raise NotImplementedError
     end
+
+    def self.get_refresh_months
+      current_date = Date.today
+      current_day = current_date.day
+      current_month = current_date.beginning_of_month
+      current_day.odd? ? [current_month, current_month << 1] : [current_month, current_month >> ((current_day / 2) + 1)]
```
igbanam, line 41:
```ruby
current_month << 1
```

could be

```ruby
current_month.prev_month
```

…and…

```ruby
current_month >> ((current_day / 2) + 1)
```

could be

```ruby
current_month.next_month((current_day / 2) + 1)
```

It's easier to grok what's going on.

igbanam: The `.next_month` bit can be further clarified as 

```ruby
month_offset = current_day / 2 + 1
current_month + month_offset.month
```

PR #5633, opened 2025-07-01
Hid recent updates and upcoming improvements

dburka:
These 2 sections were on the HOME screen of the dashboard but hadn't been updated in >1 year. It looks like Simple is dead if we don't use these. I have removed them but left the files so we can bring them back later if we want.

No story.


jamiecarter7:
LGTM :shipit:

PR #5640, opened 2025-07-16
bring back hidden pre-init initializer config

igbanam:


igbanam:
Things that are brought back so far are

- **simple_server_extensions** for things like `Simple.env`
- **metrics** for all things Prometheus
- **current_enabled_features** because we throw this into the global html for debugging purposes

PR #5643, opened 2025-07-31
feat: Dr. Rai Titration Indicator

igbanam:
**Story card:** [sc-15931](https://app.shortcut.com/simpledotorg/story/15931/dr-rai-titration-indicator)

## Because

Titration Indicator is the second indicator we have chosen to implement for Dr. Rai Reports. This indicator has SQL as its datasource — as opposed to the data in the Dashboard.

## This addresses

- Creating the mechanism to use SQL as a data source
- Creating the Titration Indicator (alongside its query)
- Adding the Titration Indicator to the Dr. Rai component
- Smoothening out component decisions made when there was just one indicator

## Test instructions

suite tests

### DIFF
--- a/app/models/dr_rai/titration_indicator.rb
+++ b/app/models/dr_rai/titration_indicator.rb
@@ -39,11 +39,11 @@ module DrRai
     def is_supported?(region)
       case CountryConfig.current[:name]
       when "Bangladesh"
-        region.path.split('.').include?("nhf")
+        region.path.split(".").include?("nhf")
       when "Ethiopia"
-        region.path.split('.').none? { |level| level.include?("non_rtsl") }
+        region.path.split(".").none? { |level| level.include?("non_rtsl") }
       when "Sri Lanka"
-        region.path.split('.').include?("sri_lanka_organization")
+        region.path.split(".").include?("sri_lanka_organization")
       else
         false
       end


app/models/dr_rai/action_plan.rb
```diff
@@ -24,9 +24,20 @@ def denominator
...
+
+  def unit
+    indicator.unit
+  end
+
+  private
+
+  def unprocessible?
+    denominator.negative? ||
+      numerator.nil?
```
aagrawalrtsl, line 41: Should we add  a check for when denominator is 0 as well ?

app/queries/dr_rai/titration_query.rb
```diff
@@ -0,0 +1,266 @@
...
+            current_month.medicine_purpose_hypertension is TRUE as medicine_purpose_hypertension,
+            current_month.actual_dosage as current_dosage,
+            previous_month.actual_dosage as previous_dosage,
+            current_month.actual_dosage IS DISTINCT FROM previous_month.actual_dosage as drug_titrated
+
+        from monthly_drugs current_month
+        join reporting_patient_states on reporting_patient_states.patient_id = current_month.patient_id and reporting_patient_states.month_date = current_month.month_date
+        left outer join monthly_drugs previous_month on current_month.patient_id = previous_month.patient_id and current_month.actual_name = previous_month.actual_name and current_month.month_date = previous_month.month_date + '1 month'::interval
+        where reporting_patient_states.month_date between timestamp '#{begins_at}' and timestamp '#{ends_at}'
+        AND reporting_patient_states.assigned_facility_slug NOT LIKE '%non-rtsl%'
```
aagrawalrtsl, line 178: I assume this condition is taken from metabase particularly from ethiopia. 
Different countries have different requirements when we run this report in metabase. For e.g in BD we only consider NHF facilities. I am not sure if we would be able to inject this condition directly based on different countries, but I think we can get rid of this condition, and let the report be generated for all the facilities.
?

aagrawalrtsl (approved):

PR #5656, opened 2025-08-22
Dr Rai Ux improvements

jamiecarter7:
**Story card:** [sc-16421](https://app.shortcut.com/simpledotorg/story/16421/fixes-to-dr-rai-mvp)

## Because

Adding a loading state when clicking the 'Create' button on the action side panel

## This addresses

It seems like nothing is happening after clicking the 'Create' button as it take 2-3 seconds to load.
This informs the user action is being performed, once complete the page refreshes

## Test instructions

Create an action and click the create button

### DIFF
--- a/app/assets/stylesheets/dr_rai.scss
+++ b/app/assets/stylesheets/dr_rai.scss
@@ -459,7 +459,7 @@ $font-condensed: "Roboto Condensed", sans-serif;
 
   .action-buttons-block {
     display: flex;
-    gap: 28px;
+    gap: 24px;
 
     button {
       border: none;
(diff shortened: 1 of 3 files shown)


igbanam (changes requested):
Thanks for doing this, @jamiecarter7. We would also want to disable the button. The code should read something like

```
const saveButton = $('css query selector for button')
saveButton.disabled = true
```

And when there's an error while creating…

```js
saveButton.disabled = false
```

Fantastic work in using the errored state to refresh the UI. Finding use for something I thought there was no use for 🥇

jamiecarter7:
> Thanks for doing this, @jamiecarter7. We would also want to disable the button. The code should read something like
> 
> ```
> const saveButton = $('css query selector for button')
> saveButton.disabled = true
> ```
> 
> And when there's an error while creating…
> 
> ```js
> saveButton.disabled = false
> ```
> 
> Fantastic work in using the errored state to refresh the UI. Finding use for something I thought there was no use for 🥇

Ok, that is much simpler than I had imagined.

I'll get this done ASAP

PR #5660, opened 2025-08-25
fix(Dr. Rai Report): Default goal to 0

igbanam:
**Story card:** [sc-16474](https://app.shortcut.com/simpledotorg/story/16474/high-calculated-statement-activity-statement-is-shown-before-a-value-is-entered-in-the-input-the-initial-value-can-be-0)

## Because

UX upgrades

## This addresses

Showing the activity statement on empty state.

## Test instructions

1. Try adding an action plan
1. Check that the activity statement is 0 without an input in the input box


jamiecarter7 (approved):
LGTM :shipit:

PR #5667, opened 2025-08-29
build(CD): Refactor Long Running Tests

igbanam:
**Story card:** [sc-16502](https://app.shortcut.com/simpledotorg/story/16502/make-cd-pipelines-faster)

## Because

We are beginning to pay a lot for running these CD pipelines

## This addresses

potentially doing things in a more efficient manner

## Test instructions

suite-tests


igbanam:
…still takes longer. It's no use.

PR #5684, opened 2025-09-15
feat(OpenSRP Exporter): Selective Exporting

igbanam:
**Story card:** [sc-16544](https://app.shortcut.com/simpledotorg/story/16544/opensrp-selective-exporting)

## Because

To reduce impact on data, we need to enable differential exports.

## This addresses

- Refactoring the `opensrp:export` task into a one-off service
- Testing that service
- Filtering selected patients for export [probably from config file]

## Test instructions

suite tests


igbanam:
The patient filter supersedes the facility filter. The cases get _nicher_ and _nicher_

PR #5687, opened 2025-09-19
fix(Dr. Rai Reports): Respect 'non-contactable' switch

igbanam:

## Because

We're using the `overdue_patients` regardless of the dashboard only considering `contactable_overdue_patients`

## This addresses

Using the right numerator and denominator keys depending on if
non-contactable patients are included in the dashboard calculations.

## Test instructions

suite tests


Gyan-Gupta-Rtsl:
Thanks for merging this, @Igbanam. 
I just feel that ideally every PR should go through a review before merging. In this case, it seems it was merged directly. Personally, I don’t think this is the best practice, since reviews help us catch issues early and keep consistency. Could we try to ensure reviews happen before merges going forward?

PR #5701, opened 2025-09-25
adding logic for doing partitioned refresh weekly

aagrawalrtsl:
**Story card:** [sc-16751](https://app.shortcut.com/simpledotorg/story/16751/update-refresh-frequeny-for-weekly-report-refresh)

## Because

India is set to do the reports refresh once per week. Our refresh for partitioned table assumes that it runs daily and hence just refreshes for two months (current month and one past month)
This would mean that in India we will miss refreshing data for some older months.

## This addresses

Adding a new environment variable REPORTING_REFRESH_FREQUENCY which when set to "weekly" will then refresh for all the months, corresponding to past 7 days, which would ideally have been refreshed if it was running daily.
For e.g. Let's assume it is running on 24 Sep 2025, then it assumes then it was last run on 17 Sep 2025, and now we have to a refresh for months that correspond to 17, 18, 19, 20, 21, 22, 23 and 24 Sep 2025.
In this case months refreshed would be Sep 2025, Aug 2025, Aug 2024, Sep 2024, Oct 2024 and Nov 2024

## Test instructions
Set REPORTING_REFRESH_FREQUENCY="weekly"
run rake "db:refresh_reporting_views". 
It would then refresh for all the months corresponding to the last week. 
This can be verified in entries under "simple_reporting.simple_reporting_runs" table


Gyan-Gupta-Rtsl (approved):

spec/models/reports/view_spec.rb
```diff
@@ -31,25 +31,52 @@ def table_descriptions_sql(table_name)
...
   describe ".get_refresh_months" do
-    it "returns current and previous month when current day is odd" do
-      travel_to Date.new(2023, 6, 15) do
-        expected_months = [Date.new(2023, 6, 1), Date.new(2023, 5, 1)]
-        expect(described_class.get_refresh_months).to eq(expected_months)
+    context "when refresh frequency is not set" do
+      before { ENV.delete("REPORTING_REFRESH_FREQUENCY") }
+
+      it "defaults to daily logic" do
+        travel_to(Date.new(2025, 9, 24)) do
```
igbanam, line 56: Hold on a second! This is possible? That's amazing. Means we can do without the Timecop gem.

https://github.com/simpledotorg/simple-server/blob/cd2124fa67eee59c519f461fc5e75e7620577e41/Gemfile#L87

igbanam (approved):
LGTM.

Thanks for the `travel_to` tip!

PR #5703, opened 2025-09-30
initial commit for replacing Materialized reporting_patient_states wi…

aagrawalrtsl:
…th a normal view

## Because

## This addresses

## Test instructions

### DIFF
--- a/db/migrate/20250925094123_discard_mat_view_reporting_patient_states.rb
+++ b/db/migrate/20250925094123_discard_mat_view_reporting_patient_states.rb
@@ -1,0 +1,849 @@
+class DiscardMatViewReportingPatientStates < ActiveRecord::Migration[6.1]
+  def change
+    # create_view :reporting_patient_states
+    drop_view :reporting_quarterly_facility_states, materialized: true
+    drop_view :reporting_facility_states, materialized: true
+    drop_view :reporting_overdue_patients, materialized: true
+    drop_view :reporting_facility_monthly_follow_ups_and_registrations, materialized: true
+    drop_view :reporting_patient_states, materialized: true
+    execute <<~SQL
+      CREATE VIEW public.reporting_patient_states AS SELECT * FROM simple_reporting.reporting_patient_states;
+    SQL
+    
+    execute <<~SQL
+      CREATE MATERIALIZED VIEW public.reporting_facility_monthly_follow_ups_and_registrations AS
+      WITH monthly_registration_patient_states AS (
+        SELECT reporting_patient_states.registration_facility_id AS facility_id,
+          reporting_patient_states.month_date,
+          reporting_patient_states.gender,
+          reporting_patient_states.hypertension,
+          reporting_patient_states.diabetes
+          FROM public.reporting_patient_states
+        WHERE (reporting_patient_states.months_since_registration = (0)::double precision)
+      ), registered_patients AS (
+        SELECT monthly_registration_patient_states.facility_id,
+          monthly_registration_patient_states.month_date,
+          count(*) AS monthly_registrations_all,
+          count(*) FILTER (WHERE (monthly_registration_patient_states.hypertension = 'yes'::text)) AS monthly_registrations_htn_all,
+          count(*) FILTER (WHERE ((monthly_registration_patient_states.hypertension = 'yes'::text) AND ((monthly_registration_patient_states.gender)::text = 'female'::text))) AS monthly_registrations_htn_female,
+          count(*) FILTER (WHERE ((monthly_registration_patient_states.hypertension = 'yes'::text) AND ((monthly_registration_patient_states.gender)::text = 'male'::text))) AS monthly_registrations_htn_male,
+          count(*) FILTER (WHERE ((monthly_registration_patient_states.hypertension = 'yes'::text) AND ((monthly_registration_patient_states.gender)::text = 'transgender'::text))) AS monthly_registrations_htn_transgender,
+          count(*) FILTER (WHERE (monthly_registration_patient_states.diabetes = 'yes'::text)) AS monthly_registrations_dm_all,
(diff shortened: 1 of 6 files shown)


spec/reporting_helpers.rb
```diff
@@ -14,6 +14,12 @@ def june_2021
     reporting_dates(2021, 6)
   end
 
+  def self.get_refresh_months_between_dates(start_date, end_date)
+    refresh_months = []
+    (start_date..end_date).each { |refresh_date| refresh_months << refresh_date.beginning_of_month }
+    refresh_months.uniq
```
igbanam, line 20: For all things date and time in Simple, I refer to [Period](https://github.com/simpledotorg/simple-server/blob/master/app/models/period.rb). I added [`Period#quarters_between`](https://github.com/simpledotorg/simple-server/blob/71c56f4bc733b1846b145d18652530baee4fbb48/app/models/period.rb#L40) for Dr. Rai Reports. First thing I thought here was "Could we have a `Period#months_between`.

Then I considered the Ruby way. What if the entire method could read like this?

```ruby
(start_date..end_date).step(1.month).to_a
```

I realized `Range#step` does not handle [`ActiveSupport::Duration`](https://api.rubyonrails.org/classes/ActiveSupport/Duration.html) properly; rightly so as well cos the Duration class is within Rails and Range is a Ruby class. But… if we monkey-patch Range within our [TimeAndDateExtensions](https://github.com/simpledotorg/simple-server/blob/master/config/initializers/time_and_date_extensions.rb), with 

```ruby
class Range
  def step_by_duration(step_value)
    raise ArgumentError, "step must be an ActiveSupport::Duration" unless step_value.is_a?(ActiveSupport::Duration)

    Enumerator.new do |a|
      current = self.begin
      while current <= self.end
        a << current
        current += step_value
      end
    end
  end
end
```

then we can have this

```ruby
(start_date..end_date).step_by_duration(1.month).to_a
```

db/migrate/20250925094123_discard_mat_view_reporting_patient_states.rb
```diff
@@ -0,0 +1,849 @@
+class DiscardMatViewReportingPatientStates < ActiveRecord::Migration[6.1]
+  def change
+    # create_view :reporting_patient_states
+    drop_view :reporting_quarterly_facility_states, materialized: true
+    drop_view :reporting_facility_states, materialized: true
+    drop_view :reporting_overdue_patients, materialized: true
+    drop_view :reporting_facility_monthly_follow_ups_and_registrations, materialized: true
+    drop_view :reporting_patient_states, materialized: true
+    execute <<~SQL
+      CREATE VIEW public.reporting_patient_states AS SELECT * FROM simple_reporting.reporting_patient_states;
```
igbanam, line 10: I'm worried about this bit. A view is virtual i.e. doesn't store anything on disk. This means the query to produce the view would be run every time this view is accessed. There's no way the query for the dashboard would load within the request lifecycle 😅 

Can we keep this as a materialized view, but deprecate it somehow. A materialized view would store the data on disk for faster retrieval. We can just choose to not update it as frequently as before, or not update it at all. But to remove it so we run the query in the request lifecycle may be costly for environments with plenty data

igbanam (changes requested):
I think if we still want to depend on Reports::RpatientState during the transition to the new way, we should keep it as a mat-view

I'm open to discussing this on a call

db/migrate/20250925094123_discard_mat_view_reporting_patient_states.rb
```diff
@@ -0,0 +1,849 @@
...
+
+    execute <<~SQL
+      CREATE INDEX index_qfs_facility_id ON public.reporting_quarterly_facility_states USING btree (facility_id);
+      CREATE UNIQUE INDEX index_qfs_quarter_string_region_id ON public.reporting_quarterly_facility_states USING btree (quarter_string, facility_region_id);
+    SQL
+
+    # reporting_facility_monthly_follow_ups_and_registrations
+    # reporting_overdue_patients
+    # reporting_facility_states
+    # reporting_quarterly_facility_states
```
igbanam, line 712: 🧹

igbanam, line 712:
```suggestion
```

db/migrate/20250925094123_discard_mat_view_reporting_patient_states.rb
```diff
@@ -0,0 +1,849 @@
...
+  #         LEFT JOIN public.medical_histories mh ON (((p.id = mh.patient_id) AND (mh.deleted_at IS NULL))))
+  #         LEFT JOIN public.reporting_prescriptions current_meds ON (((current_meds.patient_id = p.id) AND (cal.month_date = current_meds.month_date))))
+  #         LEFT JOIN public.reporting_prescriptions past_meds ON (((past_meds.patient_id = p.id) AND (cal.month_date = (past_meds.month_date + '1 mon'::interval)))))
+  #         JOIN public.reporting_facilities registration_facility ON ((registration_facility.facility_id = p.registration_facility_id)))
+  #         JOIN public.reporting_facilities assigned_facility ON ((assigned_facility.facility_id = p.assigned_facility_id)))
+  #       WHERE (p.deleted_at IS NULL)
+  #       ORDER BY p.id, cal.month_date
+  #       WITH NO DATA;
+  #   SQL
+  # end
```
igbanam, line 848:
```suggestion
```

igbanam (commented):
🧹

spec/reporting_helpers.rb
aagrawalrtsl: This method is only in spec as a helper and will not be used much later. Hence we have decided to keep it like this.

spec/features/admin/users/index_spec.rb
```diff
@@ -84,6 +84,7 @@
       edit_page.edit_registration_facility(facility_hoshiarpur.name)
 
       # user should be displayed in hoshiarpur as we have edited its registration facility
+      sleep(5)
```
Gyan-Gupta-Rtsl, line 87: Could you please explain why we need to add `sleep` here in test case? How it helps here?

PR #5790, opened 2026-03-26
Deploy merge

sshinde-rtsl:

## Because

## This addresses

## Test instructions


app/controllers/admin/deduplicate_patients_controller.rb
```diff
@@ -1,18 +1,31 @@
...
 
+    # Apply facility filter if selected and feature flag is enabled
+    filtered_facilities = if filter_enabled? && @selected_facility.present?
+      facilities.where(id: @selected_facility.id)
+    elsif filter_enabled? && @selected_district.present?
+      facilities.where(id: @selected_district.facilities)
+    else
+      facilities
+    end
+
```
igbanam, line 18: Do we still need #5789 if we have this here as well?

config/routes.rb
```diff
@@ -71,6 +71,10 @@
       get "sync", to: "cvd_risks#sync_to_user"
       post "sync", to: "cvd_risks#sync_from_user"
     end
+
+    scope :patient_scores do
+      get "sync", to: "patient_scores#sync_to_user"
+    end
```
igbanam, line 77: Why do we need a sync route for this? `PatientScore` is not data we collect from the field per se; it's calculated from the data collected. A simple `GET` route would suffice here. Am I missing something?

db/migrate/20260209112204_create_patient_scores.rb
```diff
@@ -0,0 +1,24 @@
+class CreatePatientScores < ActiveRecord::Migration[6.1]
+  def change
+    unless table_exists?(:patient_scores)
+      create_table :patient_scores, id: :uuid do |t|
+        t.references :patient, null: false, foreign_key: true, type: :uuid
```
igbanam, line 5: This is a potential breakage. See the fixes in #5768 and #5795.

Please fix this

igbanam (changes requested):
From what I gather looking through implementation here, it looks like we are trying to vend the `patient_scores` resource to the mobile somehow. If this is **_why_** we are doing this, please update the PR description and the JIRA ticket so it's clear.

A couple comments

1. I think a simple GET endpoint will suffice here. Patient Scores don't need to be sync'd from the field. Think of it as a "materialized view" of sorts from data collected from the field.
2. If you've been instructed to do this as a sync route, please do two things (1) document that instruction somewhere so future collaborators can understand the decision path here, and (2) fix the link to patients, because that causes an order bug.

igbanam:
Oh and there's significant overlap with #5789. You may want to look into that

PR #5797, opened 2026-04-09
feat(ET): Use country config for supported genders in breakdown table

sshinde-rtsl:
- Add CountryConfig.supported_genders helper method with fallback
- Update breakdown_table_component to dynamically render genders based on country config
- Add specs for supported_genders method

This ensures Ethiopia (ET) only shows male/female genders while other countries display all configured genders including transgender.

**Story card:** [SIMPLEETH-27](https://rtsl.atlassian.net/browse/SIMPLEETH-27)

## Because

## This addresses

## Test instructions


spec/initializers/country_config_spec.rb
```diff
@@ -11,4 +11,15 @@
       ).to eq true
     end
   end
+
+  describe ".supported_genders" do
```
aagrawalrtsl, line 15: we should previously set a value for the supported genders to a test value, and then expect if `CountryConfig.supported_genders` is returning the correct value or not.

config/initializers/country_config.rb
```diff
@@ -149,6 +149,10 @@ def self.country_environment_file
   def self.current_country?(country)
     CountryConfig.current[:name] == country
   end
+
+  def self.supported_genders
+    current[:supported_genders] || %w[male female transgender]
+  end
```
aagrawalrtsl, line 155: should we also set a value for supported genders in ET config as well with just male and female ?

aagrawalrtsl: > should we also set a value for supported genders in ET config as well with just male and female ?

This is not required, we already have supported_genders specified for each country here.
Ignore this comment

aagrawalrtsl (approved):

app/components/progress_tab/breakdown_table_component.html.erb
```diff
@@ -7,28 +7,15 @@
...
-  </div>
-  <div class="d-flex ai-center jc-space-between <% if include_bottom_border %>pb-3px bb-grey-mid<% end %> mb-4px">
-    <p class="m-0px p-0px pl-16px ta-left fw-normal fs-16px c-grey-dark">
-      <%= t("progress_tab.genders.transgender") %>
-    </p>
-    <p class="m-0px p-0px ta-right fw-normal fs-16px c-grey-dark">
-      <%= breakdown[:transgender] %>
-    </p>
-  </div>
+  <% CountryConfig.supported_genders.each_with_index do |gender, index| %>
```
igbanam: Hide this within the component. The view need not know about `CountryConfig` it only cares about `genders`.

igbanam (changes requested):
This change brings in some [inappropriate intimacy](https://refactoring.guru/smells/inappropriate-intimacy).

app/components/progress_tab/breakdown_table_component.html.erb
sshinde-rtsl: Got it. I’ll move the logic inside the component so the view doesn’t directly depend on CountryConfig. The component will handle fetching and shaping the data, and the view will only work with the genders it receives.

igbanam (commented):
LGTM

igbanam (approved):

PR #5798, opened 2026-04-13
Simpleback 34

sshinde-rtsl:
**Story card:** [SIMPLEBACK-34](https://rtsl.atlassian.net/browse/SIMPLEBACK-34?search_id=e983b2fc-450c-4e46-a57c-2c0a5f0bea35)

## Because

We need a sync API for patient scores so that the mobile app can retrieve patient risk scores for prioritization and follow-up workflows.

## This addresses

 - Adds a read-only sync API for patient scores
  - Implements keyset pagination using (updated_at, id) to prevent duplicate/missed records when multiple scores share the same timestamp
  - Uses next_page token for PatientScore sync with proper cursor tracking
  - Parses process_token timestamps in UTC for correct keyset pagination across timezones
  - Spreads clustered PatientScore updated_at timestamps to avoid pagination issues
  - Fixes a flaky test by excluding diagnosed_confirmed_at from deduped patient test comparison

## Test instructions

 1. Run the patient scores controller tests:
  bundle exec rspec spec/controllers/api/v3/patient_scores_controller_spec.rb
  2. Run the patients controller tests to verify the test fix:
  bundle exec rspec spec/controllers/api/v3/patients_controller_spec.rb
  3. Verify sync API works correctly by making requests with different process_token values and confirming pagination returns correct records without duplicates


igbanam:
Please fill in the PR template with actual details of this change in the description.

db/migrate/20260209112204_create_patient_scores.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    unless index_exists?(:patient_scores, [:patient_id, :score_type])
+      add_index :patient_scores, [:patient_id, :score_type]
+    end
+
+    unless index_exists?(:patient_scores, :updated_at)
+      add_index :patient_scores, :updated_at
+    end
+  end
```
aagrawalrtsl, line 20: Migration should have an up and down method. It is helpful in case we want to revert the migration

app/controllers/api/v4/patient_scores_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+        .where(patient: current_facility.prioritized_patients.select(:id))
+        .order(:updated_at, :id)
+        .limit(limit)
+        .offset((current_page - 1) * limit)
+        .to_a
+  end
+
+  def other_facility_records
+    []
+  end
```
aagrawalrtsl, line 19: Do we need this method

db/structure.sql
```diff
@@ -748,9 +748,7 @@ CREATE TABLE IF NOT EXISTS simple_reporting.reporting_patient_prescriptions (
     hypertension_drug_changed boolean,
     diabetes_drug_changed boolean,
     other_drug_changed boolean,
-    prescribed_statins boolean,
-    latest_cvd_risk_score_lower_range integer,
-    latest_cvd_risk_score_upper_range integer
```
aagrawalrtsl, line 753: why are these changes required ?

aagrawalrtsl (changes requested):

db/data/20260414103256_spread_patient_scores_updated_at.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+class SpreadPatientScoresUpdatedAt < ActiveRecord::Migration[6.1]
+  def up
```
aagrawalrtsl, line 4: Is this migration not supposed to be in schema_migration?

db/structure.sql
sshinde-rtsl: These columns were added in PRs that are already merged to master:

  - prescribed_statins → PR #5776 (Controlled BP card)
  - latest_cvd_risk_score_lower_range, latest_cvd_risk_score_upper_range → PR #5785

  They appear in this PR's diff because of merge commit f25a949d (Merge branch 'master' into SIMPLEBACK-34). These are not changes from this PR.

db/data/20260414103256_spread_patient_scores_updated_at.rb
sshinde-rtsl: Yes, it is in schema_migrations - see line in db/structure.sql:
  ('20260212195326'),

igbanam:
Why do we need a sync route here?

A sync route is for resources collected on the field, through the mobile app, so it can be kept _in sync_ with the server and other mobile devices. The patient score is not collected from the field, and we don't need to keep this in sync with anything since it's generated on the server.

app/controllers/api/v4/patient_scores_controller.rb
sshinde-rtsl: Yes, it's required. The SyncToUser concern (line 36) calls current_facility_records + other_facility_records in records_to_sync. Without this method, sync would fail.
  Returning [] means patient scores only sync for the current facility's patients.

db/migrate/20260209112204_create_patient_scores.rb
sshinde-rtsl: Done. Updated to use up and down methods.

db/structure.sql
aagrawalrtsl: These changes should not be here. The latest code in master adds these columns in reporting_patient_prescriptions table.
With this PR it will be undone, which we don't want.
It should be removed from here

db/data/20260414103256_spread_patient_scores_updated_at.rb
aagrawalrtsl: 20260212195326 refers to db/migrate/20260212195326_unlink_patients_from_cvd_risks.rb which is a completely different PR.
This migration is added in this PR with version 20260414103256 which is not added in schema migration.
My question is do we need this specifically under schema_migration or is it just a one time data fix ?
Also, how are we planning to run this migration in prod. Since it is not under migrate directory it will not be run automatically through rails db:migrate.

aagrawalrtsl:
> Why do we need a sync route here?
> 
> A sync route is for resources collected on the field, through the mobile app, so it can be kept _in sync_ with the server and other mobile devices. The patient score is not collected from the field, and we don't need to keep this in sync with anything since it's generated on the server.

Agreed, if this is generated on the server, instead of sync, we can provide a normal fetch call

db/structure.sql
sshinde-rtsl: Fixed. I've updated db/structure.sql to include the CVD risk columns from master. Now merging this PR will preserve those columns.

sshinde-rtsl:
We have made decision as per the Siddharth's request to make it similar to how sync works. Though this is 1 time activity for the mobile app to pull the facility level data.

igbanam:
The decision to continue with the sync route is because of the setup we have on the mobile side. We've built in benefits on the sync path which is could be needed elsewhere, but would be a lot to refactor. So we're keeping this as a sync route to benefit from the mobile setup.

db/data/20260414103256_spread_patient_scores_updated_at.rb
sshinde-rtsl: Apologies for the confusion with the wrong migration version.
 1. No, it doesn't need to be in schema_migrations. It's a one-time data fix - same pattern as other data migrations in db/data/.
 2. It runs via rails data:migrate using the data_migrate gem.

db/structure.sql
```diff
@@ -9509,6 +9556,14 @@ ALTER TABLE ONLY public.dr_rai_targets
     ADD CONSTRAINT fk_rails_f0398a9ae0 FOREIGN KEY (dr_rai_indicators_id) REFERENCES public.dr_rai_indicators(id);
 
 
+--
+-- Name: patient_attributes fk_rails_fc46ae3757; Type: FK CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.patient_attributes
+    ADD CONSTRAINT fk_rails_fc46ae3757 FOREIGN KEY (patient_id) REFERENCES public.patients(id);
+
```
aagrawalrtsl: This is removed in current master, should not be re added

sshinde-rtsl: Fixed. Removed the fk_rails_fc46ae3757 constraint.

aagrawalrtsl (approved):

PR #5801, opened 2026-04-22
fix(dm-dashboard): statins and BP controlled queries in v14 view

Gyan-Gupta-Rtsl:
**Story card:** [Audit and release DM dashboard cards [Controlled BP & Statins]](https://rtsl.atlassian.net/browse/SIMPLEBACK-68)

## **Because**

Two bugs were found on the Diabetes (DM) dashboard:

1. **Statins card** was excluding DM patients aged 40+ who were registered within the last 3 months, even though the indicator should cover all DM patients over 40 regardless of registration date.
2. **BP Controlled card** was over-counting patients — anyone with a recent visit (e.g. a blood sugar record) but a stale BP measurement from > 3 months ago was being counted as "controlled," because the numerator used manual `months_since_visit` + `systolic/diastolic IS NOT NULL` checks that didn't verify the BP itself was recent.

## **This addresses**

Introduces `reporting_facility_states` view **v14** with two fixes inside the `adjusted_diabetes_outcomes` / `dm_statins_outcomes` CTEs:

1. **Statins card**: removed `AND rps.months_since_registration >= 3` from the `rps_with_age_for_statins` CTE.
2. **BP Controlled card**: the `dm_bp_below_140_90_under_care` numerator now uses `htn_treatment_outcome_in_last_3_months = 'controlled'` (the same pre-calculated flag the HTN dashboard uses), which correctly requires a recent visit *and* a recent BP < 140/90. The `dm_bp_below_130_80_under_care` numerator uses the same `controlled` flag plus an additional `systolic < 130 AND diastolic < 80` filter.

Files:
- `db/views/reporting_facility_states_v14.sql` (new)
- `db/migrate/20260422120000_update_reporting_facility_states_to_version14.rb` (new)
- `db/structure.sql` (updated)

## **Test instructions**

1. Run `rails db:migrate` and confirm the view rebuilds cleanly and `db/structure.sql` matches the committed diff.
2. Refresh the reporting views and load the Diabetes dashboard for a region.
3. Add filpper flag: **diabetes_cvd_indicators**
4. **Statins card**: verify DM patients 40+ registered within the last 3 months now appear in the denominator.
5. **BP Controlled card**: verify a DM patient whose only recent encounter in the last 3 months is a blood sugar (no BP in last 3 months, even if an older BP exists) is **not** counted in the <140/90 numerator.
6. Spot-check the <130/80 numerator is a subset of the <140/90 numerator.
7. Run existing specs:
   - `bundle exec rspec spec/models/reports/region_summary_schema_spec.rb`
   - `bundle exec rspec spec/models/reports/repository_spec.rb`

### DIFF
--- a/db/migrate/20260422120000_update_reporting_facility_states_to_version14.rb
+++ b/db/migrate/20260422120000_update_reporting_facility_states_to_version14.rb
@@ -1,0 +1,27 @@
+class UpdateReportingFacilityStatesToVersion14 < ActiveRecord::Migration[6.1]
+  def up
+    drop_view :reporting_facility_states, materialized: true
+
+    execute <<-SQL
+      CREATE MATERIALIZED VIEW public.reporting_facility_states AS
+      #{File.read(Rails.root.join("db/views/reporting_facility_states_v14.sql"))}
+      WITH NO DATA;
+    SQL
+
+    add_index :reporting_facility_states, [:month_date, :facility_region_id], name: :index_fs_month_date_region_id, unique: true
+    add_index :reporting_facility_states, [:block_region_id, :month_date], name: :index_fs_block_month_date
+    add_index :reporting_facility_states, [:district_region_id, :month_date], name: :index_fs_district_month_date
+    add_index :reporting_facility_states, [:state_region_id, :month_date], name: :index_fs_state_month_date
+    add_index :reporting_facility_states, [:organization_region_id, :month_date], name: :index_fs_organization_month_date
+  end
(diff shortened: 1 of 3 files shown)


aagrawalrtsl (approved):
Looks ok to me.
Please make sure that after merging this, we deploy this to all the environments, 
and then run a manual REFRESH for reporting_facility_states everywhere as well.
If it's deployed and a refresh is not done, the dashboard doesn't load

PR #5803, opened 2026-05-05
fix: Make CreatePatientScores migration idempotent

sshinde-rtsl:
Prevent migration failure when patient_scores table already exists by checking table/index existence before creating them.

**Story card:** [SIMPLEBACK-111](https://rtsl.atlassian.net/browse/SIMPLEBACK-111)

## Because

The migration `20260209112204_create_patient_scores` fails with `PG::DuplicateTable: ERROR: relation "patient_scores" already exists` when the table already exists in the database but the migration hasn't been recorded in `schema_migrations`.

## This addresses

 - Makes the `CreatePatientScores` migration idempotent
  - Adds `table_exists?` check before creating the table
  - Adds `index_exists?` checks before creating indexes
  - Adds `if_exists: true` to `drop_table` in rollback

## Test instructions

 1. Ensure `patient_scores` table exists in your database
  2. Remove the migration record: `DELETE FROM schema_migrations WHERE version = '20260209112204';`
  3. Run `rake db:migrate`
  4. Verify migration completes without error
  5. Verify existing data in `patient_scores` is preserved

### DIFF
--- a/db/migrate/20260209112204_create_patient_scores.rb
+++ b/db/migrate/20260209112204_create_patient_scores.rb
@@ -1,21 +1,23 @@
 class CreatePatientScores < ActiveRecord::Migration[6.1]
   def up
-    create_table :patient_scores, id: :uuid do |t|
-      t.references :patient, null: false, foreign_key: true, type: :uuid
-      t.string :score_type, null: false, limit: 100
-      t.decimal :score_value, precision: 5, scale: 2, null: false
-      t.datetime :device_created_at, null: false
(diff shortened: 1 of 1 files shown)


Gyan-Gupta-Rtsl (commented):
LGTM but ensure on BD environment, where patient_scores exists with an incomplete schema (e.g., missing deleted_at), the table must be dropped manually before running this migration. The table_exists? guard will skip create_table, preserving the broken schema.

Gyan-Gupta-Rtsl (approved):

PR #5815, opened 2026-06-09
For Ethiopia change the wording to visits rather than patients

jamiecarter7:
Affects:
- Comparison table title and tooltip at all levels
- Footer notes at all levels

**Story card:** [SIMPLEETH-26](https://rtsl.atlassian.net/browse/SIMPLEETH-26?search_id=ad88c9f4-2f56-4432-a83d-7c47199b3005&referrer=quick-find)

## Because

ET request the wording of the title is changed from Follow-up patients to Follow-up visits. The word patients to visits but leaving existing wording intact.

## This addresses

An issue in ET with the report being questioned.

## Test instructions

Example:

reg-fol
reg-fol-2

- Ensure your instance is set to 'Ethiopia'
- Navigate to BOTH HTN and DM pages at all levels
Check the following 3 items:
- Table column title is updated to state 'visits'
- Table tooltip (item 2) states 'visits'
- Footer details shows 'visits'
**NOTE: At facility level the wording is 'Follow-up visits per user' but at all other levels its 'Follow-up visits'**

reg-fol-footer
reg-fol-tooltip
This tooltip is just an example so you know what to look for, but it should say 'Follow-up visits'.

### DIFF
--- a/app/components/dashboard/diabetes/registrations_and_follow_ups_by_user_component.html.erb
+++ b/app/components/dashboard/diabetes/registrations_and_follow_ups_by_user_component.html.erb
@@ -1,9 +1,8 @@
-<% follow_up_label = "Follow-up #{CountryConfig.current_country?("Ethiopia") ? "visits" : "patients"} per user" %>
 <div class="card mt-0 pr-0 pr-md-3 pb-inside-avoid">
   <%= render Dashboard::Card::TitleComponent.new(title: "Patient registrations and follow-ups") do |c| %>
     <% c.tooltip(
          { "Monthly registered patients" => t("registered_diabetes_patients_copy.monthly_registered_patients", region_name: region.name),
-           follow_up_label => t(:diabetes_follow_up_patients_copy, region_name: region.name) }) %>
+           t(:follow_up_patients_per_user_label) => t(:diabetes_follow_up_patients_copy, region_name: region.name) }) %>
   <% end %>
 
   <div class="table-responsive-md">
@@ -29,7 +28,7 @@
 
       <%= table.header("Total") %>
       <%= table.header("Monthly registered patients", colspan: 6) %>
-      <%= table.header(follow_up_label, colspan: 6) %>
+      <%= table.header(t(:follow_up_patients_per_user_label), colspan: 6) %>
 
       <%= table.sub_header("Users", sort_default: true, sort_method: :string) %>
       <%= table.sub_header("Registered patients") %>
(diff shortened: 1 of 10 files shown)


app/components/dashboard/diabetes/registrations_and_follow_ups_by_user_component.html.erb
```diff
@@ -1,8 +1,9 @@
+<% follow_up_label = "Follow-up #{CountryConfig.current_country?("Ethiopia") ? "visits" : "patients"} per user" %>
```
aagrawalrtsl: Instead of checking for country config every where, we can put the text in the locale files.
In this case it would be en_ET.yml
We can use the key that is present in the locale file. The same key would be added in the default en.yml file, so that for countries other than ET, it will show the default text.
This makes it easier for code maintenance because we are not required to do the Country config change everywhere.

aagrawalrtsl (changes requested):

aagrawalrtsl (commented):
Not a blocker - but would be great if we could add some specs around this change.

jamiecarter7:
How do I add specs?

app/components/dashboard/diabetes/registrations_and_follow_ups_by_user_component.html.erb
jamiecarter7: I agree, on doing further work for an LK issue I think your suggestion is the best way as its the only in the en-LK.yaml file

jamiecarter7: To clarify, are these files translated?
It seems like a waste as the dashboard is not translated, but these files possibly are?

aagrawalrtsl: I think they are. For Ethiopia the dashboard locale is set to en-ET, so anything that is present under en_Et.yml file should show up in the dashboard
We can also verify this locally by setting the dashboard locale to en-ET and then check if the required translations are visible or not

aagrawalrtsl:
> How do I add specs?

So we can add a spec file under spec folder, with the corresponding path of the original file
As a reference see this 
https://github.com/simpledotorg/simple-server/blob/master/spec/views/reports/regions/_diabetes_footnotes.html.erb_spec.rb
We can also use any AI tool for this as well.
But if it takes too much time/effort we can ignore for now

jamiecarter7:
I've pushed the suggested changes, minus the spec file.

aagrawalrtsl (approved):

jamiecarter7:
@aagrawalrtsl Are you able to check the reason the tests are failing?
Thank you 🙇

aagrawalrtsl:
> @aagrawalrtsl Are you able to check the reason the tests are failing? Thank you 🙇

It was just a flaky test - I have rerun the tests now. It's all good

jamiecarter7:
Thank you!

PR #5817, opened 2026-06-10
DM Dashboard treatment status chart fixes - WHO standards

jamiecarter7:
- Standardised language `controlled` and `uncontrolled`, removing `<126` and `≥126` bringing into line with HEARTS360
- Removed LK-specific language; this is now the default language for any instance with this flipper flag enabled

**Story card:** [SIMPLEBACK-125](https://rtsl.atlassian.net/browse/SIMPLEBACK-125)

## Because

The 'Treatment status' card (containing a bar chart) was incorrectly showing wording from the non-WHO standards when the WHO dm indicator standards flipper flag was enabled.
Flag: `diabetes_who_standard_indicator`

This section of the website did not contain any flipper flags for this feature.

## This addresses

Issues reported by the Sri Lanka team.

Core fixes:
- Standard format and wording implemented for all countries
- Enabled by flipper flag
- Removed county-specific language (currently only used in LK)

## Test instructions

- Familiarise yourself with the DM dashboard wording, focusing on the 2 charts (controlled and uncontrolled) and the treatment status for patients under care card and also the footer definitions (titles).
- Understand that the base setup allows for RBS and PPBS to be included in the indicators, while the next WHO standard removes RBS and PPBS from indicators (leaving only FBS and HbA1c)
- Enable the `diabetes_who_standard_indicator` flipper flag
- Test all 3 cards and footer show expected wording and title, also check tool tips match expected copy.
- Test the dashboard on the LK instance to ensure it's identical to other country instances.


aagrawalrtsl (commented):
Looks ok to me.
But the same changes needs to be done for specs as well. It's failing because it is expecting the previous texts.

PR #5819, opened 2026-06-18
feat(Inactive Facilities): Add breadcrumbs

igbanam:
**Story card:** [SIMPLEETH-39](https://rtsl.atlassian.net/browse/SIMPLEETH-39)

## Because

Inactive Facilities are hard to identify per region.

## This addresses

This helps with easier identification on facilities with their blocks and districts.

## Test instructions

UI confirmation


aagrawalrtsl (approved):
Not blocker, but would be good if we could add some specs around it

PR #5849, opened 2026-09-21
Replace deprecated dhis2 gem with REST API client

sshinde-rtsl:
Problem:
- DHIS2ExporterJob fails even when data import is successful
- Deprecated dhis2 gem raises BulkCreationError on success
- Infinite Sidekiq retry loops cause queue saturation
- Critical jobs blocked, dashboards don't load on 1st of month
- Only works with DHIS2 API v2.28 (requires version hack)

Solution:
- Create Dhis2HttpClient using Net::HTTP for direct REST API calls
- POST to /api/dataValueSets endpoint (official DHIS2 REST API)
- Correctly parse success responses (handles modern & legacy formats)
- Transform snake_case keys to camelCase (DHIS2 API requirement)
- Remove dependency on deprecated dhis2 gem from Gemfile
- Remove DHIS2_VERSION environment variable hack
- Remove Ethiopia-specific error handling workaround
- Update all exporter jobs to use new HTTP client
- Add comprehensive RSpec tests (263 lines, 15+ scenarios)

Changes:
- New: app/services/dhis2_http_client.rb (HTTP client)
- New: spec/services/dhis2_http_client_spec.rb (tests)
- Modified: app/jobs/dhis2/dhis2_exporter_job.rb (use new client)
- Modified: app/jobs/dhis2/ethiopia_exporter_job.rb (remove workaround)
- Modified: Gemfile (comment out dhis2 gem)
- Modified: .env.* files (remove DHIS2_VERSION)
- Modified: All job specs (mock new client)

Impact:
- Jobs complete successfully when data imported (no false errors)
- Sidekiq only retries actual failures (no infinite loops)
- Queue stays healthy (no saturation)
- Dashboards load immediately (no delays)
- Works with all DHIS2 API versions (future-proof)

Fixes queue saturation and dashboard loading issues.

## Because

## This addresses

## Test instructions


app/jobs/dhis2/ethiopia_exporter_job.rb
```diff
@@ -87,14 +87,6 @@ def format_cohort_registered_data(registered_patients)
...
-    # DHIS2 library we use works for old versions of DHIS2. Since Ethiopia is on much advanced version, this is a hack to prevent
-    # false errors being raised from the client.
-    def export(data_values)
-      @client.data_value_sets.bulk_create(data_values: data_values)
-    rescue Dhis2::BulkCreationError => error
-      response = JSON.parse(error.import_summary.to_json)["hash"]
-      raise error if response["http_status_code"] != 200
-      Rails.logger.info("Exported to Dhis2 with response: ", response)
-    end
+    # No need for the export override anymore - the new HTTP client handles modern DHIS2 API responses correctly
```
aagrawalrtsl: We don't need this message in the code, since the code is already removed. 
We should test this once with the version Ethiopia is using, just to be sure that it won't fail when it goes to production

Gemfile
```diff
@@ -25,7 +25,7 @@ gem "data_migrate"
 gem "data-anonymization", require: false
 gem "devise_invitable", "~> 2.0.6"
 gem "devise", ">= 4.7.1"
-gem "dhis2", require: false
+# gem "dhis2", require: false # Removed - replaced with Dhis2HttpClient using direct REST API calls
```
aagrawalrtsl: If we don't need it, I would prefer just removing rather than commenting out.
Also, we need to remove this from Gemfile.lock as well.

aagrawalrtsl (approved):

